PR #664, opened 2017-01-26
Video streaming improvments

AlexeyBulatov:
This PR continues work on messages for video streaming.
It adds following:
1. 2 new parameters for video: ```bit rate``` and ```video image rotation```
2. new parameter ```video format ID```
3. messages for setting parameters of video streaming target address ```VIDEO_STREAM_TARGET```
4. command for request this options ```MAV_CMD_REQUEST_VIDEO_STREAM_TARGET```
5. fix some typos

@LorenzMeier @julianoes please review

PR #677, opened 2017-02-27
Video streaming improvements v2

AlexeyBulatov:
v1 - https://github.com/mavlink/mavlink/pull/664
Changes:
1. Use separate commands to start/stop video streaming.
2. Use URL instead if IP and port.
3. Add a new capability for video streaming to MAV_PROTOCOL_CAPABILITY bitmask.

Unfortunately, I don’t see a clear and concise way to overcome the issue of dynamically requested camera capabilities. One of the options might be to add new MAVLink messages that request list for every capability. For example:
Command MAV_CMD_REQUEST_CAMERA_RESOLUTIONS requests mavlink message CAMERA_RESOLUTIONS which contains array of C strings with all possible values (i.e. [640x480 1280x720]).

@LorenzMeier @julianoes @dogmaphobic please review

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1422,6 +1422,16 @@
         <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
         <param index="2">Reserved</param>
       </entry>
+      <entry value="2502" name="MAV_CMD_VIDEO_START_STREAMING">
+        <description>WIP: Start video streaming</description>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="2">Reserved</param>
+      </entry>
+      <entry value="2503" name="MAV_CMD_VIDEO_STOP_STREAMING">
+        <description>WIP: Stop the current video streaming</description>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="2">Reserved</param>
+      </entry>
       <entry value="2510" name="MAV_CMD_LOGGING_START">
         <description>Request to start streaming logging data over MAVLink (see also LOGGING_DATA message)</description>
         <param index="1">Format: 0: ULog</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1432,6 +1432,12 @@
         <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
         <param index="2">Reserved</param>
       </entry>
+      <entry value="2504" name="MAV_CMD_REQUEST_VIDEO_STREAM_INFORMATION">
+        <description>WIP: Request video stream information (VIDEO_STREAM_INFORMATION)</description>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="2">0: No Action 1: Request video stream information</param>
+        <param index="3">Reserved (all remaining params)</param>
+      </entry>
       <entry value="2510" name="MAV_CMD_LOGGING_START">
         <description>Request to start streaming logging data over MAVLink (see also LOGGING_DATA message)</description>
         <param index="1">Format: 0: ULog</param>
@@ -4039,6 +4045,29 @@
       <field type="uint8_t" name="target_system">system ID of the target</field>
       <field type="uint8_t" name="target_component">component ID of the target</field>
       <field type="uint16_t" name="sequence">sequence number (must match the one in LOGGING_DATA_ACKED)</field>
+    </message>
+    <message id="269" name="VIDEO_STREAM_INFORMATION">
+      <description>WIP: Information about video stream</description>
+      <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="status">Current status of video streaming (0: not running, 1: in progress)</field>
+      <field type="float" name="framerate" units="Hz">Frames per second</field>
+      <field type="uint16_t" name="resolution_h" units="pix">Resolution horizontal in pixels</field>
+      <field type="uint16_t" name="resolution_v" units="pix">Resolution vertical in pixels</field>
+      <field type="uint32_t" name="bitrate" units="b/s">Bit rate in bits per second</field>
+      <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise</field>
+      <field type="char[230]" name="uri">Video stream URI</field>
+    </message>
+    <message id="270" name="SET_VIDEO_STREAM_SETTINGS">
+      <description>WIP: Message that sets video stream settings</description>
+      <field type="uint8_t" name="target_system">system ID of the target</field>
+      <field type="uint8_t" name="target_component">component ID of the target</field>
+      <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
+      <field type="float" name="framerate" units="Hz">Frames per second (set to -1 for highest framerate possible)</field>
+      <field type="uint16_t" name="resolution_h" units="pix">Resolution horizontal in pixels (set to -1 for highest resolution possible)</field>
+      <field type="uint16_t" name="resolution_v" units="pix">Resolution vertical in pixels (set to -1 for highest resolution possible)</field>
+      <field type="uint32_t" name="bitrate" units="b/s">Bit rate in bits per second (set to -1 for auto)</field>
+      <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise (0-359 degrees)</field>
+      <field type="char[230]" name="uri">Video stream URI</field>
     </message>
   </messages>
 </mavlink>


dogmaphobic:
Overall it all looks fine.

There is one instance where things need to be explained a bit somewhere. When using RTP/UDP, everything as is will work fine and it's pretty straight forward. However for RTSP, things are a bit different.

For RTSP, you would send a `MAV_CMD_REQUEST_VIDEO_STREAM_TARGET` and wait for the target URL. Once the client has that, it will decide when it wants to start and/or stop the streaming. It starts the streaming by making the `rtsp://<target>` connection and ends it by closing that connection. In theory, it would never need to send a `MAV_CMD_VIDEO_START_STREAMING` command. However, there is no way to tell the camera what resolution and frame rate to use.

We have two options: Create a new message that tells the camera what resolution and frame rate to use or use the existing `MAV_CMD_VIDEO_START_STREAMING` to do so, but the camera would ignore the "START" part of it as the client is responsible for starting it. An argument could be made to keep `MAV_CMD_VIDEO_START_STREAMING` alone and not create a new message. Cameras that have the RTSP server running all the time would simply ignore the "START" part (end the entire `MAV_CMD_VIDEO_STOP_STREAMING` message). However, if the camera has the option of starting/stoping the RTSP server, then these message become relevant. With that said, I'm not aware of any system that has RTSP start/stop on demand.

>Unfortunately, I don’t see a clear and concise way to overcome the issue of dynamically requested camera capabilities. One of the options might be to add new MAVLink messages that request list for every capability.

Yes, this is an issue here and for other areas as well. Whenever you have a pre-defined set of options for something, it's not easy or practical to use MAVLink messages. In this case, we are talking about only two options: **Resolution** and **Frame Rate**. For video/photo capture however, we have a dozen or so similar options with their own set of possible values. That would make it impractical to have a MAVLink message to get a list for each and everyone of them.

Video resolution could be just about anything just as frame rates. It's not uncommon to have 1, 2, 5, 10, 12, 15, 20, 24, 25, 30, 50, 60, 90, 120 for frame rate alone.

I'm not sure what the best solution for this would be either. Request a Json file?

AlexeyBulatov:
> For RTSP, you would send a MAV_CMD_REQUEST_VIDEO_STREAM_TARGET and wait for the target URL. Once the client has that, it will decide when it wants to start and/or stop the streaming. It starts the streaming by making the rtsp://<target> connection and ends it by closing that connection. In theory, it would never need to send a MAV_CMD_VIDEO_START_STREAMING command. However, there is no way to tell the camera what resolution and frame rate to use.

We can send a start/stop condition along with other settings (resolution, fps). Then in case you use RTSP this value will be false and GCS will render the appropriate GUI, i.e. "Change settings" instead of "Start/Stop" button. 

> I'm not sure what the best solution for this would be either. Request a Json file?

I think it is rather convenient to use files for that goal (xml, json or another format). 
For example:
{
  "resolution": [...],
  "fps": [...]
}
And there are two ways how to get this file:
1. Store it in GCS and have list of cameras.
2. GCS will request this file and receive using messages like for request log (LOG_REQUEST_DATA).

AlexeyBulatov:
Guys, it'd be awesome to hear some feedback on the topic!

dogmaphobic:
I guess the only pending issue here is in regard to requesting a list of options from the camera (resolution and frame rate). That can be decided on a different PR. As far as handling the basics, this PR should be fine as is.

There is a conflict within CAMERA_CAPTURE_STATUS, which I don't think belong to "streaming" any way. That message is for taking photos and/or recording video. It's not meant for video streaming. Once that is resolved, I think we can merge this. Opinions?

otaviobp:
This series looks good for me. I just have one question and one suggestion.

I am working on a Daemon to run on a drone companion board to stream multiple video cameras, using mainly RTSP. (https://github.com/01org/camera-streaming-daemon). It is not using MAVLink for communication, but I am working right now to add this support.
And I am planning to use the messages your are proposing to do that.

I saw that in your qgroundcontrol pull request you are always using 1 for Camera ID. If more than one camera are installed on the drone, how are you intending to get all camera IDs? Using MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS with camera_id = 0 or adding a new message to get all available cameras?

And why not setting the target URI in the MAV_CMD_VIDEO_START_STREAMING message? If you do that, you could start the RTSP server when VIDEO_START is received and stop the server when VIDEO_STOP is received. The same for UDP video. 
And you could keep the GET_TARGET, to return current video target.

dogmaphobic:
>I saw that in your qgroundcontrol pull request you are always using 1 for Camera ID. If more than one camera are installed on the drone, how are you intending to get all camera IDs? Using MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS with camera_id = 0 or adding a new message to get all available cameras?

This and many, many others like it, including what is being discussed earlier in this PR will need something else. This is becoming more and more of an issue that cannot be resolved simply by MAVLink messages alone. That MAV_PROTOCOL_CAPABILITY flag will get pretty crowded quite soon. We need a message that, if the vehicle supports it, it would return a full list of capabilities, settings, and settings options. It could return an URL so we could `GET` it, or some other method of returning it. It should be a file (Json, XML, whatever is decided) with a standard schema. It's the only way to have all this stuff defined in a sane manner. It could even be an URL *served* from some place on the web, not necessarily on the vehicle itself in case of mass produced vehicles, though by the time you have all these features on a vehicle, you ought to have a companion computer and this computer should be able to *serve* these files easily.

>And why not setting the target URI in the MAV_CMD_VIDEO_START_STREAMING message? 

That's fine for UDP, where the URL is on the client side but not for RTSP. More often then not, the URL is fixed on the server side. You don't *set* it. Most systems start the RTSP server when they boot and have the URL already defined.

AlexeyBulatov:
Thanks for reply @dogmaphobic @otaviobp.
It seems that it is necessary to supplement the protocol, so I want to add some more messages for better setup:
First of all if CAMERA_CAPTURE_STATUS is used for recording photo/video it will be better to create new message VIDEO_STREAM_STATUS. Where will be the following fields:
camera_id, stream_status, fps, resolution_h, resolution_v, bitrate, rotation, streaming_time.
Also I don't see any message which can tell how many cameras are available at the vehicle.
For getting list of options from the camera I suggest to use messages like for downloading logs. Options will be stored in json file. So I want add the following messages:
CAMERA_OPTIONS_REQUEST_LIST (like LOG_REQUEST_LIST) with fields:
target_system, target_component, start, end (from here can get number of cameras).
On this request will be send messages CAMERA_OPTIONS_ENTRY (like LOG_ENTRY) with the following fields:
camera_id, num_cameras, size (size of file with options).
For getting this file will be used messages CAMERA_OPTIONS_REQUEST_DATA and CAMERA_OPTIONS_DATA similarly LOG_REQUEST_DATA and LOG_DATA.
What do you think about this?

dogmaphobic:
> First of all if CAMERA_CAPTURE_STATUS is used for recording photo/video it will be better to create new message VIDEO_STREAM_STATUS. Where will be the following fields:
> camera_id, stream_status, fps, resolution_h, resolution_v, bitrate, rotation, streaming_time.

I'm not opposed to the creation of this message but what would it tell you that you don't already know?

> So I want add the following messages:
> CAMERA_OPTIONS_REQUEST_LIST (like LOG_REQUEST_LIST) with fields:

That's an overly complex, not to mention inefficient protocol. If MAVLink is the only way, yes, we would need something like this but why not just get an URL and use HTTP to get the file? It's not like we're talking to a resource constrained flight controller. By the time we need this, there ought to be a separate computer handling this. If you can *serve* a video stream, serving files should be trivial. Implementing a protocol similar to log download would be orders of magnitude more complex and error prone.

otaviobp:
> That's an overly complex, not to mention inefficient protocol. If MAVLink is the only way, yes, we would need something like this but why not just get an URL and use HTTP to get the file? 
I am ok with that for extra information, but it would be good to have a basic camera support using MAVLink.

otaviobp:
Looking at the messages that are already merged in mavlink and the new messages proposed in this pull request I have a suggestion of using the following flow of communication to use UDP and RTSP video streaming.
Any other protocol that that works in a way simillar to UDP (drone sends video to QGC) or to RTSP(QGC do a request for the video) would also be supported by that flow.

MAVLink communication for video Streaming using UDP

 - QGC sends a MAV_CMD_REQUEST_CAMERA_INFORMATION message with camera id == 0.
 - Drone send one CAMERA_INFORMATION response for each camera available. It would be nice to have a field with a list of supported protocols for streaming (UDP or RTSP).
 - QGC sends a MAV_CMD_VIDEO_START_STREAMING with the camera ID, resolution and other settings and target_URI
 - Drone streams video to the QGC
 - QGC sends a MAV_CMD_VIDEO_STOP_STREAMING
 - Drone stops the video stream

MAVLink communication for video Streaming using RTSP

 - QGC sends a MAV_CMD_REQUEST_CAMERA_INFORMATION message with camera id == 0.
 - Drone send one CAMERA_INFORMATION response for each camera available. It would be nice to have a field with a list of supported protocols for streaming (UDP or RTSP).
 - QGC sends a MAV_CMD_REQUEST_VIDEO_STREAM_TARGET
 - Drone sends the a VIDEO_STREAM_TARGET message with the RTSP URI of the video stream (rtsp://ip:port/resource)
 - QGC makes a rtsp request to the RTSP URI, with all needed settings as a query parameter.
(rtsp://ip:port/resource?resolution=640x480&framerate=30fps)
 - QGC close the rtsp connection to stop the video stream

Using this flow we could have multiple cameras basic support without needing a json file. We would still need it to retrieve extra information about the cameras (like all supported resolutions, framerate, etc).

Does it makes sense. Is there a scenario that this flow is not covering?
@AlexeyBulatov  Is this covering the use cases you need to support?

AlexeyBulatov:
> > First of all if CAMERA_CAPTURE_STATUS is used for recording photo/video it will be better to create new message VIDEO_STREAM_STATUS. Where will be the following fields:
> > camera_id, stream_status, fps, resolution_h, resolution_v, bitrate, rotation, streaming_time.
> 
> I'm not opposed to the creation of this message but what would it tell you that you don't already know?

As you said earlier we can't use `CAMERA_CAPTURE_STATUS` because this message for taking photos and recording video. So we need replacement, which can tell us the state of video streaming: id of camera, video streaming state (running or not), type of it (udp, rtsp, rtp, ...), settings of quality (fps, resolution, etc.). We may not know this parameters after connection to vehicle, so this message help to get them.

> That's an overly complex, not to mention inefficient protocol. If MAVLink is the only way, yes, we would need something like this but why not just get an URL and use HTTP to get the file? It's not like we're talking to a resource constrained flight controller. By the time we need this, there ought to be a separate computer handling this. If you can serve a video stream, serving files should be trivial. Implementing a protocol similar to log download would be orders of magnitude more complex and error prone.

Yes, it's more easy to get this file using URL. But this parameters also may be helpful for video recording. And from this point of view we will have problems with radio link. But if we don't need this, it will be better to use your solution.

Thanks @otaviobp. I like your proposed method of start/stop streaming. So using `REQUEST_CAMERA_INFORMATION` message for getting number and ids of cameras is more preferable.

dogmaphobic:
> As you said earlier we can't use CAMERA_CAPTURE_STATUS because this message for taking photos and recording video. So we need replacement, which can tell us the state of video streaming: id of camera, video streaming state (running or not), type of it (udp, rtsp, rtp, ...), settings of quality (fps, resolution, etc.). We may not know this parameters after connection to vehicle, so this message help to get them.

I guess a bit of redundancy doesn't hurt. Unlike video/image capture, you are the receiving end of video streaming. As such, you know if you are receiving the stream or not, if it's UDP or RTSP, fps, etc. This information is part of the stream itself.

Do we have an agreement on the overall spec? Can you update it with what we have so far?

dogmaphobic:
> QGC makes a rtsp request to the RTSP URI, with all needed settings as a query parameter.
> (rtsp://ip:port/resource?resolution=640x480&framerate=30fps)

Just noticed this. It won't work for most cameras that do the streaming themselves. Something like the GoPro (or anything using the Ambarella SOC). The URL is fixed and it's only for the stream. You must define resolution and FPS through other means (different commands).

otaviobp:
> Just noticed this. It won't work for most cameras that do the streaming themselves. Something like the GoPro (or anything using the Ambarella SOC). The URL is fixed and it's only for the stream. You must define resolution and FPS through other means (different commands).

Good point. We can use MAV_CMD_VIDEO_START_STREAMING with target_uri as an opitional parameter. In RTSP we can ignore it. And we could make VIDEO_STREAM_TARGET an optional response for the MAV_CMD_VIDEO_START_STREAMING, only for rtsp.

This way we would have:

MAVLink communication for video Streaming using UDP
 - same as previous post

MAVLink communication for video Streaming using RTSP

 - QGC sends a MAV_CMD_REQUEST_CAMERA_INFORMATION message with camera id == 0.
 - Drone send one CAMERA_INFORMATION response for each camera available. It would be nice to have a field with a list of supported protocols for streaming (UDP or RTSP).
 - QGC sends a MAV_CMD_VIDEO_START_STREAMING with streaming parameters and empty target_uri
 - Drone configures camera with parameters from MAV_CMD_VIDEO_START_STREAMING and sends a VIDEO_STREAM_TARGET message with the RTSP URI of the video stream (rtsp://ip:port/resource)
 - QGC makes a rtsp request to the RTSP URI informed in VIDEO_STREAM_TARGET
 - If MAV_CMD_VIDEO_STOP_STREAMING is received and the video is streaming, server closes the connection.
 - QGC can close the rtsp connection to stop the video stream

AlexeyBulatov:
So, I have updated PR.
Support for capabilities of video streaming can be added with the next PR.
Now this PR will add basic support for video streaming:
1. Set up target by URL.
2. Start/stop. Start command also have settings of video streaming (fps, resolution, etc.)

otaviobp:
I still have one concert about this proposal. How can a client knows if the streaming is going to use udp or rtsp?
If I understood your patches correctly, for UDP, the client would send a SET_VIDEO_STREAM_TARGET and then a MAV_CMD_VIDEO_START_STREAMING.
For RTSP the client would need to send a SET_VIDEO_STREAM_GET_TARGET and a MAV_CMD_VIDEO_START_STREAMING (to set parameters).
But how can a client know what is the method supported by that camera? And what if both methods are supported? Is there a way for the client to choose what will be used?

otaviobp:
I have a proposal to make both protocols supported with just a few changes in your proposed messages. I have pushed them to: https://github.com/otaviobp/mavlink/tree/video_stream

Instead of having a get/set target, I am using a VIDEO_STREAM_INFO to get the protocol and URI(if rtsp) from drone and SET_VIDEO_STREAM_PARAMS to set video streaming parameters as long as the target_uri (for udp).

So, first we could use MAV_CMD_REQUEST_VIDEO_STREAM_INFO to check if the camera is using rtsp or udp.

IF rtsp we would use SET_VIDEO_STREAM_PARAMS to set video parameters and the uri from MAV_CMD_REQUEST_VIDEO_STREAM_INFO to do the rtsp request.

IF UDP, we would use SET_VIDEO_STREAM_PARAMS to set video parameters and target_uri. To control the video we would use MAV_CMD_VIDEO_START_STREAMING and MAV_CMD_VIDEO_START_STREAMING.

What do you think about that? Does this cover your use cases?

AlexeyBulatov:
Yeah, we can also add parameter for the stream type. But I think it will be better to use ENUM for this (1-udp, 2-rtsp, ...). Or there's another way - send it type in URL, something like that: `udp:127.0.0.1:1234`.

Move parameters from command to mav message allows to use more then 6 params (max number of command's fields is 7) and it's good. Merging of params and url to one message will reduce length of URL field, but it will still be enough.

otaviobp:
> Yeah, we can also add parameter for the stream type. But I think it will be better to use ENUM for this (1-udp, 2-rtsp, ...). Or there's another way - send it type in URL, something like that: udp:127.0.0.1:1234.

I think we can use the URL field for that. The only problem is that we don't have a real URI for udp (when server is telling client that it is using udp), but we could have a uri like "udp::" or use a default uri like you said "udp:127.0.0.1:5600".

> Move parameters from command to mav message allows to use more then 6 params (max number of command's fields is 7) and it's good. Merging of params and url to one message will reduce length of URL field, but it will still be enough.

I haven't think about that. What is the maximum size of a MAVLink packet? I was expecting to fit a url with the same size it was proposed when using a message with just the URI.

Have we agreed on that protocol?  Can we work on merging it?

AlexeyBulatov:
Okay, I have updated PR and marked new messages as WIP.
@dogmaphobic Could you please take a look?

otaviobp:
lgtm :)

dogmaphobic:
It looks fine to me as long as you know what the options are, which is fine for the time being. Getting the options is something I still don't know how to deal with and it's not just a problem here but also elsewhere.

What I mean, when "setting" fields like *framerate*, *resolution*, etc. you (the GSC) don't (normally)  know what the possible options are.

otaviobp:
> It looks fine to me as long as you know what the options are, which is fine for the time being. Getting the options is something I still don't know how to deal with and it's not just a problem here but also elsewhere.

I would merge this patch and start another discussion on how to get all possible camera settings. So far we could follow the approach of the GSC sending for a desired resolution/fps and the drone streaming the video in the closest resolution/framerate possible.

AlexeyBulatov:
@dogmaphobic Could this be merged?
Messages for getting camera's options will be added in the next PR.

PR #626, opened 2016-09-27
GeoFence/Rally Point: Initial proposal [do not merge]

DonLakeFlyer:
Think of this as a conceptual proposal for #617 as opposed to something that is completely worked out. The main concepts are:
- Fence and Rally support are a 2.0 extension of the existing mission protocol.
- Vehicle support is indicated by capability bits.
- Mission/Fence/Rally have their own separate command space. You can send/request/clear/update them independantly. Within that command space you use the normal mission send/request protocol. You just need to specify the "type" you are working with. So for example to get rally points, you do a MISSION_REQUEST_LIST with mission_type = MAV_MISSION_TYPE_RALLY.

There are many other things to discuss (actual commands supported, what about the things that are in parameters currently, ...), but we need to get this high level discussion done first.

PR #681, opened 2017-03-16
Definitions for new GeoFence, Rally Point spec

DonLakeFlyer:
This should give us enough to get started with new support. PX4 will implement and work through issues.

Discussion is here: #626

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1445,6 +1445,39 @@
         <param index="5">Unscaled target latitude of center of circle in CIRCLE_MODE</param>
         <param index="6">Unscaled target longitude of center of circle in CIRCLE_MODE</param>
       </entry>
+      <entry value="5000" name="MAV_CMD_FENCE_RETURN_POINT">
+          <description>Fence return point. There can only be one fence return point.
+          </description>
+          <param index="1">Reserved</param>
+          <param index="2">Reserved</param>
+          <param index="3">Reserved</param>
+          <param index="4">Reserved</param>
+          <param index="5">Latitude</param>
+          <param index="6">Longitude</param>
+          <param index="7">Altitude</param>
+      </entry>
+      <entry value="5001" name="MAV_CMD_FENCE_POLYGON_VERTEX">
+          <description>Fence polygon vertex. Minimum of 3 required.
+          </description>
+          <param index="1">Polygon vertex count</param>
+          <param index="2">Reserved</param>
+          <param index="3">Reserved</param>
+          <param index="4">Reserved</param>
+          <param index="5">Latitude</param>
+          <param index="6">Longitude</param>
+          <param index="7">Reserved</param>
+      </entry>      
+      <entry value="5100" name="MAV_CMD_RALLY_POINT">
+          <description>Rally point. You can have multiple rally points defined.
+          </description>
+          <param index="1">Reserved</param>
+          <param index="2">Reserved</param>
+          <param index="3">Reserved</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2624,12 +2682,16 @@
       <field type="float" name="x">PARAM5 / local: x position, global: latitude</field>
       <field type="float" name="y">PARAM6 / y position: global: longitude</field>
       <field type="float" name="z">PARAM7 / z position: global: altitude (relative or absolute, depending on frame.</field>
+      <extensions/>
+      <field type="uint8_t" name="mission_type">Mission type, see MAV_MISSION_TYPE</field>
```
WickedShell: Need to extend MISSION_ITEM_INT as well.

message_definitions/v1.0/common.xml
```diff
@@ -1445,6 +1445,39 @@
...
+        <param index="1">Polygon vertex count</param>
+        <param index="2">Reserved</param>
+        <param index="3">Reserved</param>
+        <param index="4">Reserved</param>
+        <param index="5">Latitude</param>
+        <param index="6">Longitude</param>
+        <param index="7">Reserved</param>
+      </entry>
+      <entry value="5100" name="MAV_CMD_RALLY_POINT">
+        <description>Rally point. You can have multiple rally points defined.
```
julianoes: Can you please add an explanation what a rally point is. I keep thinking of a car rally race.

message_definitions/v1.0/common.xml
```diff
@@ -2062,6 +2095,27 @@
...
+        <description>Autopilot supports mission rally point protocol.</description>
+      </entry>
+    </enum>
+    <enum name="MAV_MISSION_TYPE">
+      <description>Type of mission items being requested/sent in mission protocol.</description>
+      <entry value="0" name="MAV_MISSION_TYPE_MISSION">
+        <description>Items are mission commands for main mission.</description>
+      </entry>
+      <entry value="1" name="MAV_MISSION_TYPE_FENCE">
+        <description>Items are MAV_CMD_FENCE_ GeoFence items.</description>
```
julianoes: `GeoFence` -> `geofence` in my eyes.

bkueng:
There are now 2 transactions needed for inclusion & exclusion polygons. How is the `MAV_CMD_NAV_FENCE_RETURN_POINT` handled? Can there be one for inclusion and one for exclusion? It might be simpler to only have one mission fence type and have `MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION`, `MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION` instead. What do you think?

DonLakeFlyer:
```MAV_CMD_NAV_FENCE_RETURN_POINT``` This is a return point which is used when the fence is breached. There is only one of them. If you look here: http://ardupilot.org/plane/docs/geofencing.html?highlight=fence. You can see how ArduPilot handles it.

With the addition of an exclusion fence I'm not really sure how you can make breach return work all the time. With a single breach return and a fence breach on exclusion fence the vehicle could easily fly through the exclusion zone to get to the breach return. I would guess when there is an exclusion fence breach the vehicle would not go to breach return but take some other action defined by a parameter.

I'm fine with doing inclusion and exclusion in single transaction.

@bkueng Can you take over updating mavlink and get this merged in so we can start implementing. Then we can adjust along the way. I'm hoping we are close enough now to begin work.

DonLakeFlyer:
Superceded by #683

PR #684, opened 2017-03-22
Common: Add a stage option to heartbeat

WickedShell:
Normally there are a number of sub flight stages associated with any given mode. For example on ArduPilot while in mode auto, you can be within the takeoff, normal navigation and landing stages, which would be very helpful to communicate to the GCS to allow the user to know what the aircraft is doing.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2518,6 +2518,8 @@
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags.</field>
       <field type="uint8_t" name="system_status" enum="MAV_STATE">System status flag, see MAV_STATE ENUM</field>
       <field type="uint8_t_mavlink_version" name="mavlink_version">MAVLink version, not writable by user, gets added by protocol because of magic data type: uint8_t_mavlink_version</field>
+      <extensions/>
+      <field type="uint8_t" name="stage">Autopilot-specific flight stage, used for providing information on sub stages of a mode.</field>
     </message>
     <message id="1" name="SYS_STATUS">
       <description>The general system state. If the system is following the MAVLink standard, the system state is mainly defined by three orthogonal states/modes: The system mode, which is either LOCKED (motors shut down and locked), MANUAL (system under RC control), GUIDED (system with autonomous position control, position setpoint controlled manually) or AUTO (system guided by path/waypoint planner). The NAV_MODE defined the current flight state: LIFTOFF (often an open-loop maneuver), LANDING, WAYPOINTS or VECTOR. This represents the internal navigation state machine. The system status shows wether the system is currently active or not and if an emergency occured. During the CRITICAL and EMERGENCY states the MAV is still considered to be active, but should start emergency procedures autonomously. After a failure occured it should first move from active to critical to allow manual intervention and then move to emergency after a certain timeout.</description>


dagar:
I've needed the same solution. Do you think the stages are sufficiently standard they could be an ENUM? Takeoff, normal nav, land?
Are we able to expand ENUMs in the future?

magicrub:
@dagar probably not :(

WickedShell:
I don't think its standard enough, I wish it was, but I think there will be a lot of different stages that get used. I suspect each project will be better served by maintaining their own enum for it, otherwise we end up needing the normal mode + custom mode field again (I think ArduCopter has a lot of sub stages as well, but I'm a plane user so I wouldn't know for sure).

WickedShell:
@tridge @LorenzMeier Any thoughts on getting this merged? I'd like to get started on the autopilot side of hooking this up.

DonLakeFlyer:
Seems like a decent idea to me. Could use to get rid off a bunch of status text messages that happen at boot.

dagar:
Could we have a set of simple predefined stages in common and leave the remainder empty for platform specific usage?

    TAKEOFF
    WAYPOINT
    LOITER
    LANDING

WickedShell:
@dagar I guess we could? However we'd have to limit the range for the common values really quickly. I'm just not sure if that makes enough sense though. I could see an autopilot wanting to express that TAKEOFF actually has several distinct further substages. I admit thats a step further then I'm intending to implement at the moment, but I'm not sure it provides enough justification for us to eat up part of the uint8_t's space. We also want to allow sending stages associated with flight modes other then auto.

Are those the only ones you'd want to reserve?

dagar:
I'm just trying to avoid having to keep things manually in sync between the autopilot and gcs like the current situation with modes.

Anything else that seems extremely general? LANDING-FLARE perhaps?

WickedShell:
I think modes is the perfect example of why I'm not to excited about enumerating stages at the moment. The list of common modes doesn't map well at all to ArduPilot and the message is essentially burned for any ArduPilot use. I think the concept of what a stage means is already explicitly tied to an autopilot implementation.

dagar:
This PR caught my eye because I wanted to add something to signal the vehicle is landing so that QGC can have a big landing abort button only visible when it's relevant. If this sounds similar to your intended usage then I think spending a few minutes to include the obvious commonalities while leaving room for optional custom behaviour would make life easier for the many GCS implementers.

WickedShell:
Heh, that's exactly why I want to use it as well :)

WickedShell:
```
TAKEOFF
WAYPOINT
LANDING
```
Those three are easy and make sense in an autonomous mode, we should reserve 0 as a "not implemented/not relevant state" since this is an extension. Loiter is a bit of a weird one, at least for example if you are doind an auto mission in ArduPilot you can have loiter waypoints, but it still makes more sense to report that as a WAYPOINT stage, as you are still progressing through the planned mission. Actually replacing ```WAYPOINT``` with ```MISSION``` would probably be better? Anything else that stands out?

LANDING-FLARE... I'm not to excited because with other landing types we will either need to maintain their own messages or far far more stages for the sub stages of landing. At the moment I think at a common level we should be happy with LANDING and allow any finer grained control to be handled in it's own message. (For example on Ardupilot we can do deepstall landings, where LANDING-FLARE doesn't map out well. I have a planned message (ArduPilot specific) that addresses the reporting of the substages + other needed information, because trying to map out all these combinations into a single enum set seems really problematic.

DonLakeFlyer:
Wouldn't landing states be here: EXTENDED_SYS_STATE? This is how QGC determines whether the vehicle is flying or not.

```
void Vehicle::_handleExtendedSysState(mavlink_message_t& message)
{
    mavlink_extended_sys_state_t extendedState;
    mavlink_msg_extended_sys_state_decode(&message, &extendedState);

    switch (extendedState.landed_state) {
    case MAV_LANDED_STATE_UNDEFINED:
        break;
    case MAV_LANDED_STATE_ON_GROUND:
        setFlying(false);
        break;
    case MAV_LANDED_STATE_IN_AIR:
        setFlying(true);
        return;
    }
}
```

dagar:
@DonLakeFlyer yes I think adding MAV_LANDED_STATE_LANDING to MAV_LANDED_STATE would work well if we're able to extend it.

WickedShell:
Agreed, sub things like landed inner states should be detailed outside of this, but documenting just that we are in a different state is worthwhile here.

EXTENDED_SYS_STATE is problematic in that I'm not sure how to go ahead and add more detailed landing information, as there are landing types that don't make sense on copters for example, and adding a per autopilot set to the generic seems poor...

@DonLakeFlyer I assume with ArduPlane at least you use the system_statue of MAV_STATE_ACTIVE vs MAV_STATE_STANDBY to tell if it's flying? That is our internal best guess of if we are flying or not.

WickedShell:
Given that #698 gave us a ```MAV_LANDED_STATE_LANDING``` I'd like to suggest we merge this without assigning any common meanings at this point. (If we want a common takeoff state one we could add it as an extension to EXTENDED_SYS_STATE).

tridge:
this looks good to me

WickedShell:
@dagar @LorenzMeier @DonLakeFlyer any thoughts on getting this merged? It seems to be stalled out.

DonLakeFlyer:
Would this sub state be better if it were an opaque value in EXTENDED_SYS_STATE instead. So EXTENDED_SYS_STATE has a few simple well known states and then an opaque vehicle specific sub state for each well known state. Seems to fit in there better than HEARTBEAT.

dagar:
Personally I don't like these undefined fields, but I don't want to go in circles. You might as well use STATUSTEXT.
I'd almost rather have a large list of stages for every submode someone actually wanted added to the spec.
We already have 32 bits in custom_mode, could some of that space be used to accomplish what you want?

WickedShell:
You can't mess with the upper bytes without causing a lot of GCS confusion/versioning mismatch. Statustext isn't even near the same utility.

I don't feel particularly strongly about EXTENDED_SYS_STATE vs heartbeat, if that makes everyone more happy then I'm quite happy to swap to it.

vooon:
Note: in extension fields value 0 should be interpreted as unknown, as that is default when other side know nothing about that new field (works via wire tail-0 reduction).

WickedShell:
Is the conclusion that I should move this to EXTENDED_SYS_STATE? Really looking to use this sooner rather then later, and just want to know which way to be pushing this.

vooon:
I think that changing extended state is better, than changes in essential messages.

WickedShell:
Closed for #736

PR #692, opened 2017-03-31
ArduPilot merge

tridge:
this merges all the recent XML changes from ArduPilot, including the enum tags and the recent message additions


DonLakeFlyer:
All looks great to me. Will be super useful when QGC gets around to using metadata from mavlink messages the same way it does for parameters.

PR #695, opened 2017-04-06
Add spec for NaN on yaw angle

DonLakeFlyer:
This specifies that NaN for yaw on these commands means no change in yaw:
```
MAV_CMD_NAV_WAYPOINT
MAV_CMD_NAV_LAND
MAV_CMD_NAV_TAKEOFF
MAV_CMD_NAV_VTOL_TAKEOFF
MAV_CMD_NAV_VTOL_LAND
```

This is the way PX4 Pro works. I believe that for a number of these, ArduPilot does not support the Yaw field. Can this be checked to work on ArduPilot as well. This allows for things like running a survey at a fixed yaw angle (relative to ground, not vehicle) without needing to specify yaw on every single waypoint in the survey as an example.

@tridge Can someone on the ArduPilot side help out with verification?


DonLakeFlyer:
Related to: https://github.com/mavlink/qgroundcontrol/issues/4902

DonLakeFlyer:
Verified that this is ok with ArduPilot since they don't support values in those commands currently.

DonLakeFlyer:
Waiting on go head from Lorenz to allow me to merge metadata changes.

LorenzMeier:
Since its validated let's do this.

PR #698, opened 2017-04-10
add MAV_LANDED_STATE_LANDING

dagar:
https://github.com/mavlink/qgroundcontrol/issues/4935

@DonLakeFlyer @WickedShell

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2227,6 +2227,12 @@
       <entry value="2" name="MAV_LANDED_STATE_IN_AIR">
         <description>MAV is in air</description>
       </entry>
+      <entry value="3" name="MAV_LANDED_STATE_TAKEOFF">
+        <description>MAV currently taking off</description>
+      </entry>
+      <entry value="4" name="MAV_LANDED_STATE_LANDING">
+        <description>MAV currently landing</description>
+      </entry>
     </enum>
     <enum name="ADSB_ALTITUDE_TYPE">
       <description>Enumeration of the ADSB altimeter types</description>


magicrub:
is there a "takeoff_state"?

dagar:
@WickedShell yes, and it changes the conversation somewhat if the extremely common stuff is covered elsewhere.

@magicrub I don't believe so. I definitely thought about adding that here as well, but it seemed questionable. What do you guys think?

magicrub:
sounds like it doesn't hurt to add a takeoff flavor

WickedShell:
@dagar cool, I'd quite like to get that in soon, but I do get why it's been held up.

I'm not terribly enthused about the LANDING_STATE/TAKEOFF_STATE ideas, but they do cover the generic cases well. I'm just not sure what the behaviour would be once you complete a takeoff/landing, do you keep sending the message?

dagar:
@WickedShell EXTENDED_SYS_STATE (http://mavlink.org/messages/common#EXTENDED_SYS_STATE) is streamed at a relatively low rate in PX4. 

I've updated the PR with a takeoff state.

@DonLakeFlyer review?

DonLakeFlyer:
Perfect. @LorenzMeier Can i merge?

rmackay9:
100% agree we need a landed-state.  Very nice to see this being added.

I'm less sure about the need for a take-off state.  I don't think in general we should try to think too far ahead.  The only place where ardupilot might need a take-off state is for traditional helicopters in order to tell the GCS if the rotors have spun up completely or not.  I haven't heard anyone actually asking for that notification though.

DonLakeFlyer:
Keep in mind that is is a "landing" state. Signalling that the vehicle is in the process of landing. Not a "landed" state. EXTENDED_SYS_STATUS already has values that indicate on the ground or in the air.

rmackay9:
Ah right, OK.  So what I want is already there! thanks.

LorenzMeier:
Thanks all!

PR #699, opened 2017-04-11
Remove WIP label from GPS_RTCM_DATA

WickedShell:
GPS_RTCM_DATA is used in releases, and is no longer a WIP message.


rmackay9:
This is a non functional change so I think this should go straight in.

WickedShell:
Yeah, I missed it when I updated the flags description, but it really should have been removed then (and even that should have been done a long time ago)

LorenzMeier:
Indeed, thanks!

PR #700, opened 2017-04-11
ardupilot: add VISION_POSITION_DELTA message definition

rmackay9:
This adds a new VISION_POSITION_DELTA mavlink message to the ArduPilot specific messages.

This message could probably be integrated into common.xml but there will be more developments that might lead us to add more fields to this message. Because ArduPilot is the only user for the moment I think it's probably ok to leave it in the ardupilot specific set.

This has been successfully flight tested with ardupilot and a stereo ZED camera.

Here is the equivalent downstream PR: https://github.com/ArduPilot/mavlink/pull/33

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -1374,5 +1374,14 @@
       <field name="f_dot" type="float">projection operator derivative</field>
       <field name="u" type="float">u adaptive controlled output command</field>
     </message>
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Have you considered using quaternions?

rmackay9:
It's actually a rotation vector so I've updated the description of the angle_delta field.
We've considered quaternions but the result of the discussion was that this is the way to go.

PR #701, opened 2017-04-11
More enums

amilcarlucas:
Please ignore all but the very last commit


rmackay9:
Similar to the other PR, I think these should be rebased on this repo's master branch.

In general, I think the last commit to add in the "enum=XXX" meta data and the "display=bitmask" makes sense.  I guess people could debate whether it should be "display" or "display-as" or whatever, but it seems like a good addition to me and will likely help the GCS developers.

amilcarlucas:
Rebased on #705

PR #703, opened 2017-04-11
Add si units

amilcarlucas:
Only the last two comits are relevant


rmackay9:
So as mentioned off line, it would be good to rebase this so that only the last two commits appear.

Re the last two commits, I think standardizing the order of the meta data makes sense and adding the "units" meta data should also be helpful for the ground stations.

So for me, I think these last two commits are good changes.

amilcarlucas:
I will rebase this as soon as #705 and #706 are in, because this patch builds upon those two

amilcarlucas:
I did a rebase of this on #707

PR #635, opened 2016-10-11
Camera and gimbal improvements

julianoes:
This is an attempt to improve and extend the mavlink camera and gimbal API.

The command `DIGICAM_CONFIGURE` is deprecated in favor of `CAMERA_INFORMATION`, `CAMERA_CAPTURE_STATUS`, `CAMERA_SETTINGS_1`, and `CAMERA_SETTINGS_2`.

For the camera storage `STORAGE_INFORMATION` and `STORAGE_FORMAT` has been added.

To control the gimbal, the lat/lon fields have been moved to param5/param6 for int32 accuracy, and a `MOUNT_STATUS` is added for the realtime feedback about a gimbal (or other mount).

Please review:
@LorenzMeier, @mcharleb

PR #709, opened 2017-04-17
mavlink_defs: image quality, metering, flicker

julianoes:
This adds more camera settings:
- image quality ID (e.g. for JPG this could translate to 0-100)
- metering mode ID (this is for different metering modes specific to a
  camera)
- flicker mode ID (this is for different flicker modes, also specific to
  a camera.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1347,6 +1347,16 @@
         <param index="4">Reserved for camera mode ID</param>
         <param index="5">Reserved for color mode ID</param>
         <param index="6">Reserved for image format ID</param>
+        <param index="7">Reserved for image quality ID</param>
+      </entry>
+      <entry value="529" name="MAV_CMD_SET_CAMERA_SETTINGS_3">
+        <description>WIP: Set the camera settings part 3 (CAMERA_SETTINGS)</description>
+        <param index="1">Camera ID</param>
+        <param index="2">Reserved for metering mode ID</param>
+        <param index="3">Reserved for flicker mode ID</param>
+        <param index="4">Reserved</param>
+        <param index="5">Reserved</param>
+        <param index="6">Reserved</param>
         <param index="7">Reserved</param>
       </entry>
       <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
@@ -3926,6 +3936,9 @@
       <field type="uint8_t" name="mode_id">Reserved for a camera mode ID</field>
       <field type="uint8_t" name="color_mode_id">Reserved for a color mode ID</field>
       <field type="uint8_t" name="image_format_id">Reserved for image format ID</field>
+      <field type="uint8_t" name="image_quality_id">Reserved for image quality ID</field>
+      <field type="uint8_t" name="metering_mode_id">Reserved for metering mode ID</field>
+      <field type="uint8_t" name="flicker_mode_id">Reserved for flicker mode ID</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
       <description>WIP: Information about a storage medium</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3908,7 +3908,7 @@
       <field type="uint8_t" name="camera_id">Camera ID if there are multiple</field>
       <field type="uint8_t[32]" name="vendor_name">Name of the camera vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the camera model</field>
-      <field type="uint8_t[16]" name="firmware_version">Version of the camera firmware</field>
+      <field type="uint32_t" name="firmware_version">Version of the camera firmware</field>
       <field type="float" name="focal_length" units="mm">Focal length in mm</field>
       <field type="float" name="sensor_size_h" units="mm">Image sensor size horizontal in mm</field>
       <field type="float" name="sensor_size_v" units="mm">Image sensor size vertical in mm</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1357,7 +1357,7 @@
         <param index="4">Reserved</param>
         <param index="5">Reserved</param>
         <param index="6">Reserved</param>
-        <param index="7">Reserved</param>
+        <param index="7">1: reset all camera settings, 0: no action.</param>
       </entry>
       <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
         <description>WIP: Request storage information (STORAGE_INFORMATION)</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1330,34 +1330,24 @@
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="523" name="MAV_CMD_SET_CAMERA_SETTINGS_1">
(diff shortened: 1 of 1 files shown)


dogmaphobic:
- Marked entire messages as WIP not all individual elements within
- Defined that NAN means an unchanged value for settings
- Removed “Locked” states in settings
- Added “Auto Exposure” mode (handle the individual locked states) and a proper handling of Auto Exposure
- Added EV value
- Added Audio Record (On/Off)
- Added separate message for Reset All Camera Settings
- Removed “Resolution in megapixels” as it doesn’t make sense.
- Made Camera ID consistent whenever it is used (always param 1)
- Added camera firmware version to CAMERA_INFORMATION

message_definitions/v1.0/common.xml
```diff
@@ -1357,33 +1357,38 @@
...
+      </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
-        <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture.</description>
-        <param index="1">Duration between two consecutive pictures (in seconds)</param>
-        <param index="2">Number of images to capture total - 0 for unlimited capture</param>
-        <param index="3">Resolution in megapixels (0.3 for 640x480, 1.3 for 1280x720, etc), set to 0 if param 4/5 are used, set to -1 for highest resolution possible.</param>
-        <param index="4">WIP: Resolution horizontal in pixels</param>
-        <param index="5">WIP: Resolution horizontal in pixels</param>
-        <param index="6">WIP: Camera ID</param>
+        <description>WIP: Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture.</description>
```
julianoes: The problem with these changes is that it could break earlier implementations (from before when I started to change camera messages, see: https://github.com/mavlink/mavlink/pull/635/files#diff-ceacc1e2390186650146a45be8fcb691L1313).

That's also why the camera ID was in an odd place. And that's why not the whole message was marked WIP.

message_definitions/v1.0/common.xml
```diff
@@ -1357,33 +1357,38 @@
...
+        <param index="1">Camera ID (0 for all cameras), 1 for first, 2 for second, etc.</param>
+        <param index="2">Duration between two consecutive pictures (in seconds)</param>
+        <param index="3">Number of images to capture total - 0 for unlimited capture</param>
+        <param index="4">Resolution horizontal in pixels (set to -1 for highest resolution possible)</param>
+        <param index="5">Resolution vertical in pixels (set to -1 for highest resolution possible)</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE">
-        <description>Stop image capture sequence</description>
-        <param index="1">Camera ID</param>
+        <description>WIP: Stop image capture sequence</description>
```
julianoes: Same here, the command already existed, only the params are WIP.

message_definitions/v1.0/common.xml
```diff
@@ -3904,28 +3908,30 @@
...
       <field type="uint8_t[32]" name="vendor_name">Name of the camera vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the camera model</field>
+      <field type="uint8_t[16]" name="firmware_version">Version of the camera firmware</field>
       <field type="float" name="focal_length" units="mm">Focal length in mm</field>
       <field type="float" name="sensor_size_h" units="mm">Image sensor size horizontal in mm</field>
       <field type="float" name="sensor_size_v" units="mm">Image sensor size vertical in mm</field>
       <field type="uint16_t" name="resolution_h" units="pix">Image resolution in pixels horizontal</field>
       <field type="uint16_t" name="resolution_v" units="pix">Image resolution in pixels vertical</field>
-      <field type="uint8_t" name="lense_id">Reserved for a lense ID</field>
+      <field type="uint8_t" name="lens_id">Reserved for a lens ID</field>
```
julianoes: :+1:

message_definitions/v1.0/common.xml
```diff
@@ -3904,28 +3908,30 @@
       <field type="uint8_t" name="camera_id">Camera ID if there are multiple</field>
       <field type="uint8_t[32]" name="vendor_name">Name of the camera vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the camera model</field>
+      <field type="uint8_t[16]" name="firmware_version">Version of the camera firmware</field>
```
julianoes: This is a string? I think we should use the same as for `AUTOPILOT_VERSION` but with less detail.

message_definitions/v1.0/common.xml
```diff
@@ -1330,24 +1330,24 @@
...
+        <description>WIP: Set the camera settings part 1 (CAMERA_SETTINGS). Use NAN for values you don't want to change.</description>
         <param index="1">Camera ID</param>
         <param index="2">Aperture (1/value)</param>
-        <param index="3">Aperture locked (0: auto, 1: locked)</param>
-        <param index="4">Shutter speed in s</param>
-        <param index="5">Shutter speed locked (0: auto, 1: locked)</param>
-        <param index="6">ISO sensitivity</param>
-        <param index="7">ISO sensitivity locked (0: auto, 1: locked)</param>
+        <param index="3">Shutter speed in s</param>
+        <param index="4">ISO sensitivity</param>
```
julianoes: Is there a ISO auto mode?

message_definitions/v1.0/common.xml
```diff
@@ -1357,33 +1357,38 @@
...
       <entry value="528" name="MAV_CMD_REQUEST_FLIGHT_INFORMATION">
         <description>WIP: Request flight information (FLIGHT_INFORMATION)</description>
         <param index="1">1: Request flight information</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
+      <entry value="529" name="MAV_CMD_RESET_CAMERA_SETTINGS">
+        <description>WIP: Reset all camera settings to Factory Default (CAMERA_SETTINGS)</description>
+        <param index="1">Camera ID (0 for all cameras), 1 for first, 2 for second, etc.</param>
+        <param index="2">0: No Action 1: Reset all settings</param>
+        <param index="3">Reserved (all remaining params)</param>
```
julianoes: Makes sense.

message_definitions/v1.0/common.xml
dogmaphobic: That would be a problem if anybody else is using it, which I tend to doubt as these messages are incomplete. If someone was trying to use them, we would have heard as they would either comment of their incomplete state or attempt to "fix" it themselves.

message_definitions/v1.0/common.xml
dogmaphobic: The reason I made into a string is that although we have control over the autopilot version, you may not have control over the camera version. You are more likely to get a string from the camera and you would have to parse the individual values off of it. I do now realize that it's too short (16 characters). Not sure why it ended up like that. It was meant to be 32 characters like the rest.

message_definitions/v1.0/common.xml
julianoes: Ok, let's change it then.

message_definitions/v1.0/common.xml
julianoes: The thing is you need to parse it on the receiving end anyway, so you might as well do it on the camera itself.

dogmaphobic: You're right. And it will save a lot of space. I will make these changes.

PR #711, opened 2017-04-22
Mixer parameter tuning

crashmatt:
**BENEFITS**
Enables downloading and editing of a mixer specific list of parameters. This parameter list is dynamic according to mixer configuration.

**FEATURES**
- Parameters are indexed and named items with mixer metadata.
- Parameters are accessed through index since names can be repeated
- Parameters value are arrays of the same type with length up to 8.
- Parameters have human readable identifiers with length up to 20 chars
- Parameters have type so they can be used to describe more metadata
- Parameters with zero length can be used as name tags
- Parameters can be read only which is set through flags
- Metadata and type support is optional depending on the particular implementation

**CHANGES SINCE INITIAL PR**
- Changed from 16 chars to 20
- Changed from max array size of 6 to 8
- Added parameter message type enumeration
- Simplified set param and param request messages
- Removed parameter_index as unused
- Removed mixer_type as that is supported by specific parameter type
- Error reported in flags instead of being encoded in the count value
- Moved from MAVLink1 to MAVLink2 namespace
- Added checksum parameter type so a gcs knows if the downloaded script is up-to-date
- Marked messages as WIP

[Mavlink mixer tuning description document](https://docs.google.com/document/d/1LeQC2qtxqdesNuJ3RAWH_MJ_30xjf-PMN3bd5wtUAUo/edit#heading=h.3cs7tuenkkib)

**TODO**
Tidy up into a single commit

Screenshot from typical shell output of all parameter data.  This is close to the complete data seen through mavlink.
image

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1465,6 +1465,14 @@
         <param index="5">Unscaled target latitude of center of circle in CIRCLE_MODE</param>
         <param index="6">Unscaled target longitude of center of circle in CIRCLE_MODE</param>
       </entry>
+      <entry value="4101" name="MAV_CMD_REQUEST_MIXER_PARAM_LIST">
+        <description>Request to send a list of all mixer data.</description>
+        <param index="1">Mixer Group</param>
+      </entry>
+      <entry value="4102" name="MAV_CMD_SAVE_MIXER_PARAMS">
+        <description>Save mixer parameters</description>
+        <param index="1">Mixer Group</param>
+      </entry>
       <entry value="5000" name="MAV_CMD_NAV_FENCE_RETURN_POINT">
         <description>Fence return point. There can only be one fence return point.
         </description>
@@ -2432,6 +2440,33 @@
       </entry>
       <entry value="7" name="GPS_FIX_TYPE_STATIC">
         <description>Static fixed, typically used for base stations</description>
+      </entry>
+    </enum>
+    <enum name="MIXER_GROUP">
+      <description>Identifies a set of mixers to access</description>
+      <entry value="0" name="MIXER_GROUP_NONE">
+        <description>No specific group selected</description>
+      </entry>
+      <entry value="1" name="MIXER_GROUP_LOCAL">
+        <description>Mixers on the local processor</description>
+      </entry>
+      <entry value="2" name="MIXER_GROUP_FAILSAFE">
+        <description>Mixers on a failsafe device</description>
+      </entry>
+    </enum>
+    <enum name="MIXER_PARAM_FLAGS">
+      <description>Flags for different parameter qualities and states</description>
+      <entry value="1" name="MIXER_PARAM_FLAG_READ_ONLY">
+        <description>Parameter is read only</description>
+      </entry>
+      <entry value="2" name="MIXER_PARAM_FLAG_MIX_CONN">
+        <description>Parameter is describing a mixer connection to a group</description>
+      </entry>
+      <entry value="4" name="MIXER_PARAM_FLAG_TYPE">
+        <description>Parameter is describing a mixer type</description>
+      </entry>
+      <entry value="8" name="MIXER_PARAM_FLAG_CONN_GROUP">
+        <description>Parameter is describing a connection group</description>
       </entry>
     </enum>
   </enums>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -3764,6 +3799,47 @@
       <field type="uint8_t" name="wp_num">current waypoint number</field>
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
+    <message id="235" name="MIXER_PARAM_VALUE">
+      <description>Mixer parameter values emitted after a read or set request</description>
+      <field type="uint8_t" name="flags" enum="MIXER_PARAM_FLAGS">Flags for the parameter</field>
+      <field type="int16_t" name="count">Total number of onboard parameters</field>
```
magicrub: Why is the count signed?

magicrub:
I feel like you're missing a long enum list describing several common mixer options.

magicrub:
What is so special about fitting in a 64byte MAVLink packet? Normal max size is 255

crashmatt:
@magicrub To be clear, is that long enum for configuration or reading capabilities?  Either way it is a good idea. A suggestion:

One of the flag enumerations is for parameters describing the mixer options.
Parameter index zero always returns as capabilities/options
A parameter has up to 24bytes of data so there is plenty of space for that long enum list

Have you an example of what might be in that enumeration?

crashmatt:
@WickedShell  Its a good thing that 64byte limit is there or I would be knocking on the door of the 255 limit.
There are two items fighting for the majority of the space, parameter data and parameter name.

Arrays of 6 are already 5 values larger than what we had before.  I have seen 7 point mixers on some radios and that seems kind of overkill.

The limit of 16 chars for parameter names is annoying and I did think about extending that.  If you want a nice human readable mixer script you are likely storing mixer parameter and variable names in ram.  I would like to have 20,24 or 32 chars but we have to put them somewhere.

I am tempted to expand mixer names to 20 chars.  We will get into writing new mixer scripts soon.  If 20 chars is not enough then we can adjust it later.
Since the larger name size takes the message over the 64byte limit I might as well expand the array size to 8 values.

Another thought.  When you are doing mixer tuning you are doing it in close range.  The radio link is not so stressed and loosing a packet is not mission critical.

crashmatt:
@WickedShell Are you thinking that array size should be 16 to cover maximum channel size?

message_definitions/v1.0/common.xml
```diff
@@ -3764,6 +3814,38 @@
...
+      <field type="uint8_t" name="flags" enum="MIXER_PARAM_FLAGS">Flags for the parameter</field>
+      <field type="uint8_t" name="type" enum="MIXER_PARAM_TYPE">Type of the parameter - Used for connections, mixer names etc..</field>
+      <field type="uint16_t" name="mixer_index">Index of the mixer in the mixer collection</field>
+      <field type="uint8_t" name="mixer_sub_index">Index of the sub mixer in the mixer</field>
+      <field type="uint8_t" name="mixer_type">Implementation specific identifier for mixer type</field>
+      <field type="uint16_t" name="parameter_index">Index of the parameter in a mixer or something else</field>
+      <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
+      <field type="uint8_t" name="param_array_size">Size of the array for this parameter</field>
+      <field type="char[20]" name="param_id">Onboard parameter id, terminated by NULL if the length is less than 20 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
```
WickedShell: Comment talks about 16 chars, the array however is 20 long.

message_definitions/v1.0/common.xml
```diff
@@ -3764,6 +3814,38 @@
...
+      <field type="int16_t" name="count">Total number of parameters in the group</field>
+      <field type="uint16_t" name="index">Index of the parameter in the group</field>
+      <field type="uint8_t" name="flags" enum="MIXER_PARAM_FLAGS">Flags for the parameter</field>
+      <field type="uint8_t" name="type" enum="MIXER_PARAM_TYPE">Type of the parameter - Used for connections, mixer names etc..</field>
+      <field type="uint16_t" name="mixer_index">Index of the mixer in the mixer collection</field>
+      <field type="uint8_t" name="mixer_sub_index">Index of the sub mixer in the mixer</field>
+      <field type="uint8_t" name="mixer_type">Implementation specific identifier for mixer type</field>
+      <field type="uint16_t" name="parameter_index">Index of the parameter in a mixer or something else</field>
+      <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
+      <field type="uint8_t" name="param_array_size">Size of the array for this parameter</field>
```
WickedShell: Should be noted that this can't be > 8

message_definitions/v1.0/common.xml
```diff
@@ -3764,6 +3814,38 @@
       <field type="uint8_t" name="wp_num">current waypoint number</field>
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
+    <message id="235" name="MIXER_PARAM_VALUE">
+      <description>Mixer parameter values emitted after a read or set request</description>
+      <field type="uint8_t" name="mixer_group" enum="MIXER_GROUP">Access to different sets of mixers</field>
+      <field type="int16_t" name="count">Total number of parameters in the group</field>
+      <field type="uint16_t" name="index">Index of the parameter in the group</field>
```
crashmatt: This came from a time when parameters were one value only.  Then I thought it might not be enough.   

Array values came from a need to write a mixer script in the form of:

input_values = [-1.0, -0.5, 0, 0.5, 1.0]
output_values = [-1.0, 0.25, 0, 0.25, 1.0]
aileron = multipoint(roll, input_values, output_values) 

The count value is the only signed value so it can be used to return a negative number indicating an error.  That reduced the total parameters to 128.
Maybe I should put the error indicator into flags?

message_definitions/v1.0/common.xml
```diff
@@ -3764,6 +3814,38 @@
...
+      <field type="uint8_t" name="flags" enum="MIXER_PARAM_FLAGS">Flags for the parameter</field>
+      <field type="uint8_t" name="type" enum="MIXER_PARAM_TYPE">Type of the parameter - Used for connections, mixer names etc..</field>
+      <field type="uint16_t" name="mixer_index">Index of the mixer in the mixer collection</field>
```
crashmatt: If you consider that a mixer can be as small as an operator then I can imagine someone going beyond the 255 limit. For example I count 18 mixers/operators here:

if(flap > 0) flapgain = 0.5 else flapgain = 0.25
aileron1 = (roll * rollgain) + (flap * flapgain) + (pitch*pitch_flap_gain)
aileron2 = - (roll * rollgain) + (flap * flapgain) + (pitch*pitch_flap_gain)

message_definitions/v1.0/common.xml
crashmatt: It was changed to 20 chars after the discussion on 64byte length not being a significant problem.

message_definitions/v1.0/common.xml
crashmatt: Good idea.  Will fix this.

crashmatt:
@WickedShell 
I will provide some script examples on how the mixers might be defined.  That will make it clearer where the different parameters are coming from.

I tried prefixes but very quickly ran out of character space while keeping the naming clear.  These parameters can't have the nice gcs description metadata that normal parameters do.  They need to have clear naming to work well for the user.

I get the feeling that the postfix is a bit of a hack because parameters don't have the right data defined.  Managing and maintaining data embedded in strings is much more work than simply having an array count or index.

There have been some significant changes recently so I need to update the description document.

WickedShell:
I guess my question boils down to how do you actually provide a mixer file or know as a GCS how that file maps out so that you can present a meaningful interface? I get that this provides a transport mechanism, but without seeing how the mixer itself is given to/from the autopilot I don't see how you arrive at anything that is fundamentally better then using parameters. (It currently appears to me like this is replicating parameters in a sub group). I'm guessing I'm just missing something that is envisioned on going with this.

message_definitions/v1.0/common.xml
crashmatt: Now I see the problem.  Will fix this.

crashmatt:
Everything you might need to know about about a particular mixer parameter
- Which group it belongs to
- A reference index or other unique identifier
- Which mixer it belongs to (unique identifier or index)
- Which submixer it belongs to (for scaleable mixers ie multirotor)
- Array index or array data
- A human readable name for the parameter
also useful to know for good gcs presentation
- The type of mixer or submixer it comes from
- If the parameter is read only
- The name of the mixer type
etc.....

Just taking the top 6 must have data items on a standard mavlink parameter.  How do we encode all of this in 16chars?
The index is already encoded in the parameter message.  
It needs a standard mixer prefix.  Maybe two if you wish to encode the group, say "MI_" and "MX_" for internal and external mixer.  
Mixer index would be limited to two chars but we can use Base36 to get 1296 mixers
Submixer could be limited to 36 with no problem.  That is one more Base36 char.
Array index - Another Base36 char
Type - Something that tells you what kind of data you are getting.  Another Base36 char.

That leaves 8 chars for a human readable name

Now write a mixer script using names of only 8 chars.
If you have some nostalgia for the days of 8.3 filename length limitations then this is the right thing for you.

message_definitions/v1.0/common.xml
crashmatt: Another thought: I recently had a prototype setup with 80 parameters.  With the new parameter types that could easily exceed 256.

crashmatt:
@WickedShell Here is a very early work in progress example of a json based mixer script.
[Early prototype json mixer script](https://docs.google.com/document/d/1YxXW7dcRLLNA23Z476KMNCKQALol9l5BpXPQapvfsxc/edit?usp=sharing)

The "parameters" array is a list of global parameters that can be used by the mixers.
Note the name lengths which will tend to grow with more complex use.
Note the arrays which allow easy grouping of data.

The script is not intended to be a definitive method for mixer configuration.  I am only using it as a method to generate test data.  Many different language styles could be used to define the mixers.  This is a specific implementation only.

[There was a QGC prototype implementation PR here - now closed due to bitrot](https://github.com/mavlink/qgroundcontrol/pull/4727)

The PR infomation gives some screenshots of the parameter values being read.  The format is not pretty but functional.

crashmatt:
@WickedShell  A clear reason why standard parameters can't be used:

1. Parameter list is requested
2. Parameter download starts
3. Mixer script is loaded and changes parameter count / content
4. Parameter download is broken or invalidated.

If we can agree that mixer parameters must be a separate entity from standard parameters then I don't see any reason to not move forward with this PR






6.

message_definitions/v1.0/common.xml
crashmatt: @magicrub count was signed as a way to indicate an error value.  It is no longer signed since an error is returned in flags.

crashmatt:
@LorenzMeier Please review this.

WickedShell:
@crashmatt until you have a way to actually roundtrip the mixer at the same time, I think the utility of this is hugely (to almost entirely) limited until there is a way to download a mixer to a GCS that doesn't yet know about it.

Your scenario about why standard parameters can't be used doesn't really make sense, given that could happen while changing mixer scripts later (which I don't see significed anywhere) and I don't see how the GCS knows which mixer script is loaded, so as far as I can tell this is still a risk.

Can you change this to live in the MAVLink2 namespace? I think we can probably agree that mixer related items will gain the need for more indexs in the future, and I'd hate to see us burn some of the last MAVLink 1 indexes on a large feature that might need more commands then are available in MAVLink1.

crashmatt:
@WickedShell 
Last thing first - You are right about not consuming the MAVlink1 namespace. I moved to MAVLink1 namespace to make some combination of ardupilot and mavproxy work.  Will move it back to MAVLink2.

Mixer parameters are not intended as a way to roundtrip the whole mixer. A roundtrip means you can completely re-create the original mixer script from the information you downloaded.  I can't predict what scripts might be defined and preferred by users.  I know that if I try to support this I will fail.
Also, most scripts have limitations round-tripping themselves.  Some scripts have significant problems reversing at all (eg. lexx-yacc based). Json seems to 

As the title suggests, this intent of the function is mixer parameter tuning.  serialization of the mixer is entirely up to the particular mixer implementation.  Mixer parameters is a standard way to modify the mixer.  This is similar to how mavlink parameters tune a specific AP implementation without having to know everything about it.

Identifying the loaded mixer script would be good.  I don't think there is a good way to get a script path through mixer parameters.  A mixer script checksum can be added to the parameter types.  That would let the gcs know if it needs to update its copy of a script, probably through ftp.

There is one complete alternative which I is somewhat appealing.  That is to do everything as json objects.  The mixer needs a bi-directional stream through mavlink to the GCS.  It receives objects as mixer modifications or requests.  It sends objects which can be the whole mixer script or parts of it.  Despite this simplicity the implementation is not so simple.  It also restricts scripts to be json only.  

/Matt

crashmatt:
Closing in favor of hard coded purpose specific mixers based on standard mavlink parameters

PR #713, opened 2017-04-28
common.xml: add progress to MAV_RESULT

julianoes:
This adds a field to display the progress in percent from 0 to 100, and
MAV_RESULT_IN_PROGRESS to signal that the command is being executed but
not yet finished.

MAV_RESULT_IN_PROGRESS can be followed by any of the other MAV_RESULTs
depending on what happened.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1817,6 +1817,9 @@
       <entry value="4" name="MAV_RESULT_FAILED">
         <description>Command executed, but failed</description>
       </entry>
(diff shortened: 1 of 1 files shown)


LorenzMeier:
@julianoes Can you mark this WIP so we can merge but make clear that this needs further discussion with all stakeholders before it can be deployed and considered agreed-on?

@tridge @rmackay9 I think we need some way of marking an in-progress command and this would be a non-intrusive way of adding this. Existing implementations will only see the final ACK, so this should be a compatible update.

julianoes:
@LorenzMeier done 931086d.

PR #714, opened 2017-04-29
scripts: Add C++11 headers to c_library_v2

vooon:
With that change [c_library_v2][1] will also provide C++11 headers.
C++ generator also call C gen, so all what need to be changed is `--lang`.

[1]: https://github.com/mavlink/c_library_v2


TSC21:
Is this addressed already with https://github.com/mavlink/mavlink/commit/dc415938cea1cde5d5e2eebbb27a920e67c1ec4e @tridge?

LorenzMeier:
@vooon Could you rebase and is this still up to date?

vooon:
@LorenzMeier rebased. Yes, it still should work.

TSC21:
@vooon still relevant, right? Can you rebase? @LorenzMeier should we bring this in?

julianoes:
What about having a different repo for C++? Is that something that has been considered?

hamishwillee:
> What about having a different repo for C++? Is that something that has been considered?

I assume that adding C++11 header to c_library_v2 will not break a C build?

FWIW C and C++ are quite different and it makes sense to let these evolve differently. I'd be tempted to generate separate cpp_library_v2 repo for now. 

In future it might make sense to separate generator core and language specific bindings, in which case cpp would also have its own binding repo.

julianoes:
> I'd be tempted to generate separate cpp_library_v2 repo for now.

I agree but we need to give it a better name, e.g. `mavlink_v2_generated_cpp`.

vooon:
C++ is not standalone, it uses same C utils, e.g. parser.
Only different part is serdes.

hamishwillee:
@vooon 

> C++ is not standalone, it uses same C utils, e.g. parser.

Yes, we know that. The questions are whether it should be allowed to _evolve_ separately, and whether including those headers causes problems for people developing in C.

vooon:
1. Reuse of existing utils reduce support cost, e.g. extensions to protocol do not require changes on two places, so i don't think that separation is actually required.
2. Placing `hpp` files along with `h` should not be a problem for C. Developer can choose what it needs H or HPP.
  And note that C headers with required modifications for C++ already used everywhere.

amilcarlucas:
Can this one be discussed at the dev meeting ?

hamishwillee:
@amilcarlucas I added the dev call label. Note however that we rarely get through more than a couple of issues, so it actually being looked at depends on having a sponsor. Specifically, someone there who cared about it and can argue the case for or against.

LorenzMeier:
Would it make sense to add a 2nd repo just for C++? I presume you're not using C and C++ in the same project?

PR #715, opened 2017-05-01
Components: Add minimum set and mark deprecated components.

LorenzMeier:
A number of the components were unused and we do not want to spend the ID space on them. At the same time we need more standard IDs to help adopters choose sensible de-confliction defaults.

The proposal would be to register classes on offsets of 10's:

- 1-9: Autopilots
- 20-29: Communication equipment
- 30-39: Routing components
- etc


LorenzMeier:
Rebased.

LorenzMeier:
Stale, removing.

PR #718, opened 2017-05-05
Extend camera trigger specs. From @mhkabir

LorenzMeier:
This extends the camera trigger specs to be overall more consistent, in particular for survey missions.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -992,7 +992,7 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="189" name="MAV_CMD_DO_LAND_START">
-        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts. It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used. The Latitude/Longitude is optional, and may be set to 0/0 if not needed. If specified then it will be used to help find the closest landing sequence.</description>
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts. It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used. The Latitude/Longitude is optional, and may be set to 0 if not needed. If specified then it will be used to help find the closest landing sequence.</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1174,6 +1175,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="214" name="MAV_CMD_DO_SET_CAM_TRIGG_INTERVAL">
+        <description>Mission command to set camera trigger interval for this flight. If triggering is enabled, the camera is triggered each time this interval expires. This command can also be used to set the shutter integration time for the camera.</description>
+        <param index="1">Camera trigger cycle time (milliseconds). -1 or 0 to ignore.</param>
```
julianoes: -1 or 0 to ignore is weird because you're now reserving 2 values for the same.
What about 0 for a default and -1 for no change?

message_definitions/v1.0/common.xml
```diff
@@ -1164,6 +1164,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <!-- 211 and 212 used in dialects -->
```
julianoes: Good to know!

message_definitions/v1.0/common.xml
```diff
@@ -1091,7 +1091,7 @@
         <param index="4">Focus Locking, Unlocking or Re-locking</param>
         <param index="5">Shooting Command</param>
         <param index="6">Command Identity</param>
-        <param index="7">Empty</param>
+        <param index="7">Test shot identifier. If set to 1, image will only be captured, but not counted towards internal frame count.</param>
```
julianoes: The messages `MAV_CMD_DO_DIGICAM_CONFIGURE` and `MAV_CMD_DO_DIGICAM_CONTROL` should be deprecated in favor of `MAV_CMD_IMAGE_START_CAPTURE`, etc. and `MAV_CMD_SET_CAMERA_SETTINGS_1` etc.

message_definitions/v1.0/common.xml
```diff
@@ -1406,8 +1417,8 @@
       <entry value="2003" name="MAV_CMD_DO_TRIGGER_CONTROL">
         <description>Enable or disable on-board camera triggering system.</description>
         <param index="1">Trigger enable/disable (0 for disable, 1 for start), -1 to ignore</param>
-        <param index="2">Shutter integration time (in ms), -1 to ignore</param>
-        <param index="3">1 to reset the trigger sequence, -1/0 to ignore</param>
+        <param index="2">1 to reset the trigger sequence, -1 or 0 to ignore</param>
+        <param index="3">1 to pause triggering, but without switching the camera off or retracting it. -1 to ignore</param>
```
julianoes: Retracting what exactly? I don't follow.

message_definitions/v1.0/common.xml
mhkabir: Okay, hmm. I was not sure which way we were planning to deprecate. But good to know. I will do that in my the second iteration of QGC and Firmware fixes. For now I made QGC and Firmware consistently use DIGICAM_CONTROL.

message_definitions/v1.0/common.xml
mhkabir: The lens.

PR #720, opened 2017-05-05
add timestamp to HOME_POSITION msgs

TSC21:
Allows tracking when the home position is changed or set.


DonLakeFlyer:
Curious as to why this would be need or how it would be used?

TSC21:
There's a plugin on mavros for home position to get and set home positions. It will allow sync and info to external systems or gcs to know when was a new home position defined.

TSC21:
ping @LorenzMeier

TSC21:
Is there anything else to be considered/modified so this can be merged?

TSC21:
@DonLakeFlyer friendly ping :)

DonLakeFlyer:
I don't see how it hurts anything.

WickedShell:
Not opposed to it by any means, but if you enable packet signing you would
have timestamped packets already ;)

TSC21:
thanks!

@WickedShell adding the packet signing is something on the TODO list (of MAVROS).

PR #722, opened 2017-05-08
Move Set Camera mode to its own command

dogmaphobic:
Also, add *flicker* mode to camera settings


message_definitions/v1.0/common.xml
```diff
@@ -1395,6 +1395,13 @@
         <param index="2">0: No Action 1: Reset all settings</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
```
julianoes: This is now not consistent with the message CAMERA_SETTINGS, other than that it's fine.

DonLakeFlyer:
Perfect for me, thanks

PR #724, opened 2017-05-11
Add covariance to estimation msgs

TSC21:
Given some position estimation/filters maybe be feeding data into onboard position and attitude estimators, and since IEKF, LPE, EFK2 may use covariance matrices to improve the estimation, this PR allows sending these same covariances through specific messages (vision, mocap).

@jgoppert this may interest you as one can send the odometry data from external sources through these specific msgs and receive feedback though `LOCAL_POSITION_NED_COV` and `ATTITUDE_QUATERNION_COV` respectively.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3450,6 +3450,8 @@
       <field type="float" name="roll" units="rad">Roll angle in rad</field>
       <field type="float" name="pitch" units="rad">Pitch angle in rad</field>
       <field type="float" name="yaw" units="rad">Yaw angle in rad</field>
+      <extensions/>
+      <field type="float[21]" name="covariance">Pose covariance matrix upper right triangular (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
     </message>
     <message id="102" name="VISION_POSITION_ESTIMATE">
       <field type="uint64_t" name="usec" units="us">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
@@ -3459,12 +3461,16 @@
       <field type="float" name="roll" units="rad">Roll angle in rad</field>
       <field type="float" name="pitch" units="rad">Pitch angle in rad</field>
       <field type="float" name="yaw" units="rad">Yaw angle in rad</field>
+      <extensions/>
+      <field type="float[21]" name="covariance">Pose covariance matrix upper right triangular (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
     </message>
     <message id="103" name="VISION_SPEED_ESTIMATE">
       <field type="uint64_t" name="usec" units="us">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
       <field type="float" name="x" units="m/s">Global X speed</field>
       <field type="float" name="y" units="m/s">Global Y speed</field>
       <field type="float" name="z" units="m/s">Global Z speed</field>
+      <extensions/>
+      <field type="float[9]" name="covariance">Linear velocity covariance matrix (1st three entries - 1st row, etc.)</field>
     </message>
     <message id="104" name="VICON_POSITION_ESTIMATE">
       <field type="uint64_t" name="usec" units="us">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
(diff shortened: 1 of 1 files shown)


jgoppert:
Yeah, looks good, would be nice to have this in.

TSC21:
Is there anything else to be considered/modified so this can be merged?

jgoppert:
I haven't tested extensions out enough to approve this myself. Do you have any test showing this in use? That would probably help.

TSC21:
Have not tried myself but for what I know extensions exist so they can also be used in MAVLink 2.0. We can probably add a modification to the odom plugin in mavros and check its use, since it will be one of the main usages of this extensions (while a odom msg is not added to the msg API).

TSC21:
@LorenzMeier is there something that can be done to avoid this being staled forever? Thank you

LorenzMeier:
@TSC21 I'm sure I commented on this, but my original post might not have made it through: The issue is that MAVLink 2 is also used on radio links and your changes more than double the required bandwidth. This has most likely detrimental effects on a lot of users. Since MAVLink 2 sends zero-filled packets with reduced length we actually might get away with it but it is a complex topic. I think the better approach is to get this data over our new FastRTPS bridge to ROS instead.

TSC21:
Dear @LorenzMeier,

Thanks for your feedback.

Some matters about your comment:

1. I'm not seeing why people will be sending vision estimates through a radio stream. I can see it happen, for example, to the mocap cases, but that will be in cases where the tests are made in controlled environments and with the FCU radio and the GCS radio close to each other in the same room. For both cases: in vision I think the most common case is to send this same message from a companion board through a serial interface (USB<->UART), as you are probably aware. For mocap cases, I think there are many using wifi modules to do it (like the esp8266 interface that Pixracer uses), where the bandwidth and throughput can be much higher.

2. In any case, this PR can be changed so it doesn't get that heavy in terms of bandwidth, 'cause since the covariance matrices are symmetrical, I can change from a full covariance matrix to only use the URT (that means that a 6x6 matrix will become an array of 19 floats - instead of 36 - and a probably we can keep the 3x3, since it's only 9 floats). Would you accept the PR with that modification? Does it work for you?

3. Are you talking about https://github.com/eProsima/Firmware? I thought this wasn't mainstream already. But this is something to be consider only for ROS2, right? Regardless, this is still a matter that deserves a close follow up.

TSC21:
@julianoes can I get your input here? Thanks

julianoes:
I don't think I have an opinion. @mrivi can you check please?

TSC21:
@mrivi friendly ping

mrivi:
@TSC21 Sorry, I completely missed this. It looks good to me.

TSC21:
@mrivi do you think the matrices sizes shouls be reduces to fit the URT of the matrix instead of the full matrix?

mrivi:
@TSC21 yes, I think we should exploit the fact that the matrices are symmetrical and reduce the bandwidth

TSC21:
Yeah I'll do that. I'm currently travelling but as soon as I can, I will tackle this.

TSC21:
@LorenzMeier, @mrivi I updated the matrices so they use only the URT, given the fact the matrices are symmetrical.

message_definitions/v1.0/common.xml
```diff
@@ -3615,12 +3617,16 @@
       <field type="float" name="roll" units="rad">Roll angle in rad</field>
       <field type="float" name="pitch" units="rad">Pitch angle in rad</field>
       <field type="float" name="yaw" units="rad">Yaw angle in rad</field>
+      <extensions/>
+      <field type="float[21]" name="covariance">Pose covariance matrix upper right triangular (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
```
mrivi: Could we add also the attitude covariance?

mrivi:
@LorenzMeier  do you agree on bringing this changes in?

message_definitions/v1.0/common.xml
TSC21: The attitude quaternion? Cause the pose covariance already includes the RPY covariance.

mrivi: yeah, sorry. I didn't read carefully enough.

mrivi:
@LorenzMeier could you please comment?

TSC21:
Thanks @julianoes!

PR #727, opened 2017-05-16
Extend EXTENDED_SYS_STATE message to include flight stage and safety switch status

mhkabir:
@LorenzMeier This adds safety switch state to the EXTENDED_SYS_STATE message.

We use this to control various things like starting VIO when pre-armed and switching the companion system's wifi off in favour of a long range link.

This change shouldn't break anything for those currently using the message, as it is implemented as a v2 extension.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2312,6 +2312,18 @@
         <description>MAV currently landing</description>
       </entry>
     </enum>
+    <enum name="MAV_SAFETY_STATE">
+      <description>Enumeration of safety switch states</description>
+      <entry value="0" name="MAV_SAFETY_STATE_UNDEFINED">
+        <description>Safety switch not connected or state unknown</description>
+      </entry>
+      <entry value="1" name="MAV_SAFETY_STATE_ON">
+        <description>Safety is on and the MAV cannot be armed</description>
+      </entry>
+      <entry value="2" name="MAV_SAFETY_STATE_OFF">
+        <description>Safety is off and the MAV can arm</description>
+      </entry>
+    </enum>
     <enum name="ADSB_ALTITUDE_TYPE">
       <description>Enumeration of the ADSB altimeter types</description>
       <entry value="0" name="ADSB_ALTITUDE_TYPE_PRESSURE_QNH">
@@ -3886,6 +3898,8 @@
       <description>Provides state for additional features</description>
       <field type="uint8_t" name="vtol_state" enum="MAV_VTOL_STATE">The VTOL state if applicable. Is set to MAV_VTOL_STATE_UNDEFINED if UAV is not in VTOL configuration.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
+      <extensions/>
+      <field type="uint8_t" name="safety_state" enum="MAV_SAFETY_STATE">The safety state. Is set to MAV_SAFETY_STATE_UNDEFINED if safety state is unknown.</field>
     </message>
     <message id="246" name="ADSB_VEHICLE">
       <description>The location and information of an ADSB vehicle</description>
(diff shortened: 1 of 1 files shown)


WickedShell:
Is the MAV_MODE_FLAG_SAFETY_ARMED not sufficient for conveying this?

mhkabir:
@WickedShell Nope. We want to know the _pre-armed_ state here. i.e before arming/motors spinning, but after the safety switch has been pressed.

mhkabir:
@LorenzMeier Can we please get this in ?

AndreasAntener:
If we do modify `EXTENDED_SYS_STATE` we should pull in other potentials at the same time, like https://github.com/mavlink/mavlink/pull/736

LorenzMeier:
@mhkabir Could you please address the comment from @AndreasAntener ? Thanks!

mhkabir:
@LorenzMeier @AndreasAntener Done. Please merge when ready.

@WickedShell I've integrated your `flight_stage` extension into this PR.

WickedShell:
@mhkabir Great!

mhkabir:
@LorenzMeier ping

Kiwa21:
@LorenzMeier  @mhkabir 
Can we merge this now ? I don't see any conflicts.
I was going to PR a rtl_state field for this message, but the flight_stage take care of it and even better.

dagar:
Sorry to be the stick in the mud, but I still don't see how an unspecified field is a good direction for the Mavlink spec. Would including enums for the states really be that bad? Systems don't have to use/support all possible states.

WickedShell:
@dagar I agree with having an enum mapping, but I think it has to be an autopilot specific mapping, at which point it moves back into being essentially unspecified in the spec. An attempt to make a single enum that all projects add to will quickly outgrow the uint8_t space.

amilcarlucas:
This needs a rebase

dagar:
> An attempt to make a single enum that all projects add to will quickly outgrow the uint8_t space.

I'm not so sure, we've already gone this long without a flight_stage and a large portion would be common. We could also easily give it an extra byte. A small number of reserved custom slots could even exist, but so far I haven't heard of any examples that would be out of place.

I don't understand why we'd want the spec to evolve in ways that make it impossible to maintain compatibility across different ground control stations or flight controllers. What's the point of mavlink (serious question)?

WickedShell:
@dagar Because when you want to convey the touples like (landing VTOL, approaching target), (landing VTOL, transition to hover over target), (landing VTOL, acquired beacon), (landing, landed) this starts to fill up very very fast. These concepts won't map out all that well across various firmwares and quickly will exceed the uint8_t namespace, as well as add friction to adding new stages to report.

LorenzMeier:
The question is: What is the value of a field that's not well-specified in the protocol but in the common message set? To me, it seems that if people wanted to report very specific information they should just go ahead and create a custom message.

I'm on the FAA committee discussing the safe return / landing operation and even there it seems possible to agree on certain vehicle states and defined behaviour (like RTL/RTH definitions). If it can be done across an industry with various different protocols we should be able to do it here.

message_definitions/v1.0/common.xml
```diff
@@ -2312,6 +2312,18 @@
...
+    <enum name="MAV_SAFETY_STATE">
+      <description>Enumeration of safety switch states</description>
+      <entry value="0" name="MAV_SAFETY_STATE_UNDEFINED">
+        <description>Safety switch not connected or state unknown</description>
+      </entry>
+      <entry value="1" name="MAV_SAFETY_STATE_ON">
+        <description>Safety is on and the MAV cannot be armed</description>
+      </entry>
+      <entry value="2" name="MAV_SAFETY_STATE_OFF">
+        <description>Safety is off and the MAV can arm</description>
```
hamishwillee: The way it reads now the vehicle can arm, but might there not be other things preventing arming - even though the switch is off?

Perhaps

> Safety switch is off (not blocking arming).

And above

> Safety switch is on (arming is blocked).

hamishwillee:
What is the status here? It looks like the safety switch is non-controversial. Is it still needed? Perhaps just get that change in and deal with flight_stage separately?

amilcarlucas:
How about following @hamishwillee advice and merging this?

hamishwillee:
My general thinking is that if no stands up to champion a change then it is unlikely to be used and just adds to general clutter. I am therefore closing this issue. 

@mhkabir @dagar If you really want this please reopen, and either attend dev meeting to discuss or send a representative.

PR #730, opened 2017-05-24
RFC: Support for Wi-Fi settings

AlexeyBulatov:
In this PR I propose to add new messages for setup Wi-Fi of the vehicle using ground control station.
Description:
Vehicle can be in 2 state: connected to selected network or create it's own AP (access point). Vehicle will have list of networks to which it will try to connect alternately. If vehicle can't connect to any network then it will start in AP mode.
Communication with GCS:
First of all GCS request for saved Wi-Fi networks from vehicle using command MAV_CMD_REQUEST_WIFI_NETWORKS. After it get N (number of saved networks) messages WIFI_NETWORK_INFORMATION, using that it can create list in UI.
User now can do:
- Start vehicle in AP mode using MAV_CMD_WIFI_START_AP command.
- Connect to selected network from the list using WIFI_NETWORK_CONNECT message.
- Add new network (SSID, protocol, password) to saved networks using WIFI_NETWORK_ADD message.
- Delete selected network from saved networks using WIFI_NETWORK_DELETE message.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1580,6 +1580,16 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
+      <entry value="6000" name="MAV_CMD_REQUEST_WIFI_NETWORKS">
+        <description>Request saved Wi-Fi networks (WIFI_NETWORK_INFORMATION).</description>
+        <param index="1">0: No Action 1: Request messages with wifi network information</param>
+        <param index="2">Reserved (all remaining params)</param>
+      </entry>
+      <entry value="6001" name="MAV_CMD_WIFI_START_AP">
+        <description>Start in AP (Access Point) mode.</description>
+        <param index="1">0: No Action 1: Start in AP mode</param>
+        <param index="2">Reserved (all remaining params)</param>
+      </entry>
       <!-- VALUES FROM 0-40000 are reserved for the common message set. Values from 40000 to UINT16_MAX are available for dialects -->
       <!-- BEGIN of payload range (30000 to 30999) -->
       <entry value="30001" name="MAV_CMD_PAYLOAD_PREPARE_DEPLOY">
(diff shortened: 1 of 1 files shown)


LorenzMeier:
@dogmaphobic Can you chime in here on the ESP8266 experience?

LorenzMeier:
@dogmaphobic Could you please provide feedback?

dogmaphobic:
I don't understand how you can send a message to discover how to connect so you can send messages.

Assuming this is all for subsequent communication, there is already an interface for most of this functionality using parameters. I'm not averse to add/replace the existing interface. I'm just curious how you connect the first time so you can send messages?

AlexeyBulatov:
> I'm just curious how you connect the first time so you can send messages?

`If vehicle can't connect to any network then it will start in AP mode.` So during first connection vehicle will start in Access Point mode, because list of saved networks is empty. Further we can connect to autopilot by predefined ip or name. And after you can send messages described above.

dogmaphobic:
Ok, from the context of simply controlling Wi-Fi association, these messages are sufficient provided a DHCP server is guaranteed to exist. It doesn't handle addressing, routing, or anything. If there is no DHCP, how are addresses going to be assigned?

In addition, what I wonder is what use case do you see for these? A flight controller alone will not likely have a network stack. The network stack would most likely be a separate controller, which would act as a bridge of some sort. Most likely an Ethernet/UART bridge. These messages allow you to define how this bridge connects to the network (Wi-Fi association) but it doesn't handle the connection between itself and flight controller. Should it?

TSC21:
8I suppose that @AlexeyBulatov is trying to use Mavlink msgs (Application Level) to configure specifications at lower levels. But as you say @dogmaphobic, unless there's a bridge to a module/radio with a network stack within, the msgs won't make much of a sense, unless for example we are working in an Ad-hoc network, which is decentralized and does not rely on a pre-existing infrastructure.
But also systems capable of interacting on a Ad-hoc network usually have their own ways of communication throughout is different network layers, but it could be interesting to see how could Mavlink be used in a network like that.

dogmaphobic:
>So during first connection vehicle will start in Access Point mode, because list of saved networks is empty.

For this to work, when the vehicle first starts and it is not configured, it would start in *Open* AP mode as there is no way to tell what the password is. The SSID would also have to be known so the user can associate the CGS's device to the vehicle's AP. Assuming this is resolved, we need one more message:

```
<message id="304" name="WIFI_CONFIG_AP">
    <description>WIP: Configure AP SSID and Password. Leave password blank for an Open AP.</description>
    <field type="char[32]" name="ssid">Name of Wi-Fi network</field>
    <field type="char[64]" name="password">Password</field>
</message>
```

Where you would set the SSID and the password for the vehicle's AP.

I would imagine this will be the most used case. Having the vehicle offering an AP as opposed to have it connected to an external AP. At boot time, the vehicle should quickly see if there is any configured network. Not finding any, it should go immediately into AP mode.

LorenzMeier:
I've merged https://github.com/mavlink/mavlink/pull/765 - can we progress further here or should I close the PR?

AlexeyBulatov:
I will update this PR.

LorenzMeier:
@AlexeyBulatov I'm closing this as stale - please open a new PR once you updated. I'm very happy to get it in and will support you in doing so, I just need to make sure to have no stale PRs in the queue.

PR #731, opened 2017-05-28
Common: Allow CMD_DO_LAND_START to change the flight mode

WickedShell:
Useful for simplifying the number of commands and state machine used to safely transition a vehicle to flying a landing sequence. An example use case would be changing the flight mode to AUTO for ArduPlane.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -993,7 +993,7 @@
       </entry>
       <entry value="189" name="MAV_CMD_DO_LAND_START">
         <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts. It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used. The Latitude/Longitude is optional, and may be set to 0 if not needed. If specified then it will be used to help find the closest landing sequence.</description>
-        <param index="1">Empty</param>
+        <param index="1">1.0 indicates if the vehicle should change into a flight mode where it can perform the landing sequence. Only applicable if sent as part of a COMMAND_LONG</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>


magicrub:
I don't understand what you're​ trying to do here

DonLakeFlyer:
My take on this one is that if the vehicle needs to go into a specific flight mode to execute this command then it should just do it. If this is inappropriate at the time, then it just fails. Seems simpler.

WickedShell:
@magicrub If I want to jump to a landing start right now I need to:

1. Send MAV_CMD_DO_LAND_START
2. Wait for an ACK/NACK and retry till I get it
3. Send a mode change
4. Wait for an ACK/NACK/mode change

The problem with this is it makes it take much longer to achieve the desired behavior in a poor telemetry situation, and as @DonLakeFlyer said, I've asked the vehicle to prep to fly a landing, I should be able to just instruct it to fly a landing in one pass.

magicrub:
So this is not the mission item version?

WickedShell:
This is the same command you plan in a mission, however a mission would ignore the param1 field (as it's obviously already in the correct mode if it's executing it), but when sent as a COMMAND_LONG you would evaluate the param1 field as a mode change request.

magicrub:
We had talked about a configure land mission cmd, but that was for auto flight. Could the same type of new command be used instead of modifying this one?

WickedShell:
Configure landing was discussed in the context of keepouts/approaches. It was never discussed as a way to start the sequence. Given that this is the existing preferred method to setup a mission for actually being a landing, I'd rather stick with it as the interface for doing so, rather then having two different interfaces that overlap in functionality.

dagar:
Is the behaviour of sending MAV_CMD_DO_LAND_START without a mode change actually desired?

WickedShell:
It can be useful to prep a mission before you actually want to fly it, so I'd say there is merit to not always forcing the mode change.

DonLakeFlyer:
> It can be useful to prep a mission before you actually want to fly it, so I'd say there is merit to not always forcing the mode change.

I don't get that. Can you explain how that would be used?

WickedShell:
@DonLakeFlyer It's a really rare scenario. I've normally performed this with just set-wp commands, and while the aircraft was in guided (or being flown manually) I've tweaked the target mission waypoint, and then waited until the aircraft was aligned with the flight path before swapping back to auto. I'm happy to let go of this use case, but I thought it was more valuable to have the flexibility.

LorenzMeier:
@WickedShell Based on the discussion here: Is this really needed or not? If possible we should avoid complex special cases and keep the execution of a mission as linear and standard as possible.

LorenzMeier:
Closing as stale - please re-open a new rebased PR if this is still needed.

PR #732, opened 2017-06-02
common.xml: added message DEBUG_ARRAY_FLOAT

jlecoeur:
Adds the message DEBUG_ARRAY. 
This is similiar to DEBUG_VECT, but with 60 entries instead of 3. It basically fills the maximum payload.
I have been using it for long and it proved really useful to debug large float arrays.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3963,7 +3963,6 @@
       <field type="uint8_t" name="ind">index of debug variable</field>
       <field type="float" name="value">DEBUG value</field>
     </message>
-
     <!-- messages with ID 256 and above are only available in MAVLink2 -->
     <message id="256" name="SETUP_SIGNING">
       <description>Setup a MAVLink2 signing key. If called with secret_key of all zero and zero initial_timestamp will disable signing</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4129,7 +4129,7 @@
     </message>
     <message id="350" name="DEBUG_ARRAY_FLOAT">
       <description>Large debug array. The use of this message is discouraged for normal packets, but a quite efficient way for testing new messages and getting experimental debug output.</description>
-      <field type="uint32_t" name="time_boot_ms">Timestamp (milliseconds since system boot)</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
       <field type="uint8_t" name="id">Unique ID used to discriminate between arrays</field>
       <field type="char[10]" name="name">Name, for human-friendly display in QGC</field>
       <field type="float[60]" name="data">data</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3958,6 +3958,7 @@
       <field type="uint8_t" name="ind">index of debug variable</field>
       <field type="float" name="value">DEBUG value</field>
     </message>
+
     <!-- messages with ID 256 and above are only available in MAVLink2 -->
     <message id="256" name="SETUP_SIGNING">
       <description>Setup a MAVLink2 signing key. If called with secret_key of all zero and zero initial_timestamp will disable signing</description>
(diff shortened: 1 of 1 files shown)


TSC21:
I agree with @magicrub. You can use `<extensions/>`

jlecoeur:
@magicrub @TSC21 Yes I would be fine with something like this:
```xml
<message id="250" name="DEBUG_VECT">
  <field type="char[10]" name="name">Name</field>
  <field type="uint64_t" name="time_usec" units="us">Timestamp</field>
  <field type="float" name="x">x</field>
  <field type="float" name="y">y</field>
  <field type="float" name="z">z</field>
  <extensions/>
  <field type="float[56]" name="data">Additional data</field>
</message>
```
But isn't there a problem regarding the size of packets on the network? If I understand correctly how `<extensions/>` work, the extended fields are sent on the network by v2.0 but not by v1.0. 
If that is right, then:
- the size of `DEBUG_VECT` messages would be radically different on v1.0 and v2.0
- people on v2.0 who use `DEBUG_VECT` as a lightweight way to send 3 floats will have to pay the cost of 56 more floats

magicrub:
I think you're right. How about you change the index to 350 (requires v2) and while you're at it change the name to DEBUG_VECT_BIG

magicrub:
actually, a char[10] seems a bit short too.

WickedShell:
It doesn't consume the cost of the 56 other floats as MAVLink2 drops all 0 bytes off the end of the packet, it makes the overhead very low. Does require MAVLink2 to get access to the 56 fields, but you should be pushing to MAVLink2 anyways.

magicrub:
oh yes, good point! mavlink 2 is zero padded so it won't consume the bandwidth unless it's needed. Let's stick to extending the current packet

TSC21:
Btw is this being used in any firmware (PX4, Ardupilot,...) ATM?

TSC21:
Ok I'm seeing this on DJI SDK and Mikrokopter also

jlecoeur:
Thanks for the input @WickedShell , very good feature that all trailing 0 bytes are dropped! Ok for extending the original message.

@TSC21 will it break DJI SDK and Mikrocopter?

TSC21:
@jlecoeur I can't tell but if this is an extension only I suppose not. A change in the current API could break but as it is an addiction to the msg, I think it is clear of problems.

TSC21:
Btw are you using this with PX4 or APM? Can you tell us what is the end app for this?

jlecoeur:
We were using this message on a custom firmware, and now PX4. 
The applications were to display on QGC kalman matrices, intermediate algorithm states, and sensor measurements. Right now I use it to prototype the communication between a PX4Flow and the autopilot, it is super convenient to be able to change what is sent by the sensor without modifying the message definition and re-generating C headers on both sides.

TSC21:
> Right now I use it to prototype the communication between a PX4Flow and the autopilot, it is super convenient to be able to change what is sent by the sensor without modifying the message definition and re-generating C headers on both sides.

Why don't you propose extensions to current msg definitions so other can benefit from that? That way we can all benefit from other fields on the msg that you think that can be useful.

magicrub:
adding an index would be nice too so it's machine friendly to sort debug messages instead of a string.

jlecoeur:
> Why don't you propose extensions to current msg definitions so other can benefit from that? That way we can all benefit from other fields on the msg that you think that can be useful.

It does not send optical flow + distance like stock PX4Flow, so it does not fit the current msg even with extensions. As I said, when developing it, it is more convenient not to rewrite msg definitions each time the kind of data sent by the sensor changes. But of course if it ends up as something useful, I will create/extend the appropriate messages and share it!

@magicrub I removed BIG_DEBUG_VECT and added the extensions to DEBUG_VECT. I also added an index, and the number of floats filled in the vector.

TSC21:
> But of course if it ends up as something useful, I will create/extend the appropriate messages and share it!

Thank you very much! We will be waiting for those and we appreciate it!

vooon:
Actually i more like new message in v2 msgid space, than extending v1 messages.
With `<extensions />` you also disable any version checks for that new fields, so it is possible to racv incorrect data.

LorenzMeier:
@jlecoeur Can you put a new message into the v2 namespace? The way you've extended it also makes it hard to iterate on. I think that's more consistent than extending the previous spec. It would also be great if you could send a PR for the Firmware and QGC.

PR #733, opened 2017-06-07
UAVCAN node status message

pavel-kirienko:
Required for https://github.com/PX4/Firmware/issues/6940.

@mhkabir @eyeam3 please review.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2608,7 +2608,7 @@
     <message id="8" name="UAVCAN_NODE_STATUS">
       <description>General status information of an UAVCAN node. Please refer to the definition of the UAVCAN message "uavcan.protocol.NodeStatus" for the background information. The UAVCAN specification is available at http://uavcan.org.</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since UNIX epoch or microseconds since system boot)</field>
-      <field type="uint32_t" name="uptime_sec">The number of seconds since the start-up.</field>
+      <field type="uint32_t" name="uptime_sec" units="s">The number of seconds since the start-up.</field>
       <field type="uint8_t" name="health" enum="UAVCAN_NODE_HEALTH">Generalized node health status.</field>
       <field type="uint8_t" name="mode" enum="UAVCAN_NODE_MODE">Generalized operating mode.</field>
       <field type="uint8_t" name="sub_mode">Not used currently.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2605,26 +2605,6 @@
       <description>Emit an encrypted signature / key identifying this system. PLEASE NOTE: This protocol has been kept simple, so transmitting the key requires an encrypted channel for true safety.</description>
       <field type="char[32]" name="key">key</field>
     </message>
-    <message id="8" name="UAVCAN_NODE_STATUS">
-      <description>General status information of an UAVCAN node. Please refer to the definition of the UAVCAN message "uavcan.protocol.NodeStatus" for the background information. The UAVCAN specification is available at http://uavcan.org.</description>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since UNIX epoch or microseconds since system boot)</field>
-      <field type="uint32_t" name="uptime_sec" units="s">The number of seconds since the start-up.</field>
-      <field type="uint8_t" name="health" enum="UAVCAN_NODE_HEALTH">Generalized node health status.</field>
-      <field type="uint8_t" name="mode" enum="UAVCAN_NODE_MODE">Generalized operating mode.</field>
-      <field type="uint8_t" name="sub_mode">Not used currently.</field>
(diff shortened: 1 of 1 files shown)


pavel-kirienko:
@mhkabir Note that this message assumes that the component ID field of outgoing messages will be equal to the node ID of the UAVCAN node whose status we're reporting. The same approach is used when reporting the node configuration parameters via `PARAM_VALUE` and such.

message_definitions/v1.0/common.xml
```diff
@@ -2570,6 +2605,14 @@
       <description>Emit an encrypted signature / key identifying this system. PLEASE NOTE: This protocol has been kept simple, so transmitting the key requires an encrypted channel for true safety.</description>
       <field type="char[32]" name="key">key</field>
     </message>
+    <message id="8" name="UAVCAN_NODE_STATUS">
+      <description>General status information of an UAVCAN node. Please refer to the definition of the UAVCAN message "uavcan.protocol.NodeStatus" for the background information. The UAVCAN specification is available at http://uavcan.org.</description>
```
mhkabir: Please add a timestamp field : `uint64_t time_usec`.

pavel-kirienko: Sure. Could you please elaborate why is it necessary though?

mhkabir: All messages should be timestamped, to e.g detect timeouts or know when the node last pinged us.

pavel-kirienko: Done.

mhkabir:
I wonder if it makes sense to move this to the MAVLink2 message space ( > 255)

pavel-kirienko:
> I wonder if it makes sense to move this to the MAVLink2 message space ( > 255)

It probably does, but I will desist from further comments here.

What this PR doesn't yet account for is transferring the extended node information, such as its name, the software and the hardware version numbers, and so on. This information is vital for many sensible use cases, so we should not ignore it.

Extending the node status message with these additional fields seems unreasonable because of the bandwidth considerations. So how do we approach this? I see two possible solutions:

1. Implement a new MAVLink message with the required fields. Send this message at a very low rate *and* upon a change of the reported parameters.
2. Implement the above message and a request message for it. In this workflow, the ground station would broadcast a request for data, which the autopilot would respond to with a sequence of these extended information messages, one per each known node.

pavel-kirienko:
I decided to go with the first approach and added a new message `UAVCAN_NODE_INFO`.

eyeam3:
The message definitions look good to me.

> 1. Implement a new MAVLink message with the required fields. Send this message at a very low rate and upon a change of the reported parameters.
> 2. Implement the above message and a request message for it. In this workflow, the ground station would broadcast a request for data, which the autopilot would respond to with a sequence of these extended information messages, one per each known node.
I would prefer the second approach as it doesn't transmit the same data repeatedly. But your proposed message would still be necessary for both cases. 

I support the usage of Mavlink 2.0 as well.

pavel-kirienko:
> I would prefer the second approach as it doesn't transmit the same data repeatedly. But your proposed message would still be necessary for both cases.

This is true, but the first approach is also much easier to implement. Can we adjust the transmission frequency dynamically, based on whether or not the system is armed?

Besides, there is one aspect that I missed in my above messages: in the second approach, the ground station would still need to periodically send request messages, at least while the respective GUI elements are displayed, which complicates things too.

mhkabir:
>  Can we adjust the transmission frequency dynamically, based on whether or not the system is armed?

Yes, but it's actually easier to implement the request-and-receive protocol.

> I support the usage of Mavlink 2.0 as well.

Yes, @pavel-kirienko please move the messages into the v2 message space.

pavel-kirienko:
@mhkabir @eyeam3 Done. The message definitions have been moved into the v2 space; new message `UAVCAN_NODE_INFO_REQUEST` has been added. It is also allowed for remote systems to emit messages `UAVCAN_NODE_INFO` asynchronously at their own discretion, e.g. when a new UAVCAN node appears online.

mhkabir:
@pavel-kirienko `UAVCAN_NODE_INFO_REQUEST` should not be a message. It should be a command. (`MAV_CMD` : https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L598)

pavel-kirienko:
Fixed.

mhkabir:
@LorenzMeier Please review these messages that we need to transport UAVCAN status to the GCS side.

LorenzMeier:
Looks good. Can you please also add the documentation for it here: https://www.gitbook.com/book/mavlink/mavlink-devguide/details

I'm trying to enforce a policy that only documented messages get merged, as we otherwise end up with undocumented sub-protocols all the time. Its also important for the spec of the messages themselves because the documentation tends to identify inconsistencies in the spec.

pavel-kirienko:
Is this ready to be merged now?

pavel-kirienko:
I would be happy to help if you told me what needs to be done before this PR can be merged.

pavel-kirienko:
@eyeam3 @mhkabir @LorenzMeier anybody?

LorenzMeier:
Yep. Let me bring that in

PR #734, opened 2017-06-07
common.xml: added message DEBUG_FLOAT_ARRAY

jlecoeur:
This PR adds the message DEBUG_ARRAY. This is a follow up of #732 with clean history.

I have been using this message for long and it proved really useful to debug large float arrays.
The message is similar to DEBUG_VECT, but with 60 entries instead of 3. It basically fills the maximum payload with a packet size of 263 bytes.

@LorenzMeier I placed it in v2 namespace, using id 350 as recommended by @magicrub. PRs for PX4 Firmware and QGC are coming.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4774,8 +4774,7 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
-    <message id="350" name="DEBUG_ARRAY">
-      <description>Large debug array. The message fills the maximum payload with a packet size of 263 bytes.</description>
+    <message id="350" name="DEBUG_FLOAT_ARRAY">
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
       <field type="uint8_t" name="array_id">Unique ID used to discriminate between arrays</field>
       <field type="char[10]" name="name">Name, for human-friendly display in a Ground Control Station</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4775,6 +4775,7 @@
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
     <message id="350" name="DEBUG_FLOAT_ARRAY">
+      <description>Large debug/prototyping array. The message uses the maximum available payload for data. The array_id and name fields are used to discriminate between messages in code and in user interfaces (respectively). Do not use in production code.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
       <field type="uint8_t" name="array_id">Unique ID used to discriminate between arrays</field>
       <field type="char[10]" name="name">Name, for human-friendly display in a Ground Control Station</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4776,11 +4776,11 @@
     </message>
     <message id="350" name="DEBUG_FLOAT_ARRAY">
       <description>Large debug/prototyping array. The message uses the maximum available payload for data. The array_id and name fields are used to discriminate between messages in code and in user interfaces (respectively). Do not use in production code.</description>
-      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
-      <field type="uint8_t" name="array_id">Unique ID used to discriminate between arrays</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
       <field type="char[10]" name="name">Name, for human-friendly display in a Ground Control Station</field>
+      <field type="uint16_t" name="array_id">Unique ID used to discriminate between arrays</field>
       <extensions/>
-      <field type="float[60]" name="data">data</field>
+      <field type="float[58]" name="data">data</field>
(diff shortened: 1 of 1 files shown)


vooon:
I think that you miss `data_len` field.

jlecoeur:
@vooon The message is already 263 bytes long so we cannot add fields. Plus 60 is a nice round number so I would rather not reduce the size of the data array :-)
I added a field `data_len` in the previous PR because it was an optional array in the `<extensions/>` section. Thus it was a way to avoid issues with mavlink versions: if the sender used v1.0, then data_len would be 0, meaning that the optional array should not be used. Now, I think `data_len` is unnecessary.

magicrub:
I had suggested 350 because the other message in the other thread was 250. Nothing special about it other than that.

magicrub:
I can see how a length variable would be helpful but if you're generating debug data then it's proprietary and so you know exactly what you're sending so the need is low.

jlecoeur:
Here is the commit for integration in PX4: https://github.com/lis-epfl/PX4Firmware/commit/5042b7f6d1cbfee29023769557828c2cefe196d0 (not in https://github.com/PX4/Firmware/issues/7373 yet)

jlecoeur:
And the QGC pull request is cited just above. I tested with the example app `px4_mavlink_debug` and it works fine. It should be good to go @LorenzMeier !

jlecoeur:
A little update, please let me know if this can move forward.

TSC21:
@jlecoeur this makes sense to me.

TSC21:
@jlecoeur can you rebase on master?

@LorenzMeier does this look good to you? I would like to have this in and help @jlecoeur with extending the support on PX4 and MAVROS.

jlecoeur:
Thanks @TSC21, I rebased.

TSC21:
@LorenzMeier is there any reason this would be staled forever?

TSC21:
@jlecoeur can you rebase it ITMT?

jlecoeur:
@TSC21 done

jlecoeur:
rebased

TSC21:
@LorenzMeier @jlecoeur should we update this?

message_definitions/v1.0/common.xml
```diff
@@ -4543,5 +4543,12 @@
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure in centimeters</field>
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure in centimeters</field>
     </message>
+    <message id="350" name="DEBUG_ARRAY">
+      <description>Large debug array. The use of this message is discouraged for normal packets, but a quite efficient way for testing new messages and getting experimental debug output.</description>
```
hamishwillee: In MAVLink 2 array truncation means that only the data that is used will be sent, so defining a full width debug packet makes sense (and to my mind, only defining a "partial use" of the payload does not).

What does this mean?

>The use of this message is discouraged for normal packets

Also I can see how you might want to get experimental debug output, but how does this help design new messages?

>, but a quite efficient way for testing new messages and getting experimental debug output.

Anyway, suggest a little more clarify on purpose.

message_definitions/v1.0/common.xml
```diff
@@ -4543,5 +4543,12 @@
...
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
+      <field type="uint8_t" name="id">Unique ID used to discriminate between arrays</field>
```
hamishwillee: Maybe `array_id` or `debug_id`? We're specific in naming in most other places.

message_definitions/v1.0/common.xml
jlecoeur: > What does this mean?
>   > The use of this message is discouraged for normal packets

Do not use in production.


> but how does this help design new messages?

Before creating a new message which content is not 100% defined yet, you can fill your data in a DEBUG_ARRAY. The advantage is that you can easily experiment with the content of the future message by updating the content of a DEBUG_ARRAY without updating MAVLink headers on tx and rx systems. When the message content is fixed, you need to create a new MAVLink message and update the headers on all systems just *once*.

message_definitions/v1.0/common.xml
jlecoeur: done

jlecoeur:
@TSC21 @hamishwillee I rebased and addressed the comments.

jlecoeur:
Thanks @hamishwillee I updated the description.

@LorenzMeier thanks for the comments. I could make sense of the three first points but I cannot decipher the fourth one, can you elaborate?

hamishwillee:
@jlecoeur 

Thanks for your fixes! Would timestamp be more useful as microseconds (question, not "request")?

>> Make sure that message name and hit an alignment increment

> I could make sense of the three first points but I cannot decipher the fourth one, can you elaborate?

Probably not helpful because I don't grok this and @LorenzMeier will have to confirm ....

This is to do with [data alignment](https://en.wikipedia.org/wiki/Data_structure_alignment). Apparently if the fields are misaligned with respect to natural word boundaries then more code needs to be generated to make everything pack and unpack properly. 

The bytes taken for each fields are shown below.
* 4 "uint32_t" timestamp
* 1 "uint8_t" array_id
* 1 char[10]" name

I'm guessing that the natural alignment is 4 bytes, and Lorenz suggesting that the name, id etc be extended so that the data is aligned.

jlecoeur:
@hamishwillee ms and us are both used, I switched to microseconds.

I was not aware that data alignment was a constraint when creating a message, it is fine like this?

jlecoeur:
rebased

hamishwillee:
> I was not aware that data alignment was a constraint when creating a message, it is fine like this?

I wasn't either. I think it is a case where you "try" to align, knowing there is a cost of some extra code generated if you don't. I suspect it is right now. Depends on what the word size is.

@LorenzMeier what boundary  are we trying to pack this on/is it right now?

TSC21:
@jlecoeur given this was merged, can you please continue with https://github.com/PX4/Firmware/issues/7770 and add support to `DEBUG_FLOAT_ARRAY` on PX4? Thanks

PR #735, opened 2017-06-08
camera: separate commands and settings

julianoes:
It turns out, it's a bad idea to have the resolution settings in the
same command as the start recording / capture command. This is because
both, the setting or the action, can fail and the result won't tell
which one did. Also applying a setting can take some time and can delay
the command to take a picture / record a video which should be
instantanteous.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1405,9 +1405,14 @@
         <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
         <param index="2">Camera mode (0: photo mode, 1: video mode)</param>
         <param index="3">Audio recording enabled (0: off 1: on)</param>
-        <param index="4">Photo resolution ID (20 Megapixel, 10 Megapixel, etc.)</param>
-        <param index="5">Video resolution and rate ID (4K 60 Hz, 4K 30 Hz, HD 60 Hz, HD 30 Hz, etc.)</param>
-        <param index="6">Reserved (all remaining params)</param>
+        <param index="4">Reserved (all remaining params)</param>
+      </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_SETTINGS_3">
+        <description>WIP: Set the camera settings part 3 (CAMERA_SETTINGS). Use NAN for values you don't want to change.</description>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="2">Photo resolution ID (20 Megapixel, 10 Megapixel, etc.)</param>
+        <param index="3">Video resolution and rate ID (4K 60 Hz, 4K 30 Hz, HD 60 Hz, HD 30 Hz, etc.)</param>
+        <param index="4">Reserved (all remaining params)</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
         <description>WIP: Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1405,15 +1405,16 @@
         <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
         <param index="2">Camera mode (0: photo mode, 1: video mode)</param>
         <param index="3">Audio recording enabled (0: off 1: on)</param>
-        <param index="4">Reserved (all remaining params)</param>
+        <param index="4">Photo resolution ID (20 Megapixel, 10 Megapixel, etc.)</param>
+        <param index="5">Video resolution and rate ID (4K 60 Hz, 4K 30 Hz, HD 60 Hz, HD 30 Hz, etc.)</param>
```
lbegani: Shouldn't the photo/video resolution be written in MAV_CMD_SET_CAMERA_SETTINGS instead of MAV_CMD_SET_CAMERA_MODE

julianoes: Good point. I think I agree. I will move it to MAV_CMD_SET_CAMERA_SETTINGS_3 probably.

dogmaphobic: I get that there were empty spaces in `MAV_CMD_SET_CAMERA_MODE` and it is not completely illogical to have the resolution there but `MAV_CMD_SET_CAMERA_SETTINGS_3` might be a bit more obvious. I'm fine either way.

julianoes: Changed in 64351c7.

lbegani:
+1. 
Corresponding changes done in MAVLink users like QGC, FlightStack etc ?

julianoes:
> Corresponding changes done in MAVLink users like QGC, FlightStack etc ?

QGC can change once the mavlink headers get updated.

julianoes:
@dogmaphobic let me know when you think this is good to go in.

message_definitions/v1.0/common.xml
```diff
@@ -1407,13 +1407,19 @@
         <param index="3">Audio recording enabled (0: off 1: on)</param>
         <param index="4">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_SETTINGS_3">
+        <description>WIP: Set the camera settings part 3 (CAMERA_SETTINGS). Use NAN for values you don't want to change.</description>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="2">Photo resolution ID (20 Megapixel, 10 Megapixel, etc.)</param>
```
dogmaphobic: I would add a -1 for *Max Resolution*, which would be used if you don't know what the options are. The options would also most likely be actual resolutions (4000 x 3000, etc.) rather than *Megapixel*.

message_definitions/v1.0/common.xml
```diff
@@ -1407,13 +1407,19 @@
...
+        <param index="3">Video resolution and rate ID (4K 60 Hz, 4K 30 Hz, HD 60 Hz, HD 30 Hz, etc.)</param>
```
dogmaphobic: Same idea, have an option of -1 for max resolution and frame rate when you don't know what the options are.

dogmaphobic:
I've added two comments that pertain to the comment section of the definition only, not the implementation. Other than that, I'm ready when you are.

julianoes:
@dogmaphobic I've fixed what you suggested in 933acf7.

dogmaphobic:
Perfect. Let's get this in.

PR #736, opened 2017-06-11
Common: Add a stage option to EXTENDED_SYS_STATE

WickedShell:
This replaces #684, allows autopilots to report custom internal stages associated with a flight mode.


WickedShell:
@LorenzMeier @DonLakeFlyer @dagar are you guys good with this?

dagar:
I personally think it's a step in the wrong direction, but if you believe it's the best solution to your problem I won't object.

DonLakeFlyer:
My only comment is that it would seem to belong here. Whether it is a good thing or now I have no opinion.

WickedShell:
I'm strongly of the opinion that it's the best approach at this point, or any proposed alternative to date. The only other solution would be to have a very long list of enums, much the way we have CMD's that each project can update as needed, but every GCS would quickly get into the stage of having to interpret the modes differently based on what autopilot they are talking to.

mhkabir:
I've added this to https://github.com/mavlink/mavlink/pull/727.

PR #737, opened 2017-06-15
Generate common message specs in gitbook format

hamishwillee:
Script generates a file message-definitions.md that can be imported into a gitbook. This depends on XSL file being added to master.

### DIFF
--- a/doc/README
+++ b/doc/README
@@ -1,6 +1,16 @@
 MAVLink Micro Air Vehicle Message Marshalling Library
 
-The mavlink_to_html_table.xsl file is used to transform the MAVLink XML into a human-readable HTML table for online documentation.
+Files:
+
+- **Doxyfile**: -
+- **MAVLink2.md**: MAVLink2 Documentation
+- **README**: This file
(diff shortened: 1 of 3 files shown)


hamishwillee:
@dagar OK, so this is PR to get the common message specs into gitbook format.

Can I leave you with the job of running this script when needed and copying the generated file into the english tree [here](https://github.com/mavlink/mavlink-devguide/tree/master/en), master branch ?

dagar:
Sure, but let's make some simple notes for all these gitbook imports.

hamishwillee:
@dagar Good idea. I've added a readme.

Are we likely to have more message definitions or related commands?
If so, then the output should perhaps be renamed to "common.md" or "common_definitions.md" and when copied into the gitbook moved to **/message_definitions/common.md**

hamishwillee:
@bkueng Just Python 3. Is that OK?

bkueng:
I'd add python 2 support as well, since it's still in use (preferred), or make sure python 3 is used: `#! /usr/bin/env python3`.

hamishwillee:
@bkueng Thanks. Testing shows it works for both :-) I added a note in the docs section of the script. Merging this now.

PR #739, opened 2017-06-18
Update mavlink_gitbook.py

hamishwillee:


hamishwillee:
Added documentation into the gitbook script about compatibility with Py2/Py3

PR #740, opened 2017-06-19
add: MAV_DISTANCE_SENSOR_RADAR & MAV_DISTANCE_SENSOR_UNKNOWN

khancyr:


TSC21:
Very useful. But analog units? They must have some kind of eletromagnetic signal rx/tx. So the type should be according to this type of signal

khancyr:
@TSC21 The problem come from ardupilot, for pwm and analog sensors we can't fix a type or auto deduce it.

TSC21:
You can't auto-deduce the sensor? If you know what is connected to the board, can't you just tell to the FCU what sensor is that?

OXINARF:
@TSC21 How can you tell what an analog or a PWM sensor type is? It is just an analog or a PWM signal, any rangefinder type can give those outputs.

TSC21:
A rangefinder type is according to the electromagnetic wave type that it uses to compute the range. The type is not the type of signal it uses to comunicate with the FCU, from what I know. So being analog or PWM, if it uses IR, then the type is `INFRARED`.

OXINARF:
I absolutely agree with you. Which is why I still don't know how you propose to auto-deduce the sensor type from an analog sensor? If you just have an analog signal, how can code know if it is infrared, laser or any other type?

TSC21:
When I mean auto deduce I'm saying the user. What I'm saying is that the user can tell the FCU what type of sensor that is, through a parameter or so.
Btw how are you extrapolating ovet the type of sensor automatically? How is Ardupilot doing that?

TSC21:
(@OXINARF saudações lusas :D )

khancyr:
@TSC21 Each driver got a fixed type according to the sensor documentation, this isn't modifiable.

TSC21:
> @TSC21 Each driver got a fixed type according to the sensor documentation, this isn't modifiable.

Each driver? You mean the rangefinder drivers? So if there is no driver for an analog rangefinder, it should be set as an `UNKNOWN` type? Is that it?

khancyr:
Yep rangefinder drivers ! 
And analog and pwm type are set to UNKNOWN

TSC21:
> And analog and pwm type are set to UNKNOWN

Assuming you connect an PWM rangefinder to an AUX port. How are you going to read the values if not through a driver? If you program the FCU to read values from PWM and publish them in the FCU stream, can't you manually state what is the type of sensor without the FCU automatically defaults it to `UKNOWN`?

khancyr:
analog and pwm are drivers (see https://github.com/ArduPilot/ardupilot/tree/master/libraries/AP_RangeFinder). But material drivers. Each driver correspond to one sensor and not a type of sensors.
Ardupilot make no use of the sensor_type parameter (radiation type etc). But to improve compatiblity with ROS, I added the sensors type to each sensor we support. That way the DISTANCE SENSOR message is fill correctly with the right sensors type for each sensor.

It was chose to not create new parameters to handle the sensor_type. It should be simplier for user (as they generally don't know what type they use), for GCS (no new box to fill), ardupilot as we don't need to add a parameter for the number for rangefinder supported (currently we support up to 4 sensors)

TSC21:
@khancyr ok sure. Makes sense then.

PR #745, opened 2017-06-29
Better define reserved argument handling and response for Storage Formatting.

dogmaphobic:
* Clarifying what to do with reserved values
 * Clarify that a STORAGE_INFORMATION message is sent once formatting storage is completed.


LorenzMeier:
@magicrub Thanks for the review!

PR #746, opened 2017-06-30
TRIGG_DIST: Support trigger before/after distance trigger start/stop

DonLakeFlyer:
The reason for these changes is to make it way easier to create survey missions which have consistent coverage over a polygonal area. The basic idea is to modify the TRIGG_DIST command such that:
* When you start distance triggering the first image is taken right away (optional) instead of waiting until the first distance passes.
* When turning off distance triggering you can also take one last image at that point (optional).

This makes it possible to simply set up missions which generate transects that have full coverage from the entrance of the survey area to the exit of the survey area. Specifically for the case where images are not taken in the turnarounds. But this also helps when taking images throughout an entire survey. Since it will close up possible holes in the coverage at the entrance and exit of the survey.

The simple use case example for a single transect works like this:
* You have a single transect which identifies the flight path for images. You want to cover the path with images from entry to exit.
* The entrance/exit points are waypoints.
* After the entrance waypoint you start trigger by distance with trigger once immediately.
* After the exit waypoint you stop trigger by distance and also set trigger once immediately.

This way you have full coverage with images at both ends of the transect.

This pull also changes the wording on how the trigger distance value works:
* I don't understand the "ignore" concept for trigger distance. If you want to ignore this value, why are you using the command at all? Maybe I am missing something.
* Setting trigger distance to 0 simply stops the triggering. Only 0 is supported since although PX4 may support -1 as well, ArduPilot does not. I don't see why -1 is needed as well?

Note: The explanation above is written assuming TRIGG_DIST does not trigger immediately but only after the distance has passed. This is how ArduPilot works. This is not how PX4 work. PX4 triggers immediately. The reason for that is because I asked for it in order to solve part of the above problem. In hindsight I think that was a bad request. And this proposal is better since it can keep both firmwares working the same way.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1151,9 +1151,9 @@
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST">
         <description>Mission command to set camera trigger distance for this flight. The camera is trigerred each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera.</description>
-        <param index="1">Camera trigger distance (meters). -1 or 0 to ignore</param>
+        <param index="1">Camera trigger distance (meters). 0 to stop triggering.</param>
         <param index="2">Camera shutter integration time (milliseconds). -1 or 0 to ignore</param>
-        <param index="3">Empty</param>
+        <param index="3">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
         <param index="4">Empty</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>


DonLakeFlyer:
@julianoes @meee1 @WickedShell FYI for discussion.

DonLakeFlyer:
The naming of "Trigger once immediately" could likely come up with better naming and/or explanation.

DonLakeFlyer:
@OXINARF Can someone from ArduPilot side look at this?

WickedShell:
Sorry I was quite busy most of last week and spaced responding on this. Implementation wise it looks straightforward.

The only argument I can think of for including -1 is if stop triggering and ignore are meant to be distinct for some reason, and with your proposed change set there is more merit to that argument, as now you could be trying to use the waypoint to force the camera to shoot a photo at a specific spot.

I'd caution that the trigger immediately makes sense on starting the mission, but there are a couple places where I think it falls apart on stopping. The two main scenarios that come to mind are:
A) We are still in the midst of triggering the camera from the previous attempt, and can't start taking another photo yet, if we are doing this do we then just say that the command is complete, or are we expected to store the command? (The former is fine and easy, the latter would be a larger change set from the current behaviour).
B) The camera just shot a photo recently and isn't ready to take another one, this will generate as a missed photo. Given the lack of control you generally have on camera scheduling this will just be a random chance regardless. I suspect this is probably tolerable for a lot of users, but I think it's worth emphasizing that this won't get you perfect coverage at the end of a photo leg. From a planning perspective ArduPlane does flyby waypoints rather then fly over out of the box, which means that the GCS will be really challenged to drop this waypoint at the right spot on the map for you, as the turn time is dynamic based upon current travel speeds.

Here is an alternate idea that I suspect won't go anywhere, but I'm going to throw it out there anyways :)  If we extended the mission commands to have a flags field, and dedicate a flag to camera trigger, that way the GCS can tag each waypoint as "if flying this waypoint at the moment enable the camera system per the last distance/settings loaded". It would help cover the mission start/end concerns without tripling (to quadrupling depending on implementation) the number of waypoints used in a mission. However I suspect that given that photogrammetry isn't the most common application there wouldn't be a lot of support for the flags field. (And it still wouldn't give you what you wanted here, unless you dedicated a flag to "trigger camera upon reaching the waypoint".

DonLakeFlyer:
> We are still in the midst of triggering the camera from the previous attempt, and can't start taking another photo yet,

Hmm, crap. That throws a wrench in the whole thing from the transect exit side. I think the entrance side is still valuable though. The problem is with waypoint acceptance and so forth you always need to  fudge survey boundaries to get full coverage. What I'm trying to solve is a way to be able to not have to fudge survey edges and guarantee coverage while not taking images in turns. Hmm.

DonLakeFlyer:
What is needed is something that won't move on to the next waypoint until after the camera triggers. That way you could guarantee a last full distance traversed and image taken on the transect exit.

WickedShell:
Entrance is straightforward. The scenario I proposed in A is a case where I think we could reasonably make the argument the photo was requested close enough to the edge where it doesn't matter, B is the one I think will cause more problems.

> What is needed is something that won't move on to the next waypoint until after the camera triggers. That way you could guarantee a last full distance traversed and image taken on the transect exit.

How would you do this on a fixed wing platform? I don't see anyway for this to not go wrong, the correct behavior is to keep flying the leg till you get to the next scheduled photo location IMO, but that causes headaches if you hadn't preplanned that on the GCS (and I'm opposed to automatically trying to do that, as the operator could put in a bad interval number that would like like a fly away IE 1km distances or something silly)

DonLakeFlyer:
Argh! It doesn't seem like there is any way to set this up correctly. Other than to make sure you have a decent sized turnaround and you don't turn off the camera in the turns. So you oversample the thing so much you should hit the full polygon. That said I still think the trigger immediate on distance start is good to use.

WickedShell:
Yeah honestly I'm a big advocate of just oversize, and fly it, trimming angled photos during turns is easy in a tagging process (and in Agisoft at least). IMO the only reasons to try and turn the camera off during turns are battery life and storage space, and given the cost of storage space (and most peoples endurance) the battery is the only one I sympathize with :)

Anyways I'm absolutely fine with the flag for trying to take a photo when you hit the waypoint, it's quite reasonable. Will require a bit of mission twiddling to allow it to directly trigger the camera, but on the first step of the MAVLink side I think thats very reasonable, most of my concerns related to the situation you are trying to address. (And I strongly agree it's a challenging one).

Do you see merit to maintaining the -1 flag as ignore so that you can stick one of these with trigger immediately after a waypoint in the middle of the mission without having to have provided a distance? I can see just asking the GCS to duplicate the desired distance number into that, but I can also see that it might be useful if they could just ignore that part?

PR #747, opened 2017-07-07
Camera Definition

dogmaphobic:
This is in preparation for adding support for a _Camera Definition File_ as discussed in [discuss.px4.io](http://discuss.px4.io/t/mavlink-camera-api-discussion/3278/26).

 * Flagging messages and commands that are to be removed
 * Adding an URI for fetching the camera definition file (if any) to the `CAMERA_INFORMATION` message
 * Adding a revision number for the definition file (to check against an already cached version)
 * Adding a capability flag to the `CAMERA_INFORMATION`.
 * Adding an enum with a bitmap describing the various capability flags (`CAMERA_CAP_FLAGS`)
 * Adding a `device_id` field to the `COMMAND_ACK` message to identify target devices that share a component ID (Camera, Storage, etc.)

@julianoes  Please check and see if the implementation of the `CAMERA_CAP_FLAGS` is correct. This is the first time I define a bitmap.

Note that this doesn’t remove any existing functionality yet. Only when the camera definition file is fully implemented and validated will the _Experimental_ messages and commands be removed.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1376,13 +1376,13 @@
       </entry>
       <entry value="521" name="MAV_CMD_REQUEST_CAMERA_INFORMATION">
         <description>WIP: Request camera information (CAMERA_INFORMATION).</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">0: No action 1: Request camera capabilities</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="522" name="MAV_CMD_REQUEST_CAMERA_SETTINGS">
         <description>WIP: Request camera settings (CAMERA_SETTINGS).</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">0: No Action 1: Request camera settings</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
@@ -1420,7 +1420,7 @@
       </entry>
       <entry value="527" name="MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS">
         <description>WIP: Request camera capture status (CAMERA_CAPTURE_STATUS)</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">0: No Action 1: Request camera capture status</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
@@ -1431,14 +1431,14 @@
       </entry>
       <entry value="529" name="MAV_CMD_RESET_CAMERA_SETTINGS">
         <description>WIP: Reset all camera settings to Factory Default (CAMERA_SETTINGS)</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">0: No Action 1: Reset all settings</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
         <description>WIP: Set camera running mode. Use NAN for reserved values and values you don't want to change.</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
-        <param index="2">Camera mode (0: photo mode, 1: video mode)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
+        <param index="2">Camera mode (0: photo/image mode, 1: video mode)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="531" name="MAV_CMD_SET_CAMERA_SETTINGS_3">
@@ -1450,19 +1450,19 @@
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
         <description>WIP: Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">Duration between two consecutive pictures (in seconds)</param>
         <param index="3">Number of images to capture total - 0 for unlimited capture</param>
         <param index="4">Reserved</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE">
         <description>WIP: Stop image capture sequence</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">Reserved</param>
       </entry>
       <entry value="2002" name="MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE">
         <description>WIP: Re-request a CAMERA_IMAGE_CAPTURE packet. Use NAN for reserved values.</description>
-        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)</param>
+        <param index="1">Camera ID (0 for all cameras, 1 for first, 2 for second, etc.) (EXPERIMENTAL: IT WILL BE REMOVED)</param>
         <param index="2">Sequence number for missing CAMERA_IMAGE_CAPTURE packet</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4078,33 +4096,36 @@
       <field type="uint8_t" name="camera_count">Number of cameras</field>
       <field type="uint8_t[32]" name="vendor_name">Name of the camera vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the camera model</field>
-      <field type="uint32_t" name="firmware_version">Version of the camera firmware</field>
+      <field type="uint32_t" name="firmware_version">Version of the camera firmware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
```
julianoes: I wonder if these & will get htmlized anyway.

message_definitions/v1.0/common.xml
```diff
@@ -4078,33 +4096,36 @@
...
       <field type="float" name="focal_length" units="mm">Focal length in mm</field>
       <field type="float" name="sensor_size_h" units="mm">Image sensor size horizontal in mm</field>
       <field type="float" name="sensor_size_v" units="mm">Image sensor size vertical in mm</field>
       <field type="uint16_t" name="resolution_h" units="pix">Image resolution in pixels horizontal</field>
       <field type="uint16_t" name="resolution_v" units="pix">Image resolution in pixels vertical</field>
       <field type="uint8_t" name="lens_id">Reserved for a lens ID</field>
+      <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS">CAMERA_CAP_FLAGS enum flags (bitmap) describing camera capabilities.</field>
+      <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration)</field>
```
julianoes: That's the schema/protocol version?

message_definitions/v1.0/common.xml
```diff
@@ -4078,33 +4096,36 @@
...
+      <field type="uint8_t[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available).</field>
```
julianoes: You don't want a local and a remote URI?

message_definitions/v1.0/common.xml
```diff
@@ -3158,6 +3175,7 @@
       <field type="uint8_t" name="result" enum="MAV_RESULT">See MAV_RESULT enum</field>
       <extensions/>
       <field type="uint8_t" name="progress">WIP: Needs to be set when MAV_RESULT is MAV_RESULT_IN_PROGRESS, values from 0 to 100 for progress percentage, 255 for unknown progress.</field>
+      <field type="uint8_t" name="device_id">WIP: Used to identify targeted device when multiple devices use same component ID (i.e. Camera, Storage, etc.)</field>
```
dogmaphobic: Take for example this message:
```
MAV_CMD_REQUEST_CAMERA_INFORMATION
Param #1 | Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)
```
I need to know what `Camera ID` this response is for. On a parallel system, sending messages from arbitrary camera controllers, the system needs to know to which (camera) controller to route the message to. In this case, both System ID and Component ID are the same. Only the Camera ID changes.

message_definitions/v1.0/common.xml
dogmaphobic: There is one way to find out :)

message_definitions/v1.0/common.xml
dogmaphobic: No, the content version. It will tell if some parameter or option (or rule, localization, etc.) changed. The GCS would have cached the file and there is no reason to fetch it again if it's the same version.

message_definitions/v1.0/common.xml
dogmaphobic: I don't see a reason to have two separate URIs. If the vehicle/camera is able to _serve_ the file, why have an external URI? If it can't, it can't and there would be only one. Not to mention the lack of space within the message. Can you think of a reason why we would need both local and external?

message_definitions/v1.0/common.xml
```diff
@@ -4078,33 +4096,36 @@
       <field type="uint8_t" name="camera_count">Number of cameras</field>
```
lbegani: Not related, what does camera_count mean? Number of sensors in the given camera?

message_definitions/v1.0/common.xml
lbegani: Having multiple URI might lead to confusion/mismatch etc at some point in time. Better to have a single URI.

message_definitions/v1.0/common.xml
lbegani: Will this change require existing users of COMMAND_ACK to make corresponding changes in their code?

message_definitions/v1.0/common.xml
```diff
@@ -2586,6 +2585,24 @@
         <description>Landing target represented by a pre-defined visual shape/feature (ex: X-marker, H-marker, square)</description>
       </entry>
     </enum>
+    <enum name="CAMERA_CAP_FLAGS">
+      <description>Camera capability flags (Bitmap).</description>
+      <entry value="1" name="CAMERA_CAP_FLAGS_CAPTURE_VIDEO">
+        <description>Camera is able to record video.</description>
+      </entry>
+      <entry value="2" name="CAMERA_CAP_FLAGS_CAPTURE_PHOTO">
```
lbegani: I think the term IMAGE sounds better than PHOTO (Subjective)

message_definitions/v1.0/common.xml
```diff
@@ -2586,6 +2585,24 @@
         <description>Landing target represented by a pre-defined visual shape/feature (ex: X-marker, H-marker, square)</description>
       </entry>
     </enum>
+    <enum name="CAMERA_CAP_FLAGS">
```
lbegani: Shouldn't all capability definition be subsumed in camera definition file? Or is this only to be used for specific cases where def file isn't available?

message_definitions/v1.0/common.xml
dogmaphobic: This would be the same changes that occurred when the `progress` field were added. If you don't use the new field, existing code would simply ignore it.

message_definitions/v1.0/common.xml
dogmaphobic: The idea is for when you don't have the camera definition file. While implementing this, I realized that most of this info within the definition file is redundant and it should not be there (in the definition file). As this is the first call you make for camera discovery, you will receive this message back with the information within. There is no reason to duplicate it within the camera definition as well.

message_definitions/v1.0/common.xml
dogmaphobic: I have no preference. If this what everybody thinks, I will change it.

message_definitions/v1.0/common.xml
dogmaphobic: When you request camera information and set `Camera ID` to 0 (All Cameras), you will receive 0 or more `CAMERA_INFORMATION` messages in response. If the vehicle has two cameras, you would receive two `CAMERA_INFORMATION` messages where one would have `camera_id` set to 1 and the other `camera_id` would be set to 2. In this case, `camera_count` would be set to 2. That's how you know how many cameras there are and to make sure you received a `CAMERA_INFORMATION` for each. If the `camera_count` is set to 2 and you only get one `CAMERA_INFORMATION` back, it means one of the messages was lost. That would allow you to request again the missing message (and target the request to the missing camera id.)

message_definitions/v1.0/common.xml
julianoes: Ok sure.

message_definitions/v1.0/common.xml
julianoes: On the other hand, PHOTO works with PHOTO mode.

message_definitions/v1.0/common.xml
julianoes: Good luck :).

message_definitions/v1.0/common.xml
julianoes: Ah, I thought the version would be included in the file string. I guess this allows to have one http URI, and then a version number on top.

message_definitions/v1.0/common.xml
julianoes: > On a parallel system

For the sake of simplicity, I would not try to parallelize the case of two camera as one mavlink node (with system and component ID). I think this will be confusing for most developers implementing mavlink commands.

message_definitions/v1.0/common.xml
dogmaphobic: That was the reason I used "Photo". But the command to "take pictures" is called "START IMAGE CAPTURE" so we sort of have two different terms for the same thing already.

message_definitions/v1.0/common.xml
dogmaphobic: The problem I'm trying to solve is to determine which camera is responding to a command when both system and component IDs are the same.

message_definitions/v1.0/common.xml
lbegani: Lets make them all same

dogmaphobic:
* Flagged all references to *Camera ID* from the various camera messages as *EXPERIMENTAL: IT WILL BE REMOVED*. Each camera should present itself using a different component ID. As far as we could tell, we could not find any concrete use case where one single camera would provide separate interfaces.

 * For consistency’s sake, I’ve changed all references to “Photo” and replaced with “Image” where appropriate. For example, `CAMERA_CAP_FLAGS_CAPTURE_PHOTO` is now `CAMERA_CAP_FLAGS_CAPTURE_IMAGE`.

 * Removed the `device_id` field from the `COMMAND_ACK` message. That was based on the assumption we had multiple cameras using one same Component ID. As that is not the case, we don’t need this `device_id` field.

Speaking of `device_id`, I’ve also noticed there is a parameter in `MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN` for an ID where the description uses *Camera ID* as an example. I’m assuming that should be renamed to something else? If you want to reboot a camera, wouldn’t you send a `MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN` command to the camera’s specific *Component ID*? Unless the autopilot is *broadcasting* this to all components? Why is it there?

message_definitions/v1.0/common.xml
dogmaphobic: I removed it based on the fact we decided to remove *Camera ID* from all camera messages. Separate cameras are to be identified using different *Component ID* instead.

message_definitions/v1.0/common.xml
dogmaphobic: We decided to use XML instead. Json simply isn't conducive for this. As such, the schema is defined in the file header.

message_definitions/v1.0/common.xml
dogmaphobic: Done. It's now IMAGE wherever appropriate.

message_definitions/v1.0/common.xml
dogmaphobic: This has been removed as we discussed.

dogmaphobic:
Putting this here for now but it has to be documented somewhere more appropriate at some point.

How to deal with storage (the various STORAGE messages).

When dealing with a camera’s storage unit, we will use the camera’s component ID when sending these messages. That is, if you want to _format_ the storage of a given camera (or query storage status, etc.) you will send that command to the camera’s component ID. If the camera has its own storage, it can just go ahead and format it (faster than erasing files). If the camera _shares_ storage with something else, it’s up to the camera to deal with it. In that case,  when it gets a _format_ command, it would _erase_ its files, leaving the storage alone. If it finds the storage is corrupted (and therefore requiring formatting) it would elect to format it instead if appropriate.

message_definitions/v1.0/common.xml
julianoes: Ok.

PR #749, opened 2017-07-09
Common: Ability to specify ground altitude for landing waypoints

AndreasAntener:
This is an attempt to consolidate approaches on vertical landings. I don't know all the methods used but I believe that this currently solved differently across implementations. Possibly someone already has an elegant solution for this problem, if so, please let me know :).

**The problem**: With only one altitude field for a landing mission item we have to decide if the vehicle approaches on that altitude (horizontally) until it's above the landing spot **OR** if that altitude marks the height where we should expect ground (can be used in two step vertical landing or as fallback for a failed distance sensor).

For fixed-wing vehicles doing a glide slope landing these two altitudes are the same in the final phase which is why I'm focusing only on vertical landings right now.

My proposal is to use parameter 3 to specify the ground altitude. That way the actual altitude field is preserved as the target flying altitude and won't change it's meaning depending on the `MAV_CMD`.

I believe that would be a sensible solution, however it is worth mentioning that there are already other options:
1. Use 2 waypoints: One for the approach and one for the actual landing where the altitude field serves as ground altitude.
2. Use actual terrain data on the flight controller.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -686,7 +686,7 @@
         <description>Land at location</description>
         <param index="1">Abort Alt</param>
         <param index="2">Empty</param>
-        <param index="3">Empty</param>
+        <param index="3">Ground altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
         <param index="4">Desired yaw angle. NaN for unchanged.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
@@ -816,7 +816,7 @@
         <description>Land using VTOL mode</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
-        <param index="3">Empty</param>
+        <param index="3">Ground altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
         <param index="4">Yaw angle in degrees. NaN for unchanged.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -686,11 +686,11 @@
         <description>Land at location</description>
         <param index="1">Abort Alt</param>
         <param index="2">Empty</param>
-        <param index="3">Ground altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
+        <param index="3">Empty</param>
         <param index="4">Desired yaw angle. NaN for unchanged.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="7">Altitude (ground level)</param>
       </entry>
       <entry value="22" name="MAV_CMD_NAV_TAKEOFF">
         <description>Takeoff from ground / hand</description>
@@ -816,11 +816,11 @@
         <description>Land using VTOL mode</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
-        <param index="3">Ground altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
+        <param index="3">Approach altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
         <param index="4">Yaw angle in degrees. NaN for unchanged.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="7">Altitude (ground level)</param>
       </entry>
       <!-- IDs 90 and 91 are reserved until the end of 2014,
                     as they were used in some conflicting proposals


DonLakeFlyer:
Is there a problem with the "Fixed wing landing pattern" item in QGC?

dagar:
I've personally always interpreted this as where you want to land, not where you go to begin landing. Then you could use different MAV_FRAMEs to specify what that means. 

This year for FW landing QGC added a complex mission item that combined a spiral descent + tangent breakout and landing waypoint. It's one thing you configure in the gui, but actually sends multiple waypoints to the vehicle. Based on how well that's worked I'm a fan of using 2 waypoints.

What do you think about the 2 waypoint solution combined with a better description of the altitude field in mavlink? This also leaves room for the controller to decide how it handles a land waypoint, similar to the FW glide slope.

WickedShell:
I have a similar problem with deepstall landings that I think this addresses nicely. Currently the altitude field is used for specifying the altitude to do the entry from (deepstall is a complex item on the autopilot side rather then multiple mission items, as the mission doesn't contain enough capability to generate the needed waypoints), but I'd like to be able to enter ground altitude offsets for the target point.

This proposal would work quite well for what I need. However you need to accept 0 as unspecified rather then NaN. There will be to many old GCS's that won't realize that this is a new meaning and would result in a lot of crashes :)

AndreasAntener:
> This year for FW landing QGC added a complex mission item that combined a spiral descent + tangent breakout and landing waypoint. It's only one item you select in the gui, but actually sends multiple waypoints to the vehicle. Based on how well that's worked I'm a fan of using 2 waypoints if you want to be explicit with the landing.

Good to know, I wasn't aware of that as it has been a while since I did the last fixed-wing landing ;).

> What do you think about the 2 waypoint solution combined with a better description of the altitude field in mavlink? This also leaves room for the controller to optionally decide how it handles a land waypoint, similar to the FW glide slope.

Basically I'm all for that. In this case the important outcome of this PR would be to clarify the altitude field on landing items.

AndreasAntener:
@WickedShell 
I would rather expect that your ground altitude should be specified on the command mission item used to trigger the deep stall.

> This proposal would work quite well for what I need. However you need to accept 0 as unspecified rather then NaN. There will be to many old GCS's that won't realize that this is a new meaning and would result in a lot of crashes :)

Excluding 0 would haunt me when I'm asleep ;). It is guaranteed that someone runs into this eventually. The other option is to add a parameter that enables the use of this field so the GCS can set it when updated.

DonLakeFlyer:
> There will be to many old GCS's that won't realize that this is a new meaning and would result in a lot of crashes :)

Really? This is usually more of a problem on the vehicle side. For that I tend to just look at PX4 and ArduPilot source and if they don't freak out over NaN in previously unused field I'd call it good. 

Why would a ground station have a problem with a previously unused field having a NaN in it? Lot of other fields already use this NaN concept.

WickedShell:
Long term I need to be able to specify landings without all the hassle/overhead of mission manipulation, which means condensing behaviour into a single command. (From the GCS side I need to be able to present the option of simply issuing a land here (where here  is some lat/long/alt picked off the map). And this needs to be done without corrupting the uploaded mission, in a poor telem environment.

Trust me, you will be far more haunted by the number of crashes when using MAV_FRAME_GLOBAL that didn't update that field. Because unused fields are all spec'd to be 0 you have to honor that on any new/future extension, or all backwards compatibility is lost. A parameter isn't sufficent as all GCS's on the network might not be up to date, GCS's shouldn't just tweak parameters like that, and it ties mission protocol to the parameter protocol, and there is now a name race condition that the param must be present/named the same.

WickedShell:
@DonLakeFlyer my point is that an old GCS would send the parameter with a value of 0, the aircraft would then interpret that as a landing altitude, and if in MAV_FRAME_GLOBAL means land at sea level if the autopilot is aware of the new meanings.

DonLakeFlyer:
Oh yeah, duh!

dagar:
Yet another option would be adding an entire lat/lon/alt for the optional starting approach point and making it clear the existing lat/lon/alt is the desired end point.

I'd rather we try to encompass all needed information into a single mission item, or go to the other extreme of using two separate items rather than something that's helpful for MC/VTOL, but redundant/confusing for FW.

DonLakeFlyer:
> Yet another option would be adding an entire lat/lon/alt for the optional starting approach point and making it clear the existing lat/lon/alt is the desired end point.

A mission item with two specified lat/lon positions is going to give ground stations fits from an implementation standpoint.

WickedShell:
I'm great with extra lat/long alt as that would support FW glide slope landings in arbitrary locations much better, but short of handling it as an extension I don't see how we could do it (unless we accept the coordinate accuracy hit on sticking the start point lat/long's in floats). But I do like that it allows more expression in a single handlable waypoint and I think actually simplifies some of the mission planning for GCS's (especially with jumping/starting landing waypoints)

sanderux:
I see several problems with the 2 waypoint approach;
1. If the FC doesnt realize it's flying towards a (VTOL)LAND it cannot apply logic to the acceptance radius/back transition duration
2. It makes the interface needlessly complex, the commands were invented to simplify the operation, not make it more complex, should we take this path then why not scrap the commands and do WP - back trans - land
3. In all cases the operator needs to be made aware of the difference between the position setpoint altitude and the expected landing altitude. Making the operator switch between 2 different commands to set this would make it more prone to errors.

I would back @AndreasAntener in his original proposal, and perhaps add that future terrain databases in QGC could pre-populate this field.

AndreasAntener:
We are dealing with two different problems:
1. single commands specifying all the information at once, not used in missions, should be atomic
2. missions, which are generated by a GCS and should be allowed to be complex (which means the GCS hides that complexity from the user)

I'm really focusing on the later right now. I think that for 1, complex commands should have their waypoint independent command definition.

@sanderux 1) is not a problem since we can do a look ahead, 2) I'm not only talking about a hybrid VTOL here but also about a normal multirotor and therefore the normal land command, 3) agreed, but the GCS can do that

dagar:
@sanderux 
1. This could almost be a separate discussion. Adding an optional transition at altitude to just VTOL_LAND makes sense.
2. I disagree, see FW spiral landing for example. The user never has to know or care if the end result is 2 simple waypoints or one slightly more complex one. On the firmware side keeping them separate is much cleaner. Is it a LAND or POSITION setpoint vs we're within the portion of a LANDING waypoint that's treated like a position setpoint.
3. Yes, we absolutely need to remove the ambiguity, and part of that is being consistent between usage, but the operator doesn't have to see any of this.

sanderux:
@dagar 
1. I think we need to decide if a VTOL_LAND and a LAND command is indeed a position waypoint. If we agree that it is (as it is now in PX4 and AFAIK in APM) then it should comply with the standards and therefor comply with the altitude field following a FOH or ZOH path towards the 3D waypoint.
Expected landing altitude would seem to me like the additional information. 

@dagar @AndreasAntener 
2 and 3. Perhaps in this reasoning i was a little too much working on assumptions on how this would be implemented in GCS's. Indeed if the interface can be made clean it could very well result in 2 commands, but i do fail to see the need for this is the additional information can be passed in one.

AndreasAntener:
If we would be dogmatic about a multi mission item approach, the VTOL_LAND and VTOL_TAKEOFF waypoints are already broken. Maybe it does make sense to only care for VTOL_LAND item in this case and let the GCS do more complex things with the traditional items.

dagar:
@sanderux 
1. This is the part that continues to surprise me. I didn't think it was even a question that they aren't positions. However, approaching them with FOH doesn't necessarily have to be the case. The mavlink spec doesn't state how you have to get to that position.

2. 3. The issue I have is that these aren't really different problems between FW and MC. You want to specify the start and the end. With FW we already do this with another waypoint. Adding a shortcut on the mavlink side for MC landing just pushes the complexity to the firmware and GCS. Let's find the solution that works consistently for all platforms and avoid additional problems down the road.

@AndreasAntener I don't see the multi mission item approach as necessarily a problem. Trying to get extra complexity out of an existing simple item with an optional parameter is the part I question.

As for VTOL what I really want is VTOL_TRANSITION to have a position, but we don't need to get into that now.

AndreasAntener:
> The issue I have is that these aren't really different problems between FW and MC. You want to specify the start and the end. With FW we already do this with another waypoint. Adding a shortcut on the mavlink side for MC landing just pushes the complexity to the firmware and GCS. Let's find the solution that works consistently for all platforms and avoid additional problems down the road.

I agree. In the multi waypoint scenario the nice thing is that the target location of a land item will be "pure" again. Even if in my head I think "ground altitude" it's actually just the location to fly to, with the addition that we'll descend even further if needed (similar to additional logic needed for FW like flaring). (and diagonal landings with FOH will be possible again ;))

However I think it's important to fix VTOL_LAND as it lacks that option in the current form. How does APM expose and implement VTOL_LAND?

DonLakeFlyer:
Just FYI to everyone: The Fixed wing landing pattern from QGC is not a two item system. It is a three item system with two lat/lon positions which define a glide slope:
* First is a MAV_CMD_DO_LAND_START item to signal this is a landing sequence
* Followed by a MAV_CMD_NAV_LOITER_TO_ALT with a tangent exit which sets the latitude for the start of the glide slide position and altititude.
* Follow by the MAV_CMD_NAV_LAND which sets the position and altitude for landing point on the ground.

sanderux:
After careful deliberation a lot of good arguments from @dagar i changed my mind. altitude defines the destination. waypoint approach altitude can be considered a behavioral aspect of the item.

AndreasAntener:
Thanks @DonLakeFlyer 

Thanks for all the input. IMHO, considering the multi waypoint strategy, it would make sense to add this special behaviour to VTOL_LAND but leave LAND untouched.

LorenzMeier:
@AndreasAntener Could you update the PR to match your last comment? Thanks!

message_definitions/v1.0/common.xml
```diff
@@ -816,11 +816,11 @@
         <description>Land using VTOL mode</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
-        <param index="3">Empty</param>
+        <param index="3">Approach altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
         <param index="4">Yaw angle in degrees. NaN for unchanged.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
```
LorenzMeier: There is an edge case here: Existing implementations will have this value at zero. The flight controller should check for this altitude to be above the default altitude.

AndreasAntener:
This is updated according to the discussions:
- waypoint position in general is considered the destination of that waypoint, in which case the altitude of the landing waypoint is ground level (0 relative)
- specifying the approach altitude can be done by using a normal waypoint before the landing waypoint and that complexity is potentially hidden by the GCS
- the VTOL transition and land command gets a parameter for the approach altitude. This one is tricky because it will be 0 by default for most GCS, so that needs to be taken care of somehow

AndreasAntener:
@LorenzMeier you were too fast in between my pushes ;) so your edge case is still there but on the vtol approach altitude.

sanderux:
With QGC new terrain elevation api link it could default this value to that reading
@DonLakeFlyer correct?

dagar:
@sanderux yes that's definitely possible.

PR #750, opened 2017-07-09
Common: Ability to specify the direction of the VTOL forwards transition

AndreasAntener:
Proposal to standardize in which direction the VTOL transitions. This is relevant for VTOL takeoff as well as for the independent transition command.

Basically there are 4 options we use so far (there might me more, please comment):
1) (default) Direction of the next waypoint. This should be expected when planning a mission on the map since we are supposed to track the line from A to B.
2) (only relevant for takeoff) Use the heading of the vehicle on the ground before takeoff, makes sense if the user positions the vehicle consciously into the wind.
3) Use the specified heading.
4) Use whatever heading the vehicle has when getting the transition command (reaching the waypoint), especially useful for takeoff with enabled weather-vaning.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -805,7 +805,7 @@
       <entry value="84" name="MAV_CMD_NAV_VTOL_TAKEOFF">
         <description>Takeoff from ground using VTOL mode</description>
         <param index="1">Empty</param>
-        <param index="2">Empty</param>
+        <param index="2">Forwards transition heading: 0: use the heading towards the next waypoint, 1: use the heading on takeoff, 2: use the specified heading in parameter 4, 3: use the current heading when reaching takeoff altitude (facing the wind when weather-vaning is active).</param>
         <param index="3">Empty</param>
         <param index="4">Yaw angle in degrees. NaN for unchanged.</param>
         <param index="5">Latitude</param>
@@ -1540,6 +1540,8 @@
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION">
         <description>Request VTOL transition</description>
         <param index="1">The target VTOL state, as defined by ENUM MAV_VTOL_STATE. Only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
+        <param index="2">Forwards transition heading: 0: use the heading towards the next waypoint, 1: ignored, 2: use the specified heading in parameter 4, 3: use the current heading when executing this command (facing the wind when weather-vaning is active).</param>
+        <param index="4">Yaw angle in degrees.</param>
       </entry>
       <entry value="4000" name="MAV_CMD_SET_GUIDED_SUBMODE_STANDARD">
         <description>This command sets the submode to standard guided when vehicle is in guided mode. The vehicle holds position and altitude and the user can input the desired velocites along all three axes.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -805,7 +805,7 @@
       <entry value="84" name="MAV_CMD_NAV_VTOL_TAKEOFF">
         <description>Takeoff from ground using VTOL mode</description>
         <param index="1">Empty</param>
-        <param index="2">Forwards transition heading, see VTOL_TRANSITION_HEADING enum.</param>
+        <param index="2">Front transition heading, see VTOL_TRANSITION_HEADING enum.</param>
         <param index="3">Empty</param>
         <param index="4">Yaw angle in degrees. NaN for unchanged.</param>
         <param index="5">Latitude</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2605,6 +2605,7 @@
       </entry>
       <entry value="4" name="VTOL_TRANSITION_HEADING_ANY">
         <description>Use the current heading when reaching takeoff altitude (potentially facing the wind when weather-vaning is active).</description>
+      </entry>
     </enum>
     <enum name="CAMERA_CAP_FLAGS">
       <description>Camera capability flags (Bitmap).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2588,17 +2588,20 @@
     </enum>
     <enum name="VTOL_TRANSITION_HEADING">
       <description>Direction of VTOL transition</description>
-      <entry value="0" name="VTOL_TRANSITION_HEADING_NEXT_WAYPOINT">
+      <entry value="0" name="VTOL_TRANSITION_HEADING_VEHICLE_DEFAULT">
+        <description>Respect the heading configuration of the vehicle.</description>
+      </entry>
+      <entry value="1" name="VTOL_TRANSITION_HEADING_NEXT_WAYPOINT">
         <description>Use the heading pointing towards the next waypoint.</description>
       </entry>
-      <entry value="1" name="VTOL_TRANSITION_HEADING_TAKEOFF">
+      <entry value="2" name="VTOL_TRANSITION_HEADING_TAKEOFF">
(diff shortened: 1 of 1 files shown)


dagar:
Great! Is it worth adding enums?

AndreasAntener:
Right :). Will do with the final list.

sanderux:
Why set this in the waypoint? why would this not be a parameter in the FC?
There are VTOL's that have a very hard time yawing and making this such a user-oriented field can result in problems IMHO.

AndreasAntener:
That's is a good point. I considered it to be dynamic enough to use on waypoints, but that might be a misconception from a testers point of view. I wouldn't expose that option to the user "just like that" but rather tied to the current mode of operation which could change between missions.

AndreasAntener:
Having said that, it is closely related to the mission plan, which is why I think it would be good to have in the protocol.

sanderux:
I think there are many parameters that affect the behavior of a mission. Although your solution will probably be more flexible i still believe that specing this as such will invite GCS programmers to expose this parameter where this is not always desired.

dagar:
Given we treat MAV_CMD_DO_VTOL_TRANSITION like a command that initiates transition immediately I'm wondering if also having a yaw requirement is a little weird.

If you're already moving forward as a multicopter do you stop in place and try to achieve the yaw setpoint? If that's the intention why not give it an optional position and skip the intermediary. On the firmware side this would really help tame the navigator vtol complexity.

AndreasAntener:
@dagar makes sense. How can we force a heading at a location with the current waypoint specs?

dagar:
Do you mean the combination of point towards waypoint on route, but yaw in place once you arrive?

AndreasAntener:
Basically stop and wait on the yaw setpoint being met at a specific location and abort the mission if the setpoint cannot be met after a timeout period.

LorenzMeier:
@AndreasAntener A yaw requirement and timeout would be feasible. Where is this standing? The extension seems sensible and I'd like to get this in. I don't think that this is a system parameter, how the vehicle should take off is specific to a mission and should be encoded in a mission - so this all makes perfect sense.

AndreasAntener:
I think too that it is definitely a mission governed parameter. However, after considering @dagar input and the discussion we had on the landing altitude it makes sense to me that for the "atomic" waypoint commands we don't want to add intermediate steps on the flight controller. Therefore I would add this extension only for VTOL_TAKEOFF as it is already a combination of actions anyway. I'll updated this and also go over https://github.com/mavlink/mavlink/pull/749 once more

LorenzMeier:
@AndreasAntener Could you have the pass so I can assist getting it in?

AndreasAntener:
Sorry for the delay. @sanderux @dagar is my proposed default "transition towards the next waypoint" what it is now?

sanderux:
On a regular transition command it is.

sanderux:
I do think the standard should not be to point towartds the takeoff waypoint as some vehicles might not be capable of achieving this. Setting this as default will make it hard to explain to users how to set this correctly.

I am still an advocate of controlling this on the param side

sanderux:
I would suggest adding VTOL_TRANSITION_HEADING_VEHICLE_DEFAULT as standard option to have this behavior governed by the FC who could use a param to store the default behavior?

sanderux:
Also VTOL_TRANSITION_HEADING_WEATHERVANE seems it does not describe correctly the behavior if i understand it correctly. It seems this value simply disables navigating to the intended yaw setpoint. In that case having weathervane enabled or not does not matter. Our vehicle defaults to holding heading as it was on the ground and transition in that direction.

I would propose calling this VTOL_TRANSITION_HEADING_HOLD

AndreasAntener:
@sanderux re your last comment: if you disable weather-vane then  VTOL_TRANSITION_HEADING_TAKEOFF and VTOL_TRANSITION_HEADING_WEATHERVANE would have the same result. But if VTOL_TRANSITION_HEADING_WEATHERVANE is used together with weather-vane the heading could change during takeoff which means it will transition into a random direction. Still unclear?

sanderux:
Why would you make this distinction? if weathervane is enabled would you not always want VTOL_TRANSITION_HEADING_WEATHERVANE and if disabled always want VTOL_TRANSITION_HEADING_TAKEOFF ?

AndreasAntener:
We normally have weather-vane enabled but transition into the direction of the next waypoint by default. That's similar to using the heading on takeoff but with weather-vane. Weather-vane is just used to cause less stress on the climb in that case. If the VTOL can't manage the heading for the transition the mission is aborted. So for me all combinations exist which is why I have those split.


> I do think the standard should not be to point towartds the takeoff waypoint as some vehicles might not be capable of achieving this. Setting this as default will make it hard to explain to users how to set this correctly.
>
> I am still an advocate of controlling this on the param side

I don't mind changing the default behaviour. My main priority is to make it clear to the user what's happening. These options require a visualization or clear mission item settings on the GCS. Additional firmware parameter would only make it less transparent. Once the GCS handles this properly it should be easy to set/explain.

sanderux:
You might be right on the weathervane takeoff behavior. But your current implementation forces the default behavior to be the same on every vehicle whereas my suggestion at least gives some control to the airframe maintainer. it could be conceivable that some vtol will have virtually no yaw control, this implementation forces the user to then always have to change this param. It also sounds like an easy param to forget setting.

message_definitions/v1.0/common.xml
```diff
@@ -2586,6 +2586,21 @@
...
+      <entry value="0" name="VTOL_TRANSITION_HEADING_NEXT_WAYPOINT">
+        <description>Use the heading pointing towards the next waypoint.</description>
+      </entry>
+      <entry value="1" name="VTOL_TRANSITION_HEADING_TAKEOFF">
+        <description>Use the heading on takeoff (while sitting on the ground).</description>
+      </entry>
+      <entry value="2" name="VTOL_TRANSITION_HEADING_SPECIFIED">
+        <description>Use the specified heading in parameter 4.</description>
+      </entry>
+      <entry value="3" name="VTOL_TRANSITION_HEADING_WEATHERVANE">
```
LorenzMeier: Let's change this to HEADING_ANY because that's what it effectively means. If the vehicle weathervanes or not does not matter.

message_definitions/v1.0/common.xml
```diff
@@ -2586,6 +2586,21 @@
         <description>Landing target represented by a pre-defined visual shape/feature (ex: X-marker, H-marker, square)</description>
       </entry>
     </enum>
+    <enum name="VTOL_TRANSITION_HEADING">
+      <description>Direction of VTOL transition</description>
+      <entry value="0" name="VTOL_TRANSITION_HEADING_NEXT_WAYPOINT">
```
LorenzMeier: If the value prescribes a heading we could use param 3 to specify the acceptance region in +- degrees. So if you set HEADING_SPECIFIED and and param3 is 15 degrees, then it will transition if it is able to hold within that region.

Or is that over-engineering it?

AndreasAntener: There is a parameter for that (haha, falling into my own trap here...).

AndreasAntener:
Updated
@sanderux acknowledged your concern
@LorenzMeier we could add the yaw error still but I have a feeling it would be too much

message_definitions/v1.0/common.xml
```diff
@@ -805,7 +805,7 @@
       <entry value="84" name="MAV_CMD_NAV_VTOL_TAKEOFF">
         <description>Takeoff from ground using VTOL mode</description>
         <param index="1">Empty</param>
-        <param index="2">Empty</param>
+        <param index="2">Forwards transition heading, see VTOL_TRANSITION_HEADING enum.</param>
```
sanderux: Is "Forwards" correct? i think it should be either Forward or Front to stay consistent

LorenzMeier:
Needs to be squashed.

LorenzMeier:
This is in.

PR #752, opened 2017-07-13
Extended parameter handling

dogmaphobic:
Extending (by creating a new set of messages) the PARAM SET/VALUE messages. This adds an extended set of result ACKs and increases the value payload from float to double. The primary use is for camera settings but nothing stops you from using for something else.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4263,7 +4263,7 @@
     <message id="322" name="PARAM_EXT_VALUE">
       <description>Emit the value of a parameter. The inclusion of param_count and param_index in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout.</description>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
-      <field type="double" name="param_value">Parameter value</field>
+      <field type="char[128]" name="param_value">Parameter value</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
       <field type="uint16_t" name="param_count">Total number of parameters</field>
       <field type="uint16_t" name="param_index">Index of this parameter</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4279,7 +4279,7 @@
     <message id="324" name="PARAM_EXT_ACK">
       <description>Response from a PARAM_EXT_SET message.</description>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
-      <field type="double" name="param_value">Parameter value (new value if PARAM_ACK_ACCEPTED, current value otherwise)</field>
+      <field type="char[128]" name="param_value">Parameter value (new value if PARAM_ACK_ACCEPTED, current value otherwise)</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code: see the PARAM_ACK enum for possible codes.</field>
     </message>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4267,7 +4267,7 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
       <field type="uint16_t" name="param_count">Total number of parameters</field>
       <field type="uint16_t" name="param_index">Index of this parameter</field>
-    </message>  
+    </message>
     <message id="323" name="PARAM_EXT_SET">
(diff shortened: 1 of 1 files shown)


WickedShell:
Why not just do this as an extension to the exsisting param message? Just add a new field to the bottom, and if the type is marked as  DOUBLE then that should be used?

If making a new message could the MAVLink part of the param id at least be extended to be more then 16 characters? This is a poor legacy to keep carrying forward. Nothing states an autopilot has to support id's longer then 16 chars yet, but one day it will be desired and it would be nice to pave that now.

dogmaphobic:
Wouldn't that break the current PARAM_SET/PARAM_VALUE implementation? Wouldn't this force everyone to update to the "updated" MAVLink? I was under that assumption. If not, yes, by all means it would make a heck lot more sense to use the existing set. But how do you know when to send the PARAM_EXT_ACK versus a PARAM_VALUE back? PARAM_SET is used by everyone and they all expect a PARM_VALUE in return.

WickedShell:
You would have to just send both messages for the forseeable future until others support the new ack/mavlink2.

dogmaphobic:
@LorenzMeier @julianoes Is this reasonable?

WickedShell:
@dogmaphobic the ack could be put in the extension potentially... And just set to 0 if its not a ACK either way.

dogmaphobic:
Do you mean adding the `param_result` to the end of `PARAM_VALUE`? Wouldn't changing the structure of the `PARAM_VALUE` message cause issues with one side using the new structure and the other using the old structure?

WickedShell:
Add it as a mavlink 2 extension, rather then editing the message definition I mean. As long as 0's mean ignore/do nothing then it will be fine.

dogmaphobic:
I'm not getting it. Are you saying that if I simply add the field to the end of the message it won't affect old code?

```
<message id="22" name="PARAM_VALUE">
  <description>Emit the value of a onboard parameter.</description>
  <field type="char[16]" name="param_id">Onboard parameter id</field>
  <field type="float" name="param_value">Onboard parameter value</field>
  <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Onboard parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
  <field type="uint16_t" name="param_count">Total number of onboard parameters</field>
  <field type="uint16_t" name="param_index">Index of this onboard parameter</field>
</message>
```

The (old) code itself won't care because it doesn't know about the field so it won't even look at it. The question is about the message unpacking. Will it handle a message with some few extra bytes at the end?

WickedShell:
@dogmaphobic Are you familiar with the extension fields for MAVLink2? If a packet is encoded in MAVLink 2 there can be a set of optional additional fields that are extensions that are packed after the rest of the packet. To add it to a packet defintion, you need to just add it after an <extension /> tag. IE if you see here: https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2829

dogmaphobic:
>Are you familiar with the extension fields for MAVLink2?

No, I'm not. I have done little to nothing related to V2.

Does it mean all I need to to is this?

```
<message id="22" name="PARAM_VALUE">
  <description>Emit the value of a onboard parameter.</description>
  <field type="char[16]" name="param_id">Onboard parameter id</field>
  <field type="float" name="param_value">Onboard parameter value</field>
  <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Onboard parameter type: see the MAV_PARAM_TYPE enum for supported data types.</field>
  <field type="uint16_t" name="param_count">Total number of onboard parameters</field>
  <field type="uint16_t" name="param_index">Index of this onboard parameter</field>
  <extensions/>
</message>
```

WickedShell:
@dogmaphobic there is some documentation on the extension stuff here: https://github.com/mavlink/mavlink/blob/master/doc/MAVLink2.md#handling-message-extensions

That would be a correct way to present an extension. Because (as noted in the linked documents) the extension fields will all be null unless the sender spoke MAVLink2 as well, you need to ensure that a value of 0 in PARAM_ACK means do not use/unknown. (And would imply the current behavior)

dogmaphobic:
Got it. I now understand how it works and it would be a good option with one exception: If instead of using a specialized `PARAM_ACK` response, we retrofit the `result` into `PARAM_VALUE`, when a new process sends a `PARAM_VALUE` with `result` set to `PARAM_ACK_IN_PROGRESS`, an older process will take it as a regular ACK and when the subsequent `PARAM_VALUE` with the actual response arrives, it won't know what to do with it.

dogmaphobic:
After having a discussion with @LorenzMeier, it makes more sense to have a new, separate protocol for this. This not only gives more granular feedback but also allows for a larger payload (I went back to the original idea of increasing the value payload to char[128]). The extended payload will allow sending entire objects at once as opposed to field by field as it is done today.

WickedShell:
@dogmaphobic why not extend the name from 16 characters? This is quite an apparent issue at the moment, even if the autopilot didn't want to work with longer ones yet it would be nice for the transport to support longer parameters.

dogmaphobic:
>why not extend the name from 16 characters? 

I have no issues with it if everybody else agrees. This is still in flux as I develop the new camera settings.

For instance, we may also need to work on some new union (or equivalent) to replace `mavlink_param_union_t` and/or `mavlink_param_union_double_t` to deal with the 128-byte value.

LorenzMeier:
The union is unused - we can just adapt it. In terms of the string lengths: We can probably safely increase, however, this adds up for every transfer - and autopilots will be for the time being bandwidth-constrained. So I would consider extending to 24, but going to 32 bytes would seem excessive.

dogmaphobic:
As I said, it doesn't make that much difference to me. In my particular case, these parameter names are only used in code (machine) and are not meant for human consumption. The current number of possible permutations already far exceeds the number of possible parameters.

But in case someone does change it, make sure to create a new `MAVLINK_MSG_PARAM_VALUE_FIELD_PARAM_ID_LEN` for these "EXT" sets.

LorenzMeier:
That matches pretty much the way I look at it. Believing that these names are descriptive is anyway overly optimistic - there are over 900 parameters in PX4 and a lot of them are somewhat inter-dependent. Either you read the documentation for these and you get what you're doing, or you're not able to configure the system successfully. So they really mostly represent a hash map from UI or documentation to a variable these days.

PR #758, opened 2017-07-20
Final step in moving camera settings to PARAM_EXT_xxx

dogmaphobic:
Clean up unused camera settings messages and commands.
Remove all references to camera ID (use component ID instead).


dogmaphobic:
>There are still a few more "camera IDs" but I guess we can't remove them.

Right, those are related to video streaming and don't quite apply here.

PR #764, opened 2017-07-25
Add a terminating state

WickedShell:
It would be nice to have an unequivocal indication to a GCS/operator that the system is performing a commanded flight termination. MAV_STATE_EMERGENCY could be sent for this, but there are other reasons to get into that state, so I'd prefer to have an unambiguous message for termination if that's fine with others.


LorenzMeier:
Thanks!

PR #767, opened 2017-08-05
Mission: Add command to delay state machine until position gate has b…

LorenzMeier:
…een reached.

This helps to structure missions where certain 3D coordinates need to be reached before continuing the mission plan, but where adding a state to the navigation controller is undesirable.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1537,6 +1537,16 @@
         <param index="5">Unscaled target latitude of center of circle in CIRCLE_MODE</param>
         <param index="6">Unscaled target longitude of center of circle in CIRCLE_MODE</param>
       </entry>
+      <entry value="4501" name="MAV_CMD_CONDITION_GATE">
+        <description>WIP: Delay mission state machine until gate has been reached.</description>
+        <param index="1">Geometry: 0: orthogonal to path between previous and next waypoint.</param>
+        <param index="2">Altitude: 0: ignore altitude</param>
+        <param index="3">Empty</param>
+        <param index="4">Empty</param>
+        <param index="5">Latitude</param>
+        <param index="6">Longitude</param>
+        <param index="7">Altitude</param>
+      </entry>
       <entry value="5000" name="MAV_CMD_NAV_FENCE_RETURN_POINT">
         <description>Fence return point. There can only be one fence return point.
         </description>


LorenzMeier:
I'm rebasing and merging as non-intrusive, but if there is feedback I'm happy to accommodate it as needed.

DonLakeFlyer:
@LorenzMeier This is the Survey transect exit image thing right?

LorenzMeier:
Correct.

LorenzMeier:
I have a branch for QGC and Firmware, will push later.

DonLakeFlyer:
@WickedShell This is to get the last image taken at the transect exit point.

WickedShell:
@DonLakeFlyer What is the vehicle expected to do with this command? I don't really see what it's proposing that the aircraft do.

DonLakeFlyer:
It forces the next mission command to wait until the vehicle has passed a line orthogonal to it's current direction of travel and the set location. You can then use this to get more predictable triggering of the last image on a transect without being affected by acceptance radius flakiness. So if you have survey polygon and a transect along it with a turnaround. You can set one of these at the survey polygon exit intersection. And then stop triggering after then and then go to your turnaround point. Then it guarantess you get the last image. If you then use this with the trigger start mode of trigger immediately. You can get full coverage at the polygon edges without needing to fudge the survey polygon large, or takes images through the turns.

WickedShell:
@DonLakeFlyer Would a reasonable summary be that this is attempting to force a fly over rather then fly by waypoint on the vehicle? (Which is so is probably easier to approach with just turning one of the normal params into fly over/fly by as needed).

The behavior as you've described has knock on effects for the overshoots/turn radius needed to be back on the line which the user could compensate for by raising their extension/lead in terms.

There are a lot of cases where as far as I can tell the behavior of this is undefined such as a loiter to altitude/turns. In fact the only command this makes any sense on really is MAV_CMD_NAV_WAYPOINT (and arguably SPLINE waypoints).

The other problem I see with this proposal is that it is just burning command storage space. I'm counting this as 7 commands per flight line (Lead in, polygon edge, camera on, polygon edge, this state machine command, camera off, lead out) from at least an ArduPilot side that is insane as it limits the vehicle to ~103 flight lines, a number I routinely map past.

DonLakeFlyer:
> Would a reasonable summary be that this is attempting to force a fly over rather then fly by waypoint on the vehicle?

That is correct.

> The other problem I see with this proposal is that it is just burning command storage space. I'm counting this as 7 commands per flight line

Almost. It's 6 commands per transect. Lead in, polygon edge, camera on, state gate (which is at polygon edge), camera off, lead out,

It's up to you to decide what you want. If you don't want images in the turns screwing up a capture then this will give you full coverage. QGC has customers which require this.

If you want a lower command count and are ok with taking images in the turns then you use trigger immediate at the first entry polygon edge and state gate and the last polygon edge exit. And then you get full coverage of the polygon for that case as well.

> The behavior as you've described has knock on effects for the overshoots/turn radius needed to be back on the line which the user could compensate for by raising their extension/lead in terms.

Don't understand what you mean here?

PR #768, opened 2017-08-08
Arm authorization messages

zehortigoza:
Allow vehicles to request flight authorization before arm.

Government entities like NASA are working in systems to managed the aero space for drones(https://utm.arc.nasa.gov/) but this also can have another uses by private companies.

More information about how it should work:
PX4 implementation: https://github.com/PX4/Firmware/pull/7377
Arm authorizer implementation: https://github.com/01org/mavlink-router/pull/94

I also plan to implement this in Ardupilot after implement in PX4.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3209,7 +3209,10 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID, as defined by MAV_CMD enum.</field>
       <field type="uint8_t" name="result" enum="MAV_RESULT">See MAV_RESULT enum</field>
       <extensions/>
-      <field type="uint8_t" name="progress">WIP: Needs to be set when MAV_RESULT is MAV_RESULT_IN_PROGRESS, values from 0 to 100 for progress percentage, 255 for unknown progress.</field>
+      <field type="uint8_t" name="progress">WIP: Also used as result_param1, it can be set with a enum containing the errors reasons of why the command was denied or the progress percentage or 255 if unknown the progress when result is MAV_RESULT_IN_PROGRESS.</field>
+      <field type="int32_t" name="result_param2">WIP: Additional parameter of the result, example: which parameter of MAV_CMD_NAV_WAYPOINT caused it to be denied.</field>
+      <field type="uint8_t" name="target_system">WIP: System which requested the command to be executed</field>
+      <field type="uint8_t" name="target_component">WIP: Component which requested the command to be executed</field>
     </message>
     <message id="81" name="MANUAL_SETPOINT">
       <description>Setpoint in roll, pitch, yaw and thrust from the operator</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1550,6 +1550,11 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
+      <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST">
+        <description>Request authorization to arm the vehicle to a external entity, the arm authorizer is resposible to request all data that is needs from the vehicle before authorize or deny the request. If approved the progress of command_ack message should be set with period of time that this authorization is valid in seconds or in case it was denied it should be set with one of the reasons in ARM_AUTH_DENIED_REASON.
+        </description>
+        <param index="1">Vehicle system id, this way ground station can request arm authorization on behalf of any vehicle</param>
+      </entry>
       <entry value="5000" name="MAV_CMD_NAV_FENCE_RETURN_POINT">
         <description>Fence return point. There can only be one fence return point.
         </description>
@@ -2630,6 +2635,26 @@
       </entry>
       <entry value="2" name="CAMERA_MODE_IMAGE_SURVEY">
         <description>Camera is in image survey capture mode. It allows for camera controller to do specific settings for surveys.</description>
+      </entry>
+    </enum>
+    <enum name="MAV_ARM_AUTH_DENIED_REASON">
+      <entry value="0" name="MAV_ARM_AUTH_DENIED_REASON_GENERIC">
+        <description>Not a specific reason</description>
+      </entry>
+      <entry value="1" name="MAV_ARM_AUTH_DENIED_REASON_NONE">
+        <description>Authorizer will send the error as string to GCS</description>
+      </entry>
+      <entry value="2" name="MAV_ARM_AUTH_DENIED_REASON_INVALID_WAYPOINT">
+        <description>At least one waypoint have a invalid value</description>
+      </entry>
+      <entry value="3" name="MAV_ARM_AUTH_DENIED_REASON_TIMEOUT">
+        <description>Timeout in the authorizer process(in case it depends on network)</description>
+      </entry>
+      <entry value="4" name="MAV_ARM_AUTH_DENIED_REASON_AIRSPACE_IN_USE">
+        <description>Airspace of the mission in use by another vehicle, second result parameter can have the waypoint id that caused it to be denied.</description>
+      </entry>
+      <entry value="5" name="MAV_ARM_AUTH_DENIED_REASON_BAD_WEATHER">
+        <description>Weather is not good to fly</description>
       </entry>
     </enum>
   </enums>


magicrub:
why not extend the existing _ACK ?

zehortigoza:
> why not extend the existing _ACK ?

I did not wanted to changed the type of progress to uint16_t(it could break something already using it) and I think for a more general purpose result parameter a uint8_t it not enough. Also it would cause more struct unalignment.

magicrub:
but progress is uint8_t in your version?

Also, MAV_ARM_AUTH_DENIED_REASON is defined but never used.

zehortigoza:
> but progress is uint8_t in your version?
>
> Also, MAV_ARM_AUTH_DENIED_REASON is defined but never used. 

In ACK2 I think a more generic field is better. If the result is PROGRESS the result_param1 can be with the percentage, if is DENIED any enum with the errors can be used, that is why MAV_ARM_AUTH_DENIED_REASON is not used here.

vooon:
Why that should be command? Why not just add two messages for that particular task?

zehortigoza:
> Why that should be command? Why not just add two messages for that particular task?

I think is better reuse the generic command and only create new messages if really needed(when it need more fields or fields of different types), also this make easy to be implemented as most implementations only will need to add one more 'case' to his command handling.

dogmaphobic:
Instead of creating a new ACK message response, wouldn't it make more sense to just use an extension to the existing one?

>Why that should be command? Why not just add two messages for that particular task?

The main reason I can think of is the automatic retry mechanism you get by using a command, whereas if you were to create a new set of messages for this, we would have to implement new timers and retry algorithms and propagate this throughout the system (QGC for instance). On the other hand, the ACK2 message seems a bit like a hack to me.

@DonLakeFlyer?

DonLakeFlyer:
What @dogmaphobic said is correct.

TSC21:
> whereas if you were to create a new set of messages for this, we would have to implement new timers and retry algorithms and propagate this throughout the system (QGC for instance).

At least that would be consistent. Is a trade-off for the ease of implementation. In the case of MAVROS side, it would be the same.

DonLakeFlyer:
The fact is that sending a non-COMMAND_LONG to ask the vehicle to do something flat out doesn't work as a protocol. A ground station has to jump through crazy hoops whenever it needs to use a message like that. If you are asking the vehicle to do something that is a "command", hence you should use a command api.

LorenzMeier:
@zehortigoza Given how much effort is in the deployed _ACK implementations already and given there is a progress field (https://github.com/zehortigoza/mavlink/blob/27656c5f60f36e316c4f63ce14ed97fa8e8c64fc/message_definitions/v1.0/common.xml#L3201-L3207), could you please try to fit your objective into the existing message?

We're talking about man-months of implementation effort to get ACK2 everywhere supported and stable (QGC, Firmware, mavlink-router, language bindings, Dronecode SDK) and while that should never be a reason to reject progress, it seems a bit over the top for what you need for the arm authorisation.

zehortigoza:
The effort to get ACK2 everywhere is lower than implement the extended version of ACK everywhere, you don't need to switch to ACK2 unless you need the additional fields. Also look at the progress extension it is still not implemented in PX4 or Ardupilot.

But for sure I could use something like this(sorting this way to keep it as aligned as possible):

```
<message id="77" name="COMMAND_ACK">
      <description>Report status of a command. Includes feedback whether the command was executed.</description>
      <field type="uint16_t" name="command" enum="MAV_CMD">Command ID, as defined by MAV_CMD enum.</field>
      <extensions/>
      <field type="uint8_t" name="progress">Also used as result_param1, it can be set with a enum containing the errors reasons of why the command was denied or the progress percentage when result is MAV_RESULT_IN_PROGRESS.</field>
      <field type="int32_t" name="result_param2">Additional parameter of the result, example: which parameter of MAV_CMD_NAV_WAYPOINT caused it to be denied.</field>
      <field type="uint8_t" name="target_system">System which requested the command to be executed</field>
      <field type="uint8_t" name="target_component">Component which requested the command to be executed</field>
</message>
```

magicrub:
I think this should either extend ACK1 or be a completely different packet named something like AUTH_RESPONSE

dogmaphobic:
>I think this should either extend ACK1 or be a completely different packet named something like AUTH_RESPONSE

That makes sense. A regular ACK stating the command was received and a new message with whatever content it wants.

zehortigoza:
I think you guys can not see the usage of this additional fields with the current commands, here two examples:
- Example 1

	In a setup with 3 systems connected:
	- vehicle
	- GSC
	- custom application

	Both GSC and custom application send each a MAV_CMD_SET_MESSAGE_INTERVAL to vehicle in a similar time, the one sent by GCS is accepted the one sent by custom application is denied because the interval is too low. The vehicle will send back 2 ACKs, both GCS and custom application will receive both ACKs and it will not know for sure what ACK to use.

- Example 2

	A GSC or a custom application controlling the vehicle servos send 3 MAV_CMD_DO_SET_SERVO one for each servo, one of the 3 messages is lost in the link, the vehicle will send back 2 acks, the GSC or a custom application can not know what MAV_CMD_DO_SET_SERVO was not executed.

Also when porting current ACK to ACK2 could help make GSC notifications more clean, right it will announce the result of command that it did not sent.

- Now why it is needed for arm authorization:

	In a setup with 4 system connected:
	- 2 vehicles
	- 1 arm authorizer
	- 1 custom application controlling the swarm

	The custom application send to the swarm different missions and ask then to arm and takeoff. Each vehicle will ask arm authorization, arm authorizer need to be able to send to each vehicle the individual answer.

COMMAND_ACK do not cover cases were more system are connected.

zehortigoza:
New version removing ack2 and extending ack.

message_definitions/v1.0/common.xml
```diff
@@ -3209,7 +3234,10 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID, as defined by MAV_CMD enum.</field>
       <field type="uint8_t" name="result" enum="MAV_RESULT">See MAV_RESULT enum</field>
       <extensions/>
-      <field type="uint8_t" name="progress">WIP: Needs to be set when MAV_RESULT is MAV_RESULT_IN_PROGRESS, values from 0 to 100 for progress percentage, 255 for unknown progress.</field>
+      <field type="uint8_t" name="progress">WIP: Also used as result_param1, it can be set with a enum containing the errors reasons of why the command was denied or the progress percentage or 255 if unknown the progress when result is MAV_RESULT_IN_PROGRESS.</field>
+      <field type="int32_t" name="result_param2">WIP: Additional parameter of the result, example: which parameter of MAV_CMD_NAV_WAYPOINT caused it to be denied.</field>
+      <field type="uint8_t" name="target_system">WIP: System which requested the command to be executed</field>
```
LorenzMeier: Have you tested if the current MAVLink 2 library gives you the correct offset of the target_system field? There is a special function to request it.

While we are at it: We have a weakness in the protocol when we loose ACK's: We have retransmission for commands, but if the ACK is lost, the command might be executed multiple times. Could you look at that too? It would require a sequence / seq field in both the command and the ACK. The command should monotonically increase it, the ack should repeat it. If the seq is not increasing the receiving entity should not repeat the command execution. @dogmaphobic FYI.

I can also already say that we will have a backwards compatibility challenge: MAVLink 1 users will have the seq default to zero, which means that from the system perspective it will look like the command sequence zero is being retransmitted all the time. We either need to start the seq at 1 (which is horrible for wrapping, as we need to wrap from uint8_t max to 1), or we need to have another field indicating if the sequence should be used, which defaults to zero. @dogmaphobic FYI

LorenzMeier:
While we are at it: We have a weakness in the protocol when we loose ACK's: We have retransmission for commands, but if the ACK is lost, the command might be executed multiple times. Could you look at that too? It would require a sequence / seq field in both the command and the ACK. The command should monotonically increase it, the ack should repeat it. If the seq is not increasing the receiving entity should not repeat the command execution. @dogmaphobic FYI.

I can also already say that we will have a backwards compatibility challenge: MAVLink 1 users will have the seq default to zero, which means that from the system perspective it will look like the command sequence zero is being retransmitted all the time. We either need to start the seq at 1 (which is horrible for wrapping, as we need to wrap from uint8_t max to 1), or we need to have another field indicating if the sequence should be used, which defaults to zero.

message_definitions/v1.0/common.xml
zehortigoza: Just tested with: const mavlink_msg_entry_t *msg_entry = mavlink_get_msg_entry(MAVLINK_MSG_ID_COMMAND_ACK);
msg_entry->target_system_ofs, msg_entry->target_component_ofs
and both are correctly set.

zehortigoza:
> While we are at it: We have a weakness in the protocol when we loose ACK's: We have retransmission for commands, but if the ACK is lost, the command might be executed multiple times. Could you look at that too? It would require a sequence / seq field in both the command and the ACK. The command should monotonically increase it, the ack should repeat it. If the seq is not increasing the receiving entity should not repeat the command execution. @dogmaphobic FYI.
I can also already say that we will have a backwards compatibility challenge: MAVLink 1 users will have the seq default to zero, which means that from the system perspective it will look like the command sequence zero is being retransmitted all the time. We either need to start the seq at 1 (which is horrible for wrapping, as we need to wrap from uint8_t max to 1), or we need to have another field indicating if the sequence should be used, which defaults to zero.

The implementations could use a int16_t and handle negative number as not having support to sequence but still wrap as uint8_t.

I can work on that but after I finish the arm authorization stuff.

PR #775, opened 2017-09-11
Fix typo in explanation of vz

h4writer:
The description of GLOBAL_POSITION_INT mentions the following: "The position is in GPS-frame (right-handed, Z-up)". Now the explanation used for vz is saying "Ground Z Speed (Altitude, positive down), expressed as m/s * 100". One of the two is fault. The simulator is giving Z-up. I.e. the vz description is fault.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2878,7 +2878,7 @@
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground in meters, expressed as * 1000 (millimeters)</field>
       <field type="int16_t" name="vx" units="cm/s">Ground X Speed (Latitude, positive north), expressed as m/s * 100</field>
       <field type="int16_t" name="vy" units="cm/s">Ground Y Speed (Longitude, positive east), expressed as m/s * 100</field>
-      <field type="int16_t" name="vz" units="cm/s">Ground Z Speed (Altitude, positive down), expressed as m/s * 100</field>
+      <field type="int16_t" name="vz" units="cm/s">Ground Z Speed (Altitude, positive up), expressed as m/s * 100</field>
       <field type="uint16_t" name="hdg" units="cdeg">Vehicle heading (yaw angle) in degrees * 100, 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
     </message>
     <message id="34" name="RC_CHANNELS_SCALED">


amilcarlucas:
Any progress on this ? is this true ?

amilcarlucas:
Was this tested with ardupilot ? There was a similar bug that got fixed recently in ardupilot master.
Can you retest with the latest master ?

amilcarlucas:
Seven months later, still no reply ... and now it needs a rebase.

hamishwillee:
This seems correct to me (it is possible but not sensible that the position and velocity might define the direction of Z differently). @amilcarlucas What is this waiting on - just someone to approve it?

amilcarlucas:
Yes, this just needs someone to confirm it and approve it

julianoes:
Thanks @h4writer.

amilcarlucas:
I messed up. I failed to ask around in the Ardupilot team. Basically it breaks the ardupilot code and a lot of ground stations. So the Ardupilot team opposes to this change.

Can this please be reverted ?

h4writer:
I don't get this sentiment? Why would it break the ardupilot code?
This is only a change in text. It corrects the text to say what it actually does, instead of saying something bogus.

amilcarlucas:
Because the ardupilot honors that comment and points down (NED).

By just changing the comment to NEU, you effectively break the code.

julianoes:
Well, I did not review this correctly. Everything in PX4 is NED as far as I know and I completely misread this diff.

WickedShell:
Yeah, my brief investigation shows that PX4 was sending NED before (I didn't spend long enough to find the structure definition in current PX4 master). Can we agree that this was the wrong change and sending NED is correct and fix the top level description to be consistent? (Since all major projects/GCS's are expecting the previous behavior I think that makes more sense).

mhkabir:
@WickedShell is correct. The field definition was fine before. The wrong part was the description of the message ("The position is in GPS-frame (right-handed, Z-up)").

julianoes:
> Can we agree that this was the wrong change and sending NED is correct and fix the top level description to be consistent?

Yes, I agree!

amilcarlucas:
Just hold a second, I need to get some input from @rmackay9 on this thing. I hope we can clarify this in the next few days.

julianoes:
Here is the PR to undo this: #892

julianoes:
> And PX4 was also using NED up until this commit 23 days ago: PX4/Firmware@deb5cbc

I don't think it was changed in there @amilcarlucas.

amilcarlucas:
ardupilot/ardupilot#8192 explains the ardupilot side of things. But I still need the feedback from @rmackay9 on this. I was at the meeting where that commit was discussed and approved. I know the background for the decision. But the final word comes from Randy. I hope he replies in the next few days.

PR #778, opened 2017-09-20
add distance sensor orientation convention

ChristophTobler:
Before there was no convention on the distance sensor convention.
See also https://github.com/PX4/Devguide/pull/269 and https://github.com/PX4/Firmware/pull/7984

I added it to `MAV_SENSOR_ORIENTATION` as the distance sensor is the only sensor using it so far. If that should stay general we can add the comments to the `DISTANCE_SENSOR` message.


ChristophTobler:
info: I wouldn't merge this before we agree in https://github.com/PX4/Firmware/pull/7984

PR #785, opened 2017-10-09
extend attitude_quaternion_cov message with covariance matrix…

mrivi:
for quaternions


mrivi:
@julianoes Could you please have a look? Thank you

PR #786, opened 2017-10-09
mavlink_autopilot_version_t: add extended uid2 hardware identifier

bkueng:
Newer HW chips provide a 128 bit identifier. To make use of that we extend
the message with a longer uid, and use an uint8 array, so that the byte
ordering is clear.
The new field is 18 bytes, so that 2 bytes can be used to differentiate
between multiple SoC's.


magicrub:
Please make it an extension instead because simply changing it breaks compatibility.

bkueng:
> Please make it an extension instead because simply changing it breaks compatibility.

I did. Did I use the wrong syntax?

magicrub:
oh, weird. I looked for it but didn't see it. Yes, that looks good.

magicrub:
+1

LorenzMeier:
@magicrub Since you're happy with that and Beat and I have discussed it extensively I'm merging. Thanks!

PR #790, opened 2017-10-27
Add messages for command access control (for offboard modules)

JonasVautherin:
Messages and their use-case are defined more precisely in [this document](https://docs.google.com/document/d/1eMYBJRMl9JzqvydidBRB3IP8ZB16Z6d-zoegr8RNuaY/edit?usp=sharing).

To summarize:

* COMPONENT_IN_CONTROL: regularly published by the FCU to indicate which component is currently in control. By default, it is the FCU itself.
* TAKE_CONTROL: request control. The FCU may or may not accept it depending on the priority of the calling component and the current state.
* RELEASE_CONTROL: release control, effectively giving it back to the FCU.

I am not sure if I should define messages for the replies, as TAKE_CONTROL might be refused. @LorenzMeier: Do you have an opinion on that?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1352,6 +1352,26 @@
         <param index="6">Longitude / Y position</param>
         <param index="7">Altitude / Z position</param>
       </entry>
+      <entry value="253" name="MAV_CMD_TAKE_CONTROL">
+        <description>Request navigation control. Depending on the situation, the FCU might accept or reject it and update the COMPONENT_IN_CONTROL.</description>
+        <param index="1">Empty</param>
+        <param index="2">Empty</param>
+        <param index="3">Empty</param>
+        <param index="4">Empty</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
+      <entry value="254" name="MAV_CMD_RELEASE_CONTROL">
+        <description>Release navigation control, effectively making the FCU become the next COMPONENT_IN_CONTROL.</description>
+        <param index="1">Empty</param>
+        <param index="2">Empty</param>
+        <param index="3">Empty</param>
+        <param index="4">Empty</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
       <entry value="300" name="MAV_CMD_MISSION_START">
         <description>start running a mission</description>
         <param index="1">first_item: the first mission item to run</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4397,5 +4397,17 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type: see the MAV_PARAM_EXT_TYPE enum for supported data types.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code: see the PARAM_ACK enum for possible codes.</field>
     </message>
+    <message id="325" name="COMPONENT_IN_CONTROL">
+      <description>Emit the id and name of the component having control, i.e. being the one allowed to navigate the drone. It can be the FCU, or an external module through MAVLink or FastRTPS.</description>
+      <field type="char[32]" name="component_name">Human readable component name</field>
+      <field type="uint8_t" name="component_id">Component ID</field>
+      <field type="uint8_t" name="system_id">System ID</field>
+    </message>
+    <message id="326" name="TAKE_CONTROL">
```
LorenzMeier: This should be a command ID instead.

message_definitions/v1.0/common.xml
```diff
@@ -4397,5 +4397,17 @@
...
+      <description>Request navigation control. Depending on the situation, the FCU might accept or reject it and update the COMPONENT_IN_CONTROL.</description>
+    </message>
+    <message id="327" name="RELEASE_CONTROL">
```
LorenzMeier: This should be a command ID instead.

message_definitions/v1.0/common.xml
```diff
@@ -4397,5 +4397,17 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type: see the MAV_PARAM_EXT_TYPE enum for supported data types.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code: see the PARAM_ACK enum for possible codes.</field>
     </message>
+    <message id="325" name="COMPONENT_IN_CONTROL">
+      <description>Emit the id and name of the component having control, i.e. being the one allowed to navigate the drone. It can be the FCU, or an external module through MAVLink or FastRTPS.</description>
+      <field type="char[32]" name="component_name">Human readable component name</field>
```
bkueng: Where does the FCU get this name from? If it's just from the [MAV_COMPONENT](http://mavlink.org/messages/common) ID then I think we don't need it here.

JonasVautherin: I don't now how the component id looks like. I was thinking about a human-readable name that we would define somewhere (e.g. in the priority list?), and that would be helpful for debugging and potentially in the UI (I'd rather see "component in control: avoidance" than "component in control: 12" in QGC).

bkueng: Search for `MAV_COMPONENT` on http://mavlink.org/messages/common: there's a predefined set of ID's that we could extend to add avoidance etc. Then QGC can have it's own string lookup table.

JonasVautherin:
@LorenzMeier: I've put them as commands, as requested. I was not completely sure about the id they should get, but I did not find a better place.

@bkueng: I removed the component name. But that means that if a module is not defined in the MAV_COMPONENT list, then one has to know the id somehow. Which is fine, I guess.

PR #791, opened 2017-10-31
Generate markdown to messages/common.md

hamishwillee:
This generates markdown into a folder and names as we wish to use it in docs. The reason for putting into a  folder is that eventually we might generate docs for other message types.


hamishwillee:
Merged - I'm currently the only "client" of this code.

PR #792, opened 2017-10-31
Obstacle Distance Message 

mrivi:
@dogmaphobic @julianoes

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4397,5 +4397,12 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type: see the MAV_PARAM_EXT_TYPE enum for supported data types.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code: see the PARAM_ACK enum for possible codes.</field>
     </message>
+    <message id="330" name="OBSTACLE_DISTANCE">
+      <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
+      <field type="uint8_t" name="estimator_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
+      <field type="uint8_t[12]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means no obstacle. In a array element, each unit corresponds to 1m.</field>
+      <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
+    </message>
   </messages>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4397,5 +4397,12 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type: see the MAV_PARAM_EXT_TYPE enum for supported data types.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code: see the PARAM_ACK enum for possible codes.</field>
     </message>
+    <message id="325" name="OBSTACLE_DISTANCE" >
+      <description>Obstacle distances in front of the camera, starting from the left in increment degrees to the right</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
+      <field type="uint8_t" name="estimator_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
+      <field type="uint8_t[12]" name="distances">Distance of obstacles in front of the camera staring on the left side. A value of 0 means no obstacle. In a array element, each unit corresponds to 1m.</field>
```
julianoes: What about also having a value for unknown such as `255` ?

mrivi: I am not sure I get what you mean. 255 to indicate no obstacle in a certain array element?

julianoes: What value would you send if you don't know if there is an obstacle, so you're uncertain.

lbegani: > staring on the left side

Typo?

lbegani: > front of the camera

Should the term be camera or sensor?

message_definitions/v1.0/common.xml
```diff
@@ -4397,5 +4397,12 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type: see the MAV_PARAM_EXT_TYPE enum for supported data types.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code: see the PARAM_ACK enum for possible codes.</field>
     </message>
+    <message id="325" name="OBSTACLE_DISTANCE" >
```
lbegani: Is it possible to have multiple sensors mounted to detect obstacles? Then, how to identify the sensor that this message corresponds to?

mrivi: My idea was to have this message to show already aggregated data from multiple sensors. @julianoes what do you think?

julianoes: @lbegani several sensors could send this message. You can then identify who sent it by the mavlink component ID.

lbegani: Sounds good.

PR #798, opened 2017-11-15
Ardupilot: Timestamp added to RPM message.

pavloblindnology:
This PR is done for better synchronization of RPMs of wheel encoders connected to autopilot on a companion computer (e.g. to compute wheel odometry).


pavloblindnology:
After some considerations, it was decided to create a new message with distance traveled by each wheel (reverse rotation reducing this distance). This way the odometry can be computed more accurately comparing to RPMs.

PR #799, opened 2017-11-15
Ardupilot: message with wheels' cumulative distances added.

pavloblindnology:
This PR is aimed to make possible computing of wheel odometry on a companion computer with encoders connected to autopilot. Instead of using existing [RPM](http://mavlink.org/messages/ardupilotmega#RPM) message, this message allows more accurate odometry calculations.
Corresponding topic is discussed in mavros PR [#863](https://github.com/mavlink/mavros/pull/863).

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -1318,6 +1318,11 @@
       <field type="float" name="rpm1">RPM Sensor1</field>
       <field type="float" name="rpm2">RPM Sensor2</field>
     </message>
+    <message id="227" name="WHEEL_DISTANCE">
+      <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
+      <field type="float[2]" name="distance">Distance for each wheel (m)</field>
+    </message>
     <!-- ardupilot specific mavlink2 messages starting at 11000-->
     <message id="11000" name="DEVICE_OP_READ">
       <description>Read registers for a device</description>
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -1318,12 +1318,6 @@
       <field type="float" name="rpm1">RPM Sensor1</field>
       <field type="float" name="rpm2">RPM Sensor2</field>
     </message>
-    <message id="227" name="WHEEL_DISTANCE">
-      <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
-      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
-      <field type="uint8_t" name="count">Number of wheels reported</field>
-      <field type="float[4]" name="distance" units="m">Distance traveled by each wheel (m)</field>
-    </message>
     <!-- ardupilot specific mavlink2 messages starting at 11000-->
     <message id="11000" name="DEVICE_OP_READ">
       <description>Read registers for a device</description>
@@ -1388,6 +1382,13 @@
       <field type="float[3]" name="angle_delta">Defines a rotation vector in body frame that rotates the vehicle from the previous to the current orientation</field>
       <field type="float[3]" name="position_delta">Change in position in meters from previous to current frame rotated into body frame (0=forward, 1=right, 2=down)</field>
       <field type="float" name="confidence">normalised confidence value from 0 to 100</field>
+    </message>
+    <!-- wheel traveled distance message -->
+    <message id="11020" name="WHEEL_DISTANCE">
+      <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
+      <field name="time_usec" type="uint64_t">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
+      <field type="uint8_t" name="count">Number of wheels reported</field>
+      <field type="float[4]" name="distance" units="m">Distance traveled by each wheel (m)</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -1321,7 +1321,8 @@
     <message id="227" name="WHEEL_DISTANCE">
       <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
-      <field type="float[2]" name="distance">Distance for each wheel (m)</field>
+      <field type="uint8_t" name="count">Number of wheels reported</field>
+      <field type="float[4]" name="distance" units="m">Distance traveled by each wheel (m)</field>
     </message>
     <!-- ardupilot specific mavlink2 messages starting at 11000-->
     <message id="11000" name="DEVICE_OP_READ">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/ardupilotmega.xml
```diff
@@ -1318,6 +1318,11 @@
       <field type="float" name="rpm1">RPM Sensor1</field>
       <field type="float" name="rpm2">RPM Sensor2</field>
     </message>
+    <message id="227" name="WHEEL_DISTANCE">
+      <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
+      <field type="float[2]" name="distance">Distance for each wheel (m)</field>
```
TSC21: Maybe make it `float[]` to the distance of N wheels.

TSC21: I don't think its possible though. Just a suggestion.

pavloblindnology: But variable-length arrays are not supported by mavlink, aren't they?

TSC21:
They are not that's why we should consider a further extension later. I guess we can set the size to 4 and have mavlink 2.0 drop the bytes it is not using.

pavloblindnology:
Don't quite understand... Is the next ok?

>     <message id="227" name="WHEEL_DISTANCE">
>      <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
>      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
>      <field type="uint8_t" name="count">Number of wheels reported</field>
>      <field type="float[4]" name="distance">Distance for each wheel (m)</field>
>    </message>

TSC21:
> Don't quite understand... Is the next ok?
```
<message id="227" name="WHEEL_DISTANCE">
 <description>Cumulative distances traveled by each wheel in meters (forward rotations increase these values, backward - decrease).</description>
 <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
 <field type="uint8_t" name="count">Number of wheels reported</field>
 <field type="float[4]" name="distance">Distance for each wheel (m)</field>
```

Yes it is.

pavloblindnology:
Can I know when it is going to be merged so that I can use it in Ardupilot?

TSC21:
> Can I know when it is going to be merged so that I can use it in Ardupilot?

Unfortunately it does not depend on me. @LorenzMeier @tridge @WickedShell can you take a look please? Thanks

mhkabir:
Please change the timestamp to microseconds.

pavloblindnology:
@mhkabir Inside Ardupilot the wheel encoder timestamps are in milliseconds, that is why I chose them - to save extra 4 bytes. If there is some convention please let me know.

vooon:
I think better to choose Msg ID from v2 space.

pavloblindnology:
@vooon 11012 would be ok?
@mhkabir So what's about time stamp, is there some need in microseconds instead of milliseconds?

vooon:
Perhaps 11020 would be better, as it leave some space for `DEVICE_xxx` and `VISION_xxx`.
But better to hear APM guys opinion. @tridge what do you think?
Also i'm not sure where you should do PR for APM dialect, because they have it's own repo which periodically sync with this one.

Microsecond time-stamp be nice if anyone else will adopt this message in their FCU.
Also it preferred for new messages.

pavloblindnology:
APM repo: do you mean https://github.com/ArduPilot/mavlink which is forked from this repo?
How often do they sync and how fast do the changes propagate to ROS distro?
How to properly update my local config to test the changes? Right now I do it by modifying ros-mavlink includes (CRC values too) and ardupilot-mavlink message-definitions by hand, which is not the best way I suppose. Maybe there is some easier way.
Does anyone know? Thanks.

vooon:
For ROS i do monthly releases from this repo. If you need more recent messages, easiest way (source install) is to replace xml definition in `ros-ws/src/mavlink`.

pavloblindnology:
Moved to v2 space and changed time stamps to microseconds.

pavloblindnology:
Ardupilot's mavlink repo just has added 11020 msg ID, so I'll move to 11030.

pavloblindnology:
Can anyone tell me how to get messages from v2 space in mavros, please?
Should I set `fcu_protocol` to `v2.0` for mavros_node?
I set `SERIAL*_PROTOCOL` to 2.
After FCU restart, if `fcu_protocol` is set to `v1.0` then I don't get v2 space messages.
If `fcu_protocol` is set to `v2.0` then I do get v2 space messages.
If `fcu_protocol` is set to `v1.0` again (after it was `v2.0`) then I again do get v2 space messages.

Also, what `time/timesync_mode` should I set to be able to use time stamps from mavlink messages, MAVLINK, PASSTHROUGH? Should I also set `conn/timesync_rate`?
Thank you.

vooon:
APM not yet support timesync. With PASSTHROUGH you get timestamp from mavlink messages, since boot.
`fcu_protocol` only affects outgoing messages (all messages will be in v2 format).
Input parser understands both versions.

pavloblindnology:
@vooon Thank you. I'll surely check again but, as I said, I don't receive v2 messages in mavros if not setting `fcu_protocol` to `v2.0` at least once. They just don't get to mavlink's buffer parsing.
And with `time/timesync_mode` set to `MAVLINK` and `conn/timesync_rate` specified I get some time synchronization (corresponding messages show up, though time lag sometimes is unrealistic). Is it smth usable\reliable?

pavloblindnology:
@vooon Also, If using `fcu_protocol v2.0` then I cannot connect APM Planner through `udp` bridge.

LorenzMeier:
Not seeing this added downstream - closing.

pavloblindnology:
@LorenzMeier @TSC21  Where this should be added - upstream (here), downstream ([ardupilot/mavlink PR 55](https://github.com/ArduPilot/mavlink/pull/55)), both? In ardupilotmega.xml, or in common.xml?
Please advise, because everyone tells opposite things and PR remains not merged forever.
Thanks.

auturgy:
No worries, I understand. My point was that it should go in upstream common, or ardupilotmega.
Changing common.xml in ArduPilot  just makes it harder to keep in sync, and imho should be avoided.

pavloblindnology:
So, to conclude, I need to move this from `ArduPilot/mavlink/ardupilotmega.xml` to `ArduPilot/mavlink/common.xml`?
Also, what msgid should I use? Will current `11030` be ok to use in `common.xml`?

auturgy:
NO!
We shouldn’t be adding new messages to common.xml through the ArduPilot fork. It makes it impossible to sync between the various projects.
To try to keep things manageable:
If it’s for common, it should go straight to upstream, not via ArduPilot (and then be pulled downstream).
If it’s for ardupilotmega it should go to ArduPilot (and then be pushed upstream).

LorenzMeier:
I think for rovers I'd like to see an RFC with *all known rover related messages* and how they fit together to form a minimum viable set of messages for rovers:

https://github.com/mavlink/rfcs/blob/master/text/0001-rfc-format.md

auturgy:
Definitely merit for that, but this particular PR has been bounced around by both ends unnecessarily imho.
I think out of respect to the author (and in acknowledgment that the confusion here isn’t his fault), we should just make a decision. My view is that it’s probably suitable for common, but if you’d prefer to wait until a broader rover rfc discussion takes place, it can go in ardupilotmega downstream first and be revisited later (if necessary).

LorenzMeier:
@auturgy That's a fair point, but this has led to feature creep in the past, so maybe we could have a quick call to establish context.

pavloblindnology:
@auturgy @LorenzMeier So, what is your decision?

pavloblindnology:
@auturgy @LorenzMeier Any updates on this? We can close this PR if there is no interest in it at all. I will use it on my own. Just tell me, so that i can organize my time more effectively.

LorenzMeier:
@pavloblindnology I've resurrected this for the common message set here: https://github.com/mavlink/mavlink/pull/1032

PR #800, opened 2017-11-16
Update scripts to generate markdown message definitions

hamishwillee:
Scripts provide much better interlinking of generated message definitions. The tools are now run over all message definitions, not just the common.xml

### DIFF
--- a/doc/mavlink_gitbook.py
+++ b/doc/mavlink_gitbook.py
@@ -1,54 +1,47 @@
 #! /usr/bin/python
 
 """
-This script generates a markdown file (message-definitions.md) that can be imported into a gitbook to display the MAVLink Common message format (HTML).
-(this is a mechanism of bringing the current docs into the gitbook: http://mavlink.org/messages/common )
+This script generates markdown files for all the MAVLink message definition XML at: 
+https://github.com/mavlink/mavlink/tree/master/message_definitions/v1.0
+  
+The files can be imported into a gitbook to display the messages as HTML
 
 The script runs on both Python2 and Python 3. The following libraries must be imported: lxml, requests, bs4.
 
-The file is generated using an XSLT transformation from the XML at "https://raw.github.com/mavlink/mavlink/master/message_definitions/v1.0/common.xml". 
+The file is run in mavlink/doc/ with no arguments. It writes the files to /messages/
 """
 
 import lxml.etree as ET
 import requests
 from bs4 import BeautifulSoup as bs
 import re
-import os
+import os # for walk
 
-xml_file_url = "https://raw.github.com/mavlink/mavlink/master/message_definitions/v1.0/common.xml";
-xsl_file_url= "https://raw.github.com/mavlink/mavlink/master/doc/mavlink_to_html_table_gitbook.xsl";
+
+xsl_file_name = "mavlink_to_html_table_gitbook.xsl"
+xml_message_definitions_dir_name = "../message_definitions/v1.0/"
+
 output_dir = "./messages/"
-output_file_name = output_dir+"common.md"
 if not os.path.exists(output_dir):
     os.makedirs(output_dir)
-    
 
-# Get files from published URLs
-r = requests.get(xml_file_url)
-xml_file = r.text
-r = requests.get(xsl_file_url)
-xsl_file = r.text
+# File for index
+index_file_name = "README.md"
+index_file_name = output_dir + index_file_name
 
-dom = ET.fromstring(xml_file)
+# Get XSLT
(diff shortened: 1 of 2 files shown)


hamishwillee:
@dagar NOTE, the script is used in the same way as our previous one except now it parses files loaded locally rather than getting them from URLs. Jenkins will need to copy multiple files across to the mavlink devguide rather than just common.html. 
What needs to happen for this version to be used by Jenkins?

dagar:
This hasn't been running on jenkins yet because we need an environment (machine or docker) with the dependencies.
http://ci.px4.io:8080/job/mavlink/job/mavlink/

doc/mavlink_gitbook.py
```diff
@@ -1,55 +1,48 @@
...
 xslt = ET.fromstring(xsl_file)
 
-transform = ET.XSLT(xslt)
-newdom = transform(dom)
-
-#Prettify the HTML using BeautifulSoup
-soup=bs(str(newdom), "lxml")
-prettyHTML=soup.prettify()
+#initialise text for index file. 
+index_text='<!-- THIS FILE IS AUTO-GENERATED (DO NOT UPDATE GITBOOK): https://github.com/mavlink/mavlink/blob/master/doc/mavlink_gitbook.py -->'
```
julianoes: In general, I would recommend to use templates for this kind of auto-generated files. It makes it easier to separate text and scripts to populate it.

It's probably fine for now but something to keep in mind for the future.

doc/mavlink_gitbook.py
```diff
@@ -62,13 +55,85 @@ def remove_space_between_content_tags(matchobj):
...
+            dom = ET.fromstring(xml_file)
+            transform = ET.XSLT(xslt)
+            newdom = transform(dom)
+
+            #Prettify the HTML using BeautifulSoup
+            soup=bs(str(newdom), "lxml")
+            prettyHTML=soup.prettify()
+
+            # Strip out all text before "<html>" - not needed in the output
+            def slicer(my_str,sub):
```
julianoes: I find it weird to have the function inside here. Personally, I'd take it outside.

doc/mavlink_gitbook.py
```diff
@@ -62,13 +55,85 @@ def remove_space_between_content_tags(matchobj):
...
+            prettyHTML = fix_content_in_tags(prettyHTML)
+            
+            #Inject a heading and doc-type intro (markdown format)
+            prettyHTML = inject_top_level_docs(prettyHTML,file)
+            
+            #Replace invalid file extensions (workaround for xslt)
+            prettyHTML = fix_include_file_extension(prettyHTML)
+            
+            #Write output file
+            output_file_name = file[:-4]+".md"
```
julianoes: What's this `-4`, that's not clear to me.

doc/mavlink_gitbook.py
hamishwillee: > I find it weird to have the function inside here. Personally, I'd take it outside.

Will do!

doc/mavlink_gitbook.py
hamishwillee: > What's this -4, that's not clear to me.

It is stripping the file extension. I'll use an rsplit instead for a bit cleaner design.

doc/mavlink_gitbook.py
hamishwillee: Will do. This was actually a conscious decision as most of the auto generation is via the XSLT script. Didn't seem reasonable to set up a template for just a couple of files.

hamishwillee:
@julianoes 

> Have you tested it with Python2 and Python3?

I have now. It works :-)

Updated as you suggested - thanks. OK to merge?

doc/mavlink_gitbook.py
julianoes: I see :). You could also have just added a comment to explain what it does.

hamishwillee: Pah, why make life easy :-)

hamishwillee:
> This hasn't been running on jenkins yet because we need an environment (machine or docker) with the dependencies.
http://ci.px4.io:8080/job/mavlink/job/mavlink/

@dagar  Who do we have to annoy to move that job along?

dagar:
One of us. The build environment needs some basic python dependencies. Maybe we could simply use px4-dev-base?

hamishwillee:
>> @dagar Who do we have to annoy to move that job along?

> One of us. 

Work harder, not smarter!

> The build environment needs some basic python dependencies. Maybe we could simply use px4-dev-base?

Whatever works :-)

PR #801, opened 2017-11-16
Remove old message definitions (message_definitions/v0.9)

hamishwillee:
The old message definitions do not appear to still be used by anyone, and are potentially confusing.
Proposal is to remove them.

@julianoes Not sure who can approve this - your suggestion, and I think a good one.


julianoes:
I'm ok with this but we need to check with @LorenzMeier.

hamishwillee:
Ah, so the messages are [tested here](https://travis-ci.org/mavlink/mavlink/jobs/302821297#L1571). 

We will have to remove from the tests (somewhere in ./scripts/test.sh"), but I won't do that until @LorenzMeier confirms that it is OK to remove this set of messages.

julianoes:
@hamishwillee I see quite a bit of scripts and tools around that still include v0.9. I think there is more harm done by deleting it and it's probably easier just to leave it here.

However, I would omit it for any docs generation.

hamishwillee:
Thanks @julianoes 
That is pragmatic. Closed. I'm happy that decision made, for a reason.

PR #807, opened 2017-11-29
Update obstacle distance

mrivi:
@TSC21 it should address your concerns.
@julianoes is it okay?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4401,10 +4401,10 @@
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint16_t[72]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means no obstacle. A value of UINT16_MAX for unknown/not used. In a array element, each unit corresponds to 1m.</field>
+      <field type="uint16_t[72]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means that the obstacle is right in front of the sensor. A value of max_distance +1 means no obstace is present. A value of UINT16_MAX for unknown/not used. In a array element, each unit corresponds to 1cm.</field>
       <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
-      <field type="uint16_t" name="min_distance" units="m">Minimum distance the sensor can measure in meters</field>
-      <field type="uint16_t" name="max_distance" units="m">Maximum distance the sensor can measure in meters</field>
+      <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure in centimeters</field>
+      <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure in centimeters</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4401,7 +4401,7 @@
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint8_t[12]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means no obstacle. In a array element, each unit corresponds to 1m.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4400,9 +4400,11 @@
     <message id="330" name="OBSTACLE_DISTANCE">
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
-      <field type="uint8_t" name="estimator_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
+      <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
       <field type="uint8_t[12]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means no obstacle. In a array element, each unit corresponds to 1m.</field>
```
TSC21: should this be `uint16_t[12]`?

julianoes:
I think that's ok since it's quite new and no QGC release has been made with this.

TSC21:
There's also one thing I think we are missing here: one thing is having measurement = 0, meaning that there is no obstacle. But there there's a case I guess where the beam on a certain direction is not used/not activated (assuming we can use this msg for a group of sensors instead of a TOF type). And in that case we will want to state which sensor is reading and which is not. For that purpose, I suggest:
1. Changing the type to `uint16_t[12]`, for increased precision;
2. No obstacle, then `distances[i] = 0`, no beam/not used, `distances[i] = UINT16_MAX`. Should be referenced on the filed description on that case.

julianoes:
@TSC21 I agree that we should have a special value for not used but we should also have some value for "unknown".

TSC21:
> Well, on second thought, that could be the same.

Yeah, `UNKNOWN` or `NA` or `NOTUSED`. Turn out to be the same.

mrivi:
@TSC21 @julianoes since we are refactoring a bit, what do you think about increasing the number of elements of the array distances? Now we have 12 elements so if we have a setup to have distance data 360 degrees we can only have a resolution of 30 degrees.

TSC21:
> @TSC21 @julianoes since we are refactoring a bit, what do you think about increasing the number of elements of the array distances? Now we have 12 elements so if we have a setup to have distance data 360 degrees we can only have a resolution of 30 degrees.

Yeah I guess you can increase till you fit the payload of the message.

mrivi:
I have increased the array size to 72 to have a potential resolution of 5 degrees

TSC21:
> I have increased the array size to 72 to have a potential resolution of 5 degrees

Is the increment higher or lower than this?

mrivi:
Lower, it would be 5 if you use all the 72 bins. If the increment is 10 the  you would need only 36 bins, etc..

TSC21:
Yeah I guess that depends on the developer to set that accordingly.

julianoes:
So what is the distance if you are basically bumping into an object? 0 and therefore no obstacle?

mrivi:
For stereo cameras once you're closer to an obstacle than `min_distance` you won't have any indicative data of you're actual distance to the obstacle

TSC21:
Yah basically you can use the min_distance as a validation value for the readings. Actually I'm not seeing any sensor that currently can give you a minimum range of 0.0.

julianoes:
What if a new sensor comes around that claims it can do 0? Sorry if I'm going into an edge case here, I just find the choice for "no obstacle" of 0 weird.

dogmaphobic:
>I just find the choice for "no obstacle" of 0 weird.

What number will it show right when you hit it? :)

message_definitions/v1.0/common.xml
```diff
@@ -4400,9 +4400,11 @@
     <message id="330" name="OBSTACLE_DISTANCE">
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
-      <field type="uint8_t" name="estimator_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint8_t[12]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means no obstacle. In a array element, each unit corresponds to 1m.</field>
+      <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
+      <field type="uint16_t[72]" name="distances">Distance of obstacles in front of the sensor starting on the left side. A value of 0 means no obstacle. A value of UINT16_MAX for unknown/not used. In a array element, each unit corresponds to 1m.</field>
```
LorenzMeier: Let's use uint16 here and use 1 cm as unit. That would make it 650m maximum distance, which is probably reasonable.

LorenzMeier:
I would propose the following:

  * 1 unit = 1 cm
  * 0 = obstacle is right in front of sensor
  * max_range + 1: no obstacle (positively confirmed)
  * UINT16_MAX: no measurement

mrivi:
@LorenzMeier @julianoes I have addressed your comments.

PR #812, opened 2017-12-21
common: add second iteration of HIGH_LATENCY message

TSC21:
So the purpose of this iteration is to create a message that actually allows to establish a bidirectional high latency link between the ground station and the autopilot, by removing some fields that I considered may be secondary and complete with all the fields of the `HEARTBEAT` message. The removed fields are `gps_fix_type` and `temperature` and the added fields are `type`, `autopilot`, `system_status` and `mavlink_version`.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4068,6 +4068,35 @@
       <field type="uint8_t" name="wp_num">current waypoint number</field>
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
+    <message id="235" name="HIGH_LATENCY2">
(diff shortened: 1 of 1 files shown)


TSC21:
@dagar, @LorenzMeier, @dogmaphobic your reviews are welcomed.

TSC21:
The total number of bytes currently fit the max Iridium char size, which is 50.

dagar:
Why do you need system_status and mavlink version?

TSC21:
The idea was actually keep all fields of the HEARTBEAT msg.

dagar:
I originally wanted the same thing, but after going through it on the QGC side I saw that the entire thing wasn't actually needed. The full base_mode and custom_mode is nice, but I think it's a big waste of space for what info we actually need to get to the ground.

Issue #813, opened 2017-12-22
Generating for v2 CS have errors


tuloski:
CS\common\ByteArrayUtils.cs could not be found 
CS\common\FrameworkBitConverter.cs could not be found 
CS\common\Mavlink.cs could not be found 

but I can find the files in the common folder, but they are only for v1, i.e. they expect the package with this anatomy
                * 0  Packet start sign	
                * 1	 Payload length	 0 - 255
                * 2	 Packet sequence	 0 - 255
                * 3	 System ID	 1 - 255
                * 4	 Component ID	 0 - 255
                * 5	 Message ID	 0 - 255
                * 6 to (n+6)	 Data	 (0 - 255) bytes
                * (n+7) to (n+8)	 Checksum (high byte, low byte) for v0.9, lowbyte, highbyte for 1.0
while v2 requires a different anatomy.

amilcarlucas:
@hamishwillee This is a duplicate of #482 and can be closed.

PR #814, opened 2017-12-22
[WIP] HIGH_LATENCY2

dagar:
Update to HIGH_LATENCY.

Overall goal is 50 bytes total as a Mavlink v1.0 packet (8 byte header).

@acfloria @sanderux @TSC21 @philipoe @DonLakeFlyer @envirover

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -174,6 +174,195 @@
         <description>official stable release</description>
       </entry>
     </enum>
+    <enum name="FLIGHT_MODE">
+      <description>Flight modes of the MAV.</description>
+      <entry value="0" name="FLIGHT_MODE_MANUAL">
+        <description>Manual flight mode</description>
+      </entry>
+      <entry value="1" name="FLIGHT_MODE_ALTCTL">
+        <description>Altitude hold flight mode</description>
+      </entry>
+      <entry value="2" name="FLIGHT_MODE_POSCTL">
+        <description>Position control flight mode</description>
+      </entry>
+      <entry value="3" name="FLIGHT_MODE_STABILIZED">
+        <description>Stabilized mode</description>
+      </entry>
+      <entry value="4" name="FLIGHT_MODE_ACRO">
+        <description>Acro mode</description>
+      </entry>
+      <entry value="5" name="FLIGHT_MODE_RATTITIDE">
+        <description>Rattitude mode</description>
+      </entry>
+      <entry value="6" name="FLIGHT_MODE_AUTO_MISSION">
+        <description>Mission mode</description>
+      </entry>
+      <entry value="7" name="FLIGHT_MODE_AUTO_LOITER">
+        <description>Loiter mode</description>
+      </entry>
+      <entry value="8" name="FLIGHT_MODE_AUTO_RTL">
+        <description>Return to land mode</description>
+      </entry>
+      <entry value="9" name="FLIGHT_MODE_AUTO_RCRECOVER">
+        <description>RC recover mode</description>
+      </entry>
+      <entry value="10" name="FLIGHT_MODE_AUTO_RTGS">
+        <description>Return to ground station mode</description>
+      </entry>
+      <entry value="11" name="FLIGHT_MODE_AUTO_LANDENGFAIL">
+        <description>Land Engine Fail mode</description>
+      </entry>
+      <entry value="12" name="FLIGHT_MODE_AUTO_LANDGPSFAIL">
+        <description>Land GPS fail mode</description>
+      </entry>
+      <entry value="13" name="FLIGHT_MODE_AUTO_TAKEOFF">
+        <description>Takeoff mode</description>
+      </entry>
+      <entry value="14" name="FLIGHT_MODE_AUTO_LAND">
+        <description>Land mode</description>
+      </entry>
+      <entry value="15" name="FLIGHT_MODE_AUTO_FOLLOW_TARGET">
+        <description>Follow target/me mode</description>
+      </entry>
+      <entry value="16" name="FLIGHT_MODE_AUTO_PRECLAND">
+        <description>Precision landing mode</description>
+      </entry>
+      <entry value="17" name="FLIGHT_MODE_DESCEND">
+        <description>Descend mode</description>
+      </entry>
+      <entry value="18" name="FLIGHT_MODE_OFFBOARD">
+        <description>Offboard mode</description>
+      </entry>
+      <entry value="19" name="FLIGHT_MODE_TERMINATION">
+        <description>Termination mode</description>
+      </entry>
+      <entry value="20" name="FLIGHT_MODE_AUTO_TUNE">
+        <description>Auto tune mode</description>
+      </entry>
+      <entry value="21" name="FLIGHT_MODE_BRAKE">
+        <description>Brake mode</description>
+      </entry>
+      <entry value="22" name="FLIGHT_MODE_CIRCLE">
+        <description>Circle mode</description>
+      </entry>
+      <entry value="23" name="FLIGHT_MODE_DRIFT">
+        <description>Drift mode</description>
+      </entry>
+      <entry value="24" name="FLIGHT_MODE_GUIDED">
+        <description>Guided mode</description>
+      </entry>
+      <entry value="25" name="FLIGHT_MODE_SPORT">
+        <description>Sport mode</description>
+      </entry>
+      <entry value="26" name="FLIGHT_MODE_THROW">
+        <description>Throw mode</description>
+      </entry>
+      <entry value="27" name="FLIGHT_MODE_SUPER_SIMPLE">
+        <description>Simplified multicopter control mode</description>
+      </entry>
+      <entry value="28" name="FLIGHT_MODE_AVIOD_TRAFFIC">
+        <description>Avoid traffic mode</description>
+      </entry>
+      <entry value="29" name="FLIGHT_MODE_FBWA">
+        <description>Fly by wire (version A) mode</description>
+      </entry>
+      <entry value="30" name="FLIGHT_MODE_FBWB">
+        <description>Flz by wire (version B) mode</description>
+      </entry>
+      <entry value="31" name="FLIGHT_MODE_TRAINING">
+        <description>Position control flight mode</description>
+      </entry>
+      <entry value="32" name="FLIGHT_MODE_CRUISE">
+        <description>Cruise mode</description>
+      </entry>
+      <entry value="33" name="FLIGHT_MODE_HOLD">
+        <description>Position hold mode</description>
+      </entry>
+      <entry value="34" name="FLIGHT_MODE_STEERING">
+        <description>Steering mode</description>
+      </entry>
+      <entry value="35" name="FLIGHT_MODE_SMART_RTL">
+        <description>Smart return-to-land-mode</description>
+      </entry>
+    </enum>
+    <enum name="MAV_FAILSAFE_FLAG">
+      <description>These flags encode if a certain failsafe mode is triggered.</description>
+      <entry value="128" name="MAV_FAILSAFE_FLAG_RC_LOSS">
+        <description>0b10000000 RC loss failsafe active.</description>
+      </entry>
+      <entry value="64" name="MAV_FAILSAFE_FLAG_TELEM_LOSS">
+        <description>0b01000000 remote control input is enabled.</description>
+      </entry>
+      <entry value="32" name="MAV_FAILSAFE_FLAG_ENGINE">
+        <description>0b00100000 Engine failure failsafe active.</description>
+      </entry>
+      <entry value="16" name="MAV_FAILSAFE_FLAG_GPS">
+        <description>0b00010000 GPS failure failsafe active.</description>
+      </entry>
+      <entry value="8" name="MAV_FAILSAFE_FLAG_BATTERY">
+        <description>0b00001000 Battery failure failsafe active.</description>
+      </entry>
+      <entry value="4" name="MAV_FAILSAFE_FLAG_GEOFENCE">
+        <description>0b00000100 Geofence failsafe active.</description>
+      </entry>
+      <entry value="2" name="MAV_FAILSAFE_FLAG_RESERVED1">
+        <description>0b00000010 Reserved for future use.</description>
+      </entry>
+      <entry value="1" name="MAV_FAILSAFE_FLAG_RESERVED2">
+        <description>0b00000001 Reserved for future use.</description>
+      </entry>
+    </enum>
+    <enum name="MAV_FAILURE_FLAG">
+      <description>These flags encode if a certain failsafe mode is triggered.</description>
+      <entry value="1" name="MAV_FAILURE_FLAG_GPS">
+        <description>GPS failure.</description>
+      </entry>
+      <entry value="2" name="MAV_FAILURE_FLAG_AIRSPEED">
+        <description>Airspeed failure.</description>
+      </entry>
+      <entry value="4" name="MAV_FAILURE_FLAG_BAROMETER">
+        <description>Barometer failure.</description>
+      </entry>
+      <entry value="8" name="MAV_FAILURE_FLAG_ACCELEROMETER">
+        <description>Accelerometer failure.</description>
+      </entry>
+      <entry value="16" name="MAV_FAILURE_FLAG_GYROSCOPE">
+        <description>Gyroscope failure.</description>
+      </entry>
+      <entry value="32" name="MAV_FAILURE_FLAG_MAGNETOMETER">
+        <description>Magnetometer.</description>
+      </entry>
+      <entry value="64" name="MAV_FAILURE_FLAG_MISSION">
+        <description>Mission failure.</description>
+      </entry>
+      <entry value="128" name="MAV_FAILURE_FLAG_ESTIMATOR">
+        <description>Estimator failure.</description>
+      </entry>
+      <entry value="256" name="MAV_FAILURE_FLAG_LIDAR">
+        <description>Lidar failure.</description>
+      </entry>
+      <entry value="512" name="MAV_FAILURE_FLAG_OFFBOARD_LINK">
+        <description>Offboard link failure.</description>
+      </entry>
+      <entry value="1024" name="MAV_FAILURE_FLAG_RESERVED1">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="2048" name="MAV_FAILURE_FLAG_RESERVED2">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="4096" name="MAV_FAILURE_FLAG_RESERVED3">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="8192" name="MAV_FAILURE_FLAG_RESERVED4">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="16384" name="MAV_FAILURE_FLAG_RESERVED5">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="32768" name="MAV_FAILURE_FLAG_RESERVED6">
+        <description>Reserved for future use.</description>
+      </entry>
+    </enum>
     <!-- WARNING: MAV_ACTION Enum is no longer supported - has been removed. Please use MAV_CMD -->
     <enum name="MAV_MODE_FLAG">
       <description>These flags encode the MAV mode.</description>
@@ -4123,6 +4312,37 @@
       <field type="uint8_t" name="failsafe">failsafe (each bit represents a failsafe where 0=ok, 1=failsafe active (bit0:RC, bit1:batt, bit2:GPS, bit3:GCS, bit4:fence)</field>
       <field type="uint8_t" name="wp_num">current waypoint number</field>
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
+    </message>
+    <message id="235" name="HIGH_LATENCY2">
+      <description>(DRAFT) Message appropriate for high latency connections like Iridium (version 2)</description>
+      <field type="uint32_t" name="timestamp" units="ms">Timestamp (milliseconds since boot or Unix epoch)</field>
+      <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
+      <field type="uint8_t" name="autopilot" enum="MAV_AUTOPILOT">Autopilot type / class. defined in MAV_AUTOPILOT ENUM</field>
+      <field type="uint8_t" name="flight_mode" enum="FLIGHT_MODE">Flight Mode of the vehicle as defined in the FLIGHT_MODE ENUM</field>
+      <field type="int32_t" name="latitude" units="degE7">Latitude, expressed as degrees * 1E7</field>
+      <field type="int32_t" name="longitude" units="degE7">Longitude, expressed as degrees * 1E7</field>
+      <field type="int16_t" name="altitude" units="m">Altitude above mean sea level</field>
+      <field type="int16_t" name="target_altitude" units="m">Altitude setpoint</field>
+      <field type="uint8_t" name="heading" units="deg / 2">Heading (degrees / 2)</field>
+      <field type="uint8_t" name="target_heading" units="deg / 2">Heading setpoint (degrees / 2)</field>
+      <field type="uint16_t" name="target_distance" units="m">Distance to target waypoint or position (meters / 10)</field>
+      <field type="uint8_t" name="throttle" units="%">Throttle (percentage)</field>
+      <field type="uint8_t" name="airspeed" units="m/s * 5">Airspeed (m/s * 5)</field>
+      <field type="uint8_t" name="airspeed_sp" units="m/s * 5">Airspeed setpoint (m/s * 5)</field>
+      <field type="uint8_t" name="groundspeed" units="m/s * 5">Groundspeed (m/s * 5)</field>
+      <field type="uint8_t" name="windspeed" units="m/s * 5">Windspeed (m/s * 5)</field>
+      <field type="uint8_t" name="wind_heading" units="deg / 2">Wind heading (deg / 2)</field>
+      <field type="uint8_t" name="eph" units="m * 10">Maximum error horizontal position since last message (m * 10)</field>
+      <field type="uint8_t" name="epv" units="m * 10">Maximum error vertical position since last message (m * 10)</field>
+      <field type="int8_t" name="temperature_air" units="degC">Air temperature (degrees C) from airspeed sensor</field>
+      <field type="int8_t" name="climb_rate" units="m/s * 10">Maximum climb rate magnitude since last message (m/s * 10)</field>
+      <field type="int8_t" name="battery" units="%">Battery (percentage, -1 for DNU)</field>
+      <field type="uint16_t" name="wp_num">Current waypoint number</field>
+      <field type="uint16_t" name="failure_flags" enum="MAV_FAILURE_FLAG">Indicates failures as defined in MAV_FAILURE_FLAG ENUM. </field>
+      <field type="uint8_t" name="failsafe" enum="MAV_FAILSAFE_FLAG">Indicates if a failsafe mode is triggered, defined in MAV_FAILSAFE_FLAG ENUM</field>
+      <field type="int8_t" name="custom0">Field for custom payload.</field>
+      <field type="int8_t" name="custom1">Field for custom payload.</field>
+      <field type="int8_t" name="custom2">Field for custom payload.</field>
     </message>
     <message id="241" name="VIBRATION">
       <description>Vibration levels and accelerometer clipping</description>


dagar:
@WickedShell as I recall you had some good input the first time. Would you be interested in reviewing this once it's ready? I haven't updated the comments or units yet (some scaling is different).

philipoe:
We discussed offline that many variables (airspeed, groundspeed, ...) need scaling (that you also mention in your second comment). I think these scaling values should be included in the comments/description of each message field.

TSC21:
Duplicating PR is not a good move but ok. Can I get your feedback first on #813?

dagar:
It's not duplicating, it was done independently. That's why I pushed this to a branch off the main repo as soon as I saw you were working on the same thing.

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
       <field type="uint8_t" name="wp_num">current waypoint number</field>
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
+    <message id="235" name="HIGH_LATENCY2">
+      <description>(DRAFT) Message appropriate for high latency connections like Iridium (version 2)</description>
+      <field type="uint32_t" name="timestamp" units="ms">Timestamp (milliseconds since boot or Unix epoch)</field>
+      <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
```
peterbarker: What's the comment getting at here?

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="int8_t" name="pitch" units="deg">Pitch (optional)</field>
+      <field type="uint16_t" name="heading" units="deg">Heading (degrees / 2)</field>
+      <field type="uint16_t" name="heading_sp" units="deg">(REVIEW) heading (degrees / 2)</field>
+      <field type="uint8_t" name="throttle" units="%">Throttle (percentage)</field>
+      <field type="uint8_t" name="airspeed" units="m/s">Airspeed (m/s * 5)</field>
+      <field type="uint8_t" name="airspeed_sp" units="m/s">Airspeed setpoint (m/s * 5)</field>
+      <field type="uint8_t" name="groundspeed" units="m/s">Groundspeed (m/s * 5)</field>
+      <field type="uint8_t" name="windspeed" units="m/s">Windspeed (m/s * 10)</field>
+      <field type="uint8_t" name="wind_heading" units="m/s">Wind heading (deg / 2)</field>
+      <field type="uint8_t" name="gps_info">Bits 0-2: GPS fix Bits 3-7: Number of observed satellites</field>
```
peterbarker: This isn't a great use of space IMO.  Some other quality measure of the vehicle position would be far preferable.

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="int8_t" name="temperature_air" units="degC">Air temperature (degrees C) from airspeed sensor</field>
+      <field type="int8_t" name="climb_rate" units="m/s">Maximum climb rate magnitude since last message (m/s * 10)</field>
+      <field type="uint8_t" name="battery" units="%">Battery (percentage)</field>
+      <field type="uint16_t" name="wp_num">Current waypoint number</field>
+      <field type="uint16_t" name="wp_distance" units="m">(REVIEW) Distance to target (meters)</field>
```
peterbarker: Yep, too small.

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="uint8_t" name="nav_state">Navigation statae of the vehicle</field>
```
peterbarker: What's the intent?

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="uint16_t" name="failure_flags">(REVIEW)</field>
```
peterbarker: What's the intent?

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
+      <field type="uint8_t" name="autopilot" enum="MAV_AUTOPILOT">Autopilot type / class. defined in MAV_AUTOPILOT ENUM</field>
+      <field type="int32_t" name="latitude" units="degE7">Latitude, expressed as degrees * 1E7</field>
+      <field type="int32_t" name="longitude" units="degE7">Longitude, expressed as degrees * 1E7</field>
+      <field type="int16_t" name="altitude" units="m">Altitude above mean sea level</field>
+      <field type="int16_t" name="altitude_sp" units="m">Altitude setpoint</field>
+      <field type="int8_t" name="roll" units="deg">Roll (optional)</field>
+      <field type="int8_t" name="pitch" units="deg">Pitch (optional)</field>
+      <field type="uint16_t" name="heading" units="deg">Heading (degrees / 2)</field>
+      <field type="uint16_t" name="heading_sp" units="deg">(REVIEW) heading (degrees / 2)</field>
```
peterbarker: Wrong type?

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="uint32_t" name="timestamp" units="ms">Timestamp (milliseconds since boot or Unix epoch)</field>
+      <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
+      <field type="uint8_t" name="autopilot" enum="MAV_AUTOPILOT">Autopilot type / class. defined in MAV_AUTOPILOT ENUM</field>
+      <field type="int32_t" name="latitude" units="degE7">Latitude, expressed as degrees * 1E7</field>
+      <field type="int32_t" name="longitude" units="degE7">Longitude, expressed as degrees * 1E7</field>
+      <field type="int16_t" name="altitude" units="m">Altitude above mean sea level</field>
+      <field type="int16_t" name="altitude_sp" units="m">Altitude setpoint</field>
+      <field type="int8_t" name="roll" units="deg">Roll (optional)</field>
+      <field type="int8_t" name="pitch" units="deg">Pitch (optional)</field>
+      <field type="uint16_t" name="heading" units="deg">Heading (degrees / 2)</field>
```
peterbarker: Wrong type?

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
+    <message id="235" name="HIGH_LATENCY2">
+      <description>(DRAFT) Message appropriate for high latency connections like Iridium (version 2)</description>
+      <field type="uint32_t" name="timestamp" units="ms">Timestamp (milliseconds since boot or Unix epoch)</field>
+      <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
+      <field type="uint8_t" name="autopilot" enum="MAV_AUTOPILOT">Autopilot type / class. defined in MAV_AUTOPILOT ENUM</field>
+      <field type="int32_t" name="latitude" units="degE7">Latitude, expressed as degrees * 1E7</field>
+      <field type="int32_t" name="longitude" units="degE7">Longitude, expressed as degrees * 1E7</field>
+      <field type="int16_t" name="altitude" units="m">Altitude above mean sea level</field>
```
peterbarker: This type isn't large enough; 32km won't cut it for balloons and the like.

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
       <field type="uint8_t" name="wp_num">current waypoint number</field>
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
+    <message id="235" name="HIGH_LATENCY2">
```
peterbarker: I can't see a compelling reason to use a mavlink1 ID for this message.

We don't have many of those left, and we should avoid using them where possible.  The mavlink1 message ids may be used in situations where intermediaries don't do mavlink2 and still try to parse messages - one of the common esp8266 firmwares, for example.

dagar: The motivation for this work is using sending this over IridiumSBD (Rockblock) which has 50 byte billing increments. The additional size of the mavlink v2 header seemed problematic.

message_definitions/v1.0/common.xml
dagar: This message will also serve as a HEARTBEAT. This is needed to construct the "vehicle" in the GCS.

message_definitions/v1.0/common.xml
dagar: Are you suggesting dropping gps_info entirely and adding estimator eph/epv instead?

message_definitions/v1.0/common.xml
dagar: This is another one needed as a HEARTBEAT replacement that we'd like to use in place of base_mode + custom_mode (5 bytes).

message_definitions/v1.0/common.xml
dagar: TBD, but so far thinking something like... (these will be included in the spec)
0. Gps failure
1. Airspeed failure
2. Barometer failure
3. Accelerometer failure
4. Gyroscope failure
5. Magnetometer failure
6. Mission failure
7. Estimator fusion numerical failure
8. Estimator innovation error
9. Engine failure
10.-15. Reserved for other failure cases
--

message_definitions/v1.0/common.xml
TSC21: we will be increasing the msg payload, and we can't actually do that given the size limit in Iridium SBD msgs.

message_definitions/v1.0/common.xml
TSC21: why? https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2921

message_definitions/v1.0/common.xml
TSC21: https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2921

TSC21:
@dagar based on @peterbarker, are we able to go forward with this?

dagar:
We need to settle on Mavlink v1 vs v2, but yes let's keep iterating.

TODO
 - carefully review sizing and packing of each field
 - build up the the failure_flags 
   - this is actually quite important as it can eliminate the need for any STATUSTEXT messages
 - replace GPS fields with horizontal and vertical position error?

TSC21:
> We need to settle on Mavlink v1 vs v2

In this situation, I think it really depends if we would want this to be a signed msg or not. Giving the type of network that needs to be established in order to retrieve and send Iridium msgs to/from the ground station, I think this would actually make sense.

> - replace GPS fields with horizontal and vertical position error?

Agreed as I don't see the need of the number of satellites available on a high latency link.

> - build up the the failure_flags

I would add two more: Lidar (or distance-to-ground measurement sensor) failure, as a direct cause for AGL estimation problems, and the Offboard link timeout it failure, which for systems with a OBC running onboard feeding and retrieving critical data, may represent also a problem. Will have to be coupled with the work on https://github.com/PX4/Firmware/pull/7985.

peterbarker:
On Sun, 24 Dec 2017, Daniel Agar wrote:

>      +  this is actually quite important as it can eliminate the need for
>         any STATUSTEXT messages

Random thought: if you put the equivalent of a statustext message at the 
end of the packet (assuming something of reasonable length fits), then you 
can rely on mavlink2's zero-bytes-truncation to save sending the 
statustext bit unless necessary.  Not sure if that helps given underlying 
packetisation etc.

DonLakeFlyer:
QGC needs the following to run:
* MAV_TYPE
* MAV_AUTOPILOT
* Some form of flight mode (doesn't need to be base_mode/custom_mode which takes too many bits to represent a few modes)

message_definitions/v1.0/common.xml
DavidIngraham: Just a note: Rockblock is the only Iridium reseller that I know of that bills in 50 byte increments like this. Most other resellers have a minimum price per message that covers the first 20-30 bytes, and a flat rate per byte for additional data in the message.

However, Rockblock's equipment is very accessible and well documented, so I think it makes sense to stick with the 50 byte target for the base message. Perhaps message extensions could be used for additional data when weird things (failures) are going on.

dagar: Thanks for the input David.

dagar:
Thanks Don. We've included type (uint8_t MAV_TYPE), autopilot (uint8_t MAV_AUTOPILOT), and a new uint8_t nav_state to replace base_mode/custom_mode.

@peterbarker that's potentially a good idea with Mavlink v2, but I'm concerned about the additional header size. I counted an additional 18 bytes we'd need for v2 in the header. Can you confirm this, or am I missing something about mavlink v2?

Mavlink v1.0

``` C
MAVPACKED(
typedef struct __mavlink_message {
	uint16_t checksum; ///< sent at end of packet
	uint8_t magic;   ///< protocol magic marker
	uint8_t len;     ///< Length of payload
	uint8_t seq;     ///< Sequence of packet
	uint8_t sysid;   ///< ID of message sender system/aircraft
	uint8_t compid;  ///< ID of the message sender component
	uint8_t msgid;   ///< ID of message in payload
	uint64_t payload64[(MAVLINK_MAX_PAYLOAD_LEN+MAVLINK_NUM_CHECKSUM_BYTES+7)/8];
}) mavlink_message_t;
```

Mavlink v2.0

``` C
MAVPACKED(
typedef struct __mavlink_message {
	uint16_t checksum;      ///< sent at end of packet
	uint8_t magic;          ///< protocol magic marker
	uint8_t len;            ///< Length of payload
	uint8_t incompat_flags; ///< flags that must be understood
	uint8_t compat_flags;   ///< flags that can be ignored if not understood
	uint8_t seq;            ///< Sequence of packet
	uint8_t sysid;          ///< ID of message sender system/aircraft
	uint8_t compid;         ///< ID of the message sender component
	uint32_t msgid:24;      ///< ID of message in payload
	uint64_t payload64[(MAVLINK_MAX_PAYLOAD_LEN+MAVLINK_NUM_CHECKSUM_BYTES+7)/8];
	uint8_t ck[2];          ///< incoming checksum bytes
	uint8_t signature[MAVLINK_SIGNATURE_BLOCK_LEN];
}) mavlink_message_t;
```

dogmaphobic:
>In this situation, I think it really depends if we would want this to be a signed msg or not.

Not only signing is not complete, you don't need it. The link is encrypted (HTTPS) and authentication is done using this encrypted link. If it's good enough to access your bank, it should be good enough for this.

>We need to settle on Mavlink v1 vs v2, but yes let's keep iterating.

@DonLakeFlyer would know this better. During fight, V1 should be just fine but during boot up and initialization, it will depend on what type of mission messages are used. Some are V2 only. Note that this is a hybrid link. There are usually two sub-links where Iridium is a failover. When the primary sub-link is first established (LTE/4G/etc.), it will determine if we're to use V1 or V2, which will remain in effect for the duration, even if the sub-link changes from primary to Iridium.

dagar:
Thanks Gus, and yes I agree about signing.

I think for the most part everyone is on the same page here, but for clarity when I say Mavlink v1 or v2 I'm really only talking about keeping the total size (msg + mavlink header) within 50 bytes, and then the v1 message id.

Beyond that the actual use of Mavlink v1 or v2 is an implementation detail.

message_definitions/v1.0/common.xml
DavidIngraham: Could we possibly condense the MAV_TYPE and MAV_AUTOPILOT into one byte?  5 bits for MAV_TYPE and 3 bits for MAV_AUTOPILOT would cover all current users of MAVLINK from what I can tell. 

It just feels a bit wasteful to have 2 bytes used for something that doesn't change at all during a flight.

Or perhaps the standard heartbeat message could be used to initially handshake with the GCS, then the GCS would have to remember the mav/autopilot type that corresponds to each SYSID.

message_definitions/v1.0/common.xml
DavidIngraham: @TSC21 Heading/2 only needs 8 bits. This should either be 16 bits for heading, or 8 bits for heading/2.

I think 2 degrees is plenty of precision in this case

message_definitions/v1.0/common.xml
DavidIngraham: I believe it would be best to scale this value rather than adding more precision. SBD Messages take at least 10 seconds to send and most vehicles fly significantly faster than 0.1m/s, so one meter accuracy isn't really relevant. A vehicle the cruises at 5m/s would cover 50 m in the period of one message.

If we scaled it by 10, we would be able to report waypoint distances up to 655 km (406 miles).

pfg:
For what its worth...


By the time your into these messages the presumption would be your in trouble of some sort so supply the most relevant data. By far the most important is the location of the vehicle - everything else is secondary. If you really want to send anything else then have two versions of the message (just use two id's) and keep the location in both and then put half the optional stuff in one and the other half in the other and alternate between the two.

________________________________

peterbarker:
On Tue, 26 Dec 2017, Daniel Agar wrote:

> @peterbarker that's potentially a good idea with Mavlink v2, but I'm concerned about the additional header size. I counted an additional 18 bytes we'd need
> for v2 in the header. Can you confirm this, or am I missing something about mavlink v2?

If signing isn't used those bytes will be zero, and thus truncated.

The overhead is certainly there, 'though.

> 
> Mavlink v1.0
> 
> MAVPACKED(
> typedef struct __mavlink_message {
> 	uint16_t checksum; ///< sent at end of packet
> 	uint8_t magic;   ///< protocol magic marker
> 	uint8_t len;     ///< Length of payload
> 	uint8_t seq;     ///< Sequence of packet
> 	uint8_t sysid;   ///< ID of message sender system/aircraft
> 	uint8_t compid;  ///< ID of the message sender component
> 	uint8_t msgid;   ///< ID of message in payload
> 	uint64_t payload64[(MAVLINK_MAX_PAYLOAD_LEN+MAVLINK_NUM_CHECKSUM_BYTES+7)/8];
> }) mavlink_message_t;
> 
> Mavlink v2.0
> 
> MAVPACKED(
> typedef struct __mavlink_message {
> 	uint16_t checksum;      ///< sent at end of packet
> 	uint8_t magic;          ///< protocol magic marker
> 	uint8_t len;            ///< Length of payload
> 	uint8_t incompat_flags; ///< flags that must be understood
> 	uint8_t compat_flags;   ///< flags that can be ignored if not understood
> 	uint8_t seq;            ///< Sequence of packet
> 	uint8_t sysid;          ///< ID of message sender system/aircraft
> 	uint8_t compid;         ///< ID of the message sender component
> 	uint32_t msgid:24;      ///< ID of message in payload
> 	uint64_t payload64[(MAVLINK_MAX_PAYLOAD_LEN+MAVLINK_NUM_CHECKSUM_BYTES+7)/8];
> 	uint8_t ck[2];          ///< incoming checksum bytes
> 	uint8_t signature[MAVLINK_SIGNATURE_BLOCK_LEN];
> }) mavlink_message_t;

I *believe* the extra overhead is 4 bytes unless you want signing (due to 
assumed zero-trimming).  Depending on how you arrange your message (and 
the content) it is possible you might make those 4 bytes back up by 
utilising the zero-trimming feature.

_mav_finalize_message_chan_send seems to confirm that.

Yours,
-- 
Peter Barker                          |   Programmer,Sysadmin,Geek.
pbarker@barker.dropbear.id.au	      |   You need a bigger hammer.
:: It's a hack! Expect underscores! - Nigel Williams

DonLakeFlyer:
Thinking a little more on MAV_AUTOPILOT and MAV_TYPE. It's certainly wasteful to send this in every packet. My new thinking is that the high latency case is special enough that the firmware type and vehicle type should be specified as part of the link metadata on the groundstation side. There is no need to support the fancy auto-detection stuff that QGC does for these links. This won't get used for connecting to random vehicles. So I'd say remove MAV_AUTOPILOT and MAV_TYPE from the message.

message_definitions/v1.0/common.xml
```diff
@@ -4068,6 +4068,34 @@
...
+      <field type="uint8_t" name="throttle" units="%">Throttle (percentage)</field>
+      <field type="uint8_t" name="airspeed" units="m/s">Airspeed (m/s * 5)</field>
+      <field type="uint8_t" name="airspeed_sp" units="m/s">Airspeed setpoint (m/s * 5)</field>
+      <field type="uint8_t" name="groundspeed" units="m/s">Groundspeed (m/s * 5)</field>
+      <field type="uint8_t" name="windspeed" units="m/s">Windspeed (m/s * 10)</field>
+      <field type="uint8_t" name="wind_heading" units="m/s">Wind heading (deg / 2)</field>
+      <field type="uint8_t" name="gps_info">Bits 0-2: GPS fix Bits 3-7: Number of observed satellites</field>
+      <field type="int8_t" name="temperature_air" units="degC">Air temperature (degrees C) from airspeed sensor</field>
+      <field type="int8_t" name="climb_rate" units="m/s">Maximum climb rate magnitude since last message (m/s * 10)</field>
+      <field type="uint8_t" name="battery" units="%">Battery (percentage)</field>
```
WickedShell: This should be `int8_t` with a DNU of -1, or an explicitly listed do not use value here.

message_definitions/v1.0/common.xml
WickedShell: I don't think this is a great use of data either for two reasons:
1. It doesn't map out a lot of useful cases that we already have with `MAV_SYS_STATUS_SENSOR` that we would still want to report to the GCS
2. You are spending payload space communicating a normally okay state to the GCS/monitoring. I submit that it is far better in a failure case to broadcast an extra message that is sent in it's own payload as soon as the error condition asserts to enable a GCS/operator to respond ASAP to the failure as well as allowing it to be a much more specific enumeration of failure reason.

message_definitions/v1.0/common.xml
WickedShell: state is spelled wrong (as well as not being at all clear how to map/interpret the value)

message_definitions/v1.0/common.xml
WickedShell: I would also think it's not unreasonable to also state that the WP_DISTANCE just clamps to the maximum value of a `uint16_t` when you exceed it. From a path planning/validation perspective this should still be a more then sufficient vector to check that the vehicle is proceeding as you expect it to be. However we do need to document thats the expectation if we choose to implement it that way.

message_definitions/v1.0/common.xml
acfloria: It would be an autopilot dependent state. For example for PX4: [NAVIGATION_STATE](https://github.com/PX4/Firmware/blob/master/msg/vehicle_status.msg#L15)

message_definitions/v1.0/common.xml
acfloria: I agree, that eph/epv would give more information. But we were thinking of the high_latency message as more of a status message which is displayed in QGC and for this purpose the GPS fix and the number of satellites is useful.
But I would vote that we transmit the eph/epv instead of the roll/pitch angles.

message_definitions/v1.0/common.xml
acfloria: What we intended here is a condensed version of `MAV_SYS_STATUS_SENSOR` as we do not want to send the full `SYS_STATUS` message. Sending it each time failure occurs can easily block the link assuming multiple failures occur in a short time.

There is not much the user can do in case of a sensor failure, taking into account the limited bandwidth and the 20 to 40 seconds delay. This failure flags should give the user the information why the plane possibly crashed.

peterbarker:
On Wed, 3 Jan 2018, Florian Achermann wrote:

> +      <field type="uint16_t" name="heading" units="deg">Heading (degrees /
>  2)</field>
> +      <field type="uint16_t" name="heading_sp" units="deg">(REVIEW) headin
> g (degrees / 2)</field>
> 
> Also here we need only a uint8_t. But do we really need it as we have the
> wp_num from which we can infer the heading setpoint.

Desired vehicle heading may not be directly towards the waypoint - 
crosstrack error.

> +      <field type="uint8_t" name="nav_state">Navigation statae of the vehi
> cle</field>
> 
> It would be an autopilot dependent state. For example for PX4:
> NAVIGATION_STATE

If a GCS is displaying this - shouldn't we have an enumeration which can 
be generally used by an autopilot?

> mber of observed satellites</field>
> 
> I agree, that eph/epv would give more information. But we were thinking of
> the high_latency message as more of a status message which is displayed in
> QGC and for this purpose the GPS fix and the number of satellites is
> useful.

The number of satellites is not *particularly* useful - it's just 
traditional :-)  It's really just a feel-good thing (which... I guess... 
could be considered "useful" in some respects)

> But I would vote that we transmit the eph/epv instead of the roll/pitch
> angles.

I'm sure we could dig up other information more useful than a satellite 
count, too.

> +      <field type="uint16_t" name="failure_flags">(REVIEW)</field>
> 
> What we intended here is a condensed version of MAV_SYS_STATUS_SENSOR as we
> do not want to send the full SYS_STATUS message. Sending it each time
> failure occurs can easily block the link assuming multiple failures occur
> in a short time.

hysteresis and a minimum-send-interval would seem to make sense here.

> There is not much the user can do in case of a sensor failure, taking into
> account the limited bandwidth and the 20 to 40 seconds delay. This failure
> flags should give the user the information why the plane possibly crashed.

If you're waiting for the next HIGH_LATENCY2 message to be sent out before 
telling the user of a sensor failure you may never get that opportunity!

message_definitions/v1.0/common.xml
```diff
@@ -4042,7 +4042,7 @@
       <field type="uint8_t[180]" name="data">RTCM message (may be fragmented)</field>
     </message>
     <message id="234" name="HIGH_LATENCY">
-      <description>Message appropriate for high latency connections like Iridium</description>
+      <description>(DEPRECATED: use HIGH_LATENCY2) Message appropriate for high latency connections like Iridium</description>
       <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitfield, see MAV_MODE_FLAG ENUM in mavlink/include/mavlink_types.h</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
```
DonLakeFlyer: What about vtol_state from EXTENDED_SYS_STATE? QGC uses that as well for VTOL transition.

vooon:
If that message needed for iridium, why keep mavlink framing?
Only what you need: start byte - msgid, and maybe system id.
I'm pretty sure, that this 50 bytes already should be protected by crc's by iridium modem, or i'm wrong?

dogmaphobic:
>If that message needed for iridium, why keep mavlink framing?
That's an interesting question. Specially given the non standard nature of the link it would come through. It's not like it will be mixed in with "regular" MAVLink traffic.

acfloria:
> I'm pretty sure, that this 50 bytes already should be protected by crc's by iridium modem, or i'm wrong?

You can send up to 340 bytes and receive up to 270 bytes in a single message. The pricing though is in 50 bytes steps.

TSC21:
Hey guys, is there a way we can move this discussion forward and reach out a consensus for the definite structure of the message? Otherwise we will keep iterating and this will never be merged.

peterbarker:
What sort of interval would this interval usually be sent at?

Trying to reduce resource consumption on ArduPilot's message-interval implementation, and wedging in 100Hz messages in with something which might be sent once and hour is going to be "interesting".

dagar:
It varies quite a bit, but I'd say once every 10-60 seconds.

acfloria:
From our experience sending a single message with RockBlock and a good connection requires around 20 seconds. However, if the connection is bad it can take up to one minute or even more.

peterbarker:
On Thu, 25 Jan 2018, Florian Achermann wrote:

> From our experience sending a single message with RockBlock and a good
> connection requires around 20 seconds. However, if the connection is bad it
> can take up to one minute or even more.

More the maximum interval between messages I'm interested in.  At least a 
minute, I guess :-)

acfloria:
This is the quote from the RockBlock webpage:

> Testing shows that it generally takes around 20 seconds from power-up to successful transmission, with a perfect view of the sky. With a very restricted view, it may take several minutes.

peterbarker:
On Thu, 25 Jan 2018, Florian Achermann wrote:

> This is the quote from the RockBlock webpage:
>
>       Testing shows that it generally takes around 20 seconds from
>       power-up to successful transmission, with a perfect view of the
>       sky. With a very restricted view, it may take several minutes.

Hmmm.

I might need to read up on this a lot more.

What happens if we send a second HIGH_LATENCY2 message over the serial 
link to the RockBlock (assuming here... ;-) ) while it is still sending 
one?

Is it discarded?  I hope it's not queued :-)

message_definitions/v1.0/common.xml
dagar: This seems like one of the only messages that actually has a compelling reason to be Mavlink v1 for the smaller header.

dagar:
Thanks for the input everyone. I propose we merge this as a DRAFT, get some hands on time actually testing it with a real modem, firmware, and GCS support, then get everyone to do one last final approval to finalize.

WickedShell:
MAV_MODE's don't map out all the modes ArduPilot supports, the failure flags still represents a terrible compromise on failure reporting compared to the existing status flags (You can't tell the GCS what mag failed which is critical for debugging/post flight analysis especially if the vehicle goes down somewhere remote). Failsafe flags also don't map out to everything ArduPilot supports either.

At a very high level enums that are flags should never have reserved values in them, as it just makes it harder to assign them meaning in the future, as a future user has to decide if he can take a reserved one.

dagar:
@WickedShell let's start an issue to keep track of these things. https://github.com/mavlink/mavlink/issues/831

Could you help us fill out the missing ardupilot modes? What would you propose specifically for the failure flags?

amilcarlucas:
@dagar The units you created are a bit ..... creative. #826 tries to integrate them in the Travis CI tests and brings in the ArduPilot flight modes enums that @WickedShell was talking about.
Please in the future use the display="bitmask" for bitmask enums, it will help in the GCS development.

PR #818, opened 2018-01-03
Remove duplicate command to MAV_CMD_DO_SET_ROI

DonLakeFlyer:
MAV_CMD_NAV_ROI is an exact duplicate of MAV_CMD_DO_SET_ROI. PX4 supports both it seems and ArduPilot only supports MAV_CMD_DO_SET_ROI. This pull removes the duplicate MAV_CMD_NAV_ROI which just ends up confusing things at least from a ground station standpoint.


DonLakeFlyer:
Yeah, oops good point. I'll close this and put in a new pull.

PR #825, opened 2018-01-16
precision land: add description for precision land mode

ndepal:
This documents the usage of `param2` of `MAV_CMD_NAV_LAND` for precision landing.


ndepal:
@dagar could you review?

magicrub:
This sounds nice in theory but the insta-merge is not appreciated without discussion. There's no explanation how to use it.

Is this for plane?
VTOL/copter-only?
What happens when you use =2 but don't have any additional sensor hardware?
Why would we ever not want precision when precision hardware/sensors are available?

ndepal:
https://docs.px4.io/en/advanced_features/precland.html

PR #826, opened 2018-01-18
More unit fixes

amilcarlucas:
Fixes and adds units attributes
Adds display="bitmask" attribute in some more bitmask fields

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4145,7 +4145,7 @@
     <message id="234" name="HIGH_LATENCY">
       <description>Message appropriate for high latency connections like Iridium</description>
       <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitfield, see MAV_MODE_FLAG ENUM in mavlink/include/mavlink_types.h</field>
-      <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags.</field>
+      <field type="uint32_t" name="custom_mode" display="bitmask">A bitfield for use for autopilot-specific flags.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
       <field type="int16_t" name="roll" units="cdeg">roll (centidegrees)</field>
       <field type="int16_t" name="pitch" units="cdeg">pitch (centidegrees)</field>
@@ -4170,7 +4170,7 @@
       <field type="uint16_t" name="wp_distance" units="m">distance to target (meters)</field>
     </message>
     <message id="235" name="HIGH_LATENCY2">
-      <description>(DRAFT) Message appropriate for high latency connections like Iridium (version 2)</description>
+      <description>WIP: Message appropriate for high latency connections like Iridium (version 2)</description>
       <field type="uint32_t" name="timestamp" units="ms">Timestamp (milliseconds since boot or Unix epoch)</field>
       <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
       <field type="uint8_t" name="autopilot" enum="MAV_AUTOPILOT">Autopilot type / class. defined in MAV_AUTOPILOT ENUM</field>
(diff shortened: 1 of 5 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -3348,7 +3348,7 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID, as defined by MAV_CMD enum.</field>
       <field type="uint8_t" name="result" enum="MAV_RESULT">See MAV_RESULT enum</field>
       <extensions/>
-      <field type="uint8_t" name="progress">WIP: Also used as result_param1, it can be set with a enum containing the errors reasons of why the command was denied or the progress percentage or 255 if unknown the progress when result is MAV_RESULT_IN_PROGRESS.</field>
+      <field type="uint8_t" name="progress" units="%">WIP: Also used as result_param1, it can be set with a enum containing the errors reasons of why the command was denied or the progress percentage or 255 if unknown the progress when result is MAV_RESULT_IN_PROGRESS.</field>
```
WickedShell: Not sure this is a valid tag. The % only holds true in one of three different use cases. In that case I think omitting the unit tag is a better approach.

message_definitions/v1.0/common.xml
```diff
@@ -4051,11 +4051,11 @@
...
       <field type="uint64_t" name="time_usec" units="us">Timestamp (micros since boot or Unix epoch)</field>
       <field type="uint16_t" name="flags" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Integer bitmask indicating which EKF outputs are valid. See definition for ESTIMATOR_STATUS_FLAGS.</field>
-      <field type="float" name="vel_ratio">Velocity innovation test ratio</field>
-      <field type="float" name="pos_horiz_ratio">Horizontal position innovation test ratio</field>
-      <field type="float" name="pos_vert_ratio">Vertical position innovation test ratio</field>
-      <field type="float" name="mag_ratio">Magnetometer innovation test ratio</field>
-      <field type="float" name="hagl_ratio">Height above terrain innovation test ratio</field>
+      <field type="float" name="vel_ratio" units="Hz">Velocity innovation test ratio</field>
```
WickedShell: Estimator status message fields are Hz? The description of ratio and Hz are not things that I see how they go together.

message_definitions/v1.0/common.xml
```diff
@@ -2881,16 +2881,16 @@
       <field type="uint32_t" name="h_acc" units="mm">Position uncertainty in meters * 1000 (positive for up).</field>
       <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty in meters * 1000 (positive for up).</field>
       <field type="uint32_t" name="vel_acc" units="mm">Speed uncertainty in meters * 1000 (positive for up).</field>
-      <field type="uint32_t" name="hdg_acc" units="mm">Heading / track uncertainty in degrees * 1e5.</field>
+      <field type="uint32_t" name="hdg_acc">Heading / track uncertainty in degrees * 1e5.</field>
```
WickedShell: I agree that mm is wrong, but why not add the correct unit?

amilcarlucas: There is no such SI unit (1E5) https://www.bipm.org/en/publications/si-brochure/chapter3.html

message_definitions/v1.0/common.xml
amilcarlucas: Good point, I missed that. Will correct it.

message_definitions/v1.0/common.xml
amilcarlucas: The vel_ratio is already documented. I assume that the ratio information has to do with the update rate of some fields.

amilcarlucas:
Thanks for reviewing @WickedShell . I addressed your issues now.

message_definitions/v1.0/common.xml
WickedShell: There is also no such unit as degE7. All you need here is a degE5. Don't let the limitation of scaling stop you from having correct documentation.

amilcarlucas:
This one needs to be merged first, so that the new CI tests introduced in mavlink/pymavlink#5 pass

amilcarlucas:
The Travis tests only fail because it timedout while downloading python 3.5

amilcarlucas:
All the new CI tests introduced by the latest pymavlink pass

amilcarlucas:
@LorenzMeier The last master PR merge you did ( [HIGH_LATENCY2 message](https://github.com/mavlink/mavlink/pull/814) ) caused this PR to fail the CI tests.
I had to fix it, but one of the the reasons for this PR is to make the CI tests more inteligent, so that the other developers can fix the issues themselves.
Please merge this in before it gets more conflicts again.

dagar:
I'm not familiar with the generate code, but this PR looks good to me.

amilcarlucas:
I had the solve the conflicts again :(

amilcarlucas:
Two CI tests pass, can someone restart the third ?

dagar:
Restarted

amilcarlucas:
@dagar Thanks, the CI tests pass again.

amilcarlucas:
@dagar Thanks, for merging but it does require mavlink/pymavlink#7.

amilcarlucas:
#843 is the permanent solution

PR #833, opened 2018-02-05
HIGH_LATENCY2 revision

acfloria:
- Use the 2 bytes custom_mode instead of a custom defined flight mode.

- Merge `failure_flags` and `failsafe` into one 2 byte bitfield.

- Change the naming of the `failure_flags` so that it is closer to `MAV_SYS_STATUS_SENSOR` wherever possible.

@dagar @DonLakeFlyer can you check it?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -174,193 +174,49 @@
         <description>official stable release</description>
       </entry>
     </enum>
-    <enum name="FLIGHT_MODE">
-      <description>Flight modes of the MAV.</description>
-      <entry value="0" name="FLIGHT_MODE_MANUAL">
-        <description>Manual flight mode</description>
-      </entry>
-      <entry value="1" name="FLIGHT_MODE_ALTCTL">
-        <description>Altitude hold flight mode</description>
-      </entry>
-      <entry value="2" name="FLIGHT_MODE_POSCTL">
-        <description>Position control flight mode</description>
-      </entry>
-      <entry value="3" name="FLIGHT_MODE_STABILIZED">
-        <description>Stabilized mode</description>
-      </entry>
-      <entry value="4" name="FLIGHT_MODE_ACRO">
-        <description>Acro mode</description>
-      </entry>
-      <entry value="5" name="FLIGHT_MODE_RATTITIDE">
-        <description>Rattitude mode</description>
-      </entry>
-      <entry value="6" name="FLIGHT_MODE_AUTO_MISSION">
-        <description>Mission mode</description>
-      </entry>
-      <entry value="7" name="FLIGHT_MODE_AUTO_LOITER">
-        <description>Loiter mode</description>
-      </entry>
-      <entry value="8" name="FLIGHT_MODE_AUTO_RTL">
-        <description>Return to land mode</description>
-      </entry>
-      <entry value="9" name="FLIGHT_MODE_AUTO_RCRECOVER">
-        <description>RC recover mode</description>
-      </entry>
-      <entry value="10" name="FLIGHT_MODE_AUTO_RTGS">
-        <description>Return to ground station mode</description>
-      </entry>
-      <entry value="11" name="FLIGHT_MODE_AUTO_LANDENGFAIL">
-        <description>Land Engine Fail mode</description>
-      </entry>
-      <entry value="12" name="FLIGHT_MODE_AUTO_LANDGPSFAIL">
-        <description>Land GPS fail mode</description>
-      </entry>
-      <entry value="13" name="FLIGHT_MODE_AUTO_TAKEOFF">
-        <description>Takeoff mode</description>
-      </entry>
-      <entry value="14" name="FLIGHT_MODE_AUTO_LAND">
-        <description>Land mode</description>
-      </entry>
-      <entry value="15" name="FLIGHT_MODE_AUTO_FOLLOW_TARGET">
-        <description>Follow target/me mode</description>
-      </entry>
-      <entry value="16" name="FLIGHT_MODE_AUTO_PRECLAND">
-        <description>Precision landing mode</description>
-      </entry>
-      <entry value="17" name="FLIGHT_MODE_DESCEND">
-        <description>Descend mode</description>
-      </entry>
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Looks good. Should let me finish up HIGH_LATENCY2.

message_definitions/v1.0/common.xml
```diff
@@ -174,193 +174,55 @@
...
+        <description>Estimator failure, for example measurement rejection or large variances.</description>
       </entry>
-      <entry value="8192" name="MAV_FAILURE_FLAG_RESERVED4">
-        <description>Reserved for future use.</description>
+      <entry value="8192" name="HL_FAILURE_FLAG_MISSION">
+        <description>Mission failure.</description>
       </entry>
-      <entry value="16384" name="MAV_FAILURE_FLAG_RESERVED5">
-        <description>Reserved for future use.</description>
+      <entry value="16384" name="HL_FAILURE_FLAG_CUSTOM1">
```
dagar: Are these needed short term, or are they reserved for future use?

acfloria: Currently those are not used but I defined them so that someone could use them.

TSC21: @acfloria new entries can be added any time so I don't see the point of those being there.

acfloria: Ok, I removed those custom flags.

message_definitions/v1.0/common.xml
```diff
@@ -4318,28 +4174,27 @@
...
       <field type="int32_t" name="latitude" units="degE7">Latitude, expressed as degrees * 1E7</field>
       <field type="int32_t" name="longitude" units="degE7">Longitude, expressed as degrees * 1E7</field>
       <field type="int16_t" name="altitude" units="m">Altitude above mean sea level</field>
       <field type="int16_t" name="target_altitude" units="m">Altitude setpoint</field>
-      <field type="uint8_t" name="heading" units="deg / 2">Heading (degrees / 2)</field>
-      <field type="uint8_t" name="target_heading" units="deg / 2">Heading setpoint (degrees / 2)</field>
-      <field type="uint16_t" name="target_distance" units="m">Distance to target waypoint or position (meters / 10)</field>
+      <field type="uint8_t" name="heading" units="deg/2">Heading (degrees / 2)</field>
+      <field type="uint8_t" name="target_heading" units="deg/2">Heading setpoint (degrees / 2)</field>
+      <field type="uint16_t" name="target_distance" units="m/10">Distance to target waypoint or position (meters / 10)</field>
```
amilcarlucas: the unit here must be "dm"

message_definitions/v1.0/common.xml
```diff
@@ -4318,28 +4174,27 @@
       <field type="uint32_t" name="timestamp" units="ms">Timestamp (milliseconds since boot or Unix epoch)</field>
       <field type="uint8_t" name="type" enum="MAV_TYPE">Type of the MAV (quadrotor, helicopter, etc., up to 15 types, defined in MAV_TYPE ENUM)</field>
       <field type="uint8_t" name="autopilot" enum="MAV_AUTOPILOT">Autopilot type / class. defined in MAV_AUTOPILOT ENUM</field>
-      <field type="uint8_t" name="flight_mode" enum="FLIGHT_MODE">Flight Mode of the vehicle as defined in the FLIGHT_MODE ENUM</field>
+      <field type="uint16_t" name="custom_mode">A bitfield for use for autopilot-specific flags (2 byte version).</field>
```
amilcarlucas: If it is a bitfield use display="bitmask"

message_definitions/v1.0/common.xml
acfloria: dm would be wrong, one unit is 10 m

message_definitions/v1.0/common.xml
acfloria: Added.

DonLakeFlyer:
Would be good to get this in. Need to finish QGC implementation.

TSC21:
@dagar can you get this in so we can move on with the implementation and testing? Thanks

dagar:
I'm wondering about HL_FAILURE_FLAG. It seems to a lot of this isn't needed. (rc receiver, offboard link, battery) and some of it overlaps with SYS_STATUS.

That being said, it's still an improvement. Let's merge and keep iterating.

amilcarlucas:
This breaks #826 again. Can you please merge that one before continuing to break it ?

acfloria:
@dagar the reason why it overlaps with SYS_STATUS is that we do not need all those flags for HIGH_LATENCY, that is why I condensed them. The additional flags are here so that we potentially do not need to transmit the STATUS_TEXT messages.

message_definitions/v1.0/common.xml
amilcarlucas: I corrected it to the proper `dam` SI Unit in #826

DonLakeFlyer:
So I screwed up on the aligment of the PX4 custom_mode usage. The struct is this:
```
union px4_custom_mode {
	struct {
		uint16_t reserved;
		uint8_t main_mode;
		uint8_t sub_mode;
	};
    struct {
        uint16_t reserved;
        uint16_t custom_mode_hl;
    };
    uint32_t data;
	float data_float;
};
```

I had struct alignment reversed in my head. The issue is that although the actual custom_mode bits used are only 16, they are in the high portion of the 32 bits, not the low portion. Hence you cannot just copy across from a 16 bit value to a 32 bit value. PX4 will need custom code to move the high latency custom_mode values into the internal structure. ArduPilot works fine with this. I don't see any way around this. Hopefully everyone is still ok with this.

DonLakeFlyer:
To pull the value in/out of 16 bits I'm using this which is going to need to match PX4 implementation:
```
union px4_custom_mode {
	struct {
		uint16_t reserved;
		uint8_t main_mode;
		uint8_t sub_mode;
	};
    struct {
        uint16_t reserved_hl;
        uint16_t custom_mode_hl;
    };
    uint32_t data;
	float data_float;
};
```

DonLakeFlyer:
Perfect, thanks. Can you sanity check my QGC pull for this?

Issue #832, opened 2018-01-29
Remove v0.9 from mavlink message_definitions?


samacumen:
I notice that v0.9 is hardly used. In QGC and PX4, v1.0 is used. Why can't we remove v0,9 from message definitions? IMHO, it does a world a lot more good.

Also, I don't really get the meaning of protocol versions Mavlink v1.0, v2.0 (packet format) versus v0.9 and v1.0 under message definitions. Which maps to which ones? Can someone explain me the difference between these? Why are there a different nomenclature in version numbering?

hamishwillee:
>  Why can't we remove v0,9 from message definitions? IMHO, it does a world a lot more good.

Eventually. We tried this and found there were a lot of dependencies in the test code. Yes it could be done, but not really worth the effort.

> Also, I don't really get the meaning of protocol versions Mavlink v1.0, v2.0 (packet format) versus v0.9 and v1.0 under message definitions. Which maps to which ones? Can someone explain me the difference between these? Why are there a different nomenclature in version numbering?

I believe that the message definitions actually changed between v0.9 and v1.0. The definitions for MAVLink 2.0 were changed in a backwards compatible way with 1.0 so there was no need to create a new tree. The current protocol version is listed in the docs https://mavlink.io/en/messages/common.html#mavlink-protocol-version

I do agree this is confusing - we probably should at least have a doc that explains this. 

@julianoes What is the most pragmatic way to help people understand they can ignore 0.9 message definitions and that the 1.0 tree of message definitions covers both v1.0 and 2.0? Options:
- Delete the 0.9 tree
- Add readme in tree explaining it is deprecated (and also in 1.0 tree explaining covers both)
- Add note [here](https://mavlink.io/en/mavlink_version.html#mavlink-version) explaining there is  legacy beta version 0.9 that can be ignored.
- Other?

julianoes:
@hamishwillee ok, just try a PR where you delete it and see how it goes in CI, etc. And then we fix it from there.

amilcarlucas:
I think this can be closed and the discussion can be continued on that PR

PR #835, opened 2018-02-07
Remove 0.9 messages

hamishwillee:
As per https://github.com/mavlink/mavlink/issues/832 the 0.9 messages are confusing. This branch removes them. Further work will be required to fix test errors etc.


hamishwillee:
@julianoes Can we/should we also move the 1.0 messages into the parent folder - ie not have any message versioning in the folder tree. As it is now the version 1.0 is confusing because the folder contains both 1.0 and 2.0 messages. 
I guess we'd leave as-is if we expect versioning in future. Even then though, I'd think we'd be better off considering keep the parent level as "current messages" and pushing legacy message formats down into a lower tree.

julianoes:
@LorenzMeier this is what I was referring to yesterday in the call.

hamishwillee:
@julianoes Can we progress this? Fine not to do right now, but when should I chase you for next steps?

julianoes:
I don't know. There is a bunch of mavlink work to be done that I have to push back for now.

hamishwillee:
OK, I'm just going to take if off my "chase Julian" list for now. We'll talk about it again when/if you have time (or can suggest someone else).

LorenzMeier:
I'll take a pass at this later this week. @amilcarlucas Any feedback from your end on removing v0.9?

amilcarlucas:
We will discuss this on Monday at the ArduPilot Team meeting. I'll get back to you then.

hamishwillee:
@LorenzMeier Yes. I will look at it again now I have a better understanding of the system.

I think it makes sense at same time to also move the whole tree up one - ie so all the definitions in 1.0 move up one level.  Thoughts? 

In dev meeting today we also discussed pushing these messages into their own repo. This is to help a few external generator developers to more easily include just the messages. ThomasG will discuss with you.

hamishwillee:
@LorenzMeier 
1. This is blocked by https://github.com/mavlink/pymavlink/pull/15 - can you merge? 
    - Is it then OK for me to update the pymavlink submodule to current master? 
    - What testing needs to happen after that to verify nothing broke?

Does not affect this PR, but:
2. More generally, I understand there is a plan to drop mavlink/pymavlink and re-sync with ArduPilot/pymavlink. Is that correct? Timing?
3. This removes the 0.9 messages. We should also remove the underlying support for 0.9 in the generators, because any any attempt to use it will fail without them. I could do that, but would be best to have single fork of pymavlink first.

@amilcarlucas Did you have a chance to discuss this in meeting as per https://github.com/mavlink/mavlink/pull/835#issuecomment-367606995 ?
If not, I will put in agenda for next mavlink dev meeting

amilcarlucas:
Going to ardupilot/pymavlink only introduces 2 regressions:
- Mavlink 2.0 Java support (that is half-broken in the fork anyways)
- Ignoring timestamps in Mavlink 2.0 signed messages (that is also a broken concept anyways)
Everything else is in there, I have kept the for up-to date, except the two features mentioned above because they did not got accepted upstream.

The last few weeks I was not able to attend the meeting, I hope that changes soon.
I would love to hear if my deprecated PR got accepted :)

amilcarlucas:
Ardupilot does not use 0.9, It can go away.

amilcarlucas:
Needs a rebase, but AFAIK upstream pymavlink changes are already in place to support this.

hamishwillee:
Damn it, merged via github rather than rebased. Hope that breaks nothing.

hamishwillee:
It fails at the same point, so is it possible mavlink/mavlink is using a changelist of pymavlink before the fix went in?

amilcarlucas:
I re-triggered the CI. Let's see if it works now. Please do a clean rebase later before merging.

hamishwillee:
@amilcarlucas Thank you. CI passed.

> Please do a clean rebase later before merging.

I had to force push but it looks good.

PR #837, opened 2018-02-10
Add camera Mavlink type

Vavooon:
It's hard to add anything :)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -155,6 +155,9 @@
       <entry value="29" name="MAV_TYPE_DODECAROTOR">
         <description>Dodecarotor</description>
       </entry>
+      <entry value="30" name="MAV_TYPE_CAMERA">
+        <description>Camera</description>
+      </entry>
     </enum>
     <enum name="FIRMWARE_VERSION_TYPE">
       <description>These values define the type of firmware release.  These values indicate the first version or release of this type.  For example the first alpha release would be 64, the second would be 65.</description>


hamishwillee:
@dagar @julianoes As per https://github.com/mavlink/mavlink-devguide/issues/33 we need to create  a contribution guide - and document things like this "Adding a new message - how to add message, where to add message, what to use for main fields, who can do it, who needs to approve it ...". We can use this as a test case - can you guys ping me to discuss?

Vavooon:
@hamishwillee Thanks, it's a good idea. I was looking for a contribution guide in README before I created the request.
BTW do you have any comments about the commit utself? Camera type is very important for us because we need to be able to distinguish them from other Mavlink nodes.

message_definitions/v1.0/common.xml
```diff
@@ -155,6 +155,9 @@
       <entry value="29" name="MAV_TYPE_DODECAROTOR">
         <description>Dodecarotor</description>
       </entry>
+      <entry value="30" name="MAV_TYPE_CAMERA">
+        <description>Camera</description>
```
julianoes: Why is this needed? Because there is also [MAV_COMP_ID_CAMERA](http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA).

julianoes: The way to distinguish mavlink nodes is using different component IDs.

Vavooon: @julianoes We find it more flexible when checking only one field for component type detection. 
There is no dedicated component ID for GCS, for example. We also can't use comp_id 0 because our GCS isn't monolithic and consists of few parts.
I would not mention cameras limit while using MAV_COMP_ID_CAMERA but I believe there is a small chance to exceed the limit on big UAVs. 
If adding camera Mavlink type isn't a good idea what value would you suggest to set to MAV_TYPE for this kind of node? It is a bit complicated because gimbals and GCSs already exist in MAV_TYPE enum.

vooon:
We definitely need contribution guide. But more we need to decide approval process, because we have lots of stale PR's which may be merged if decided at a time.

message_definitions/v1.0/common.xml
julianoes: > I would not mention cameras limit while using MAV_COMP_ID_CAMERA but I believe there is a small chance to exceed the limit on big UAVs.

I don't understand what you mean, can you explain?

Vavooon: There are 6 reserved comp IDs for cameras, so that number may be exceeded one day by someone.
Anyway I don't say it's the main argument for adding MAV_TYPE_CAMERA.

julianoes: Well, if you ever exceed the 6 component IDs, you can always add other component IDs that are not taken, or you can differentiate by system ID.

julianoes: What other explanations? You need to make your point very clear, sorry.

Vavooon: Sorry, let me try to explain it better.
1. MAV_TYPE includes all components we use (GCS, GIMBAL, AUTOPILOT) excluding only camera. So it is a good way to use MAV_TYPE in order to determine component type. You may argue that we can use comp_id, but as I wrote there is no GCS in MAV_COMP_ID enum (and we can't just use 0 because our GCS consists of 2 components).
2. Once again, there is MAV_TYPE_GIMBAL and MAV_TYPE_GCS for such kinds of components and no an entry for camera.

vooon: @julianoes what if drone have more than one camera?

julianoes: @vooon: You can use
http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA
http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA2
http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA3
http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA4
http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA5
http://mavlink.org/messages/common#MAV_COMP_ID_CAMERA6

> @Vavooon:
>  but as I wrote there is no GCS in MAV_COMP_ID enum

For the ground station, you could use:
http://mavlink.org/messages/common#MAV_COMP_ID_MAPPER
http://mavlink.org/messages/common#MAV_COMP_ID_MISSIONPLANNER
http://mavlink.org/messages/common#MAV_COMP_ID_PATHPLANNER
http://mavlink.org/messages/common#MAV_COMP_ID_SYSTEM_CONTROL
or even some other number that is not taken yet.

> Once again, there is MAV_TYPE_GIMBAL and MAV_TYPE_GCS for such kinds of components and no an entry for camera.

I can see how there is overlap, and how `MAV_TYPE` is already used for things which are not "MAVs", so I'm not strongly against it, I just wanted to make clear that the use of component ID should be encouraged because mavlink commands for instance can be addressed to a specific component. `MAV_TYPE` is more of a description to inform others what you are.

Vavooon: @julianoes Of course, component ID is important for addressing and we must use it anyway.
You properly noticed that MAV_TYPE describes device type. 
So in our case if we want to detect all nodes in network it is easier to write 
```
switch (msg.comp_id): {
case MAV_TYPE_CAMERA:
comp = (MavComponent*)new MavCamera;
break;

case MAV_TYPE_GIMBAL:
comp = (MavComponent*)new MavGimbal;
break;
```
than checking for MAV_COMP_ID range, for example

```
if (msg.comp_id >= MAV_COMP_ID_CAMERA && msg.comp_id  <= MAV_COMP_ID_CAMERA6) {
comp = (MavComponent*)new MavCamera;
} else if (msg.comp_id == MAV_COMP_ID_GIMBAL) {
comp = (MavComponent*)new MavGimbal;
}
```

The problem with limited number of components of some type is that we can't easily determine type of component which exceeded the limit.
For example, if a system has 2 gimbals (it's more close to reality than 7 cameras because DJI already has systems with 2 gimbals, also remember that gimbals aren't so lucky as cameras and they have only one MAV_COMP_ID_GIMBAL reserved) we can't detect another gimbal by checking comp_id field. If we use other comp_id we need to hardcode its value somewhere or implement another checks in order to be sure that we just detected a gimbal.

IMHO comp_id is better for addressing purposes and MAV_TYPE is better for determining component type because it allows to avoid some mentioned limitations. I'm not trying to say  everyone to use suggested way, just trying to complete the protocol in order to allow easier way to develop Plug-n-Play component type detection.

P.S. We're working on complex software suite (GCS, cameras and gimbals controls over mavlink with multiple vehicles support) so we want to do it much flexible as we can. And that's why using another system ID isn't a solution because it would be considered as a separate system.

julianoes: @Vavooon I understand, I see what you mean now about why it can be limiting.

hamishwillee:
> But more we need to decide approval process, because we have lots of stale PR's which may be merged if decided at a time.

@julianoes YOu just approved this so I guess our approval process is "if some core engineer thinks it makes sense?
Either way, who should now press the merge button, and what is blocking this? I can certainly press it and would normally do so, but are other checks needed?

LorenzMeier:
@hamishwillee I let usually the subject matter expert take the final review and then merge it.

PR #839, opened 2018-02-14
common.xml: add global yaw angle, clarify comments

julianoes:
The current yaw field was implemented as a yaw relative to the vehicle's
heading in QGC and is therefore stabilized that way.
However, we sometimes also would like to know the absolute orientation
of a mount/gimbal relative to North, so that's why the field
`yaw_global` is added.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4426,6 +4426,7 @@
       <field type="float" name="roll" units="deg">Roll in global frame in degrees (set to NaN for invalid).</field>
       <field type="float" name="pitch" units="deg">Pitch in global frame in degrees (set to NaN for invalid).</field>
       <field type="float" name="yaw" units="deg">Yaw relative to vehicle in degrees (set to NaN for invalid).</field>
+      <extensions/>
       <field type="float" name="yaw_absolute" units="deg">Yaw in absolute frame in degrees, North is 0 (set to NaN for invalid).</field>
     </message>
     <message id="266" name="LOGGING_DATA">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4426,7 +4426,7 @@
       <field type="float" name="roll" units="deg">Roll in global frame in degrees.</field>
       <field type="float" name="pitch" units="deg">Pitch in global frame in degrees.</field>
       <field type="float" name="yaw" units="deg">Yaw relative to vehicle in degrees</field>
-      <field type="float" name="yaw_global" units="deg">Yaw in global frame in degrees, North is 0.</field>
+      <field type="float" name="yaw_absolute" units="deg">Yaw in absolute frame in degrees, North is 0.</field>
     </message>
     <message id="266" name="LOGGING_DATA">
       <description>A message containing logged data (see also MAV_CMD_LOGGING_START)</description>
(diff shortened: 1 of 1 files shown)


LorenzMeier:
@magicrub There is an additional field and the message is still marked as WIP, so it should not yet be committed in terms of deployment (we should check that though).

@julianoes How do I know which field is valid?

julianoes:
@magicrub that's what I did. I did not change anything but added a field.
@LorenzMeier unfortunately WIP gets ignored, so that's already in QGC according to @DonLakeFlyer.

@LorenzMeier I would send `NaN` for invalid. I could add that to the comment.

657793213:
@julianoes  HI，I think there should be six value in this migID. roll_relative,pitch_relative,yaw_relative,roll_global,pitch_global,yaw_global.

magicrub:
>@magicrub that's what I did. I did not change anything but added a field.

There was no adding. You changed "yaw" to "yaw_absolute". Doesn't that change the hash?

julianoes:
@magicrub look at the whole diff again.

magicrub:
Ahh, you're right. However, still breaks compatibility. Why not extend it?

DonLakeFlyer:
QGC knows nothing about MOUNT_ORIENTATION. It uses DO_MOUNT_CONTROL to change gimbal angles.

magicrub:
The problem with treating common.xml as PX4's own personal spec is it gets inherited by other projects. So, just because PX4/QGC doesn't use a given packet doesn't mean it's OK to trample on the protocol and break compatibility with other GCS and autopilots. When maintaining a spec, thought should go into long-term compatibility.

julianoes:
> The problem with treating common.xml as PX4's own personal spec is it gets inherited by other projects. So, just because PX4/QGC doesn't use a given packet doesn't mean it's OK to trample on the protocol and break compatibility with other GCS and autopilots. When maintaining a spec, thought should go into long-term compatibility.

Again, I'm not trying to break here. I'm just adding one field.

LorenzMeier:
@julianoes If you add it behind a <extensions/> tag then it will not break backwards-compatibility.

julianoes:
@LorenzMeier oh yes, I forgot to add the `<extensions/>`, will add that.

julianoes:
> @DonLakeFlyer: QGC knows nothing about MOUNT_ORIENTATION. It uses DO_MOUNT_CONTROL to change gimbal angles.

I asked you on Slack and you told me that this is already in QGC, however, looking at the code I can't find `MOUNT_ORIENTATION` or `mount_orientation` being used anywhere, so given that, we could still change this message.

LorenzMeier:
Nobody is using it, so let's merge this as-is.

LorenzMeier:
@julianoes Sorry, I can't merge this yet - the use of NaN for invalid to de-conflict these fields is not documented yet - that needs to be in for this change to be self-consistent.

julianoes:
> Nobody is using it, so let's merge this as-is.

That means we can actually change it and don't just need to add one field though.

julianoes:
@magicrub is this already used by something that we're not aware of?

julianoes:
@magicrub would be great if you could comment:

> @magicrub is this already used by something that we're not aware of?

Thanks!

julianoes:
Ok, I've rebased, and I vote we leave it exactly as is. This way with the `<extensions/>` tag nothing is broken and we have the values required.

PR #841, opened 2018-02-15
Expand battery status to include more information from a fuel gauge

AlexKlimaj:
Added more information to the battery status message to include parameters read from a fuel gauge IC.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4058,12 +4058,12 @@
       <field type="int32_t" name="current_consumed" units="mAh">Consumed charge, in milliampere hours (1 = 1 mAh), -1: autopilot does not provide mAh consumption estimate</field>
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, in HectoJoules (intergrated U*I*dt)  (1 = 100 Joule), -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy: (0%: 0, 100%: 100), -1: autopilot does not estimate the remaining battery</field>
-	  <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in 10*milliamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
-	  <field type="uint16_t" name="serial_number">Battery serial number, 0: battery does not have a serial number</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4058,6 +4058,12 @@
...
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy: (0%: 0, 100%: 100), -1: autopilot does not estimate the remaining battery</field>
+	  <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in 10*milliamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
+	  <field type="uint16_t" name="serial_number">Battery serial number, 0: battery does not have a serial number</field>
+	  <field type="uint16_t" name="capacity" units="mAh">Actual capacity of the battery measured by the fuel gauge, 0: battery does not report actual capacity</field>
+	  <field type="uint16_t" name="cycle_count">Number of charge-discharge cycles reported by the fuel gauge, UINT16_MAX: battery does not report cycle count</field>
+	  <field type="uint16_t" name="run_time_to_empty" units="min">Predicted remaining battery capacity based on the present rate of discharge in min, 0: battery does not report predicted remaining capacity</field>
```
WickedShell: This seems unneeded as the GCS can easily do this math.

message_definitions/v1.0/common.xml
```diff
@@ -4058,6 +4058,12 @@
...
+	  <field type="uint16_t" name="average_time_to_empty" units="min">Predicted remaining battery capacity based on the average rate of discharge in min, 0: battery does not report predicted remaining capacity</field>
```
WickedShell: This is unneeded as it can be easily calculated on the GCS side.

mcsauder: Hi @WickedShell ,

As it pertains to this message, the battery monitor is coulomb counting, so this is a much more accurate representation based off of battery cell direct measurement and calculations made by the smart batter monitor.

Does this help to explain why it is important to pass this information back to the GCS via a message and why this message is important if you have have a smart battery?

Thanks for looking at the PR.

-Mark

AlexKlimaj:
This also fixes those "dam" units...

https://github.com/mavlink/mavlink/pull/826/files#diff-ceacc1e2390186650146a45be8fcb691R4184

amilcarlucas:
dam is m/10
dm is m*10

Which one is correct ?? My guess is that m/10 was wrong.

PR #855, opened 2018-03-08
common: extend RC_CHANNELS_OVERRIDE to 18 channels

julianoes:
This raises the channel count possible for RC override channels to 18,
so on par with RC_CHANNELS.
These channels can be used to send RC input via mavlink (e.g. over wifi)
to a drone.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3348,15 +3348,15 @@
       <field type="uint16_t" name="chan8_raw" units="us">RC channel 8 value, in microseconds. A value of UINT16_MAX means to ignore this field.</field>
       <extensions/>
       <field type="uint16_t" name="chan9_raw" units="us">RC channel 9 value, in microseconds. A value of 0 means to ignore this field.</field>
-      <field type="uint16_t" name="chan10_raw" units="us">RC channel 10 value, in microseconds. A value of 0 means to ignore this field.</field>
-      <field type="uint16_t" name="chan11_raw" units="us">RC channel 11 value, in microseconds. A value of 0 means to ignore this field.</field>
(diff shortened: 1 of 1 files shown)


LorenzMeier:
@amilcarlucas @magicrub Any feedback on how you would like to send RC channel data from the ground to the vehicle? We're trying to not use the same message as we use for reporting RC values to avoid confusion in a multi-node setup.

message_definitions/v1.0/common.xml
```diff
@@ -3346,6 +3346,17 @@
       <field type="uint16_t" name="chan6_raw" units="us">RC channel 6 value, in microseconds. A value of UINT16_MAX means to ignore this field.</field>
       <field type="uint16_t" name="chan7_raw" units="us">RC channel 7 value, in microseconds. A value of UINT16_MAX means to ignore this field.</field>
       <field type="uint16_t" name="chan8_raw" units="us">RC channel 8 value, in microseconds. A value of UINT16_MAX means to ignore this field.</field>
+      <extensions/>
+      <field type="uint16_t" name="chan9_raw" units="us">RC channel 9 value, in microseconds. A value of UINT16_MAX means to ignore this field.</field>
```
WickedShell: Because it's an extension if this is received from a MAVLink 1 source (or MAVLink2 that didn't send this) it will be set to 0. This means that a value of 0 should also be accepted on the extensions as ignore the field.

julianoes: Ok, we'll have to put 0 as "ignore" then. I'll change it.

julianoes:
@WickedShell are you ok with this now? I appreciate your feedback.

julianoes:
@WickedShell ok both it is.

PR #856, opened 2018-03-08
avoidance msg

ChristophTobler:
This is a first draft of the avoidance message according to the [obstacle avoidance interface proposal](https://docs.google.com/document/d/1BQp1a6yszl9f6LDrxrkKUDDGyXxBs5C86BzvJwVbRrU/edit#heading=h.huol20joi641)

The `enum` MAV_TRAJECTORY_REPRESENTATION can still be extended.

The field `field_of_view` probably needs to change a bit as it's currently only possible to have one continuous field of view. A solution would be to have e.g. up to 6 of these (array)?

[RFC](https://github.com/mavlink/rfcs/pull/4)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2921,7 +2921,6 @@
         <param index="3">Z-coordinate of starting bezier point [m], set to NaN if not being used</param>
         <param index="4">Bezier time horizon [s], set to NaN if velocity/acceleration should not be incorporated</param>
         <param index="5">Yaw [rad], set to NaN for unchanged</param>
-        <param index="6">Yaw-rate [rad/s], set to NaN for unchanged</param>
       </entry>
     </enum>
   </enums>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4689,8 +4689,9 @@
       <field type="float" name="yawspeed" units="rad/s">Yaw angular speed</field>
       <field type="float[21]" name="pose_covariance">Pose (states: x, y, z, roll, pitch, yaw) covariance matrix upper right triangle (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
       <field type="float[21]" name="twist_covariance">Twist (states: vx, vy, vz, rollspeed, pitchspeed, yawspeed) covariance matrix upper right triangle (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
-    <message id="332" name="OBSTACLE_AVOIDANCE">
-      <description>Obstacle avoidance message with trajectory points and field of view</description>
+    </message>
+    <message id="332" name="TRAJECTORY">
+      <description>Message to describe a trajectory. Supported trajectory types are enumerated in MAV_TRAJECTORY_REPRESENTATION</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch).</field>
       <field type="uint8_t" name="type" enum="MAV_TRAJECTORY_REPRESENTATION">Waypoints, Bezier etc. see MAV_TRAJECTORY_REPRESENTATION</field>
       <field type="float[11]" name="point_1">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
@@ -4699,10 +4700,6 @@
       <field type="float[11]" name="point_4">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
       <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
       <field type="uint8_t[5]" name="point_valid">States if respective point is valid</field>
-      <field type="int16_t[4]" name="field_of_view">Field of View of the sensor/s being used. The first two parameters are used to indicate the direction (Azimuth: 0 - 360 degrees and elevation: -90 - 90 degrees) and the last two indicate the range (Azimuth and elevation). The range is applied symmetrically in the direction. Set first element to -1 to mark as unknown.</field>
     </message>
   </messages>
 </mavlink>
-
-
-listener trajectory_waypoint 100
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2898,6 +2898,30 @@
         <description>Spektrum DSMX</description>
       </entry>
     </enum>
+    <enum name="MAV_TRAJECTORY_REPRESENTATION">
+      <description>Enumeration of possible waypoint/trajectory representation</description>
+      <entry value="0" name="MAV_TRAJECTORY_REPRESENTATION_WAYPOINTS">
+        <description>Array of waypoints with the following order</description>
+        <param index="1">X-coordinate of waypoint</param>
+        <param index="2">Y-coordinate of waypoint</param>
+        <param index="3">Z-coordinate of waypoint</param>
+        <param index="4">X-velocity of waypoint</param>
+        <param index="5">Y-velocity of waypoint</param>
+        <param index="6">Z-velocity of waypoint</param>
+        <param index="7">X-acceleration of waypoint</param>
+        <param index="8">Y-acceleration of waypoint</param>
+        <param index="9">Z-acceleration of waypoint</param>
+        <param index="10">Yaw</param>
+        <param index="11">Yaw-rate</param>
+      </entry>
+      <entry value="1" name="MAV_TRAJECTORY_REPRESENTATION_BEZIER">
+        <description>Array of bezier points with the following order</description>
+        <param index="1">X-coordinate of starting bezier point</param>
+        <param index="2">Y-coordinate of starting bezier point</param>
+        <param index="3">Z-coordinate of starting bezier point</param>
+        <param index="4">Bezier time horizon</param>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="0" name="HEARTBEAT">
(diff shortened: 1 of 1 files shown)


ChristophTobler:
We first wanted to have something similar (PCA) to be part of the msg but that doesn't really work because of the size. So we will probably have a separate msg that represent nearby obstacles in 3D space. We still need to decide if that replaces `OBSTACLE_DISTANCE` or if `OBSTACLE_DISTANCE` will be extended to 3D or if there will be both.

TSC21:
I am open to discuss a way of extending to a 3D spatial ocupation.

dagar:
Can you resolve the conflict and rebase?

mrivi:
@dagar done!

ChristophTobler:
@TSC21 @dagar Can we merge this one?

TSC21:
Is there an intent to extend to the 3d space representation?

ChristophTobler:
@TSC21 The trajectory is in 3D space. Do you mean the PCA / obstacle representation? If so, I think we could move that to a separate discussion because it shouldn't block this PR.

TSC21:
Yes PCA. Ok can you open a RFC issue so we can discuss it?

ChristophTobler:
@TSC21 https://github.com/mavlink/rfcs/pull/5

thomasgubler:
Looks good to me in general. Can we add units everywhere?

TSC21:
@ChristophTobler @thomasgubler I will leave some comments tomorrow. Thanks for opening the RFC @ChristophTobler

message_definitions/v1.0/common.xml
```diff
@@ -2894,6 +2894,32 @@
         <description>Spektrum DSMX</description>
       </entry>
     </enum>
+    <enum name="MAV_TRAJECTORY_REPRESENTATION">
+      <description>Enumeration of possible waypoint/trajectory representation</description>
+      <entry value="0" name="MAV_TRAJECTORY_REPRESENTATION_WAYPOINTS">
+        <description>Array of waypoints with the following order</description>
+        <param index="1">X-coordinate of waypoint</param>
```
TSC21: If we just use position coordinates, what values are set to the other indexes? `NaN`? I think that needs to be clarified, so the FCU knows what values to process.

message_definitions/v1.0/common.xml
```diff
@@ -4660,5 +4686,16 @@
...
+    <message id="332" name="TRAJECTORY">
+      <description>Message to describe a trajectory. Supported trajectory types are enumerated in MAV_TRAJECTORY_REPRESENTATION</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch).</field>
+      <field type="uint8_t" name="type" enum="MAV_TRAJECTORY_REPRESENTATION">Waypoints, Bezier etc. see MAV_TRAJECTORY_REPRESENTATION</field>
+      <field type="float[11]" name="point_1">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
+      <field type="float[11]" name="point_2">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
+      <field type="float[11]" name="point_3">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
+      <field type="float[11]" name="point_4">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
+      <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
+      <field type="uint8_t[5]" name="point_valid">States if respective point is valid</field>
```
TSC21: How's this set?

message_definitions/v1.0/common.xml
```diff
@@ -4660,5 +4686,16 @@
       <field type="float[21]" name="pose_covariance">Pose (states: x, y, z, roll, pitch, yaw) covariance matrix upper right triangle (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
       <field type="float[21]" name="twist_covariance">Twist (states: vx, vy, vz, rollspeed, pitchspeed, yawspeed) covariance matrix upper right triangle (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
     </message>
+    <message id="332" name="TRAJECTORY">
+      <description>Message to describe a trajectory. Supported trajectory types are enumerated in MAV_TRAJECTORY_REPRESENTATION</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch).</field>
```
TSC21: Shouldn't there be a field for the frame where the navigation will happen?

message_definitions/v1.0/common.xml
```diff
@@ -4660,5 +4686,16 @@
...
+      <field type="uint8_t" name="type" enum="MAV_TRAJECTORY_REPRESENTATION">Waypoints, Bezier etc. see MAV_TRAJECTORY_REPRESENTATION</field>
+      <field type="float[11]" name="point_1">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
```
TSC21: What lead you to define only five points? The message payload size restriction? Or other considerations?

message_definitions/v1.0/common.xml
ChristophTobler: Yes, the ones that are not being used are supposed to be `NaN`. I'll add that to the description. Thanks!

message_definitions/v1.0/common.xml
ChristophTobler: This states which of the 5 points are valid -> boolean. This way we don't need to check for all `NaN`'s

message_definitions/v1.0/common.xml
ChristophTobler: The frame is always the local frame that is being used by the estimator and position controller etc. I can state that in the description. Or should it even be part of the msg name?
The problem is that if we want to allow for global positions as well we need to have `doubles` which makes the msg a lot bigger.

message_definitions/v1.0/common.xml
ChristophTobler: That has been answered in https://github.com/mavlink/rfcs/pull/4#pullrequestreview-110378851 already

ChristophTobler:
General question: How do we deal with the RFC and PR? Now it's a bit messy with all the conversations...

thomasgubler:
> General question: How do we deal with the RFC and PR? Now it's a bit messy with all the conversations...

yes I agree, I was thinking this as well. In my opinion, the RFC should be merged first (which means the mavlink community agrees on the change) and afterwards a PR is opened for the implementation. So on the mavlink implementation PR there should be no conceptual discussion.

ChristophTobler:
>yes I agree, I was thinking this as well. In my opinion, the RFC should be merged first (which means the mavlink community agrees on the change) and afterwards a PR is opened for the implementation. So on the mavlink implementation PR there should be no conceptual discussion.

Makes sense. So this one is an exception as RFC didn't exist at the beginning of this PR.

message_definitions/v1.0/common.xml
```diff
@@ -2898,26 +2898,26 @@
...
-        <param index="1">X-coordinate of starting bezier point</param>
-        <param index="2">Y-coordinate of starting bezier point</param>
-        <param index="3">Z-coordinate of starting bezier point</param>
-        <param index="4">Bezier time horizon</param>
-        <param index="5">Yaw, set to NaN for unchanged</param>
-        <param index="6">Yaw-rate, set to NaN for unchanged</param>
+        <param index="1" units="m">X-coordinate of starting bezier point, set to NaN if not being used</param>
+        <param index="2" units="m">Y-coordinate of starting bezier point, set to NaN if not being used</param>
+        <param index="3" units="m">Z-coordinate of starting bezier point, set to NaN if not being used</param>
+        <param index="4" units="s">Bezier time horizon, set to NaN if not being used</param>
```
TSC21: For bezier, don'y you always need the time horizon?

Stifael: For splines you always need to provide the range of the independent variable `t` (= time horizon). Usually `t` is in between [0,1]. If you are only interested in `Position` information, then having the default range is fine. However, if you want to extract velocity information, then `t` needs to be provided.

Maybe we can say that if time_horizon is set to NAN, then the independent variable assumes the default range [0,1] and the spline will then not incorporate velocity / acceleration information.

ChristophTobler: I'll add that :+1:

MaEtUgR:
I just calculated sizes and write them down here such that not everyone has to repeat:
- `TRAJECTORY` packet: 234 bytes (new)
- `CONTROL_SYSTEM_STATE` packet: 100 bytes (I guess biggest common right now)
- `FOLLOW_TARGET` packet: 93 bytes (I guess second biggest common right now)

ChristophTobler:
Can we get this one in?

message_definitions/v1.0/common.xml
```diff
@@ -2895,6 +2895,32 @@
...
+        <param index="2">Y-coordinate of waypoint [m], set to NaN if not being used</param>
+        <param index="3">Z-coordinate of waypoint [m], set to NaN if not being used</param>
+        <param index="4">X-velocity of waypoint [m/s], set to NaN if not being used</param>
+        <param index="5">Y-velocity of waypoint [m/s], set to NaN if not being used</param>
+        <param index="6">Z-velocity of waypoint [m/s], set to NaN if not being used</param>
+        <param index="7">X-acceleration of waypoint [m/s/s], set to NaN if not being used</param>
+        <param index="8">Y-acceleration of waypoint [m/s/s], set to NaN if not being used</param>
+        <param index="9">Z-acceleration of waypoint [m/s/s], set to NaN if not being used</param>
+        <param index="10">Yaw [rad], set to NaN for unchanged</param>
+        <param index="11">Yaw-rate [rad/s], set to NaN for unchanged</param>
```
mhkabir: We should also have the other dimensions (roll, pitch and rates) in the enum.

Stifael:
> Makes sense. So this one is an exception as RFC didn't exist at the beginning of this PR.

Agree, but still let's move discussion about what should be part of the message and what not to RFC (https://github.com/mavlink/rfcs/pull/4)  and implementation review to here.

mrivi:
@julianoes @TSC21 can we merge this ?

julianoes:
I've been merging stuff too early in the past for mavlink so I'm staying out of this. I suggest to bring this up in the strategy or dev call with @dagar and @LorenzMeier.

thomasgubler:
Let's make sure the new RFC meeting of the maintainers next week happens (as discussed in the strategy call) and approves the RFC. Then to me this looks like it can be merged against the RFC.

@LorenzMeier @mrpollo

mrpollo:
Hey @thomasgubler I'll help you setup the call and promote it ASAP, I'm also worried we are delaying the progress of the project by not defining clear leadership / governance behind the project. I'll help push this through!

LorenzMeier:
As there has been no opposition I will merge this but flag the messages as WIP and DO NOT DEPLOY so that we can revisit and adjust after a proper formal discussion on a call.

PR #860, opened 2018-03-13
[ASLUAV msg subset] update battery management system messages

ASM3:
modify ASLs battery management system messages


message_definitions/v1.0/ASLUAV.xml
```diff
@@ -130,13 +130,15 @@
     </message>
     <message id="209" name="SENS_BATMON">
       <description>Battery pack monitoring data for Li-Ion batteries</description>
+      <field type="uint64_t" name="batmon_timestamp">Timestamp</field>
```
philipoe: 1) Rename name to "timestamp" (to be consistent with all other messages) 2) please give unit [ms, us, ...]

PR #863, opened 2018-03-15
Common: Add an enum for RC reciever types

WickedShell:
Adds an enum for GCS's to use rather then simply hardcoding the values in MAV_CMD_START_RX_PAIR.


amilcarlucas:
Thanks Michael

PR #867, opened 2018-03-18
common: fix MAV_FRAME_LOCAL_NED and MAV_FRAME_LOCAL_ENU definitions

korigod:
It looks like typo in `MAV_FRAME_LOCAL_NED` and `MAV_FRAME_LOCAL_ENU` definitions: it tells NED is Z-up, ENU is Z-down but in fact the opposite is true.


hamishwillee:
@julianoes I think @korigod is correct. Can you please review/approve.

Issue #721, opened 2017-05-06
Support ROS REP 147 Odometry Message


jgoppert:
http://www.ros.org/reps/rep-0147.html

If we support this I think it will be very helpful. Currently we are using local_position_ned_cov + attitude_quaternion_cov to send the same info. We may still have to split the message into the attitude and position parts due to size, but we should at least add the frame_id and child_frame_id so that the messages can be used for various sensors (vision/mocap etc.) These frame ids could be implemented in mavlink using an enum to save space.

@tfoote
@LorenzMeier 
@mhkabir 
@TSC21

TSC21:
Agree!

jgoppert:
I currently am blocking on this for my ROS SLAM system, so I'm going to start adding it.

### ROS side

```bash
# This represents an estimate of a position and velocity in free space.
# The pose in this message should be specified in the coordinate frame given by header.frame_id.
# The twist in this message should be specified in the coordinate frame given by the child_frame_id
Header header
string child_frame_id
geometry_msgs/PoseWithCovariance pose
geometry_msgs/TwistWithCovariance twist
```

### Data Contained

* header
  * frame_id: frame that pose is expressed in, frame with which velocity is respect to
  * stamp: timestamp
  * seq: monotonic sequence number
* child_frame_id: frame which velocity vector is expressed in
* pose
  * pose: [x, y, z], [qw, qx, qy, qz] double
  * covariance: [6x6] double
* twist
  * twist: [vx, vy, vz], [p, q, r]
  * covariance: [6x6] double

* pose: [x, y, z, ], expressed in frame_id

### MAVROS
* Support this odom msg in mavros_extras odometry plugin
* Add support for messages to pose publisher, publish to vision/mocap etc based on frame

### Proposed MAVLink message

* frame enum
  * FRAME_MOCAP_NED [ origin for mocap]
  * FRAME_VISION_NED [ origin for vision ]
  * FRAME_ODOM_NED [ origin for odometry ]
  * FRAME_ESTIM_NED [ origin for estimate ]
  * FRAME_BODY_FRD (forward right down, fixed in body)
  * FRAME_ENU  (want to support? could just throw an error if user gives another frame)
  * ? do we need any others
   * example:
    * to send mocap measurement to vehicle:
      * frame_id: FRAME_MOCAP_NED
      * child_frame_id: FRAME_BODY_FRD
    * to send vision measurement to vehicle:
      * frame_id: FRAME_VISION_NED
[...]

Issue #861, opened 2018-03-14
Local and body frames confusion


korigod:
We have now four main local frames in MAVLink:
- LOCAL_NED,
- LOCAL_OFFSET_NED,
- BODY_NED,
- BODY_OFFSET_NED.

Unfortunately, description of these frames in MAVLink itself is not very clear. [APM docs](http://ardupilot.org/dev/docs/copter-commands-in-guided-mode.html#set-position-target-local-ned) say:
> In frames, `_OFFSET_` means “relative to vehicle position” while `_LOCAL_` is “relative to home position” (these have no impact on velocity directions). `_BODY_` means that velocity components are relative to the heading of the vehicle rather than the NED frame.

Everyone agrees on this definition? What's with positions in BODY_NED? Same docs say:
> **BODY_NED:** Positions are relative to the vehicle’s home position in the North, East, Down (NED) frame. Use this to specify a position x metres north, y metres east and (-) z metres above the home position. Velocity directions are relative to the current vehicle heading. Use this to specify the speed forward, right and down (or the opposite if you use negative values).

It seems quite inconsistent to use different heading in one frame for positions and velocities, isn't it?
The most confusing part, though, is `_NED` as part of `BODY_OFFSET_NED`, which is, as I understand, _never_ really mean NED.

hamishwillee:
The simple way of looking at this is that "Body" implies that velocity is relative to vehicle heading, while "Offset" implies "Position" is relative to vehicle current position. If those aren't specified the position and velocity are relative to NED.

1. So positions in Body NED are NED. 
2. Yes, BODY_OFFSET_NED is a misnomer - the only part of "NED" that is used is the "D" and the way that the axis are defined relative to each other.

It is consistent. I would have quite liked the first frame to just be called NED.

I also think that some of this could be captured in the MAVLink docs: https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_OFFSET_NED

hamishwillee:
Some of this could be captured in the MAVLink docs -e.g. https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_OFFSET_NED

Won't do anything for now though - lets see what other's say. E.g. might be a more clear way than "BODY" to say "Velocity relative to vehicle heading frame".

korigod:
> The simple way of looking at this is that "Body" implies that velocity is relative to vehicle heading, while "Offset" implies "Position" is relative to vehicle current position. 

[...]

PR #869, opened 2018-03-22
[WIP] common: add ODOMETRY msg and Odometry related MAV_FRAMEs

TSC21:
Adresses https://github.com/mavlink/mavlink/issues/721. I added some related Odometry frames to `MAV_FRAME` but we probably should discuss the RFC proposal for local frames of reference in https://github.com/mavlink/rfcs/pull/2.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -601,6 +601,30 @@
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
         <description>Global coordinate frame with above terrain level altitude. WGS84 coordinate system, relative altitude over terrain with respect to the waypoint coordinate. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
+      <entry value="12" name="MAV_FRAME_BODY_FRD">
+        <description>Body fixed frame of reference, Z-down (x: forward, y: right, z: down).</description>
+      </entry>
+      <entry value="13" name="MAV_FRAME_BODY_FLU">
+        <description>Body fixed frame of reference, Z-up (x: forward, y: left, z: up).</description>
+      </entry>
+      <entry value="14" name="MAV_FRAME_MOCAP_NED">
+        <description>Odometry local coordinate frame of data given by a motion capture system, Z-down (x: north, y: east, z: down).</description>
+      </entry>
+      <entry value="15" name="MAV_FRAME_MOCAP_ENU">
+        <description>Odometry local coordinate frame of data given by a motion capture system, Z-up (x: east, y: north, z: up).</description>
+      </entry>
+      <entry value="16" name="MAV_FRAME_VISION_NED">
+        <description>Odometry local coordinate frame of data given by a vision estimation system, Z-down (x: north, y: east, z: down).</description>
+      </entry>
+      <entry value="17" name="MAV_FRAME_VISION_ENU">
+        <description>Odometry local coordinate frame of data given by a vision estimation system, Z-up (x: east, y: north, z: up).</description>
+      </entry>
+      <entry value="18" name="MAV_FRAME_ESTIM_NED">
+        <description>Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-down (x: north, y: east, z: down).</description>
+      </entry>
+      <entry value="19" name="MAV_FRAME_ESTIM_ENU">
+        <description>Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-up (x: east, y: noth, z: up).</description>
+      </entry>
     </enum>
     <enum name="MAVLINK_DATA_STREAM_TYPE">
       <entry name="MAVLINK_DATA_STREAM_IMG_JPEG">
@@ -4571,6 +4595,23 @@
       <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure in centimeters.</field>
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure in centimeters.</field>
+    </message>
+    <message id="331" name="ODOMETRY">
+      <description>Odometry message to communicate odometry information with an external interface. Fits ROS REP 147 standard for aerial vehicles (http://www.ros.org/reps/rep-0147.html).</description>
+      <field type="uint8_t" name="frame_id" enum="MAV_FRAME">Coordinate frame of reference for the pose data, as defined by MAV_FRAME enum.</field>
+      <field type="uint8_t" name="child_frame_id" enum="MAV_FRAME">Coordinate frame of reference for the velocity in free space (twist) data, as defined by MAV_FRAME enum.</field>
+      <field type="float" name="x" units="m">X Position</field>
+      <field type="float" name="y" units="m">Y Position</field>
+      <field type="float" name="z" units="m">Z Position</field>
+      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation)</field>
+      <field type="float" name="vx" units="m/s">X linear speed</field>
+      <field type="float" name="vy" units="m/s">Y linear speed</field>
+      <field type="float" name="vz" units="m/s">Z linear speed</field>
+      <field type="float" name="rollspeed" units="rad/s">Roll angular speed</field>
+      <field type="float" name="pitchspeed" units="rad/s">Pitch angular speed</field>
+      <field type="float" name="yawspeed" units="rad/s">Yaw angular speed</field>
+      <field type="float[21]" name="pose_covariance">Pose (states: x, y, z, roll, pitch, yaw) covariance matrix upper right triangle (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
+      <field type="float[21]" name="twist_covariance">Twist (states: vx, vy, vz, rollspeed, pitchspeed, yawspeed) covariance matrix upper right triangle (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
     </message>
   </messages>
 </mavlink>


TSC21:
@jgoppert, @LorenzMeier, @thomasgubler, @ChristophTobler FYI and review.

message_definitions/v1.0/common.xml
```diff
@@ -4572,5 +4596,24 @@
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure in centimeters.</field>
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure in centimeters.</field>
     </message>
+    <message id="330" name="ODOMETRY">
+      <description>Odometry message to communicate odometry information with an external interface. Fits ROS REP 147 standard for aerial vehicles (http://www.ros.org/reps/rep-0147.html).</description>
+      <field type="uint8_t" name="frame_id" enum="MAV_FRAME">Coordinate frame of reference for the pose data, as defined by MAV_FRAME enum in mavlink_types.h.</field>
```
korigod: Why reference `mavlink_types.h`? `MAV_FRAME` enum is here in `common.xml`.

TSC21: Example: https://github.com/TSC21/mavlink/blob/8a98846a6eaa7fbe3ffa176c4b233ac47a49bb16/message_definitions/v1.0/common.xml#L3421

message_definitions/v1.0/common.xml
```diff
@@ -4572,5 +4596,24 @@
...
+      <field type="uint8_t" name="child_frame_id" enum="MAV_FRAME">Coordinate frame of reference for the velocity data, as defined by MAV_FRAME enum in mavlink_types.h.</field>
+      <field type="float" name="x" units="m">x Position</field>
+      <field type="float" name="y" units="m">y Position</field>
+      <field type="float" name="z" units="m">z Position</field>
+      <field type="float" name="q1">Quaternion component 1, w (1 in null-rotation)</field>
+      <field type="float" name="q2">Quaternion component 2, x (0 in null-rotation)</field>
+      <field type="float" name="q3">Quaternion component 3, y (0 in null-rotation)</field>
+      <field type="float" name="q4">Quaternion component 4, z (0 in null-rotation)</field>
```
korigod: Isn't `float[4]` better for the quaternion? Not sure about x, y, z though. Anyway, it looks really messed up that it's done differently in different messages, maybe we need a standard for this.

TSC21: It needs a std. But actually this can be found the same way on other msgs (like ATTITUDE_QUATERNION).

korigod: That's correct, however it is `float[4]` in the majority of messages (in fact in all except `MAV_CMD_DO_MOUNT_CONTROL_QUAT` and `ATTITUDE_QUATERNION`, and in both these messages the quaternion is the main or the only payload).

message_definitions/v1.0/common.xml
korigod: I believe it's just from MAVLink v0.9, where `MAV_FRAME` is not in `common.xml`. At first glance this should be corrected in v1.0 definitions.

TSC21: Yeah you are right. I will fix the comment.

message_definitions/v1.0/common.xml
TSC21: Yeah I guess it does not matter. But looks better as `float[4]`.

TSC21:
@korigos based on your RFC proposal, do you think the frames of reference I added respect what you propose? I saw you proposed a deeper modification, but you probably are aware that changing whatever in the Mavlink spec is a nightmare. At a max, you can extend the spec.

korigod:
@TSC21 
`MAV_FRAME_BODY_FRD` and `MAV_FRAME_BODY_FLU` are ok but I'd like to clarify their meaning. Is it correct that the z axis of this frames is always normal to the ground, even if the vehicle is tilted (have non-zero pitch/roll angles)?
Your `ODOM` frames does not match the proposed convention. It's ok to amend it of course, let's discuss. First of all, do you really mean these frames are `NED`/`ENU` (geo oriented) or you are just specifying z axis direction?
If `MOCAP`/`VISION` are specifying the origin point location (i.e. frame of reference translation), it may be more consistent to name it like `MAV_FRAME_MOCAP_NED_ODOM` which will comply with the proposed `MAV_FRAME_origin_rotation[_modifier]` convention (e.g. in `MAV_FRAME_LOCAL_NED` we specify the translation before rotation, not after).

message_definitions/v1.0/common.xml
```diff
@@ -601,6 +601,30 @@
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
         <description>Global coordinate frame with above terrain level altitude. WGS84 coordinate system, relative altitude over terrain with respect to the waypoint coordinate. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
+      <entry value="12" name="MAV_FRAME_BODY_FRD">
+        <description>Body fixed frame of reference, Z-down (x: forward, y: right, z: down).</description>
+      </entry>
+      <entry value="12" name="MAV_FRAME_BODY_FLU">
```
korigod: These integer values are duplicated (same for `17`). `Mavgenerate` will fail.

TSC21:
> `MAV_FRAME_BODY_FRD` and `MAV_FRAME_BODY_FLU` are ok but I'd like to clarify their meaning. Is it correct that the z axis of this frames is always normal to the ground, even if the vehicle is tilted (have non-zero pitch/roll angles)?

It should not be consider normal to the ground. It is normal with the vehicle fixed frame.

> First of all, do you really mean these frames are NED/ENU (geo oriented) or you are just specifying z axis direction?

I am considering short-range Cartesian representations, as on ROS REP 103.

> If MOCAP/VISION are specifying the origin point location (i.e. frame of reference translation), it may be more consistent to name it like MAV_FRAME_MOCAP_NED_ODOM which will comply with the proposed MAV_FRAME_origin_rotation[_modifier] convention (e.g. in MAV_FRAME_LOCAL_NED we specify the translation before rotation, not after).

I am not actually trying to fully follow your convention but rather have an approximation. The convention you propose needs to be reviewed by others, so I will stick with this naming until actually a common convention is applied.

message_definitions/v1.0/common.xml
```diff
@@ -601,6 +601,30 @@
...
+        <description>Odometry local coordinate frame of data given by a motion capture system, Z-up (x: east, y: north, z: up).</description>
+      </entry>
+      <entry value="16" name="MAV_FRAME_ODOM_VISION_NED">
+        <description>Odometry local coordinate frame of data given by a vision estimation system, Z-down (x: north, y: east, z: down).</description>
+      </entry>
+      <entry value="17" name="MAV_FRAME_ODOM_VISION_ENU">
+        <description>Odometry local coordinate frame of data given by a vision estimation system, Z-up (x: east, y: north, z: up).</description>
+      </entry>
+      <entry value="18" name="MAV_FRAME_ODOM_ESTIM_NED">
+        <description>Odometry local coordinate frame of data given by a estimator running onboard the vehicle, Z-down (x: north, y: east, z: down).</description>
```
korigod: What is this frame? How does it differ from `MAV_FRAME_LOCAL_NED`? I've read REP-147 and #721 but still don't get it.

korigod:
@TSC21
> I am not actually trying to fully follow your convention but rather have an approximation. The convention you propose needs to be reviewed by others, so I will stick with this naming until actually a common convention is applied.
Of course, I'm just telling that if `MOCAP` specifies frame translation then it would be more consistent with current frame naming to specify it before rotation (like in `MAV_FRAME_LOCAL_NED` etc.)

message_definitions/v1.0/common.xml
TSC21: `MAV_FRAME_LOCAL_NED` may actually come from outside the vehicle. The `MAV_FRAME_ODOM_ESTIM_NED` is a frame that is for sure related with the estimation onboard the vehicle. It's a matter of actually setting the source of the data.

TSC21:
`MOCAP` does not specify the frame translation but rather the data source.

korigod:
@TSC21 

> It should not be consider normal to the ground. It is normal with the vehicle fixed frame.

Ok. I've proposed to use `RPY` (Roll-Pitch-Yaw) to specify this frame orientation as there can be also need in true Forward-Right-Down frame (in which z axis is always down, normal to the ground), e.g. if one wants to command the drone to fly 10 meters forward. I'm not sure how to distinguish z-up and z-down variants though. 

> `MOCAP` does not specify the frame translation but rather the data source.

Yes, but if I understand correctly it specifies the origin as well (it's mocap system origin). Or the origin must be the same as the `MAV_FRAME_LOCAL_NED` origin?

korigod:
@TSC21 By the way, why `VISION`/`MOCAP`/`ESTIM` frame names also include `ODOM`? There is `odom` frame in ROS (introduced in REP-105) with quite a special meaning, it's discussed in mavlink/rfcs#2 as well. Can't it confuse someone who is familiar with ROS?

message_definitions/v1.0/common.xml
korigod: > It needs a std. 

Please feel free to suggest changes or expand. https://github.com/mavlink/rfcs/pull/3

TSC21:
> @korigod agree that odom and vision is a bit confusing. Maybe we shouldn't include odom if we are going to be sensor specific.

I guess I can remove the `ODOM` from the naming.

TSC21:
@korigod @jgoppert OK I gave up of the `ODOM` naming now.

korigod:
Nice. 
@TSC21 @jgoppert Could you please clarify the question about the origins of the proposed frames? Do I understand correctly that they don't necessarily coincide with the `LOCAL_NED` origin but must be fixed w.r.t. the Earth? I need it to add these frames to the RFC proposal (mavlink/rfcs#2).

TSC21:
> Do I understand correctly that they don't necessarily coincide with the LOCAL_NED origin but must be fixed w.r.t. the Earth?

You basically answered your own question :D

julianoes:
@mrivi and @ChristophTobler you might want to review this too.

message_definitions/v1.0/common.xml
```diff
@@ -601,6 +601,30 @@
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
         <description>Global coordinate frame with above terrain level altitude. WGS84 coordinate system, relative altitude over terrain with respect to the waypoint coordinate. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
+      <entry value="12" name="MAV_FRAME_BODY_FRD">
+        <description>Body fixed frame of reference, Z-down (x: forward, y: right, z: down).</description>
```
julianoes: Isn't this the same as `MAV_FRAME_BODY_NED`?

message_definitions/v1.0/common.xml
```diff
@@ -4572,5 +4596,22 @@
...
+      <field type="float" name="y" units="m">Y Position</field>
+      <field type="float" name="z" units="m">Z Position</field>
+      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation)</field>
+      <field type="float" name="vx" units="m/s">X linear speed</field>
+      <field type="float" name="vy" units="m/s">Y linear speed</field>
+      <field type="float" name="vz" units="m/s">Z linear speed</field>
+      <field type="float" name="rollspeed" units="rad/s">Roll angular speed</field>
+      <field type="float" name="pitchspeed" units="rad/s">Pitch angular speed</field>
+      <field type="float" name="yawspeed" units="rad/s">Yaw angular speed</field>
+      <field type="float[21]" name="pose_covariance">Pose covariance matrix upper right triangular (first six entries are the first ROW, next five entries are the second ROW, etc.)</field>
```
julianoes: Can you specify the states of the matrix?

message_definitions/v1.0/common.xml
```diff
@@ -601,6 +601,30 @@
...
+        <description>Odometry local coordinate frame of data given by a vision estimation system, Z-down (x: north, y: east, z: down).</description>
+      </entry>
+      <entry value="17" name="MAV_FRAME_VISION_ENU">
+        <description>Odometry local coordinate frame of data given by a vision estimation system, Z-up (x: east, y: north, z: up).</description>
+      </entry>
+      <entry value="18" name="MAV_FRAME_ESTIM_NED">
+        <description>Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-down (x: north, y: east, z: down).</description>
+      </entry>
+      <entry value="19" name="MAV_FRAME_ESTIM_ENU">
+        <description>Odometry local coordinate frame of data given by an estimator running onboard the vehicle, Z-up (x: east, y: noth, z: up).</description>
```
julianoes: I'm scared by all these frames because I suppose we then need to implement them all on the PX4 side, otherwise it's kind of useless to add them all.

jgoppert:
I think there is some confusion on how this would be used, so I'm reposting this from the original issue to clarify:

### Proposed MAVLink message

* frame enum
  * FRAME_MOCAP_NED [ origin for mocap]
  * FRAME_VISION_NED [ origin for vision ]
  * FRAME_ESTIM_NED [ origin for estimate ]
  * FRAME_BODY_FRD (forward right down, fixed in body)
  * ? do we need any others, can have 256 in a 1 byte enum, but probably don't want to support them all on the firmware side
* **examples**:
  * these 2 byte frames will remove a lot of the duplication we have in various pose sensors that we current have (VISION_POSITION, VISION_VELOCITY, MOCAP_POSITOIN etc.)
  * to send mocap measurement to vehicle:
    * frame_id: FRAME_MOCAP_NED
    * child_frame_id: FRAME_BODY_FRD
  * to send vision measurement to vehicle:
    * frame_id: FRAME_VISION_NED
    * child_frame_id: FRAME_BODY_FRD
  * to receive estimate from vehicle:
    * frame_id: FRAME_ESTIM_NED
    * child_frame_id: FRAME_BODY_FRD

message_definitions/v1.0/common.xml
jgoppert: Yes, I believe they are the same.

message_definitions/v1.0/common.xml
jgoppert: I think for the sensors it makes sense, we can have one mesasge with flags to say what sensor we are talking about and avoid all of the duplicate message. But for ENU and NED it doesn't make sense for PX4 to handle both.

TSC21: @julianoes PX4 does not actually have to handle all the conversions, But, when you set the frame, it will be clear to both sides which are the `frame_id` and the `child_frame_id` and if a frame transform is required or not. For example, if the external interface sends the data in `NED`, then PX4 can handle it directly.

message_definitions/v1.0/common.xml
TSC21: X,Y,Z,R,P,Y cross-covariance.

message_definitions/v1.0/common.xml
TSC21: They are. Though the naming should be `BODY_FRD`, so I added this new one instead of replacing the spec.

message_definitions/v1.0/common.xml
julianoes: I would add that to the description.

TSC21:
> I think there is some confusion on how this would be used, so I'm reposting this from the original issue to clarify:

It was clear on my side @jgoppert. I just added the respective `NED` and `ENU` variants in case you actually want to handle the conversions in the autopilot side. This is a more generic approach and does not harm anyone.

message_definitions/v1.0/common.xml
TSC21: `Pose (position and attitude) covariance matrix upper right triangular (first six entries are the first ROW, next five entries are the second ROW, etc.)`?

julianoes: position is clear but what about attitude? Is that meant for rollspeed or roll angle?

TSC21: It is roll angle. So `orientation` is probably better right?

message_definitions/v1.0/common.xml
TSC21: @korigod I agree with you. But one thing that is certain is that it would be extremely difficult to change the current spec, even if we agree with a correct convention for it.

jgoppert:
Yeah, I agree, just think we would have to document what frames we are supporting.

korigod:
@TSC21 

> You basically answered your own question :D

Great, I'll add these frames (`MOCAP`, `VISION`, `ESTIM`) to the RFC proposal. The only remaining question concerns `BODY_FRD` and `BODY_FLU` frames naming, I think it should be discussed prior merge.

TSC21:
> The only remaining question concerns `BODY_FRD` and `BODY_FLU` frames naming, I think it should be discussed prior merge.

I think that we can follow that discussion after merging and then update if needed. But we definitely should reach a consensus and a base convention for this.

korigod:
> I think that we can follow that discussion after merging and then update if needed. But we definitely should reach a consensus and a base convention for this.

If the update is the changes in the `MAV_FRAME` enum (not an extension or update of the description) I don't think it would be possible, is it? Are we in a hurry here?

These frames are consistent with the convention I proposed, but it's possible someone will suggest something better, and if we merge now we'll have a bunch of these frames which we'll have to deprecate, that's what I'm afraid of.

TSC21:
> I don't think it would be possible, is it? Are we in a hurry here?

No hurry and it is possible while you actually do not add support for it anywhere. So it does not make a big difference.

> if we merge now we'll have a bunch of these frames which we'll have to deprecate, that's what I'm afraid of.

What frames are you afraid can be deprecated? The ones that already exist or the ones being added in this PR?

korigod:
> What frames are you afraid can be deprecated? The ones that already exist or the ones being added in this PR?

Talking about this PR now.

TSC21:
> Talking about this PR now.

Then it matters even less, cause these are not being used anywhere yet.

message_definitions/v1.0/common.xml
korigod: @TSC21 I agree. However, the description of current frames is unclear so I think we actually don't have some solid spec now for `BODY` frames. These frames are interpreted differently by different developers. The only real spec on the matter is the APM docs and it looks like they are not perfectly aligned with the original meaning of these frames (#861). I'm not suggesting breaking changes in the enum anyway.

TSC21: APM developers are open to make their comments and suggest one concept. But, in any situation, the Mavlink spec should go according to a specific meaning and implementation of an autopilot firmware. A good reference to what a "body-fixed"  frame is can be found here: https://ccrma.stanford.edu/~jos/pasp/Body_Fixed_Space_Fixed_Frames_Reference.html. So, at my view, `FLU` and `FRD` do not have to be normal to the ground, but actually should be coincident with principal directions set for the body. So, everything that is actually aligned with a space-fixed frame should be earth aligned, meaning that it is `NED` or `ENU` for our use case. If it is related to the body frame, then it is `FLU` or  `FRD`.

korigod: @julianoes @jgoppert @TSC21 

> Isn't this the same as `MAV_FRAME_BODY_NED`?

I've checked the code of both APM and PX4 and no, the proposed `BODY_FRD` is not the same as `BODY_NED`. In both APM and PX4 `BODY_NED` is the `NED` frame rotated about z axis for the vehicle yaw angle. So in `BODY_NED` z axis is always vertical (at least that is how it's implemented) and in the suggested `BODY_FRS` z axis tilts with the vehicle (like REP-105 `base_link` frame).

julianoes:
@LorenzMeier I propose to merge this unless you have objections.

mhkabir:
In general this is great, but I'm against having ENU frame options since I suspect that we will not end up supporting it.

Just aerospace NED is enough for MAVLink.

TSC21:
@mhkabir if you actually send the data in the ENU frame without transforming it first, then you need to at least reference it (that's the reason for the suffix). For the MAVROS case, for example, we don't have that problem cause we take care of that. But any other API working in the ENU frame and wanting to parse that data, needs to take this into consideration.

I am not actually saying that we need to support frame conversions on the autopilot side, but at least we give the option to do it so. We actually have `MAV_FRAME_LOCAL_ENU` also for a long time, and I believe with the same purpose.

Not adding this option also obliges the developer to implement (if not available) the frame conversion first, considering the autopilots usually work on the aerospace NED. But again, that doesn't mean that the autopilots cannot support that transform task.

mrivi:
@julianoes LGTM

korigod:
> We actually have MAV_FRAME_LOCAL_ENU also for a long time, and I believe with the same purpose.

It's been there for seven years, and still neither APM nor PX4 support it. I just think it can be better not to merge entries which have no support by any firmware at all (at least on WIP stage), it would be really easy to add them later if someone implements corresponding support in the firmwares.

PR #870, opened 2018-03-22
common: remove references to mavlink_types.h

korigod:
In MAVLink v0.9 some enums were defined in `mavlink_types.h` instead of `common.xml`. These references are no longer valid for the v1.0 message set.


korigod:
@julianoes please review.

PR #872, opened 2018-03-23
common.xml: add capture sequence to START_CAPTURE

julianoes:
The sequence/ID as a param allows to distinguish if a command was
retransmitted or if actually yet another capture has been requested.

This can avoid double captures but stil allows to retransmit the
command to make sure it arrives over a lossy link.


julianoes:
@dogmaphobic could you comment?

dogmaphobic:
> @dogmaphobic could you comment?

Makes sense to me. How will you differentiate a starting sequence of 0 and a *not defined* parameter?

julianoes:
@dogmaphobic I've added a commit so that a starting sequence needs to be >= 1 to be valid/used.

PR #667, opened 2017-01-27
Generic estimator support

jgoppert:
This closes https://github.com/mavlink/mavlink/issues/665.

PR #873, opened 2018-03-24
ESTIMATOR msgs by James Goppert

TSC21:
This is a rebased version of #667.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4835,53 +4835,6 @@
       <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure in centimeters.</field>
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure in centimeters.</field>
-    <message id="335" name="ESTIMATOR_STATE">
-      <description>An estimator agnostic state message</description>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
-      <field type="uint8_t" name="n">Number of states, max 21</field>
-      <field type="uint8_t[21]" name="id" enum="MAV_FIELD">An array describing field type (see MAV_FIELD)</field>
-      <field type="uint8_t[21]" name="sensor" enum="MAV_SENSOR_TYPE">An array describing the sensor associated with the field (see MAV_SENSOR_TYPE)</field>
-      <field type="float[21]" name="state">The estimator state</field>
-    </message>
-    <message id="336" name="ESTIMATOR_STATE_STD">
-      <description>An estimator agnostic state standard deviation message</description>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot or since UNIX epoch)</field>
-      <field type="uint8_t" name="n">Number of states, max 21</field>
-      <field type="uint8_t[21]" name="id" enum="MAV_FIELD">An array describing field type (see MAV_FIELD)</field>
-      <field type="uint8_t[21]" name="sensor" enum="MAV_SENSOR_TYPE">An array describing the sensor associated with the field (see MAV_SENSOR_TYPE)</field>
-      <field type="float[21]" name="std">The estimator state standard deviation sqrt(diag(P))</field>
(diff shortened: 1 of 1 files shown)


TSC21:
@jgoppert @LorenzMeier can you please review?

jgoppert:
Thanks for the rebase. I think this would be nice to have.

dagar:
@priseborough you might be interested in this.

message_definitions/v1.0/common.xml
```diff
@@ -2837,6 +2837,222 @@
...
+      <entry value="25" name="MAV_FIELD_ANGVEL_X">
+        <description>Euler Body x rate, radians/second</description>
+      </entry>
+      <entry value="26" name="MAV_FIELD_ANGVEL_Y">
+        <description>Euler Body y rate, radians/second</description>
+      </entry>
+      <entry value="27" name="MAV_FIELD_ANGVEL_Z">
+        <description>Euler Body z rate, radians/second</description>
+      </entry>
+      <entry value="28" name="MAV_FIELD_ROLLRATE">
```
julianoes: What's different between Euler Body rates and Tait-Bryan angle rates?

message_definitions/v1.0/common.xml
```diff
@@ -2837,6 +2837,222 @@
...
+      <!--Geodetic Position -->
+      <entry value="31" name="MAV_FIELD_LAT">
+        <description>Latitude, radians</description>
+      </entry>
+      <entry value="32" name="MAV_FIELD_LON">
+        <description>Longitude, radians</description>
+      </entry>
+      <!-- Sensor Bias NAV Frame -->
+      <entry value="33" name="MAV_FIELD_BIAS_N">
+        <description>North bias</description>
```
julianoes: Unit?

message_definitions/v1.0/common.xml
jgoppert: Rollrate is the same, but here for clarity. For pitch yaw it is different due to incremental frame.

julianoes: Hm ok, I didn't know.

message_definitions/v1.0/common.xml
jgoppert: I'm open to suggestions on what would be best. A lot of the mag measurement stuff is normalized so I hesitated on what to put.

julianoes: Ok, if it is normalized then you could add that in the comment.

TSC21:
@julianoes @jgoppert I think your comments are now addressed. Should this be merged then?

message_definitions/v1.0/common.xml
priseborough: Just to be clear, you are saying 'Euler Body x rate'  is not equivalent to the time derivative of 'Euler Body 3-2-1 (Tait-Bryan) roll angle'? If not then what is it a derivative of?

julianoes: I was confused and didn't realize that we actually use Tait-Bryan angles everywhere and not Euler angles.

priseborough: Sorry Juilan, my question was directed at James. If 'Euler Body x rate' =  d'Euler Body 3-2-1 (Tait-Bryan) roll angle' / dt then shouldn't we use the Tait-Bryan descriptor in the rate field to avoid confusion.

jgoppert:
@priseborough my logic was if some wanted body angular rates , make sure they have consistent labelling. If they euler angle derivatives, also make a consistently labeled set for that. Open to naming adjustments. If we only have one roll rate descriptor some people might get confused finding roll rate when they are looking for body x for example.

jgoppert:
We should probably remove euler angle from body rates. It is really just the angular velocity in the body frame.

priseborough:
The description and your response still leaves me confused about what that rate quantity is. It it:

a) Angular rate in body frame as measured by a rate gyro aligned with the the X,Y and Z axis
b) The rate of change of the Tait-Bryan Angles  as implied by the use of the Euler term.
c) Something else

jgoppert:
We want to provide both and (a) and (b) with names that don't leave you wondering.

priseborough:
If fields 25 to 27 are body rates, ie a), then why does the descriptor contain the term 'Euler' ?  We should be using the term 'Body Axis Rates' to describe what a gyro would measure.

PR #875, opened 2018-03-26
Common: Extend Battery Status with Remaining Time and Low Extent State

### LINKED ISSUE #809: Common low battery state message
Currently there is the common battery status message http://mavlink.org/messages/common#BATTERY_STATUS which is of course very useful. But **it does not contain any low battery state** but only the raw values. There should be **support to allow the autopilot to publish it's low battery state**.

I'm thinking of reflecting following properties: https://github.com/PX4/Firmware/blob/master/msg/battery_status.msg#L16-L20

The problem is that if the GCS only knows the remaining battery percentage and not what warning level this translates to according to the configuration of the autopilot. This means the UI has to duplicate the low battery state by it's own settings again. E.g. currently QGC has it's own low battery percentage threshold to start warning the user and it has to be set every time you switch GCS device or vehicle manually.

The goal use case should be:
The user connects an arbitrary vehicle to the GCS, uses it and the UI can reflect warning levels with e.g. red warnings, beeping or whatever based on the decision of the autopilot and its configuration. This way also the configuration of such levels would stay the same even after using another GCS device.

MaEtUgR:
Allow the autopilot to optionally publish its:
- estimation of the remaining time in seconds the battery can still be used before it's empty.
This information can be showed to the user or used by an external algorithm to further plan the usage of the battery/vehicle.
- low battery state telling the user or external components to what extent the battery is empty.
This information can be used to update the UI with appropriate warnings and inform external components to trigger energy saving or other risk minimizing behaviour.

FYI @julianoes @dogmaphobic 

**EDIT:** fixes #809

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2498,25 +2498,28 @@
         <description>Payload battery</description>
       </entry>
     </enum>
-    <enum name="MAV_BATTERY_LOW_STATE">
+    <enum name="MAV_BATTERY_CHARGE_STATE">
       <description>Enumeration for states of low battery extent</description>
-      <entry value="0" name="MAV_BATTERY_LOW_STATE_UNDEFINED">
+      <entry value="0" name="MAV_BATTERY_CHARGE_STATE_UNDEFINED">
         <description>Low battery state is not provided</description>
       </entry>
-      <entry value="1" name="MAV_BATTERY_LOW_STATE_OK">
+      <entry value="1" name="MAV_BATTERY_CHARGE_STATE_OK">
         <description>Battery is not nearly empty, normal operation</description>
       </entry>
-      <entry value="2" name="MAV_BATTERY_LOW_STATE_LOW">
+      <entry value="2" name="MAV_BATTERY_CHARGE_STATE_LOW">
         <description>Battery state is low, warn and monitor close</description>
       </entry>
-      <entry value="3" name="MAV_BATTERY_LOW_STATE_CRITICAL">
+      <entry value="3" name="MAV_BATTERY_CHARGE_STATE_CRITICAL">
         <description>Battry state is critical, return / abort immediately</description>
       </entry>
-      <entry value="4" name="MAV_BATTERY_LOW_STATE_EMERGENCY">
+      <entry value="4" name="MAV_BATTERY_CHARGE_STATE_EMERGENCY">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2474,6 +2474,27 @@
         <description>Payload battery</description>
       </entry>
     </enum>
+    <enum name="MAV_BATTERY_LOW_STATE">
```
julianoes: I would call it `BATTERY_CHARGE_STATE` because `LOW_STATE` implies that it's already low.

message_definitions/v1.0/common.xml
```diff
@@ -2474,6 +2474,27 @@
...
+      </entry>
+      <entry value="3" name="MAV_BATTERY_LOW_STATE_CRITICAL">
+        <description>Battry state is critical, return / abort immediately</description>
+      </entry>
+      <entry value="4" name="MAV_BATTERY_LOW_STATE_EMERGENCY">
+        <description>Battry state is too low for ordinary abortion, fastest possible emergency stop preventing damage</description>
+      </entry>
+      <entry value="5" name="MAV_BATTERY_LOW_STATE_FAILED">
+        <description>Battry failed, damage unavoidable</description>
+      </entry>
```
julianoes: For smart batteries the state `UNHEALTHY` could also be good which prevents taking off with a "broken" battery.

message_definitions/v1.0/common.xml
MaEtUgR: Your reasoning makes sense, seems I was too burried in my own naming scheme.

message_definitions/v1.0/common.xml
MaEtUgR: Yes or at least lets the user know he needs to order a new one.

MaEtUgR: I extended the definition to errors which prevent usage in general.

MaEtUgR:
I rebased on master and addressed the comment see commit message:
> Thanks to a review comment by @WickedShell I learned that extensions
only available in MAVLink 2 get filled with zeroes in older MAVLink
versions. Hence 0 has to be the value meaning no useful data is provided.
It's a small scrifice to have 1 second battery usage remaining defined
as the value for a completely empty battery.

Thanks for mentioning @WickedShell!

PR #877, opened 2018-03-28
Revert "Merge pull request #873 from TSC21/est_msgs"

julianoes:
This reverts commit b4efea1609631683c7b29f85571ba2f34a10116e, reversing
changes made to 56e28570a7557e62e95613d7bd13a96e8cf5e02e.

This is required because the ESTIMATOR_INNOV_COV message is way too long
and this slipped the review and CI.

@TSC21 @jgoppert sorry this slipped my review. I'm removing all for now. We have to add it again as part of a PR with another review and discussion.

PR #878, opened 2018-03-29
Add bitmask to indicate EKF logic state

jiejiezhang:
add bitmask to indicate EKF logic state

PR #880, opened 2018-03-31
[RFC] [WIP] add ESTIMATOR msgs, MAV_FIELD and MAV_SENSOR_TYPE enum

TSC21:
I am reopening this to have it available for discussion.

Associated issues/PRs:
https://github.com/mavlink/mavlink/pull/667
https://github.com/mavlink/mavlink/pull/873
https://github.com/mavlink/mavlink/pull/877

Note: the reported issue on https://github.com/mavlink/mavlink/pull/877 was actually a rebase problem, as `ESTIMATOR_INNOV_COV` shouldn't have got in.

@LorenzMeier, @julianoes, @priseborough, @jgoppert, @thomasgubler.

Edited: Some actions from discussion:

* [ ] rename euler body rate to just body rate (https://github.com/mavlink/mavlink/pull/880#issuecomment-451058999)


TSC21:
@LorenzMeier can we open a discussion about this?

TSC21:
@jgoppert your input here is also important

hamishwillee:
@TSC21 If you have no joy otherwise, one thing that can be done to help move a PR along is to add the [Dev Call](https://mavlink.io/en/about/support.html#dev_call) label and attend the meeting.

TSC21:
@hamishwillee my problem is actually be able to attend the devcall :D

hamishwillee:
@TSC21 
- What needs to happen to progress this? 
- Does this address all the cases of https://github.com/mavlink/mavlink/pull/878? - if so can we close that as superseded by this?

TSC21:
> - What needs to happen to progress this?

@hamishwillee this has been merged before. I think at this point what we need to address are the interfaces on the consumers side (I can spoke by the PX4 side, but maybe Ardupilot has a word to say about this as well).

> - Does this address all the cases of #878? - if so can we close that as superseded by this?

The idea behind this is that you have a generic estimator interface that allows the autopilot Firmware to pass and receive estimator data. It's agnostic, in the sense that it's not limited to the number of states and the specific estimator interfaces. #878 is an implementation that will continue the focus on the convergence of the interfaces to specific estimators, while what we are trying to achieve here is a more broad approach to the estimators interfaces.
By any chance PR replaces #878, since they have different purposes. What needs to happen, in this sense, is that whatever extension we do to `ESTIMATOR_STATUS` message is a generic approach to an estimator status message, not a specific approach to a commonly used estimator. So, given that, that extension could be matched according to what this PR offers in terms of the considered states. Then, if there are more states to consider, those can be extended on both sides.

hamishwillee:
Hi @TSC21 

Thanks very much. In particular this was not at all obvious:

> The idea behind this is that you have a generic estimator interface that allows the autopilot Firmware to pass and receive estimator data. It's agnostic, in the sense that it's not limited to the number of states and the specific estimator interfaces. 

Why is this useful? Who is the consumer?

You talk about the Firmware receiving estimator data - is the idea you might have an offboard estimator? Is that likely?
Similarly, this can be output by firmware - who would consume it? GCS? 

Basically it would be useful to have all the things expected in the generic RFC process: https://github.com/mavlink/rfcs

That makes it much easier to engage with ArduPilot and other interested parties. At the moment I can't engage anyone because I don't see why this is considered useful.

TSC21:
> You talk about the Firmware receiving estimator data - is the idea you might have an offboard estimator? Is that likely?

Yes that is likely. And that the same time you can have the data being sent to the ground stations or SLAM platforms that require closed-loop pose estimation with covariance to work properly. The broad situation here is to actually allow the usage of onboard or offboard estimators that are not tide to the firmware specifications but rather to an accepted common interface - in this case, it's Mavlink. Narrow the message interfacing to specific estimators results in removing flexibility to anyone trying to move forward with their own estimators. At least, from this point, we can agree on a common estimator interface.

I am out of bandwidth to move to a full spec RFC at this moment. But there may be several people working with GNC projects interested on this. @jgoppert can bring is thoughts as well if he as some time, since he was the first to bring this in.

TSC21:
@hamishwillee the first tackle on this matter was done in https://github.com/mavlink/mavlink/pull/667. You have there a basic explanation for an example where this can become handy.

hamishwillee:
Thanks @TSC21 . I was hoping for the full RFC because eventually we'll want to document this protocol, and the RFC would have fed directly into that. In addition, RFC processes are intended to force discipline on the design. Oh well. I guess all we can do is ping people until either they answer or we get bored.

hamishwillee:
@auturgy @WickedShell Do you know anyone in the ArduPilot community who has an interest in [Simultaneous Localisation and Mapping (SLAM)](https://www.kudan.eu/kudan-news/an-introduction-to-slam/) and might be able to provide feedback on the proposed APIs ([discussed in #667](https://github.com/mavlink/mavlink/pull/667#issuecomment-275555572)) to share estimator data in a generic way?

 @jgoppert @priseborough You've had a couple of years to think about this now - has anything changed that would make you approach this problem differently? Is your recommendation that we push these messages is as is?

jgoppert:
Paul suggested that we rename euler body rate to just body rate, I'm fine with that change.

mhkabir:
@hamishwillee @jgoppert @TSC21 Interest in reviving this?

TSC21:
@mhkabir yep

message_definitions/v1.0/common.xml
```diff
@@ -2885,6 +2885,236 @@
         <description>North, East, Down</description>
       </entry>
     </enum>
+    <enum name="MAV_FIELD">
```
hamishwillee: MAV_FIELD_TYPE ?

hamishwillee:
From what @jgoppert says, this should go in "as is". The only suggestion is https://github.com/mavlink/mavlink/pull/880#issuecomment-451058999

- Is there anyone who thinks that is a bad idea? 
- Any further suggestions?
- If we get this in, who is planning on using it, and when will there be an implementation?

If not, then propose we put a `<wip>` on the message and accept it pending evaluation against a real implementation.

amilcarlucas:
ping

hamishwillee:
Thanks @amilcarlucas ! I'll add to list to discuss in dev call tonight. BUt if no one shows interested will merge anyway.

hamishwillee:
Discussed in the dev call last night. We decided that there is not much pull or interest in this. So closing.

If someone wants to reopen and champion this, then they can feel free, but we'd need answer to the question:

> If we get this in, who is planning on using it, and when will there be an implementation?"

Issue #883, opened 2018-04-10
Video stream settings: Invalid values in the message elements


shakthi-prashanth-m:
In [MAVLink Video stream settings message](https://mavlink.io/en/messages/common.html#SET_VIDEO_STREAM_SETTINGS) below:

Field Name | Type | Description
-- | -- | --
target_system | uint8_t | system ID of the target
target_component | uint8_t | component ID of the target
camera_id | uint8_t | Camera ID (1 for first, 2 for second, etc.)
framerate | float | Frames per second (set to -1 for highest framerate possible)      (Units: Hz)
resolution_h | uint16_t | Resolution horizontal in pixels (set to -1 for highest resolution possible)      (Units: pix)
resolution_v | uint16_t | Resolution vertical in pixels (set to -1 for highest resolution possible)      (Units: pix)
bitrate | uint32_t | Bit rate in bits per second (set to -1 for auto)      (Units: bits/s)
rotation | uint16_t | Video image rotation clockwise (0-359 degrees)      (Units: deg)
uri | char[230] | Video stream URI

Default values for `resolution_h`, `resolution_v` and `bitrate` are -1! All these are of **unsigned type**. Hence it must have **zero** or similar to represent these default settings.

shakthi-prashanth-m:
Also, what is the use case of Camera client providing Video stream URI ? Anyway Camera server can start streaming on any port and provide the URI via `VIDEO_STREAM_INFORMATION`; isn't it ?

julianoes:
Good point. Also, note that `camera_id` was removed again in general, so it needs to be removed from this message as well, or made empty/reserved.

shakthi-prashanth-m:
PR #885 which fixes it, is in [pending ](https://github.com/mavlink/mavlink/pull/885#issuecomment-380792940) state. We can close this soon the discussion in its PR is concluded.

hamishwillee:
Closing as linked items are closed (ie decided not to keep messages)

PR #885, opened 2018-04-11
common.xml: Correct default values of SET_VIDEO_STREAM_SETTINGS

shakthi-prashanth-m:
Default values for `resolution_h`, `resolution_v` and `bitrate` are -1. However, all these are of `unsigned` type. Hence it must have zero or similar to represent these default settings.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4557,9 +4557,9 @@
       <field type="uint8_t" name="target_component">component ID of the target</field>
       <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
       <field type="float" name="framerate" units="Hz">Frames per second (set to -1 for highest framerate possible)</field>
-      <field type="uint16_t" name="resolution_h" units="pix">Resolution horizontal in pixels (set to -1 for highest resolution possible)</field>
-      <field type="uint16_t" name="resolution_v" units="pix">Resolution vertical in pixels (set to -1 for highest resolution possible)</field>
-      <field type="uint32_t" name="bitrate" units="bits/s">Bit rate in bits per second (set to -1 for auto)</field>
+      <field type="uint16_t" name="resolution_h" units="pix">Resolution horizontal in pixels (set to UINT16_MAX for highest resolution possible)</field>
+      <field type="uint16_t" name="resolution_v" units="pix">Resolution vertical in pixels (set to UINT16_MAX for highest resolution possible)</field>
+      <field type="uint32_t" name="bitrate" units="bits/s">Bit rate in bits per second (set to 0 for auto)</field>
       <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise (0-359 degrees)</field>
       <field type="char[230]" name="uri">Video stream URI</field>
     </message>


shakthi-prashanth-m:
Addresses #883

amilcarlucas:
Due to the nature of twos-complement binary numerical representation, -1 is the way to represent the maximum possible value in a straight binary coding. In other words -1 gets coded to 65535 in this particular case. and into 255 in the case of uint8.

AFAIK the original code is already correct.

shakthi-prashanth-m:
@amilcarlucas I got your point. But it is not right (logically and compiler's strong typing point of view) for developers to assign an out of boundary value to objects. In that case, we may recommend assigning UINT8_MAX  for `uint8_t` and UINT16_MAX for type `uint16_t`. Makes sense ?

amilcarlucas:
IMHO, yes, it makes sense to use UINT16_MAX instead of -1 in this case, yes.

shakthi-prashanth-m:
Thanks @amilcarlucas for confirming.
In case of resolution (Horizontal & Vertical) we'd to represent highest resolution possible; so UINT16_MAX makes sense.

But in case of  `bitrate`, we need to indicate value for "auto" setting. Value of 0 seems more intuitive than UINT32_MAX. What do you think ?

shakthi-prashanth-m:
I am sorry if I was not clear in explaining it.

What I meant is that, Instead of setting `bitrate` to -1 for auto, it is better to use value 0. Because in cases of resolution (horizontal & vertical) values UINT16_MAX represents Highest resoultion, so max values makes sense. 


Field Name | Type | Description
-- | -- | --
bitrate | uint32_t | Bit rate in bits per second (set to -1 for auto)

But in case of `bitrate` field auto value is not necessarily max bit rate, so setting 0 would be ideal and intuitive. Sounds good ?

PS: I heard during Dronecode Platform Strategy call that, no body is using MAVLink video stream interface until now (Not even in QGroundControl). Probably we would be first. We are adding Video streaming interface in DroneCore (https://github.com/dronecore/DroneCore/issues/292#issue-300633959):

- start video streaming
- configure video streaming
- request video stream info
- stop video streaming

amilcarlucas:
This needs a rebase, but the issue here is that the change to use "0 for auto" represents a functional change. The other changes are only from a documentation point of view.

hamishwillee:
@dogmaphobic You recently touched SET_VIDEO_STREAM_SETTINGS (https://github.com/mavlink/mavlink/pull/1017). Are you OK with the suggested changes in this PR to change from -1 as the default to UINT16_MAX 

Note, I wouldn't change the set auto value to 0 but perhaps to UINT32_MAX (although I would prefer to change this to zero, as Amlicar points out, it is a break.).

dogmaphobic:
Yes, those changes predate my recent additions. Indeed, it makes no sense to ask for a -1 for an unsigned value. However, it would be a bit more intuitive to use "0" for _default_ resolution as well. Not only it makes it consistent, the use of a _default_ value is a better use case than a _highest resolution possible_, which might be way higher than a given link/GCS may handle.

This is all for "basic" camera/streaming handling where you don't know much about the camera/stream. If you need finer control, you would need to use the PARAM_EXT set of commands.

Also note that this is still WIP (these changes are perfectly valid in this phase) and I am still working on its implementation within QGC. I will also be adding a new "stream type" field (along with an enum for it) as well as some new flags. I was ready to push the PR but I will now wait for this one to merge to avoid unnecessary conflicts.

The new field is:
```
      <field type="uint8_t" name="type" enum="VIDEO_STREAM_TYPE">Type of stream</field>
```

And its enum:
```
    <enum name="VIDEO_STREAM_TYPE">
      <description>Video stream types</description>
      <entry value="0" name="VIDEO_STREAM_TYPE_RTSP">
        <description>Stream is RTSP</description>
      </entry>
      <entry value="1" name="VIDEO_STREAM_TYPE_RTPUDP">
        <description>Stream is RTP UDP (uri gives the port number)</description>
      </entry>
      <entry value="2" name="VIDEO_STREAM_TYPE_TCP_MPEG">
        <description>Stream is MPEG on TCP</description>
      </entry>
      <entry value="3" name="VIDEO_STREAM_TYPE_MPEGTS_H264">
        <description>Stream is h.264 on MPEG TS (uri gives the port number)</description>
      </entry>
    </enum>
```

I'm also adding a new bit to the `VIDEO_STREAM_STATUS_FLAGS` flag:
```
      <entry value="4" name="VIDEO_STREAM_HAS_BASIC_ZOOM">
        <description>Stream has basic zoom control (MAV_CMD_SET_CAMERA_ZOOM)</description>
      </entry>
```

This last one allows QGC to be aware the stream supports basic zoom. This is for cases where you only have a video stream and no "camera" per se. I still don't like how streams and cameras are decoupled but it's too late for that.

dogmaphobic:
> I was ready to push the PR but I will now wait for this one to merge to avoid unnecessary conflicts.

With that said, I need this done within this week :)

dogmaphobic:
Well, now that I actually looked at the changes in this PR and the current state of that message, this PR is quite out of date. Let me push mine through and then we can sort out what people want. The changes are basically just the descriptions.

dogmaphobic:
The discussion above is a bit confusing (my fault) as I was mixing SET and GET. I think of video streams (in their basic forms) as a passive system. You query what there is and you react to it accordingly. That's the *GET* command.

From QGC's perspective, the *SET* command doesn't make a whole lot of sense to me as there is no way for it to know what to enter in those fields. It would need to know what the options are. It's very unlikely you will be able to enter arbitrary values for resolution for example. The great majority of cases, these are fixed to a few choices (480P, 720P and 1080P). Entering -1 for a "maximum available" is rather pointless and anything else would be at best a guess.

I guess this message has a place for custom systems where the GCS knows what to enter but that sort of defeats the purpose of "common".

I'm not planning on implementing this within QGC. If stream configuration is desired, it would have to present a proper XML interface exposing the parameters and their options. All using the PARAM_EXT set of messages.

hamishwillee:
@dogmaphobic Understand your points, and why you don't plan on implementing this.

Even though the message is not well conceived, it still needs to be fixed because -1 is not a proper value for a uint. Either we decide that:
- it is a minor convenience to be able to set UINT16_MAX for "maximum supported resolution" and "0" for default resolution. OR
- That bit of text gets deleted.

If I don't hear from you tomorrow I'll do the first option because that is semantically the same as what is present now - ie I don't need to think about breaks and can just merge it if the tests pass.

dogmaphobic:
Using -1 is certainly wrong. I'm a bit ambivalent about the use of UINT16_MAX or 0 for _"maximum supported resolution"_ but I don't have a strong opinion one way or another. What I do have a strong opinion is about the need for this message altogether. I don't think it makes sense for it to exist. It won't ever be used in a "generic" way.

hamishwillee:
OK, I'm happy to delete it and then close this one. Discussion/approval sought in https://github.com/mavlink/mavlink/pull/1053

hamishwillee:
Closing as https://github.com/mavlink/mavlink/pull/1053 deleted the relevant parts.

PR #887, opened 2018-04-13
common.xml: Remove redundant "No Action" elements

### LINKED ISSUE #886: Remove "No Action" field in several MAVLink commands
### Consider below list of commands:

|CMD ID | Field Name | Description|
|----|----|----|
|521 | MAV_CMD_REQUEST_CAMERA_INFORMATION | Request camera information (CAMERA_INFORMATION).|
|| Mission Param #1 | 0: No action 1: Request camera capabilities|
|| Mission Param #2 | Reserved (all remaining params)|
|522 | MAV_CMD_REQUEST_CAMERA_SETTINGS | Request camera settings (CAMERA_SETTINGS).|
|| Mission Param #1 | 0: No Action 1: Request camera settings|
|| Mission Param #2 | Reserved (all remaining params)|
|525 | MAV_CMD_REQUEST_STORAGE_INFORMATION | WIP: Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.|
|| Mission Param #1 | Storage ID (0 for all, 1 for first, 2 for second, etc.)|
|| Mission Param #2 | 0: No Action 1: Request storage information|
|| Mission Param #3 | Reserved (all remaining params)|
|526 | MAV_CMD_STORAGE_FORMAT | WIP: Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.|
|| Mission Param #1 | Storage ID (1 for first, 2 for second, etc.)|
|| Mission Param #2 | 0: No action 1: Format storage|
|| Mission Param #3 | Reserved (all remaining params)|
|527 | MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS | Request camera capture status (CAMERA_CAPTURE_STATUS) |
|| Mission Param #1 | 0: No Action 1: Request camera capture status|
|| Mission Param #2 | Reserved (all remaining params)|
|529 | MAV_CMD_RESET_CAMERA_SETTINGS | Reset all camera settings to Factory Default|
|| Mission Param #1 | 0: No Action 1: Reset all settings|
|| Mission Param #2 | Reserved (all remaining params)|
|2504 | MAV_CMD_REQUEST_VIDEO_STREAM_INFORMATION | WIP: Request video stream information (VIDEO_STREAM_INFORMATION)|
|| Mission Param #1 | Camera ID (0 for all cameras, 1 for first, 2 for second, etc.)|
|| Mission Param #2 | 0: No Action 1: Request video stream information|
|| Mission Param #3 | Reserved (all remaining params)|

Why do we have **0: No Action** in all the above MAVLink commands ?
For eg:

|521 | MAV_CMD_REQUEST_CAMERA_INFORMATION | Request camera information (CAMERA_INFORMATION).|
|----|----|----|
|| Mission Param #1 | 0: No action 1: Request camera capabilities|
|| Mission Param #2 | Reserved (all remaining params)|

What does it mean to pass Param 1 value as 0 ?
Passing value 1 (Request camera capabilities) should be implicit until we add a command for getting a specific camera information.
So, it is better to remove **No Action** field value from all these commands.

Please correct me if I am wrong or missed any information.

Thanks.

shakthi-prashanth-m:
Fixes #886.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1507,30 +1507,25 @@
       </entry>
       <entry value="521" name="MAV_CMD_REQUEST_CAMERA_INFORMATION">
         <description>Request camera information (CAMERA_INFORMATION).</description>
-        <param index="1">0: No action 1: Request camera capabilities</param>
-        <param index="2">Reserved (all remaining params)</param>
+        <param index="1">Reserved (all params)</param>
       </entry>
       <entry value="522" name="MAV_CMD_REQUEST_CAMERA_SETTINGS">
         <description>Request camera settings (CAMERA_SETTINGS).</description>
-        <param index="1">0: No Action 1: Request camera settings</param>
-        <param index="2">Reserved (all remaining params)</param>
+        <param index="1">Reserved (all params)</param>
       </entry>
       <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
         <description>WIP: Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.</description>
         <param index="1">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
-        <param index="2">0: No Action 1: Request storage information</param>
-        <param index="3">Reserved (all remaining params)</param>
+        <param index="2">Reserved (all params)</param>
       </entry>
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT">
         <description>WIP: Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.</description>
         <param index="1">Storage ID (1 for first, 2 for second, etc.)</param>
-        <param index="2">0: No action 1: Format storage</param>
-        <param index="3">Reserved (all remaining params)</param>
+        <param index="2">Reserved (all params)</param>
       </entry>
       <entry value="527" name="MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS">
         <description>Request camera capture status (CAMERA_CAPTURE_STATUS)</description>
-        <param index="1">0: No Action 1: Request camera capture status</param>
-        <param index="2">Reserved (all remaining params)</param>
+        <param index="1">Reserved (all params)</param>
       </entry>
       <entry value="528" name="MAV_CMD_REQUEST_FLIGHT_INFORMATION">
         <description>WIP: Request flight information (FLIGHT_INFORMATION)</description>
@@ -1539,8 +1534,7 @@
       </entry>
       <entry value="529" name="MAV_CMD_RESET_CAMERA_SETTINGS">
         <description>Reset all camera settings to Factory Default</description>
-        <param index="1">0: No Action 1: Reset all settings</param>
-        <param index="2">Reserved (all remaining params)</param>
+        <param index="1">Reserved (all params)</param>
       </entry>
       <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
         <description>Set camera running mode. Use NAN for reserved values.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes - this changes spec so that if a "request for camera information" is received the actual parameters are ignored (all parameters are reserved). It makes much better sense to me. Can you please merge if happy?

shakthi-prashanth-m:
@julianoes OK, if we're going to add more actions in future, be it.
But, what is the meaning of no actions ? What does MAVLink Camera server do when it receives "No Action" ? If its not supposed to do anything, then why should we send it in the first place ?

julianoes:
> If its not supposed to do anything, then why should we send it in the first place ?

Yes in that case it should not do anything. I agree that's a bit pointless but it allows to add other actions in the future. And it is how some of these requests were already implemented which is why I just copied it.

hamishwillee:
@julianoes I think we can and should change this. There is nothing stopping us adding other events in future following this change (ie unreserving the parameters) - and in the meantime the message is clearer and easier to use.

julianoes:
@hamishwillee this has been in the specs too long now and is already stabilized and implemented. It is too late now to change it.

hamishwillee:
> @hamishwillee this has been in the specs too long now and is already stabilized and implemented. It is too late now to change it.

I don't see how this change could affect any non-idiotic implementation. Your call though.

julianoes:
> I don't see how this change could affect any non-idiotic implementation. Your call though.

If it is changed to `Reserved` it means you can send `NaN` or `0`. In both cases the command won't work anymore if the command handler is implemented correctly. (Unless I'm not getting it?)

hamishwillee:
> If it is changed to Reserved it means you can send NaN or 0. In both cases the command won't work anymore if the command handler is implemented correctly. (Unless I'm not getting it?)

Thanks for explanation - now your point makes perfect sense to me.

shakthi-prashanth-m:
I agree that it would affect the implementation. But its worth removing "0. No Action" entry as I don't think any would ever use that option. We can keep "1. Request camera capabilities", so that we may add in future: "2. Request something else..", etc.

julianoes:
@shakthi-prashanth-m how would that make it clearer?

shakthi-prashanth-m:
@julianoes I meant that no body will ever use "No Action" value for params. I would like to know what makes me pass "No Action" in requesting camera information.

julianoes:
> @julianoes I meant that no body will ever use "No Action" value for params. I would like to know what makes me pass "No Action" in requesting camera information.

That's true but at least it means that it is fully specified and the behavior is not unknown for 0.

LorenzMeier:
@julianoes @shakthi-prashanth-m Escalated in Slack trying to get a final conclusion.

lbegani:
The `No Action` flag makes sense if there is a use-case where the message flows through a pipeline and one of the block enables it, signalling downstream blocks to not handle the message. Not sure if we have any such use case. Also, do not see messages other than camera related using the concept of `No Action` , so there is a bit of inconsistency which IMO is a bigger concern.

hamishwillee:
@LorenzMeier @lbegani Was there ever a resolution to this discussion? Would be good to merge or close this.

FWIW I would prefer it be changed, but this will require co-ordination with Gus (QGC), Julian (SDK) and Lalit (Camera Manager)

julianoes:
I think we have to close this. It's already been implemented this way, and changing it can cause errors and confusion.

hamishwillee:
@julianoes We have to be able to change things, but I guess you're right - this is not broken "enough" to justify the effort of fixing it. If no one disagrees, I will close this next time I come round to it (probably in the next couple of days).

hamishwillee:
OK. Closing. People can reopen in future if they wish to continue the discussion.

PR #892, opened 2018-04-19
Revert "Fix typo in explanation of vz"

julianoes:
This reverts commit 69ff05623272fd1d12da6fec807d744d375af0d4.

This was an oversight in the review and is therefore reverted.

This has all been in NED (North, East, Down) for a long time and there
is no point in changing it.

Reverts #775.


WickedShell:
@julianoes the original PR was opened because the description indicated z-up. Can that be fixed in this PR to indicate Z-down? (Or just remove that whole sentence about GPS frame entirely as the fields document their reference frames)

julianoes:
@WickedShell right, I see!

I agree we should remove `The position is in GPS-frame (right-handed, Z-up).` entirely, it doesn't make any sense.

PR #897, opened 2018-04-25
common.xml: Add COMP_ID and MAV_TYPE for FLARM periphery

umaeder:
FLARM is a cooperative collision avoidance device comparable to ADS-B. It outputs ADSB_VEHICLE packets, like other ADS-B receivers. To distinguish from "real" ADS-B receivers, and to operate both FLARM and ADS-B on the same vehicle, these changes are proposed.


LorenzMeier:
Looking good for me!

PR #900, opened 2018-05-04
Add support for Wi-Fi settings

vldpro:
In this PR I propose to add new messages to setup Wi-Fi of the vehicle using GCS.

Description:
Vehicle can be in 2 states: connected to a selected network (Client mode) or create its own AP (Access Point). Vehicle will have a list of saved (by user) networks to which it may try to connect. If vehicle can't connect to any network then it will start in AP mode.

Communication with GCS (Initially vehicle is in AP):
- GCS requests wifi status (MAV_CMD_GET_WIFI_STATUS)
- GCS requests for number of saved networks (MAV_CMD_GET_WIFI_NETWORKS_COUNT, WIFI_NETWORKS_COUNT)
- GCS requests for networks info (MAV_CMD_GET_WIFI_NETWORKS_INFO, WIFI_NETWORS_INFO)
- User connects Vehicle to a selected network (WIFI_NETWORK_CONNECT)
- Vehicle switches to Client Mode
- After that user can connect to Vehicle again through WiFi 

Additionally, GCS can:
- Start vehicle in AP mode using MAV_CMD_WIFI_START_AP command.
- Add new network (SSID, protocol, password) to saved networks using WIFI_NETWORK_ADD message.
- Delete selected network from saved networks using WIFI_NETWORK_DELETE message.

Also GCS can use messages MAV_CMD_GET_WIFI_NETWORKS_INFO, WIFI_NETWORKS_INFO, MAV_CMD_GET_WIFI_NETWORKS_COUNT, WIFI_NETWORKS_COUNT for obtaining list of scanned networks (Networks which are available for connection)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1770,6 +1770,16 @@
         <param index="7">Reserved (set to 0)</param>
       </entry>
       <!-- END of UAVCAN range -->
+      <entry value="6000" name="MAV_CMD_GET_WIFI_NETWORKS_INFO">
+        <description>Commands the vehicle to respond with a sequence of messages WIFI_NETWORK_INFO. The losses can be detected with MAV_CMD_REQUEST_WIFI_NETWORKS_COUNT</description>
+        <param index="1">0: No Action 1: Request messages with saved Wi-Fi networks 2: Request messages with scanned Wi-Fi network info</param>
+        <param index="2">Reserved (all remaining params)</param>
+      </entry>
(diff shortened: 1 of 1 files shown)


TSC21:
Is there an [RFC](https://github.com/mavlink/rfcs) open for this? https://mavlink.io/en/contributing/contributing.html

vldpro:
[RFC](https://github.com/mavlink/rfcs/pull/6) is created

hamishwillee:
Associated RFC has stalled for more than 6 months, with an end status that it may not make sense to implement at all, but definitely should be done differently. I am therefore closing this - it can be reopened (if that makes sense) when the RFC converges on agreement.

PR #907, opened 2018-05-14
Create BATTERY_INFO to communicate smart battery info.

AlexKlimaj:
Added messages to BATTERY_STATUS extensions to communicate smart battery info such as capacity, serial number, cycle count, etc.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4225,7 +4225,7 @@
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
-      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, -1: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
     </message>
     <message id="148" name="AUTOPILOT_VERSION">
@@ -4771,6 +4771,12 @@
       <field type="float[5]" name="pos_z" units="m">Z-coordinate of starting bezier point, set to NaN if not being used</field>
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
+    </message>
+    <message id="334" name="BATTERY_INFO">
+      <description>Battery information to communicate once per power cycle.</description>
+      <field type="uint32_t" name="serial_number">Battery serial number, UINT32_MAX: battery does not have a serial number</field>
+      <field type="uint32_t" name="capacity" units="mAh">Actual capacity of the battery measured by the fuel gauge, UINT32_MAX: battery does not report actual capacity</field>
+      <field type="uint16_t" name="cycle_count">Number of charge-discharge cycles reported by the fuel gauge, UINT16_MAX: battery does not report cycle count</field>
     </message>
   </messages>
 </mavlink>


message_definitions/v1.0/common.xml
```diff
@@ -4705,6 +4705,14 @@
...
+    <message id="333" name="BATTERY_EXTENDED">
+      <description>Message to communicate smart battery information</description>
+	  <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in 10*milliamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
+	  <field type="uint16_t" name="serial_number">Battery serial number, 0: battery does not have a serial number</field>
+	  <field type="uint16_t" name="capacity" units="mAh">Actual capacity of the battery measured by the fuel gauge, 0: battery does not report actual capacity</field>
+	  <field type="uint16_t" name="cycle_count">Number of charge-discharge cycles reported by the fuel gauge, UINT16_MAX: battery does not report cycle count</field>
+	  <field type="uint16_t" name="run_time_to_empty" units="s">Predicted remaining battery capacity based on the present rate of discharge in seconds, 0: battery does not report predicted remaining capacity</field>
+	  <field type="uint16_t" name="average_time_to_empty" units="s">Predicted remaining battery capacity based on the average rate of discharge in seconds, 0: battery does not report predicted remaining capacity</field>
```
vooon: You break prev message (close tag).

AlexKlimaj: @vooon Thanks, fixed and pushed.

amilcarlucas:
This is interesting for @WickedShell

amilcarlucas:
@LorenzMeier can you add @WickedShell  to the mavlink repository ?

message_definitions/v1.0/common.xml
```diff
@@ -4193,6 +4193,11 @@
       <extensions/>
       <field type="int32_t" name="time_remaining" units="s">Remaining battery time, in seconds (1 = 1s = 0% energy left), 0: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
+      <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in 10*milliamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
```
amilcarlucas: Instead of `10*milliamperes` please use `centiampere`

AlexKlimaj: @amilcarlucas The unit is in cA. I got the wording from the existing current_battery line. 

https://github.com/mavlink/mavlink/blob/467b6bceb896d1032bb863ca861e17ada00c62a0/message_definitions/v1.0/common.xml#L4189

amilcarlucas: Yes, and I changed that wording in ardupilotmega.xml today :)
https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/ardupilotmega.xml#L1308

Kiwa21:
@AlexKlimaj @LorenzMeier 

If that's ok can you also integrate the change to the voltages array ( size 10 -> 12) in this PR ?
It's usual now to have up to a 12S battery, why limit it to 10 ?

` <field type="uint16_t[12]" name="voltages" units="mV">Battery voltage of cells, in millivolts (1 = 1 millivolt). Cells above the valid cell count for this battery should have the UINT16_MAX value.</field>
`

WickedShell:
@Kiwa21 you can't  change the array size without breaking all existing usage of this message, which has seen fairly wide usage.

message_definitions/v1.0/common.xml
hamishwillee: And shouldn't the units be removed from the description? ie >Average battery current. -1: autopilot does not measure the current.

amilcarlucas: Yes, please!

AlexKlimaj: Changed to centiamperes.

AlexKlimaj:
@WickedShell @Kiwa21 What if we add another voltage array to the extended messages to support higher cell count?

message_definitions/v1.0/common.xml
```diff
@@ -4186,13 +4186,18 @@
...
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy: (0%: 0, 100%: 100), -1: autopilot does not estimate the remaining battery</field>
       <extensions/>
       <field type="int32_t" name="time_remaining" units="s">Remaining battery time, in seconds (1 = 1s = 0% energy left), 0: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
+      <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in centiamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
+      <field type="uint16_t" name="serial_number">Battery serial number, 0: battery does not have a serial number</field>
+      <field type="uint16_t" name="capacity" units="mAh">Actual capacity of the battery measured by the fuel gauge, 0: battery does not report actual capacity</field>
```
amilcarlucas: Does it not make more sense to use UINT16_MAX instead of zero here ?

message_definitions/v1.0/common.xml
```diff
@@ -4186,13 +4186,18 @@
...
+      <field type="uint16_t" name="cycle_count">Number of charge-discharge cycles reported by the fuel gauge, UINT16_MAX: battery does not report cycle count</field>
+      <field type="uint16_t" name="average_time_remaining" units="s">Predicted remaining battery capacity based on the average rate of discharge in seconds, 0: battery does not report predicted remaining capacity</field>
```
amilcarlucas: Does it not make more sense to use UINT16_MAX instead of zero here ?

message_definitions/v1.0/common.xml
AlexKlimaj: Hm, my only concern is that a companion computer might mistake UINT16_MAX as an actual reported capacity.

amilcarlucas:
One value needs special treatment anyway, so it makes no difference for the code complexity.
And if you are not careful enough, I guess the battery capacity can go to 0, as well as the average time remaining

AlexKlimaj:
@amilcarlucas Updated the requested fields to UINT16_MAX.

mcsauder:
@amilcarlucas , bump.

message_definitions/v1.0/common.xml
```diff
@@ -4190,8 +4190,13 @@
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
-      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, UINT16_MAX: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
+      <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in centiamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
```
WickedShell: What's the time constant this is averaged over? Presumably not since boot since that would involve a very long period of time with no load.

message_definitions/v1.0/common.xml
```diff
@@ -4190,8 +4190,13 @@
...
+      <field type="uint16_t" name="serial_number">Battery serial number, UINT16_MAX: battery does not have a serial number</field>
+      <field type="uint16_t" name="capacity" units="mAh">Actual capacity of the battery measured by the fuel gauge, UINT16_MAX: battery does not report actual capacity</field>
```
WickedShell: Some monitors report the status in Wh, rather then mAh. Are we expecting the values to always be mAh (the most common I admit) and then have the vehicle make a conversion estimate?

This is also inconsistent as we report current consumed as `int32_t`, which means we are accepting a far larger range of current consumed then capacity reporting.

message_definitions/v1.0/common.xml
```diff
@@ -4190,8 +4190,13 @@
...
+      <field type="uint16_t" name="cycle_count">Number of charge-discharge cycles reported by the fuel gauge, UINT16_MAX: battery does not report cycle count</field>
+      <field type="uint16_t" name="average_time_remaining" units="s">Predicted remaining battery capacity based on the average rate of discharge in seconds, UINT16_MAX: battery does not report predicted remaining capacity</field>
```
WickedShell: What's the time constant?

What's the value for when the capacity exceeds the reportable value? (Either no load or a charging battery). I'm fine with saying it's `UINT16_MAX - 1`, I'd just like to see it documented.

message_definitions/v1.0/common.xml
```diff
@@ -4190,8 +4190,13 @@
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
-      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, UINT16_MAX: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
+      <field type="int16_t" name="average_current_battery" units="cA">Average battery current, in centiamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
+      <field type="uint16_t" name="serial_number">Battery serial number, UINT16_MAX: battery does not have a serial number</field>
```
WickedShell: Given the low frequency with which this changes (as well as `cycle_count` and `capacity`) should these go into a separate message? They increase the overhead by 6 bytes for data that can generally be expected to not change in a flight (except if you hot swap batteries).

mcsauder: @wickedshell, for a driver that hasn't seen any changes for years, I'm not sure where the resistance to change is originating.

WickedShell: Because it's in the outgoing telemetry stream already per battery connected? It increases in size cost must always be paid since the autopilot can't selectively elide messages.It's not a deal breaker, but for constant information like this I'm not sure it makes sense in a streaming message that is otherwise constantly changing data.

mcsauder: Fair enough, and I understand!  We are hoping to accomplish a more generic message implementation this through this PR and your feedback is much appreciated!  Thanks!

WickedShell: I do owe an apology for not getting back quicker on the initial changes on this, I looked briefly at the first version of this and the RFC, but I missed everything after my initial quick comments, and had missed it till the recent ping.

mcsauder: We'll get it right!  Thanks again for your feedback!

message_definitions/v1.0/common.xml
AlexKlimaj: All of the fuel gauges I've seen report both mAh and Wh. The BQ40Z50 for instance, can be configured to report in mAh or "10 mWh". I think leaving this as mAh will cover the most use cases. 

image

Current reported from the fuel gauge is an int32_t. It actually reports negative current when it's being discharged and positive when it's being charged. 

I agree, we can switch capacity to a uint32_t. That will bump the max capacity from 65.535Ah to 4294967.295Ah.

message_definitions/v1.0/common.xml
amilcarlucas: In what product are you working? Can one buy your BMS?

lvale:
This message would be useful if extended to support at least a 12S battery, preferably a 14S. Both of these batteries are in current utilisation (more 12S than 14S) and there are sensors that report the status of each cell via telemetry (using normal RC transmitter and receivers) so adding the possibility of having a message that allows to use these sensors is something that we really need.

AlexKlimaj:
@lvale See my latest commit. I've added a voltages_extended to the extended messages to handle up to 20s batteries.

message_definitions/v1.0/common.xml
AlexKlimaj: Teal One. It will be available soon.

message_definitions/v1.0/common.xml
AlexKlimaj: I've switched the capacity to a uint32_t in the latest commit.

message_definitions/v1.0/common.xml
AlexKlimaj: See the latest commit. I've broken out the one time per flight information to a new BATTERY_INFO message.

lvale:
Thanks, although why 20S ?? I'm not aware of anything running that high voltage, and perhaps when that comes would be as a multiple of 6S or 7S, because then you can use battery packs in series, which is quite common, so would be either 18S or 21S.

AlexKlimaj:
Just an arbitrary value I picked because the other voltage array is size 10.

lvale:
so perhaps getting the array to a value that makes engineering sense ? 21 ?

AlexKlimaj:
Sounds good to me. I'll take it to 11.

lvale:
11 or 21 ?

AlexKlimaj:
My thought is that someone with a battery over 10S would populate the original voltage array and use the extended one for the remaining cells.

lvale:
Again, I would stay within what the industry is mostly using to make it easier, and you get 2/3/4S batteries on RTF vehicles, then usually 6S and then jump to 12S (2 x 6S series) and also 14S (2x7S)

AlexKlimaj:
Okay, so do you want the extended voltage array to be of size 4? So 10+4 cells supported?

AlexKlimaj:
The existing voltage array is of size 10. To stay consistent, adding another array of size 10 makes sense.

message_definitions/v1.0/common.xml
```diff
@@ -4205,8 +4205,11 @@
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
-      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, UINT16_MAX: autopilot does not provide remaining battery time estimate</field>
```
dagar: Something negative instead? Otherwise the field doesn't even need to be a signed integer.

message_definitions/v1.0/common.xml
```diff
@@ -4205,8 +4205,11 @@
...
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
+      <field type="int16_t" name="average_current_battery" units="cA">Average battery current, -1: autopilot does not measure the current</field>
```
dagar: centiamps? Maybe either a signed field or deciamps. It's not common, but I already know systems that exceed this at peek (~320 Amps).

message_definitions/v1.0/common.xml
AlexKlimaj: That came from https://github.com/mavlink/mavlink/pull/875

I can switch it to uint32_t.

message_definitions/v1.0/common.xml
```diff
@@ -4735,5 +4735,11 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="334" name="BATTERY_INFO">
+      <description>Battery one time information</description>
+      <field type="uint16_t" name="serial_number">Battery serial number, UINT16_MAX: battery does not have a serial number</field>
```
dagar: This doesn't seem big enough for most serial numbers.

AlexKlimaj: You're right. I'll update this.

message_definitions/v1.0/common.xml
```diff
@@ -4205,7 +4205,7 @@
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
-      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="uint32_t" name="time_remaining" units="s">Remaining battery time, UINT32_MAX: autopilot does not provide remaining battery time estimate</field>
```
WickedShell: This can't be changed without breaking everyone who is decoding the value already.

hamishwillee: Probably something we'd have to leave for a revision number update/formal release.  

@WickedShell @dagar Perhaps we should start a register of these sorts of changes we'd like to make but can't until we're able to break.

message_definitions/v1.0/common.xml
```diff
@@ -4772,5 +4772,11 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="334" name="BATTERY_INFO">
+      <description>Battery one time information</description>
```
amilcarlucas: I think the description can be improved

mcsauder:
@LorenzMeier , @amilcarlucas , is there any additional work required in this PR we need to take care of?

hamishwillee:
@dagar Anything you'd like to add?

lvale:
Hi @hamishwillee 

12 and 14S are usual these days on the larger vehicles like 700/800 class Helis, large MR like the DJI M600, etc, so saving some "headroom" would be a "kind" of future proof

`Is there a trend to increase "S" numbers used? If so, do we need to future proof further?`

hamishwillee:
lvale - with proper versioning process "coming" a buffer of a few years would be fine. Thanks!

@dagar Anything to add? @MaEtUgR Anything to add for battery message. Otherwise I'm good with this once rebased.

dagar:
Is uint32_t big enough for most serial numbers?

LorenzMeier:
@AlexKlimaj Happy to merge a more holistic PR that exposes all the significant fields of a smart battery.

hamishwillee:
FYI, discussion moved to https://github.com/mavlink/mavlink/issues/988

message_definitions/v1.0/common.xml
```diff
@@ -4225,7 +4225,7 @@
       <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, -1: autopilot does not provide energy consumption estimate</field>
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
-      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="int32_t" name="time_remaining" units="s">Remaining battery time, -1: autopilot does not provide remaining battery time estimate</field>
```
MaEtUgR: JFYI I was told `-1` is not allowed as default here because it's an extension and if the particular piece of software does not support the extension it will get filled with zeroes. Hence 0 has to be the default/unsupported meaning otherwise it might get interpreted wrong. But I think 1 second remaining to capture the battery died now is a reasonable compromise.

hamishwillee: FYI it is true that any empty field(s) at the *end* of a message will be truncated when the message is sent and zero filled again on receipt. While it is true that extensions are ordered after other fields when the message is serialised (so are "most likely" to be empty) this rule also applies to non-extension fields at the end of the message. 

In MAVLink 1 the extension will not be sent at all. If the system sends this message in MAVLink 2 it will *need* to populate the value in order to call the send method - and it *should* use the appropriate value for not supported if the field is not support (in this case -1). 

Upshot, I think this is not a valid point. It is possible I am wrong of course!

[Also note, obviously using 0 is desirable value  for indicating "not supported" because that's the sort of value that people normally plug into the arguments when they don't want to think .... but it isn't mandatory]

FYI, Otherwise this discussion has moved to https://github.com/mavlink/mavlink/issues/988

PR #910, opened 2018-05-17
Mark messages as deprecated using machine readable xml elements and attributes

amilcarlucas:
This allows the pymavlink code generator to optionally drop this messages and
allows the documentation generator to mark this messages in a standard uniform way

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "pymavlink"]
 	path = pymavlink
-	url = https://github.com/mavlink/pymavlink.git
+	url = https://github.com/ardupilot/pymavlink.git
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -1027,7 +1027,8 @@
       <field type="float" name="accel_cal_z">Accel Z calibration.</field>
     </message>
     <message id="151" name="SET_MAG_OFFSETS">
-      <description>Deprecated. Use MAV_CMD_PREFLIGHT_SET_SENSOR_OFFSETS instead. Set the magnetometer offsets.</description>
+      <deprecated since="2014-07" replaced_by="MAV_CMD_PREFLIGHT_SET_SENSOR_OFFSETS"/>
+      <description>Set the magnetometer offsets</description>
       <field type="uint8_t" name="target_system">System ID.</field>
       <field type="uint8_t" name="target_component">Component ID.</field>
       <field type="int16_t" name="mag_ofs_x">Magnetometer X offset.</field>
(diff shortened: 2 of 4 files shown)


amilcarlucas:
This also need some matching pymavlink code changes, but first I want to gather some comments about this.

I also plan to create a `<wip>` tag for work-in-progress messages. and again pymavlink would optionally (via command line parameter) not generate code for those messages.

Comments ?

amilcarlucas:
@tridge using those WIP labels will make it easier to merge the downstream fork (ardupilot would completely ignore all WIP marked messages)

LorenzMeier:
This and the WIP flag are a great idea. Does the WIP flag thinking justify introducing min and max versions per packet so we could start working towards a proper API versioning? That would allow a range of messages to be acceptable by a client (e.g. v1700 - v2122) and would simplify version checking based on expected message range.

hamishwillee:
From a docs point of view wip and deprecated are much better than text. 
1. Is the assumption that these only affect docs - ie they don't affect generated libraries?
2. Is it ever possible to have something that is deprecated and then restored? I guess in that case you just remove the tag?
3. This will require an update to some of the tools in https://github.com/mavlink/mavlink/tree/master/doc

amilcarlucas:
@LorenzMeier A min and max version should allow to deprecate stuff and do WIP as well. So we should decide what method to use. and ´just stick to one method.

@hamishwillee 1. This should affect both the generated docs, and the generated code. I did not do those changes yet. I first need to gather opinions if this is the right way to go and if I missed something.
2. Yes, like you said.
3. Yes, that and pymavlink as well

WickedShell:
@amilcarlucas this shouldn't affect the generated code

hamishwillee:
> @amilcarlucas this shouldn't affect the generated code

Yes. Deprecated is different from versioning. It is a hint to users that something might be removed later.

Re wip vs versioning, I think you probably need both. The reason being that versioning indicates that something is built into the libraries, while wip is again a hint to users that they can't rely on it. It does the same thing as deprecated. WIP should probably have a version "introduced in" as well, so we have a reminder that we need to delete or include more permanently. 

How are versions assigned? Ie does adding a new method cause an iteration of our version "flag"? What is that flag?

amilcarlucas:
@WickedShell the plan is to add command line parameters to pymavlink

- `--no-deprecated` causes the all deprecated messages to not generate any code
- `--no-deprecated-before YYYY-MM` causes the deprecated messages before the date YYY-MM to not generate any code
- `--with-wip` generates code for wip messages

Default will then be: generate deprecated, but not wip

amilcarlucas:
I started working on the pymavlink side in ArduPilot/pymavlink#182

amilcarlucas:
This needs ardupilot/pymavlink#182 to get approved and merged (the XML schema needs to be updated). After that we need to get rid of the pymavlink downstream fork. And I need to add the functionality described two comment above to pymavlink.

But most of all, we need to discuss if this is the way we want to go !!

hamishwillee:
Also, "Default will then be: generate deprecated, but not wip". NOrmally a change would be "default as things are now" - so everyone else should run `--no-wip`

amilcarlucas:
The point is normal users (the majority) do not need nor want WIP messages. The ones that do want then should be savvy enough to add the `--with-wip`  after all they are developers and are messing with XML and C++ code.
The upstream repo has been updated, I will rebase now and merge it myself. Thanks for the input on this!

PR #915, opened 2018-05-18
common.xml: replace the enum indirection layer, with a second message.

amilcarlucas:
This makes the generated code more readable (variables get a proper name)
makes the documentation more readable (less indirection)
Saves mavlink bandwidth when transmiting bezier data

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2914,31 +2914,6 @@
         <description>Spektrum DSMX</description>
       </entry>
     </enum>
-    <enum name="MAV_TRAJECTORY_REPRESENTATION">
-      <description>WORK IN PROGRESS! DO NOT DEPLOY! Enumeration of possible waypoint/trajectory representation</description>
-      <entry value="0" name="MAV_TRAJECTORY_REPRESENTATION_WAYPOINTS">
-        <description>Array of waypoints with the following order</description>
-        <param index="1">X-coordinate of waypoint [m], set to NaN if not being used</param>
-        <param index="2">Y-coordinate of waypoint [m], set to NaN if not being used</param>
-        <param index="3">Z-coordinate of waypoint [m], set to NaN if not being used</param>
-        <param index="4">X-velocity of waypoint [m/s], set to NaN if not being used</param>
-        <param index="5">Y-velocity of waypoint [m/s], set to NaN if not being used</param>
-        <param index="6">Z-velocity of waypoint [m/s], set to NaN if not being used</param>
-        <param index="7">X-acceleration of waypoint [m/s/s], set to NaN if not being used</param>
-        <param index="8">Y-acceleration of waypoint [m/s/s], set to NaN if not being used</param>
-        <param index="9">Z-acceleration of waypoint [m/s/s], set to NaN if not being used</param>
-        <param index="10">Yaw [rad], set to NaN for unchanged</param>
-        <param index="11">Yaw-rate [rad/s], set to NaN for unchanged</param>
-      </entry>
-      <entry value="1" name="MAV_TRAJECTORY_REPRESENTATION_BEZIER">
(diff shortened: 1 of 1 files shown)


amilcarlucas:
@ChristophTobler please review this.

message_definitions/v1.0/common.xml
```diff
@@ -4695,16 +4670,31 @@
...
-      <field type="float[11]" name="point_4">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
-      <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
-      <field type="uint8_t[5]" name="point_valid">States if respective point is valid (boolean)</field>
+    <message id="332" name="TRAJECTORY_REPRESENTATION_WAYPOINTS">
+      <description>WORK IN PROGRESS! DO NOT DEPLOY! Describe a trajectory using an array of up-to 5 waypoints in the local frame.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (since system boot or since UNIX epoch).</field>
+      <field type="uint8_t" name="valid_points">Number of valid points (up-to 5 waypoints are possible)</field>
+      <field type="float[5]" name="pos_x" units="m">X-coordinate of waypoint, set to NaN if not being used</field>
+      <field type="float[5]" name="pos_y" units="m">Y-coordinate of waypoint, set to NaN if not being used</field>
+      <field type="float[5]" name="pos_z" units="m">Z-coordinate of waypoint, set to NaN if not being used</field>
```
ChristophTobler: why is all of this an array -> `float[5]`?

amilcarlucas: This IS an improvement, for the reasons I stated above. There are over 64000 free messages ID's, there is no need to add a indirection layer, that will just complicate things in order to save a couple of those IDs.
Your original version supports up-to 5 data points. This just preserves that.

amilcarlucas: Point_1 to point_5

ChristophTobler:
@LorenzMeier 
>MAVLink always specifies format as well as semantics and the trajectory message essentially "hacked" multiple semantics.

Like e.g. http://mavlink.org/messages/common#COMMAND_LONG ...?

But if we only support 2 types (waypoints and bezier) as discussed then it makes sense to use this suggested representation, I guess.

LorenzMeier:
The command interface is fundamentally different because the nature of a command is to transport different types of action calls through a single interface. The command ID has always the same semantic and the message only carries optional parameters for each command. The way to understand that intuitively is to imagine a recipient for all possible command messages: The recipient will always be able to decode the full command and hand it off to an onboard interface using a *single semantic*. The comparison actually supports the argument, rather than contradicting it.

The avoidance interface, by contrast, is a continuous stream like many others in the system representing a state of the vehicle. That drives the desire to architect it like all the other streams and make the semantic well and tightly defined - because that is driving at the end of the day the interoperability and maintainability of MAVLink.

amilcarlucas:
@LorenzMeier can this please be merged?

TSC21:
@amilcarlucas can you rebase please?

hamishwillee:
@amilcarlucas @LorenzMeier So what is this waiting on?

amilcarlucas:
This is waiting for @LorenzMeier to press the "rebase and merge" button since 26 May.

LorenzMeier:
Not quite. I’d like to hear back from @ChristophTobler if he is now happy with it.

ChristophTobler:
I'm happy with it.
But this needs to be synced with Firmware and MAVROS. I'll coordinate with @mrivi and post the respective PRs in here.

mrivi:
@LorenzMeier if the Firmware mavlink submodule is updated the avoidance interface (that is in masyer since yestarday) won't work because we don't have a PR ready that makes it compatible with this new message structure. Same for Mavros.

PR #916, opened 2018-05-18
Add timestamp to the COMMAND_INT and COMMAND_LONG message

acfloria:
Adding a timestamp to the command messages helps to identify outdated commands and rejecting/ignoring them.

This is important once there are multiple links with different latencies connected to the vehicle. In this setup it can happen that a command sent earlier from the ground station appears after a later sent one and therefore the order of execution is different.

Here is one example which happened during our flight:

- Our setup includes a radio link with a negligible latency and a satellite link with a latency of minimum ten seconds up to a couple of minutes. What happened during one flight is that we sent a "camera door open" command over the satellite link due to bad signal it took over one minute to transmit the message. In the meantime we sent the "camera door close" command. In the end, due to the different latencies, the close command arrived before the open command on the plane so for the rest of the flight the camera door was open.

The same could happen with more safety critical commands such as: `MAV_CMD_DO_CHANGE_ALTITUDE` or `MAV_CMD_DO_REPOSITION` where the vehicle suddenly flies at the wrong altitude or to the wrong position.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3484,6 +3484,7 @@
     </message>
     <message id="75" name="COMMAND_INT">
       <description>Message encoding a command with parameters as scaled integers. Scaling depends on the actual command value.</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the COMMAND, as defined by MAV_FRAME enum</field>
@@ -3500,6 +3501,7 @@
     </message>
     <message id="76" name="COMMAND_LONG">
       <description>Send a command with up to seven parameters to the MAV</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
       <field type="uint8_t" name="target_system">System which should execute the command</field>
       <field type="uint8_t" name="target_component">Component which should execute the command, 0 for all components</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID, as defined by MAV_CMD enum.</field>


message_definitions/v1.0/common.xml
```diff
@@ -3484,6 +3484,7 @@
     </message>
     <message id="75" name="COMMAND_INT">
       <description>Message encoding a command with parameters as scaled integers. Scaling depends on the actual command value.</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (microseconds, synced to UNIX time or since system boot)</field>
```
LorenzMeier: Because of MAVLink compatibility checks this will break compatibility. You would need to add this to the end of the message with the right extension tag.

Is the timestamp here really helpful or is what you want a command sequence number?

acfloria: Ok, I will change it.
In general a sequence number could also work for only one ground station. But if we also want to allow using multiple groundstations for one vehicle then I think the timestamp would be the way to go.

WickedShell: How are you proposing vehicles use this timestamp? Every GCS will have a different sense of time, which can lead to some GCS's being denied entirely. You'd have to always break it down to a looking at the timestamp per sender ID, which still doesn't help if one senders link is heavily delayed. It's made even worse since you are proposing to accept 2 different types of time stamping in this field, without any indication of what version is being sent to the aircraft.

I think this needs a lot more protocol thought/description then just an extra field on the messages that will be unclear how to use/how the vehicle will be using it.

acfloria: @WickedShell I agree with you. The timestamp alone is not enough but it is necessary for the solution I have in mind.

My Idea is to basically use the timestamp of the vehicle for that field. To do so the Ground Control Station needs to be synchronized with the vehicle time. The idea was to use the `SYSTEM_TIME` message from the vehicle to get the vehicle time to GCS. This would synchronize the GCS with the vehicle at the accuracy of the low latency link which should be accurate enough for the command timestamp. As long as the synchronization with the vehicle is not completed I would always fill in a timestamp of 0.

On the vehicle side I would track the timestamps of the received commands per command ID and reject if a command with the same ID and a larger timestamp is received. To not break the current behaviour the vehicle would always execute commands with a timestamp of 0.

LorenzMeier: Could you please send a PR for an RFC detailing the whole design?

LorenzMeier: https://github.com/mavlink/rfcs/blob/master/README.md

acfloria: Sure, @LorenzMeier can you give me write access to that repo?

acfloria: @LorenzMeier @WickedShell https://github.com/mavlink/rfcs/pull/9

PR #917, opened 2018-05-18
common: update covariances descriptions

TSC21:
This PR:
1. Updates the description of the covariance matrices fields with the right and more informative explanation of the fields. Also, fits the proposed in https://github.com/mavlink/rfcs/pull/8 for identifying the matrix when it is unknown;
2. Updates the `ODOMETRY` msg by replacing the `twist_covariance` field with `velocity_covariance` field, it order to support a future addition of acceleration and its respective cross-covariance (according to [ROS REP 147](http://www.ros.org/reps/rep-0147.html) and the addition of a new proposed [nav_msgs/OdometryWithAcceleration`](https://github.com/ros/common_msgs/pull/87).
3. Corrects a misconception on `DISTANCE_SENSOR`, since where it's read `covariance` should instead be `variance`. This change thought would affect the msg API so I would like to get the feedback from its consumers in terms of changing its naming and description, or just the description (taking into account that the naming is wrong).


TSC21:
@LorenzMeier at your consideration for review. Thanks!

@amilcarlucas, @WickedShell can you evaluate this PR in terms of what is done currently on APM side, specially regarding the `DISTANCE_SENSOR` msg? Thanks!

TSC21:
@WickedShell friendly ping.

message_definitions/v1.0/common.xml
```diff
@@ -4057,7 +4057,7 @@
       <field type="uint8_t" name="type" enum="MAV_DISTANCE_SENSOR">Type from MAV_DISTANCE_SENSOR enum.</field>
       <field type="uint8_t" name="id">Onboard ID of the sensor</field>
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces from MAV_SENSOR_ORIENTATION enum. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
-      <field type="uint8_t" name="covariance" units="cm">Measurement covariance in centimeters, 0 for unknown / invalid readings</field>
+      <field type="uint8_t" name="variance" units="cm">Measurement variance in centimeters. If unknown, assign NaN value.</field>
```
WickedShell: ArduPilot has shipped versions that are providing 0 for the covariance. I think for now we would need to continue to accept 0 for unknown/invalid covariance.

WickedShell: I'm fine with changing the message to be either is accepted for unknown now, and pretty comfortable with sending NaN's from the aircraft. The only person who is going to get confused by the definition change will be a GCS talking to an exsisting vehicle.

As far as input to the vehicle ArduPilot dosen't use the covariance at all, so from that perspective the major issue is will that confuse/bother GCS's?

TSC21: > As far as input to the vehicle ArduPilot dosen't use the covariance at all, so from that perspective the major issue is will that confuse/bother GCS's?

Good question. @DonLakeFlyer @dogmaphobic any comments about this?

TSC21:
@LorenzMeier can we push this in?

TSC21:
@LorenzMeier good to go?

message_definitions/v1.0/common.xml
priseborough: A variance has units of 'measurement unit'^2

priseborough: If we want to keep this as a uint8_t type, then it wil need to be specified as error_sd, not variance where error_sd is the 1-SD error in the distance measurement and has units of cm.

thomasgubler:
@hamishwillee @LorenzMeier could you please discuss next steps for this PR in the next mavlink call?

hamishwillee:
@thomasgubler I've added the Dev Call label, which is what automatically gets this discussed in the call. 

@TSC21 @WickedShell There is a lot of back and forth on this, but from your discussion it looked like the main impact was on GCS:  https://github.com/mavlink/mavlink/pull/917#discussion_r189690145

If that is still the case, then has anyone been able to coordinate with @DonLakeFlyer  and @dogmaphobic. Presumably Mission Planner and others would also be affected?

@TSC21 There is also comment from Paul in https://github.com/mavlink/mavlink/pull/917#discussion_r201181619 - was that addressed?

TSC21:
@priseborough I addressed your comments and changed the field type to `float` and set the units to `cm2`.

> @TSC21 @WickedShell There is a lot of back and forth on this, but from your discussion it looked like the main impact was on GCS: #917 (comment)

@hamishwillee from what I could inspect, there's not current impact on the GCS side (at least in QGC).

>@TSC21 There is also comment from Paul in #917 (comment) - was that addressed?

Yes, it was. Done on my latest commits.

message_definitions/v1.0/common.xml
TSC21: @priseborough changed to `float`.

amilcarlucas:
done ArduPilot/pymavlink#219 used `cm^2`

amilcarlucas:
AFAIK it is better to add the pymavlink submodule update to this PR. This makes it "atomic" :)

TSC21:
@LorenzMeier done

TSC21:
@LorenzMeier good to go?

WickedShell:
The type change in distance sensor is breaking and can't come in.

TSC21:
The name change would also break. So what is the difference? The consumers will always have to update its interfaces

WickedShell:
@TSC21 Right, that breakage is unacceptable. There are a large number of deployed use cases that send and recieve the `DISTANCE_SENSOR` message as it stands. If you break the CRC seed on them they all break. All that can be done to that message now is to either add an extension or change the description, changing the actual fields types or names in the message cannot be done.

WickedShell:
@hamishwillee We have support for both rangefinders and proximity sensors using this message published from ROS nodes, as well as publishing the message to GCS's for visualization and flight monitoring. That's a large set of places to be breaking at once in my opinion.

DonLakeFlyer:
FYI: QGC doesn't use DISTANCE_SENSOR.covariance, only id, orientation and current_distance.

hamishwillee:
Thanks @DonLakeFlyer . Unfortunately that doesn't help us since the CRC_EXTRA is across the whole message, not just a field. QGC will still see any changes as a bad CRC.

message_definitions/v1.0/common.xml
```diff
@@ -4091,7 +4091,7 @@
       <field type="uint8_t" name="type" enum="MAV_DISTANCE_SENSOR">Type of distance sensor.</field>
       <field type="uint8_t" name="id">Onboard ID of the sensor</field>
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
-      <field type="uint8_t" name="covariance" units="cm">Measurement covariance, 0 for unknown / invalid readings</field>
+      <field type="float" name="variance" units="cm^2">Measurement variance in centimeters squared. If unknown, assign NaN value.</field>
```
hamishwillee: @TSC21 This has dragged on a bit sorry. Given that this change to DISTANCE_SENSOR is the only controversial bit, do you think you could revert that bit and rebase - I'll happily merge the rest.

I don't recall discussing this in the dev meeting. We can raise again in the next one if you like, but I do expect that as a group people will be opposed. 
One option would be to add a new field as an message extension. We can't remove the old field without breaking things, but we could change the description text to "Deprecated". 

Thoughts?

TSC21: Well adding a new field will probably add unnecessary overhead. The correct way of doing this was to replace the naming, since it's completely wrong. But given that the assumption is that this will break deployed platforms, I guess the only thing we can do is actually change the description.

TSC21: Also a note that type of the field is completely wrong as well. So this will probably have to consider what @priseborough suggested in https://github.com/mavlink/mavlink/pull/917#discussion_r216565966.

hamishwillee: @TSC21 Can you please revert the DISTANCE_SENSOR change here so that all the rest of this useful information can go in. We can then address the controversial discussion in #1047.

TSC21: @hamishwillee done!

hamishwillee:
@WickedShell @auturgy In order to push all the docs changes in this PR to master we've reverted the (desirable) breaking change to DISTANCE_SENSOR for separate discussion in #1047.

Have you any other feedback before I merge this? @thomasgubler @priseborough ditto?

hamishwillee:
Actually @TSC21 Paul says "If we want to keep this as a uint8_t type, then it wil need to be specified as error_sd, not variance where error_sd is the 1-SD error in the distance measurement and has units of cm." 

- I interpret that to mean that units have to be `cm` and text has to refer to it as the 1-SD error.

PR #924, opened 2018-05-25
Add vehicle command for Orbit

MaEtUgR:
Add a command to make the vehicle orbit around a target position.
The executing code for that task on PX4 is located here: https://github.com/PX4/Firmware/blob/master/src/lib/FlightTasks/tasks/FlightTaskOrbit.hpp
And is currently being transformed from a draft to a supported feature here:
https://github.com/PX4/Firmware/pull/9434

This message allows the UI to access this functionality and provide all the necessary parameters to determine the behaviour like:
- Orbit center latitude, longitude, altitude (with option for local coordinates)
defaults to where the vehicle currently is
- Orbit radius (in meters)
defaults to vehicle configuration e.g. 1m for safety
- Orbit velocity (tangential in m/s, positive CW, negative CCW)
defaults to vehicle cruise velocity or less if the radius is too small
- What the yaw of the vehicle does with options
defaults to front of the vehicle always heads to towards the center
option to leave yaw uncontrolled or hold a certain heading

Setting a value NAN will execute default.

Any corrections, hints or comments appreciated.

FYI @julianoes @DonLakeFlyer

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -888,7 +888,7 @@
         <param index="1">Radius of the circle in meters. positive: Orbit clockwise. negative: Orbit counter-clockwise. </param>
         <param index="2">Velocity tangential in m/s. NaN: Vehicle configuration default.</param>
         <param index="3">Yaw behaviour of the vehicle. 0: vehicle front points to the center (default). 1: Hold last heading. 2: Leave yaw uncontrolled.</param>
-        <param index="4">MAV_FRAME coordinate frame of center point.</param>
+        <param index="4">Reserved (e.g. for dynamic center beacon options)</param>
         <param index="5">Center point latitude / X coordinate. NaN: Use current vehicle position or current center if already orbiting.</param>
         <param index="6">Center point longitude / Y coordinate.</param>
         <param index="7">Center point altitude / Z coordinate.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -883,6 +883,16 @@
         <param index="6">X offset from target (m)</param>
         <param index="7">Y offset from target (m)</param>
       </entry>
+      <entry value="34" name="MAV_CMD_DO_ORBIT">
+        <description>Start orbiting on the circumference of a circle defined by the parameters. Setting any value NaN results in using defaults.</description>
+        <param index="1">Radius of the circle in meters.</param>
+        <param index="2">Velocity tangential in m/s. positive: Clockwise. negative: Counter clockwise. NaN: Vehicle configuration cruise speed clockwise.</param>
+        <param index="3">Yaw behaviour of the vehicle. 0: default vehicle front points to the center 1: Hold last heading. 2: Leave yaw uncontrolled.</param>
+        <param index="4">MAV_FRAME coordinate frame of center point.</param>
+        <param index="5">Center point latitude / X coordinate. NaN: Use current vehicle position or current center if already orbiting.</param>
+        <param index="6">Center point longitude / Y coordinate.</param>
+        <param index="7">Center point altitude / Z coordinate.</param>
+      </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI">
         <description>THIS INTERFACE IS DEPRECATED AS OF JANUARY 2018. Please use MAV_CMD_DO_SET_ROI_* messages instead. Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicles control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
         <param index="1">Region of intereset mode. (see MAV_ROI enum)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3227,9 +3227,9 @@
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
       <field type="float" name="param4">PARAM4, see MAV_CMD enum</field>
-      <field type="float" name="x">PARAM5 / local: x position, global: latitude</field>
-      <field type="float" name="y">PARAM6 / y position: global: longitude</field>
-      <field type="float" name="z">PARAM7 / z position: global: altitude (relative or absolute, depending on frame.</field>
+      <field type="float" name="x">PARAM5 / local: X coordinate, global: latitude</field>
+      <field type="float" name="y">PARAM6 / local: Y coordinate, global: longitude</field>
+      <field type="float" name="z">PARAM7 / local: Z coordinate, global: altitude (relative or absolute, depending on frame).</field>
       <extensions/>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type, see MAV_MISSION_TYPE</field>
     </message>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -885,9 +885,9 @@
       </entry>
       <entry value="34" name="MAV_CMD_DO_ORBIT">
         <description>Start orbiting on the circumference of a circle defined by the parameters. Setting any value NaN results in using defaults.</description>
-        <param index="1">Radius of the circle in meters.</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -883,6 +883,16 @@
         <param index="6">X offset from target (m)</param>
         <param index="7">Y offset from target (m)</param>
       </entry>
+      <entry value="34" name="MAV_CMD_DO_ORBIT">
+        <description>Start orbiting on a circle defined by the parameters. Setting any value NaN results in using defaults and center at current vehicle position or keeping current parameters if orbit is already running.</description>
+        <param index="1">Radius in meters.</param>
+        <param index="2">Velocity tangential clockwise in m/s. NaN: Vehicle configuration cruise speed.</param>
+        <param index="3">Center Latitude/Local X-coordinate</param>
```
amilcarlucas: Lat/lon/alt must be sent in Params 5, 6, 7 like in all other MAV_CMD messages.
Take a look at other MAV_CMD to see which is which

MaEtUgR: Thanks! That's exactly the kind of feedback I was hoping for.
I took an example from the definition of `MAV_CMD_OVERRIDE_GOTO` and used the `MAV_FRAME` which looks future proof for handling local frames to me.
What do you think now?

DonLakeFlyer:
Do you really need MAV_FRAME and support for local coords? The the speed and clockwise/counterclockwise need to be split into separate values. You need to be able to specify direction without needing to specify speed.

DonLakeFlyer:
The other problem with including MAV_FRAME in here is that it screws up the usage of this command when used from a mission which has MAV_FRAME in a different place. If you want MAV_FRAME on a COMMAND_LONG then I think you should implement a new COMMAND_LONG which support MAV_FRAME just like MISSION_ITEM does.

MaEtUgR:
@DonLakeFlyer Good insights. I would prefer to support local coordinates because I think a maneuver like Orbit is useful without GPS localization (outside of a global mission of course). Would the sign of the radius for the turning direction (like it's done in other messages that have no velocity field) be acceptable for you?

Sorry but I don't understand your second message. What needs to be met such that it's compatible to a mission? Probably not only order? You presume we need a command with more than 7 parameters and therefore another message which is compaitble with a `COMMAND_LONG` or does it need to fit a `MISSION_ITEM`? I'm confused.

DonLakeFlyer:
Can this be marked as WIP in the headers and be merged in now. I'm ready to start implementing the QGC side but need the message.

DonLakeFlyer:
There is also this thing: MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE which maybe should be removed?

DonLakeFlyer:
Also I was talking with the ArduPIlot folks about this. ORBIT ends up being inches away from the LOITER_UNLIM command. I wonder it LOITER_UNLIM is tweakable to support ORBIT. Not sure. Maybe worth a look.

MaEtUgR:
@DonLakeFlyer Thanks for your investigations. It seems to me that `MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE` is a less precise message to do the same but only in some specific case. I would also like to replace it.

I looked at `MAV_CMD_NAV_LOITER_UNLIM` before and thought about reusing it. In practise I don't think it's a good idea because:
- I don't want to screw up existing e.g. ardupilot functionality
- Right now loiter is used to just hold position right? I think it's super confusing to everyone that the same word is used for a multicopter staying where it is and a fixed wing circling a position. It should be hold when a position setpoint is kept and a FW does loiter, a MC just stays.
- The loiter command does not contain circling velocity probably because that's not useful for fixed wing but still a desired heading which is only achievable for a multicopter.

I would prefer to
- bring the dedicated message `MAV_CMD_DO_ORBIT` as WIP
- mark `MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE` as deprecated
- use `MAV_CMD_NAV_LOITER_UNLIM` in its original sense which I would rather call `HOLD`

DonLakeFlyer:
All sounds fine. I would leave LOITER_UNLIM alone though since it is already in used. Can you merge so I can start the QGC work.

MaEtUgR:
I have no merge right. Could we kindly ask @julianoes?

julianoes:
@amilcarlucas are you ok with this now?

message_definitions/v1.0/common.xml
```diff
@@ -883,6 +883,16 @@
...
       </entry>
+      <entry value="34" name="MAV_CMD_DO_ORBIT">
+        <description>WIP: Start orbiting on the circumference of a circle defined by the parameters. Setting any value NaN results in using defaults.</description>
+        <param index="1">Radius of the circle in meters. positive: Orbit clockwise. negative: Orbit counter-clockwise. </param>
+        <param index="2">Velocity tangential in m/s. NaN: Vehicle configuration default.</param>
+        <param index="3">Yaw behaviour of the vehicle. 0: vehicle front points to the center (default). 1: Hold last heading. 2: Leave yaw uncontrolled.</param>
+        <param index="4">Reserved (e.g. for dynamic center beacon options)</param>
+        <param index="5">Center point latitude / X coordinate. NaN: Use current vehicle position or current center if already orbiting.</param>
+        <param index="6">Center point longitude / Y coordinate.</param>
+        <param index="7">Center point altitude / Z coordinate.</param>
```
julianoes: Is this relative or absolute altitude?

MaEtUgR: That should be defined by `MAV_FRAME` or did I get it wrong now? @DonLakeFlyer

DonLakeFlyer: This runs into a fundamental flaw of mavlink commands, mission items and command long. A command can be used both in a mission and a command_long. WHen used as a MISSION_ITEM the MAV_FRAME is specified in the MISSION_ITEM message (external to the command params). When used in a COMMAND_LONG there is no MAV_FRAME. Hence everything gets strange. I

DonLakeFlyer: I would say just define it as one or the other for now and then you acn discuss whether you want this command to work in missions as well.

DonLakeFlyer: You can see the same exact problem with MAV_CMD_NAV_TAKEOFF. This command is used both in missions and command_long. In this case (at least for PX4) the altitude is an absolute altitude.

julianoes: Hm, interesting. I had not realized that.

DonLakeFlyer: I guess the simplest approach is to assume DO_ORBIT is for use with COMMAND_LONG only. Then @MaEtUgR's original spec would be correct. Sorry for confusing this whole thing.

julianoes: @DonLakeFlyer but don't we want COMMAND_INT in order to have good accuracy for the lat/lon?

DonLakeFlyer: Hmm, I've never looked at COMMAND_INT. QGC sends COMMAND_LONG currently for all commands. I can use COMMAND_INT for this one.

julianoes:
@LorenzMeier was this discussed in yesterday's mavlink call? Is this good to be merged?

MaEtUgR:
@DonLakeFlyer I got confused by now. Clearly you have more experience with how MAVLink is/schould be used and I was hoping you have a clear idea how to correct my initial suggestion. While I had no intention to hurry I think we need to conclude at some point.

**Accuracy**
Can we support double/integer resacled accuracy with the current framework? Otherwise we would get a relatively coarse resolution of ~2m right? I just saw that the normal mavlink command has float parameters while the message forwarding it inside the autopilot has two doubles for lat/long (https://github.com/YUNEEC/Firmware/blob/develop/msg/vehicle_command.msg#L107-L108).

**Mission/Command**
It's nice to be able to also use it inside a mission. Is that currently possbile or does it require other mavlink syntax changes?

If none of the above are possible I would opt for a WIP message we can use to support the feature now instead of waiting until all possibilities and generalizations are fixed. In that case I should probably revert https://github.com/mavlink/mavlink/pull/924/commits/4edaa1aa687b563852255b29a6b97bdffb915cea or is there other things missing?

DonLakeFlyer:
I"m not exactly quite sure what question(s) is/are waiting to be answered by me. Maybe some of this:
* QGC can use COMMAND_INT to send this message to get better accuracy
* The ORBIT messages doesn't need to include MAV_FRAME since COMMAND_INT supports that
* Whether ORBIT can be used inside a mission is a firmware question

As far as I can tell from the discussion the current state of this pull is fine.

Issue #153, opened 2013-12-03
Mavlink SERVO_OUTPUT_RAW Message not correctly defined in COMMON


PDSantos:
In the Common Mavlink Message definitions SERVO_OUTPUT_RAW uses an uint32_t to represent a microsecond time (time_usec). This seams to be wrong because in all other messages, a microseconds time is represented by an uint64_t. Also, given the range of an uint32 (0 to 4294967296), the time will overflow in approximately 71.5minutes.

Reference: https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L1193

Pedro Santos

amilcarlucas:
@LorenzMeier he has a good point here.

LorenzMeier:
This is pretty bad. I'm considering a breaking change to resolve. A time overflow could be safety critical and worth forcing an upgrade - any opinions?

WickedShell:
Depends what way you want to update it. Changing it to be millis instead of micros doesn't break parsers, it can cause some issues, but it doesn't lead to packets as being rejected. Any change to the actual message packing though I view as unacceptable. It breaks the point of a write only description, and to the best of my knowledge none of the current generators can load 2 definitions for this packet, which would cause large version issues when talking to older firmwares.

amilcarlucas:
Create a new and proper `SERVO_OUTPUT_STATUS` message, and deprecate `SERVO_OUTPUT_RAW`?

LorenzMeier:
@amilcarlucas I think that makes a lot of sense. I added a new message - please comment here: https://github.com/mavlink/mavlink/pull/926

hamishwillee:
Closing. This was superseded by https://mavlink.io/en/messages/common.html#ACTUATOR_OUTPUT_STATUS

PR #926, opened 2018-05-29
Add new PWM output message

LorenzMeier:
This new message is streamlined and ensures correct timestamping. It also has improved wording to not be specific to servos.

It fixes https://github.com/mavlink/mavlink/issues/153#issuecomment-392452592

Please comment on the fields and make suggestions for improvements.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5107,6 +5107,12 @@
       <field type="uint16_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
+    <message id="375" name="ACTUATOR_OUTPUT_STATUS">
+      <description>The raw values of the actuator outputs.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (since system boot).</field>
+      <field type="uint32_t" name="active" display="bitmask">Active outputs</field>
+      <field type="float[32]" name="actuator">Servo / motor output array values. Zero values indicate unused channels.</field>
+    </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
       <description>Cumulative distance traveled for each reported wheel.</description>


message_definitions/v1.0/common.xml
```diff
@@ -4706,5 +4706,25 @@
       <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
       <field type="uint8_t[5]" name="point_valid">States if respective point is valid (boolean)</field>
     </message>
+    <message id="340" name="PWM_OUTPUT_STATUS">
+      <description>The RAW values of the PWM outputs. The standard PPM modulation is as follows: 1000 microseconds: 0%, 2000 microseconds: 100%.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (microseconds since system boot)</field>
```
amilcarlucas: Remove "microseconds "

message_definitions/v1.0/common.xml
```diff
@@ -4706,5 +4706,25 @@
...
+      <field type="uint16_t" name="pwm1" units="us">Servo / motor output 1 value, in microseconds</field>
```
amilcarlucas: remove ", in microseconds"

amilcarlucas:
This PR should also mark the old message as deprecated

message_definitions/v1.0/common.xml
```diff
@@ -4706,5 +4706,25 @@
       <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
       <field type="uint8_t[5]" name="point_valid">States if respective point is valid (boolean)</field>
     </message>
+    <message id="340" name="PWM_OUTPUT_STATUS">
+      <description>The RAW values of the PWM outputs. The standard PPM modulation is as follows: 1000 microseconds: 0%, 2000 microseconds: 100%.</description>
```
WickedShell: We should define a no data value for all the fields (ideally 0 to leverage the trimming).

amilcarlucas: Something line "set to zero for unused channels" ?

LorenzMeier: Fair, doing that now.

hamishwillee: Done in https://github.com/mavlink/mavlink/pull/926/commits/6fd8d8bced699533c265d620183ed515f7c0b564

message_definitions/v1.0/common.xml
hamishwillee: Removed everywhere in https://github.com/mavlink/mavlink/pull/926/commits/d7310f49a1de89437813ba3fc55363f96c360466

hamishwillee:
OK, I added the zero values text and removed the units from descriptions.

Just FMI, can you confirm that
* the actual mapping of motor number here to physical pins is flight stack and hardware dependent
* why are we interested in raw PWM values? Ie these would only be useful if you know what the range is, which implies the precise servo/actuator attached.

> This PR should also mark the old message as deprecated.

Doesn't that happen after we've started implementing it in Autopilots?

hamishwillee:
Can everyone re-review and confirm they are happy for this to be merged?

message_definitions/v1.0/common.xml
```diff
@@ -4774,5 +4774,25 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="340" name="PWM_OUTPUT_STATUS">
+      <description>The RAW values of the PWM outputs. Zero values indicate unused channels. The standard PPM modulation is as follows: 1000 microseconds: 0%, 2000 microseconds: 100%.</description>
```
WickedShell: This is describing PWM not PPM channels. The standard description also doesn't map to a lot of exsisting hardware or actuators, and I'd probably say it's worth just removing that part.

message_definitions/v1.0/common.xml
```diff
@@ -4774,5 +4774,25 @@
...
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (since system boot).</field>
+      <field type="uint16_t" name="pwm1" units="us">Servo / motor output 1 value.</field>
```
WickedShell: No unused/unprovided field value?

hamishwillee: What do you mean? In description it says `Zero values indicate unused channels.`

message_definitions/v1.0/common.xml
hamishwillee: How would you re-write it?

hamishwillee:
> I know we discussed on the call making this an array instead (which allows users to easily iterate over it). Was that rejected for some reason? (I don't recall if we locked in that we wanted it or not at this point)

I don't recall the discussion. If this was last week, I wasn't there :-) FWIW it sounds like a good idea. Would unused (zero value) elements be truncated from the array? If so, that might well be another benefit.

message_definitions/v1.0/common.xml
WickedShell: So it does... Totally read over it, sorry!

message_definitions/v1.0/common.xml
WickedShell: I'd drop the entire sentence about standard PPM modulation. Probably not worth capitalzing the `RAW` part in the first sentence either.

WickedShell:
> I don't recall the discussion. If this was last week, I wasn't there :-) FWIW it sounds like a good idea. > Would unused (zero value) elements be truncated from the array? If so, that might well be another benefit.

I was suggesting it during the call since it allows GCS's/AutoPilots to easily just iterate over the array (which you can technically do by casting right now, but that's a dirty hack that only happens to work at the moment because of the packing). The zero truncation would happen either way, the main advantage of using 0 for the no value is that it works well with any future extensions.

message_definitions/v1.0/common.xml
hamishwillee: Makes sense to me. Done in https://github.com/mavlink/mavlink/pull/926/files#diff-ceacc1e2390186650146a45be8fcb691

magicrub:
Why not make them all signed and add a format enum like UAVCAN?

hamishwillee:
@magicrub Can you please respond to  https://github.com/mavlink/mavlink/pull/926#issuecomment-416446611 -

@LorenzMeier @amilcarlucas Thoughts on https://github.com/mavlink/mavlink/pull/926#issuecomment-416419336

amilcarlucas:
I vote for an array. And no enum, use different message for % if needed.
Both code and documentation are easier to maintain without having the meaning of the fields depend on some enum.

magicrub:
The UAVCAN msg I was referring to is [equipment/actuator/Command](https://github.com/UAVCAN/dsdl/blob/master/uavcan/equipment/actuator/Command.uavcan) which is which is of a single output:
```
uint8 actuator_id

# Whether the units are linear or angular depends on the actuator type.
uint8 COMMAND_TYPE_UNITLESS     = 0     # [-1, 1]
uint8 COMMAND_TYPE_POSITION     = 1     # meter or radian
uint8 COMMAND_TYPE_FORCE        = 2     # Newton or Newton metre
uint8 COMMAND_TYPE_SPEED        = 3     # meter per second or radian per second
uint8 command_type

# Value of the above type
float16 command_value

```

And then that's in a variable sized array of [equipment/actuator/ArrayCommand](https://github.com/UAVCAN/dsdl/blob/master/uavcan/equipment/actuator/1010.ArrayCommand.uavcan)
```
# Actuator commands.
# The system supports up to 256 actuators; up to 15 of them can be commanded with one message.

Command[<=15] commands

```


This allows for mixed actuator output units with variable length and since the _id is in there you can send them out-of-order or partial packets with servos that update at different rates. However, I agree it's a lot of data and in MAVLink it likely won't be used to it's potential and will just slow down the link. Also what's missing is unit type of "RC PWM" or something that would represent that 1000-2000us PWM signal that this would typically be used for in the near/medium future.

hamishwillee:
Sorry, accidental close!

hamishwillee:
@magicrub - thanks for the suggestion and explanation. I tend to agree with @amilcarlucas that having multiple messages provides a "cleaner" definition -and it is certainly easier for GCS to parse. We do have some cases where we do use the enum approach, but they tend to be more for internal protocol messages that won't ever have to be exposed via the UI.

That said, is it useful to expose the values as a percentage too?

hamishwillee:
I have made this an array now. 

1. Is there a defined mapping between index/servo numbers and ports on the device?

My assumption is that numbers 1-8 map must map to MAIN and 9-16 map to AUX (if present) - and we should say that. 

Also that if Aux is not present that does not matter because the values will be 0 (truncated).

2. Would there ever be a case where we just want to send MAIN or just AUX ports?
3. Can there ever be more than 16 ports?

If either of these are true, then we would be better off with a field for the index of the affected port and an 8-value array.

WickedShell:
> My assumption is that numbers 1-8 map must map to MAIN and 9-16 map to AUX (if present) - and we should say that.

I think this is out of scope for us to be specifying in MAVLink. The output numbering should always be considered to be an attribute of what you are talking to. The concept of main/aux outputs is not something that should be documented here, particularly if it's just a dense array of outputs. I think this meaning will always have to be specific to the autopilot you are communicating with.

0 indicating unused works, but you have an overlap then that we also send 0 as outputs when we are sending no pulses on an otherwise valid channel, and we may start sending values later. If 0 simply means unused a GCS won't be able to differentiate between not sending pulses and not a valid channel.

hamishwillee:
>> My assumption is that numbers 1-8 map must map to MAIN and 9-16 map to AUX (if present) - and we should say that.

> I think this is out of scope for us to be specifying in MAVLink. The output numbering should always be considered to be an attribute of what you are talking to. The concept of main/aux outputs is not something that should be documented here, particularly if it's just a dense array of outputs. I think this meaning will always have to be specific to the autopilot you are communicating with.

@WickedShell I'm OK with that. 

Still, this does assume a system with max 16 ports. Is it better to have 8 ports per message and have index for which array you're using?

> 0 indicating unused works, but you have an overlap then that we also send 0 as outputs when we are sending no pulses on an otherwise valid channel, and we may start sending values later. If 0 simply means unused a GCS won't be able to differentiate between not sending pulses and not a valid channel.

I guess we could add an index for the max valid array index. Or we could use UINT16_MAX for unused - but that then means we can't take advantage of zero-byte truncation. Thoughts?

message_definitions/v1.0/common.xml
```diff
@@ -5107,6 +5107,12 @@
       <field type="uint16_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
+    <message id="375" name="ACTUATOR_OUTPUT_STATUS">
+      <description>The raw values of the actuator outputs.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (since system boot).</field>
+      <field type="uint32_t" name="active" display="bitmask">Active outputs</field>
+      <field type="float[32]" name="actuator">Servo / motor output array values. Zero values indicate unused channels.</field>
```
WickedShell: Do actively disabled channels get set to NaN?

amilcarlucas: I guess that is not set by NaN, but by the "active" bitmask?

WickedShell: Gah! I was on the call and even made the suggestion then forgot about it when looking at the pushed changes... You're right!

hamishwillee:
Did we decide against adding a uint32 index to allow larger numbers of actuators?

Otherwise looks as discussed?

hamishwillee:
Replaced by #1148

PR #932, opened 2018-06-11
add CELLULAR_STATUS message to report cellular network status

chobitsfan:
Some drones use cellular network (3G, LTE, WCDMA, etc...) to communicate with GCS. This message can be sent from drone to GCS to indicate its cellular connection status. This new message is requied for celluar network becuase RADIO_STATUS has been used by existed telemetry.

I use char[5]  for the file "technology" for 3 reasons:
1. celluar network modem usually report celluar technology in string (for example, qualcomm). And it do not list all possible values in document.
2. There are many cellular technology and it is not easy to list all possible string reported by carrier and test it
3. Cellular technology is continuous advancing (for example: 5G is coming). If we use an enumeration, we need to keep it update.


message_definitions/v1.0/common.xml
```diff
@@ -4706,5 +4706,14 @@
       <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
       <field type="uint8_t[5]" name="point_valid">States if respective point is valid (boolean)</field>
     </message>
+    <message id="333" name="CELLULAR_STATUS">
+      <description>report current used cellular network status</description>
+      <field type="char[5]" name="technology">current cellular technology: 3G, HSPDA, LTE, WCDMA, etc...</field>
```
vooon: Why not use enum here?

chobitsfan: there are 2 reasons:
1. cellular network modem (for example, qualcomm) report cellular technology in string format. And it do not list all possible values in document. 
2. cellular technology is  continuous advancing and vary from country to country (for example, WCDMA is not available in Taiwan). 
If we use enum, it will be difficult to list all values and need to keep it updated.

Vavooon:
Why not to send another RADIO_STATUS message from different component id?
However I do agree that current RADIO_STATUS may be extended with additional fields like proposed (but I'd vote for using enum).

chobitsfan:
sik telemetry radio inject RADIO_STATUS to mavlink stream. If we extend/modify RADIO_STATUS definition, I think it will break something?

Vavooon:
You`re right, but we can`t add a new message every time when we just can add a field. 
You will need to rebuild SiK firmware in order to let it generate the message with updated structure.

chobitsfan:
>  we can't add a new message every time when we just can add a field.
Yes, you are right

But sik telemetry radio is very popular among ardupilot community. If we modify RADIO_STATUS, every one need to update their firmware. I am afraid that it may confuse a lot of people

chobitsfan:
OK, I change radio type to enum. Current enum values are referred from google location api

message_definitions/v1.0/common.xml
```diff
@@ -4730,5 +4737,14 @@
       <field type="float[11]" name="point_5">Depending on the type (see MAV_TRAJECTORY_REPRESENTATION)</field>
       <field type="uint8_t[5]" name="point_valid">States if respective point is valid (boolean)</field>
     </message>
+    <message id="333" name="CELLULAR_STATUS">
+      <description>report current used cellular network status</description>
+      <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">cellular network radio type: gsm, cdma, lte...</field>
+      <field type="uint8_t" name="rssi">cellular network signal RSSI, without negative sign</field>
```
amilcarlucas: What are the units here dBµV, dBi, dB, .... ?

chobitsfan: It's dBm. I  forget the unit, thank you for reminding me.

hamishwillee:
@thomasgubler Given that Auterion is providing live data over LTE (as per [product page](https://auterion.com/product/)), do you think you could seek some advice from the team on whether this message would meet the auterion use cases? For example:
- Does it cover the cellular network types you see?
- Do you need all these fields - quality (rssi/rsrp in dBm absolute), mcc, mnc, lac, cell id?
- Are there any other fields you need to know about cellular status than signal strength? 
- Might a system have multiple cellular radios? i.e. do we need a device index?
- Is a uint8_t sufficient for the range of dBm we might get?
- Is CELLULAR_STATUS a good name? Is there a better one?

Is there anyone else you might recommend for this discussion?

@chobitsfan Sorry for not engaging with this message earlier. I presume you are still interested ?

thomasgubler:
@bkueng could you have look here please. (see @hamishwillee s comment above)

hamishwillee:
Thanks @chobitsfan !  Did you see the note above from Lorenz: https://github.com/mavlink/mavlink/pull/932#pullrequestreview-184706249 ?

Hopefully @bkueng  will be able to offer some advice too. 

Note, just FYI, I am certain the message does exactly what you want and need. As a message in common.xml is hard to change after release we try to make sure that we  consider all the information people might reasonably want for similar use cases, and to future proof this a little.

chobitsfan:
Hi @LorenzMeier  thank you for commenting

I took https://play.google.com/store/apps/details?id=com.gyokovsolutions.gnettracklite as reference.  

unnamed

chobitsfan:
hi @bkueng thank you for comment

I am not sure about the meaning of "time since last connection". Would you mind further explain it? Thank you very much

hamishwillee:
> hi @bkueng thank you for comment
> I am not sure about the meaning of "time since last connection". Would you mind further explain it? 

Good question. @bkueng Obviously if you're disconnected it would be nice to know the time since messages stopped getting through ... but then how would you get the response if you're disconnected?

Below sounds good IFF it can programmatically be extracted the sending component. Otherwise the GCS has to do quite a lot of work that _should_ be easy for the LTE device.

> whether roaming is active (if that cannot already be inferred from the mobile country/network codes)
time since last connection

@LorenzMeier is there anything else you expect to be added?

message_definitions/v1.0/common.xml
```diff
@@ -3107,6 +3107,20 @@
         <description>The fields next_lat, next_lon and next_alt contain valid data.</description>
       </entry>
     </enum>
+    <enum name="CELLULAR_NETWORK_RADIO_TYPE">
+      <description>cellular network radio type</description>
```
hamishwillee: Can we please use sentence capitalisation everywhere? ie this would be:

> Cellular network radio type.

hamishwillee:
@bkueng @LorenzMeier Anything to add? Can this be pushed in?

message_definitions/v1.0/common.xml
chobitsfan: Yes. thank you

hamishwillee:
Thanks for that @chobitsfan .

What did you think about @bkueng suggestions? Specifically

* whether roaming is active (if that cannot already be inferred from the mobile country/network codes)
* amount of data uploaded/downloaded over the cellular link

chobitsfan:
Hi @hamishwillee 

> whether roaming is active (if that cannot already be inferred from the mobile country/network codes)

I added an uint16_t status bitmap, which bit 0 means roaming active

> amount of data uploaded/downloaded over the cellular link

I am not sure about this. With all due respect, maybe it is a little out of scope?

hamishwillee:
Thanks @chobitsfan - let's see if @bkueng wants to respond on the point above. 

Sorry for the delays. Personally I think this looks good to go, but I don't want to push it through without people who raised issues agreeing they are OK with the responses. 

I have added the DevCall label. We'll be meeting early Jan, so if @LorenzMeier and others have not responded by then I will suggest to the meeting that this gets merged regardless.

hamishwillee:
Thanks @bkueng , Just to be very clear, I think you are agreeing that 
- data over link is reasonably "out of scope" and we don't need time since last connection
- This is good to merge.

@chobitsfan Last thoughts 
- will every system be able to supply all the information in CELLULAR_STATUS - e.g. cell id, mnc etc? If not we will need to specify a value for "not supplied" - typically UINT16_MAX for int16 and NaN for float.
- Is the data specified sufficient to hold those values (ie is mnc always storable in a uint16 - I just don't know the format).

@auturgy @LorenzMeier @amilcarlucas Anything to add? If not, I propose to merge this at the end of the week or Monday.

chobitsfan:
Hi @bkueng , Does "time since last connection" means "time since last connection with GCS" or "time since last connection with cellular base station"? Thank you.

Hi @hamishwillee, thank you for commenting
> will every system be able to supply all the information in CELLULAR_STATUS - e.g. cell id, mnc etc? If not we will need to specify a value for "not supplied" - typically UINT16_MAX for int16 and NaN for float.

I agree wit you. I will add uint32_max for cell id, unit16_max for mcc and mnc, 0 for lac to indicate "not supplied"

 > Is the data specified sufficient to hold those values (ie is mnc always storable in a uint16 - I just don't know the format).

mnc and mcc is 3 decimal digits (0~999) , lac is 16 bit , cell id is 32 bit. I think it is sufficient for them

hamishwillee:
> I agree wit you. I will add uint32_max for cell id, unit16_max for mcc and mnc, 0 for lac to indicate "not supplied"

Thanks @chobitsfan . Looks good. I presume you're using unit16_max because 0 is a valid value for the other cases?

> @bkueng Does "time since last connection" means "time since last connection with GCS" or "time since last connection with cellular base station"? Thank you.

I think he meant "since last connection with GCS" . I don't think it was fully thought out.

I'm good with this, but we can't submit until some problems with CI are fixed.

chobitsfan:
Hi @hamishwillee 

> I presume you're using unit16_max because 0 is a valid value for the other cases?

Yes, that is correct.

>  but we can't submit until some problems with CI are fixed.

I check the ci build log, build in python 2.7 and 3.5 is passed. but failed in python 3.3

Collecting lxml>=3.6.0 (from -r pymavlink/requirements.txt (line 1))
    Complete output from command python setup.py egg_info:
    This lxml version requires Python 2.7, 3.4 or later.    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-0lr9e7/lxml/

hamishwillee:
Yes, I know! I've created  a fix in #1056 - just waiting on a sanity check to removal of Python 3.3 from someone else in the team (It makes perfect sense to me, but best to get checked).

hamishwillee:
@LorenzMeier The comment you made was addressed. Beat checked it. This appears ready to merge. I plan to do so very soon.

message_definitions/v1.0/common.xml
```diff
@@ -4967,6 +4981,16 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="334" name="CELLULAR_STATUS">
+      <description>Report current used cellular network status</description>
```
LorenzMeier: @hamishwillee Let's flag this WIP and get it in so we get some real usage. I think otherwise it's impossible to settle the discussion for good.

@RomanBapst @julianoes We should look at how we can add support for this in mavlink-router for LTE (not a priority, but should be in the architecture backlog and should be logged by the autopilot in ulog).

hamishwillee:
@chobitsfan Last change, as requested by Lorenz in https://github.com/mavlink/mavlink/pull/932#discussion_r246534988

Can you please add the [`<wip />`](https://mavlink.io/en/guide/xml_schema.html#wip) tag immediately below the message tag? 
This means "work in progress" and means that we can still (easily) iterate this if we find we missed something obvious during implementation. As soon as that is done I will merge this.

chobitsfan:
> Can you please add the <wip /> tag immediately below the message tag?

Yes, thank you everyone for comments and suggestions.

PR #939, opened 2018-06-26
Add stamped commands to ASLUAV

acfloria:
Add a stamped version of the commands to test the command handling system.


dagar:
Merged as requested. It would still be great to promote a version of this to common if we can reach consensus on https://github.com/mavlink/rfcs/pull/9.

PR #940, opened 2018-06-29
Add a MAV_CMD version of MISSION_SET_CURRENT

acfloria:
This is a command version of MISSION_SET_CURRENT. This is useful as one can make sure with an acknowledge that the command arrived on the vehicle and is accepted. MAV_CMD_MISSION_START can not be used as that one is supposed also to start the mission.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1411,6 +1411,17 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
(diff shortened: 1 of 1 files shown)


hamishwillee:
FMI Possibly a stupid question, but is the acknowledgment that is expected "standardised"? For example, expected response if there is no current mission, or if the mission number used is outside the total number? Does it matter if the mission item is not a waypoint?

acfloria:
I think the standard is that if the index could be set an acknowledge with MAV_RESULT_ACCEPTED is answered. If it failed then an ack with MAV_RESULT_FAILED is sent.

PR #944, opened 2018-07-06
common.xml: support absolute mount/gimbal angles

julianoes:
This change 1) clearly specifies the frame of an angle to set (body or
absolute) and 2) adds support to set absolute angles.

This is as suggested by @bkueng.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1288,9 +1288,9 @@
         <param index="2">stabilize roll? (1 = yes, 0 = no)</param>
         <param index="3">stabilize pitch? (1 = yes, 0 = no)</param>
         <param index="4">stabilize yaw? (1 = yes, 0 = no)</param>
-        <param index="5">roll input (0 = angle, 1 = angular rate)</param>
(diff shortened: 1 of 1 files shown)


bkueng:
I'm wondering if we can just set `stabilize yaw` to 1, then we should achieve the same effect, since it substracts the vehicle's yaw.

julianoes:
@bkueng fair point but we should somehow document that in the spec so it's clear how to use it.

hamishwillee:
Probably a stupid question, but can't an angular rate (1 = angular rate) also be respective to a particular frame.

julianoes:
> Probably a stupid question, but can't an angular rate (1 = angular rate) also be respective to a particular frame.

Yes you would need a frame but the frame is actually the same for all and we only change the reference for the angle (body or absolute). For angular rate you don't need a reference, or rather the reference is 0 velocity.

julianoes:
@bkueng now that I think about this: shouldn't the stabilize flags be parameters instead of command params? Basically, the stabilize flags depend on the drone and how the gimbal works. A ground station should not need to know that and just be able to set the angle without having to worry about that.

PR #945, opened 2018-07-11
Message markdown improvements

hamishwillee:
This improves the rendering of message output:
1. Changes column heading for enum value to "Value" (from command ID) - both HTML and markdown
2. Moves field units and enums from description into their own column. BUT the columns are only created if some field in the message has the unit/enum.


hamishwillee:
Next step would be to pull units and enums out of the actual message definition description in common.xml.

LorenzMeier:
Thanks!

PR #914, opened 2018-05-17
ardupilotmega.xml: remove redundant units information

amilcarlucas:

PR #948, opened 2018-07-13
common.xml: remove redundant units info 

hamishwillee:
Like https://github.com/mavlink/mavlink/pull/914

@amilcarlucas 
A couple of questions/statements:
1. A lot of descriptions also include the name of the ENUM, which is also captured in the enum= field. I propose we remove the description from the ENUM?
1. Percents are stored a uint8s, but only 0 - 100 is used. We should make this clear in a common way.
   How about from:
   ```
   <field type="uint8_t" name="rssi" units="%">Receive signal strength indicator, 0: 0%, 100: 100%, 255: invalid/unknown.</field>
   ```
   TO
   ```
   <field type="uint8_t" name="rssi" units="%">Receive signal strength indicator. Values: [0-100], 255: invalid/unknown.</field>
   ```
1. There are a whole bunch of variations of "Timestamp (time since system boot or since UNIX epoch)". How about we make this consistent:
   * Timestamp (since system boot)
   * Timestamp (UNIX Epoch time)
   * Timestamp. (UNIX Epoch time or time since system boot).


LorenzMeier:
@hamishwillee Couple clarifications:

Enums: Please remove redundant information

Percent clarification: Looks good.

Timestamp: All new messages should use:

  * time_usec Timestamp. (UNIX Epoch time or time since system boot). The receiving end can infer the used timestamp (since 1.1.1970 or since boot in 2018) by checking for the magnitude the number.

For old messages:

  * time_boot_ms Timestamp: Milliseconds since system boot (always)

hamishwillee:
But a bunch of old messages say that they support epoch. And where is
old/new cutoff?

Can't all messages support the new format?

amilcarlucas:
Ardupilot does not support the 2018 stuff.

message_definitions/v1.0/common.xml
```diff
@@ -1526,7 +1526,7 @@
       <entry value="511" name="MAV_CMD_SET_MESSAGE_INTERVAL">
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM</description>
         <param index="1">The MAVLink message ID</param>
-        <param index="2">The interval between two messages, in microseconds. Set to -1 to disable and 0 to request default rate.</param>
```
amilcarlucas: I would really, really, like that the PR ardupilot/pymavlink#7 would be discussed in the Mavlink meetings and be implemented ASAP

message_definitions/v1.0/common.xml
```diff
@@ -3175,39 +3175,39 @@
...
-      <field type="uint16_t" name="servo13_raw" units="us">Servo output 13 value, in microseconds</field>
-      <field type="uint16_t" name="servo14_raw" units="us">Servo output 14 value, in microseconds</field>
-      <field type="uint16_t" name="servo15_raw" units="us">Servo output 15 value, in microseconds</field>
-      <field type="uint16_t" name="servo16_raw" units="us">Servo output 16 value, in microseconds</field>
+      <field type="uint16_t" name="servo9_raw" units="us">Servo output 9 value</field>
+      <field type="uint16_t" name="servo10_raw" units="us">Servo output 10 value</field>
+      <field type="uint16_t" name="servo11_raw" units="us">Servo output 11 value</field>
+      <field type="uint16_t" name="servo12_raw" units="us">Servo output 12 value</field>
+      <field type="uint16_t" name="servo13_raw" units="us">Servo output 13 value</field>
+      <field type="uint16_t" name="servo14_raw" units="us">Servo output 14 values</field>
```
amilcarlucas: One "s" too many

LorenzMeier:
@amilcarlucas I doubt that, because you don't have a system time source onboard. I think this is a misunderstanding.

@hamishwillee It's very simple: If the microseconds in the timestamp amount to less than 40 years, it is since boot. If its more than 40 years (= 40 * 365 * 24 * 60 * 60 * 1000 * 1000) then its obviously since UNIX epoch.

message_definitions/v1.0/common.xml
```diff
@@ -4164,18 +4164,18 @@
     </message>
     <message id="143" name="SCALED_PRESSURE3">
       <description>Barometer readings for 3rd barometer</description>
-      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (milliseconds since system boot)</field>
-      <field type="float" name="press_abs" units="hPa">Absolute pressure (hectopascal)</field>
-      <field type="float" name="press_diff" units="hPa">Differential pressure 1 (hectopascal)</field>
-      <field type="int16_t" name="temperature" units="cdegC">Temperature measurement (0.01 degrees celsius)</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot)</field>
+      <field type="float" name="press_abs" units="hPa">Absolute pressure</field>
+      <field type="float" name="press_diff" units="hPa">Differential pressure 1</field>
```
amilcarlucas: Remove the 1

message_definitions/v1.0/common.xml
```diff
@@ -4164,18 +4164,18 @@
...
       <description>current motion information from a designated system</description>
-      <field type="uint64_t" name="timestamp" units="ms">Timestamp in milliseconds since system boot</field>
+      <field type="uint64_t" name="timestamp" units="ms">Timestamp (time since system boot)</field>
       <field type="uint8_t" name="est_capabilities">bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT + RATES = 3)</field>
-      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84), in degrees * 1E7</field>
-      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84), in degrees * 1E7</field>
-      <field type="float" name="alt" units="m">AMSL, in meters</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">AMSL</field>
```
amilcarlucas: AMSL altitude

message_definitions/v1.0/common.xml
```diff
@@ -4208,14 +4208,14 @@
...
-      <field type="int16_t" name="current_battery" units="cA">Battery current, in 10*milliamperes (1 = 10 milliampere), -1: autopilot does not measure the current</field>
-      <field type="int32_t" name="current_consumed" units="mAh">Consumed charge, in milliampere hours (1 = 1 mAh), -1: autopilot does not provide mAh consumption estimate</field>
-      <field type="int32_t" name="energy_consumed" units="hJ">Consumed energy, in HectoJoules (intergrated U*I*dt)  (1 = 100 Joule), -1: autopilot does not provide energy consumption estimate</field>
+      <field type="int16_t" name="temperature" units="cdegC">Temperature of the battery. INT16_MAX for unknown temperature.</field>
+      <field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells. Cells above the valid cell count for this battery should have the UINT16_MAX value.</field>
+      <field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
+      <field type="int32_t" name="current_consumed" units="mAh">Consumed charge, -1: autopilot does not provide mAh consumption estimate</field>
```
amilcarlucas: Replace "mAh" with "charge"

message_definitions/v1.0/common.xml
```diff
@@ -4505,75 +4505,75 @@
...
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot)</field>
+      <field type="uint64_t" name="time_utc" units="us">Timestamp (time since UNIX epoch) in UTC. 0 for unknown.</field>
       <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
-      <field type="int32_t" name="lat" units="degE7">Latitude, expressed as degrees * 1E7 where image was taken</field>
-      <field type="int32_t" name="lon" units="degE7">Longitude, expressed as degrees * 1E7 where capture was taken</field>
-      <field type="int32_t" name="alt" units="m">Altitude in meters, expressed as * 1E3 (AMSL, not WGS84) where image was taken</field>
-      <field type="int32_t" name="relative_alt" units="m">Altitude above ground in meters, expressed as * 1E3 where image was taken</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude where image was taken</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude where capture was taken</field>
+      <field type="int32_t" name="alt" units="m">Altitude (AMSL, not WGS84) where image was taken</field>
```
amilcarlucas: This one is probably millimeters, not meters

message_definitions/v1.0/common.xml
```diff
@@ -4505,75 +4505,75 @@
...
+      <field type="int32_t" name="relative_alt" units="m">Altitude above ground</field>
```
amilcarlucas: Again mm

message_definitions/v1.0/common.xml
hamishwillee: 1. I have added to the mavlink meeting [here](https://github.com/mavlink/mavlink/wiki/20180725-Weekly-Meeting) BUT that is a fairly "busy" PR. Can we get a summary post at the end of it that explains the specific points remaining for agreement/implications of those changes?
2. I did not intend to make this change. Apologies - my brain was glazing over by the end of this. Will revert.

message_definitions/v1.0/common.xml
hamishwillee: Good. That was there originally, but didn't make sense.

message_definitions/v1.0/common.xml
hamishwillee: Thanks. Might make this format global.

message_definitions/v1.0/common.xml
hamishwillee: Just removed the word altogether.

message_definitions/v1.0/common.xml
hamishwillee: Yes - this is the sort of thing that leads to confusion. Fixed.

hamishwillee: Note, I removed (and will remove globally) the "not WGS84" clarification - so "AMSL altitude where image was taken".

message_definitions/v1.0/common.xml
```diff
@@ -3962,85 +3962,85 @@
...
+      <field type="int16_t" name="ygyro" units="mrad/s">Angular speed around Y axis</field>
+      <field type="int16_t" name="zgyro" units="mrad/s">Angular speed around Z axis</field>
+      <field type="int16_t" name="xmag" units="mT">X Magnetic field</field>
+      <field type="int16_t" name="ymag" units="mT">Y Magnetic field</field>
+      <field type="int16_t" name="zmag" units="mT">Z Magnetic field</field>
     </message>
     <message id="130" name="DATA_TRANSMISSION_HANDSHAKE">
-      <field type="uint8_t" name="type">type of requested/acknowledged data (as defined in ENUM DATA_TYPES in mavlink/include/mavlink_types.h)</field>
-      <field type="uint32_t" name="size" units="bytes">total data size in bytes (set on ACK only)</field>
+      <field type="uint8_t" name="type">Type of requested/acknowledged data (as defined in ENUM DATA_TYPES in mavlink/include/mavlink_types.h)</field>
```
hamishwillee: @amilcarlucas This does not have the enum= - it should right?

message_definitions/v1.0/common.xml
```diff
@@ -3665,138 +3665,138 @@
       <field type="float" name="aux2">Aux 2, -1 .. 1</field>
       <field type="float" name="aux3">Aux 3, -1 .. 1</field>
       <field type="float" name="aux4">Aux 4, -1 .. 1</field>
-      <field type="uint8_t" name="mode" enum="MAV_MODE">System mode (MAV_MODE)</field>
+      <field type="uint8_t" name="mode" enum="MAV_MODE">System mode.</field>
       <field type="uint8_t" name="nav_mode">Navigation mode (MAV_NAV_MODE)</field>
```
hamishwillee: @amilcarlucas Should this be an enum? If so, which one - MAV_NAV_MODE is not found.

message_definitions/v1.0/common.xml
```diff
@@ -2957,13 +2957,13 @@
     </message>
     <message id="2" name="SYSTEM_TIME">
       <description>The system time is the time of the master clock, typically the computer clock of the main onboard computer.</description>
-      <field type="uint64_t" name="time_unix_usec" units="us">Timestamp of the master clock in microseconds since UNIX epoch.</field>
-      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp of the component clock since boot time in milliseconds.</field>
+      <field type="uint64_t" name="time_unix_usec" units="us">Timestamp of the master clock since UNIX epoch.</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp of the master clock since system boot time.</field>
```
hamishwillee: @LorenzMeier @amilcarlucas Is this correct? "Component" clock was not defined - so I assume this is just a case where the message provides separate fields for Unix vs system boot timestamps, both from the same source (typically)

message_definitions/v1.0/common.xml
```diff
@@ -2957,13 +2957,13 @@
...
     </message>
     <!-- FIXME to be removed / merged with SYSTEM_TIME -->
     <message id="4" name="PING">
       <description>A ping message either requesting or responding to a ping. This allows to measure the system latencies, including serial port, radio modem and UDP connections.</description>
-      <field type="uint64_t" name="time_usec" units="us">Unix timestamp in microseconds or since system boot if smaller than MAVLink epoch (1.1.2009)</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
```
hamishwillee: @LorenzMeier @amilcarlucas This one referred to MAVLink epoch. I THINK that this was just in order to provide a good date for evaluating the magnitude. Can you confirm change is OK?

message_definitions/v1.0/common.xml
```diff
@@ -3846,72 +3846,72 @@
     </message>
     <message id="112" name="CAMERA_TRIGGER">
       <description>Camera-IMU triggering and synchronisation message.</description>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp for the image frame in microseconds</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp for image frame (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
```
hamishwillee: Just FYI, this still includes "for image frame" in the timestamp description because it isn't otherwise clear.

hamishwillee:
I've made a bunch of changes as discussed. These are broken into individual commits to make them easier to review.

Re Timestamps

-  Just FYI @LorenzMeier I see now that you have a consistent naming for "old" (time_boot_ms ) and new (time_usec) - I had assumed all timestamps had the same name, so my misunderstanding was how I should differentiate them (not the actual numbers)

- The 2018 confused me a bit too at first.  It is not necessary, because of course next year the boot will be 2019 ... I used a slightly different description:

> Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.

message_definitions/v1.0/common.xml
amilcarlucas: yes, please add it

message_definitions/v1.0/common.xml
amilcarlucas: I think you will need to do git blame on this and ask the original author

message_definitions/v1.0/common.xml
amilcarlucas: I would remove the "of the master clock" altogether

message_definitions/v1.0/common.xml
amilcarlucas: I like the new wording. And sure hope there is no 2009 hardcoded somewhere.

message_definitions/v1.0/common.xml
hamishwillee: @dagar Can you help on this one - ie what is MAV_NAV_MODE supposed to be?

dagar: Good question, PX4 doesn't seem to touch that field.

hamishwillee: @amilcarlucas @dagar There is something called NAV_MODE in the [SYSTEM_STATUS](https://github.com/hamishwillee/mavlink/blob/master/message_definitions/v1.0/common.xml#L2968) description which might be relevant. However it is then not referenced.

Does that help at all? If not, I propose we ignore that, and get the other changes merged.

hamishwillee:
@LorenzMeier @amilcarlucas Can we merge this? There are a couple of open questions, but IMO those can either be deferred or ignored (they are not the main point of this PR).

amilcarlucas:
Yes, please merge.

PR #952, opened 2018-07-19
Remove units from bitmask descriptions

hamishwillee:
Remove a few additional enums in descriptions. Follows on from #948


LorenzMeier:
Thanks!

PR #953, opened 2018-07-19
Update MAV_SENSOR_ORIENTATION for more entries

WickedShell:
Replaces the value of 38 to match the ArduPilot usage, which is the only current user that I could find of the enum. Tridge and I looked recently and couldn't find anyone using `MAV_SENSOR_ROTATION_ROLL_315_PITCH_315_YAW_315` in production publicly. If we can change this value to match the current ArduPilot map of rotations that would be quite useful as there are a couple of places we are sending the internal rotations out via MAVLink at the moment, and I'd like to be able to tag them as the enum (and allow GCS's to have proper documentation on the meanings).

I'd also like to note that this enum violates the normal naming rules we have. The enum is called `MAV_SENSOR_ORIENTATION` while the entries are all `MAV_SENSOR_ROTATION_`

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2405,8 +2405,17 @@
       <entry value="37" name="MAV_SENSOR_ROTATION_ROLL_90_YAW_270">
         <description>Roll: 90, Pitch: 0, Yaw: 270</description>
       </entry>
-      <entry value="38" name="MAV_SENSOR_ROTATION_ROLL_315_PITCH_315_YAW_315">
-        <description>Roll: 315, Pitch: 315, Yaw: 315</description>
+      <entry value="38" name="MAV_SENSOR_ROTATION_ROLL_90_PITCH_68_YAW_293">
+        <description>Roll: 90, Pitch: 68, Yaw: 293</description>
+      </entry>
+      <entry value="39" name="MAV_SENSOR_ROTATION_PITCH_315">
+        <description>Pitch: 315</description>
+      </entry>
+      <entry value="40" name="MAV_SENSOR_ROTATION_ROLL_90_PITCH_315">
+        <description>Roll: 90, Pitch: 315</description>
+      </entry>
+      <entry value="100" name="MAV_SENSOR_ROTATION_CUSTOM">
+        <description>Custom orientation</description>
       </entry>
     </enum>
     <enum name="MAV_PROTOCOL_CAPABILITY">


hamishwillee:
Are you saying that you want to use these enum values because they map to your internal definitions for those orientations/rotations? ie if you were to not do this, you'd have to do code to map the enums to particular orientations?

> I'd also like to note that this enum violates the normal naming rules we have. The enum is called MAV_SENSOR_ORIENTATION while the entries are all MAV_SENSOR_ROTATION_

- Where are those naming rules defined? (ie should I add this to my list?)
- Can the enum name be changed now? I presume not, without breaking any clients of the existing API?

WickedShell:
>Are you saying that you want to use these enum values because they map to your internal definitions for those orientations/rotations? ie if you were to not do this, you'd have to do code to map the enums to particular orientations?

The onboard compass orientation code now estimates the orientation of the compasses and can correct them for you. The main mechanism for informing the GCS of this is a MAVLink extension on the `MAG_CAL_REPORT` (see ardupilot.xml). At the moment the vehicle is sending the internal rotation number which matches the parameter used in the settings/setup page. I'd like to be able to tag this as being from the `MAV_SENSOR_ENUM` but currently value 38 conflicts. We could potentially special case the value 38 (and up), however I'm loathe to do this as a GCS author may choose to just consult the XML documentation for the parameters when interpreting this number which could lead to subtle errors. (In general I just think if we can keep these matched that it would be a lot better, I do get that the XML has been deployed with the current value for quite awhile, but from looking at the original commit I think it was just intended to be the last/maximum orientation you could get to with only 45 degree increments)

> - Where are those naming rules defined? (ie should I add this to my list?)
> - Can the enum name be changed now? I presume not, without breaking any clients of the existing API?

This is mostly just a manually enforced convention, mainly needed for the C/C++ generated files, it provides easy clarity what enum one is referring to without having to manually type hint it every time to avoid unintended affects.

WickedShell:
> - Can the enum name be changed now? I presume not, without breaking any clients of the existing API?

If anyone has it in code then changing the name would be an issue. With that being said a quick git grep of PX4/firmware only shows it in comments (and a TODO to actually use it), ArduPilot also only has it in comments at the moment so we could possibly change `MAV_SENSOR_ORIENTATION` to `MAV_SENSOR_ROTATION` without any conflicts.

WickedShell:
I also just found that ArduPilot already does emit our internal rotation definitions directly in `DISTANCE_SENSOR` to the GCS, so we are already sending incorrect data. If no one else is using it then I think canonicalizing on the deployed one is correct. (And if we do that I will probably go ahead and add a giant comment in the rotations definition header of ArduPilot to tell people to ensure that any additions match/make it into MAVLink)

hamishwillee:
This seems reasonable to me and you've done your due diligence. But I feel uncomfortable that we've we don't have a better process for managing/notifying of potentially breaking changes like this. Another good reason to get versioned releases in place. 

We should also work on a notification process.

@LorenzMeier Thoughts?

LorenzMeier:
There is a version field that we used for this purpose. I think it’s time to revive the usage of that. It was intended as sub-version field for potentially breaking changes. We could try to build a dictionary of affected messages.

hamishwillee:
> There is a version field that we used for this purpose. I think it’s time to revive the usage of that. It was intended as sub-version field for potentially breaking changes. 

You mean `version` here? 
```
<mavlink>
  <version>3</version>
  <dialect>0</dialect>
```
Sounds good, but can we also update it for new messages (non-breaking) and make an an associated tag release on github?. This will make it easier for people to track changes, and will also give us a place to put that dictionary of messages. 

If we're all in agreement, presumably you @WickedShell can wait for next Dev meeting to confirm process for this to merge?

WickedShell:
I was under the impression that the version there was to support major protocol revisions (IE MAVLink 0.9/1/2). This goes into the `uint8_t_mavlink_version` which as an implementer I have to say I really really don't like. More importantly though this is uint8_t, so we only have a limited number of times we can increase it)

hamishwillee:
> I was under the impression that the version there was to support major protocol revisions (IE MAVLink 0.9/1/2). This goes into the uint8_t_mavlink_version which as an implementer I have to say I really really don't like. More importantly though this is uint8_t, so we only have a limited number of times we can increase it)

It is possible @LorenzMeier is talking about some other number. However I do know that this is what is used for the minor revision in the [docs](https://mavlink.io/en/messages/common.html#mavlink-protocol-version) - currently 3. 

The fact it can only iterate to 256 is certainly a problem. 

How would you solve this if you were all powerful?

hamishwillee:
@WickedShell Please update branch.

LorenzMeier:
Thanks, I rebased and yes, let's get that in.

PR #957, opened 2018-07-26
Update .gitmodules to use ardupilot/pymavlink

hamishwillee:
@LorenzMeier Is there anything else that needs to be done to switch to the ardupilot pymavlink? (EDITED, obviously there is, because this failed).

- I don't see a changelist entered, so I assume it just takes the latest?
- What tests need to be run?

It may be that you prefer to manage the switch yourself. If so, can you close an point me to item I can use to track it?


LorenzMeier:
It needs an upstream PR to be merged - so that CI is passing.

LorenzMeier:
@hamishwillee Can you use the latest head from upstream *now*? It should pass CI then.

LorenzMeier:
@hamishwillee I think the commit hash you're using isn't quite right. The repo URL and the commit hash have to match.

hamishwillee:
@LorenzMeier Clearly I don't know what I am doing. I am going to close this, and watch you do it properly.

PR #958, opened 2018-07-26
Update mavlink_gitboot to remove APM reference

khancyr:


hamishwillee:
@LorenzMeier This is aligned. It may of course change.

PR #964, opened 2018-07-31
pymavlink update

amilcarlucas:


amilcarlucas:
OK this fixes master by pointing at a pymavlink branch that contains ArduPilot/pymavlink#194

dagar:
Thanks @amilcarlucas this is a good immediate solution, but if/when that exact pymavlink commit disappears later (branch deleted) then mavlink at this point will be broken.

This tends to bite adopters that have grabbed mavlink at some arbitrary point.

amilcarlucas:
I know that this is not a good solution. But it does get master unstuck.

PR #966, opened 2018-08-02
Un-deprecate messages actively used by ardupilot

amilcarlucas:
Otherwise the documentation will mislead the ardupilot users


hamishwillee:
@amilcarlucas 
FMI, do ArduPilot intend to move to MESSAGE_INTERVAL at some point? (and if so when)? 

Yes, having deprecation is confusing for users of systems that have not implemented the new message. We need to align better and make sure that this doesn't happen - discussion in https://github.com/mavlink/mavlink/pull/967#issuecomment-410088116

What is missing is that we have a new message we'd like people to be able to use on systems where it is implemented, but there is no pointer to the recommended alternative.

Short term, I think we should remove this (as you have done) but also add a note about the alternative:

> `MESSAGE_INTERVAL` is used instead of this message on some systems (e.g. PX4).

Going forward can you think of a good mechanism to indicate that there is an alternative message that may be implemented on some systems. I would like to be able to imply that the alternative message is recommended (ie that we are on the path to deprecation of current message).

Slightly outside the above discussion, I think the root cause of issues like this is that there is no way to tell if any message has been implemented on a particular system. I'm not sure how we'd maintain the list properly, but having something like `<supportedby type="MAV_AUTOPILOT_ARDUPILOTMEGA" />` 
 (and not supported by could be useful).

hamishwillee:
@amilcarlucas Thanks for clarifications. Then for now I think we should remove this (as you have done) but also add a note about the alternative:

>MESSAGE_INTERVAL is used instead of this message on some systems (e.g. PX4).

Can you make that change and rebase?

@LorenzMeier , anyone else got any objections to undeprecating the MAV_DATA_STREAM message? Let's discuss in dev meeting.

hamishwillee:
ALso, perhaps "not-supported-by" would be less controversial? If applied to a message that replaces a deprecated message, at least you'd know you couldn't use it on a particular target. Also the absence of such a message does not necessarily imply that something "is supported by".

amilcarlucas:
I like the idea of not-supported-by.... and one could use the dialect enums to specify the autopilot type.

hamishwillee:
Apparently these will be supported by ArduPilot in the coming weeks, so opinion in dev meeting was probably not worth the churn. So propose close. OK?

hamishwillee:
> I like the idea of not-supported-by.... and one could use the dialect enums to specify the autopilot type.

Only problem is that sometime support is at a vehicle type level. However this might be a good way to indicate "not supported at all".

Raised issue for discussion in:  https://github.com/mavlink/mavlink/issues/973

PR #967, opened 2018-08-02
Proper deprecation of DO_DIGICAM_CONFIGURE and DO_DIGICAM_CONTROL

amilcarlucas:


hamishwillee:
Looks good to me. 

@amilcarlucas Am I correct that a deprecated message is always built by the toolchain - ie marking as deprecated does not change generator output? 

If so, then in this case the field is already deprecated, and all we are doing is marking it "properly" as such - no adverse impact on users?

If however a message was "newly"  being deprecated what would be the process? 
Basically I'm trying to work out what I would need to do in order to be sure I could reasonably merge such a change.
- Who are stakeholders?
- What impact would this have on ArduPilot or other users (say)
- Would I have to update a version of any kind/notify people specially?
- After deprecation, what is the next step? Would I ever delete a message from common.xml? 

Yes, I am aware I am moving back into the version discussion. I'm not sure it was finished

amilcarlucas:
The idea is that now that the XML tags are in both upstream and downstream mavlink repositories, we can change the generator to not build wip messages per default.
We can also change it to stop building messages that have been deprecated for more than 4 years.

Once that is done we can also change the documentation generator to add warnings to deprecated and wip messages.

In the future marking messages as deprecated needs to be coordinated between all autopilot dialects. A message can only be deprecated once _ALL_ autopilot agree that it can/should be deprecated.

As for leaving the wip state I would advise that it can only be done once the message has been approved by at least 6 reviewers.

hamishwillee:
Hi @amilcarlucas 
I think this change is a no brainer - assigned to Lorenz to approve though.

Thanks for explanations. That is pretty much what I understood. 

FWIW I'm not sure that marking messages as deprecated must be based purely on the agreement of "all" dialects (a bit of a recipe for freezing evolution). We should have clear rules and a path for discussion/voting so that no engaged community member gets this happen by surprise.

auturgy:
Don't deprecate this message.  It is used in production, and transition to the new message is not necessary for working devices.

amilcarlucas:
I like the rationale behind it. But it does make things more complex

hamishwillee:
> I like the rationale behind it. But it does make things more complex

Are you talking about this text change or embedding the concept of fallback into the toolchain?

Can you make the change to the text? It is supported by the people who attended the dev call.

Actually supporting the concept of fallback into the toolchain would be more complex. And yes, using this fallback as a mechanism for avoiding deprecation is poor. Want to propose an acceptable-to-all-parties alternative?

auturgy:
My understanding was that rather than tag as deprecated, it would be a documentation note.

Regards,

James

hamishwillee:
@auturgy That is correct. The request is to update the text as in https://github.com/mavlink/mavlink/pull/967#issuecomment-423009298. 

The rest of the discussion was an "aside" with @amilcarlucas about whether the concept of fallback has any use as part of the toolchain. Probably shouldn't have mixed that into this discussion.

amilcarlucas:
I was talking of both "embedding the concept of fallback into the toolchain" and adding the process documentation explaining the "fallback" procedures.

hamishwillee:
@amilcarlucas Fair enough. So can you make the change to the text now? Let's discuss the concept of fallback in the infrastructure as a separate issue. Because the one thing we know now is that the current text is wrong and the new text is "approved" by the stakeholders who attended the dev meeting.

hamishwillee:
@amilcarlucas Ping! https://github.com/mavlink/mavlink/pull/967#issuecomment-429141628

amilcarlucas:
created a new one #1036

hamishwillee:
Closing. Superseded by https://github.com/mavlink/mavlink/pull/1036

PR #970, opened 2018-08-03
RC_CHANNELS_OVERRIDE: ignore value fix

MaEtUgR:
Every channel should have the same definition. Ignoring with 0 is just there because it's an extension and extensions are filled with zero if they are not supported. UINT16_MAX is IMHO the far better way to skip the field because it's unambiguous and cannot be set by accident.

FYI @julianoes

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3475,7 +3475,7 @@
       <field type="uint16_t" name="chan7_raw" units="us">RC channel 7 value. A value of UINT16_MAX means to ignore this field.</field>
       <field type="uint16_t" name="chan8_raw" units="us">RC channel 8 value. A value of UINT16_MAX means to ignore this field.</field>
       <extensions/>
-      <field type="uint16_t" name="chan9_raw" units="us">RC channel 9 value. A value of 0 means to ignore this field.</field>
+      <field type="uint16_t" name="chan9_raw" units="us">RC channel 9 value. A value of 0 or UINT16_MAX means to ignore this field.</field>
       <field type="uint16_t" name="chan10_raw" units="us">RC channel 10 value. A value of 0 or UINT16_MAX means to ignore this field.</field>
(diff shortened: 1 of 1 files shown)


WickedShell:
A slight tangent I'd like to discuss as well, there are 3 states I'd like to be able to do with the RC_CHANNELS_OVERRIDE message per channel.

1. Set the channel to the override value
2. Ignore the RC_CHANNELS_OVERRIDE message and don't update the value on the channel
3. Return control to the RC receiver physically connected

1 is what everyone does with the message already, and per the spec number 2 can be achieved by sending UINT16_MAX. I'd loosely like to propose that the value of 0 means return control to the RC receiver, but I haven't given it a lot of thought, and it means that the upper 8 channels would always be under RC control unless you were sending it as MAVLink 2 with the extensions.

MaEtUgR:
@WickedShell Don't you automatically take the RC receivers values if the override doesn't provide any? Ah it's per channel and you want to mix controls up between override and not? You would like to do that to the channels 1-8 as well then right?

In our use case we don't use 1000-2000 values but rather 0-4095... so zero could be a value and we have to be careful with the channels 9-18.

Wouldn't it be possible to use something like UINT16_MAX - 1 to mix in the receivers value for that channel?

julianoes:
I made this change initially:
https://github.com/mavlink/mavlink/commit/101cca833dc0727f38b455d1e61c893b1b9cf4a8

I said it was for compatibility with mavlink 1 but I'm not sure about what that exactly means anymore.

What is for sure is that we can save bandwidth by not having to set all channels to `UINT16_MAX` but can let them 0 and have them trimmed.

MaEtUgR:
Is the particular change in this pr still up to discussion? We could open an issue for any further concerns on a similar topic...

hamishwillee:
Merging this to get the obvious bug fix in. Discussion moved to https://github.com/mavlink/mavlink/issues/981

PR #971, opened 2018-08-03
common: allow for longer tunes in PLAY_TUNE

amilcarlucas:


LorenzMeier:
Conclusion from dev call: The extension is harmless, but its hard for adopters to split between tune and tune2 fields and the message is a bit clunky. So the proposal would be to merge as-is, but also define a new message with the full field length and deprecate the old message end of this year.

hamishwillee:
@amilcarlucas You OK with this proposal? If so, can you update branch so can be merged?

@LorenzMeier Did anyone take action to propose the new message? (and if not, who are the interested stakeholders who might do so?)

amilcarlucas:
Rebased

PR #974, opened 2018-08-09
Remove all unnecessary html/classes from output

hamishwillee:
This removes all the inline html styling and classes that aren't used by the gitbook system. The style for text alignment in html is added back as a style block via the python script.
Comments replace the classes that were removed - these don't render in the output.

This reduces the output size significantly, reducing download size. See https://github.com/mavlink/mavlink-devguide/issues/103


hamishwillee:
Thanks!

PR #976, opened 2018-08-10
Revert previous change to highlight this is not official

hamishwillee:
Broken. Will do more testing/retry on Monday

### DIFF
--- a/doc/mavlink_to_html_table.xsl
+++ b/doc/mavlink_to_html_table.xsl
@@ -8,6 +8,10 @@
 </xsl:template>
 
 <xsl:template match="//enums">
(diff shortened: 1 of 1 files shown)


doc/mavlink_to_html_table.xsl
```diff
@@ -55,11 +59,6 @@
 </xsl:template>
 
 <xsl:template match="//version">
-   <h2 style="color:red;">MAVLink Documentation</h1>
```
amilcarlucas: You mixed h2 with h1 here

hamishwillee: Thank you. Fixed.

PR #670, opened 2017-01-31
Added enum for better use of bitmask on the set_position_target_local…

PeterQFR:
…_ned message

This PR addresses #669. It will provide a common location for enums relating to offboard control using mavlink in the px4 flight controller stack.

Issue #669, opened 2017-01-31
Add enumerator in common.xml for mavlink_set_position_target


PeterQFR:
The type_mask for the mavlink_set_position_target is poorly defined including that the bit numbering in the description starts at 1 instead of 0 and that the values of the highest nibble are not well defined (and are actually poorly implemented in px4 FC).

Add an enumerator in the common.xml to define these masks and provide a common enum for use of the bitmask.
@dagar

amilcarlucas:
Good point, and very simple to fix. Can you please provide a pull request ?

amilcarlucas:
Fixed in #997

PR #978, opened 2018-08-14
common.xml: add enum for position target mask

khancyr:
This enum will help to write correct usage of set_position_target either global or local

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3647,7 +3647,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_LOCAL_NED = 1, MAV_FRAME_LOCAL_OFFSET_NED = 7, MAV_FRAME_BODY_NED = 8, MAV_FRAME_BODY_OFFSET_NED = 9</field>
-      <field type="uint16_t" name="type_mask" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration. Mapping: bit 1: x, bit 2: y, bit 3: z, bit 4: vx, bit 5: vy, bit 6: vz, bit 7: ax, bit 8: ay, bit 9: az, bit 10: is force setpoint, bit 11: yaw, bit 12: yaw rate</field>
+      <field type="uint16_t" name="type_mask" enum="POSITION_TARGET_TYPEMASK" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle.</field>
       <field type="float" name="x" units="m">X Position in NED frame</field>
       <field type="float" name="y" units="m">Y Position in NED frame</field>
       <field type="float" name="z" units="m">Z Position in NED frame (note, altitude is negative in NED)</field>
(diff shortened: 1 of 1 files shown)


amilcarlucas:
The attribute `enum="POSITION_TARGET_MASK"` surely needs to be added to some messages.

khancyr:
update to include enum="POSITION_TARGET_MASK"

message_definitions/v1.0/common.xml
```diff
@@ -2959,6 +2959,60 @@
         <description>Spektrum DSMX</description>
       </entry>
     </enum>
+    <enum name="POSITION_TARGET_MASK">
+      <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration. Mapping: bit 1: x, bit 2: y, bit 3: z, bit 4: vx, bit 5: vy, bit 6: vz, bit 7: ax, bit 8: ay, bit 9: az, bit 10: is force setpoint, bit 11: yaw, bit 12: yaw rate</description>
```
hamishwillee: Suggest we drop the bit `Mapping: bit ...` and make sure that the enum information is sufficient. No need to duplicate descriptions.

message_definitions/v1.0/common.xml
```diff
@@ -3593,7 +3647,7 @@
...
-      <field type="uint16_t" name="type_mask" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration. Mapping: bit 1: x, bit 2: y, bit 3: z, bit 4: vx, bit 5: vy, bit 6: vz, bit 7: ax, bit 8: ay, bit 9: az, bit 10: is force setpoint, bit 11: yaw, bit 12: yaw rate</field>
+      <field type="uint16_t" name="type_mask" enum="POSITION_TARGET_MASK" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration. Mapping: bit 1: x, bit 2: y, bit 3: z, bit 4: vx, bit 5: vy, bit 6: vz, bit 7: ax, bit 8: ay, bit 9: az, bit 10: is force setpoint, bit 11: yaw, bit 12: yaw rate</field>
```
hamishwillee: Suggest remove everything after `ignored by the vehicle`. The rest is defined in the (linked) enum and we should avoid duplicating it here.

hamishwillee: Also below ...

message_definitions/v1.0/common.xml
khancyr: ok

message_definitions/v1.0/common.xml
khancyr: yep, it will be cleaner

khancyr:
rebase with @hamishwillee suggestion !

hamishwillee:
Thanks @khancyr . One more concern - how/are the enum description strings going to render in QGC? If they do get rendered then something like `<description>0b0000000000000000 Ignore all</description>` is very ugly. If not, then it doesn't matter.

@amilcarlucas I'm sure you'll have something helpful to add to this discussion.

message_definitions/v1.0/common.xml
```diff
@@ -2959,6 +2959,60 @@
         <description>Spektrum DSMX</description>
       </entry>
     </enum>
+    <enum name="POSITION_TARGET_MASK">
```
amilcarlucas: In order to be cosistent with other display="bitmask" fields, and in order to make the life easier for GCS programmers and general people please:
Remove the descriptions for 0 and non-power of two bitmasks. (0, 7, 56, 448, 65535)

khancyr: Then, it is better to remove the bitmask field ... POSITION_IGNORE, VELOCITY_IGNORE, and ACCELERATION_IGNORE are far simplier to use than to set every field by hand...

amilcarlucas: The idea is that "display=bitmask" guides the GCS to do:
grafik

Without it it just displays a cryptic number. Or in this case a combobox (with only the combinations explicitly defined in the enum)

So I like the "display=bitmask" better.

amilcarlucas: Each box is a single bit. The GCS calculates the decimal number by adding the codes together.
If a box is multiple bits and someone selects the a bit twice (once in it's pure state, one in its combined-with-other-bits state, the wrong number gets calculated). This can be of course addressed by carefully coding it on the GCS side. But I would really like to keep it simple. So far no other bitmask has had such corner cases.

amilcarlucas:
This supersedes #670 and implements #669 Those two should be closed once this gets merged.
There has been some discussion in #670 about C/C++ namespacing on the enum text. That should be addressed here IMHO.

message_definitions/v1.0/common.xml
khancyr: As I said this is an ENUM for programmer not for GCS GUI ... but having non-2-powered should be a problem in any sane programming language ... just do bitwise operation ... X_IGNORE | POSITION_IGNORE = POSITION_IGNORE ... that is the goal of bitmask ....
And POSITION_IGNORE, VELOCITY_IGNORE etc are the normal case ... X_IGNORE, VX_IGNORE are the corner one's ...

khancyr:
Should I close this ?

amilcarlucas:
@khancyr this PR is better than #670 , but it does need to take the comments in #670 into consideration.

khancyr:
@amilcarlucas there are no issue related to #670 , PX4 mode handling had no place this enum.

khancyr:
changes done, rename to POSITION_TARGET_TYPEMASK, add this string in from of each field to prevent namespacing issue for to generic field

amilcarlucas:
Once this gets merged #669 and #670 can be closed

hamishwillee:
I'm not keen on the descriptions including the bitmask (`0b0000000111000000 Ignore acceleration x,y,z`) but not sure we have any choice since this is "non standard". Why not just have all of these setting single bits (ie `Ignore acceleration x` and `Ignore acceleration y` ... this would be completely consistent and require no work from any GCS to support.

amilcarlucas:
I've merged part of this PR that everybody agrees upon with #997 
The part that we seam to disagree, I've separated to #998

khancyr:
you merge the useless part...

PR #986, opened 2018-09-17
common: clarify BATTERY_STATUS description

MaEtUgR:
The enum MAV_BATTERY_CHARGE_STATE is mentioned in XML by enum="MAV_BATTERY_CHARGE_STATE" but does not show up in documentation as far as I see. So I made it clear in the description.

FYI @julianoes

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4226,7 +4226,7 @@
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
       <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
No, this should be rejected. 

It is documented where it is supposed to be: https://mavlink.io/en/messages/common.html#BATTERY_STATUS

hamishwillee:
And yes, I know that it isn't in mavlink.org - that is deprecated, see comment in red: mavlink.org/messages/common

It may be time to think about redirecting that.

julianoes:
I agree you should redirect to the new page. Nothing worse than conflicting information of the same thing.

hamishwillee:
I agree. But there are some sensitivities. Discussing.

hamishwillee:
Either way, this can be closed

MaEtUgR:
👍 You're right. I was always using the old page like presumably a lot of other people.

> It may be time to think about redirecting that.

I think that's absolutely crucial because the google index ranks the old page quite high and hence people will find it first.

hamishwillee:
> I think that's absolutely crucial because the google index ranks the old page quite high and hence people will find it first.

Will raise in the devcall tonight.

PR #991, opened 2018-09-21
Modify markdown generator to split output

hamishwillee:
This splits the generated markdown into an md file (the intro) and html file (the html format of the xml). The markdown includes the html so the rendered output is identical. However this allows the file to be translated (the translation software choked on it). It is also better for translators to work in html. 

This is tested, and I already dumped the output into the devguide.


hamishwillee:
Fixed as part of #1021 - closing.

PR #995, opened 2018-09-26
[WIP] Added fields for RTL and LAND time estimates

potaito:
[**don't merge**]

There has frequently been a mention of "Smart RTL" and similar things, where the drone does not only return based on battery capacity, but also the distance to home. I'm not working on smart RTL, but a feature that estimates the time to perform RTL and LAND actions given the current location. This can then be used to trigger RTL in time, in case the drone is far away from home. 

In my tests, the drone publishes a uorb message at 1Hz for the RTL/Land time estimates. 

Now, for a GroundStation to be able to display the remaining effectively remaining flight time, which is "battery-based remaining flight time" minus "RTL time estimate", we need to convey the data via mavlink. And I am not sure where it best belongs. 

The estimated time to perform RTL would in my opinion best fit into the message `<message id="242" name="HOME_POSITION">`, since it is data depending on and derived from the home position. But the LAND time estimate does not fit in there, since LAND just happens wherever the drone is currently located.

@MaEtUgR suggested to add it next to the total remaining flight time, which is inside the `BATTERY_STATUS` message.

Thoughts?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5116,11 +5116,11 @@
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
-      <field type="int32_t" name="rtl" units="s">Estimated time duration for RTL given current vehicle position. The estimate reflects whatever type of RTL is curently configured. -1 means already landed or no time estimate available.</field>
-      <field type="int32_t" name="land" units="s">Estimated time duration for LAND action given current vehicle position. -1 means already landed or no time estimate available.</field>
-      <field type="int32_t" name="mission_next_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
-      <field type="int32_t" name="mission_end" units="s">Estimated time duration for completing the current mission. -1 means no mission active and/or no estimate available.</field>
-      <field type="int32_t" name="non_mission_action" units="s">Estimated time duration for completing the current action that is not a mission (i.e. Go To, Orbit, takeoff, land, etc.). -1 means no action active and/or no estimate available.</field>
+      <field type="int32_t" name="rtl" units="s">Estimated time to complete the vehicle's configured "safe return" action (e.g. RTL, Smart RTL, etc.) from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
+      <field type="int32_t" name="land" units="s">Estimated time for vehicle to complete the LAND action from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
+      <field type="int32_t" name="mission_next_item" units="s">Estimated time for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
+      <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
+      <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4472,6 +4472,8 @@
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
       <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
+      <field type="int32_t" name="time_estimate_rtl" units="s">Estimated time for RTL action, 0: Already landed or autopilot does not provide RTL time estimate</field>
+      <field type="int32_t" name="time_estimate_land" units="s">Estimated time for LAND action, 0: Already landed or autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
     </message>
     <message id="148" name="AUTOPILOT_VERSION">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5118,7 +5118,7 @@
       <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
       <field type="int32_t" name="rtl" units="s">Estimated time duration for RTL given current vehicle position. The estimate reflects whatever type of RTL is curently configured. -1 means already landed or no time estimate available.</field>
       <field type="int32_t" name="land" units="s">Estimated time duration for LAND action given current vehicle position. -1 means already landed or no time estimate available.</field>
-      <field type="int32_t" name="mission_nex_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
+      <field type="int32_t" name="mission_next_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
       <field type="int32_t" name="mission_end" units="s">Estimated time duration for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="non_mission_action" units="s">Estimated time duration for completing the current action that is not a mission (i.e. Go To, Orbit, takeoff, land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5132,7 +5132,7 @@
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
-      <field type="int32_t" name="rtl" units="s">Estimated time to complete the vehicle's configured "safe return" action (e.g. RTL, Smart RTL, etc.) from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
+      <field type="int32_t" name="safe_return" units="s">Estimated time to complete the vehicle's configured "safe return" action from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
       <field type="int32_t" name="land" units="s">Estimated time for vehicle to complete the LAND action from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
       <field type="int32_t" name="mission_next_item" units="s">Estimated time for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5113,10 +5113,14 @@
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
     <message id="380" name="TIME_ESTIMATE_TO_TARGET">
-      <description>Time estimates until specific targets are reached or actions performed, given the current vehicle state and position</description>
-      <field type="int32_t" name="rtl" units="s">Estimated time for RTL action given current vehicle position, 0: Already landed or autopilot does not provide RTL time estimate</field>
-      <field type="int32_t" name="land" units="s">Estimated time for LAND action given current vehicle position, 0: Already landed or autopilot does not provide LAND time estimate</field>
-      <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission, 0: Mission complete or no mission active</field>
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
+      <field type="int32_t" name="rtl" units="s">Estimated time duration for RTL given current vehicle position. The estimate reflects whatever type of RTL is curently configured. -1 means already landed or no time estimate available.</field>
+      <field type="int32_t" name="land" units="s">Estimated time duration for LAND action given current vehicle position. -1 means already landed or no time estimate available.</field>
+      <field type="int32_t" name="mission_nex_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
+      <field type="int32_t" name="mission_end" units="s">Estimated time duration for completing the current mission. -1 means no mission active and/or no estimate available.</field>
+      <field type="int32_t" name="non_mission_action" units="s">Estimated time duration for completing the current action that is not a mission (i.e. Go To, Orbit, takeoff, land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4472,8 +4472,6 @@
       <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
       <extensions/>
       <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
-      <field type="int32_t" name="time_estimate_rtl" units="s">Estimated time for RTL action, 0: Already landed or autopilot does not provide RTL time estimate</field>
-      <field type="int32_t" name="time_estimate_land" units="s">Estimated time for LAND action, 0: Already landed or autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
     </message>
     <message id="148" name="AUTOPILOT_VERSION">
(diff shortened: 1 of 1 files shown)


WickedShell:
My first inclination is that this information should be published in a different message. `BATTERY_STATUS` is sent for every battery on the vehicle. If you have 4 (6 to 10 isn't even a stretch these days) batteries onboard the data in this field will be the same on all the messages you send and adds overhead (particularly when battery monitoring is run at a faster rate like 5Hz, but you only need this information at 1Hz). Probably worth seeing what others think before changing the PR though.

potaito:
Right, I didn't even think about that. Thanks for the input @WickedShell

MaEtUgR:
> `BATTERY_STATUS` is sent for every battery on the vehicle.

I didn't know that, I was expecting the message to be the summary of the whole battery situation on the vehicle. If it's the case that it's used per battery then it doesn't make sense to add distance related failsafe action times. But then I'm honestly also not sure how the battery remaining time should be defined for each battery if there are multiple...

WickedShell:
@MaEtUgR The `id` field is used so you can tell which battery on the system is being reported. The remaining time field is used to indicate the remaining time that battery has at it's current discharge rate. If you consider an aircraft with a propulsion battery and a payload battery it becomes easy to see how these can be very different times. (And even an all electric quadplane where the VTOL pack is powered from one battery and the forward is a different pack).

potaito:
Okay it's quite obvious now that `BATTERY_STATUS` is not the right place for it. But I don't see where else it would fit in. Unless we leave out the estimated time for `LAND` and put the estimated time for `RTL` inside the `HOME_POSITION` message.

But then the next person wanting to add the `LAND` time estimate will have an even more difficult time. 
@bkueng halp plz :smiley:

WickedShell:
You could add a message for TIME_ESTIMATE_TO_TARGET with a field for time and an enum to say what the target is (IE RTL, fly a landing, complete mission) but that might be more complicated to manage

bkueng:
TIME_ESTIMATE_TO_TARGET sounds reasonable to me, but probably with a fixed set of time estimate fields (RTL, Land, mission, safe point?, more?) to simplify the streaming/configuration of the message.

potaito:
Sounds good. I thought adding new mavlink messages is rarely acceptable. But in this case there really doesn't seem to be a better way. 

So we would have something like:

- RTL
- LAND
- MISSION_NEXT
- MISSION_END (probably cumbersome to implement)
- what's safe point? :D
- GEOFENCE_BOUNDS

For serious commercial applications it's probably a good thing to always have an estimate of when the drone will be done with the current task such as taking off, rising to a designated altitude etc. Would you put actions like takeoff also under MISSION_NEXT ?

Also, would you go with `uint16` or `uint32`? If the units of the message elements are seconds, we have a maximum time frame of 

- `uint16`: 18.2 hours
- `uint32`: 136 years

I guess with winged aircraft in mind it's perfectly possible that the vehicle stays airborne for longer than 18 hours.

bkueng:
> what's safe point? :D

Essentially an alternative RTL position (or safe landing spot, e.g. for longer missions). Also known as Rally points.

> For serious commercial applications it's probably a good thing to always have an estimate of when the drone will be done with the current task such as taking off, rising to a designated altitude etc. Would you put actions like takeoff also under MISSION_NEXT ?

So maybe something like NEXT_TARGET? Which is the next waypoint in a mission.
I wouldn't add too many things though as it increases the efforts to implement it and we can still extend the message later on if needed.

> I guess with winged aircraft in mind it's perfectly possible that the vehicle stays airborne for longer than 18 hours.

Yes I'd go for uint32.

potaito:
Okay I added a new message. But I am not sure if the ID is on the correct/legal range of IDs.

hamishwillee:
@potaito 
1. Guidance on [id ranges here](https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges). Essentially you should choose around 340. 150 is in the MAVLink 1 range and should not be allocated (I can explain why if the docs are not clear, but suggest you take it as read).
2. 
   > I thought adding new mavlink messages is rarely acceptable. But in this case there really doesn't seem to be a better way.

   We love to add new MAVLink messages :-). We just want to make sure that they are future proofed as much as possible and work well with existing message. This does seem a good idea. 
3. Is the idea that the message is sent when you are in (say) an RTL OR would it be sent all the time - so that I could as a reader decide whether or not to RTL based on the time?
4. The text descriptions still mention battery but battery capacity does not seem to be a condition for estimating time to target.

potaito:
@hamishwillee Thanks, I'll make the adjustments to the ID :+1: 

> Is the idea that the message is sent when you are in (say) an RTL OR would it be sent all the time - so that I could as a reader decide whether or not to RTL based on the time?

The latter. Currently PX4 has the capability of executing RTL/LAND based on battery levels and various thresholds (warning, danger, emergency...). When far away from home, RTL will not make it back in time and the drone will emergency land somewhere in between. The time estimates for the RTL action for instance would make it possible to execute RTL early enough to make it home before emergency levels are reached. Perhaps the mavlink message is not clear enough about that :thinking: 

> The text descriptions still mention battery but battery capacity does not seem to be a condition for estimating time to target.

Oh those are copy/paste mistakes :see_no_evil:

hamishwillee:
> Would you suggest skipping to 340 to leave some room for additional trajectory related message?

Yes, I suggested for exactly that reason. There are no firm recommendations in the spec for inserting spacing, but IMO it is a good idea to do so, and I suspect that 7 spare slots will be more than enough.

@bkueng I know some messages are streamed all the time, and in other cases information is triggered - ie there are a pair of messages where the first is a request and the second is a response with the information. When should you use each type, and which would this be?
If it should be streamed *by default* what would the rate be?

Should the RTL be generic or do we need multiple forms of RTL? Specifically RTL means Return To Land, and typically is interpreted as "Return to Home Position and Land". However on some systems the configuration might mean that you return and loiter. On others you might support some form of smart return that will take another path than direct to home (SmartRTL). On others you might have a rally point, and RTL will take you to that if it is closer than home. 

I'm not sure what is best. For the stated purpose of failsafe handling you might just call this "Return" and let the system tell you the time for whatever its failsafe return behaviour is. However if you're also interested in possibility that they might Return manually using a switch to select the "type" then you'd need variants for all common return types.

@bkueng Sounds like smarter return behaviour or warnings based on capacity might even be useful within PX4.

hamishwillee:
> TIME_ESTIMATE_TO_TARGET sounds reasonable to me, but probably with a fixed set of time estimate fields (RTL, Land, mission, safe point?, more?) to simplify the streaming/configuration of the message.

@bkueng Giving this further thought, do you really think that this difficulty outweighs the benefits of future compatibility and bandwidth from using an enum? 

From a MAVLink perspective I'd much prefer something like ...

```
    <enum name="MAV_TIME_ESTIMATE_EVENT">
      <description>Type of event for which a time estimate is required (in TIME_ESTIMATE_TO_TARGET).</description>
      <entry value="0" name="MAV_TIME_ESTIMATE_EVENT_RETURN">
        <description>Return operation complete.</description>
      </entry>
      <entry value="1" name="MAV_TIME_ESTIMATE_EVENT_LANDING">
        <description>Vehicle is landed.</description>
      </entry>
      <entry value="2" name="MAV_TIME_ESTIMATE_EVENT_MISSION_END">
        <description>Mission is completed (completion of last waypoint).</description>
      </entry>
   </enum>

	   <message id="340" name="TIME_ESTIMATE_TO_TARGET">
       <description>Time estimates until specific target/action complete, given the current vehicle state and position.</description>
       <field type="uint8_t" name="event_type" enum="MAV_TIME_ESTIMATE_EVENT">Type of event/action for which remaining-time-estimate required,</field>
       <field type="int32_t" name="estimate" units="s">Estimated time for current event type from current vehicle position, -1: Operation already complete or not supported.</field>
     </message>
```

@potaito I assume you are still working on this? Apologies that this stalled.

potaito:
@hamishwillee I would still like to have the time estimates in some form communicated over mavlink, yes :)
But I'm not actively working on a solution since I don't really know how to properly do it other than the suggestion in the PR.

hamishwillee:
Thanks @potaito 

We'll discuss in the next devcall, and also hopefully @bkueng will have some more useful feedback.

potaito:
Great, thanks. Then I should probably join the call :sweat_smile:

bkueng:
> @bkueng I know some messages are streamed all the time, and in other cases information is triggered - ie there are a pair of messages where the first is a request and the second is a response with the information. When should you use each type, and which would this be?
If it should be streamed by default what would the rate be?

Streamed messages contain information that regularly gets updated, and can be displayed in a UI accordingly. An example is the vehicle's attitude.
Requested messages generally contain information that is static, i.e. does not change later on (or only rarely). An example is the autopilot version.
To me the time estimate should clearly be a streamed message, and 1Hz (or even somewhat lower) seems reasonable.

@hamishwillee we can go for the enum-based version as well, but from an implementation perspective we would very likely end up sending all of them together (and you cannot configure the rate separately for different types). Also you would see the message arriving with a higher rate than you actually specify - not sure what the spec says on that.

MaEtUgR:
Streaming this message as long as there is at least one useful time estimate contained in it with `1Hz (or even somewhat lower)` sounds bearable to me.

message_definitions/v1.0/common.xml
```diff
@@ -4774,5 +4774,11 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="340" name="TIME_ESTIMATE_TO_TARGET">
+      <description>Time estimates until specific targets are reached or actions performed, given the current vehicle state and position</description>
```
LorenzMeier: This is a cool concept, but is this sufficient to build a complete UI for the user or is more information needed? What else would you be using? Battery status?

Would you be willing to join the MAVLink dev call in two weeks to discuss?

potaito: Can do

potaito: So the critical information we need is 
1. remaining flight time based on current battery level
2./3. Duration for emergency actions RTL and LAND
4. Duration for reaching next mission item
5. Duration for reaching mission end

The remaining flight time is already part of the battery status message. But since the battery message conveys information for a single battery, and there are multi-battery setups, it make sense to move the estimate total remaining flight time over here.

hamishwillee: + @MaEtUgR @auturgy FYI

@potaito I edited your post slightly just to my taste.

A few more points from devcall:
- I assume that even though there can be muliple RTLs available to a system (normal RTL, SMART RTL, RTL to rally points etc etc) only one will be configured at a time. The RTL we are reporting is whatever is currently configured (right?). I am also assuming that it is OK if a system is configured for SMART_RTL but for whatever reason decides it has to use a normal RTL - and so what the message outputs might change quite substantially?
- Messages normally have a value that indicates a field is not used/invalid/not populated. For floats this is often NaN. Ideally it is zero. For UInts it may be set to max value. For ints it may be set to -1. 
- Can you please update PR as discussed and make it [wip](https://mavlink.io/en/guide/xml_schema.html#wip).

hamishwillee: @potaito @MaEtUgR please see above, and in particular "Can you please update PR as discussed and make it wip."? 

Please ETA!

potaito: @hamishwillee Many apologies. I had two really stressful weeks. 

I updated the message now. 
* Rebased on newest master
* Added `<wip/>` tag
* Added wip comment

I am not too happy with the wording of the RTL message and the fact that it can reflect RTL, smart RTL or whatever kind of RTL. I'm also not sure if we have a better name for what I called "non-mission actions".

message_definitions/v1.0/common.xml
```diff
@@ -5112,6 +5112,16 @@
...
     </message>
+    <message id="380" name="TIME_ESTIMATE_TO_TARGET">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
+      <field type="int32_t" name="rtl" units="s">Estimated time duration for RTL given current vehicle position. The estimate reflects whatever type of RTL is curently configured. -1 means already landed or no time estimate available.</field>
+      <field type="int32_t" name="land" units="s">Estimated time duration for LAND action given current vehicle position. -1 means already landed or no time estimate available.</field>
+      <field type="int32_t" name="mission_nex_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
```
hamishwillee: typo mission_nex_item

potaito: :man_facepalming: thanks

message_definitions/v1.0/common.xml
```diff
@@ -5112,6 +5112,16 @@
       <field type="uint16_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
+    <message id="380" name="TIME_ESTIMATE_TO_TARGET">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
+      <field type="int32_t" name="rtl" units="s">Estimated time duration for RTL given current vehicle position. The estimate reflects whatever type of RTL is curently configured. -1 means already landed or no time estimate available.</field>
```
hamishwillee: I don't think this description is bad. Is this better?

> Estimated time to complete the vehicle's configured "safe return" action (e.g. RTL, Smart RTL, etc.) from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.

message_definitions/v1.0/common.xml
hamishwillee: @potaito Thanks very much

No worries re "tardiness". I'm just generally trying to make sure that the person who posted an issue knows that to some extent they are going to have to push for resolution - I won't do it by default.

> I am not too happy with the wording of the RTL message and the fact that it can reflect RTL, smart RTL or whatever kind of RTL. 

I have tried to reword. Please check.

> I'm also not sure if we have a better name for what I called "non-mission actions".

Let me see!

message_definitions/v1.0/common.xml
```diff
@@ -5112,6 +5112,16 @@
...
+      <field type="int32_t" name="land" units="s">Estimated time duration for LAND action given current vehicle position. -1 means already landed or no time estimate available.</field>
```
hamishwillee: Slight modification suggested (just to bring inline with above suggestion)

> Estimated time for vehicle to complete the LAND action from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.

message_definitions/v1.0/common.xml
potaito: Yes, a lot better in my opinion :+1:

message_definitions/v1.0/common.xml
```diff
@@ -5112,6 +5112,16 @@
...
+      <field type="int32_t" name="mission_nex_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
+      <field type="int32_t" name="mission_end" units="s">Estimated time duration for completing the current mission. -1 means no mission active and/or no estimate available.</field>
+      <field type="int32_t" name="non_mission_action" units="s">Estimated time duration for completing the current action that is not a mission (i.e. Go To, Orbit, takeoff, land, etc.). -1 means no action active and/or no estimate available.</field>
```
hamishwillee: How about calling these commanded actions? Generally you kick them off in offboard/guided mode, often with a mission command (though that depends on the flight stack/implementation) as they may be a proper "mode".

potaito: Yeah unfortunately I only have the PX4 perspective. @WickedShell do you have a preference for naming this field? Does Ardupilot have autonomous actions that are not part of a mission? And do they have a name?

Commanded actions would be fine for me.

message_definitions/v1.0/common.xml
```diff
@@ -5116,11 +5116,11 @@
...
-      <field type="int32_t" name="land" units="s">Estimated time duration for LAND action given current vehicle position. -1 means already landed or no time estimate available.</field>
-      <field type="int32_t" name="mission_next_item" units="s">Estimated time duration for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
-      <field type="int32_t" name="mission_end" units="s">Estimated time duration for completing the current mission. -1 means no mission active and/or no estimate available.</field>
-      <field type="int32_t" name="non_mission_action" units="s">Estimated time duration for completing the current action that is not a mission (i.e. Go To, Orbit, takeoff, land, etc.). -1 means no action active and/or no estimate available.</field>
+      <field type="int32_t" name="rtl" units="s">Estimated time to complete the vehicle's configured "safe return" action (e.g. RTL, Smart RTL, etc.) from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
```
hamishwillee: Just as an aside, should we change name to safe_return? Just a thought

message_definitions/v1.0/common.xml
hamishwillee: @auturgy Please see ^^^ - does the term "commanded action" also work for ArduPilot to refer to non-mission items for which you want a time estimate?

Note that IMO it does though it isn't "perfect" for either stack in that some "actions" are actually modes (e.g. LAND), while others might be MAV_CMD actions in guided or offboard mode. 

Also some cases don't quite work - e.g. Orbit on PX4 is a task, but it has no end point, and similarly CIRCLE mode is a permanent loiter. I don't think that matters - it is up to the stack/GCS to understand the context of the current mode to interpret the "commanded action".

message_definitions/v1.0/common.xml
hamishwillee: @potaito I think we should call this safe return, get out of all that RTL, Smart RTL etc naming history. Can you do that. Then ping me and I will merge. Let's get this live.

potaito: I agree and changed it.

potaito:
@hamishwillee I updated the PR (see last commit).

message_definitions/v1.0/common.xml
```diff
@@ -5128,6 +5128,16 @@
       <field type="uint16_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
+    <message id="380" name="TIME_ESTIMATE_TO_TARGET">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Time/duration estimates for various events and actions given the current vehicle state and position.</description>
+      <field type="int32_t" name="safe_return" units="s">Estimated time to complete the vehicle's configured "safe return" action from its current position. -1 indicates that the vehicle is landed, or that no time estimate available.</field>
```
hamishwillee: I'm going to merge and add a clarification here of what safe return means.

hamishwillee:
Thanks @potaito ! Merging now. FYI @auturgy @LorenzMeier et al. This is WIP, so opportunity to change still exists.

@DonLakeFlyer FYI, this adds a mechanism for a vehicle to report how long it will take to complete a particular action from its current position. For example, you could use this to work out if you have enough time to complete an RTL given current battery capacity.

potaito:
Thanks for your help and persistence @hamishwillee :D

PR #998, opened 2018-10-01
common.xml: Add multibit entries to POSITION_TARGET_TYPEMASK bitmap

amilcarlucas:
This is the contradictory part of #978.

I'm against this, but I place it here for further discussion.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2961,6 +2961,9 @@
     </enum>
     <enum name="POSITION_TARGET_TYPEMASK">
       <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration.</description>
+      <entry value="0" name="POSITION_TARGET_TYPEMASK_NONE_IGNORE">
+        <description>0b0000000000000000 Ignore none</description>
+      </entry>
       <entry value="1" name="POSITION_TARGET_TYPEMASK_X_IGNORE">
         <description>0b0000000000000001 Ignore position x</description>
       </entry>
@@ -2969,6 +2972,9 @@
       </entry>
       <entry value="4" name="POSITION_TARGET_TYPEMASK_Z_IGNORE">
         <description>0b0000000000000100 Ignore position z</description>
+      </entry>
+      <entry value="7" name="POSITION_TARGET_TYPEMASK_POSITION_IGNORE">
+        <description>0b0000000000000111 Ignore position x,y,z</description>
       </entry>
       <entry value="8" name="POSITION_TARGET_TYPEMASK_VX_IGNORE">
         <description>0b0000000000001000 Ignore velocity x</description>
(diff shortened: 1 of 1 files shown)


khancyr:
those are not contradictory... It is the usefull part of my pr...
the use of setpoint mostly use those enum and not the other one... 
this is a bitmask enum not power of 2 bitmask...

DonLakeFlyer:
Nothing that QGC uses as far as I remember. What mavlink message are those bits used for? I don't see it here: https://mavlink.io/en/messages/common.html

hamishwillee:
@DonLakeFlyer The change is incomplete. The enum is supposed to be added as a *new field* in `SET_POSITION_TARGET_LOCAL_NED` to indicate which dimensions can be ignored in the vehicle. The full discussion is [here](https://github.com/mavlink/mavlink/pull/978/files) - best to start there to understand this.

The new field has values that looks like a bitmask, but has some values which are combinations of the other bits. My concerns here was how this would work with QGC:
- Does it matter that some of the values are combinations of bits rather than just a single value to bit mapping? The combined bit mappings are consistent - ie the ignore value for x,y,z IS the logical OR of the x, y, and z values.
- Do you agree that having the bitmap in the description is not needed in the UI - i.e. why "0b0000000000000001 Ignore position x" rather than "Ignore position x" is consistent. 

@khancyr I think @amilcarlucas was trying to _help_ you by getting this PR progressed and leaving us to discuss only the bits that are "contentious". For my part I don't have a real problem with what you've done (other than the descriptions), I just wanted to ensure that GCS can render this usefully. You didn't respond to that .... https://github.com/mavlink/mavlink/pull/978#issuecomment-417965948

khancyr:
As I said previously, I don't make this enum to be a checkbox for GCS ... that was intend to be an helper for developers to prevent multiplication those enum see : https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/defines.h#L421 and https://github.com/mavlink/mavros/blob/master/mavros/src/plugins/setpoint_position.cpp#L132 .
Moreover the bits you are calling contentious are the only one's that are usefull ... 

About the GCS part, I really don't see the problem. If you are using bitwise operation, there won't be any problem at having more field. If you are using addition ... then use bitwise operation when manipulation bitmask ... 

Yes, we could remove the 0b0000000000X00000 part from description. But the thing that what merge isn't my work, and I don't have the hand this one...

khancyr:
This can be close, I don't need it anymore

hamishwillee:
@amilcarlucas Still useful for others? If so then we need to add the enum to `SET_POSITION_TARGET_LOCAL_NED`. If not, then we need to revert https://github.com/mavlink/mavlink/pull/978.

I vote revert. Thoughts?

amilcarlucas:
No, no reverts, the #997 fixed an issue and superseded another PR. Please see #997 for more info.

Regarding this particular PR what we need is information from @DonLakeFlyer on how he intends to implement the `display="bitmask"` and `enum` inside QGC

amilcarlucas:
Let's close this one, and wait until @DonLakeFlyer implements it in QGC. Witl the available information it should also be possible for @meee1 to implement `display="bitmask"` and `enum` inside mission Planner and to the UGCS people to implement it.

PR #1001, opened 2018-10-08
update CI script

khancyr:
update syntax with https://www.shellcheck.net/
test all xml and not only common.xml


khancyr:
Still need https://github.com/ArduPilot/pymavlink/pull/225
I have update to generate C++11 too as that was missing and highly needed for ROS!

hamishwillee:
@khancyr Just FYI, I added [WIP] to title to give people pause before they merge this (so that https://github.com/ArduPilot/pymavlink/pull/225 goes in first)

khancyr:
rebase since the pymavlink PR goes in ! I should be good to go now !

PR #1004, opened 2018-10-16
Add missing message descriptions

hamishwillee:
All messages should have descriptions. This PR adds missing descriptions so that we can start enforcing the requirement (see https://github.com/ArduPilot/pymavlink/pull/215)

I've taken a shot at descriptions. 
- Does anyone know the difference between GLOBAL_VISION_POSITION_ESTIMATE, VISION_POSITION_ESTIMATE, VICON_POSITION_ESTIMATE? They are all very similar. Don't really see why we don't have GLOBAL_POSITION_ESTIMATE and provide a field for the source type.
- These all put position in terms of metres. Where is the origin of this "global" system?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3781,6 +3781,7 @@
       <field type="float" name="flow_rate_y" units="rad/s">Flow rate about Y axis</field>
     </message>
     <message id="101" name="GLOBAL_VISION_POSITION_ESTIMATE">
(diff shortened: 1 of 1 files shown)


hamishwillee:
@TSC21 Can you advise?

hamishwillee:
@TSC21 I'd still love your answer to my questions above: https://github.com/mavlink/mavlink/pull/1004#issue-223082345

PR #1006, opened 2018-10-16
common: add orbit execution telemetry message

MaEtUgR:
References:
- https://github.com/mavlink/mavlink/pull/924 introduced a message to command the vehicle to do an orbit
- QGroundcontrol added support for this command https://github.com/mavlink/qgroundcontrol/pull/6608, https://github.com/mavlink/qgroundcontrol/pull/6711, https://github.com/mavlink/qgroundcontrol/pull/6753 thanks to @DonLakeFlyer 
- On PX4 autopilot the implementation to execute the orbit can be found here: https://github.com/PX4/Firmware/tree/master/src/lib/FlightTasks/tasks/Orbit

New:
This PR adds a message for the vehicle to report the current status, parameters and therefore planned trajectory to the ground control station while flying an orbit to allow the UI to show a live visualization for the pilot.

Use case:
After commanding an orbit the circle in the qgc UI disappears and the user can only remember where it was and see the tail line of the vehicle forming the same circle again. Even if the circle was kept on the UI after successful command acknowledgement the user might change the parameters of the orbit e.g. tangential speed and radius by sticks or SDK which would then not be reflected in the UI anymore.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4829,5 +4829,16 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
+    <message id="360" name="ORBIT_EXECUTION_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Vehicle status report that is sent out while orbit execution is in progress (see MAV_CMD_DO_ORBIT).</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="float" name="radius" units="m">Radius of the orbit circle. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
+      <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
+      <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
+      <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
+    </message>
   </messages>
 </mavlink>


message_definitions/v1.0/common.xml
```diff
@@ -4829,5 +4829,13 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
+    <message id="360" name="ORBIT_EXECUTION_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Vehicle status report which is only sent out while orbit execution is in progress (see MAV_CMD_DO_ORBIT).</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="state">State of execution. 0: Executing orbit 1: Paused and waiting 2: Approaching Circle 3: Error </field>
```
hamishwillee: Commas between options? ie `0: Executing orbit, 1: Paused ...`

hamishwillee: Why not an enum?

hamishwillee: Do we need to know possible errors?

hamishwillee:
There is info in the original command that isn't in the output - ie centre of the circle. What do you need/not need from that?

message_definitions/v1.0/common.xml
MaEtUgR: I was thinking about that but the error is just for completeness, it should never occur because if there's an error there will be some other task taking over... and should be a message to the pilot.

MaEtUgR:
@hamishwillee Thanks for taking a look, your input is valued. I think there should be more information in the telemetry, it's not finished yet but the radius is the most important one and I wanted to test that first and the whole setup with this new message I propose.

@DonLakeFlyer 's input is appreciated. Tangential velocity should in the best case be equal to the vehicle velocity while orbiting, is there any advantage from knowing the exact setpoint value or what is planned while still approaching the circle? How could that get displayed visually? I think the center of the circle definitely makes sense to add becuase it's easy to display, could be dragged around on the map during orbit execution (commanded again by the mavlink command) and also change dynamically in the future if we support target orbiting. Other configuration options in use might make sense to report in the future but I hope we can still add them to the message or is that a problem that it needs to be all fixed from the beginning?

DonLakeFlyer:
For QGC the location lat/lon/alt and the radius are what is needed. I wouldn't go too crazy with the status bits. For example I think I'd leave out paused and so forth for now. Wondering if those are needed at all? Simpler can be better for generic mavlink since too many bits tends to get firmware specific.

hamishwillee:
FWIW then I would remove the error. 

> ... but the radius is the most important one

in case you missed this comment: "shouldn't this also include the centre of the circle?"

The radius does not do what you need if the UI does not know where to draw the circle.

MaEtUgR:
> "shouldn't this also include the centre of the circle?"

> location lat/lon/alt and the radius are what is needed

sure, we'll keep it simple and only add what's needed thanks for the input!

MaEtUgR:
I updated now according to the requirements:
- timestamp
- radius
- coordinate frame
- x coordinate
- y coordinate
- altitude

Coordinates are based on the `COMMAND_INT` message which is also used for best results with the orbit command as we discussed earlier with @DonLakeFlyer .

hamishwillee:
Looks great to me. @amilcarlucas @auturgy at anyone else have any comment about this?

@LorenzMeier As you know, I've been trying to work out rules that will allow "the team" to review and merge at least some changes to common/dialects without you having to review/approve.

From now on I plan to merge **wip** messages that pass a small checklist:
- The are _wip_, and hence can be excluded from builds
- I understand what they are trying to achieve and they have no glaring errors
- The dev team have had a week to comment

If no one comments in the next couple of days, I plan to merge this on that basis.

amilcarlucas:
Unfortunately current pymavlink still ignores the wip tag :(

DonLakeFlyer:
Looks good enough to me to start working with

hamishwillee:
@amilcarlucas 

> Unfortunately current pymavlink still ignores the wip tag :(

Bugger. Is there a plan/resources to add support for this in the generator? Or just something we want to happen?

hamishwillee:
@MaEtUgR Ping. Haven't had further comments. But this can't be merged without rebase.

MaEtUgR:
@hamishwillee Thanks so much for pushing forward, let me quickly rebase.

message_definitions/v1.0/common.xml
```diff
@@ -4829,5 +4829,16 @@
...
+      <description>Vehicle status report which is only sent out while orbit execution is in progress (see MAV_CMD_DO_ORBIT).</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="float" name="radius" units="m">Radius of the circle. positive: Orbit clockwise. negative: Orbit counter-clockwise.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the coordinates in the message.</field>
+      <field type="int32_t" name="x">Center point X coordinate according to frame field. local: x position in meters * 1e4, global: latitude in degrees * 1e7</field>
+      <field type="int32_t" name="y">Center point Y coordinate according to frame field. local: x position in meters * 1e4, global: latitude in degrees * 1e7</field>
+      <field type="float" name="z">Center point altitude according to frame field. global: altitude in meters (relative or absolute, depending on frame)</field>
```
hamishwillee: @MaEtUgR A few more little thoughts (if this branch was on this repo I'd make them myself. 
-  This should be `units="m"` 
- Descriptions might be better with the "difference" first - so Altitude of centre point etc.
- Use sentence capitalisation for sentences. 

Here is a marked up version you can copy-paste in:
```
    </message>
    <message id="360" name="ORBIT_EXECUTION_STATUS">
      <wip/>
      <description>Vehicle status report that is sent out while orbit execution is in progress (see MAV_CMD_DO_ORBIT).</description>
      <field type="float" name="radius" units="m">Radius of the orbit circle. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
      <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
    </message>
```

hamishwillee:
@LorenzMeier All, I plan on merging this as soon as @MaEtUgR has made my last suggestions. Speak now, or forever hold your peace

message_definitions/v1.0/common.xml
amilcarlucas: Please remove meters from the z description text. Avoid redundant information, especially if it is not machine parse-able ;)

hamishwillee: @amilcarlucas Yes, I've provided an alternative wording and that is included: https://github.com/mavlink/mavlink/pull/1006#discussion_r229925305

MaEtUgR: @hamishwillee Thanks a lot for your suggestion!! They all make sense to me.

> if this branch was on this repo I'd make them myself.

I would have liked to push the branch for cooperation but I have no rights at all so I have to keep the fork workflow.

MaEtUgR:
@hamishwillee Thanks for your help. I encorporated your suggestions. Does it now make sense? I could fill the message for a first test with UI support.

hamishwillee:
@MaEtUgR Thank you. CI was a glitch. As discussed, I am merging this now - everyone who has any interest has had more than enough time to comment.

PR #1008, opened 2018-10-18
Ardupilot downstream changes

amilcarlucas:


message_definitions/v1.0/common.xml
```diff
@@ -2960,42 +2960,42 @@
       </entry>
     </enum>
     <enum name="POSITION_TARGET_TYPEMASK">
-      <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration.</description>
+      <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 9 is set the floats afx afy afz should be interpreted as force instead of acceleration.</description>
```
hamishwillee: What's up with "If bit 9 is set" vs "If bit 10 is set"?

message_definitions/v1.0/common.xml
```diff
@@ -15,7 +15,7 @@
         <description>SLUGS autopilot, http://slugsuav.soe.ucsc.edu</description>
       </entry>
       <entry value="3" name="MAV_AUTOPILOT_ARDUPILOTMEGA">
-        <description>ArduPilotMega / ArduCopter, http://diydrones.com</description>
+        <description>ArduPilot - Plane/Copter/Rover/Sub/Tracker, http://ardupilot.org</description>
```
hamishwillee: Shouldn't all changes to common.xml be made upstream (here) and moved downstream (to ArduPilot)? The actual change is great.

message_definitions/v1.0/common.xml
```diff
@@ -2960,42 +2960,42 @@
       </entry>
     </enum>
     <enum name="POSITION_TARGET_TYPEMASK">
-      <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 10 is set the floats afx afy afz should be interpreted as force instead of acceleration.</description>
+      <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b0000000000000000 or 0b0000001000000000 indicates that none of the setpoint dimensions should be ignored. If bit 9 is set the floats afx afy afz should be interpreted as force instead of acceleration.</description>
       <entry value="1" name="POSITION_TARGET_TYPEMASK_X_IGNORE">
-        <description>0b0000000000000001 Ignore position x</description>
+        <description>Ignore position x</description>
```
hamishwillee: FYI very happy with this.

hamishwillee:
So my general take on this is that for dialect changes from the dialect repo we should just accept them - after checking of course whether they might reasonably be better in Common (I assume this is not the case?)

For the common changes, generally better if they come here first and propagate down to ArduPilot. That said, they are mostly good/obvious except: https://github.com/mavlink/mavlink/pull/1008#discussion_r226500007

message_definitions/v1.0/common.xml
amilcarlucas: The LSB is in 99,9% of the books numbered as bit0, so this is bit9
The original description started count at bit 1 hence it called bit 9 bit 10

amilcarlucas:
@LorenzMeier can I rebase and merge this?

LorenzMeier:
Missed this over the weekend - good to go!

Issue #624, opened 2016-09-20
(IMPROVEMENT) Add kinematic uncertainty (e.g., NIC and NACp) to ADSB_VEHICLE message


marclecerf:
The ADS-B specification includes requirements for aircraft to transmit a 
quantification of the [uncertainty of their position reports](http://adsb-decode-guide.readthedocs.io/en/latest/nicnac.html).

The current MAVLink message used to pack ADS-B position information received from another aircraft ("ADSB_VEHICLE") does not include the ability to transmit this uncertainty information.

Uncertainty quantification is very important for fusion of ADS-B position reports with other forms of kinematic information, and it should be considered as an improvement to the current ADSB_VEHICLE message to add these fields.

magicrub:
We can amend the packet via MAVLink v2.0 but would a flag suffice? If so
that could be added to v1

marclecerf:
Hi Tom,

Thanks for the quick response!

Details are below, but in summary, we could likely encode all the uncertainty information in a uint16. (not sure exactly what you meant by a flag? I’m guessing this change could only be incorporated into v2.0.)

I’m also CC’ing Matt and Jeff at uAvionix — I ran into this issue while parsing MAVLink messages from their ADS-B receiver, they might have more insight than me concerning the ADS-B spec. (Matt and Jeff, I found your emails in the uAvionix.xml file.)

The FAA-required uncertainty characterizations are:

- the Navigation Accuracy Category for Position (NACp), 4 bits (values 0-11)
- the Navigation Accuracy Category for Velocity (NACv), 3 bits (values 0-4)
- the Navigation Integrity Category (NIC), 4 bits (values 0-11)
- the Source Integrity Level (SIL), 2 bits (values 0-3)
- the System Design Assurance (SDA), 2 bits (values 0-3)
- (Total = 15 bits)

I’m assuming we wouldn’t need to add any flags saying eg “there is a NACp value for this message” but otherwise, that would be an additional 5 bits, one for each metric.

I’ve been referencing an FAA AC and a public MITRE report:

http://www.faa.gov/documentLibrary/media/Advisory_Circular/AC%2020-165.pdf
https://www.mitre.org/sites/default/files/pdf/10_3026.pdf

Regards,
-Marc

marclecerf:
Hey Marc,

Thanks for looping Matt and myself into the conversation.

We've received similar requests from other sources and we're considering
adding a few new MAVLink2 messages to expand the information we're able to
[...]

PR #1009, opened 2018-10-22
UTM_GLOBAL_POSITION message

eyeam3:
Introduction of UTM_GLOBAL_POSITION as discussed in https://github.com/mavlink/rfcs/pull/10.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4872,22 +4872,22 @@
       <description>The global position resulting from GPS and sensor fusion.</description>
       <field type="uint64_t" name="time" units="us">Time of applicability of position (microseconds since UNIX epoch).</field>
       <field type="uint8_t[18]" name="uas_id">Unique UAS ID.</field>
-      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84), in degrees * 1E7.</field>
-      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84), in degrees * 1E7.</field>
-      <field type="int32_t" name="alt" units="mm">Altitude (WGS84), in meters * 1000.</field>
-      <field type="int32_t" name="relative_alt" units="mm">Altitude above ground in meters * 1000.</field>
-      <field type="int16_t" name="vx" units="cm/s">Ground X speed (latitude, positive north), expressed as m/s * 100.</field>
-      <field type="int16_t" name="vy" units="cm/s">Ground Y speed (longitude, positive east), expressed as m/s * 100.</field>
-      <field type="int16_t" name="vz" units="cm/s">Ground Z speed (altitude, positive down), expressed as m/s * 100.</field>
-      <field type="uint16_t" name="h_acc" units="mm">Horizontal position uncertainty in meters * 1000. (standard deviation)</field>
-      <field type="uint16_t" name="v_acc" units="mm">Altitude uncertainty in meters * 1000. (standard deviation)</field>
-      <field type="uint16_t" name="vel_acc" units="cm/s">Speed uncertainty in m/s * 100. (standard deviation)</field>
-      <field type="int32_t" name="next_lat" units="degE7">Next waypoint, latitude (WGS84), in degrees * 1E7.</field>
-      <field type="int32_t" name="next_lon" units="degE7">Next waypoint, longitude (WGS84), in degrees * 1E7.</field>
-      <field type="int32_t" name="next_alt" units="mm">Next waypoint, altitude (WGS84), in meters * 1000.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (WGS84)</field>
+      <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
+      <field type="int16_t" name="vx" units="cm/s">Ground X speed (latitude, positive north)</field>
+      <field type="int16_t" name="vy" units="cm/s">Ground Y speed (longitude, positive east)</field>
+      <field type="int16_t" name="vz" units="cm/s">Ground Z speed (altitude, positive down)</field>
+      <field type="uint16_t" name="h_acc" units="mm">Horizontal position uncertainty (standard deviation)</field>
+      <field type="uint16_t" name="v_acc" units="mm">Altitude uncertainty (standard deviation)</field>
+      <field type="uint16_t" name="vel_acc" units="cm/s">Speed uncertainty (standard deviation)</field>
+      <field type="int32_t" name="next_lat" units="degE7">Next waypoint, latitude (WGS84)</field>
+      <field type="int32_t" name="next_lon" units="degE7">Next waypoint, longitude (WGS84)</field>
+      <field type="int32_t" name="next_alt" units="mm">Next waypoint, altitude (WGS84)</field>
       <field type="uint16_t" name="update_rate">Seconds * 1E2 until next update. Set to 0 if unknown or in data driven mode.</field>
-      <field type="uint8_t" name="flight_state" enum="UTM_FLIGHT_STATE">The flight state defined in enum \verb|UTM_FLIGHT_STATE|.</field>
-      <field type="uint8_t" name="flags" enum="UTM_DATA_AVAIL_FLAGS">Bitwise OR combination of the enums \verb|UTM_DATA_AVAIL_FLAGS|.</field>
+      <field type="uint8_t" name="flight_state" enum="UTM_FLIGHT_STATE">Flight state</field>
+      <field type="uint8_t" name="flags" enum="UTM_DATA_AVAIL_FLAGS">Bitwise OR combination of the data available flags.</field>
     </message>
     <message id="350" name="DEBUG_FLOAT_ARRAY">
       <description>Large debug/prototyping array. The message uses the maximum available payload for data. The array_id and name fields are used to discriminate between messages in code and in user interfaces (respectively). Do not use in production code.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3016,16 +3016,19 @@
         <description>The field time contains valid data.</description>
       </entry>
       <entry value="2" name="UTM_DATA_AVAIL_FLAGS_UAS_ID_AVAILABLE">
-        <description>The field uasId contains valid data.</description>
+        <description>The field uas_id contains valid data.</description>
       </entry>
       <entry value="4" name="UTM_DATA_AVAIL_FLAGS_POSITION_AVAILABLE">
-        <description>The fields lat, lon, alt and relativeAlt contain valid data.</description>
+        <description>The fields lat, lon and h_acc contain valid data.</description>
       </entry>
       <entry value="8" name="UTM_DATA_AVAIL_FLAGS_ALTITUDE_AVAILABLE">
-        <description>The fields lat, lon, alt and relativeAlt contain valid data.</description>
+        <description>The fields alt and v_acc contain valid data.</description>
       </entry>
-      <entry value="16" name="UTM_DATA_AVAIL_FLAGS_NEXT_WAYPOINT_AVAILABLE">
-        <description>The fields next_at, next_lon and next_alt contain valid data.</description>
+      <entry value="16" name="UTM_DATA_AVAIL_FLAGS_RELATIVE_ALTITUDE_AVAILABLE">
+        <description>The field relative_alt contains valid data.</description>
+      </entry>
+      <entry value="32" name="UTM_DATA_AVAIL_FLAGS_NEXT_WAYPOINT_AVAILABLE">
+        <description>The fields next_lat, next_lon and next_alt contain valid data.</description>
       </entry>
     </enum>
   </enums>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2998,6 +2998,36 @@
         <description>Ignore yaw rate</description>
       </entry>
     </enum>
+    <enum name="UTM_FLIGHT_STATE">
+      <description>Airborne status of UAS.</description>
+      <entry value="1" name="UTM_FLIGHT_STATE_UNKNOWN">
+        <description>The flight state can't be determined.</description>
+      </entry>
+      <entry value="2" name="UTM_FLIGHT_STATE_GROUND">
+        <description>UAS on ground.</description>
+      </entry>
+      <entry value="3" name="UTM_FLIGHT_STATE_AIRBORNE">
+        <description>UAS airborne.</description>
+      </entry>
+    </enum>
+    <enum name="UTM_DATA_AVAIL_FLAGS">
+      <description>Flags for the global position report.</description>
+      <entry value="1" name="UTM_DATA_AVAIL_FLAGS_TIME_VALID">
+        <description>The field time contains valid data.</description>
+      </entry>
+      <entry value="2" name="UTM_DATA_AVAIL_FLAGS_UAS_ID_AVAILABLE">
+        <description>The field uasId contains valid data.</description>
+      </entry>
+      <entry value="4" name="UTM_DATA_AVAIL_FLAGS_POSITION_AVAILABLE">
+        <description>The fields lat, lon, alt and relativeAlt contain valid data.</description>
+      </entry>
+      <entry value="8" name="UTM_DATA_AVAIL_FLAGS_ALTITUDE_AVAILABLE">
+        <description>The fields lat, lon, alt and relativeAlt contain valid data.</description>
+      </entry>
+      <entry value="16" name="UTM_DATA_AVAIL_FLAGS_NEXT_WAYPOINT_AVAILABLE">
+        <description>The fields next_at, next_lon and next_alt contain valid data.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="0" name="HEARTBEAT">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2998,6 +2998,45 @@
...
+      <entry value="1" name="UTM_FLIGHT_STATE_UNKNOWN">
+        <description>The flight state can't be determined.</description>
+      </entry>
+      <entry value="2" name="UTM_FLIGHT_STATE_GROUND">
+        <description>UAS on ground.</description>
+      </entry>
+      <entry value="3" name="UTM_FLIGHT_STATE_AIRBORNE">
+        <description>UAS airborne.</description>
+      </entry>
+    </enum>
```
auturgy: Suggest additional state "UTM_FLIGHT_STATE_EMERGENCY".

message_definitions/v1.0/common.xml
```diff
@@ -2998,6 +2998,45 @@
...
+      <entry value="32" name="UTM_DATA_AVAIL_FLAGS_HORIZONTAL_VELO_AVAILABLE">
+        <description>The fields vx and vy contain valid data.</description>
+      </entry>
+      <entry value="64" name="UTM_DATA_AVAIL_FLAGS_VERTICAL_VELO_AVAILABLE">
+        <description>The field vz contains valid data.</description>
+      </entry>
+      <entry value="128" name="UTM_DATA_AVAIL_FLAGS_NEXT_WAYPOINT_AVAILABLE">
+        <description>The fields next_lat, next_lon and next_alt contain valid data.</description>
+      </entry>
+    </enum>
```
auturgy: How the flags have been split up doesn't flow logically to me.  In fact some groupings make little sense at all.
Needs rework.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4860,29 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="334" name="UTM_GLOBAL_POSITION">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>The global position resulting from GPS and sensor fusion.</description>
+      <field type="uint64_t" name="time" units="us">Time of applicability of position (microseconds since UNIX epoch).</field>
+      <field type="uint8_t[18]" name="uas_id">Unique UAS ID.</field>
```
auturgy: Existing ADS-B identifier uses a single uint32_t for the ICAO address.  Acknowledging the likely need for a very large number of identifiers, what is the specific use case driving this to be an array?  Some justification is required, as this is a critical field that won't be easy to change once the use of this message proliferates.

message_definitions/v1.0/common.xml
eyeam3: Good idea, I will add the state.

message_definitions/v1.0/common.xml
eyeam3: Can you please elaborate on your comment a bit more? What split would you propose instead?

message_definitions/v1.0/common.xml
eyeam3: The choice of an array with 18 byte is motivated by the utilization of UUID from the chip (which is up to 128 bit on some architectures) plus 2 byte to distinguish different architectures. The reasoning is discussed in this issue: https://github.com/PX4/Firmware/issues/7891#issuecomment-326608612.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4860,29 @@
...
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (WGS84), in meters * 1000.</field>
+      <field type="int32_t" name="relative_alt" units="mm">Altitude above ground in meters * 1000.</field>
+      <field type="int16_t" name="vx" units="cm/s">Ground X Speed (Latitude, positive north), expressed as m/s * 100.</field>
+      <field type="int16_t" name="vy" units="cm/s">Ground Y Speed (Longitude, positive east), expressed as m/s * 100.</field>
+      <field type="int16_t" name="vz" units="cm/s">Ground Z Speed (Altitude, positive down), expressed as m/s * 100.</field>
+      <field type="uint32_t" name="h_acc" units="mm">Horizontal Position uncertainty in meters * 1000. (standard deviation)</field>
+      <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty in meters * 1000. (standard deviation)</field>
+      <field type="uint32_t" name="vel_acc" units="mm/s">Speed uncertainty in m/s * 1000. (standard deviation)</field>
```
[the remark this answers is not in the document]
eyeam3: I can make them uint16.

message_definitions/v1.0/common.xml
auturgy: When I first read through this, my brain assumed v_acc and h_acc were accelerations.  Which made no sense.  Accuracy makes more sense :)
There is relevant discussion re accuracy/uncertainty here: https://github.com/mavlink/mavlink/issues/624

message_definitions/v1.0/common.xml
auturgy: OK.  The obvious logical flaw is that the UUID is unique to the MCU of the flight controller, not the air vehicle.  Whilst it sort of works conceptually for RTF manufacturers, this not a good idea.
It changes ID's unnecessarily if you change the flight controller on a vehicle (not uncommon, although infrequent), it shifts ID's if you move the flight controller to another vehicle (again, infrequent but not uncommon), and it isn't domain controllable.
Obviously as a 128bit x 2byte number we aren't going to run out of ID's, but using the UUID is a convenient hack that forgets the actual purpose of the field: a unique identifier for the UAV.  Needs more discussion, and input from the relevant regulators.

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
auturgy: Just a further comment: having the measurement and it's related uncertainty in different units could result in implementation errors.  Not absolutely necessary, but consistency is a good thing.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4860,29 @@
       <field type="float[5]" name="delta" units="s">Bezier time horizon, set to NaN if velocity/acceleration should not be incorporated</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw, set to NaN for unchanged</field>
     </message>
+    <message id="334" name="UTM_GLOBAL_POSITION">
```
hamishwillee: Does it make sense to leave a few empty ids in case we need some new trajectory messages and want them adjacent? We don't  have any rules on this, but it is obviously easier for readers if associated messages are together, and the available message space in MAVLink 2 is large.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4860,29 @@
...
+      <description>The global position resulting from GPS and sensor fusion.</description>
+      <field type="uint64_t" name="time" units="us">Time of applicability of position (microseconds since UNIX epoch).</field>
+      <field type="uint8_t[18]" name="uas_id">Unique UAS ID.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (WGS84), in meters * 1000.</field>
+      <field type="int32_t" name="relative_alt" units="mm">Altitude above ground in meters * 1000.</field>
+      <field type="int16_t" name="vx" units="cm/s">Ground X Speed (Latitude, positive north), expressed as m/s * 100.</field>
+      <field type="int16_t" name="vy" units="cm/s">Ground Y Speed (Longitude, positive east), expressed as m/s * 100.</field>
+      <field type="int16_t" name="vz" units="cm/s">Ground Z Speed (Altitude, positive down), expressed as m/s * 100.</field>
```
hamishwillee: not a big deal, but think about capitalisation here - it is odd and inconsistent.
ie

> Ground X Speed (Latitude, positive north), expressed as m/s * 100. 

could be

> Ground X speed (latitude, positive north), expressed as m/s * 100.

Similarly "Horizontal Position uncertainty in meters * 1000." should proabably be "Horizontal **p**osition uncertainty in meters * 1000."

thomasgubler:
@auturgy could you join the next mavlink devcall? https://mavlink.io/en/about/support.html#dev_call

message_definitions/v1.0/common.xml
eyeam3: Maybe the naming is unfortunate, I will check other messages and rename to stay consistent in MAVLink.
The discussion is very interesting, in particular the categories. I prefer to publish the information as standard deviation since it is more precise and could be converted to the categories afterwards.

message_definitions/v1.0/common.xml
eyeam3: You are absolutely correct about the flaws concerning the UUID and it might not be used in the final UTM system. As far as I know, in the foreseeable future there is no global regulation that organizes unique identifiers for UAVs, which in turn means we cannot directly adapt such a standard. Using the UUID would give us a unique identifier without requiring the developer or pilot to change anything, move the discussion forward and enable various demonstrations. 
It is important to note that the definition is considered as "work in progress" also after the merge and the message should be used with care, as there will be modifications to acknowledge outside requirements. 
To come back to the size of the field, an array with 18 byte should be large enough to accommodate any type of identifier and should have a more consistent interface compared to a uint32_t. Yet, the size of the array might change when the topic evolves.
I understand that such modifications can break existing implementations and we have to avoid to scatter the identifier over to many different types.

message_definitions/v1.0/common.xml
eyeam3: I am open to change the id and would follow the best practice from MAVLink. Do we have a recommendation here?

message_definitions/v1.0/common.xml
eyeam3: Noted, I will adjust the capitalization.

auturgy:
@thomasgubler I’ve been to almost every one :)
I can’t guarantee the next call - I’ll be traveling - but will try.
I attend the first call. Feel free to tag me in the minutes if I don’t make it.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4860,29 @@
...
+      <field type="uint32_t" name="h_acc" units="mm">Horizontal Position uncertainty in meters * 1000. (standard deviation)</field>
+      <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty in meters * 1000. (standard deviation)</field>
+      <field type="uint32_t" name="vel_acc" units="mm/s">Speed uncertainty in m/s * 1000. (standard deviation)</field>
+      <field type="int32_t" name="next_lat" units="degE7">Next waypoint, Latitude (WGS84), in degrees * 1E7.</field>
```
WickedShell: No DNU value if there isn't a target waypoint?

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4860,29 @@
...
+      <field type="int32_t" name="next_lon" units="degE7">Next waypoint, Longitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="next_alt" units="mm">Next waypoint, Altitude (WGS84), in meters * 1000.</field>
+      <field type="uint16_t" name="update_rate">Seconds * 1E2 until next update. Set to 0 if unknown or in data driven mode.</field>
+      <field type="uint8_t" name="flight_state">The flight state defined in enum \verb|UTM_FLIGHT_STATE|.</field>
+      <field type="uint8_t" name="flags">Bitwise OR combination of the enums \verb|UTM_DATA_AVAIL_FLAGS|.</field>
```
WickedShell: Tag the field with `enum="UTM_DATA_AVAIL_FLAGS"`

message_definitions/v1.0/common.xml
WickedShell: Missed the flags.

eyeam3:
@hamishwillee I addressed the comments from this PR and from the MAVLink call and rebased onto current master.

Please have a look and let me know if there is anything I missed.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4866,29 @@
...
     </message>
+    <message id="340" name="UTM_GLOBAL_POSITION">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>The global position resulting from GPS and sensor fusion.</description>
+      <field type="uint64_t" name="time" units="us">Time of applicability of position (microseconds since UNIX epoch).</field>
+      <field type="uint8_t[18]" name="uas_id">Unique UAS ID.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (WGS84), in meters * 1000.</field>
```
hamishwillee: For `units` attributes generally we shouldn't need to state what units are in the description. So here 
```
<field type="int32_t" name="lat" units="degE7">Latitude (WGS84).</field>
      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84).</field>
      <field type="int32_t" name="alt" units="mm">Altitude (WGS84).</field>
      <field type="int16_t" name="vx" units="cm/s">Ground X speed (latitude, positive north).</field>
...
```
@auturgy @amilcarlucas Are there any exceptions? 
The only one I think we might consider would be when the description states a standard that is always expected to be in a particular format. So if (for example) WGS84 mandates metres be used for altitude, then it might make sense to hightlight the non standard format. Don't think that applies here.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4866,29 @@
...
+      <field type="int16_t" name="vz" units="cm/s">Ground Z speed (altitude, positive down), expressed as m/s * 100.</field>
+      <field type="uint16_t" name="h_acc" units="mm">Horizontal position uncertainty in meters * 1000. (standard deviation)</field>
+      <field type="uint16_t" name="v_acc" units="mm">Altitude uncertainty in meters * 1000. (standard deviation)</field>
+      <field type="uint16_t" name="vel_acc" units="cm/s">Speed uncertainty in m/s * 100. (standard deviation)</field>
+      <field type="int32_t" name="next_lat" units="degE7">Next waypoint, latitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="next_lon" units="degE7">Next waypoint, longitude (WGS84), in degrees * 1E7.</field>
+      <field type="int32_t" name="next_alt" units="mm">Next waypoint, altitude (WGS84), in meters * 1000.</field>
+      <field type="uint16_t" name="update_rate">Seconds * 1E2 until next update. Set to 0 if unknown or in data driven mode.</field>
+      <field type="uint8_t" name="flight_state" enum="UTM_FLIGHT_STATE">The flight state defined in enum \verb|UTM_FLIGHT_STATE|.</field>
```
hamishwillee: @eyeam3 

> The flight state defined in enum \verb|UTM_FLIGHT_STATE|.

TO

> Flight state.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4866,29 @@
...
+      <field type="uint8_t" name="flags" enum="UTM_DATA_AVAIL_FLAGS">Bitwise OR combination of the enums \verb|UTM_DATA_AVAIL_FLAGS|.</field>
```
hamishwillee: Hi @eyeam3 

description to 

> Bitmap of display availability flags.

Add attribute  `display="bitmask"`  (see [here for explanation](https://mavlink.io/en/guide/xml_schema.html#messages)

Do any fields have a default value you'd like to seed? You can specify that with `default=""`

hamishwillee:
@eyeam3 Thanks. A few more comments sorry!

message_definitions/v1.0/common.xml
eyeam3: I am not sure that this applies here. The purpose of this field is more for the logic of UTM and less for displaying purposes. Moreover, the available flags can change during the flight, e.g. when switching into mission mode, the UTM_DATA_AVAIL_FLAGS_NEXT_WAYPOINT_AVAILABLE flag would toggle.
And with regards to the explanation, it is not a checkbox that the user could select in the ground station, but defined by the FC software based on the flight state, available sensors and estimation.

Because of this, there are also no default values.

eyeam3:
@hamishwillee Thanks for the feedback. I addressed most of your additional comments, except for the `display` and `default` (see explanation above).

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4866,29 @@
...
+      <field type="int16_t" name="vz" units="cm/s">Ground Z speed (altitude, positive down)</field>
+      <field type="uint16_t" name="h_acc" units="mm">Horizontal position uncertainty (standard deviation)</field>
+      <field type="uint16_t" name="v_acc" units="mm">Altitude uncertainty (standard deviation)</field>
+      <field type="uint16_t" name="vel_acc" units="cm/s">Speed uncertainty (standard deviation)</field>
+      <field type="int32_t" name="next_lat" units="degE7">Next waypoint, latitude (WGS84)</field>
+      <field type="int32_t" name="next_lon" units="degE7">Next waypoint, longitude (WGS84)</field>
+      <field type="int32_t" name="next_alt" units="mm">Next waypoint, altitude (WGS84)</field>
+      <field type="uint16_t" name="update_rate">Seconds * 1E2 until next update. Set to 0 if unknown or in data driven mode.</field>
+      <field type="uint8_t" name="flight_state" enum="UTM_FLIGHT_STATE">Flight state</field>
+      <field type="uint8_t" name="flags" enum="UTM_DATA_AVAIL_FLAGS">Bitwise OR combination of the enum \verb|UTM_DATA_AVAIL_FLAGS|.</field>
```
hamishwillee: @eyeam3 YOu missed this one :-). Specifically, enum is already referenced, and I don't think the system/rendering will like `\verb|UTM_DATA_AVAIL_FLAGS|.`

> Bitwise OR combination of the data available flags.

message_definitions/v1.0/common.xml
hamishwillee: Fair enough on both of these, but as per other post, would be good if the description did not include the \verb ENum syntax.

eyeam3: Okay, makes sense.

message_definitions/v1.0/common.xml
eyeam3: I changed the comment.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4866,29 @@
...
+      <field type="int16_t" name="vx" units="cm/s">Ground X speed (latitude, positive north)</field>
+      <field type="int16_t" name="vy" units="cm/s">Ground Y speed (longitude, positive east)</field>
+      <field type="int16_t" name="vz" units="cm/s">Ground Z speed (altitude, positive down)</field>
+      <field type="uint16_t" name="h_acc" units="mm">Horizontal position uncertainty (standard deviation)</field>
+      <field type="uint16_t" name="v_acc" units="mm">Altitude uncertainty (standard deviation)</field>
+      <field type="uint16_t" name="vel_acc" units="cm/s">Speed uncertainty (standard deviation)</field>
+      <field type="int32_t" name="next_lat" units="degE7">Next waypoint, latitude (WGS84)</field>
+      <field type="int32_t" name="next_lon" units="degE7">Next waypoint, longitude (WGS84)</field>
+      <field type="int32_t" name="next_alt" units="mm">Next waypoint, altitude (WGS84)</field>
+      <field type="uint16_t" name="update_rate">Seconds * 1E2 until next update. Set to 0 if unknown or in data driven mode.</field>
```
amilcarlucas: is the update rate in centi-seconds ?

message_definitions/v1.0/common.xml
amilcarlucas: I like the current descriptions, they do not include units, just directions and reference frames. I am fine with that.

message_definitions/v1.0/common.xml
amilcarlucas: display="bitmask" also applies when data gets displayed, that way the GCS can display that multiple bits are active at the same time, and can explain which ones are.

message_definitions/v1.0/common.xml
```diff
@@ -4821,6 +4866,29 @@
...
+      <field type="uint8_t" name="flight_state" enum="UTM_FLIGHT_STATE">Flight state</field>
+      <field type="uint8_t" name="flags" enum="UTM_DATA_AVAIL_FLAGS">Bitwise OR combination of the data available flags.</field>
```
amilcarlucas: Please add display="bitmask" on this one

message_definitions/v1.0/common.xml
eyeam3: The field has only 16 bit and I don't want to increase its size. Given the size limit, I would like to leave the possibility to set a next update 10 minutes in the future and that would leave less precision for the next update, but I think this would be less important.

message_definitions/v1.0/common.xml
eyeam3: Did you see the comment above regarding the same tag? Is it then still required here?

message_definitions/v1.0/common.xml
hamishwillee: @amilcarlucas Were you asking in order to get correct units? Unfortunately we don't define this unit - perhaps we should? https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L69

In any case I think this reason is good.

hamishwillee:
@eyeam3 As agreed in Dev meeting I'm going to merge this now. Thanks for fixing the additional feedback. Remember that this is wip so you could well have to modify it again before final acceptance :-)

message_definitions/v1.0/common.xml
amilcarlucas: My comment is that the cs units needs to be added to pymavlink and used here.
Please do a PR adding it here
https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd
and I will merge that ASAP.

After that the `units="cs"` can be added here, and the text `Seconds * 1E2` should be replaced with `time`

message_definitions/v1.0/common.xml
amilcarlucas: yes, it is

eyeam3:
Yes, it is WIP and I expect changes in the future both from the community and from changes in requirements.
Thanks for all the feedback!

eyeam3:
@amilcarlucas I will follow up in a few minutes and make another PR.

message_definitions/v1.0/common.xml
hamishwillee: I pushed it through because it is WIP and there is usually significant lag on getting changes in. Thanks so much for sorting this.

PR #1010, opened 2018-10-22
[ASLUAV dialect] add GSM_LINK message

mantelt:
GSM_LINK_STATUS message to report GSM link status.
atm tailored for use with HUAWEI E3372 LTE stick


hamishwillee:
@amilcarlucas @LorenzMeier @auturgy What are the rules for merging a change to a dialect file? The tests pass and there are no ID clashes so IMO there are only potentially two reasons to refuse to merge such a change:
1. The poster is not from the organisation "owning" the dialect.
2. The message is generically useful and should go in the common messages.

I presume that there is no reason not to just merge this one (even if we want the message we'd put it in the MAVLink 2 range)?

auturgy:
I’m really not qualified to have a real opinion in this. With that caveat out of the way....
On the basic premise of “does it help the author/user and not break anything”, I think it should be merged.
It does have a more general use case, but in the present form is a little too specific for Common without some work.  I suggest that’s a separate discussion and shouldn’t hold this up.

Regards,

James

mantelt:
I think a lot of people could benefit from something like we're doing, but to make this work in a general case, a lot of work is needed IMHO. basically every 4G modem reports the RSSI differently. This is why we stick to our dialect and hope @LorenzMeier (or the person who is responsible to merge things) is OK with that.

hamishwillee:
@auturgy I like that approach. Let's give  @LorenzMeier a day or two to comment because I'd like to hear his opinion on the "general rule". After that I'm happy to merge this. 

> My only concern here is that it uses up a Mavlink1 ID.

This is actually not a big concern. Dialects are allowed to use ids from 180 - 229 for MAVLink 1. 
@mantelt if you guys are using MAVLink 2 you might be better off using a number in the MAVLink 2 range though. 

@mantelt Thanks. Is it going to be a problem for you if we keep this open a couple of days?

mantelt:
@hamishwillee thanks for looking into this PR. 
on the msgID: I can change the ID if this would be needed. But as you stated, the range 180 to 229 is reserved for dialect specific IDs, so I don't see an issue there. 

So far changes to our dialect have been accepted by @LorenzMeier quiet quickly. As we would like to continue the work on our GSM rssi feedback, I would appreciate if this PR could be merge asap. 
If needed, I could also change the ID to MAVLink 2 range. What would be the range for dialect specific messages then?

PR #1012, opened 2018-10-29
Clarify global frames

hamishwillee:
The MAVLink frames typically use the WGS84 ellipsoid for horizontal co-ordinates (either scaled for use in INT or not) and and some other altitude reference. The current descriptions were confusing because the initial part of the description did not always mention the altitude was not part of the position frame. ie

> **Global coordinate frame, WGS84 coordinate system**. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).

This PR changes the first part of the description to always mention the altitude reference:

> Global (WGS84) coordinate frame **+ MSL altitude**. ...

When the inputs are scaled this information is added (allowing the _INT to be differentiated)

> Global (WGS84) coordinate frame **(scaled)** + MSL altitude. ...

Also simplifies the text describing relative altitudes. 

> Global (WGS84) coordinate frame + **altitude relative to the home position**.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -578,7 +578,7 @@
     </enum>
     <enum name="MAV_FRAME">
       <entry value="0" name="MAV_FRAME_GLOBAL">
-        <description>Global coordinate frame, WGS84 coordinate system. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
+        <description>Global (WGS84) coordinate frame + MSL altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="1" name="MAV_FRAME_LOCAL_NED">
         <description>Local coordinate frame, Z-down (x: north, y: east, z: down).</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
> Is it relative to home location or to EKF origin ?????

Hi @amilcarlucas - Is "what" relative to EKF origin ? Everything local should be relative to home.

All this change does is make it very clear what frame we are talking about in the initial description. It should not change any meaning of the messages.

@LorenzMeier Would be good if you could review this change too.

amilcarlucas:
Are the relative frames relative to (x,y) home position or to (x,y) EKF position ?
Are the relative altitudes relative to (z) home altitude or to (z) EKF altitude ?

amilcarlucas:
I am aware that this PR just tries to document the existing status.
I am just trying to make sure that it fully reflects the way the current code works.

hamishwillee:
Hi @amilcarlucas, I don't really know what the EKF position or altitude would mean, but I do know that not every system uses EKF. We can't change a definition to something else (e.g. from "home") without agreement of the main stakeholders. And that is why I am confused, because you know that, so I assume that you must be suggesting something else.

amilcarlucas:
What I am suggesting here is the input from @tridge

hamishwillee:
@amilcarlucas I'm sorry, but that doesn't help. I don't understand what is required, or why it makes sense. Perhaps @tridge can explain it to me rather than we play Chinese whispers :-).

amilcarlucas:
I am not 100% sure how Ardupilot interprets these fields. And that is why I would like @tridge, @WickedShell or @rmackay9 to confirm that the changes you did are OK from an Ardupilot point of view.

hamishwillee:
> I am not 100% sure how Ardupilot interprets these fields. And that is why I would like @tridge, @WickedShell or @rmackay9 to confirm that the changes you did are OK from an Ardupilot point of view.

Thanks for clarification.

This is IMO a case where "point of view" is kind or irrelevant, because the PR doesn't change the previous meaning (it just makes it more obvious). i.e. if ArduPilot or PX4 do something different then they were non-compliant before and would still be non compliant. That is a much bigger discussion than this PR.

Always really happy to have feedback from @tridge, @WickedShell or @rmackay9 - in particular if you guys have a better way of wording this. I plan to merge this next time it comes to my attention even if no one bothers to comment because the current text is pretty poor and I have no real doubt of the validity of this change.

hamishwillee:
This was discussed in the [dev meeting](https://github.com/mavlink/mavlink/wiki/20181212-Dev-Meeting#call-1) and agreed can merge. The reason is that this change simply makes the current spec easier to read - it doesn't change the meaning.

There was some discussion about whether a system should use the position where EKF2 gets a valid position instead of home. If that is required it is an actual change to the meaning, and would need to be a separate and justified PR or RFC.

PR #1015, opened 2018-10-31
Ci plus

khancyr:
reformat script with https://www.shellcheck.net/
remove submodules update from test.sh : this is always handle by CI jobs. It will allow to have smaller CI image to run mavlink test by removing explicite git dependency.
add CI variable to test.sh to notify that we use a CI system the same way TRAVIS one is done.


hamishwillee:
@khancyr Can you please update the branch. @LorenzMeier This OK to merge?

PR #1017, opened 2018-11-01
Work on video streams, camera zoom and focusing

dogmaphobic:
#### Zoom
If a camera sets `CAMERA_CAP_FLAGS_HAS_BASIC_ZOOM` in its capabilities flag, it tells us it can handle basic zoom functions and supports the `MAV_CMD_SET_CAMERA_ZOOM` command.

For the `MAV_CMD_SET_CAMERA_ZOOM` command, you either use Step (one step at a time) or continuous zooming (start/stop zooming). For step zoom, you define if its Tele or Wide using `param1` (-1 for wide or 1 for tele). The camera will move the zoom back or forward one step. For continuous zoom, you send the `MAV_CMD_SET_CAMERA_ZOOM` command with `param2` set to -1 to start zooming out (wide) or 1 for zooming in (tele). To stop zooming, you send a new `MAV_CMD_SET_CAMERA_ZOOM` command with `param2` set to 0.

#### Focus
The same idea applies to basic focusing. If a camera sets `CAMERA_CAP_FLAGS_HAS_BASIC_FOCUS` in its capabilities flag, it tells us it can handle basic focusing and supports the `MAV_CMD_SET_CAMERA_FOCUS` command, which behaves in the same way as the zoom command above but handling focusing instead.

#### Tweaks to video streaming messages and commands.
For all video streaming messages and commands, I've renamed "Camera" with "Stream". You are processing streams and not cameras. A single camera may contain multiple streams.

I've added the stream count and FOV to the `VIDEO_STREAM_INFORMATION` message. The count tells you how many `VIDEO_STREAM_INFORMATION` messages you should expect. The FOV allows the ground station to overlay multiple video streams adjusting the size to match the FOV of each stream.

I don't quite get the `SET_VIDEO_STREAM_SETTINGS` message. There is no support for ACK from the camera. If the message is lost, the ground station doesn't really have any way to tell. I'm implementing all of the above in QGC with the exception of `SET_VIDEO_STREAM_SETTINGS` until this is sorted out better.

For all the above, suggestions and criticism are welcome. More of the former :)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2950,7 +2950,7 @@
         <description>Continuous zoom up/down until stopped (-1 for wide, 1 for tele, 0 to stop zooming)</description>
       </entry>
       <entry value="2" name="ZOOM_TYPE_RANGE">
-        <description>Zoom value as proportion of full camera range (a value between 0.0 and 1.0)</description>
+        <description>Zoom value as proportion of full camera range (a value between 0.0 and 100.0)</description>
       </entry>
     </enum>
     <enum name="SET_FOCUS_TYPE">
@@ -2962,7 +2962,7 @@
         <description>Continuous focus up/down until stopped (-1 for focusing in, 1 for focusing out towards infinity, 0 to stop focusing)</description>
       </entry>
       <entry value="2" name="FOCUS_TYPE_RANGE">
-        <description>Zoom value as proportion of full camera range (a value between 0.0 and 1.0)</description>
+        <description>Zoom value as proportion of full camera range (a value between 0.0 and 100.0)</description>
       </entry>
     </enum>
     <enum name="PARAM_ACK">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1592,12 +1592,16 @@
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+      <wip/>
+        <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
         <description>Set camera zoom. Use NAN for reserved values.</description>
         <param index="1">Step zoom up/down (-1 for wide, 1 for tele, 0 for no action)</param>
         <param index="2">Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
+      <wip/>
+        <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
         <description>Set camera focus. Use NAN for reserved values.</description>
         <param index="1">Step focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for no action)</param>
         <param index="2">Continuous focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for stop focusing)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1592,7 +1592,7 @@
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
-      <wip/>
+        <wip/>
         <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
         <description>Set camera zoom. Use NAN for reserved values.</description>
         <param index="1">Step zoom up/down (-1 for wide, 1 for tele, 0 for no action)</param>
@@ -1600,7 +1600,7 @@
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
-      <wip/>
+        <wip/>
         <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
         <description>Set camera focus. Use NAN for reserved values.</description>
         <param index="1">Step focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for no action)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4693,8 +4693,8 @@
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="mode_id" enum="CAMERA_MODE">Camera mode</field>
       <extensions/>
-      <field type="float" name="zoomLevel">Current zoom level (0.0 to 1.0)</field>
-      <field type="float" name="focusLevel">Current focus level (0.0 to 1.0)</field>
+      <field type="float" name="zoomLevel">Current zoom level (0.0 to 100.0)</field>
+      <field type="float" name="focusLevel">Current focus level (0.0 to 100.0)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1594,7 +1594,7 @@
       <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Set camera zoom. Use NAN for reserved values.</description>
+        <description>Set camera zoom. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
         <param index="1">Zoom type (See SET_ZOOM_TYPE enum)</param>
         <param index="2">Zoom value</param>
         <param index="3">Reserved (all remaining params)</param>
@@ -1602,9 +1602,9 @@
       <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Set camera focus. Use NAN for reserved values.</description>
-        <param index="1">Step focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for no action)</param>
-        <param index="2">Continuous focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for stop focusing)</param>
+        <description>Set camera focus. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
+        <param index="1">Focus type (See SET_FOCUS_TYPE enum)</param>
+        <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
@@ -2899,12 +2899,12 @@
       </entry>
     </enum>
     <enum name="CAMERA_CAP_FLAGS">
-      <description>Camera capability flags (Bitmap).</description>
+      <description>Camera capability flags (Bitmap)</description>
       <entry value="1" name="CAMERA_CAP_FLAGS_CAPTURE_VIDEO">
-        <description>Camera is able to record video.</description>
+        <description>Camera is able to record video</description>
       </entry>
       <entry value="2" name="CAMERA_CAP_FLAGS_CAPTURE_IMAGE">
-        <description>Camera is able to capture images.</description>
+        <description>Camera is able to capture images</description>
       </entry>
       <entry value="4" name="CAMERA_CAP_FLAGS_HAS_MODES">
         <description>Camera has separate Video and Image/Photo modes (MAV_CMD_SET_CAMERA_MODE)</description>
@@ -2926,23 +2926,35 @@
       </entry>
     </enum>
     <enum name="VIDEO_STREAM_STATUS_FLAGS">
-      <description>Stream status flags (Bitmap).</description>
+      <description>Stream status flags (Bitmap)</description>
       <entry value="1" name="VIDEO_STREAM_STATUS_FLAGS_RUNNING">
-        <description>Stream is active (running).</description>
+        <description>Stream is active (running)</description>
       </entry>
       <entry value="2" name="VIDEO_STREAM_STATUS_FLAGS_THERMAL">
-        <description>Stream is thermal imaging.</description>
+        <description>Stream is thermal imaging</description>
       </entry>
     </enum>
     <enum name="SET_ZOOM_TYPE">
-      <description>Zoom types for MAV_CMD_SET_CAMERA_ZOOM.</description>
+      <description>Zoom types for MAV_CMD_SET_CAMERA_ZOOM</description>
       <entry value="0" name="ZOOM_TYPE_STEP">
-        <description>Zoom one step increment (-1 for wide, 1 for tele).</description>
+        <description>Zoom one step increment (-1 for wide, 1 for tele)</description>
       </entry>
       <entry value="1" name="ZOOM_TYPE_CONTINUOUS">
         <description>Continuous zoom up/down until stopped (-1 for wide, 1 for tele, 0 to stop zooming)</description>
       </entry>
       <entry value="2" name="ZOOM_TYPE_RANGE">
+        <description>Zoom value as proportion of full camera range (a value between 0.0 and 1.0)</description>
+      </entry>
+    </enum>
+    <enum name="SET_FOCUS_TYPE">
+      <description>Focus types for MAV_CMD_SET_CAMERA_FOCUS</description>
+      <entry value="0" name="FOCUS_TYPE_STEP">
+        <description>Focus one step increment (-1 for focusing in, 1 for focusing out towards infinity).</description>
+      </entry>
+      <entry value="1" name="FOCUS_TYPE_CONTINUOUS">
+        <description>Continuous focus up/down until stopped (-1 for focusing in, 1 for focusing out towards infinity, 0 to stop focusing)</description>
+      </entry>
+      <entry value="2" name="FOCUS_TYPE_RANGE">
         <description>Zoom value as proportion of full camera range (a value between 0.0 and 1.0)</description>
       </entry>
     </enum>
@@ -4628,6 +4640,9 @@
       <description>Settings of a camera, can be requested using MAV_CMD_REQUEST_CAMERA_SETTINGS.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="mode_id" enum="CAMERA_MODE">Camera mode</field>
+      <extensions/>
+      <field type="float" name="zoomLevel">Current zoom level (0.0 to 1.0)</field>
+      <field type="float" name="focusLevel">Current focus level (0.0 to 1.0)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4693,8 +4693,8 @@
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="mode_id" enum="CAMERA_MODE">Camera mode</field>
       <extensions/>
-      <field type="float" name="zoomLevel">Current zoom level (0.0 to 100.0)</field>
-      <field type="float" name="focusLevel">Current focus level (0.0 to 100.0)</field>
+      <field type="float" name="zoomLevel">Current zoom level (0.0 to 100.0, NaN if not known)</field>
+      <field type="float" name="focusLevel">Current focus level (0.0 to 100.0, NaN if not known)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2926,8 +2926,6 @@
       </entry>
     </enum>
     <enum name="VIDEO_STREAM_STATUS_FLAGS">
-      <wip/>
-      <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments -->
       <description>Stream status flags (Bitmap).</description>
       <entry value="1" name="VIDEO_STREAM_STATUS_FLAGS_RUNNING">
         <description>Stream is active (running).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4720,7 +4720,7 @@
       <field type="uint16_t" name="resolution_v" units="pix">Vertical resolution</field>
       <field type="uint32_t" name="bitrate" units="bits/s">Bit rate in bits per second</field>
       <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise</field>
-      <field type="uint16_t" name="fov" units="deg">Field of view</field>
+      <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
       <field type="char[160]" name="uri">Video stream URI</field>
     </message>
     <message id="270" name="SET_VIDEO_STREAM_SETTINGS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1596,7 +1596,7 @@
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Set camera zoom. Use NAN for reserved values.</description>
         <param index="1">Zoom type (See SET_ZOOM_TYPE enum)</param>
-        <param index="2">Zoom Value</param>
+        <param index="2">Zoom value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
@@ -2937,13 +2937,13 @@
     <enum name="SET_ZOOM_TYPE">
       <description>Zoom types for MAV_CMD_SET_CAMERA_ZOOM.</description>
       <entry value="0" name="ZOOM_TYPE_STEP">
-        <description>Step zoom up/down (-1 for wide, 1 for tele)</description>
+        <description>Zoom one step increment (-1 for wide, 1 for tele).</description>
       </entry>
       <entry value="1" name="ZOOM_TYPE_CONTINUOUS">
-        <description>Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</description>
+        <description>Continuous zoom up/down until stopped (-1 for wide, 1 for tele, 0 to stop zooming)</description>
       </entry>
       <entry value="2" name="ZOOM_TYPE_RANGE">
-        <description>Zoom range (a value between 0 and 1)</description>
+        <description>Zoom value as proportion of full camera range (a value between 0.0 and 1.0)</description>
       </entry>
     </enum>
     <enum name="PARAM_ACK">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1591,6 +1591,18 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+        <description>Set camera zoom. Use NAN for reserved values.</description>
+        <param index="1">Step zoom up/down (-1 for wide, 1 for tele, 0 for no action)</param>
+        <param index="2">Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</param>
+        <param index="3">Reserved (all remaining params)</param>
+      </entry>
+      <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
+        <description>Set camera focus. Use NAN for reserved values.</description>
+        <param index="1">Step focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for no action)</param>
+        <param index="2">Continuous focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for stop focusing)</param>
+        <param index="3">Reserved (all remaining params)</param>
+      </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4669,28 +4696,30 @@
...
       <field type="uint8_t" name="target_component">component ID of the target</field>
-      <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="camera_id">Stream ID (1 for first, 2 for second, etc.)</field>
       <field type="float" name="framerate" units="Hz">Frame rate (set to -1 for highest framerate possible)</field>
       <field type="uint16_t" name="resolution_h" units="pix">Horizontal resolution (set to -1 for highest resolution possible)</field>
       <field type="uint16_t" name="resolution_v" units="pix">Vertical resolution (set to -1 for highest resolution possible)</field>
       <field type="uint32_t" name="bitrate" units="bits/s">Bit rate (set to -1 for auto)</field>
       <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise (0-359 degrees)</field>
-      <field type="char[230]" name="uri">Video stream URI</field>
+      <field type="char[160]" name="uri">Video stream URI (mostly for UDP/RTP)</field>
```
LorenzMeier: This is a breaking change for anything that is deployed already. What is the rationale?

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+        <wip/>
+        <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+        <description>Set camera zoom. Use NAN for reserved values.</description>
```
hamishwillee: I understand the concept of "not a number" but what does this mean in terms of what data/bytes gets sent in a message? The send APIs take values for every param, so what do you actually set?

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
...
+        <param index="1">Step zoom up/down (-1 for wide, 1 for tele, 0 for no action)</param>
+        <param index="2">Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</param>
```
hamishwillee: FYI, it is now possible to set an enum of possible values for a param: https://mavlink.io/en/guide/xml_schema.html#param

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+        <wip/>
+        <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+        <description>Set camera zoom. Use NAN for reserved values.</description>
+        <param index="1">Step zoom up/down (-1 for wide, 1 for tele, 0 for no action)</param>
```
hamishwillee: why not values 0, 1, 2 etc?

message_definitions/v1.0/common.xml
```diff
@@ -2902,6 +2918,21 @@
...
         <description>Camera has image survey mode (MAV_CMD_SET_CAMERA_MODE)</description>
       </entry>
+      <entry value="64" name="CAMERA_CAP_FLAGS_HAS_BASIC_ZOOM">
+        <description>Camera has basic zoom control (MAV_CMD_SET_CAMERA_ZOOM)</description>
+      </entry>
+      <entry value="128" name="CAMERA_CAP_FLAGS_HAS_BASIC_FOCUS">
+        <description>Camera has basic focus control (MAV_CMD_SET_CAMERA_FOCUS)</description>
+      </entry>
+    </enum>
+    <enum name="STREAM_STATUS_FLAGS">
```
hamishwillee: 1. Should these be WIP too?
2. Is this intended for all streams of anything, or just video? If it is for just video, does it need more specific name?

message_definitions/v1.0/common.xml
```diff
@@ -4669,28 +4700,30 @@
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Information about video stream</description>
-      <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
-      <field type="uint8_t" name="status">Current status of video streaming (0: not running, 1: in progress)</field>
+      <field type="uint8_t" name="stream_id">Stream ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="count">Number of streams available</field>
```
hamishwillee: Fullstop on description.

message_definitions/v1.0/common.xml
hamishwillee: Fair enough - but I assume you are still syncing with @lbegani/camera API people? 

FYI, I have a PR in progress in pymavlink that should strip wip messages out of a library build by default.

message_definitions/v1.0/common.xml
mhkabir: IEEE754 actually defines values for NaN.

hamishwillee: @mhkabir Thanks. Pragmatically though, is this just saying that you need to (in Python, say) pass in float('NaN') for reserved values?

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
...
+        <param index="2">Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</param>
+        <param index="3">Reserved (all remaining params)</param>
+      </entry>
+      <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
+        <wip/>
+        <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+        <description>Set camera focus. Use NAN for reserved values.</description>
+        <param index="1">Step focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for no action)</param>
+        <param index="2">Continuous focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for stop focusing)</param>
+        <param index="3">Reserved (all remaining params)</param>
```
hamishwillee: @amilcarlucas "most" MAV_CMDs take 7 params, and specify all of the unused as reserved. Some of these just specify index 1, 2, 3. Is there any implication to *not* actually specifying the additional params? Looking at the code there shouldn't be - the params don't generate any code for C. They do generate an entry for Python, but AFAIK this is just for documentation purposes:
```
MAV_CMD_SET_CAMERA_FOCUS = 532 # Set camera focus. Use NAN for reserved values.
enums['MAV_CMD'][532] = EnumEntry('MAV_CMD_SET_CAMERA_FOCUS', '''Set camera focus. Use NAN for reserved values.''')
enums['MAV_CMD'][532].param[1] = '''Step focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for no action)'''
enums['MAV_CMD'][532].param[2] = '''Continuous focus up/down (-1 for focusing in, 1 for focusing out towards infinity, 0 for stop focusing)'''
enums['MAV_CMD'][532].param[3] = '''Reserved (all remaining params)'''
```

message_definitions/v1.0/common.xml
dogmaphobic: The idea of using NaN is for a theoretical future proofing. If one day we use those parameters for something else, old software would still send NaN and new software would then ignore it.

message_definitions/v1.0/common.xml
dogmaphobic: I left these "open" with the idea that the sign indicates a direction (negative goes wide and positive goes tele). It is possible then to have a "rate". If you start continuous zoom with -2 instead, it would go faster than if using -1. I'm not sure about that yet. For now, within the camera handler, I simply check for < 0, 0 and > 0.

message_definitions/v1.0/common.xml
dogmaphobic: See above

message_definitions/v1.0/common.xml
mhkabir: Correct.

message_definitions/v1.0/common.xml
dogmaphobic: I guess it wouldn't hurt to flag them as WIP, even though they are inherently so given that the message that uses it is marked as such.

Yes, you are right. It should be a bit more specific as this is Video Streaming only. I will change that.

message_definitions/v1.0/common.xml
dogmaphobic: Done (removed only instance with a full stop)

message_definitions/v1.0/common.xml
dogmaphobic: It looks like enums don't allow `<wip>`.

message_definitions/v1.0/common.xml
amilcarlucas: I'm going to leave this one for @tridge to answer.

message_definitions/v1.0/common.xml
hamishwillee: OK. Then you can't :-)

message_definitions/v1.0/common.xml
hamishwillee: Makes sense if this is potentially a continuum, rather than an enumeration value.

hamishwillee:
I've sent a slack message to ask @lbegani to engage with this. Do you have any idea when you might want to think about these messages moving no longer wip? (shouldn't be WIP forever).

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
```
hamishwillee: this can specify `enum = "FLIGHT_INFORMATION"` now.

hamishwillee: this can specify `enum = "CAMERA_MODE"` now.

message_definitions/v1.0/common.xml
hamishwillee: Fair enough. I sent Lalit message on slack. @julianoes any comments?

julianoes:
FYI: @657793213

dogmaphobic:
>Do you have any idea when you might want to think about these messages moving no longer wip? (shouldn't be WIP forever).

Once we have enough input and no more changes, I will implement the QGC side of this. Once that's done, it will no longer be "WIP".

message_definitions/v1.0/common.xml
julianoes: I'm ok with this change. We'll have to change it in the SDK but I doubt this is used yet.

dogmaphobic:
> These changes are ok but from what I know we also need an API to set zoom as a factor, e.g. `1x` to `10x`.

Agreed but we don't know what the possible range is. Some cameras are 1x to 30x. How about instead use a 0 - 1 range and let the camera fit that within its actual range? If you truly need specific values (such as 2x, 8x or 15mm, 20mm, etc.), which are not knowable by the GS, you should use extended parameters.

Note that normally, 2x, 3x, etc. are only used by consumer cameras. For professional cameras, these imply digital magnification. Change in (optical) focal length is usually specified by an actual focal length in mm. A camera may have a variable focal length lens (zoom lens) AND a _Digital Zoom_ feature. You would have in this case two ranges. An optical range in mm and a digital range (magnification) expressed as 2x, 3x, etc. This usually only starts once you reach the end of the optical range.

We could use something like this, where we move the type of zoom to an enum and have a single value field:

```
    <enum name="SET_ZOOM_TYPE">
      <description>.</description>
      <entry value="0" name="ZOOM_TYPE_STEP">
        <description>Step zoom up/down (-1 for wide, 1 for tele)</description>
      </entry>
      <entry value="1" name="ZOOM_TYPE_CONTINUOUS">
        <description>Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</description>
      </entry>
      <entry value="2" name="ZOOM_TYPE_RANGE">
        <description>Zoom range (a value between 0 and 1)</description>
      </entry>
    </enum>


    <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
        <wip/>
        <description>Set camera zoom. Use NAN for reserved values.</description>
        <param index="1">Zoom type (See SET_ZOOM_TYPE enum)</param>
        <param index="2">Zoom Value</param>
    </entry>
```

julianoes:
> How about instead use a 0 - 1 range and let the camera fit that within its actual range?

That could work.

message_definitions/v1.0/common.xml
```diff
@@ -2934,6 +2934,18 @@
         <description>Stream is thermal imaging.</description>
       </entry>
     </enum>
+    <enum name="SET_ZOOM_TYPE">
```
hamishwillee: @dogmaphobic So just FMI user calls MAV_CMD_SET_CAMERA_ZOOM with one of these and a value. 
- For ZOOM_TYPE_STEP they would call this as many times as they wanted to step through values, with a value of -1 or 1.
- For ZOOM_TYPE_CONTINUOUS they would call it once with -1 or 1, and then again later when ready with 0 (to stop zooming)
- For ZOOM_TYPE_RANGE they would call it once with a value between 0 and 1, indicating the zoom value as a proportion of the total range.


Is that all correct?

Can they find out what the current zoom setting is? That would seem useful if you're displaying it somewhere or want to know that you've hit the limit.

Is there some return value to indicate that a setting is invalid (and is that needed?)

dogmaphobic: > * For ZOOM_TYPE_STEP they would call this as many times as they wanted to step through values, with a value of -1 or 1.
> * For ZOOM_TYPE_CONTINUOUS they would call it once with -1 or 1, and then again later when ready with 0 (to stop zooming)

Correct. The sign indicates the direction of zoom travel.

> * For ZOOM_TYPE_RANGE they would call it once with a value between 0 and 1, indicating the zoom value as a proportion of the total range.

Correct, and the camera would compute the actual range based on its capabilities.

> Can they find out what the current zoom setting is? That would seem useful if you're displaying it somewhere or want to know that you've hit the limit.

No, not currently. There are two options: A specific, new set of request/response for the current zoom/focus settings (values) or add as an extension to [CAMERA_SETTINGS](
https://mavlink.io/en/messages/common.html#CAMERA_SETTINGS)

>Is there some return value to indicate that a setting is invalid (and is that needed?)

That comes from using commands. All commands receive a response based on the [MAV_RESULT](
https://mavlink.io/en/messages/common.html#MAV_RESULT) enum. Yes, if something is out of bounds or not supported, it gets a proper error response.

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Set camera zoom. Use NAN for reserved values.</description>
+        <param index="1">Zoom type (See SET_ZOOM_TYPE enum)</param>
+        <param index="2">Zoom Value</param>
```
hamishwillee: lower case "V" in value

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Set camera zoom. Use NAN for reserved values.</description>
+        <param index="1">Zoom type (See SET_ZOOM_TYPE enum)</param>
```
hamishwillee: This can be set as `<param index="1" enum="SET_ZOOM_TYPE">Zoom type</param>`.

Maybe: `<param index="1" enum="SET_ZOOM_TYPE">Zoom type (e.g. step increment, continuous, specific value)</param>`

message_definitions/v1.0/common.xml
```diff
@@ -2902,6 +2918,33 @@
...
+    <enum name="SET_ZOOM_TYPE">
+      <description>Zoom types for MAV_CMD_SET_CAMERA_ZOOM.</description>
+      <entry value="0" name="ZOOM_TYPE_STEP">
+        <description>Step zoom up/down (-1 for wide, 1 for tele)</description>
+      </entry>
+      <entry value="1" name="ZOOM_TYPE_CONTINUOUS">
+        <description>Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</description>
+      </entry>
+      <entry value="2" name="ZOOM_TYPE_RANGE">
+        <description>Zoom range (a value between 0 and 1)</description>
```
hamishwillee: How about "Zoom value as proportion of full camera range (a value between 0 and 1)"

message_definitions/v1.0/common.xml
```diff
@@ -2902,6 +2918,33 @@
...
+        <description>Stream is active (running).</description>
+      </entry>
+      <entry value="2" name="VIDEO_STREAM_STATUS_FLAGS_THERMAL">
+        <description>Stream is thermal imaging.</description>
+      </entry>
+    </enum>
+    <enum name="SET_ZOOM_TYPE">
+      <description>Zoom types for MAV_CMD_SET_CAMERA_ZOOM.</description>
+      <entry value="0" name="ZOOM_TYPE_STEP">
+        <description>Step zoom up/down (-1 for wide, 1 for tele)</description>
```
hamishwillee: How about:

> Zoom step increment (-1 for wide, 1 for tele).

message_definitions/v1.0/common.xml
```diff
@@ -2902,6 +2918,33 @@
...
+      </entry>
+      <entry value="1" name="ZOOM_TYPE_CONTINUOUS">
+        <description>Continuous zoom up/down (-1 for wide, 1 for tele, 0 for stop zooming)</description>
```
hamishwillee: Maybe:

> Continuous zoom up/down until stopped (-1 for wide, 1 for tele, 0 to stop zooming)

message_definitions/v1.0/common.xml
hamishwillee: Thanks very much. 
1. I have tried to suggest enum text that better captures the above. Not sure how successfully - you can be judge of whether it is improvement or not. 

> No, not currently. There are two options: A specific, new set of request/response for the current zoom/focus settings (values) or add as an extension to CAMERA_SETTINGS

If it were me, I would extend CAMERA_SETTINGS. Rather than forcing clients to request this I would also require of the protocol that this message is automatically emitted immediately after any increment change to the zoom value. ie whenever MAV_CMD_SET_CAMERA_ZOOM is called, and as the camera goes through "increments" while in continuous zoom.

message_definitions/v1.0/common.xml

message_definitions/v1.0/common.xml
dogmaphobic: I don't think `param` accepts an `enum` attribute. All other references to enums follow the format I used above.

message_definitions/v1.0/common.xml
dogmaphobic: Changed to `Zoom value as proportion of full camera range (a value between 0.0 and 1.0)` to reinforce the idea that this value is a floating point value.

message_definitions/v1.0/common.xml

message_definitions/v1.0/common.xml

message_definitions/v1.0/common.xml
dogmaphobic: Added as full comment below to make it easier to track.

dogmaphobic:
* Specified that both MAV_CMD_SET_CAMERA_ZOOM and MAV_CMD_SET_CAMERA_FOCUS commands will return a CAMERA_SETTINGS message, which now includes the current value for both.
* Added current focus and zoom values (as an extension) to the CAMERA_SETTINGS message.
* Made MAV_CMD_SET_CAMERA_FOCUS consistent with MAV_CMD_SET_CAMERA_ZOOM, also using an enum for the different types.
* Removed a few spurious full stop periods.

message_definitions/v1.0/common.xml
hamishwillee: The [schema](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L152) was updated recently to allow enums and other value information. This was to allow the toolchain and QGC to do things like provide a selection list for params that take enums.

That said QGC does nothing with this yet. There is a small negative that I don't get add this to the wiki documentation, and I suspect that neither does mavgen.

Upshot, I'd prefer this was updated to the new format, but can't really justify pushing it.

message_definitions/v1.0/common.xml
```diff
@@ -1594,17 +1594,17 @@
       <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Set camera zoom. Use NAN for reserved values.</description>
+        <description>Set camera zoom. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
```
hamishwillee: This works for me. If a user wants to monitor the changing increment on continuous zoom I guess they can send the "request message" for the settings.

Do we need constraints on the continuous zoom to ensure that it is actually possible/practical for a user to control the zoom with message? The problem I envisage is that user sets zoom through QGC and 0.1s later the system hits max zoom (or visa versa).

amilcarlucas:
Maybe it would make sense to use a bit more of the new schema tags, not only the ENUM tag ?

message_definitions/v1.0/common.xml
dogmaphobic: Continuous zoom is meant for mechanical, optical zoom. It takes time to go from one end to the other. If the camera only supports digital zoom, you would only use step up/down commands or specific magnification values.

message_definitions/v1.0/common.xml
dogmaphobic: That link above points to Ardupilot. Unless the schema change is added to our repo, I can't really use it.

hamishwillee: You use MAVLink/mavlink repo right? That has the above repo as a submodule. It is the official and only schema used by mavgen.

hamishwillee: And I'm going to backtrack on "Upshot, I'd prefer this was updated to the new format, but can't really justify pushing it."

I'd really like you to use the new schema. They are a "good thing" and this is a good test case to make us start using them properly.

message_definitions/v1.0/common.xml
hamishwillee: Good fix. We'll have to update the [guide docs](https://mavlink.io/en/services/camera.html) too at some point.

message_definitions/v1.0/common.xml
dogmaphobic: That was due to my not knowing where that schema was and how this is all integrated. Yes, I will certainly use it, I just didn't know it was available.

message_definitions/v1.0/common.xml
```diff
@@ -1591,6 +1591,22 @@
...
+        <param index="1">Zoom type (See SET_ZOOM_TYPE enum)</param>
+        <param index="2">Zoom value</param>
+        <param index="3">Reserved (all remaining params)</param>
+      </entry>
+      <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Set camera focus. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
+        <param index="1">Focus type (See SET_FOCUS_TYPE enum)</param>
```
amilcarlucas: Please remove the enum from the description and ass it as an XML tag.

amilcarlucas:
Here are the list of valid `param` attributes:
 `
<xs:attribute name="index" type="xs:unsignedByte"/>                    
<xs:attribute name="label" type="xs:string"/>                          
<xs:attribute name="units" type="xs:string"/>                          
<xs:attribute name="decimalPlaces" type="xs:unsignedByte"/>            
<xs:attribute name="increment" type="xs:float"/>                       
<xs:attribute name="minValue" type="xs:float"/>                        
<xs:attribute name="maxValue" type="xs:float"/>                        
`

message_definitions/v1.0/common.xml
dogmaphobic: I've added an "enum" attribute as requested, though I don't quite understand the list of valid `param` attributes below as it doesn't show "enum" as one of them.

hamishwillee: That's because it isn't the full list/best list to show. This is the definition of param:
```
<!-- definition of complex elements -->
<xs:element name="param">
    <xs:complexType mixed="true">
        <xs:attribute ref="index" use="required"/>
        <xs:attribute ref="label"/>
        <xs:attribute ref="units"/>
        <xs:attribute ref="enum" />
        <xs:attribute ref="decimalPlaces"/>
        <xs:attribute ref="increment"/>
        <xs:attribute ref="minValue"/>
        <xs:attribute ref="maxValue"/>
    </xs:complexType>
</xs:element>
```
All of them are optional except index. The description of purpose is [documented here](https://mavlink.io/en/guide/xml_schema.html#param). I would guess you might want to supply label too and possibly max/min values.

hamishwillee:
@WickedShell Any interest in the camera API?

LorenzMeier:
@dogmaphobic rebase please

message_definitions/v1.0/common.xml
```diff
@@ -2902,6 +2918,45 @@
...
+      <entry value="64" name="CAMERA_CAP_FLAGS_HAS_BASIC_ZOOM">
+        <description>Camera has basic zoom control (MAV_CMD_SET_CAMERA_ZOOM)</description>
+      </entry>
+      <entry value="128" name="CAMERA_CAP_FLAGS_HAS_BASIC_FOCUS">
+        <description>Camera has basic focus control (MAV_CMD_SET_CAMERA_FOCUS)</description>
+      </entry>
+    </enum>
+    <enum name="VIDEO_STREAM_STATUS_FLAGS">
+      <description>Stream status flags (Bitmap)</description>
+      <entry value="1" name="VIDEO_STREAM_STATUS_FLAGS_RUNNING">
```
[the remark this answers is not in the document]
dogmaphobic: It's a bitmap. If the bit is set, it's running. If it's not set, it's not running.

message_definitions/v1.0/common.xml
```diff
@@ -2902,6 +2918,45 @@
...
+    <enum name="SET_ZOOM_TYPE">
+      <description>Zoom types for MAV_CMD_SET_CAMERA_ZOOM</description>
+      <entry value="0" name="ZOOM_TYPE_STEP">
+        <description>Zoom one step increment (-1 for wide, 1 for tele)</description>
+      </entry>
+      <entry value="1" name="ZOOM_TYPE_CONTINUOUS">
+        <description>Continuous zoom up/down until stopped (-1 for wide, 1 for tele, 0 to stop zooming)</description>
+      </entry>
+      <entry value="2" name="ZOOM_TYPE_RANGE">
+        <description>Zoom value as proportion of full camera range (a value between 0.0 and 1.0)</description>
```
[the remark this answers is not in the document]
dogmaphobic: Is there some pattern within the MAVLink specs for this? I usually think in terms of 0-1. Never 0-100. The former is usually what you have at the API level. Any API. 0-100 is usually just for UI.

dogmaphobic:
If everybody is happy with this current state, I will start developing both sides with this new API. Once I can verify that this is sufficient, I will remove the WIP.

message_definitions/v1.0/common.xml
```diff
@@ -1592,6 +1592,22 @@
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Set camera zoom. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
+        <param index="1" enum="SET_ZOOM_TYPE">Zoom type</param>
+        <param index="2">Zoom value</param>
```
hamishwillee: You could constrain this value from -1 to +1 using maxValue, minValue.

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: There are cases both ways and no clear rule - e.g. https://mavlink.io/en/messages/common.html#HIGH_LATENCY

Is there a way to tell the UI that this should be displayed as a percentage? If not, I _personally_ prefer percentage to be used.

hamishwillee:
@lbegani Did you see Gus response? https://github.com/mavlink/mavlink/pull/1017#issuecomment-439377699

message_definitions/v1.0/common.xml
dogmaphobic: The idea is to allow some sort of speed control. I'm still working on this. If a given lens has a lot of steps, I can control the rate of speed of change based on how fast the user "pinches" the screen in/out. That is, -1/1 is one step at a time, -2/2 two steps, etc. Yes, there should be some limit to this (-1000/1000 won't do any good) but I need to test with a few different cameras to find a usable range.

dogmaphobic: Only after I know that this works. Not only from a hardware perspective but also from an UI perspective. For now, it's just -1/1. I will experiment with larger values and only then I will know if that makes sense. If it doesn't, I will set min/max to -1/1.

dogmaphobic: To make it clearer, when you use "pinch" zoom, the UI sends a stream of values as you pinch in and out. It's way too much data to send these values as they come. I have to "round" them and only send a zoom command when a certain threshold is reached. I want to experiment with handling pinch speed. The faster you pinch in and out, the greater the value I will use -2/2, -3/3, etc.

dogmaphobic: In other words, I have to find a compromise between the rate of commands sent and the amount of zoom change based on the touch movement speed.

lbegani:
> > 1. Now there will be two ways of setting zoom/focus. Will setting of zoom/focus using camera protocol reflect in the QGC UI which is based on basic zoom/focus mavlink messages.
> 
> What is the _other_ way? The idea of these command is to standardize focus and zoom. The only time you would have these through parameters would be in case you have something that falls outside the normal and need something custom. That would imply you also need custom code within the ground station to deal with it, in which case, you would handle it that way. Can you think of something that is "normal" and doesn't fit within these commands?
> 
By two ways I meant - a) Zoom features set using zoom commands b) Zoom features exported by camera definition file and set using set/get param commands. If a & b are mutually exclusive, it will work fine, otherwise QGC UI must be updated for changes due to (b)
> > 2. Related to the first comment, setting of some advanced parameter using camera protocol may affect zoom/focus value set. How will the vehicle notify QGC about the change in value(unsolicited)
> 
> Good question. I can't off the top of my head think of any such scenario but it's not impossible. The extended parameter definition handles these through rules expressed within the XML file but those only apply to other extended parameters. Maybe add rules that can handle commands as well? Something like: _if this parameter change, this command must be updated_?
> 
I guess automatic modes of camera(like Macro/Portrait/Landscape/Night etc) plays with all the parameters of Camera. May be adding a rule to refresh the params would work. Not sure how QGC zoom/focus UI component will be connected with these rules.

> > 3. For video streaming, is there a difference in sequence of messages sent between rtsp vs udp protocol.
> 
> Can you elaborate on this? Those two protocols are quite different. RTSP is a client initiated connection and RTP/UDP is a server initiated connection. For RTSP, the client needs an URI to connect to. For RTP/UDP, the server needs an URI to connect to. The sequence is inherently inverted.

As RTSP and UDP are different, the sequence of message flow will differ. I guess the messages that are defined will suffice for both. Explanation can happen in documentation.

lbegani:
> If everybody is happy with this current state, I will start developing both sides with this new API. Once I can verify that this is sufficient, I will remove the WIP.

I think we can merge the changes, corner cases can be addressed as we hit them.

dogmaphobic:
> If a & b are mutually exclusive, it will work fine

Yes, they are mutually exclusive. I decided against the "common" parameters for these because it felt they belonged to a dedicated command instead. For Zoom and Focus, only weird cases would need to be implemented through extended parameters and those will require custom code within QGC to handle it.

> I guess automatic modes of camera(like Macro/Portrait/Landscape/Night etc) plays with all the parameters of Camera. May be adding a rule to refresh the params would work. Not sure how QGC zoom/focus UI component will be connected with these rules.

Normally only Macro could have an impact and even then is not normal. Not to mention having a macro lens on a drone is a bit outside the "normal" in its own right. With that said, adding rules to handle commands is not that complicated and it would solve this.

LorenzMeier:
@dogmaphobic Could you please rebase? Thanks!

dogmaphobic:
>Could you please rebase? Thanks!

It already is. I did it earlier today.

dogmaphobic:
Merging for now. Leaving it as WIP until I know for sure this works as supposed to.

TSC21:
This is breaking `sitl_gazebo` builds: https://travis-ci.org/PX4/sitl_gazebo/jobs/458320375. The `gazebo_geotagged_images_plugin` requires an update on the source code so to handle the two new fields of `CAMERA_SETTINGS` message.

dogmaphobic:
> This is breaking `sitl_gazebo` builds

https://github.com/PX4/sitl_gazebo/pull/250

PR #1018, opened 2018-11-05
Clarify that PARAM_SET is a permanent write

hamishwillee:
The previous definition indicates that the write was temporary and that the setting would not survive reboot. It also indicated that you could send MAV_ACTION_STORAGE_WRITE to write the value to permanent storage, but this value does not appear to exist.

On PX4, PARAM_SET writes the value to RAM and then immediately to permanent storage (a cached write).

This update is to reflect the actual behaviour.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3084,7 +3084,7 @@
       <field type="uint16_t" name="param_index">Index of this onboard parameter</field>
     </message>
     <message id="23" name="PARAM_SET">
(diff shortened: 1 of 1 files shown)


hamishwillee:
@auturgy This is true for ArduPilot too right? I think the definition just way out of date.

hamishwillee:
> True for ArduPilot. Might be worth seeing if you can check with SmartAP, SLUGS, and whatever other consumers of mavlink are out there (cameras?)

FYI @auturgy I did send out some emails on this but got no response yet. Alternative was to make ambiguous about whether params are permanently stored or make it clear that this is for temporary writes and make PX4 and ArduPilot implementations non-compliant. Both of those seem less useful. 

It does mean that if there is a valid use case for temporary writes, we will need to invent something new. 
That would be a good candidate for an extension.

PR #1021, opened 2018-11-11
Make MAV_CMDs own section of messages with field markup

hamishwillee:
Enum schema now allows params to have markup like fields that define possible values, units, and display hints to GCS - e.g.: enum value, max/min/increment, units, label, display with X decimal points.

There is no really clean way to display all these fields in the current layout (see https://github.com/mavlink/mavlink/pull/1019 for best shot).

This PR does two things.
1. It splits MAV_CMD into its own group as a peer of messages and enums
image
2. It mimics the message format to allow columns for units and values.
image
But these are only displayed if the fields are defined. If not, it looks like:
image

The only question is where should label and decimal point fields go. I have currently put them in the description because those two don't matter to all users (they mostly matter to GCS authors).
Is that good?

You can play with this here:
https://hamishwillee.gitbooks.io/ham_mavdevguide/content/v/just_messages/en/messages/common.html

### DIFF
--- a/doc/mavlink_gitbook.py
+++ b/doc/mavlink_gitbook.py
@@ -22,8 +22,10 @@ xsl_file_name = "mavlink_to_html_table_gitbook.xsl"
 xml_message_definitions_dir_name = "../message_definitions/v1.0/"
 
 output_dir = "./messages/"
-if not os.path.exists(output_dir):
-    os.makedirs(output_dir)
+output_dir_html=output_dir+"_html/"
+if not os.path.exists(output_dir_html):
+    os.makedirs(output_dir_html)
+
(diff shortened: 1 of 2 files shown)


hamishwillee:
@LorenzMeier @auturgy @amilcarlucas FYI, this update IMO allows us to properly support the new markup schema for MAV_CMDs by rendering them similarly to messages. 
This is I guess "controversial" because they are really a kind of enum. However I think highlighting MAV_CMDs separately is a really good idea. 
Your thoughts on this welcome.

hamishwillee:
I am merging. It doesn't break anything else and is IMO a move forward.

PR #1023, opened 2018-11-15
Follow up PR for UTM_GLOBAL_POSITION

eyeam3:
There have been requested changes open from #1009. This is the follow up PR to address them as well.

- Add centisecond as unit to update_rate field
- Add missing display tag

Related to ArduPilot/pymavlink#242


hamishwillee:
@eyeam3 Thanks for being so fast. There doesn't seem to be a process for automatically updating the submodule dependency. I've created a PR for this (see link above).

hamishwillee:
@eyeam3 I've pulled in latest Pymavlink and tests now pass - so this can be approved. Can you confirm nothing else needs to change/that this can be merged?

eyeam3:
Should be good to go. These were the open requests from the previous PR and the discussion around it.

hamishwillee:
Merging as this has previously been reviewed/approved, it is a WIP message, and it builds without error.

PR #1026, opened 2018-11-19
add additional component_id fields for obstacle avoidance and VIO

baumanta:
To monitor individual processes on the companion computer, the component ID enum has to be extended to include the process types which should be monitored over the heartbeat meassage.


baumanta:
FYI @TSC21

hamishwillee:
FYI I'm going to add comments to MAV_COMPONENT in the next day or so. 
I know you "followed the pattern", but as a general rule, even if not enforced by the system, every enum and enum value should have a description in order to define the intention and reduce the change of mis-use.

PR #1028, opened 2018-11-21
Add MAV_COMPONENT documentation

hamishwillee:
As part of general push to get all common.xml entities documented, this pushes some proposed text for MAV_COMPONENT. The descriptions should cover how and when each value is used, and any exceptions. Ideally we should attempt to restrict use and intent so it is clear how it should be used and extended, and remove ambiguity that might later result in different uses. 

I have put in a bunch of placeholders for discussion. Please highlight anything you think should be more precise, or could cause potential issue.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -369,81 +369,83 @@
       </entry>
     </enum>
     <enum name="MAV_COMPONENT">
-      <description>Component ids for the different types and instances of onboard hardware/software that might make up a MAVLink system (autopilot, cameras, servos, GPS systems, avoidance systems etc.). Where possible, the appropriate ID should be used for the target/source address of each type of component. The MAV_COMP_ID_ALL value is used to indicate messages that must be processed by all components. New ids may be created for any component that must be separately addressed. Where possible, values for components that can have multiple instances (e.g. cameras, servos etc.) should be allocated sequentially (and gaps left in the allocation rage to allow the number of instances to be expanded).</description>
+      <description>Component ids for the different types and instances of onboard hardware/software that might make up a MAVLink system (autopilot, cameras, servos, GPS systems, avoidance systems etc.). 
+      Components must use the appropriate ID in their source address when sending messages and use the same ID to determine if they are the intended recipient of an incoming message. The MAV_COMP_ID_ALL value is used to indicate messages that must be processed by all components.
+      When creating new entries for this enum, values for components that can have multiple instances (e.g. cameras, servos etc.) should be allocated sequentially (and gaps left in the allocation rage to allow the number of instances to be expanded).</description>
       <entry value="0" name="MAV_COMP_ID_ALL">
         <description>Used to broadcast messages to all components of the receiving system. Components should attempt to process messages with this component ID and forward to components on any other interfaces.</description>
       </entry>
       <entry value="1" name="MAV_COMP_ID_AUTOPILOT1">
-        <description>System flight controller ("autopilot") component. Only one autopilot is expected in a particular system.</description>
+        <description>System flight controller component ("autopilot"). Only one autopilot is expected in a particular system.</description>
       </entry>
       <entry value="100" name="MAV_COMP_ID_CAMERA">
-        <description>Camera ID #1.</description>
+        <description>Camera #1.</description>
       </entry>
       <entry value="101" name="MAV_COMP_ID_CAMERA2">
-        <description>Camera ID #2.</description>
+        <description>Camera #2.</description>
       </entry>
       <entry value="102" name="MAV_COMP_ID_CAMERA3">
-        <description>Camera ID #3.</description>
+        <description>Camera #3.</description>
       </entry>
       <entry value="103" name="MAV_COMP_ID_CAMERA4">
-        <description>Camera ID #4.</description>
+        <description>Camera #4.</description>
       </entry>
       <entry value="104" name="MAV_COMP_ID_CAMERA5">
-        <description>Camera ID #5.</description>
+        <description>Camera #5.</description>
       </entry>
       <entry value="105" name="MAV_COMP_ID_CAMERA6">
-        <description>Camera ID #6.</description>
+        <description>Camera #6.</description>
       </entry>
       <entry value="140" name="MAV_COMP_ID_SERVO1">
-        <description>Servo ID #1.</description>
+        <description>Servo #1.</description>
       </entry>
       <entry value="141" name="MAV_COMP_ID_SERVO2">
-        <description>Servo ID #2.</description>
+        <description>Servo #2.</description>
       </entry>
       <entry value="142" name="MAV_COMP_ID_SERVO3">
-        <description>Servo ID #3.</description>
+        <description>Servo #3.</description>
       </entry>
       <entry value="143" name="MAV_COMP_ID_SERVO4">
-        <description>Servo ID #4.</description>
+        <description>Servo #4.</description>
       </entry>
       <entry value="144" name="MAV_COMP_ID_SERVO5">
-        <description>Servo ID #5.</description>
+        <description>Servo #5.</description>
       </entry>
       <entry value="145" name="MAV_COMP_ID_SERVO6">
-        <description>Servo ID #6.</description>
+        <description>Servo #6.</description>
       </entry>
       <entry value="146" name="MAV_COMP_ID_SERVO7">
-        <description>Servo ID #7.</description>
+        <description>Servo #7.</description>
       </entry>
       <entry value="147" name="MAV_COMP_ID_SERVO8">
-        <description>Servo ID #8.</description>
+        <description>Servo #8.</description>
       </entry>
       <entry value="148" name="MAV_COMP_ID_SERVO9">
-        <description>Servo ID #9.</description>
+        <description>Servo #9.</description>
       </entry>
       <entry value="149" name="MAV_COMP_ID_SERVO10">
-        <description>Servo ID #10.</description>
+        <description>Servo #10.</description>
       </entry>
       <entry value="150" name="MAV_COMP_ID_SERVO11">
-        <description>Servo ID #11.</description>
+        <description>Servo #11.</description>
       </entry>
       <entry value="151" name="MAV_COMP_ID_SERVO12">
-        <description>Servo ID #12.</description>
+        <description>Servo #12.</description>
       </entry>
       <entry value="152" name="MAV_COMP_ID_SERVO13">
-        <description>Servo ID #13.</description>
+        <description>Servo #13.</description>
       </entry>
       <entry value="153" name="MAV_COMP_ID_SERVO14">
-        <description>Servo ID #14.</description>
+        <description>Servo #14.</description>
       </entry>
       <entry value="154" name="MAV_COMP_ID_GIMBAL">
-        <description>Generic Gimbal ID. Use ID for specific gimbal type if one exists (e.g. MAV_COMP_ID_QX1_GIMBAL).</description>
+        <description>Gimbal component.</description>
       </entry>
       <entry value="155" name="MAV_COMP_ID_LOG">
-        <description>Logging component ID.</description>
+        <description>Logging component.</description>
       </entry>
       <entry value="156" name="MAV_COMP_ID_ADSB">
-        <description>Automatic Dependent Surveillance-Broadcast (ADS-B) component ID.</description>
+        <description>Automatic Dependent Surveillance-Broadcast (ADS-B) component.</description>
       </entry>
       <entry value="157" name="MAV_COMP_ID_OSD">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
-        <description/>
+        <description>Servo ID #12.</description>
       </entry>
       <entry value="152" name="MAV_COMP_ID_SERVO13">
-        <description/>
+        <description>Servo ID #13.</description>
       </entry>
       <entry value="153" name="MAV_COMP_ID_SERVO14">
-        <description/>
+        <description>Servo ID #14.</description>
```
hamishwillee: Just FYI, if there will ever be more servos, it is inconvenient that the next value has been used.

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
-        <description/>
+        <description>Gimbal ID for QX1.</description>
       </entry>
       <entry value="160" name="MAV_COMP_ID_FLARM">
-        <description/>
+        <description>ID for FLARM collision alert component.</description>
       </entry>
       <entry value="180" name="MAV_COMP_ID_MAPPER">
-        <description/>
+        <description></description>
```
hamishwillee: What is this for?

hamishwillee:
@auturgy @WickedShell FYI.

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
       </entry>
       <entry value="190" name="MAV_COMP_ID_MISSIONPLANNER">
-        <description/>
+        <description>ID for a component that supports the Mission microservice.</description>
```
hamishwillee: In enum description I suggest components should use the most appropriate value, which is fine for cameras, servos, whatever. But what should an autopilot do for a case like this? It supports missions, but is not separately addressed for that purpose. You could similarly consider this case for a system that supports obstacle avoidance etc.

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
       </entry>
       <entry value="195" name="MAV_COMP_ID_PATHPLANNER">
-        <description/>
+        <description></description>
```
hamishwillee: @mrivi @baumanta How would you characterise a PATHPLANNER component for someone who did not know of your work? - "It is a component that ..." Presumably this is for the companion side code so you can address just a particular node with your messages?

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
       </entry>
       <entry value="196" name="MAV_COMP_ID_OBSTACLE_AVOIDANCE">
-        <description/>
+        <description></description>
```
hamishwillee: @mrivi @baumanta Ditto for MAV_COMP_ID_OBSTACLE_AVOIDANCE

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
-        <description/>
+        <description>GPS #1.</description>
       </entry>
       <entry value="221" name="MAV_COMP_ID_GPS2">
-        <description/>
+        <description>GPS #2.</description>
       </entry>
       <entry value="240" name="MAV_COMP_ID_UDP_BRIDGE">
-        <description/>
+        <description></description>
```
hamishwillee: A bridge connects two transports - is a UDP bridge TO UDP, and what is it from. Same question for MAV_COMP_ID_UART_BRIDGE below

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
       </entry>
       <entry value="241" name="MAV_COMP_ID_UART_BRIDGE">
-        <description/>
+        <description></description>
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
-        <description/>
+        <description></description>
```
hamishwillee: What does this component do?

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
-        <description/>
+        <description>Servo ID #14.</description>
       </entry>
       <entry value="154" name="MAV_COMP_ID_GIMBAL">
-        <description/>
+        <description>Generic Gimbal ID. Use ID for specific gimbal type if one exists (e.g. MAV_COMP_ID_QX1_GIMBAL).</description>
       </entry>
       <entry value="155" name="MAV_COMP_ID_LOG">
-        <description/>
+        <description>Logging component ID.</description>
```
hamishwillee: Are there any particular expectations of a logging component?

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,132 @@
...
-        <description>On Screen Display (OSD) devices for video links</description>
+        <description>On Screen Display (OSD) devices for video links.</description>
       </entry>
       <entry value="158" name="MAV_COMP_ID_PERIPHERAL">
-        <description>Generic autopilot peripheral component ID. Meant for devices that do not implement the parameter sub-protocol</description>
+        <description>Generic autopilot peripheral component ID. Meant for devices that do not implement the parameter microservice.</description>
       </entry>
       <entry value="159" name="MAV_COMP_ID_QX1_GIMBAL">
-        <description/>
+        <description>Gimbal ID for QX1.</description>
```
hamishwillee: Why does QX1 need separate addressing?

julianoes: It shouldn't in my opinion.

message_definitions/v1.0/common.xml
julianoes: Not sure :smile:

message_definitions/v1.0/common.xml
TSC21: @hamishwillee this component is currently set in MAVROS to identify the link between MAVROS and the GCS, which is made through an UDP bridge, forwards the packages between the GCS and the FCU.

auturgy:
Mavesp8266 also identifies as udp bridge (240).

message_definitions/v1.0/common.xml
```diff
@@ -148,7 +148,7 @@
         <description>VTOL reserved 5</description>
       </entry>
       <entry value="26" name="MAV_TYPE_GIMBAL">
-        <description>Onboard gimbal</description>
+        <description>Gimbal (standalone)</description>
```
hamishwillee: FYI, MAV_TYPE is for systems, not component. This was supposed to be fixed in another PR, but will combine it here.

message_definitions/v1.0/common.xml
hamishwillee: Thanks. I've marked it as deprecated. FYI @auturgy too. This is clearly wrong and there is a clear right way to do it, so I think deprecation is in order for now, ideally with deletion at some point in future.

message_definitions/v1.0/common.xml
hamishwillee: @TSC21 So to clarify, is this a component of the GCS system - so you address the message to GCS and MAV_COMP_ID_UART_BRIDGE, and this tells the GCS to forward the packet to PX4 (say). And the link is via UDP, between MAVROS and QGC (but not necessarily between QGC and PX4)?

Is the MAV_COMP_ID_UART_BRIDGE the same thing?

IN that case, what would you propose for a description, remembering that the user should not have to be ROS?

message_definitions/v1.0/common.xml
hamishwillee: Research shows it is used in [AntennaTracker](https://github.com/ArduPilot/ardupilot/blob/master/AntennaTracker/GCS_Mavlink.cpp#L379) to identify the target that should accept an arm command (MAV_CMD_COMPONENT_ARM_DISARM).
Digging a bit further, it seems the requirement to check this component was removed elsewhere from ArduPilot in [2014](https://github.com/ArduPilot/ardupilot/pull/909) and the command will accept MAV_COMP_ID_ALL as a target now. So probably should also be removed from AntennaTracker.

@auturgy I propose to deprecate this (as not used by PX4 and only accidentally by AntennaTracker) and add a description. It would look like the below (OK?)
```
      <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
        <description>ID for component to handle system commands (e.g. takeoff, arm).</description>
        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">There is no need to separately address system commands to a specific component for handling.</deprecated>
      </entry>
```

message_definitions/v1.0/common.xml
hamishwillee: @mrivi Ping :-)

message_definitions/v1.0/common.xml
hamishwillee: @auturgy I can't find this on ArduPilot, PX4 or via google in anything other than header file listings. Other than sending out messages on our forums and google discussions, do you have any ideas? 
Could I add a description "Purpose unknown" and deprecate it after doing that basic due-diligence?

message_definitions/v1.0/common.xml
hamishwillee: @auturgy noted that that *Mavesp8266 also identifies as udp bridge (240)*. [Comments](https://github.com/dogmaphobic/mavesp8266/blob/master/src/mavesp8266_component.cpp#L33) say "MavLink UART/UDP Bridge", so essentially this is used for a bridge from UART to UDP. 

WOuld that be an OK description for ROS as well?

message_definitions/v1.0/common.xml
julianoes: Yer `MAV_TYPE` is a mess. I think we should probably get rid of it. The only thing that it tells is really the frame type.

hamishwillee: Probably not - this is apparently what is used for determining the icon for airframe in QGC :-)
I think it is fine if used properly.

message_definitions/v1.0/common.xml
hamishwillee: @julianoes Any thoughts on this?

message_definitions/v1.0/common.xml
hamishwillee: Thanks. Updated!

message_definitions/v1.0/common.xml
hamishwillee: FYI, I have updated these. Please check you are happy.

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,136 @@
...
       <entry value="159" name="MAV_COMP_ID_QX1_GIMBAL">
-        <description/>
+        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_GIMBAL">All gimbals should use MAV_COMP_ID_GIMBAL.</deprecated>
+        <description>Gimbal ID for QX1.</description>
       </entry>
       <entry value="160" name="MAV_COMP_ID_FLARM">
-        <description/>
+        <description>FLARM collision alert component.</description>
       </entry>
       <entry value="180" name="MAV_COMP_ID_MAPPER">
```
hamishwillee: Lorenz says this can be deleted.

message_definitions/v1.0/common.xml
```diff
@@ -369,131 +369,136 @@
       </entry>
     </enum>
     <enum name="MAV_COMPONENT">
+      <description>Component ids for the different types and instances of onboard hardware/software that might make up a MAVLink system (autopilot, cameras, servos, GPS systems, avoidance systems etc.). 
+      Components must use the appropriate ID in their source address when sending messages and use the same ID to determine if they are the intended recipient of an incoming message. The MAV_COMP_ID_ALL value is used to indicate messages that must be processed by all components.
+      When creating new entries for this enum, values for components that can have multiple instances (e.g. cameras, servos etc.) should be allocated sequentially (and gaps left in the allocation rage to allow the number of instances to be expanded).</description>
```
hamishwillee: allocation rage -> allocation range. 
Perhaps provide guidance on what is a sensible range.

message_definitions/v1.0/common.xml
hamishwillee: No answer here. But this is IMO sufficient for now.

hamishwillee:
`MAV_COMPONENT` documentation merged as discussed in dev call.

message_definitions/v1.0/common.xml
julianoes: @hamishwillee I think your guess about the history of this is correct.

> The messages can be used by multiple systems, and you want to make it clear which system should handle it.

I don't think we need that.

I rather thought that this component ID would be for a component which plans the mission. So let's say this is a companion computer which is responsible for waypoint generation based on some higher level goal.

hamishwillee: Thanks very much! I have clarified this further in https://github.com/mavlink/mavlink/pull/1031

PR #1029, opened 2018-11-23
Update minimal.xml to use definitions in common.xml

hamishwillee:
The version in minimal.xml was quite out of date compared to common.xml. This iterates the version, and main types to match.

Questions:
1. Should this *add* `MAV_COMPONENT`. It is something you should set in all messages. I think so.
2. This is always at risk of getting out of sync. It is a good example of something that should be included into common.xml for that reason. Is this something we should consider?
3. In the docs I call this the [Heartbeat/Connection Protocol](https://mavlink.io/en/services/heartbeat.html). IF this was a microservice under your new proposal, what would you call it?


amilcarlucas:
1. yes
2. good idea
3. keep your name, I think it is a good one.

hamishwillee:
Thanks @amilcarlucas . I'll add `MAV_COMPONENT` (assume @LorenzMeier also OK with it) after https://github.com/mavlink/mavlink/pull/1028 is merged.

PR #1030, opened 2018-11-28
add timestamp to the SENS_ATMOS message

ASM3:
fix: add timestamp to the ASLUAV SENS_ATMOS message

### DIFF
--- a/message_definitions/v1.0/ASLUAV.xml
+++ b/message_definitions/v1.0/ASLUAV.xml
@@ -184,6 +184,7 @@
     </message>
     <message id="208" name="SENS_ATMOS">
       <description>Atmospheric sensors (temperature, humidity, ...) </description>
+      <field type="uint64_t" name="atmos_timestamp">Time since system start [us]</field>
       <field type="float" name="TempAmbient" units="degC"> Ambient temperature</field>
       <field type="float" name="Humidity" units="%"> Relative humidity</field>
     </message>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/ASLUAV.xml
```diff
@@ -184,6 +184,7 @@
     </message>
     <message id="208" name="SENS_ATMOS">
       <description>Atmospheric sensors (temperature, humidity, ...) </description>
+      <field type="uint64_t" name="atmos_timestamp">Time since system start [us]</field>
```
hamishwillee: Propose you add units attribute, and remove unitsfrom description. Is *atmos* in the name implied?
```
<field type="uint64_t" name="timestamp"  units="us">Time since system start.</field>
```
Just FYI, within common we standardised on timestamps like below. This is OK except the name, which really should not include the units (but which we can't change for compatibility reasons).
```
<field type="uint64_t" name="time_usec">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
```

hamishwillee:
Hi @ASM3,
I've added a comment for your attention. 
Otherwise happy to merge this based on the general rule that we merge anything that dialect owners want that causes no harm to others (ie I am assuming you are managing any compatibility issues for your users).

ASM3:
Hi @hamishwillee,  
We will probably make a code revision at some later point for aligning our dialect style with the standardized one.

hamishwillee:
> We will probably make a code revision at some later point for aligning our dialect style with the standardized one.

Probably a good idea. There are a whole lot of new features in the xml-schema - for example params can now be associated with enums and units in the same way as fields. This hasn't yet propagated through to  QGC, but it will.

PR #1033, opened 2018-12-06
common: Add a relative jump command

WickedShell:
This provides support for a variant of `MAV_CMD_DO_JUMP` that instead uses relative indexes. The motivation for this is when working with a complex mission a jump may make the most sense in a local contect (IE go back 4 waypoints), and having to remember to edit the jump target on the absolute command when you delete an earlier unassociated waypoint is a problem. Similarly it's not actually clear if a jump index should automatically be adjusted by a GCS in this scenario either. By allowing an explicit relative mode of jumping a user can plan for this case easily.

This could have been a flag on the existing jump command, but it was thought to be easier to drop into most GCS/UI's more easily as a seperate command, allowing a distinct visual styling. (It also allows a way to ensure that a mission uploaded that depends upon this command being relative will actuallly be interpreted as a relative command).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1616,6 +1616,15 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="600" name="MAV_CMD_JUMP_TAG">
+        <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
+        <param index="1">Tag.</param>
+      </entry>
+      <entry value="601" name="MAV_CMD_DO_JUMP_TAG">
+        <description>Jump to the matching tag in the mission list. Repeat this action for the specified number of times. A mission should contain a single matching tag for each jump. If this is not the case then a jump to a missing tag should complete the mission, and a jump where there are multiple matching tags should always select the one with the lowest mission sequence number.</description>
+        <param index="1">Target tag to jump to.</param>
+        <param index="2">Repeat count</param>
+      </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>


amilcarlucas:
Needs a rebase

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,16 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="1000" name="MAV_CMD_DO_JUMP_RELATIVE">
```
hamishwillee: FMI Why 1000? Maybe 550? I know there are no clear rules, but gap seems overly large.

hamishwillee: Possible alternative for your consideration:  `MAV_CMD_DO_JUMP_OFFSET`

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,16 @@
...
+        <description>Jump to the desired command in the mission list.  Repeat this action only the specified number of times</description>
```
hamishwillee: Could perhaps be more specific.

> Jump to the mission list command at the specified offset from the current mission item.  Repeat this action only the specified number of times

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,16 @@
...
+        <param index="1">Relative jump from the current sequence number. Negative indicates jump backwards, 0 is a noop. A jump to an index that is outside of the mission boundary shall not wrap, and the mission will be considered complete.</param>
```
hamishwillee: > A jump to an index that is outside of the mission boundary shall not wrap, and the mission will be considered complete.

Shouldn't the recipient reject the mission? It has the mission count, so it can work out if the sequence is invalid.

If not, then perhaps the "shall not wrap" is redundant. 

> A jump to an index that is outside of the mission boundary will complete the mission.

Also FMI does it complete with mission number at the last mission item or do systems have a "we're finished" state. Basically I'm hoping that implementations don't require the last mission item to be recorded as a valid value.

hamishwillee:
@julianoes @DonLakeFlyer @meee1 You guys might have to implement this - got any opinions?

message_definitions/v1.0/common.xml
WickedShell: I wasn't sure how many more camera related things wanted to be up there :)

hamishwillee: :-) Me either. But I'm sure we won't need 500.

message_definitions/v1.0/common.xml
WickedShell: > Shouldn't the recipient reject the mission? It has the mission count, so it can work out if the sequence is invalid.

If it's a forward jump and you are uploading a mission you don't yet know if it's invalid or not, so you can't assess at that time. You might also later trim an already uploaded mission and a previously valid jump becomes invalid. (We have the same problems with a normal jump command, I just explicitly listed what the error handling should be here rather then leaving it as an implementation decision)

> Also FMI does it complete with mission number at the last mission item or do systems have a "we're finished" state.

ArduPilot has an explicit state for being complete with a mission, and at a minimum a vehicle has to do something when it finishes the mission, so this would indicate "just do your completion action and stop following the mission"

message_definitions/v1.0/common.xml
WickedShell: Personally I'd prefer `RELATIVE` but happy to defer to consensus if others don't.

DonLakeFlyer:
Why not implement a labeled jump instead? The fact is that any sort of index based jump is crappy since it gets out of sync so easily except in the simplest of cases. Two items: One for the jump label with some sort of id, one for the jump to the label.

WickedShell:
@DonLakeFlyer I'm not opposed, but if your mission storage is on anything that doesn't respond quickly/known time the look up time of the jump index might be painful, as compared to a known relative jump. You could prefetch your jump destination labels I guess. That's really more of an implementation detail though then a protocol problem.

message_definitions/v1.0/common.xml
peterbarker: `RELATIVE` is more appropriate as it implies `OFFSET`.

`OFFSET` can be from start-of-mission, end-of-mission, current-position or most colourful mission item.

julianoes:
> You guys might have to implement this - got any opinions?

I find this actually quite hard to implement and also visualize in a ground station but maybe that's just me. If someone can do it cleanly for the Dronecode SDK I certainly won't object. This applies to both absolute and relative gotos actually.

As an alternative, I usually just use generation of the waypoints, so basically just send a loop a couple of times if that's what is required. It means that the mission progress is clear as well.

WickedShell:
> I find this actually quite hard to implement and also visualize in a ground station but maybe that's just me.

Yeah, I draw a line like a normal waypoint, but if the target isn't itself a nav target it gets problematic. I have to admit that being able to do jump tag targets is probably a lot more helpful for UI's.

> As an alternative, I usually just use generation of the waypoints, so basically just send a loop a couple of times if that's what is required. It means that the mission progress is clear as well.

@julianoes I'm not following this. For us the vehicle will jump to a landing sequence and execute it wherever it is in the mission, and ArduPilot is gaining the ability to execute planned aborts in the same manner. I want to coordinate between these, which means that unwinding the waypoints directly to avoid jumps doesn't work, as there is no fixed number of possible jumps between these, particularly when the vehicle selects the closest landing sequence. You could plan it with absolute waypoints, but it's a lot harder to keep synced correctly then a relative (or tagged jump).

LorenzMeier:
@WickedShell Given the potential for error (and the catastrophic consequences when you skip a critical part of a landing sequence), I believe the tags/markers are a lot safer. Could you check if that's workable on the Firmware side? The way I would implement it, I would store a new linked list of tags while uploading the mission with their associated index onboard. When a command to jump then comes I can go through this short list of 1-10 items, pull out the index and jump there. The list would only take a few dozen bytes for a couple of jump labels, as you just need to store a 32bit key and a 16bit jump index plus the data struct overhead.

**With this, only the firmware side of the mission handling is safety-critical. With a relative jump, the GCS needs to ensure a correct index and also becomes safety-critical, which I'd love to avoid.**

julianoes:
> I'm not following this. For us the vehicle will jump to a landing sequence and execute it wherever it is in the mission, and ArduPilot is gaining the ability to execute planned aborts in the same manner.

Oh I understand, that's what this is for. Didn't we add rally points as separate "missions" for exactly that?

WickedShell:
> Oh I understand, that's what this is for. Didn't we add rally points as separate "missions" for exactly that?

We have DO_LAND_START tags in missions already and the vehicle jumps to the nearest one. I have an open PR for ArduPilot to store the DO_GO_AROUND's in the mission and jump to the nearest one of those when an abort is complete. I'm just looking for a form of jumping to control where I go from my landing abort back into shooting the approach again (as the initial part of the approach encoded in DO_LAND_START may not make sense to fly over when resuming from an abort).


> With this, only the firmware side of the mission handling is safety-critical. With a relative jump, the GCS needs to ensure a correct index and also becomes safety-critical, which I'd love to avoid.

Yeah. Technically this is already true for all the jumps we have already.

> When a command to jump then comes I can go through this short list of 1-10 items, pull out the index and jump there. The list would only take a few dozen bytes for a couple of jump labels, as you just need to store a 32bit key and a 16bit jump index plus the data struct overhead.

It's both the key length (given that it would be using a float I think forcing a shorter key is probably acceptable), jump target, and number of jumps to that target executed. (I'm happy to have no counters, but I'm sure others would be upset to lose them). I'd also worry about about imposing a limit on this, but given that we currently scroll through the entire mission when looking for DO_LAND_START/DO_GO_AROUND I don't think it's any worse, and can always be improved later if it does become a large pain point.

Given the above discussion I'll rework this into a `JUMP_TARGET` and `DO_JUMP_TO_TARGET`. Other then a number of repeats is there any other data we want to be tagging on to it?

LorenzMeier:
@WickedShell Not that I could think of right now. Thanks for taking this into consideration, I promise I will be quick in the review turnaround.

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,15 @@
...
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="600" name="MAV_CMD_JUMP_TAG">
+        <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
+        <param index="1">Tag.</param>
+      </entry>
+      <entry value="601" name="MAV_CMD_DO_JUMP_TAG">
+        <description>Jump to the matching tag in the mission list. Repeat this action only the specified number of times. If the target tag is not found this shall be treated as if the mission commands had been completed. Handling of multiple matching tags is implementation defined.</description>
+        <param index="1">Target tag to jump to.</param>
+        <param index="2">Repeat count</param>
```
hamishwillee: Add full stop.

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,15 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="600" name="MAV_CMD_JUMP_TAG">
+        <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
+        <param index="1">Tag.</param>
+      </entry>
+      <entry value="601" name="MAV_CMD_DO_JUMP_TAG">
+        <description>Jump to the matching tag in the mission list. Repeat this action only the specified number of times. If the target tag is not found this shall be treated as if the mission commands had been completed. Handling of multiple matching tags is implementation defined.</description>
+        <param index="1">Target tag to jump to.</param>
```
hamishwillee: We should perhaps set [increments](https://mavlink.io/en/guide/xml_schema.html#param) and minimum values for all of these to 1

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,15 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="600" name="MAV_CMD_JUMP_TAG">
```
hamishwillee: _TAG - or maybe _TARGET?

message_definitions/v1.0/common.xml
```diff
@@ -1616,6 +1616,15 @@
...
+        <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
+        <param index="1">Tag.</param>
+      </entry>
+      <entry value="601" name="MAV_CMD_DO_JUMP_TAG">
+        <description>Jump to the matching tag in the mission list. Repeat this action only the specified number of times. If the target tag is not found this shall be treated as if the mission commands had been completed. Handling of multiple matching tags is implementation defined.</description>
```
hamishwillee: Why "Handling of multiple matching tags is implementation defined.".? (or rather why not state that multiple labels of same value are invalid.

message_definitions/v1.0/common.xml
WickedShell: If we want an integer increment then leveraging one of the int transported fields is better. I'm actually inclined to let this be whatever people want and not reduce the tag space. If we send it as a float lets keep it as a float, and accept negative numbers. (If you are doing stuff with a GCS you could potentially use positives in the mission and stick a unique negative tag somewhere as a database to the local GCS that associates the mission with local meta data. This is a serious stretch case, but I'd rather not overconstrain these uses out.)

message_definitions/v1.0/common.xml
WickedShell: I'd actually say jump to the first one, such that you don't have to scan for duplicates later (others may think jump to closest one in the mission). Given that I thought there was a lot of options here I was taking the cowardly approach of not trying to dictate a universal behavior. I think it's probably a malformed mission, but I wasn't going to force the issue.

Although I may have made it worse by raising the issue for discussion at all :D

hamishwillee: Definitely made it better. We need to constrain differences that can possibly result in incompatible versions. 

Isn't it easier to reject duplicates than work out which of multiples are "closer" (potentially in both directions)?

WickedShell: > Isn't it easier to reject duplicates than work out which of multiples are "closer" (potentially in both directions)?

Somewhat. The problem is you can do this before starting a mission which is fine, but doing it during mission upload is a problem. Because the state of the mission on the vehicle is undefined while uploading you get into a runtime problem of rejecting at upload time.

Consider a mission which had a tag `foo` (using names rather then numbers for clarity) at index 100 of 200 items. You delete the first 5 waypoints for some reason and upload this mission, you have to do a full mission upload as a partial can't remove, just replace or extend. An implementation may (ArduPilot will) see the new total waypoint count is 195, it will immediately drop the last 5 waypoints from the list, but retain the rest of the mission, as items are uploaded they replace the older items. Tag `foo` is now sequence 95 instead of 100. This gets dropped in the mission, and at this time there are now 2 tags of `foo` in the vehicle representation of the mission at the same time. 5 waypoints from now the next one will be removed, but until then the vehicle could be flying the mission and hit the jump.

This is a very long scenario for how to get 2 tags in, but it could very easily happen when modifying a mission.

hamishwillee: OK. I leave it as an exercise to everyone to work out what approach is best, but as above "We need to constrain differences that can possibly result in incompatible versions."

message_definitions/v1.0/common.xml
hamishwillee: I don't have a strong opinion so "fair enough"

message_definitions/v1.0/common.xml
hamishwillee: FYI Dev meeting decided your _TAG is good option.

message_definitions/v1.0/common.xml
hamishwillee: FYI, Dev meeting discussion was that ideal is that duplicate tags in mission is not desirable, and I'd add that neither is a jump command without a matching tag. GCS/Developer API  should prevent these from being uploaded or saved (and in the case of duplicate tags, ideally from being added in the first place). 

We also discussed that there are ways that an Autopilot might get a mission with these problems, so we should spec handling. The thought was that the autopilot should:
- move to a "mission completed" state from a jump to an missing tag
- iterate forward from the current item to the end and then loop up to current number, picking the first matching tag it finds. BUTbelow I suggest we always select the tag that is closest to the end of the mission - that is just a reverse search through the list, without having to worry about looping.

And we might word this as:

>  Jump to the matching tag in the mission list. Repeat this action for the specified number of times. Mission planning software should not upload missions that have duplicate tags or jump commands without matching tags. On the autopilot a jump to a missing target tag should complete the mission. A jump to a duplicate tag should always select the tag with the highest mission sequence number.

Thoughts? @WickedShell @DonLakeFlyer

hamishwillee: Ping @WickedShell @DonLakeFlyer - would https://github.com/mavlink/mavlink/pull/1033#discussion_r241277491 wording change work for you guys?

@WickedShell This needs to be rebased and if above text is OK that needs to be inserted.

WickedShell: Sorry just got back to this. I'm not a fan of the proposed wording as it's both very long/detailed from a protocol side. And more problematically "must not (where possible)" then collides with the later comments of "must not upload", and the autopilot/mission planning software feels like an overload here. I'd rather see us just specify what happens if there are multiple matching tags here.

PR #1034, opened 2018-12-06
extend distance_sensor message with field of view

mrivi:
This field is useful when using distance sensors for collision prevention.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4247,6 +4247,8 @@
       <field type="uint8_t" name="id">Onboard ID of the sensor</field>
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
       <field type="uint8_t" name="covariance" units="cm">Measurement covariance, 0 for unknown / invalid readings</field>
+      <extensions/>
+      <field type="float" name="fov" units="rad">Angle where the distance measurement is valid.</field>
     </message>
     <message id="133" name="TERRAIN_REQUEST">
       <description>Request for terrain data and terrain status</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4248,9 +4248,9 @@
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
       <field type="uint8_t" name="covariance" units="cm">Measurement covariance, 0 for unknown / invalid readings</field>
       <extensions/>
-      <field type="float" name="h_fov" units="rad">Horizontal Field of View (angle) where the distance measurement is valid.</field>
-      <field type="float" name="v_fov" units="rad">Vertical Field of View (angle) where the distance measurement is valid.</field>
-      <field type="float[4]" name="q">Quaternion of the sensor orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0). This field is required if the orientation is set to MAV_SENSOR_ROTATION_CUSTOM. Otherwise set it to NaN."</field>
+      <field type="float" name="horizontal_fov" units="rad">Horizontal Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to NaN.</field>
+      <field type="float" name="vertical_fov" units="rad">Vertical Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to NaN.</field>
+      <field type="float[4]" name="quaternion">Quaternion of the sensor orientation in vehicle body frame (w, x, y, z order, zero-rotation is 1, 0, 0, 0). Zero-rotation is along the vehicle body x-axis. This field is required if the orientation is set to MAV_SENSOR_ROTATION_CUSTOM. Otherwise set it to NaN."</field>
     </message>
     <message id="133" name="TERRAIN_REQUEST">
       <description>Request for terrain data and terrain status</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4247,6 +4247,8 @@
       <field type="uint8_t" name="id">Onboard ID of the sensor</field>
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
       <field type="uint8_t" name="covariance" units="cm">Measurement covariance, 0 for unknown / invalid readings</field>
+      <extensions/>
+      <field type="float" name="fov" units="rad">Angle where the distance measurement is valid.</field>
```
hamishwillee: Perhaps for consistency:

> Field of view (angle) where the distance measurement is valid.

A few thoughts/questions:
- I assume this defines a cone around `orientation`?
- Will/must this always be set? What if it is NaN? Basically I'm wondering what happens for all existing distance sensors [and I'm curious what your collision prevention code will do if you don't have this information].

mrivi: - yes, do I need to specify something along the lines of `objects in the range [-fov/2, fov/2] determine the sensor measurements. 0 corresponds to the sensor x-axis` ?
- we would assume a default field of view for the collision prevention library. The fov information would come from the sensor driver.

auturgy:
Not a blocker, just a comment for consideration: horizontal and vertical (or X/Y) FOV. Some sensors are asymmetric (eg https://aerotenna.com/datasheets/AerotennauSharppatchDatasheet.pdf).

LorenzMeier:
@mrivi Could you please split that in horizontal and vertical? Thanks! In terms of orientation: I would propose to add a quaternion for this and an enum value for the angle enums that says ANGLE_CUSTOM or something like that, in that case the quaternion field should be evaluated. With this we're taking care of all the corner cases, yes? Any type of variance or signal strength required as well?

We have met all these issues now often enough that I feel solid about being able to spec a message that is closer to final.

mrivi:
@LorenzMeier For what I have tested, adding quaternion orientation and field of view solves a lot of issues. As for covariance, the field is already present.

auturgy:
@rmackay9 FYI.

mrivi:
@hamishwillee  @LorenzMeier any other comments ?

message_definitions/v1.0/common.xml
```diff
@@ -4247,6 +4247,10 @@
       <field type="uint8_t" name="id">Onboard ID of the sensor</field>
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
       <field type="uint8_t" name="covariance" units="cm">Measurement covariance, 0 for unknown / invalid readings</field>
+      <extensions/>
+      <field type="float" name="h_fov" units="rad">Horizontal Field of View (angle) where the distance measurement is valid.</field>
+      <field type="float" name="v_fov" units="rad">Vertical Field of View (angle) where the distance measurement is valid.</field>
+      <field type="float[4]" name="q">Quaternion of the sensor orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0). This field is required if the orientation is set to MAV_SENSOR_ROTATION_CUSTOM. Otherwise set it to NaN."</field>
```
mhkabir: How is the zero rotation defined? With respect to vehicle body frame? This should be specified.

message_definitions/v1.0/common.xml
```diff
@@ -4247,6 +4247,10 @@
       <field type="uint8_t" name="id">Onboard ID of the sensor</field>
       <field type="uint8_t" name="orientation" enum="MAV_SENSOR_ORIENTATION">Direction the sensor faces. downward-facing: ROTATION_PITCH_270, upward-facing: ROTATION_PITCH_90, backward-facing: ROTATION_PITCH_180, forward-facing: ROTATION_NONE, left-facing: ROTATION_YAW_90, right-facing: ROTATION_YAW_270</field>
       <field type="uint8_t" name="covariance" units="cm">Measurement covariance, 0 for unknown / invalid readings</field>
+      <extensions/>
+      <field type="float" name="h_fov" units="rad">Horizontal Field of View (angle) where the distance measurement is valid.</field>
```
hamishwillee: is there any reason not to use more descriptive names? It might not matter here, but could be useful in code and doc. E.g. horizontal_fov, vertical_fov, quaternion?

hamishwillee:
@mrivi Looks pretty good to me (though see additional comments).

@auturgy @rmackay9 Do you have further feedback on this? I believe the discussion is nearing conclusion and this will be merged soon.

message_definitions/v1.0/common.xml
```diff
@@ -4247,6 +4247,10 @@
...
+      <field type="float" name="v_fov" units="rad">Vertical Field of View (angle) where the distance measurement is valid.</field>
```
hamishwillee: > we would assume a default field of view for the collision prevention library. The fov information would come from the sensor driver.

My point is that how should the library know that it must use a default field of view? The answer would be that the v_fov is set to some set of values. What are they? I expect NaN. In which case, you might change this to something like:

> Vertical Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to NaN.

Ditto for other fields. The reason this is important is that otherwise they might choose 0 (say) and users have to code for both.

mhkabir: Agree with @hamishwillee here. Also NaN is not a good choice here, since components running older Mavlink will "send" 0 for these fields, not NaN.

mrivi:
@hamishwillee @mhkabir I have addressed your comments

hamishwillee:
@mrivi What @mhkabir was suggesting is that I was wrong - the value for "not used" has to be 0 in _extension_ fields, if at all possible.

The reason for this is that if a mavlink 2 system receives a message that does not contain an extension field that it expects, then it just zero-fills the values. This allows a system to be able to handle messages from systems that have not yet been updated with the new field. 

This is obviously irritating, because you don't know if you're getting an FOV of actually 0 or an indication of not supported. Generally this is fine though - any useful FOV will be non-zero, so you can assume if you get zero that it means not supported.

Does this make sense?

hamishwillee:
And you'll have to rebase.

mrivi:
@hamishwillee any other comments?

hamishwillee:
@mrivi Looks great. @LorenzMeier @auturgy @mhkabir Any final words before this gets pushed live?

auturgy:
Looks good. Thanks @mrivi

hamishwillee:
@mrivi Do you need this in urgently? I think everyone has had a chance to comment and I am happy to merge if you do. But otherwise will wait until Friday for last minute comments.

hamishwillee:
Merged, as discussed in December. Thanks @mrivi for your patience.

PR #1036, opened 2018-12-12
Better description of outdated digicam messages (mark as fallback, in…

amilcarlucas:
…stead of deprecated)

This kind of replaces #967 with a weaker version


hamishwillee:
Thanks @amilcarlucas . The URL of the services changed, but I'll fix that up as a post push. Merging as is "pre approved" by the dev call team.

PR #1037, opened 2018-12-13
Fix up links to protocol/microservices URLs

hamishwillee:
A bunch of links moved. There are redirects, but fixing the link is better.


hamishwillee:
Self merging. This is a no brainer.

PR #1038, opened 2018-12-14
Allow bigger STATUSTEXT messages

hamishwillee:
The MAVLINK1 STATUSTEXT message allows only 50 char messages, which is restrictive. While we don't want the message abused in terms of bandwidth usage and overly long strings for display in UIs, we know that if there had been the ability to have message empty-byte truncation when the message was defined, a larger number would have been specified. 

This PR adds a second (extension) field `extra_text` that can be appended to the `text` field for longer messages:
- It recommends users should craft the `text` content to be stand alone so that if used with systems that don't understand extra_text the message can still be understood.
- QGC is moved from the description as recipient could be anything - another GCS or a developer API.

The alternative was to define a new full width message. The [dev call](https://github.com/mavlink/mavlink/wiki/20181212-Dev-Meeting) proposed this approach because we assume it retains the greatest compatibility, and may encourage users to keep messages concise. We're open to a completely new message if @meee1 and @DonLakeFlyer feel strongly that is better.

@DonLakeFlyer @auturgy @WickedShell @LorenzMeier et al, thoughts?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4642,9 +4642,11 @@
       <field type="int32_t" name="value">Signed integer value</field>
     </message>
     <message id="253" name="STATUSTEXT">
-      <description>Status text message. These messages are printed in yellow in the COMM console of QGroundControl. WARNING: They consume quite some bandwidth, so use only for important status and error messages. If implemented wisely, these messages are buffered on the MCU and sent only at a limited rate (e.g. 10 Hz).</description>
+      <description>Status text message. These messages should only be used for important status and error messages. Keep them concise to ensure they can be displayed by the recipient (and to reduce bandwidth). WARNING: Status updates consume significant bandwidth. They should be buffered and sent at a limited rate (e.g. 10 Hz).</description>
       <field type="uint8_t" name="severity" enum="MAV_SEVERITY">Severity of status. Relies on the definitions within RFC-5424.</field>
-      <field type="char[50]" name="text">Status text message, without null termination character</field>
+      <field type="char[50]" name="text">Status text message, without null termination character.</field>
+      <extensions/>
+      <field type="char[204]" name="extra_text">Additional status text, without null termination character. Append to 'text' field content for display. NOTE: Not all systems support extra_text, so consider populating 'text' with sensible stand-alone content.</field>
     </message>
     <message id="254" name="DEBUG">
       <description>Send a debug value. The index is used to discriminate between values. These values show up in the plot of QGroundControl as DEBUG N.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4642,7 +4642,7 @@
       <field type="int32_t" name="value">Signed integer value</field>
     </message>
     <message id="253" name="STATUSTEXT">
-      <description>Status text message. These messages should only be used for important status and error messages. Keep them concise to ensure they can be displayed by the recipient (and to reduce bandwidth). WARNING: Status updates consume significant bandwidth. They should be buffered and sent at a limited rate (e.g. 10 Hz).</description>
+      <description>Status text message. These messages should only be used for important status and error messages (and kept concise).</description>
       <field type="uint8_t" name="severity" enum="MAV_SEVERITY">Severity of status. Relies on the definitions within RFC-5424.</field>
       <field type="char[50]" name="text">Status text message, without null termination character.</field>
       <extensions/>


WickedShell:
> It recommends users should craft the text content to be stand alone so that if used with systems that don't understand extra_text the message can still be understood.

I'm not sure this is practical. A common place to overflow at the moment is a format string who's length is not clear to begin with, and there isn't a good way to break it up. I suspect if we tried to split messages we'd end up having to always choose a short message that fits regardless, and then a fragment to glob on that may not make sense, or a reduced content version in the short one that is worse then the current situation with a string in the extension that makes sense. I agree that if you can stay below that's great and we should do it.

The other consideration is that these aren't null terminated strings, so a C based application would have to do a `strnlen()` on the first part of the string to determine if it could use it as is (a 50 character `text` section would flow directly into a `extra_text` without any `null` breaks), or if it needs to glob on an  extension in a new area, which is made worse as you now have to create another storage location. This is of course doable, but I'm not sure it's really worth the hassle/complexity to force on the people receiving a `statustext`).

message_definitions/v1.0/common.xml
```diff
@@ -4642,9 +4642,11 @@
       <field type="int32_t" name="value">Signed integer value</field>
     </message>
     <message id="253" name="STATUSTEXT">
-      <description>Status text message. These messages are printed in yellow in the COMM console of QGroundControl. WARNING: They consume quite some bandwidth, so use only for important status and error messages. If implemented wisely, these messages are buffered on the MCU and sent only at a limited rate (e.g. 10 Hz).</description>
+      <description>Status text message. These messages should only be used for important status and error messages. Keep them concise to ensure they can be displayed by the recipient (and to reduce bandwidth). WARNING: Status updates consume significant bandwidth. They should be buffered and sent at a limited rate (e.g. 10 Hz).</description>
```
WickedShell: I'm not sure that we should really be sticking warnings about MCU's buffering/bandwidth usage of a statustext... It's very application dependent on if that happens/is a real factor. And it's really a MAVLink users job to know if they are using up their bandwidth or not as that will always a concern. The warning here makes it look like `STATUSTEXT` does something special at a protocol level that makes it bigger then it otherwise would be. (Otherwise this implies we should be annotating all the messages that are >= 51 bytes of payload).

DonLakeFlyer:
QGC doesn't care what the text lenght is. 

I always thought the reason STATUS_TEXT was small was for two reasons:
* Make it fit into a single transmission packet over a radio or something like that. Just a guess.
* Since mavlink 1 messages are fixed size limit the bandwidth used for every STATUS_TEXT message

I'm not sure it's a goot idea to modify the existing STATUS_TEXT to make it large because you are affecting the bandwidth usage of every existing system. I would opt for a new mavlink 2 messages, which supports variable length messages doesn't it?

WickedShell:
> Make it fit into a single transmission packet over a radio or something like that. Just a guess.

We actually emit a warning in the compilier if a message is >64 bytes as it might not fit into a radio packet, but this really depends on the radio that is used, and doesn't seem to be a design concern on new messages.

> Since mavlink 1 messages are fixed size limit the bandwidth used for every STATUS_TEXT message

Because you can't use extensions with MAVLink1 the MAVLink1 transported `STATUSTEXT` would be unchanged from what it already is, it's only with MAVLink2 that the size could go up, and even there your average `STATUSTEXT` should actually be shorter then the MAVLink1, and if it is larger you should only be paying a cost on the non 0 bytes of the extension.

> I would opt for a new mavlink 2 messages, which supports variable length messages doesn't it?

All MAVLink2 encoding can do 0 stripping, extensions have to be last in  packet, and the empty characters in a string are all 0, so they would be stripped with the extension.

A new message has the downside that the application can't know if the other side can parse the new message, so it will generally have to send a short STATUSTEXT + the new message which ends up being longer then just sending the extension.

DonLakeFlyer:
Sorry I missed the extensions thing. This all seems fine to me then.

message_definitions/v1.0/common.xml
hamishwillee: I agree. Will fix.

hamishwillee:
@WickedShell Sorry, took me a long time to get back to this:

>> It recommends users should craft the text content to be stand alone so that if used with systems that don't understand extra_text the message can still be understood.

> I'm not sure this is practical. A common place to overflow at the moment is a format string who's length is not clear to begin with, and there isn't a good way to break it up. I suspect if we tried to split messages we'd end up having to always choose a short message that fits regardless, and then a fragment to glob on that may not make sense, or a reduced content version in the short one that is worse then the current situation with a string in the extension that makes sense. I agree that if you can stay below that's great and we should do it.

Good points. So what exactly are you proposing? That we just remove the comment to make this: *Additional status text, without null termination character. Append to 'text' field content for display.*

What would happen on a mavlink 1 system then that receives a partially truncated format string?

>The other consideration is that these aren't null terminated strings, so a C based application would have to do a strnlen() on the first part of the string to determine if it could use it as is (a 50 character text section would flow directly into a extra_text without any null breaks), or if it needs to glob on an extension in a new area, which is made worse as you now have to create another storage location. This is of course doable, but I'm not sure it's really worth the hassle/complexity to force on the people receiving a statustext).

So what happens here then? They just append the two strings, one of which might be empty, and everything works?

@DonLakeFlyer @WickedShell 

> A new message has the downside that the application can't know if the other side can parse the new message, so it will generally have to send a short STATUSTEXT + the new message which ends up being longer then just sending the extension.

Generally isn't the assumption that the flight stack will just support one format, and the burden of supporting multiple types falls on the GCS?
If the main GCS authors are happy to support both messages for a while I'd actually prefer to have a new message (or in an ideal word just update STATUSTEXT and rev the version number).

DonLakeFlyer:
It's not an issue for QGC to support a new status text message (or both old and new).

hamishwillee:
I'm fairly convinced that longer term we're better off just adding a new full-width MAVLink 2 message to replace STATUSTEXT (as opposed to having a MAVLink 2 extension field for appended/extra text. )
This is much simpler for both the sender and the recipient, and worth a little migration pain given we can expect it to be in use for a long time. 

I assume that 
- autopilots would choose to upgrade and support just new message (eventually)
- GCS and MAVLink APIs (drone

New message mirrors the old one but uses full width of available characters.
```xml
<message id="359" name="STATUSTEXT2">
    <description>Status text message (v2). These should only be used for important updates and error messages (and kept concise).</description>
    <field type="uint8_t" name="severity" enum="MAV_SEVERITY">Severity of status. Relies on the definitions within RFC-5424.</field>
    <field type="char[254]" name="text">Status text message, without null termination character</field>
</message>
```
**Do we need to think about Unicode/Internationalization?**

@peterbarker , @julianoes @meee1 Given that SDK, DroneKit, Mission Planner might need to be updated to understand this message, do you have any thoughts?

@LorenzMeier @dagar @auturgy Given that PX4 / ArduPilot should also update at some point to emit the new messages, what do you think?

LorenzMeier:
I've waited for the discussion to shake out a bit since I didn't want my quite opinionated position to bias it: For any software architect looking at MAVLink and the manually-coded way the old text field would need to overflow into the new text field must be horrible. If we prioritize hacking quickly something together for the codebases we care about today, we inevitably will more and more loose support as a quasi-standard.

peterbarker:
Just thinking out loud here....

```
uint8_t continuation_group_id;
uint8_t expect_continuation;
uint8_t is_continuation;
```

(actual types TBD e.g. bitmask)

`group_id` just tries to account for packet loss, being a (temporally) unique identifier for a group.

This might allow a GCS to delay emitting a message in the interface until all of a very long statustext is received, while allowing non-mavlink2-aware/non-new-mechanism-aware GCSs to work just fine.  This is handy for things like OSDs, which I've learnt the hard way are a PITA for backwards compatability...

hamishwillee:
Thanks @peterbarker . Just to be clear, I think you are considering an approach to dynamically construct arbitrary length STATUSTEXT messages in MAVLink 2 by sending them in pieces that get reconstructed at the other end.
If that is the case, then I think it is a clever idea, and would be good if we needed arbitrary length statustext (but I don't think we do).

The problem is that this looks even harder for the sender/recipient to construct/reconstruct than having a single additional field, and adds the complexity of waiting for message fragments and reconstructing them.

@LorenzMeier as per my comments previously I agree strongly with you on https://github.com/mavlink/mavlink/pull/1038#issuecomment-451147802.

But is there anything we can/should do to mitigate compatibility issues that Peter talks about above (e.g. taking the case of OSDs). If PX4 move to STATUSTEXT2 - does that mean that no OSD will work with PX4 for some time?

dagar:
> @LorenzMeier @dagar @auturgy Given that PX4 / ArduPilot should also update at some point to emit the new messages, what do you think?


My personal opinion is that the vast majority of STATUSTEXT usage would be better served by a structured message. Yes this is harder short term for a variety of reasons, but it's necessary if we want robust polished systems used internationally.

auturgy:
Completely agree that a new mavlink2 message is the better solution.
This PR may be a worthwhile transition path though.

auturgy:
@dagar I think you’re suggesting that we identify key use cases where statustext is basically abused for convenience (I think we’d all agree that it’s use exceeds it’s original intent), and come up with some new messages?
That’s also very sound, and ultimately worthwhile, but will need some thought.

LorenzMeier:
@auturgy The idea would be to build an events microservice that I meant to build out for a year now - its all in my head and I need to sit down and just do it.

Basically what it is is a linked list of events with a sequence, relative priorities and re-transmission of lost events (through buffering the last 2-3 which are in transit). It can be any type of event.

hamishwillee:
@dagar By "structured message" do you mean the same thing that @LorenzMeier is talking about? And if not, what did you have in mind? I know that we need something that:
- allows long(er) strings
- allows internationalisation

It sounds like @LorenzMeier adds to that a mechanism for reliable delivery (not clear to me what the rest of the concept is, and what "any type of event" might mean other than status that we have currently). 

@LorenzMeier ETA for getting idea out of your head into ours? Or do we discuss an intermediate message?

dagar:
> @dagar By "structured message" do you mean the same thing that @LorenzMeier is talking about? And if not, what did you have in mind?

What I was referring to is different, but it's something I'd like in addition to what Lorenz is talking about.

I'm talking about establishing well defined interfaces to replace some of things we've crammed into STATUSTEXT over the years. Things like preflight checks, reasons for failing to arm, general sensor status, etc. Yes these things vary quite a bit between flight controllers, but it still wouldn't be unreasonable to capture a large portion of it in the mavlink spec itself along with mechanisms for checking which parts are supported.

I don't think sending strings across these slow unreliable links should have become part of normal operation in the first place.

hamishwillee:
@dagar So basically you're envisaging separate messages for some of the stuff people are using STATUSTEXT for? Or possibly something that builds on the "events" architecture to allow the payload to be configured and structured to contain the messages we want. Sounds good to me, provided that the payload isn't opaque - ie can be inferred progammatically.

[As an aside, I'm a bit wary of this unless we can get versioning under control. There are unexpected consequences of encoding information in anything other than separate messages - ie things like the fact that you're forced to decide whether a param that is reserved can be used later for a message with 0 or NaN as the "not changed" value up front, because there is no way to tell that the definition has changed in the parser itself).]

But I'm not sure how we would approach that. Case by case? ie start with preflight checks and define preflight checks message? Feels like something that should spawn off this discussion.

LorenzMeier:
Conclusion on the call was to go with a new, separate message.

hamishwillee:
Awesome. I'll close this and a PR for new full length message now for discussion at #1058. 

@dagar @LorenzMeier This does not of course preclude an event framework or making more explicit structured messages to convey specific information, but I expect you'll drive that through some other forum than this PR.

Issue #929, opened 2018-06-01
No way to transmit airspeed reference


philipoe:
MAVLINK currently has no message to transmit den current airspeed reference. However, the airspeed reference is very important due to safety reasons (e.g. stall). The airspeed reference is also internally adapted by many autopilots, i.e. the airspeed reference that a user expects based on what he sets in the waypoint items (e.g. via QGC) is NOT the one that the vehicle is actually tracking, e.g. because of the "groundspeed undershoot" or "accelerated stall" functionality of PX4. Surprisingly, the [HIGH_LATENCY](https://mavlink.io/en/messages/common.html#HIGH_LATENCY) and [HIGH_LATENCY2](https://mavlink.io/en/messages/common.html#HIGH_LATENCY2) messages actually _do_ contain the airspeed reference.

So overall we think it is really important to add that. Question: Into what message should it be added? [NAV_CONTROLLER_OUTPUT](https://mavlink.io/en/messages/common.html#NAV_CONTROLLER_OUTPUT) would work for example. There is already an airspeed error in there, but in our opinion an airspeed ref makes more sense. The Only complication is that we'd then also need to add the true airspeed (TAS), because what is transmitted in the VFR_HUD message (at least by Pixhawk) is the indicated airspeed (IAS).

dagar:
I agree it would be much more valuable to have in place of aspd_error and alt_error.

Options
 1. Add new extension fields (mavlink v2 only) to NAV_CONTROLLER_OUTPUT (https://mavlink.io/en/messages/common.html#NAV_CONTROLLER_OUTPUT)
 2. Review the entire thing and create a new message.

I don't have a preference at this point.

Thoughts? @WickedShell @auturgy

WickedShell:
I'd prefer to extend rather then create a new message at the moment. Regarding `VFR_HUD` IAS is what you want sent to the GCS, although the field is not (and should be) annotated for what is being sent.

peterbarker:
We did discuss on a recent ArduPilot devcall an `airspeed` mavlink message.  Notably including an instance number and a type (which might be synthetic).  tridge suggested it could be an array in a single message instead ('though would you then need an array of types to go with it?).

hamishwillee:
I've added DevCall label (thought that is a month away) to make sure this does not stall again. 

I'm a bit confused about some of this. 

@philipoe @WickedShell :

[...]

PR #1041, opened 2018-12-17
Clarify VFR_HUD.airspeed is IAS

hamishwillee:
Discussion originates in https://github.com/mavlink/mavlink/issues/929#issuecomment-446777948.

The airspeed in the HUD could be true airspeed, indicated airspeed, airspeed setpoint, etc etc. This clarifies that it is "indicated airspeed". This is definitely the case for ArduPilot, and I THINK it is also the case for PX4 (@dagar ?)

In addition, minor tidy removes units information from descriptions.


dagar:
Yes it's indicated airspeed.

PR #1043, opened 2018-12-24
Common.xml:Partial mission read/write startindex=1

hamishwillee:
This changes the default start index for a partial read/write of a mission from 0 to 1.

My reasoning is that mission sequences are indexed from 1 to count, not from 0 to count. Having 0 here would fetch the zero home value on ArduPilot, but is otherwise confusing because it implies that sequences normally start at zero.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3368,7 +3368,7 @@
       <description>Request a partial list of mission items from the system/component. https://mavlink.io/en/services/mission.html. If start and end index are the same, just send one waypoint.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="int16_t" name="start_index">Start index, 0 by default</field>
+      <field type="int16_t" name="start_index">Start index, 1 by default</field>
       <field type="int16_t" name="end_index">End index, -1 by default (-1: send list to end). Else a valid index of the list</field>
       <extensions/>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@WickedShell @auturgy FYI/for discussion

auturgy:
I’m not sure that this change really adds any value, but ignoring technical merit, the description at ln3377 should be updated to reflect the intent of the pr.

WickedShell:
TL;DR: I think we should just remove all mention of "default" as you have to logically provide a value for these to make sense, and you still have to know what ArduPilot does with item 0 either way.
___

@hamishwillee I'm not sure this (or actually the entire sentence helps). You already have to know the specialness of index 0 since you get it on a full mission request, and when you go to upload you have to already know that you can't upload index 0 to ArduPilot. More importantly the sender of those messages is either requesting an index to read/write from. If it's out of scope surely the best thing to do is NACK it rather then auto warp it to something else?

WickedShell:
@hamishwillee Just checked, ArduPilot doesn't do any fixing of your index on an upload, if it's out of range we reject the transaction (although we don't issue a NACK which we should do), and we don't support the `REQUEST_PARTIAL_LIST` (which is actually a weird message since the mission protocol is built around the GCS just asking for items it cares about, so partial downloads can be done pretty much adhoc once you know the count, and even without the count you can discover it when you try and download.

hamishwillee:
@auturgy  @WickedShell
>the description at ln3377 should be updated to reflect the intent of the pr.

What is ln3377?

The value is definitely minor. But a value of zero is definitely "wrong" in terms of the spec, so having a default with that value is potentially confusing. 

Arguably removing a default does make sense for the lower range, particularly as we're not providing a hint for it. I'll remove default for that altogether. 

NACK makes sense. Thanks for that change.

Re, REQUEST_PARTIAL_LIST , the "right protocol" is awaiting your discussion in https://github.com/mavlink/mavlink-devguide/pull/157 .
I agree with what you are saying though - there is no need for MISSION_REQUEST_PARTIAL_LIST because you should be able to just request any item from any system once you have the count.

PR #1046, opened 2019-01-02
Allow defining stream type.

dogmaphobic:
This adds a field telling the GCS the type of video stream for a given stream. In most cases, you don't "set" anything. The vehicle has a fixed type of video stream and it will tell you what it is. The field allows the GCS to use a proper pipeline for receiving and rendering the video stream.

In addition, I've also added a flag telling if the stream supports zoom. As streams are decoupled from "cameras" (which have their own API), if your system has video stream with zoom support but no other "camera" feature (image/video capture, etc.), there is no way to let the GCS know the stream handles zoom.

@julianoes, another change in the structure. I'm implementing this in QGC (basic, experimental support is already in master).

Note that this is unrelated to #885 as this is only for `VIDEO_STREAM_INFORMATION` while #885 is dealing with `SET_VIDEO_STREAM_SETTINGS`, which I have no plans to implement within QGC.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2952,13 +2952,13 @@
         <description>Stream is RTSP</description>
       </entry>
       <entry value="1" name="VIDEO_STREAM_TYPE_RTPUDP">
-        <description>Stream is RTP UDP (uri gives the port number)</description>
+        <description>Stream is RTP UDP (URI gives the port number)</description>
       </entry>
       <entry value="2" name="VIDEO_STREAM_TYPE_TCP_MPEG">
         <description>Stream is MPEG on TCP</description>
       </entry>
-      <entry value="3" name="VIDEO_STREAM_TYPE_MPEGTS_H264">
-        <description>Stream is h.264 on MPEG TS (uri gives the port number)</description>
+      <entry value="3" name="VIDEO_STREAM_TYPE_MPEG_TS_H264">
+        <description>Stream is h.264 on MPEG TS (URI gives the port number)</description>
       </entry>
     </enum>
     <enum name="SET_ZOOM_TYPE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2942,6 +2942,24 @@
       <entry value="2" name="VIDEO_STREAM_STATUS_FLAGS_THERMAL">
         <description>Stream is thermal imaging</description>
       </entry>
+      <entry value="4" name="VIDEO_STREAM_HAS_BASIC_ZOOM">
+        <description>Stream has basic zoom control (MAV_CMD_SET_CAMERA_ZOOM)</description>
+      </entry>
+    </enum>
+    <enum name="VIDEO_STREAM_TYPE">
+      <description>Video stream types</description>
+      <entry value="0" name="VIDEO_STREAM_TYPE_RTSP">
+        <description>Stream is RTSP</description>
+      </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2942,6 +2942,24 @@
...
+        <description>Stream has basic zoom control (MAV_CMD_SET_CAMERA_ZOOM)</description>
+      </entry>
+    </enum>
+    <enum name="VIDEO_STREAM_TYPE">
+      <description>Video stream types</description>
+      <entry value="0" name="VIDEO_STREAM_TYPE_RTSP">
+        <description>Stream is RTSP</description>
+      </entry>
+      <entry value="1" name="VIDEO_STREAM_TYPE_RTPUDP">
+        <description>Stream is RTP UDP (uri gives the port number)</description>
```
hamishwillee: uri -> URI

message_definitions/v1.0/common.xml
```diff
@@ -2942,6 +2942,24 @@
...
+      </entry>
+      <entry value="2" name="VIDEO_STREAM_TYPE_TCP_MPEG">
+        <description>Stream is MPEG on TCP</description>
+      </entry>
+      <entry value="3" name="VIDEO_STREAM_TYPE_MPEGTS_H264">
```
hamishwillee: VIDEO_STREAM_TYPE_MPEG_TS_H264 - maybe extra underscore before TS. Up to you.

hamishwillee:
@auturgy @WickedShell FYI, this is a (sensible) break to the WIP message `VIDEO_STREAM_INFORMATION`. Anything to add?

auturgy:
Nothing of value. 
Does this suggest that the camera API doesn’t cover the necessary use cases, or is too complex?

dogmaphobic:
> Nothing of value.
> Does this suggest that the camera API doesn’t cover the necessary use cases, or is too complex?

Did you even read the description? If you have only a video stream and no "camera", there is no camera API. At best, it doesn't make any sense to implement an entire camera API just to tell the GCS it has zoom available.

The use case: Vehicle has a video stream. It does not capture images nor does it record video. It has no function other than streaming video while allowing you to zoom in and out. It does not fit the context of a camera API. The video stream is fixed (as it is usually the case). You don't "set" any setting. The GCS simply queries what it is, receives the information and uses it to build the pipeline and display the stream.

hamishwillee:
FWIW @dogmaphobic I always found it a bit weird that camera API covers video capture and still capture but not video streaming - so I thought the question reasonable even after reading the description.

Let's see if anyone else has a comment. I'm happy to push this in tomorrow if not (as it is `<wip>` so this is not the last opportunity for feedback).

dogmaphobic:
> I always found it a bit weird that camera API covers video capture and still capture but not video streaming - so I thought the question reasonable even after reading the description.

That's because there were two, distinct efforts done not in coordination. I was handling basic camera operations when this whole video streaming API sprouted, completely decoupled from the camera API. We now have two separate APIs. One handles a camera (capture/recording and its settings) and video streaming. Both independent of each other, which is weird. Short of scrapping this whole streaming API, I'm not sure else can be done. With that said however, it is not the end of the world to keep them separate and distinct if we must. It serves a purpose for cases such as this where you have video streaming and no "camera" functions. It's only harder to grasp when you have a camera that does both. It is certainly not making my life any easier implementing all of this within QGC.

LorenzMeier:
Given that contributors to this micro service have reviewed this I think it’s ok to merge and revisit if necessary.

PR #1048, opened 2019-01-04
Cleanup

dogmaphobic:
Removing WIP from messages that are no longer WIP (they are now in production)

Better explaining the _Capture sequence ID_ for `MAV_CMD_IMAGE_START_CAPTURE`. If _Number of images to capture total_ is not 1, the sequence ID seems meaningless. Correct me if I'm wrong.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1617,7 +1617,7 @@
         <param index="1">Reserved (Set to 0)</param>
         <param index="2">Duration between two consecutive pictures (in seconds)</param>
         <param index="3">Number of images to capture total - 0 for unlimited capture</param>
-        <param index="4">Capture sequence (ID to prevent double captures when a command is retransmitted, 0: unused, &gt;= 1: used)</param>
+        <param index="4">Capture sequence ID (to prevent double captures when a command is retransmitted, 0: unused, &gt;= 1: used - Only valid for single capture)</param>
         <param index="5">Reserved (all remaining params)</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1562,16 +1562,12 @@
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.</description>
         <param index="1">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
         <param index="2">0: No Action 1: Request storage information</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.</description>
         <param index="1">Storage ID (1 for first, 2 for second, etc.)</param>
         <param index="2">0: No action 1: Format storage</param>
@@ -4721,8 +4717,6 @@
       <field type="float" name="focusLevel">Current focus level (0.0 to 100.0, NaN if not known)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Information about a storage medium.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="storage_id">Storage ID (1 for first, 2 for second, etc.)</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1615,7 +1615,7 @@
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
         <description>Start image capture sequence. Camera must send CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
-        <param index="2" label="Duration" units="s">Duration between two consecutive pictures (in seconds)</param>
+        <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Capture Count" minValue="0">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE</param>
         <param index="4" label="Sequence Number" minValue="0">Capture sequence number starting from 1. This is only valid for single-capture (param3==1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it.</param>
         <param index="5">Reserved (all remaining params)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1563,14 +1563,14 @@
       </entry>
       <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
         <description>Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.</description>
-        <param index="1">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
-        <param index="2">0: No Action 1: Request storage information</param>
+        <param index="1" label="Storage ID" minValue="0">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
+        <param index="2" label="Confirm Request" minValue="0" maxValue="1">0: No Action 1: Request storage information</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT">
         <description>Format a storage medium. The formatting system must respond with STORAGE_INFORMATION when the operation is complete. Use the command's target_component to target a specific component's storage.</description>
-        <param index="1">Storage ID (1 for first, 2 for second, etc.)</param>
-        <param index="2">0: No action 1: Format storage</param>
+        <param index="1" label="Storage ID" minValue="0">Storage ID (1 for first, 2 for second, etc.)</param>
+        <param index="2" label="Confirm Request" minValue="0" maxValue="1">0: No action 1: Format storage</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="527" name="MAV_CMD_REQUEST_CAMERA_CAPTURE_STATUS">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1621,7 +1617,7 @@
         <param index="1">Reserved (Set to 0)</param>
```
hamishwillee: @dogmaphobic For description we have:
> Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.

Perhaps be specific about what sends CAMERA_IMAGE_CAPTURED. Also is it NAN or NaN generally? Anyway, up to you, but how about:

> Start image capture sequence. Use NaN for reserved values. Camera must send CAMERA_IMAGE_CAPTURED after each capture.

message_definitions/v1.0/common.xml
```diff
@@ -1621,7 +1617,7 @@
...
         <param index="2">Duration between two consecutive pictures (in seconds)</param>
         <param index="3">Number of images to capture total - 0 for unlimited capture</param>
```
hamishwillee: @dogmaphobic It might be good to cross reference MAV_CMD_IMAGE_STOP_CAPTURE here. Ie

> `<param index="3">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE</param>`

message_definitions/v1.0/common.xml
```diff
@@ -1621,7 +1617,7 @@
...
-        <param index="4">Capture sequence (ID to prevent double captures when a command is retransmitted, 0: unused, &gt;= 1: used)</param>
+        <param index="4">Capture sequence ID (to prevent double captures when a command is retransmitted, 0: unused, &gt;= 1: used - Only valid for single capture)</param>
```
hamishwillee: Are the bit `0: unused, &gt;= 1: used` useful? I.e. you wouldn't change code either way and you can't tell if the value is zero whether the sequence is invalid or just not being used.

How about?
> `<param index="4">Capture sequence number starting from 0. This is only valid for single-capture (param3==1) and is used to prevent double captures when a command is re-transmitted.</param>`

Params can now take values: https://mavlink.io/en/guide/xml_schema.html#param
You might want to apply min, increment and label. 

More generally, you could apply this to all the messages.

message_definitions/v1.0/common.xml
```diff
@@ -1621,7 +1617,7 @@
         <param index="1">Reserved (Set to 0)</param>
         <param index="2">Duration between two consecutive pictures (in seconds)</param>
```
hamishwillee: https://mavlink.io/en/guide/xml_schema.html#param

Units, minvalue?

message_definitions/v1.0/common.xml
```diff
@@ -1562,16 +1562,12 @@
...
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.</description>
         <param index="1">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
         <param index="2">0: No Action 1: Request storage information</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.</description>
```
hamishwillee: It is good to specify who sends what. So even though we can guess the sender here "Once format is complete, a STORAGE_INFORMATION message is sent.", perhaps

> The formatting system should respond with STORAGE_INFORMATION when the operation is complete.

hamishwillee:
@auturgy @LorenzMeier @amilcarlucas This is changing from WIP to "in production" so last chance for us to make changes without a lot more effort. The messages still seem reasonable to me - though I've suggested adding parameter metadata and some possible improvements to descriptions.

message_definitions/v1.0/common.xml
dogmaphobic: I've changed it to:

_Capture sequence number starting from 1. This is only valid for single-capture (param3==1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it._

This can be tricky though. Image capture are sent both programmatically (during a mission) and manually (when the user hits the shutter button). If somehow the command is sent from both (the user hitting the shutter button during a mission for example), the sequence number would be out of sync. That's why the "manual" mode sets the sequence to 0 (ignore it).

message_definitions/v1.0/common.xml
dogmaphobic: I've added `units="s"` but I'm not what would be an useful minvalue. 0.001s?

message_definitions/v1.0/common.xml
dogmaphobic: Changed to:

_Format a storage medium. The formatting system must respond with STORAGE_INFORMATION when the operation is complete. Use the command's target_component to target a specific component's storage._

message_definitions/v1.0/common.xml
```diff
@@ -1595,44 +1591,44 @@
...
+        <description>Set camera focus. Returns CAMERA_SETTINGS message. Use NaN for reserved values.</description>
         <param index="1" enum="SET_FOCUS_TYPE">Focus type</param>
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
-        <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.</description>
+        <description>Start image capture sequence. Camera must send CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
-        <param index="2">Duration between two consecutive pictures (in seconds)</param>
```
amilcarlucas: you might as well add name attributes,

hamishwillee: No, the attribute for params is `label` - a hint to the GUI how it can mark up the parameter name: https://mavlink.io/en/guide/xml_schema.html#param

The example predates the schema change. I will update it. No firm rule exists in the spec stating whether you have to use the markup or not, so in that sense it is optional. However most of that originates from Don and is clearly useful. So if you can add it I would very much appreciate it. 

I would do: label, units.
If the param is an int there isn't much point setting increment since default is 1. If you don't allow negative time than I'd set minValue to 0

message_definitions/v1.0/common.xml
hamishwillee: We haven't really got any precedent for these cases where a realistic minimum is "a little bit above zero".

My take on this is that this is a hint for the UI to set up a UI component for setting the value. So while we would like it to be realistic, I think using 0 in this case would be fine. Depends on what the system will do - ie we could specify that 0 is "minimum duration supported by hardware"?

Alternatively, we could add a comment in the description like:

> The minimum value depends on the hardware but will be slightly greater than 0.

Thoughts?

message_definitions/v1.0/common.xml
dogmaphobic: Got it. Check https://github.com/mavlink/mavlink/pull/1048/commits/0852343f591266621a0f81dd7fe12ccadd42f96b and see if it makes sense. It almost feels like there should also be a type (like _Boolean_ for things like `minValue="0" maxValue="1"`). Also, if this is to be used to build an UI,  there should be a way to prevent some parameters from being presented within the UI (such as those _Reserved_ values).

dogmaphobic:
Looks like CI is croaking on its own:
```
PYMAVLINK INSTALL
##############################################
Collecting lxml>=3.6.0 (from -r pymavlink/requirements.txt (line 1))
  Downloading https://files.pythonhosted.org/packages/16/4a/b085a04d6dad79aa5c00c65c9b2bbcb2c6c22e5ac341e7968e0ad2c57e2f/lxml-4.3.0.tar.gz (2.5MB)
    Complete output from command python setup.py egg_info:
    This lxml version requires Python 2.7, 3.4 or later.
```

message_definitions/v1.0/common.xml
hamishwillee: I'm still a bit confused on this one. You could be saying that every time you call MAV_CMD_IMAGE_START_CAPTURE for a single capture you iterate the value by one - so after 10 captures the value is 10? And if there is no ack back on a particular message you would do what?

Or you could be saying that every MAV_CMD_IMAGE_START_CAPTURE sets the value to one. If there is no `MAV_CMD_ACK` back you would then resend the message with a value of 2 (and so on).

dogmaphobic: You start at 1 and for each new IMAGE_START_CAPTURE command you would increment it by one. Retransmission is done automatically at a lower level, sending the same message again up to 3 times. The receiving end, in case it gets more than one of these retransmissions would notice the sequence is the same and ignore it.

Technically, the receiving end is only concerned about a same sequence. You could increment by one and wrap it around every 10 (starting at 1 again) and it would work in the same way.

hamishwillee: OK, that is clear to me now. The only part of this that is not very clear to a reader is:  *Use 0 to ignore it.*
A bit wordy, but how about:

> Single-capture sequence number starting from 1. Increment the number every time this command is sent for a single-capture command (param3==1), in order to prevent double captures if a command is re-transmitted. Set to 0 for multiple and manual capture commands (0 is ignored by recipient sequence checks).

message_definitions/v1.0/common.xml
dogmaphobic: For what's worth, nowhere within QGC are these commands presented in their raw form. QGC doesn't at all process these XML files in any way. These are all handled manually (added in code). I can't imagine them ever being used for building an UI automatically within the context of QGC.

message_definitions/v1.0/common.xml
hamishwillee: OK, thinking about this a bit more, for a UI minvalue, maxvalue and interval serve 2 purposes:
-  prevent obvious invalid values. minValue=0 allows invalid values for specific hardware but we can't stop that. However it does stop entry of negative values, which is a good thing. 
-Creating UI controls that show a sensible range. For this you probably always need both minvalue and maxvalue, so I don't think that is relevant.

Anyway, here is a proposal for you based on what we know:
```
```

dogmaphobic: Done in https://github.com/mavlink/mavlink/pull/1048/commits/fb9f2285472ffa6f8d6d8b53fbeacb111f82e6a7

hamishwillee:
Thanks @dogmaphobic , I'm good with this. We can't merge though until #1051 is resolved (CI). Anyone else got any comments?

Note also that at some point we might delete the reserved params or update them. I'm still trying to work out what the rules are for them in https://github.com/mavlink/mavlink/issues/1050#issuecomment-452173152 - but my desire is to remove them altogether so that "not defined==reserved" (and the code might automatically generate documentation to that effect).

message_definitions/v1.0/common.xml
DonLakeFlyer: The way QGC handles this is that I put in some small number for the minimum which I've pulled out of thin air. Also from a current QGC standpoint if 0 is a valid value then you can't have a min which is higher than that.

message_definitions/v1.0/common.xml
```diff
@@ -1563,14 +1563,14 @@
...
+        <param index="1" label="Storage ID" minValue="0">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
+        <param index="2" label="Confirm Request" minValue="0" maxValue="1">0: No Action 1: Request storage information</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT">
         <description>Format a storage medium. The formatting system must respond with STORAGE_INFORMATION when the operation is complete. Use the command's target_component to target a specific component's storage.</description>
-        <param index="1">Storage ID (1 for first, 2 for second, etc.)</param>
-        <param index="2">0: No action 1: Format storage</param>
+        <param index="1" label="Storage ID" minValue="0">Storage ID (1 for first, 2 for second, etc.)</param>
+        <param index="2" label="Confirm Request" minValue="0" maxValue="1">0: No action 1: Format storage</param>
```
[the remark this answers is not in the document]
hamishwillee: @dogmaphobic If this can happen, either in this or #1060 it would be good.

message_definitions/v1.0/common.xml
```diff
@@ -1563,14 +1563,14 @@
       </entry>
       <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
         <description>Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.</description>
-        <param index="1">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
-        <param index="2">0: No Action 1: Request storage information</param>
+        <param index="1" label="Storage ID" minValue="0">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
+        <param index="2" label="Confirm Request" minValue="0" maxValue="1">0: No Action 1: Request storage information</param>
```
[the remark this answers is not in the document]
hamishwillee: @dogmaphobic If this can happen, either in this or #1060 it would be good.

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
dogmaphobic: It's too late for this one. This is in use already and the code on the camera side specifically tests for param1 == 1 before acting on it.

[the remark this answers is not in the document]
dogmaphobic: Only for new messages and/or messages marked as WIP. These storage messages are already in use and it's too late to change them like this. That's the main reason I'm removing the "WIP" flag for these.

dogmaphobic:
@hamishwillee This PR is too stale as is. I will close it and reintroduce it with just the WIP removal for the storage messages.

PR #1049, opened 2019-01-06
Use of uninitialised memory

conition:
You're passing uninitialised memory to recvfrom() this is very bad, as this can cause the packages to be cut off.
From [man recvfrom](https://linux.die.net/man/2/recvfrom):

> The argument addrlen is a value-result argument, which the caller should initialize before the call to the size of the buffer associated with src_addr, and modified on return to indicate the actual size of the source address. The returned address is truncated if the buffer provided is too small; in this case, addrlen will return a value greater than was supplied to the call.


hamishwillee:
Thanks @conition. This makes sense. Merged.

PR #1053, opened 2019-01-07
Remove SET_VIDEO_STREAM_SETTINGS from Common.xml

hamishwillee:
Proposal is to delete the wip message `SET_VIDEO_STREAM_SETTINGS`. 

Discussion is in https://github.com/mavlink/mavlink/pull/885#issuecomment-451292906 but essentially goes that video streams (in their basic forms) are a passive system. You query what there is and you react to it accordingly. 

The set command does not make sense because there is no way to know *what* to set. It isn't really a common interface, and there are better ways this might be done. 

The message has a number of other problems, but since it isn't used by anyone and is still wip, suggestion is to remove it. 

IF this is accepted we can close https://github.com/mavlink/mavlink/pull/885

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4815,20 +4815,7 @@
       <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
       <field type="char[160]" name="uri">Video stream URI (TCP or RTSP URI ground station should connect to) or port number (UDP port ground station should listen to)</field>
     </message>
-    <message id="270" name="SET_VIDEO_STREAM_SETTINGS">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Message that sets video stream settings</description>
-      <field type="uint8_t" name="target_system">system ID of the target</field>
-      <field type="uint8_t" name="target_component">component ID of the target</field>
-      <field type="uint8_t" name="camera_id">Stream ID (1 for first, 2 for second, etc.)</field>
-      <field type="float" name="framerate" units="Hz">Frame rate (set to -1 for highest framerate possible)</field>
-      <field type="uint16_t" name="resolution_h" units="pix">Horizontal resolution (set to -1 for highest resolution possible)</field>
-      <field type="uint16_t" name="resolution_v" units="pix">Vertical resolution (set to -1 for highest resolution possible)</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @shakthi-prashanth-m 

@julianoes Gus thinks this message is pointless. Can you please analyse its use in the context of the SDK and let me know if you:
* agree - in which case what is the best way to co-ordinate removing this that won't break the SDK?
* disagree - in which case we discuss until we reach a resolution, which might be just to fix the bugs in the message as per #885

julianoes:
All I can say is that this was implemented for the SDK but I have never seen it in action. @shakthi-prashanth-m does it work with the Intel Aero?

hamishwillee:
> All I can say is that this was implemented for the SDK but I have never seen it in action. @shakthi-prashanth-m does it work with the Intel Aero?

@shakthi-prashanth-m It isn't just whether it works - ie you can make the API send data. It is whether it would make sense for anyone to actually do it this way as per Gus comments.

shakthi-prashanth-m:
> All I can say is that this was implemented for the SDK but I have never seen it in action. @shakthi-prashanth-m does it work with the Intel Aero?

Sorry, I couldn't test it on Intel Aero, but only verified using PX4-SITL and Camera Manager.

hamishwillee:
> With this command, we pass Camera Manager resolution settings, video stream URI, etc. Isn't it ?

@shakthi-prashanth-m (& @lbegani) Have a read of @dogmaphobic comments. Essentially (I believe) he is saying that most of these things are and should be set by the camera system and not from an external API or ground station.
The reason is that it makes no sense. You wouldn't tell the camera any of this because you know nothing about what it supports/enables or the limitations on what URIs it could publish etc.  Instead you the camera does what it wants, and you query for the available information/streams.

If a camera does require some form of configuration then Gus took the view that you'd have this in the camera manager configuration file, and this would be the sort of thing you might get/set via the extended parameter protocol in the same way as other camera parameters. 

So yes you can set these things, but do you see it as a realistic use case that you would want to. Is there anything that you know that calls the SDK and uses this.

The more I think about this the more I tend to take Gus point of view. There is perhaps an argument that for some cases you might want your getter filtering to be more flexible - ie rather than iterating through available streams to find one that does what you want, it might be nice to be able to query - stream that has lowest resolution, ideally in this orientation (but that is another discussion).

dogmaphobic:
Moved to #1060

Issue #935, opened 2018-06-21
Remove unused parameters from MAV_CMD in definition XML


benbarkay:
Given that `CommandLong` and `CommandInt` already specify the payload structure, there is no need to specify `<param index="N">Empty</param>`. A benefit from removing these is that programs that parse the XML files could tell which params are used and which params aren't.

I wouldn't mind doing the work once #934 is done.

hamishwillee:
This seems simpler to me than parsing params for the text "Empty". Have you tested that the various generators are smart enough to cope with non-defined params (ie don't require a param to be defined?).

@amilcarlucas Gut feeling - good idea or not? IF we decide this is a good idea how would we test it - run the main generators from mavgen before and after and check that generated output doesn't change?

amilcarlucas:
Yes, that would be a good test, ... but I have a gut feeling that this will be a problem when param1 is empty and param2 is not.

hamishwillee:
Looking at generated Python libraries, you can see that each enum gets a dict of enum values, and each enum value has a dict of parameters.  In the original version we had indexed values for each parameter, and now we're missing all the empty values.
```
enums['MAV_CMD'][30] = EnumEntry('MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT', '''Continue on the current course and climb/descend to specified altitude.  When the altitude is reached continue to the next command (i.e., don't proceed to the next command until the desired altitude is reached.''')
enums['MAV_CMD'][30].param[1] = '''Climb or Descend (0 = Neutral, command completes when within 5m of this command's altitude, 1 = Climbing, command completes when at or above this command's altitude, 2 = Descending, command completes when at or below this command's altitude. '''
enums['MAV_CMD'][30].param[7] = '''Desired altitude in meters'''
```

Any code that iterates the param dict by index will fail with key errors. I do suspect that we will see code that does this.

hamishwillee:
For C. 
The docs get a bit screwed up because they assume all params populated. E.g. for MAV_CMD_CONDITION_GATE
*  | Altitude: 0: ignore altitude| Empty| Empty| Latitude| Longitude| Altitude|  */
* | Altitude: 0: ignore altitude| Latitude| Longitude| Altitude|  */

Other than that, on brief code inspection, it looks like everything should work, because as far as I can see there is no data structure that people might try access index. Could be wrong. I'd want to test a bit.

hamishwillee:
[...]

PR #1054, opened 2019-01-08
Clarify MAV_CMD_NAV_LAND

hamishwillee:
Remove ambiguity from MAV_CMD_NAV_LAND

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -806,14 +806,14 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="21" name="MAV_CMD_NAV_LAND">
-        <description>Land at location</description>
-        <param index="1">Abort Alt</param>
-        <param index="2">Precision land mode. (0 = normal landing, 1 = opportunistic precision landing, 2 = required precsion landing)</param>
-        <param index="3">Empty</param>
-        <param index="4">Desired yaw angle. NaN for unchanged.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude (ground level)</param>
+        <description>Land at location.</description>
+        <param index="1" label="Abort Alt" units="m">Minimum target altitude if landing is aborted (0 = undefined/use system default).</param>
+        <param index="2" label="Land Mode" enum="PRECISION_LAND_MODE">Precision land mode.</param>
+        <param index="3" label="Empty">Empty.</param>
+        <param index="4" label="Yaw Angle" units="deg">Desired yaw angle. NaN for unchanged.</param>
+        <param index="5">Latitude.</param>
+        <param index="6">Longitude.</param>
+        <param index="7" units="m">Landing altitude (ground level in current frame).</param>
       </entry>
       <entry value="22" name="MAV_CMD_NAV_TAKEOFF">
         <description>Takeoff from ground / hand</description>
@@ -3149,6 +3149,18 @@
       <description>These flags encode the cellular network status</description>
       <entry value="1" name="CELLULAR_NETWORK_STATUS_FLAG_ROAMING">
         <description>Roaming is active</description>
+      </entry>
+    </enum>
+    <enum name="PRECISION_LAND_MODE">
+      <description>Precision land modes (used in MAV_CMD_NAV_LAND).</description>
+      <entry value="0" name="PRECISION_LAND_MODE_DISABLED">
+        <description>Normal (non-precision) landing.</description>
+      </entry>
+      <entry value="1" name="PRECISION_LAND_MODE_OPPORTUNISTIC">
+        <description>Use precision landing if beacon detected when land command accepted, otherwise land normally.</description>
+      </entry>
+      <entry value="2" name="PRECISION_LAND_MODE_REQUIRED">
+        <description>Use precision landing, searching for beacon if not found when land command accepted (land normally if beacon cannot be found).</description>
       </entry>
     </enum>
   </enums>


message_definitions/v1.0/common.xml
```diff
@@ -806,14 +806,14 @@
...
-        <description>Land at location</description>
-        <param index="1">Abort Alt</param>
-        <param index="2">Precision land mode. (0 = normal landing, 1 = opportunistic precision landing, 2 = required precsion landing)</param>
-        <param index="3">Empty</param>
-        <param index="4">Desired yaw angle. NaN for unchanged.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude (ground level)</param>
+        <description>Land at location.</description>
+        <param index="1" label="Abort altitude" units="m">Abort altitude (0 = undefined/use system default).</param>
```
hamishwillee: Setting to 0 uses system default value on ArduPilot, and this param is ignored on PX4. Discussion here: https://github.com/mavlink/qgroundcontrol/issues/7098#issuecomment-452104308.

@DonLakeFlyer What is the abort altitude? ie what does it mean?

message_definitions/v1.0/common.xml
```diff
@@ -806,14 +806,14 @@
...
+        <param index="2" label="Land mode" >Precision land mode (0 = normal landing, 1 = opportunistic precision landing, 2 = required precision landing).</param>
```
hamishwillee: @DonLakeFlyer Is there any reason not to turn this into an enum?

message_definitions/v1.0/common.xml
```diff
@@ -806,14 +806,14 @@
...
+        <param index="3" label="Empty">Empty.</param>
+        <param index="4" label="Yaw angle" units="deg">Desired yaw angle. NaN for unchanged.</param>
+        <param index="5" label="Latitude">Latitude.</param>
```
hamishwillee: I added labels for the position params. Is this necessary? The GCS can always infer this from the fact that it is a MAV_CMD with hasLocation not set (and hence defaulting to true). 

note, I can't set units on the Lat/Lon because these might be sent in a COMMAND_INT.

message_definitions/v1.0/common.xml
```diff
@@ -806,14 +806,14 @@
...
+        <param index="6" label="Longitude">Longitude.</param>
+        <param index="7" label="Altitude" units="m">Altitude (ground level).</param>
```
hamishwillee: What is this? ie is this the expected altitude of ground in the current frame?  In that case, should it be `Landing altitude (ground level in current frame).`

message_definitions/v1.0/common.xml
hamishwillee: Also, for GCS, is there any preferred label capitalisation and length of string?- ie is it "Abort Altitude" or "Abort altitude" or "Abort alt" or Abort Alt." etc?

hamishwillee:
Are there defaults/undefineds for anything other than Abort altitude? Ie can this MAV_CMD be sent as a command? And if so, if I set lat, lon, alt to NaN will this correspond to "Land at current position"?

WickedShell:
> And if so, if I set lat, lon, alt to NaN will this correspond to "Land at current position"?

lat/lon are packed as int32 in `MISSION_ITEM_INT` so it can't be set to NaN.

message_definitions/v1.0/common.xml
```diff
@@ -806,14 +806,14 @@
...
-        <param index="2">Precision land mode. (0 = normal landing, 1 = opportunistic precision landing, 2 = required precsion landing)</param>
-        <param index="3">Empty</param>
-        <param index="4">Desired yaw angle. NaN for unchanged.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude (ground level)</param>
+        <description>Land at location.</description>
+        <param index="1" label="Abort altitude" units="m">Abort altitude (0 = undefined/use system default).</param>
+        <param index="2" label="Land mode" >Precision land mode (0 = normal landing, 1 = opportunistic precision landing, 2 = required precision landing).</param>
+        <param index="3" label="Empty">Empty.</param>
```
WickedShell: Can we just not list empty fields?

hamishwillee:
@WickedShell Good point. No way to indicate "land right here then" using this message. I guess you can always just set the appropriate land mode.

message_definitions/v1.0/common.xml
hamishwillee: What do you mean?

WickedShell: Why not just delete param index 3 out of this one entirely? It's then impliclty unlisted/a GCS does not have to special case that a label of "Empty" means do not show the field.

hamishwillee: Because it screws up generated C library documentation and in Pymavlink results in holes in the enum index that I believe can be ignored, but which I can't get confirmation for: https://github.com/mavlink/mavlink/issues/935

If we are allowed to remove these, then awesome! 
1. Is this implicit assumption documented anywhere.
2. How does Empty differ from reserved?

@DonLakeFlyer would appreciate your thoughts on this too. If you both know it is OK I'd rather delete the field.

message_definitions/v1.0/common.xml
DonLakeFlyer: No. It's an enum in QGC.

message_definitions/v1.0/common.xml
DonLakeFlyer: > is this the expected altitude of ground in the current frame?

Yes

DonLakeFlyer:
> Are there defaults/undefineds for anything other than Abort altitude? Ie can this MAV_CMD be sent as a command? And if so, if I set lat, lon, alt to NaN will this correspond to "Land at current position"?

The answer to that is all firmware specific

message_definitions/v1.0/common.xml
DonLakeFlyer: I have no idea what the implications of a "deleted" field mean to the system.

DonLakeFlyer: Also as far as I know Empty = Reserved

message_definitions/v1.0/common.xml
hamishwillee: The answer is no. I have removed the lat, long, alt labels as these should be inferred as true unless explicitly turned off (in the MAV_CMD level metadata)

message_definitions/v1.0/common.xml
hamishwillee: Thanks. I'll leave this in for now. If I delete them, will do it as a global change.

message_definitions/v1.0/common.xml
hamishwillee: Advice I got was every-word capitalisation, shortened format, no period indicating truncation. So "Abort Alt"

message_definitions/v1.0/common.xml
```diff
@@ -806,14 +806,14 @@
...
-        <description>Land at location</description>
-        <param index="1">Abort Alt</param>
-        <param index="2">Precision land mode. (0 = normal landing, 1 = opportunistic precision landing, 2 = required precsion landing)</param>
-        <param index="3">Empty</param>
-        <param index="4">Desired yaw angle. NaN for unchanged.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude (ground level)</param>
+        <description>Land at location.</description>
+        <param index="1" label="Abort Alt" units="m">Minimum target altitude if landing is aborted (0 = undefined/use system default).</param>
```
hamishwillee: Note the description. The only implementation I found of this was in ArduPilot "Altitude to climb to if landing is aborted". I've changed as indicated because that assumes that the vehicle is below the altitude, and doesn't really tell you what the vehicle should do if it is above. The new version is more generic - it basically allows vehicles to be able to keep current altitude if above the abort alt.

hamishwillee:
@DonLakeFlyer et al. This is updated now with the new enum and all other comments addressed (or explicitly chosen not to for now). You happy for it to merge?

message_definitions/v1.0/common.xml
```diff
@@ -3151,6 +3151,18 @@
...
+    <enum name="PRECISION_LAND_MODE">
+      <description>Precision land modes (used in MAV_CMD_NAV_LAND).</description>
+      <entry value="0" name="PRECISION_LAND_MODE_DISABLED">
+        <description>Normal (non-precision) landing.</description>
+      </entry>
+      <entry value="1" name="PRECISION_LAND_MODE_OPPORTUNISTIC">
+        <description>Use precision landing if beacon detected when land command accepted, otherwise land normally.</description>
+      </entry>
+      <entry value="2" name="PRECISION_LAND_MODE_REQUIRED">
+        <description>Use precision landing, searching for beacon if not found when land command accepted (land normally if beacon cannot be found).</description>
```
amilcarlucas: Is the description correct? I would expect the land command to be rejected is beacon is not found.

hamishwillee: Thanks very much! I checked with Randy and this is correct.

hamishwillee:
@DonLakeFlyer Does this meet your requirements/is this good to merge (once branch updated?)

DonLakeFlyer:
Looks fine to me

hamishwillee:
Thanks. Merged.

message_definitions/v1.0/common.xml
hamishwillee: @amilcarlucas Opportunistic mode lands immediately when it gets a land command, using a beacon if one is detected. Required will wait/search for a beacon if it doesn't detect one when the land command is received - but will eventually land anyway.

I think the docs now say that, but welcome any improvement suggestion.

PR #1058, opened 2019-01-10
Common.xml:Add STATUSTEXT2 message

hamishwillee:
Adds STATUSTEXT2 message that allows full width status text, as I believe was suggested in dev meeting: https://github.com/mavlink/mavlink/pull/1038#issuecomment-452848604.

Note: 
- This uses a MAVLink2 id - assume no need to try find a MAVLink 1 id?
- Is the name OK? Could be STATUSTEXT_LONG or STATUS_TEXT_2 etc.
- I have stated *Note: The message is intended as a less restrictive replacement for STATUSTEXT.* My assumption being that systems will update to the new message. Systems that need to communicate specifically with MAVLink 1 devices would use STATUSTEXT.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5012,6 +5012,11 @@
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
+    <message id="365" name="STATUSTEXT2">
+      <description>Status text message (use only for important status and error messages). While the full message payload may be used for status text, we recommend that updates be kept concise. Note: The message is intended as a less restrictive replacement for STATUSTEXT.</description>
+      <field type="uint8_t" name="severity" enum="MAV_SEVERITY">Severity of status. Relies on the definitions within RFC-5424.</field>
+      <field type="char[254]" name="text">Status text message, without null termination character.</field>
+    </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
       <description>Cumulative distance traveled for each reported wheel.</description>


hamishwillee:
@DonLakeFlyer @auturgy @WickedShell @LorenzMeier @dagar @mee FYI. This is what I believe was discussed in the dev meeting - supersedes #1058

hamishwillee:
Absolutely agree. Changed to STATUSTEXT_LONG everywhere.

message_definitions/v1.0/common.xml
```diff
@@ -5012,6 +5012,11 @@
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
+    <message id="365" name="STATUSTEXT_LONG">
```
dagar: Why the ID jump from 360 to 365?

hamishwillee: The creating a message guidelines suggest that it is good to leave some space in the range to allow services to add additional messages. I figured 5 would be more than sufficient. Happy to move it if we don't think more messages might ever be needed for the orbit task.

dagar: Just curious what thought went into it. Seems fine.

amilcarlucas:
@hamishwillee Please use the "Rebase and Merge" option when merging, otherwise it is a pain to port this downstream

dagar:
> @hamishwillee Please use the "Rebase and Merge" option when merging, otherwise it is a pain to port this downstream

I'd recommend "Squash and merge" as an even better alternative in this case. It will squash the entire change into a single clean commit and link back to this discussion + full history.

hamishwillee:
@amilcarlucas I always do squash and merge. Given Dan's point, is this OK with you?

hamishwillee:
@LorenzMeier @auturgy @DonLakeFlyer et al. Any further comments/suggestions? Can this be merged?

auturgy:
Nothing from me.
@wickedshell FYI 

Regards,

James

amilcarlucas:
Yes, squash and merge is even better.

hamishwillee:
Thanks all!

Issue #1050, opened 2019-01-07
Additional param metadata


hamishwillee:
Don, we're starting to use the param metadata in some real cases. Gus made some points about particular cases that might not be as useful as we hope for display in UI. Would appreciate your thoughts:

1. Example below is actually a boolean. Do we need a way to represent that, and how would that be done best?
    `<param index="2" label="Confirm Request" minValue="0" maxValue="1">0: No action 1: Format storage</param>`
2. In fact if you're sending the message you would always specify `1`. I tend to think we need `default` too. Thoughts?
3. A lot of params are reserved. My thinking is that if we're going to do this in descriptions we should do it via markup. HOWEVER, I'd rather we made an explicit statement that any param that is not declared is automatically assumed to be reserved and we generate whatever code is needed to support that (deleting any things we have currently declared as reserved "manually").

dogmaphobic:
I'm not sure I understand these instances where you need a confirmation parameter (1 to confirm, 0 to ignore). Wouldn't _not_ sending the command be simpler? :)

DonLakeFlyer:
> nanUnchanged bool false True: value can be set to NaN to signal unchanged

If this is true the QGC will show an entry field with a checkbox. You can see an example of it with the Heading field of a waypoint:
![screen shot 2019-01-07 at 1 29 28 pm](https://user-images.githubusercontent.com/5876851/50794567-51437500-1280-11e9-8d9a-a4a5c1f18e43.png)

This is because there is no reasonable way for an end user to enter a NaN into a numeric field. Nor should they have to understand that NaN means, leave the heading alone.

hamishwillee:
> I'm not sure I understand these instances where you need a confirmation parameter (1 to confirm, 0 to ignore). Wouldn't not sending the command be simpler? :)

Yes. It is madness, but a lot of messages do this. Worse, then you need to add other options, and those don't make sense in the context of the original options.
I don't think bool in this case actually helps. I like the idea, but lets park until we find a case where we actually need it. 

In future, let's make sure that we don't have such annoying design.

hamishwillee:
@DonLakeFlyer But I still think params could use a default, and it looks like you support them, right?

hamishwillee:
With respect to "eserved", I think your point in https://github.com/mavlink/mavlink/issues/1050#issuecomment-452086899 is that a reserved parameter needs to send a value that can be interpreted as "not used" when the param is implemented. 

[...]

PR #1060, opened 2019-01-10
Video streaming is now part of the camera API

dogmaphobic:
I've moved the video streaming commands and messages into the fold of the camera API. What that means is that now a "camera component" is responsible for video streaming handling. The existing video streaming commands and messages are somewhat the same but now they are targeted to a camera component ID.

If a camera supports video streaming, it tells you by settings `CAMERA_CAP_FLAGS_HAS_VIDEO_STREAM` in its capabilities flag. When the GCS detects this flag, it will send a `MAV_CMD_REQUEST_VIDEO_STREAM_INFORMATION` command to the camera. From then on, all video streaming discovery and setup is done automatically.

A vehicle may contain 0 to many cameras (the total will depend on component ID availability as each camera has its own component ID). Each camera may contain up to 255 video streams (the value 0 is used for _all cameras_ whenever a stream ID is used).

Video streaming is inherently passive. That is, the GCS queries the camera for the video stream configuration and it then builds an appropriate pipeline to receive it. If your camera allows changing video streaming settings, you must use the `PARAM_EXT_XXX` set of commands (and a camera definition file exposing the parameters). The GCS will then present these parameters (along with their options) to the user.

Needless to say, this all needs to be documented within the camera API documentation.

* Make video streaming as part of the camera API
* Removed the `SET_VIDEO_STREAM_SETTINGS` command
* Added video stream status (request and response)

Also changed all instances of _NAN_ to _NaN_ for consistency.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4836,16 +4836,17 @@
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Information about video stream</description>
       <field type="uint8_t" name="stream_id">Video Stream ID (1 for first, 2 for second, etc.)</field>
-      <field type="uint8_t" name="count">Number of streams available</field>
-      <field type="uint8_t" name="type" enum="VIDEO_STREAM_TYPE">Type of stream</field>
-      <field type="uint16_t" name="flags" enum="VIDEO_STREAM_STATUS_FLAGS">Bitmap of stream status flags</field>
-      <field type="float" name="framerate" units="Hz">Frame rate</field>
-      <field type="uint16_t" name="resolution_h" units="pix">Horizontal resolution</field>
-      <field type="uint16_t" name="resolution_v" units="pix">Vertical resolution</field>
-      <field type="uint32_t" name="bitrate" units="bits/s">Bit rate</field>
-      <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise</field>
-      <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
-      <field type="char[160]" name="uri">Video stream URI (TCP or RTSP URI ground station should connect to) or port number (UDP port ground station should listen to)</field>
+      <field type="uint8_t" name="count">Number of streams available.</field>
+      <field type="uint8_t" name="type" enum="VIDEO_STREAM_TYPE">Type of stream.</field>
+      <field type="uint16_t" name="flags" enum="VIDEO_STREAM_STATUS_FLAGS">Bitmap of stream status flags.</field>
+      <field type="float" name="framerate" units="Hz">Frame rate.</field>
+      <field type="uint16_t" name="resolution_h" units="pix">Horizontal resolution.</field>
+      <field type="uint16_t" name="resolution_v" units="pix">Vertical resolution.</field>
+      <field type="uint32_t" name="bitrate" units="bits/s">Bit rate.</field>
+      <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise.</field>
+      <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view.</field>
+      <field type="char[32]" name="name">Stream name.</field>
+      <field type="char[160]" name="uri">Video stream URI (TCP or RTSP URI ground station should connect to) or port number (UDP port ground station should listen to).</field>
     </message>
     <message id="270" name="VIDEO_STREAM_STATUS">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1619,9 +1619,9 @@
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
-        <param index="2">Duration between two consecutive pictures (in seconds)</param>
-        <param index="3">Number of images to capture total - 0 for unlimited capture</param>
-        <param index="4">Capture sequence (ID to prevent double captures when a command is retransmitted, 0: unused, &gt;= 1: used)</param>
+        <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
+        <param index="3" label="Capture Count" minValue="0">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE</param>
+        <param index="4" label="Sequence Number" minValue="0">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it.</param>
         <param index="5">Reserved (all remaining params)</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE">
@@ -1632,8 +1632,8 @@
       <entry value="2002" name="MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Re-request a CAMERA_IMAGE_CAPTURE packet. Use NAN for reserved values.</description>
-        <param index="1">Sequence number for missing CAMERA_IMAGE_CAPTURE packet</param>
+        <description>Re-request a CAMERA_IMAGE_CAPTURE message. Use NaN for reserved values.</description>
+        <param index="1">Sequence number for missing CAMERA_IMAGE_CAPTURE message</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="2003" name="MAV_CMD_DO_TRIGGER_CONTROL">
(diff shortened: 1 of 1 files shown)


dogmaphobic:
I should mention I already have a branch of QGC with all this implemented and working.

message_definitions/v1.0/common.xml
```diff
@@ -1595,23 +1595,23 @@
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
-        <description>Set camera running mode. Use NAN for reserved values.</description>
+        <description>Set camera running mode. Use NaN for reserved values. GCS will send a MAV_CMD_REQUEST_VIDEO_STREAM_STATUS command after a mode change if the camera supports video streaming.</description>
         <param index="1">Reserved (Set to 0)</param>
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
```
hamishwillee: Add enum metadata: `<param index="2" enum="CAMERA_MODE">Camera mode.</param>`

message_definitions/v1.0/common.xml
```diff
@@ -1595,23 +1595,23 @@
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
-        <description>Set camera running mode. Use NAN for reserved values.</description>
+        <description>Set camera running mode. Use NaN for reserved values. GCS will send a MAV_CMD_REQUEST_VIDEO_STREAM_STATUS command after a mode change if the camera supports video streaming.</description>
```
hamishwillee: "Use NaN for reserved values" conflicts with `<param index="1">Reserved (Set to 0)</param>` (set to 0)

If we add a `default` metadata we can set NaN in that, otherwise I believe that defaults are set to 0 automatically by GCS.

hamishwillee: Regarding `<param index="3">Reserved (all remaining params)</param>` and similar  ...
As discussed in https://github.com/mavlink/mavlink/issues/1050#issuecomment-452508021 I'm tempted to remove all explicit reserved items  - these will exist implicitly, with defaults of zero. We'd add the ones that we want to have as NaN, and it make sense to add at least one of those, because then you have the option to later turn the reserved into something that uses NaN as the "not changed" signal.

message_definitions/v1.0/common.xml
```diff
@@ -1595,23 +1595,23 @@
...
         <param index="1">Reserved (Set to 0)</param>
         <param index="2">Camera mode (see CAMERA_MODE enum)</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="531" name="MAV_CMD_SET_CAMERA_ZOOM">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Set camera zoom. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
+        <description>Set camera zoom. Returns CAMERA_SETTINGS message. Use NaN for reserved values.</description>
```
hamishwillee: Presumably there is still an ACK? Is the CAMERA_SETTINGS returned on both ACK and NACK?
Perhaps be explicit about what responds and under what conditions - ie *Returns CAMERA_SETTINGS message* TO *Camera must respond with a CAMERA_SETTINGS message (on success or failure).*

And as discussed above, "Use NaN for reserved values." is fine here, but there is no way for QGC to know, so it will probably use default of zeros. The only way to fix this (in future) will be to individually declare these as NaNs.

message_definitions/v1.0/common.xml
```diff
@@ -1595,23 +1595,23 @@
...
         <param index="1" enum="SET_ZOOM_TYPE">Zoom type</param>
```
hamishwillee: Re `<param index="1" enum="SET_ZOOM_TYPE">Zoom type</param>` should enum be `SET_ZOOM_TYPE`? I would expect `ZOOM_TYPE` or `CAMERA_ZOOM_TYPE`

message_definitions/v1.0/common.xml
```diff
@@ -1595,23 +1595,23 @@
...
         <param index="2">Zoom value</param>
```
hamishwillee: Perhaps `<param index="2">Zoom value</param>` to `<param index="2">Zoom value. The range of valid values depend on the zoom type.</param>`

message_definitions/v1.0/common.xml
```diff
@@ -1595,23 +1595,23 @@
...
         <param index="3">Reserved (all remaining params)</param>
       </entry>
       <entry value="532" name="MAV_CMD_SET_CAMERA_FOCUS">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Set camera focus. Returns CAMERA_SETTINGS message. Use NAN for reserved values.</description>
+        <description>Set camera focus. Returns CAMERA_SETTINGS message. Use NaN for reserved values.</description>
```
hamishwillee: As above, questions re 
- ACK and what returns CAMERA_SETTINGS 
- SET_FOCUS_TYPE or FOCUS_TYPE or CAMERA_FOCUS_TYPE

message_definitions/v1.0/common.xml
```diff
@@ -1626,23 +1626,23 @@
...
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE">
-        <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NAN for reserved values.</description>
+        <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
-        <param index="2">Duration between two consecutive pictures (in seconds)</param>
-        <param index="3">Number of images to capture total - 0 for unlimited capture</param>
-        <param index="4">Capture sequence (ID to prevent double captures when a command is retransmitted, 0: unused, &gt;= 1: used)</param>
+        <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
+        <param index="3" label="Capture Count" minValue="0">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE</param>
```
hamishwillee: probably set increment metadata to force as integer.
Fullstop at end?

message_definitions/v1.0/common.xml
```diff
@@ -1626,23 +1626,23 @@
...
+        <param index="4" label="Sequence Number" minValue="0">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it.</param>
         <param index="5">Reserved (all remaining params)</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE">
-        <description>Stop image capture sequence Use NAN for reserved values.</description>
+        <description>Stop image capture sequence Use NaN for reserved values.</description>
```
hamishwillee: Missing full stop after "Stop image capture sequence". Same question about NaNs. And I guess "why" NAN?

message_definitions/v1.0/common.xml
hamishwillee: Actually, on further thought, perhaps this is irrelevant - the values are never visible in the UI and you can make QGC do what it likes. I would still like to handle reserved value types as metadata for the general case though.

message_definitions/v1.0/common.xml
dogmaphobic: Done. I think enums didn't exist for `param` at some point.

message_definitions/v1.0/common.xml
dogmaphobic: Most of these things I've just copied and pasted from other command and/or messages. In case of commands, QGC specifically fills all unused params with NaN (those it uses). The case for param1 here is specific. There used to be something there with no function if set to zero. I can't quite remember what it was. It's no longer used but because it existed at some point (even though it was way before these things were in use), it guarantees it is not some value that would be interpreted as valid by some possible old code.

message_definitions/v1.0/common.xml
```diff
@@ -4841,19 +4847,18 @@
       <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
       <field type="char[160]" name="uri">Video stream URI (TCP or RTSP URI ground station should connect to) or port number (UDP port ground station should listen to)</field>
     </message>
-    <message id="270" name="SET_VIDEO_STREAM_SETTINGS">
+    <message id="270" name="VIDEO_STREAM_STATUS">
```
hamishwillee: This appears to have large overlap over `VIDEO_STREAM_INFORMATION`
- is there an intention to delete VIDEO_STREAM_INFORMATION
- It lacks the URI field. Where should stream URI come from?

message_definitions/v1.0/common.xml
dogmaphobic: All commands _must_ return an ACK before returning some possible message in response. Explaining this should be at a higher level and not within one single command.

When you send a `MAV_CMD_SET_CAMERA_ZOOM` command, the camera will first ACK it (because it is a MAV_CMD_XXX) and then respond with a `CAMERA_SETTINGS` message.

Regarding to the _Reserved_ stuff and NaNs, it's probably better to do a separate pass just for that instead of doing it in this PR. That way it's a global do over instead of a few individual commands and messages.

message_definitions/v1.0/common.xml
dogmaphobic: Makes sense. I'm not sure what I was thinking at the time.

message_definitions/v1.0/common.xml

message_definitions/v1.0/common.xml
```diff
@@ -4841,19 +4847,18 @@
...
-      <field type="uint32_t" name="bitrate" units="bits/s">Bit rate (set to -1 for auto)</field>
-      <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise (0-359 degrees)</field>
-      <field type="char[160]" name="uri">Video stream URI (mostly for UDP/RTP)</field>
+      <description>Information about the status of a video stream.</description>
+      <field type="uint8_t" name="stream_id">Video Stream ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint16_t" name="flags" enum="VIDEO_STREAM_STATUS_FLAGS">Bitmap of stream status flags</field>
+      <field type="float" name="framerate" units="Hz">Frame rate</field>
+      <field type="uint16_t" name="resolution_h" units="pix">Horizontal resolution</field>
+      <field type="uint16_t" name="resolution_v" units="pix">Vertical resolution</field>
+      <field type="uint32_t" name="bitrate" units="bits/s">Bit rate in bits per second</field>
```
hamishwillee: Remove units from description.

hamishwillee:
FWIW I think this is logically much better than what we had before, because it centralises all camera related activities through a consistent interface - rather than spread across APIs as previously. As discussed offline with Gus, one further point is that you must query streams individually by camera - you can't specify a component ID of "All" in a request for information and get all camera components responding.

The only thing is that this does break some messages so changes will need to be coordinated with systems that have implemented the API - presumably that means Camera Manager, QGC and Dronecode only?

message_definitions/v1.0/common.xml
hamishwillee: Correct. They only came into the schema in the last year.

message_definitions/v1.0/common.xml
hamishwillee: Are you sure about "QGC specifically fills all unused params with NaN (those it uses).". @DonLakeFlyer says *QGC sets reserved params to 0 unless otherwise specified by a default. I think MP is the same.* in https://github.com/mavlink/mavlink/issues/1050#issuecomment-452389033

The rest makes sense.

message_definitions/v1.0/common.xml
hamishwillee: Yes, re ACK, but is the CAMERA_SETTINGS  sent even if the ACK is a NACK? ie should the text be from *Returns CAMERA_SETTINGS message* TO *Camera must respond with a CAMERA_SETTINGS message (on success or failure)*.

Yes re reserved stuff. I plan to go through it globally when finally clear in my head.

message_definitions/v1.0/common.xml
dogmaphobic: A stream may change at run time. Specially when you switch a camera from Photo mode to Video mode (photos tend to be 4:3 while videos are usually 16:9 aspect ratio). You need a way to request and receive its current configuration so you can adapt to it accordingly. I chose to create a new message specific to "status" and exclude the fields that can _not_ change. Had I used the original `VIDEO_STREAM_INFORMATION`, people could decide to change things like URI or Type and lead to a mess (or a great deal of documentation that can be avoided by simply not giving the option).

message_definitions/v1.0/common.xml
dogmaphobic: Done. Note that here, none of the descriptions have full stop (this message was written a while back by a different group). Another case for a separate do over and deal with all full stops for consistency.

dogmaphobic:
> The only thing is that this does break some messages so changes will need to be coordinated with systems that have implemented the API - presumably that means Camera Manager, QGC and Dronecode only?

Only the video portion. The camera API itself is unchanged. Things were just added but nothing changed. The video commands and messages were added to the SDK but simply because they existed. I doubt it's in use.

message_definitions/v1.0/common.xml
dogmaphobic: At least for the camera commands: https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/CameraSection.cc#L127

dogmaphobic: General commands do set all to zero by default: https://github.com/mavlink/qgroundcontrol/blob/master/src/Vehicle/Vehicle.h#L983-L984

message_definitions/v1.0/common.xml
dogmaphobic: No, in all cases I know off, you only get a response (for those commands that return a message) if you get an ACK.

message_definitions/v1.0/common.xml
hamishwillee: Yes, agreed. Problem with big-picture changes is that they irritate the people who want to do git blame because it can be hard to see wood from trees. Worth it to me though.

message_definitions/v1.0/common.xml
hamishwillee: Thanks for clarification. Makes sense.

message_definitions/v1.0/common.xml
hamishwillee: Then propose change that bit to *Camera must respond with a CAMERA_SETTINGS message (on success).*

message_definitions/v1.0/common.xml
hamishwillee: Thanks for clarification. You still can't specify "Use NaN for reserved values" because someone else can send these as MISSION_INT or COMMAND_INT. The have to use zero. Probably means QGC should change what it sends for those params by default.

hamishwillee: The fact that you can't encode a NaN in an Int is already sorted out :-) But I agree that we do need to clarify and perhaps leaving for a bulk change makes sense.

That said, you could change to *Use NaN for reserved values where possible*

dogmaphobic:
I've added a `name` field to the `VIDEO_STREAM_INFORMATION` message (
https://github.com/mavlink/mavlink/pull/1060/commits/34d7eaaa5de14a0355137564935911a84c2bca0f). This will allow the GCS to build a list of video streams for the user to chose from.

LorenzMeier:
Good to go on a technical level - could you please fix the typo / style aspects Hamish has raised? Thanks!

dogmaphobic:
> could you please fix the typo / style aspects Hamish has raised? Thanks!

I think everything that was raised has been dealt with. At least within the context of this PR. There are some other issues that apply to the message definitions as a whole (not specific to video/camera), which should be handled on a separate PR.

hamishwillee:
@dogmaphobic Can you please also fix up name of `SET_FOCUS_TYPE` to `CAMERA_FOCUS_TYPE`

> could you please fix the typo / style aspects Hamish has raised? Thanks!

All the typos were fixed. One issue that was not resolved is that all descriptions say *Use NaN for reserved values.* but that is not possible if messages are sent using a COMMAND_INT. I suggested using *Use NaN for reserved values where possible*, but an alternative is to add that this message should ONLY be used with COMMAND_INT.

PR #1061, opened 2019-01-11
common.xml: MAV_CMD_NAV_PAYLOAD_PLACE param metadata

hamishwillee:
This may look unimportant, but I use it in an example, so I want to make it right

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -983,14 +983,14 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="94" name="MAV_CMD_NAV_PAYLOAD_PLACE">
-        <description>Descend and place payload.  Vehicle descends until it detects a hanging payload has reached the ground, the gripper is opened to release the payload</description>
-        <param index="1">Maximum distance to descend (meters)</param>
+        <description>Descend and place payload. Vehicle descends until it detects a hanging payload has reached the ground, the gripper is opened to release the payload</description>
+        <param index="1" label="Max Descent" minValue="0" units="m">Maximum distance to descend</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
         <param index="5">Latitude (deg * 1E7)</param>
         <param index="6">Longitude (deg * 1E7)</param>
-        <param index="7">Altitude (meters)</param>
+        <param index="7">Altitude</param>
       </entry>
       <entry value="95" name="MAV_CMD_NAV_LAST">
         <description>NOP - This command is only used to mark the upper limit of the NAV/ACTION commands in the enumeration</description>


message_definitions/v1.0/common.xml
```diff
@@ -983,14 +982,14 @@
...
-        <param index="1">Maximum distance to descend (meters)</param>
+        <description>Descend and place payload. Vehicle descends towards the specified location until it detects a hanging payload has reached the ground. The gripper is then opened to release the payload.</description>
+        <param index="1" label="Max Descent" minValue="0" units="m">Maximum distance to descend</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5">Latitude (deg * 1E7)</param>
-        <param index="6">Longitude (deg * 1E7)</param>
-        <param index="7">Altitude (meters)</param>
+        <param index="5">Latitude</param>
```
hamishwillee: FYI, I removed the units here because these are supplied by the MAV_FRAME or the type of message you are using. ie if sent in a COMMAND_LONG the `deg * 1E7` makes no sense.

message_definitions/v1.0/common.xml
```diff
@@ -983,14 +982,14 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="94" name="MAV_CMD_NAV_PAYLOAD_PLACE">
-        <description>Descend and place payload.  Vehicle descends until it detects a hanging payload has reached the ground, the gripper is opened to release the payload</description>
-        <param index="1">Maximum distance to descend (meters)</param>
+        <description>Descend and place payload. Vehicle descends towards the specified location until it detects a hanging payload has reached the ground. The gripper is then opened to release the payload.</description>
```
hamishwillee: The message is ambiguous. 
- It has a max descent param (1) and you might interpret that vehicle must descend where it is currently located and then drop payload if ground is detected.
- It has lat, long, alt fields. So it can be interpreted as "Go to this position, then descend and place payload, obeying the max descend value" .... but then why specify a max descend if you know the landing position?

Either way, what happens if it drops param 1 metres and hasn't reached ground? Does it just wait for further instructions?

message_definitions/v1.0/common.xml
```diff
@@ -983,14 +983,14 @@
...
+        <param index="1" label="Max Descent" minValue="0" units="m">Maximum distance to descend</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5">Latitude (deg * 1E7)</param>
-        <param index="6">Longitude (deg * 1E7)</param>
-        <param index="7">Altitude (meters)</param>
+        <param index="5">Latitude</param>
```
amilcarlucas: Add label

message_definitions/v1.0/common.xml
```diff
@@ -983,14 +983,14 @@
...
+        <param index="6">Longitude</param>
```
amilcarlucas: Add label

message_definitions/v1.0/common.xml
```diff
@@ -983,14 +983,14 @@
...
+        <param index="7">Altitude</param>
```
amilcarlucas: Add label

message_definitions/v1.0/common.xml
hamishwillee: I also didn't include labels, because these are implied.

message_definitions/v1.0/common.xml
hamishwillee: @amilcarlucas Thanks for this suggestion, however this is a deliberate omission. A MAV_CMD enum may have isDestination or hasLocation and these default to "true". So unless these are set false a GCS can infer that these have the labels lat, lon, alt. I don't think it makes sense to force things to be specified where the schema lets the UI determine them automatically.

message_definitions/v1.0/common.xml
hamishwillee: Answered by Lorenz. 
1. The location is the target you go to to ascend. 
2. If it doesn't detect landed it won't drop the payload and will proceed to next mission item

As an aside, Lorenz suggested that we might want to consider whether mission protocol could provide mission behaviour that is conditional on success of operation, so the following waypoint could do a try-again, or whatever.

message_definitions/v1.0/common.xml
hamishwillee: As above, deliberate omission https://github.com/mavlink/mavlink/pull/1061/#discussion_r247720230

message_definitions/v1.0/common.xml
hamishwillee: As above, deliberate omission https://github.com/mavlink/mavlink/pull/1061/#discussion_r247720230

hamishwillee:
Self merged. Updated in line with discussion with Lorenz and CI passed.

PR #1064, opened 2019-01-15
Document recording time within CAMERA_CAPTURE_STATUS message.

dogmaphobic:
If the camera does not keep track of recording time (or doesn't know it), set the `recording_time_ms` field to 0. The `CAMERA_CAPTURE_STATUS` is only sent periodically so the GCS keeps track of recording time on its own any way. But when the camera does support reporting of the recording time, the GCS will update its internal timer to reflect what the camera tells it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6051,7 +6051,7 @@
       <field type="uint8_t" name="image_status">Current status of image capturing (0: idle, 1: capture in progress, 2: interval set but idle, 3: interval set and capture in progress)</field>
       <field type="uint8_t" name="video_status">Current status of video capturing (0: idle, 1: capture in progress)</field>
       <field type="float" name="image_interval" units="s">Image capture interval</field>
-      <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started</field>
+      <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started. Set to 0 if not supported (GCS to compute recording time on its own.)</field>
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
       <extensions/>
       <field type="int32_t" name="image_count">Total number of images captured ('forever', or until reset using MAV_CMD_STORAGE_FORMAT).</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4766,7 +4766,7 @@
       <field type="uint8_t" name="image_status">Current status of image capturing (0: idle, 1: capture in progress, 2: interval set but idle, 3: interval set and capture in progress)</field>
       <field type="uint8_t" name="video_status">Current status of video capturing (0: idle, 1: capture in progress)</field>
       <field type="float" name="image_interval" units="s">Image capture interval</field>
-      <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started</field>
+      <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started. Set to 0 if not supported (GCS to compute recording time on its own.)</field>
```
hamishwillee: 1. If we are still able to break compatibility of this message, then field name should change to `recording_time`.
1. Why not max value of Int for "not supported"? After all, 0 would seem a valid and far more likely value for recording time.
1. Perhaps (?)
   > Elapsed recording time (0: Not supported/available). Recipients may also compute the time (e.g. to compensate for dropped messages) but should respect non-zero values from camera.

dogmaphobic: We're not "breaking" anything per se. It is extremely unlikely the camera would send this message at time = 0. Even if it does, it won't change anything. Think of it as your computer internal clock. It is set but the computer periodically sends a request to an NTP server for the "real" time. It then adjusts its internal clock. If the GCS gets a 0 in this field, it does nothing. It only uses its internal clock. If it gets a time (elapsed time since the recording started), it adjusts its local recording counter.

dogmaphobic: The whole idea is that, if the camera doesn't keep track of the recording time, make sure to keep this field set to 0. Anything other than 0 is considered elapsed time in ms.

dogmaphobic: * We can't change the message. This is already in use. 
* I think 0 is the most logical value to set recording_time_ms unless you have an actual elapsed time. Using something like 0xFFFFFFFF for "not used" only makes it confusing.
* The GCS is always computing the recording time on its own. It uses the elapsed time in this message (if any) to update any discrepancy.

hamishwillee: OK - your call. Last go hopefully.

> Elapsed time since recording started (0: Not supported/available). A GCS should compute recording time and use non-zero values of this field to correct any discrepancy.

hamishwillee:
@dogmaphobic I'm happy to merge this if you are. Are you OK with the current text/do you still want this?

PR #1065, opened 2019-01-18
MOTOR_TEST service updates to common.xml

hamishwillee:
This adds new param metadata to motor test command. I'm hoping to clarify the operation and document the service. 

Some assumptions - can anyone confirm:
- This is a simple command protocol message MAV_CMD_DO_MOTOR_TEST. There are no other messages involved other than the ACK/NACK.
- This cannot/must not be used in missions.
- ACK is returned with progress updates, and with success and on completion. Otherwise if test can't be run or if it fails NACK returned.
- There are two modes - single motor test and sequential motor test.
  - Single motor test uses param 1 and tests just the specified motor
  - Sequence uses param 5 to indicate a count of motors to test, with sequence specified in other params.
- I assume this is expected to work such that the user either tells GCS to start one motor or a sequence of motors. The message is sent and the user watches the motors start, run etc. The GCS shows motors under test based on ACK progress, and knows when finished based on ACK complete. The GCS may report an error if test can't run, but otherwise this is just user watching the motors spin and deciding if they are happy with the order. 

Questions:
 - What if both single (param 1) and sequence (param 5) tests are specified? Which should the command respect?
- In sequence, what is the timeout? Ie I guess that test runs, motor is started and this timeout is how long the test will run for. But it could be that some feedback is expected from "somewhere" and this is timeout that says "we're not waiting for confirmation".
- If timeout is really just "motor run time" then assume this also applies to single motor test mode?
- Where is this implemented? ArduPilot, PX4?

Further comments inline!

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1410,9 +1410,8 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-
       <entry value="209" name="MAV_CMD_DO_MOTOR_TEST" hasLocation="false" isDestination="false">
-        <description>Mission command to perform motor test.</description>
+        <description>Command to perform motor test.</description>
         <param index="1" label="Instance" minValue="1" increment="1">Motor instance number (from 1 to max number of motors on the vehicle).</param>
         <param index="2" label="Throttle Type" enum="MOTOR_TEST_THROTTLE_TYPE">Throttle type (whether the Throttle Value in param3 is a percentage, PWM value, etc.)</param>
         <param index="3" label="Throttle">Throttle value.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1410,13 +1410,14 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+
       <entry value="209" name="MAV_CMD_DO_MOTOR_TEST" hasLocation="false" isDestination="false">
         <description>Mission command to perform motor test.</description>
-        <param index="1" label="Instance" minValue="1" increment="1">Motor instance number. (from 1 to max number of motors on the vehicle)</param>
-        <param index="2" label="Throttle Type" enum="MOTOR_TEST_THROTTLE_TYPE">Throttle type.</param>
-        <param index="3" label="Throttle">Throttle.</param>
-        <param index="4" label="Timeout" units="s" minValue="0">Timeout.</param>
-        <param index="5" label="Motor Count" minValue="0" increment="1">Motor count. (number of motors to test to test in sequence, waiting for the timeout above between them; 0=1 motor, 1=1 motor, 2=2 motors...)</param>
+        <param index="1" label="Instance" minValue="1" increment="1">Motor instance number (from 1 to max number of motors on the vehicle).</param>
+        <param index="2" label="Throttle Type" enum="MOTOR_TEST_THROTTLE_TYPE">Throttle type (whether the Throttle Value in param3 is a percentage, PWM value, etc.)</param>
+        <param index="3" label="Throttle">Throttle value.</param>
+        <param index="4" label="Timeout" units="s" minValue="0">Timeout between tests that are run in sequence.</param>
+        <param index="5" label="Motor Count" minValue="0" increment="1">Motor count. Number of motors to test in sequence: 0/1=one motor, 2= two motors, etc. The Timeout (param4) is used between tests.</param>
         <param index="6" label="Test Order" enum="MOTOR_TEST_ORDER">Motor test order.</param>
         <param index="7">Empty</param>
       </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1346,13 +1346,13 @@
...
       </entry>
       <entry value="209" name="MAV_CMD_DO_MOTOR_TEST">
-        <description>Mission command to perform motor test</description>
-        <param index="1">motor number (a number from 1 to max number of motors on the vehicle)</param>
-        <param index="2">throttle type (0=throttle percentage, 1=PWM, 2=pilot throttle channel pass-through. See MOTOR_TEST_THROTTLE_TYPE enum)</param>
-        <param index="3">throttle</param>
-        <param index="4">timeout (in seconds)</param>
-        <param index="5">motor count (number of motors to test to test in sequence, waiting for the timeout above between them; 0=1 motor, 1=1 motor, 2=2 motors...)</param>
-        <param index="6">motor test order (See MOTOR_TEST_ORDER enum)</param>
+        <description>Mission command to perform motor test.</description>
```
hamishwillee: Not "mission" command right? ie this can't be used in missions.

message_definitions/v1.0/common.xml
```diff
@@ -2752,29 +2752,31 @@
...
-        <description>default autopilot motor test method</description>
+        <description>Default autopilot motor test method.</description>
       </entry>
       <entry name="MOTOR_TEST_ORDER_SEQUENCE" value="1">
-        <description>motor numbers are specified as their index in a predefined vehicle-specific sequence</description>
+        <description>Motor numbers are specified as their index in a predefined vehicle-specific sequence.</description>
       </entry>
       <entry name="MOTOR_TEST_ORDER_BOARD" value="2">
-        <description>motor numbers are specified as the output as labeled on the board</description>
+        <description>Motor numbers are specified as the output as labeled on the board.</description>
```
hamishwillee: Does this actually mean that if board labels MAIN 1-8, AUX1-8 then motors will be run in that order?

message_definitions/v1.0/common.xml
```diff
@@ -1346,13 +1346,13 @@
...
-        <param index="1">motor number (a number from 1 to max number of motors on the vehicle)</param>
-        <param index="2">throttle type (0=throttle percentage, 1=PWM, 2=pilot throttle channel pass-through. See MOTOR_TEST_THROTTLE_TYPE enum)</param>
-        <param index="3">throttle</param>
-        <param index="4">timeout (in seconds)</param>
-        <param index="5">motor count (number of motors to test to test in sequence, waiting for the timeout above between them; 0=1 motor, 1=1 motor, 2=2 motors...)</param>
-        <param index="6">motor test order (See MOTOR_TEST_ORDER enum)</param>
+        <description>Mission command to perform motor test.</description>
+        <param index="1" label="Motor Number" minValue="1" increment="1">Number of motor to test (from 1 to max number of motors on vehicle).</param>
+        <param index="2" label="Throttle Type" enum="MOTOR_TEST_THROTTLE_TYPE">Throttle type. Defines whether the Throttle Value (param3) is a percentage, PWM value, etc.).</param>
+        <param index="3" label="Throttle Value">Throttle value.</param>
```
hamishwillee: Is this the throttle value to be used when running the test?

message_definitions/v1.0/common.xml
```diff
@@ -2752,29 +2752,31 @@
...
-        <description>throttle as an absolute PWM value (normally in range of 1000~2000)</description>
+        <description>Throttle as an absolute PWM value (normally in range of 1000~2000).</description>
       </entry>
       <entry value="2" name="MOTOR_TEST_THROTTLE_PILOT">
-        <description>throttle pass-through from pilot's transmitter</description>
+        <description>Throttle pass-through from pilot's transmitter.</description>
       </entry>
       <entry value="3" name="MOTOR_TEST_COMPASS_CAL">
-        <description>per-motor compass calibration test</description>
+        <description>Per-motor compass calibration test.</description>
```
hamishwillee: This seems odd - it isn't a throttle value. What is the intent/usage here?

message_definitions/v1.0/common.xml
hamishwillee: @bkueng You know about motor tests right? 
- Does this PR seem sane?
- Do you know how motor numbers to board map? My feeling/guess is that this is probably rubbish - the numbers probably map to however the underlying flight stack labels them - for PX4 this would be the motor order as listed in the Airframe reference. Can you clarify?

bkueng: You're asking the questions I had as well during implementation. I opted to only implement the simplest cases, and return unsupported for the rest. See https://github.com/PX4/PX4-Autopilot/blob/master/src/modules/commander/Commander.cpp#L1391.
- there's no progress update, just ack/nack the command
- needs to be sent outside of arming (so outside of missions as well)
- mapping is just directly MAIN 1-N, AUX is not supported. The message is IMO too limited in that regard, and a GCS needs to know about supported outputs (UAVCAN, virtual outputs, etc), which can then be specified in the test message.

hamishwillee: @bkueng Thanks for this - the PX4 implementation looks pragmatic. I do wonder  where we should document PX4 implementation details like this. Thoughts?

Do you think it is reasonable to just clarify the mapping as it is (because I think ArduPilot use the same approach). So we might say for the motor instance field:

> Motor instance number. The numbers correspond directly to PWM MAIN but outputs (only motors on MAIN bus can be tested using this command).

Further, is it worth designing a better message, and any suggestions for how to do that? One thing to consider though is getting rid of all the motor test order/timing stuff from the message - the GCS can construct tests however it likes if it has some idea of what motor are available.

I'm going to merge this for now, because currently all the changes are non-controversial. I will reopen another based on your comments.

hamishwillee:
I am merging this because all the changes are effectively "typo" fixes/readability improvements. The only real additional context added is that the description says this is a "Command" rather than a "Mission command". That is important because this cannot be used in missions.

The message is pretty flawed. I will include any follow up in a separate PR.

message_definitions/v1.0/common.xml
bkueng: Not sure what ArduPilot does.

> Further, is it worth designing a better message, and any suggestions for how to do that?

Not yet, I want to get to that when a GCS has more output information (bigger topic) for the motor & output indexing. Yes the rest can go away, with some sort of timing for safety.

hamishwillee: OK. Ping me when you have something on this you want me to do.

PR #1068, opened 2019-01-25
Add MAV_BATTERY_CHARGE_STATE_CHARGING

okalachev:
I propose adding a charging state of the battery.

This enum value would be useful for systems with automatic charging stations.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2603,7 +2603,7 @@
       </entry>
     </enum>
     <enum name="MAV_BATTERY_CHARGE_STATE">
-      <description>Enumeration for low battery states.</description>
+      <description>Enumeration for battery charge states.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Makes sense to me, and there is no compatibility issue. @auturgy @LorenzMeier @MaEtUgR - any thoughts on this one?

MaEtUgR:
Ah but I just saw it should probably say "Enumeration for ~low~ battery states." in the description of the enum.

hamishwillee:
Thanks @MaEtUgR !

@okalachev IMO better even to say "Enumeration for battery charge states." in the description. THoughts?

auturgy:
Sounds good 

Regards,

James

okalachev:
@hamishwillee, ok, done!

hamishwillee:
Thanks @okalachev , all. Merged.

PR #1070, opened 2019-01-31
SERVO_OUTPUT_RAW et al port mapping clarification

hamishwillee:
This updates description of the port field used to differentiate between "banks of servos/channels" in RC_CHANNELS_SCALED, RC_CHANNELS_RAW, SERVO_OUTPUT_RAW to recommend particular values for the known ports on Pixhawk. It does not attempt to define for other systems.

Specifically 
- it removes the comment "Most MAVs will just use one, but this allows for more than 8 servos." (not necessarily true, but only the bit "allows for more than 8 servos potentially useful".
- Adds "Flight stacks running on Pixhawk should use: 0 = MAIN, 1 = AUX." - I think common for PX4 and ArduPilot

OK @auturgy @bkueng ?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3375,7 +3375,7 @@
     <message id="34" name="RC_CHANNELS_SCALED">
       <description>The scaled values of the RC channels received: (-100%) -10000, (0%) 0, (100%) 10000. Channels that are inactive should be set to UINT16_MAX.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint8_t" name="port">Servo output port (set of 8 outputs = 1 port). Most MAVs will just use one, but this allows for more than 8 servos.</field>
+      <field type="uint8_t" name="port">Servo output port (set of 8 outputs = 1 port). Flight stacks running on Pixhawk should use: 0 = MAIN, 1 = AUX.</field>
       <field type="int16_t" name="chan1_scaled">RC channel 1 value scaled.</field>
       <field type="int16_t" name="chan2_scaled">RC channel 2 value scaled.</field>
       <field type="int16_t" name="chan3_scaled">RC channel 3 value scaled.</field>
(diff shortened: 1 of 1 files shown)


auturgy:
For Aux and Main, perhaps we need to articulate that a bit better: it confuses many.
On a board with an iomcu, main = iomcu, aux = fmu. 
On an fmu only board (pixracer, PH4-mini, v5-nano, omnibus, crazyflie etc), main = fmu. We need to catch that inconsistency.

Regards,

James

hamishwillee:
@auturgy 

> For Aux and Main, perhaps we need to articulate that a bit better: it confuses many.
> On a board with an iomcu, main = iomcu, aux = fmu.
> On an fmu only board (pixracer, PH4-mini, v5-nano, omnibus, crazyflie etc), main = fmu. We need to catch that inconsistency.

Thanks James
-  FMU=Flight Management Unit. What is IOMCU acronym? (I assume just means IO board).
- My first reaction was that this isn't really confusing unless you're saying that  port numbers actually map to FMU or IOMCU - ie *Flight stacks running on Pixhawk should use: 0 = FMU, 1 = IOMCU*.
  Otherwise this is less confusing because MAIN is always labeled main, always exists, and is always indexed using 0. Similarly AUX may not be present, but if so, is accessed with 1. I guess an implementer might then need to do the mapping to FMU/IOMCU ...
  - Thoughts?

auturgy:
Yeah. Mapping and documentation. The main confusion it causes is with users setting up DShot, which is only available on fmu pins (obviously ArduPilot specific).

Regards,

James

hamishwillee:
@auturgy So just to be clear, I think you're agreeing that this is OK? I'd prefer to avoid autopilot specifics and talking about board internals _here_ if at all possible. 

@auturgy @bkueng As an aside though, I'm not sure where to capture this information. In the past it would have gone somewhere on pixhawk.org, but now that is a site rather than a wiki. 
While I now put the user-facing [hardware](https://docs.px4.io/en/flight_controller/pixhawk_series.html) docs in PX4 user guide (not very helpful for ArduPilot, but it's what I've got) there doesn't appear to be a good place for this type of generic board design "rules" - ie overview of the FMU/IO architecture. 

I guess I could put a heading/section in doc for every flight controller which explains the mapping - e.g. "MAIN/AUX Ports"

auturgy:
Ideally we’d rename the io outputs aux, so fmu is consistently main, but that’s beyond the scope of this and an entirely different discussion. 
So yes, fine as is.

Regards,

James

PR #1071, opened 2019-01-31
MAV_CMD_NAV_VTOL_TAKEOFF - add param metadata

hamishwillee:
This further improves #1066, adding param metadata that can help GCS better structure UI.

@amilcarlucas Have I missed anything?


hamishwillee:
Thanks for review @julianoes. @amilcarlucas FYI this is merged, but can be further updated if you spot something I missed.

Issue #988, opened 2018-09-19
MAVLink battery information


LorenzMeier:
Questions the message should answer:

 * Charge level
 * Health
 * How long it will last
 * Age / max charge level / capacity
 * Battery ID
 * Any fault indications

```xml
    <message id="334" name="BATTERY_INFO">
      <description>Battery information to communicate once per power cycle.</description>
      <field type="uint32_t" name="serial_number">Battery serial number, UINT32_MAX: battery does not have a serial number</field>
      <field type="uint32_t" name="capacity" units="mAh">Actual capacity of the battery measured by the fuel gauge, UINT32_MAX: battery does not report actual capacity</field>
      <field type="uint16_t" name="cycle_count">Number of charge-discharge cycles reported by the fuel gauge, UINT16_MAX: battery does not report cycle count</field>
    </message>
```

hamishwillee:
There is an [old spec here](http://sbs-forum.org/specs/sbdat110.pdf) that might seed additional information we could include. In particular I suspect it would be useful to be able to remotely monitor the progress of a battery as it is being charged (i.e. if it is an internal battery on a remotely managed charging system).

The messages talked about in this spec are:
```
ManufacturerAccess() (0x00) 13
5.1.2. RemainingCapacityAlarm() (0x01) 13
5.1.3. RemainingTimeAlarm() (0x02) 13
5.1.4. BatteryMode() (0x03) 14
5.1.5. AtRate() (0x04) 21
5.1.6. AtRateTimeToFull() (0x05) 22
5.1.7. AtRateTimeToEmpty() (0x06) 22
5.1.8. AtRateOK() (0x07) 23
5.1.9. Temperature() (0x08) 23
5.1.10. Voltage() (0x09) 24
5.1.11. Current() (0x0a) 24
5.1.12. AverageCurrent() (0x0b) 24
5.1.13. MaxError() (0x0c) 25
5.1.14. RelativeStateOfCharge() (0x0d) 25
 RemainingCapacity() (0x0f) 26
5.1.17. FullChargeCapacity() (0x10) 26
5.1.18. RunTimeToEmpty() (0x11) 27
5.1.19. AverageTimeToEmpty() (0x12) 27
5.1.20. AverageTimeToFull() (0x13) 27
5.1.21. BatteryStatus() (0x16) 28
5.1.22. CycleCount() (0x17) 32
5.1.23. DesignCapacity() (0x18) 32
5.1.24. DesignVoltage() (0x19) 32
[...]

PR #1072, opened 2019-02-01
SMART_BATTERY_* messages: Added

hamishwillee:
This implements the messages/enum discussed in #988.

Minor delta since then 
- now have both `capacity_full` and `capacity_full_specification`
- The messages are marked `wip`

@LorenzMeier
- You indicated there might be more faults to [include from the spec](http://sbs-forum.org/specs/sbdat110.pdf)? These can be added as needed to the enum, which has plenty of spare space
- You also suggested `cell_offset`, `battery_id` might be changed to `uint16_t` so that voltage array will be ordered last, and may be truncated if empty. Is this something you still want?

Anyone else got any feedback?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2626,6 +2626,21 @@
         <description>Battery is diagnosed to be defective or an error occurred, usage is discouraged / prohibited.</description>
       </entry>
     </enum>
+    <enum name="MAV_SMART_BATTERY_FAULT">
+      <description>Smart battery supply status/fault flags (bitmask) for health indication.</description>
+      <entry value="0" name="MAV_SMART_BATTERY_FAULT_OK">
+        <description>Battery has no fault indications.</description>
+      </entry>
+      <entry value="1" name="MAV_SMART_BATTERY_FAULT_DEEP_DISCHARGE">
+        <description>Battery has deep discharged.</description>
+      </entry>
+      <entry value="2" name="MAV_SMART_BATTERY_FAULT_SPIKES">
+        <description>Voltage spikes.</description>
+      </entry>
+      <entry value="4" name="MAV_SMART_BATTERY_FAULT_SINGLE_CELL_FAIL">
+        <description>Single cell has failed.</description>
+      </entry>
+    </enum>
     <enum name="MAV_VTOL_STATE">
       <description>Enumeration of VTOL states</description>
       <entry value="0" name="MAV_VTOL_STATE_UNDEFINED">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2626,6 +2626,21 @@
         <description>Battery is diagnosed to be defective or an error occurred, usage is discouraged / prohibited.</description>
       </entry>
     </enum>
+    <enum name="MAV_SMART_BATTERY_FAULT">
+      <description>Smart battery supply status/fault flags (bitmask) for health indication.</description>
+      <entry value="0" name="MAV_SMART_BATTERY_FAULT_OK">
```
amilcarlucas: If this is a bitmask then 0 should not be defined

message_definitions/v1.0/common.xml
```diff
@@ -5054,6 +5069,33 @@
       <field type="uint8_t" name="severity" enum="MAV_SEVERITY">Severity of status. Relies on the definitions within RFC-5424.</field>
       <field type="char[254]" name="text">Status text message, without null termination character.</field>
     </message>
+    <!-- Smart battery messages -->
+    <message id="370" name="SMART_BATTERY_INFO">
+      <wip/>
```
amilcarlucas: Please add the same wip comment that other wip messages also have.

hamishwillee:
@MaEtUgR,  @apache405, @Naterater, @WickedShell , @AlexKlimaj FYI, I added devcall label, so hopefully this will get final review and merged this week. If there is any particular remaining feedback on this, please let me know in the next day or so.

AlexKlimaj:
It's looking great! Glad to see this moving forward. 

My only comment is maybe adding a couple more fault states. One for over current and one under/over temperature.

hamishwillee:
Thanks @AlexKlimaj  - let's add them along with any others under discussion following dev meeting.

FYI, Tests failing because grams not yet supported as a unit. Updated in https://github.com/ArduPilot/pymavlink/pull/265

julianoes:
How does this relate to `BATTERY_STATUS`? Does it deprecate it or is it an alternative next to it? Or should we extend `BATTERY_STATUS` alternatively?

hamishwillee:
> How does this relate to `BATTERY_STATUS`? Does it deprecate it or is it an alternative next to it? Or should we extend `BATTERY_STATUS` alternatively?

Good question. I've been thinking of these as running in parallel for smart batteries. However generally I think this could replace the existing message - @MaEtUgR  @LorenzMeier - what was your thinking?

We can't really extend BATTERY_STATUS - this is two messages.

hamishwillee:
Note, 
- Works for both smart battery to flight controller, and from flight controller to ground station. 
- Replacement for BATTERY_STATUS for smart batteries.

message_definitions/v1.0/common.xml
```diff
@@ -5057,6 +5069,35 @@
...
+      <field type="int16_t" name="current" units="cA">Battery current (through all cells/loads). Positive if discharging, negative if charging. UINT16_MAX: field not provided.</field>
+      <field type="int16_t" name="temperature" units="cdegC">Battery temperature. -1: field not provided.</field>
+      <field type="uint8_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
+      <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages.  Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
+      <field type="int32_t" name="fault_bitmask" display="bitmask" enum="MAV_SMART_BATTERY_FAULT">Fault/health indications. -1: field not provided.</field>
+      <field type="int32_t" name="time_remaining" units="s">Estimated remaining battery time. -1: field not provided.</field>
```
amilcarlucas: Would it not make more sense to move the voltages array to the last field in order to let mavlink2 automatically truncate the array ?

message_definitions/v1.0/common.xml
```diff
@@ -5057,6 +5069,35 @@
...
+    <message id="370" name="SMART_BATTERY_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Smart Battery information (static/infrequent update).</description>
+      <field type="uint8_t" name="id">Battery ID</field>
+      <field type="int32_t" name="capacity_full_specification" units="mAh">Capacity when full according to manufacturer, -1: field not provided.</field>
+      <field type="int32_t" name="capacity_full" units="mAh">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
+      <field type="uint16_t" name="cycle_count">Charge/discharge cycle count. -1: field not provided.</field>
+      <field type="int32_t" name="serial_number">Serial number. -1: field not provided.</field>
+      <field type="char[50]" name="device_name">Static device name. Encode as manufacturer and product names separated using an underscore.</field>
```
amilcarlucas: Would it not make more sense to move the device_name to the last field in order to let mavlink2 automatically truncate the array ?

hamishwillee: @amilcarlucas No need - reordering automatically does that for us (smallest items are reordered to the end, with items of the same size in the same order - as long as this is below the `id` (which has the same size) it will be ordered last.

message_definitions/v1.0/common.xml
hamishwillee: @amilcarlucas This one is more complicated. The ordering is by size first, and only based on declaration order within the common sizes.

So to move the voltages to the end I would need to change `id` and `cell_offset` from `uint8_t` to `uint16_t`. So basically 2 bytes extra for every message. That said, I think it is worth it, because we know that as long as we have less than 14 cells we're saving space.

Seems a reasonable tradeoff (until someone decides to create and extension message that breaks this. I'll change.

hamishwillee: Done

hamishwillee:
OK, I made some final changes to:
- Add the new faults suggested by @AlexKlimaj 
- Increase size of `SMART_BATTERY_STATUS.id` and `cell_offset` to uint_16 so that battery voltage array can be truncated if needed (cost is +2 bytes per message likely unused, saving up to 15x2 bytes depending on how many cells used.
- Added notes about when used to both this and BATTERY_STATUS (ie use this for smart batteries, and BATTERY_STATUS for non-smart batteries)
- Added note that this is used for both battery to flight stack, and flight stack to GCS. 
- BATTERY_STATUS - added note that this is for flight stack to GCS.

Most of that was approved in dev call. The rest is non-controversial. I am therefore merging.

PR #1073, opened 2019-02-05
Radio link information and settings related messages

aviaks:
Added new message to get the information about radio link status and change settings. 

**Question / Suggestion:** 
New `RADIO_AP_SETTINGS` message to configure access point currently duplicates [WIFI_CONFIG_AP](https://mavlink.io/en/messages/common.html#WIFI_CONFIG_AP). 
It makes sense to leave only one and depending on how widely `WIFI_CONFIG_AP` is currently used I would suggest to refactor it to `RADIO_AP_SETTINGS`. If it's not possible - existing `WIFI_CONFIG_AP` would also be okay and `RADIO_AP_SETTINGS` can be removed.


message_definitions/v1.0/common.xml
```diff
@@ -1612,6 +1612,26 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="550" name="MAV_CMD_REQUEST_RADIO_INFORMATION">
+        <description>Request radio module information. System will respond with RADIO_INFORMATION message.</description>
+        <param index="1">0: No action 1: Request radio module settings</param>
```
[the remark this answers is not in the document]
aviaks: Yes, I agree this parameter is a bit useless. It was added similar to e.g. existing MAV_CMD_REQUEST_CAMERA_SETTINGS (522) but it has second parameter reserved.

message_definitions/v1.0/common.xml
```diff
@@ -4870,6 +4890,34 @@
       <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise</field>
       <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
     </message>
+    <message id="280" name="RADIO_INFORMATION">
```
LorenzMeier: This is pretty similar to: https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L4073

message_definitions/v1.0/common.xml
```diff
@@ -1612,6 +1612,26 @@
...
+      </entry>
+      <entry value="552" name="MAV_CMD_SET_RADIO_RF_SETTINGS">
+        <description>Set radio module RF settings. Use NaN for unchanged values. </description>
+        <param index="1">Frequency in MHz</param>
+        <param index="2">Bandwidth in MHz</param>
+        <param index="3">Local TX power in dBm</param>
+        <param index="4">Remote TX power in dBm</param>
+        <param index="5">Wireless distance in meters </param>
+      </entry>
+      <entry value="553" name="MAV_CMD_REQUEST_RADIO_AP_SETTINGS">
```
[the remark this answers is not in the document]
aviaks: The link is secured so setting SSID / password is safe. We definitely need to query at least SSID to show the current name in the app. I think, getting the password should be available as an option but querying it or not will depend on the security of the link and particular implementation.

message_definitions/v1.0/common.xml
aviaks: Agree, it's very similar but doesn't contain transmitted/received bytes/packets. Can they be added to RADIO_STATUS?

message_definitions/v1.0/common.xml
```diff
@@ -1612,6 +1612,26 @@
...
+        <description>Request radio module information. System will respond with RADIO_INFORMATION message.</description>
+        <param index="1">0: No action 1: Request radio module settings</param>
+      </entry>
+      <entry value="551" name="MAV_CMD_REQUEST_RADIO_RF_SETTINGS">
+        <description>Request radio module RF settings. System will respond with RADIO_RF_SETTINGS message.</description>
+        <param index="1">0: No action 1: Request radio module RF settings</param>
+      </entry>
+      <entry value="552" name="MAV_CMD_SET_RADIO_RF_SETTINGS">
+        <description>Set radio module RF settings. Use NaN for unchanged values. </description>
+        <param index="1">Frequency in MHz</param>
```
amilcarlucas: Remove the units from the description and add them using the units attribute.
also add a label text to each parameter.

aviaks:
Thanks for the feedback! Requested changes done.

message_definitions/v1.0/common.xml
hamishwillee: They might be added as [MAVLink2 message extensions](https://mavlink.io/en/guide/define_xml_element.html#message_extensions) - essentially this would mean that you would only get the new fields when communicating with MAVLink 2 systems - they would not work on MAVLink 1 only systems.

message_definitions/v1.0/common.xml
```diff
@@ -1612,6 +1612,17 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="550" name="MAV_CMD_REQUEST_RADIO_SETTINGS">
```
hamishwillee: I believe the model to request a particular message be streamed is to call [MAV_CMD_SET_MESSAGE_INTERVAL](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL) (and MAV_CMD_GSET_MESSAGE_INTERVAL to find the current interval)

message_definitions/v1.0/common.xml
hamishwillee: @LorenzMeier I assume adding extensions is "generally" better for all concerned?  

NOte that the RADIO_STATUS message is not well documented - https://github.com/mavlink/mavlink/pull/1078.

aviaks: @hamishwillee This approach is fine too, RADIO_INFORMATION removed and new fields added to RADIO_STATUS as extension

message_definitions/v1.0/common.xml
aviaks: This message is actually not supposed to be streamed all the time. The idea is to request this message only when we want to sync and show the current settings in GCS app. 
The approach for this message should be similar to already implemented https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_CAMERA_SETTINGS

peterbarker: FWIW I've been thinking it would be nice to have a `MAV_CMD_REQUEST_MESSAGE` which takes a mavlink message ID.... basically a one-shot equivalent of `MAV_CMD_SET_MESSAGE_INTERVAL`

hamishwillee: Then I think what you're doing here works.

hamishwillee: @peterbarker Yes, I'll add a PR for that to be discussed.

message_definitions/v1.0/common.xml
```diff
@@ -4093,13 +4104,19 @@
...
-      <field type="uint8_t" name="remrssi">Remote signal strength</field>
-      <field type="uint8_t" name="txbuf" units="%">Remaining free buffer space.</field>
-      <field type="uint8_t" name="noise">Background noise level</field>
-      <field type="uint8_t" name="remnoise">Remote background noise level</field>
-      <field type="uint16_t" name="rxerrors">Receive errors</field>
-      <field type="uint16_t" name="fixed">Count of error corrected packets</field>
+      <field type="uint8_t" name="rssi">Local received signal strength indication. Values: [0-254], 255: invalid/unknown.</field>
+      <field type="uint8_t" name="remrssi">Remote received signal strength indication. Values: [0-254], 255: invalid/unknown.</field>
+      <field type="uint8_t" name="txbuf" units="%">Remaining free transmitter buffer space.</field>
+      <field type="uint8_t" name="noise" units="dB">Local background noise level.</field>
```
hamishwillee: I've been getting clarification on the units. For noise these are not db but device dependent rssi units - you can see the new definitions here: https://github.com/mavlink/mavlink/pull/1078

Are these still OK for  you?

message_definitions/v1.0/common.xml
peterbarker: @hamishwillee Amusing!

I was doing that in parallel: https://github.com/peterbarker/mavlink/commit/802f2795aa218958d699102f9517ca0ee5008b95

... and ArduPilot support for same: https://github.com/ArduPilot/ardupilot/pull/10457

hamishwillee: @peterbarker :-) Well, thanks for proposing it. Better for everyone if it is in common!

aviaks: That's very good idea, then I'm removing `MAV_CMD_REQUEST_CAMERA_SETTINGS`

message_definitions/v1.0/common.xml
aviaks: Physical units would have been more relevant but existing units are also fine. I changed and committed according to your definitions in #1078

message_definitions/v1.0/common.xml
hamishwillee: Thanks. It is still going to take a little while before that gets in!

message_definitions/v1.0/common.xml
hamishwillee: Great. I've merged the RADIO_STATUS changes, so you'll have to resolve those changes. It's a good thing - makes it easier to see the new extensions on RADIO_STATUS

@LorenzMeier @amilcarlucas @auturgy Anything else to add to this?

message_definitions/v1.0/common.xml
```diff
@@ -4891,6 +4905,16 @@
...
       <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
     </message>
+    <message id="281" name="RADIO_SETTINGS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Current settings of radio module. This message is sent in response to MAV_CMD_REQUEST_RADIO_SETTINGS command.</description>
+      <field type="float" name="frequency" units="MHz">Radio module operating frequency</field>
+      <field type="float" name="bandwidth" units="MHz">Radio module operating bandwidth</field>
+      <field type="float" name="local_tx_power" units="dB">Transmission power of local module (ground)</field>
+      <field type="float" name="remote_tx_power" units="dB">Transmission power of remote module (air)</field>
```
vooon: Would you mean dBm?

message_definitions/v1.0/common.xml
```diff
@@ -4093,13 +4101,19 @@
...
+      <field type="uint8_t" name="noise">Local background noise level. These are device dependent RSSI values (scale as approx 2x dB on SiK radios).</field>
+      <field type="uint8_t" name="remnoise">Remote background noise level. These are device dependent RSSI values (scale as approx 2x dB on SiK radios).</field>
+      <field type="uint16_t" name="rxerrors">Count of receive errors (since boot).</field>
+      <field type="uint16_t" name="fixed">Count of error corrected packets (since boot).</field>
+      <extensions/>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot)</field>
+      <field type="uint64_t" name="tx_bytes" units="bytes">The number of transmitted bytes</field>
+      <field type="uint64_t" name="rx_bytes" units="bytes">The number of received bytes</field>
+      <field type="uint32_t" name="tx_packets">The number of transmitted packets</field>
+      <field type="uint32_t" name="rx_packets">The number of received packets</field>
```
vooon: What packets exactly: mavlink or radio itself?

message_definitions/v1.0/common.xml
aviaks: Right, dBm is more relevant here. Thanks!

message_definitions/v1.0/common.xml
aviaks: In my case radio packets. But depending on implementation I think it can be mavlink too since not specified for the other fields of the message, for instance:
```
<field type="uint16_t" name="fixed">Count of error corrected packets (since boot).</field>
```

hamishwillee: It probably should be specified to be "radio" - I'll sort that out for the original fields. 

Here *The number of received packets* to *The number of received radio packets* and same for tx

message_definitions/v1.0/common.xml
```diff
@@ -4896,6 +4910,16 @@
       <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise</field>
       <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view</field>
     </message>
+    <message id="281" name="RADIO_SETTINGS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Current settings of radio module. This message is sent in response to MAV_CMD_REQUEST_RADIO_SETTINGS command.</description>
```
hamishwillee: in response to a MAV_CMD_REQUEST_MESSAGE command ...

message_definitions/v1.0/common.xml
```diff
@@ -1617,6 +1617,14 @@
         <param index="2">Focus value</param>
         <param index="3">Reserved (all remaining params)</param>
       </entry>
+      <entry value="551" name="MAV_CMD_SET_RADIO_SETTINGS">
+        <description>Set radio module settings. Use NaN for unchanged values.</description>
+        <param index="1" label="Frequency" units="MHz">Radio module operating frequency</param>
```
hamishwillee: Use consistent full stops across descriptions.

message_definitions/v1.0/common.xml
aviaks: Thanks! Now specified to "radio".

message_definitions/v1.0/common.xml
aviaks: Thanks, fixed

message_definitions/v1.0/common.xml
aviaks: Sorry, could you specify more details on this please? Everything looks good to me and consistent with other descriptions.

message_definitions/v1.0/common.xml
```diff
@@ -4105,6 +4115,12 @@
       <field type="uint8_t" name="remnoise">Remote background noise level. These are device dependent RSSI values (scale as approx 2x dB on SiK radios).</field>
       <field type="uint16_t" name="rxerrors">Count of receive errors (since boot).</field>
       <field type="uint16_t" name="fixed">Count of error corrected packets (since boot).</field>
+      <extensions/>
```
amilcarlucas: This requires support from SiK Firmware. @tridge is this doable ?

message_definitions/v1.0/common.xml
hamishwillee: Actually you are right. I would like all content to be proper sentences, but we inconsistently only add full stops when there are two sentences.

Ignore.

message_definitions/v1.0/common.xml
hamishwillee: @amilcarlucas Do we need to define a "not supported value" for these?

hamishwillee:
@aviaks We discussed this PR in the dev call last night.

The conclusion was that:
- the fields will not solve the whole class of radios and in many cases aren't all that useful except during testing- they don't really belong in the common message set
- once you start working on this you will discover that the fields aren't sufficient.
- this is probably not the best way to approach the problem

A similar problem was solved in [ESP8226 configuration](https://github.com/dogmaphobic/mavesp8266) using the parameter protocol. This allows us to scale the problem and compensate for the differences in what radios need. It also means that you can add new radio support without having to implement protocol handling - just parameters.

If you would like to attend the [dev call](https://mavlink.io/en/about/support.html#dev_call) in two weeks someone could talk you through the approach for the ESP8226?

You can of course still use messages if you like, but please put them into your own [dialect](https://mavlink.io/en/messages/). I am closing the PR because not suitable for common. Thanks for posting it and working with us - sorry it took so long to get to a proper response.

aviaks:
@hamishwillee thanks for the comments and absolutely no worries about that - the goal it to make sure that it's properly integrated, scalable and everybody is happy with the solution. 
[ESP8226 configuration](https://github.com/dogmaphobic/mavesp8266) looks as a bit of overhead but thanks for pointing to that.

I wanted to attend dev call this Wednesday and joined [this](https://zoom.us/j/990208968) meeting at 9AM CET as per [here](https://mavlink.io/en/about/support.html#dev_call), however after waiting for about 20 minutes nobody else joined. Could it be that there is a different link or scheduled time?

hamishwillee:
@aviaks Very sorry to hear that - would have been great to be able to work through this with you. 

You did exactly the right thing. 
Unfortunately when everyone else tried to join that meeting it was closed. After a short while we co-ordinated a new meeting room using the PX4 #mavlink slack channel: https://px4.slack.com/messages/C52DXKYLV - it must have been opened afterwards (we've been fighting ongoing problems with Zoom meeting setup).
It is worth joining that slack channel so you can see any last minute problems.

PR #1076, opened 2019-02-06
initial stalebot configuration

dagar:


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+daysUntilClose: 30
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
```
hamishwillee: As a suggestion, we could use this as an opportunity to ask for help?

> This issue has been automatically marked as stale, and will be closed if no further activity occurs. Add a comment if it is still affecting you - or if you can help us fix it!

hamishwillee:
Looks great! Do we want to separate out feature requests from bugs (ie make the features exempt or on a different cycle?). 

This is immediately going to make a lot of items stale. Can you quickly review all of those items and reactivate the ones you consider still useful?

dagar:
You mean different thresholds for issues vs PRs?

hamishwillee:
> You mean different thresholds for issues vs PRs?

No, I mean that a bug probably is interesting for a certain amount of time, but perhaps a feature request can live in a backlog for longer. I would expect those to be labels.

However there are no good concrete examples, so I'd be happy for this to go forward and we update if we get a real case.

PR #1077, opened 2019-02-06
Reserving part of the component ID range for private components.

edwinhayes:
Per https://github.com/mavlink/qgroundcontrol/issues/7049#issuecomment-447726922 (apologies for the delay in actioning this), I propose the C25-C99 range be reserved for private component definitions. A quick look at the existing definition files indicates this should not cause any conflicts.

To clarify, the definition of 'private components' is essentially "things which can be incompatible with other private components". This should exclude anything which is sold as a component to the general public. In Aeronavics' case, we have a number of avionics components, both physical and logical, which don't match any of the existing definitions, are important to us, but would be meaningless to other people if they were taking up address space in the public definitions.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -378,6 +378,7 @@
       <entry value="1" name="MAV_COMP_ID_AUTOPILOT1">
         <description>System flight controller component ("autopilot"). Only one autopilot is expected in a particular system.</description>
       </entry>
+      <!-- Component ids from 25-99 are reserved for private OEM component definitions; components in this range may be incompatible with other private components. -->
       <entry value="100" name="MAV_COMP_ID_CAMERA">
         <description>Camera #1.</description>
       </entry>


hamishwillee:
@auturgy @LorenzMeier The idea of a private range of components makes sense to me, provided (as indicated) the person managing the network can be sure there are no other things using those private ids.

@edwinhayes 
- Any particular reason for reserving this many ids?
- It might be safer to reserve these by explicitly declaring each item in the range. I guess the downside is that you can't then define them how you like in your own dialect. Thoughts?

edwinhayes:
@hamishwillee regarding the size of the reserved region:

I left some space at the bottom for 'special' definitions which which a low number might be desirable for some reason, then took a large contiguous block up to the next occupied address.

I reason that making the region too large doesn't seem too great a concern, since the region can always be made smaller against in the future; users of this region are (basically by definition) those for whom it's feasible to reallocate the addresses they use, the presence of the region just helps reduce how often they might be forced to do so.

Regarding how they are reserved: I just copied a comment I found further down which denoted some other reserved range (for commands I think).  Definitely need for individual dialects to be able to define the individual addresses how they want.

auturgy:
Seems reasonable to me.

hamishwillee:
Thanks @edwinhayes . Sounds reasonable to me too. But let's give this a little while to see if anyone else has anything else to add.

edwinhayes:
@hamishwillee - It's been a while without complain; can this be progressed?

hamishwillee:
@edwinhayes Sorry for the delay. This fell off my radar.

@julianoes Does this make sense to you? Is the range reasonable? Ie approx 75 component IDs reserved for private use?

julianoes:
I'm not opposed but it does sound like (too) many.

hamishwillee:
Cheers. @edwinhayes Any particular reason for that size range?/how many you really need. Of course we don't know how many we will need either. I was thinking of maybe allocating the top 25 or so.

edwinhayes:
I checked; we currently have 28 non-standard IDs defined.  

My reasoning for the size of the block proposed is essentially:

* The common set is only using something like 40 out of 254 at the moment, so it's not like reserving this large a chunk will impact the availability of public IDs for allocation in the foreseeable future.
* The size of the private reserved space _can_ be reduced in size in the future if there is a need for additional public IDs; this just requires the people using the private region to reallocate the IDs used in their ecosystem - but by definition people using this region should have that capability.
* The size of the private reserved space _cannot_ (easily) be increased in the future if someone needs additional private IDs, since that would require reallocating existing public IDs, which may break things used by consumers in the wild.

Thus, it makes sense to reserve as large a contiguous block as possible for this purpose.

hamishwillee:
The reasoning is good. I'm happy to allocate this range, but I propose to create my own PR that individually allocates each item as this allows tooling to prevent reallocation in another file.

Do you think worth adding a note that users should allocate the private ids sequentially from the lower values - so that if we are forced to reallocate this space we will cut from the top and reduce impact on them?

edwinhayes:
Ok sure.  Yes, I think that's probably sensible.

hamishwillee:
OK, so before I extend out the copies. How about this definition:
```
      <!-- Component ids from 25-99 are reserved for private OEM component definitions (and may be incompatible with other private components). Note that if this range is later reduced, higher ids will be reallocated first. -->
      <entry value="25" name="MAV_COMP_ID_USER1">
        <description>Id1 for component on privately managed MAVLink network. Can be used for any purpose but may not be published by components outside of the private network.</description>
      </entry>
```

edwinhayes:
Seems good to me.

hamishwillee:
Replaced by #1184

PR #1078, opened 2019-02-07
[WIP] RADIO_STATUS: Add field definitions

hamishwillee:
This adds field definitions and units where appropriate. It is WIP because I've guessed at some of these and have questions inline. Note that the `rssi` is as discussed in the dev call. 

@auturgy Appreciate your thoughts too.


message_definitions/v1.0/common.xml
```diff
@@ -4093,13 +4093,13 @@
...
     <message id="109" name="RADIO_STATUS">
       <description>Status generated by radio and injected into MAVLink stream.</description>
-      <field type="uint8_t" name="rssi">Local signal strength</field>
-      <field type="uint8_t" name="remrssi">Remote signal strength</field>
-      <field type="uint8_t" name="txbuf" units="%">Remaining free buffer space.</field>
-      <field type="uint8_t" name="noise">Background noise level</field>
-      <field type="uint8_t" name="remnoise">Remote background noise level</field>
-      <field type="uint16_t" name="rxerrors">Receive errors</field>
-      <field type="uint16_t" name="fixed">Count of error corrected packets</field>
+      <field type="uint8_t" name="rssi">Local recieved signal strength indication. Values: [0-254], 255: invalid/unknown.</field>
```
hamishwillee: What is difference between local and remote rssi?

I might guess that the local value is for the sending system and the rem value is for the receiving system, but I don't know how the sender would know the rssi of the remote system, and why it would care.

message_definitions/v1.0/common.xml
```diff
@@ -4093,13 +4093,13 @@
...
+      <field type="uint8_t" name="remrssi">Remote recieved signal strength indication. Values: [0-254], 255: invalid/unknown.</field>
+      <field type="uint8_t" name="txbuf" units="%">Remaining free transmitter buffer space.</field>
+      <field type="uint8_t" name="noise" units="dB">Local background noise level.</field>
```
hamishwillee: Here I have assumed dB units, and that noise is actually signal to noise ratio
- is this correct?
- If so, should I document as *Local background signal to noise ratio*?
- Again, how does this work for noise vs remnoise.

message_definitions/v1.0/common.xml
```diff
@@ -4093,13 +4093,13 @@
...
+      <field type="uint8_t" name="remnoise" units="dB">Remote background noise level.</field>
+      <field type="uint16_t" name="rxerrors">Count of receive errors.</field>
```
hamishwillee: I assume counts are since boot? Will these ever overflow, and if so, what should the system do?

hamishwillee:
Also, what are the send semantics? Is this always streamed, and if so, at what rate? Or is the idea that this is requested and set using MAV_CMD_GET_MESSAGE_INTERVAL and MAV_CMD_SET_MESSAGE_INTERVAL  (or both) ?

message_definitions/v1.0/common.xml
amilcarlucas: These messages are filled by the SiK FW that @tridge wrote. @tridge can you take a quick look at this ?

message_definitions/v1.0/common.xml
edwinhayes: Well for bidirectional radios, 'sending' and 'receiving' isn't well defined.  But in the case of some devices, RSSI information is sent back from the 'other' end of the link as part of the underlying radio protocol, so that (assuming there actually is a link) the radio at each end knows both its own RSSI and also what the other end is seeing.

Including both the local and remote RSSI in one message is helpful, because it only requires the radio at one end to be connected to (or natively) a device which understands MAVLink, though I guess an argument could be made that this should instead be exposed as two different MAVLink components.

message_definitions/v1.0/common.xml
tridge: no, units are not dB, they are device dependent RSSI values. They scale as approx 2x dB on SiK radios

message_definitions/v1.0/common.xml
tridge: yes, count since boot, as as these are just for logging, wrap isn't really of concern

message_definitions/v1.0/common.xml
hamishwillee: THanks very much @tridge . I have updated appropriately.

hamishwillee:
I am self-merging following clarifications from tridge and approval from amilcar, and definition of rssi as agreed in devcall.

PR #1079, opened 2019-02-08
Add MAV_CMD_REQUEST_MESSAGE

hamishwillee:
This adds a mavlink command to request a particular message back from a system (as suggested by @peterbarker here: https://github.com/mavlink/mavlink/pull/1073#discussion_r254929733). This is intended to request information that is occasionally needed/doesn't need to be streamed. It is like a one-shot version of MAV_CMD_SET_MESSAGE_INTERVAL.

Alternative is to update MAV_CMD_SET_MESSAGE_INTERVAL with a new parameter that does the same thing - but I don't think that is as clean.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1541,6 +1541,10 @@
         <param index="1">The MAVLink message ID</param>
         <param index="2">The interval between two messages, in microseconds. Set to -1 to disable and 0 to request default rate.</param>
       </entry>
+      <entry value="512" name="MAV_CMD_REQUEST_MESSAGE">
+        <description>Request target system(s) respond with a specified message. This is a "one-shot" version of MAV_CMD_SET_MESSAGE_INTERVAL.</description>
+        <param index="1">The MAVLink message ID of the requested message.</param>
+      </entry>
       <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION">
         <description>Request MAVLink protocol version compatibility</description>
         <param index="1">1: Request supported protocol versions by all nodes on the network</param>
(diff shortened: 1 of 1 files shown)


peterbarker:
This would replace the need for some poll-the-autopilot messages, like these:

```
      <entry value="520" name="MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES">
        <description>Request autopilot capabilities</description>
        <param index="1">1: Request autopilot version</param>
      </entry>
      <entry value="521" name="MAV_CMD_REQUEST_CAMERA_INFORMATION">
        <description>Request camera information (CAMERA_INFORMATION).</description>
        <param index="1">0: No action 1: Request camera capabilities</param>
      </entry>
      <entry value="522" name="MAV_CMD_REQUEST_CAMERA_SETTINGS">
        <description>Request camera settings (CAMERA_SETTINGS).</description>
        <param index="1">0: No Action 1: Request camera settings</param>
      </entry>
```

OTOH, the storage one:

```
      <entry value="525" name="MAV_CMD_REQUEST_STORAGE_INFORMATION">
        <description>Request storage information (STORAGE_INFORMATION). Use the command's target_component to target a specific component's storage.</description>
        <param index="1">Storage ID (0 for all, 1 for first, 2 for second, etc.)</param>
        <param index="2">0: No Action 1: Request storage information</param>
      </entry>
```

does use a parameter.

Perhaps we could make the first parameter "instance id, if appropriate"?

hamishwillee:
> Perhaps we could make the first parameter "instance id, if appropriate"?

That is a damn good idea. Default zero. I will add.

LorenzMeier:
Makes a lot of sense.

PR #1085, opened 2019-02-14
Markup reserved values with NaN default values

hamishwillee:
There was previously no formal way to markup reserved parameters and define their default. The definition is now such that:
- if you omit a parameter then it is considered reserved with a default of zero
- you can mark up a parameter as reserved with a default of NaN using: `<param index="3" reserved="True" default="NaN" />`

This PR updates the small number of params that previously just stated "all params are NaN".

Note, params 5,6 also marked as reserved/NAN. This means the message can ONLY be sent in the float variant of the command, because obviously you can't have a NaN in an int field.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2222,7 +2222,7 @@
         <param index="3" label="Pause" minValue="-1" maxValue="1" increment="2">1 to pause triggering, but without switching the camera off or retracting it. -1 to ignore</param>
       </entry>
       <entry value="2500" name="MAV_CMD_VIDEO_START_CAPTURE" hasLocation="false" isDestination="false">
-        <description>Starts video capture (recording).</description>      
+        <description>Starts video capture (recording).</description>
         <param index="1" label="Stream ID" minValue="0" increment="1">Video Stream ID (0 for all streams)</param>
         <param index="2" label="Status Frequency" minValue="0" units="Hz">Frequency CAMERA_CAPTURE_STATUS messages should be sent while recording (0 for no messages, otherwise frequency)</param>
         <param index="3" reserved="True" default="NaN"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1598,9 +1598,9 @@
       <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
         <description>Set camera running mode. GCS will send a MAV_CMD_REQUEST_VIDEO_STREAM_STATUS command after a mode change if the camera supports video streaming.</description>
         <param index="2" enum="CAMERA_MODE">Camera mode</param>
-        <param index="3" reserved="True" default="NaN" />
-        <param index="4" reserved="True" default="NaN" />
-        <param index="7" reserved="True" default="NaN" />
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1596,26 +1596,31 @@
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="530" name="MAV_CMD_SET_CAMERA_MODE">
-        <description>Set camera running mode. Use NaN for reserved values. GCS will send a MAV_CMD_REQUEST_VIDEO_STREAM_STATUS command after a mode change if the camera supports video streaming.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <description>Set camera running mode. GCS will send a MAV_CMD_REQUEST_VIDEO_STREAM_STATUS command after a mode change if the camera supports video streaming.</description>
         <param index="2" enum="CAMERA_MODE">Camera mode</param>
-        <param index="3">Reserved (all remaining params)</param>
+        <param index="3" reserved="True" default="NaN" />
```
hamishwillee: @dogmaphobic **Note the pattern here** (and in following) - I skip index 5 and 6, which will default to reserved with "0". This is because otherwise the message can't be sent in an **INT** variant of the message. Is that OK?

If 5,6, must be an NaN I can update this so that they are, but we should in the command description note that this message may only be sent in a command message that uses floats for all param values - ie a COMMAND_LONG.

Thoughts?

message_definitions/v1.0/common.xml
```diff
@@ -1715,11 +1726,12 @@
         <description/>
         <param index="1">Landing gear ID (default: 0, -1 for all)</param>
         <param index="2">Landing gear position (Down: 0, Up: 1, NaN for no change)</param>
-        <param index="3">Reserved, set to NaN</param>
-        <param index="4">Reserved, set to NaN</param>
+        <param index="2" reserved="True" default="NaN" />
+        <param index="3" reserved="True" default="NaN" />
+        <param index="4" reserved="True" default="NaN" />
         <param index="5">Reserved, set to NaN</param>
```
hamishwillee: @dogmaphobic @auturgy Not sure about this one (`MAV_CMD_AIRFRAME_CONFIGURATION`). Here the message already sets index 4,5 to NaN, which means that this can only be sent in a COMMAND_LONG.

I think we probably have to add that note to the description and also set these properly to NaN (?) because this is already implemented in a system that would bet testing against NaN?

OR can we change this so that these two are defaults of 0?

hamishwillee:
PS I leave stripping out the Empty and Reserved parameters (which will then be reserved with value 0 to a second PR). I want to do that when https://github.com/ArduPilot/pymavlink/pull/269/files is in place to absolutely ensure that generated output does not lose any code definitions (it won't anyway in C, but for Python it might lose a defined constant for the param).

hamishwillee:
FYI @auturgy @WickedShell

message_definitions/v1.0/common.xml
hamishwillee: Decided to add these but not add "need" to send in command long.

hamishwillee:
Self merging. This is same meaning as previously but ensures that a system reading the information knows whether NaN or 0 is the default null value

PR #1089, opened 2019-02-19
common : Add reset counter to external estimation messages

mhkabir:
This is to enable support for deeper integration of external VIO/SLAM pipelines in real-world situations where they may reset and/or the estimate jump while in flight. (e.g when a SLAM system detects a loop-closure). 

The onboard estimator needs to know a reset occurred and appropriately handle it for robust flight control.

Please let me know if anyone can think of anything else we should add here. We have been flying with an equivalent change in a custom dialect, and I believe this is highly needed upstream to make VIO/SLAM a default feature.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3998,6 +3998,7 @@
       <field type="float" name="yaw" units="rad">Yaw angle</field>
       <extensions/>
       <field type="float[21]" name="covariance">Row-major representation of pose 6x6 cross-covariance matrix upper right triangle (states: x_global, y_global, z_global, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
+      <field type="uint8_t" name="reset_counter">Estimate reset counter. This should be incremented when the estimate resets in any of the dimensions (position, velocity, attitude, angular speed). This is designed to be used when e.g an external SLAM system detects a loop-closure and the estimate jumps.</field>
     </message>
     <message id="102" name="VISION_POSITION_ESTIMATE">
       <description>Global position/attitude estimate from a vision source.</description>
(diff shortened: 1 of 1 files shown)


dagar:
You don't want more granularity?

TSC21:
Also I have been considering this for a while now, and it was a feature that was requested several times now, which is add a command (MAV_CMD) that resets the estimator on the autopilot side. This could be something that could work together (or even replace) the addition of this field.

mhkabir:
@dagar

mhkabir:
@dagar 

> You don't want more granularity?

mhkabir:
@dagar 

> You don't want more granularity?

I initially thought about this, but after actually using this for some time (across many different offboard estimators), I found that for almost all the use cases, you don't need more dimensional granularity. For the ones which are designed to support partial resets, transmitting a "full" reset doesn't really hurt.

@TSC21 

> Adding this field to all messages that contain external estimation data;

Agreed, will do.

> Create a new message specific for details regarding the external estimation states

Any solution involving multiple messages becomes non-atomic and prone to race conditions. We tried this before, and there's no way to guarantee that you will receive information about a reset before receiving the first packet of data post-reset, unless you use delays after the reset or other hacks. This does not allow high performance and robustness, especially in case of in-flight resets.

A generic message conveying more information about external estimation states is fine, and this field can be included in there, but this is beyond the scope of what I'm looking to solve here.

> This could be something that could work together (or even replace) the addition of this field.

Work together yes, but not replace, see above.

mhkabir:
@hamishwillee @LorenzMeier @thomasgubler @priseborough thoughts?

hamishwillee:
I don't know enough to usefully comment. My thoughts are:
- I assume the question about granularity was whether 256 is a large enough reset count before the field loops?
- Is there a need to indicate in some way when/that the field is "not supported"?
- If you're going to add the field to other messages, would be good if you could do so.

priseborough:
Unsigned 8 bit value should be OK and field looping is not an issue provided the programmer at the consumer end knows how to take advantage of the properties of unsigned integer operations.

hamishwillee:
@dagar , @jgoppert , @LorenzMeier and @mhkabir @mhkabir @thomasgubler

Is this waiting on anything? @TSC21 Are you convinced by @mhkabir arguments above.

Are you all happy to merge?

thomasgubler:
@jkflying FYI

hamishwillee:
@mhkabir See https://github.com/mavlink/mavlink/pull/1089#issuecomment-467187784 - "if you're going to add to other messages good to do so now".

BTW, When you are happy that everyone is happy to merge, can you ping Lorenz. I'm away after next week for a while, so I won't be following up.

mhkabir:
Rebased and added the change to all external estimation messages.

mhkabir:
@LorenzMeier @thomasgubler Please review.

hamishwillee:
@thomasgubler Says good to go too. So merging now.

PR #1100, opened 2019-03-23
Common.xml: Add attributes to all MAV_CMD parameters

amilcarlucas:
This allows GCSs to parse the .xml files directly and extract all info required to present a GUI to the users.
No more need to manually update some (guessed) information on some local files specific to each GCS.
The information is now explicit and central in the XML files. No duplication, no guesswork!!!!

This should greatly improve the user experience on MissionPlanner and QGC. And reduce the workload for GCS developers, no need to add new code when a new message gets added, just re-parse the .xml flie and regenerate the .exe.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -754,7 +754,7 @@
     </enum>
     <enum name="MAV_CMD">
       <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data.</description>
-      <entry value="16" name="MAV_CMD_NAV_WAYPOINT">
+      <entry value="16" name="MAV_CMD_NAV_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint.</description>
         <param index="1" label="hold" units="s" minValue="0" increment="1">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
         <param index="2" label="accept radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
@@ -764,7 +764,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="17" name="MAV_CMD_NAV_LOITER_UNLIM">
+      <entry value="17" name="MAV_CMD_NAV_LOITER_UNLIM" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint an unlimited amount of time</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
@@ -774,7 +774,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS">
+      <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint for X turns</description>
         <param index="1" label="turns" minValue="0" increment="1">Turns</param>
         <param index="2">Empty</param>
@@ -784,7 +784,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="19" name="MAV_CMD_NAV_LOITER_TIME">
+      <entry value="19" name="MAV_CMD_NAV_LOITER_TIME" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint for X seconds</description>
         <param index="1" label="time" units="s" minValue="0" increment="1">Loiter time.</param>
         <param index="2">Empty</param>
@@ -794,7 +794,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="20" name="MAV_CMD_NAV_RETURN_TO_LAUNCH">
+      <entry value="20" name="MAV_CMD_NAV_RETURN_TO_LAUNCH" hasLocation="false" isDestination="false">
         <description>Return to launch location</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
@@ -804,7 +804,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="21" name="MAV_CMD_NAV_LAND">
+      <entry value="21" name="MAV_CMD_NAV_LAND" hasLocation="true" isDestination="true">
         <description>Land at location.</description>
         <param index="1" label="Abort Alt" units="m">Minimum target altitude if landing is aborted (0 = undefined/use system default).</param>
         <param index="2" label="Land Mode" enum="PRECISION_LAND_MODE">Precision land mode.</param>
@@ -814,7 +814,7 @@
         <param index="6">Longitude.</param>
         <param index="7" units="m">Landing altitude (ground level in current frame).</param>
       </entry>
-      <entry value="22" name="MAV_CMD_NAV_TAKEOFF">
+      <entry value="22" name="MAV_CMD_NAV_TAKEOFF" hasLocation="true" isDestination="true">
         <description>Takeoff from ground / hand</description>
         <param index="1" label="pitch">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
         <param index="2">Empty</param>
@@ -824,7 +824,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="23" name="MAV_CMD_NAV_LAND_LOCAL">
+      <entry value="23" name="MAV_CMD_NAV_LAND_LOCAL" hasLocation="true" isDestination="true">
         <description>Land at local position (local frame only)</description>
         <param index="1" label="target" minValue="0" increment="1">Landing target number (if available)</param>
         <param index="2" label="offset" units="m" minValue="0">Maximum accepted offset from desired landing position - computed magnitude from spherical coordinates: d = sqrt(x^2 + y^2 + z^2), which gives the maximum accepted distance between the desired landing position and the position where the vehicle is about to land</param>
@@ -834,7 +834,7 @@
         <param index="6" label="X position" units="m">X-axis position</param>
         <param index="7" label="Z position" units="m">Z-axis / ground level position</param>
       </entry>
-      <entry value="24" name="MAV_CMD_NAV_TAKEOFF_LOCAL">
+      <entry value="24" name="MAV_CMD_NAV_TAKEOFF_LOCAL" hasLocation="true" isDestination="true">
         <description>Takeoff from local position (local frame only)</description>
         <param index="1" label="pitch" units="rad">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
         <param index="2">Empty</param>
@@ -844,7 +844,7 @@
         <param index="6" label="X position" units="m">X-axis position</param>
         <param index="7" label="Z position" units="m">Z-axis position</param>
       </entry>
-      <entry value="25" name="MAV_CMD_NAV_FOLLOW">
+      <entry value="25" name="MAV_CMD_NAV_FOLLOW" hasLocation="true" isDestination="false">
         <description>Vehicle following, i.e. this waypoint represents the position of a moving vehicle</description>
         <param index="1" label="following" increment="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
         <param index="2" label="ground speed">Ground speed of vehicle to be followed</param>
@@ -854,7 +854,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="30" name="MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT">
+      <entry value="30" name="MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT" hasLocation="false" isDestination="true">
         <description>Continue on the current course and climb/descend to specified altitude.  When the altitude is reached continue to the next command (i.e., don't proceed to the next command until the desired altitude is reached.</description>
         <param index="1" label="action" minValue="0" maxValue="2" increment="1">Climb or Descend (0 = Neutral, command completes when within 5m of this command's altitude, 1 = Climbing, command completes when at or above this command's altitude, 2 = Descending, command completes when at or below this command's altitude.</param>
         <param index="2">Empty</param>
@@ -864,7 +864,7 @@
         <param index="6">Empty</param>
         <param index="7" label="altitude" units="m">Desired altitude</param>
       </entry>
-      <entry value="31" name="MAV_CMD_NAV_LOITER_TO_ALT">
+      <entry value="31" name="MAV_CMD_NAV_LOITER_TO_ALT" hasLocation="true" isDestination="true">
         <description>Begin loiter at the specified Latitude and Longitude.  If Lat=Lon=0, then loiter at the current position.  Don't consider the navigation command complete (don't leave loiter) until the altitude has been reached.  Additionally, if the Heading Required parameter is non-zero the  aircraft will not leave the loiter until heading toward the next waypoint.</description>
         <param index="1" label="heading required" minValue="0" maxValue="1" increment="1">Heading Required (0 = False)</param>
         <param index="2" label="radius" units="m">Radius. If positive loiter clockwise, negative counter-clockwise, 0 means no change to standard loiter.</param>
@@ -874,7 +874,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="32" name="MAV_CMD_DO_FOLLOW">
+      <entry value="32" name="MAV_CMD_DO_FOLLOW" hasLocation="false" isDestination="false">
         <description>Being following a target</description>
         <param index="1" label="system ID" minValue="0" increment="1">System ID (the system ID of the FOLLOW_TARGET beacon). Send 0 to disable follow-me and return to the default position hold mode.</param>
         <param index="2">RESERVED</param>
@@ -884,7 +884,7 @@
         <param index="6">RESERVED</param>
         <param index="7" label="time to land" units="s" minValue="0">Time to land in which the MAV should go to the default position hold mode after a message RX timeout.</param>
       </entry>
-      <entry value="33" name="MAV_CMD_DO_FOLLOW_REPOSITION">
+      <entry value="33" name="MAV_CMD_DO_FOLLOW_REPOSITION" hasLocation="false" isDestination="false">
         <description>Reposition the MAV after a follow target command has been sent</description>
         <param index="1" label="camera q1">Camera q1 (where 0 is on the ray from the camera to the tracking device)</param>
         <param index="2" label="camera q2">Camera q2</param>
@@ -894,7 +894,7 @@
         <param index="6" label="X offset" units="m">X offset from target</param>
         <param index="7" label="Y offset" units="m">Y offset from target</param>
       </entry>
-      <entry value="34" name="MAV_CMD_DO_ORBIT">
+      <entry value="34" name="MAV_CMD_DO_ORBIT" hasLocation="true" isDestination="true">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Start orbiting on the circumference of a circle defined by the parameters. Setting any value NaN results in using defaults.</description>
@@ -906,7 +906,7 @@
         <param index="6">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
         <param index="7">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
       </entry>
-      <entry value="80" name="MAV_CMD_NAV_ROI">
+      <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
         <description>Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicles control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
         <param index="1" label="ROI mode" enum="MAV_ROI">Region of interest mode.</param>
@@ -917,7 +917,7 @@
         <param index="6" label="y">y</param>
         <param index="7" label="z">z</param>
       </entry>
-      <entry value="81" name="MAV_CMD_NAV_PATHPLANNING">
+      <entry value="81" name="MAV_CMD_NAV_PATHPLANNING" hasLocation="true" isDestination="true">
         <description>Control autonomous path planning on the MAV.</description>
         <param index="1" label="local ctrl" minValue="0" maxValue="2" increment="1">0: Disable local obstacle avoidance / local path planning (without resetting map), 1: Enable local path planning, 2: Enable and reset local path planning</param>
         <param index="2" label="global ctrl" minValue="0" maxValue="3" increment="1">0: Disable full path planning (without resetting map), 1: Enable, 2: Enable and reset map/occupancy grid, 3: Enable and reset planned route, but not occupancy grid</param>
@@ -927,7 +927,7 @@
         <param index="6">Longitude/Y of goal</param>
         <param index="7">Altitude/Z of goal</param>
       </entry>
-      <entry value="82" name="MAV_CMD_NAV_SPLINE_WAYPOINT">
+      <entry value="82" name="MAV_CMD_NAV_SPLINE_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint using a spline path.</description>
         <param index="1" label="hold" units="s" minValue="0" increment="1">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
         <param index="2">Empty</param>
@@ -937,7 +937,7 @@
         <param index="6">Longitude/Y of goal</param>
         <param index="7">Altitude/Z of goal</param>
       </entry>
-      <entry value="84" name="MAV_CMD_NAV_VTOL_TAKEOFF">
+      <entry value="84" name="MAV_CMD_NAV_VTOL_TAKEOFF" hasLocation="true" isDestination="true">
         <description>Takeoff from ground using VTOL mode, and transition to forward flight with specified heading.</description>
         <param index="1">Empty</param>
         <param index="2" label="Transition Heading" enum="VTOL_TRANSITION_HEADING">Front transition heading.</param>
@@ -947,7 +947,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="85" name="MAV_CMD_NAV_VTOL_LAND">
+      <entry value="85" name="MAV_CMD_NAV_VTOL_LAND" hasLocation="true" isDestination="true">
         <description>Land using VTOL mode</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
@@ -961,7 +961,7 @@
                     as they were used in some conflicting proposals
                     between PX4 and ArduPilot and need to be kept
                     unused to prevent errors -->
-      <entry value="92" name="MAV_CMD_NAV_GUIDED_ENABLE">
+      <entry value="92" name="MAV_CMD_NAV_GUIDED_ENABLE" hasLocation="false" isDestination="false">
         <description>hand control over to an external controller</description>
         <param index="1" label="enable" minValue="0" maxValue="1" increment="1">On / Off (&gt; 0.5f on)</param>
         <param index="2">Empty</param>
@@ -971,7 +971,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="93" name="MAV_CMD_NAV_DELAY">
+      <entry value="93" name="MAV_CMD_NAV_DELAY" hasLocation="false" isDestination="false">
         <description>Delay the next navigation command a number of seconds or until a specified time</description>
         <param index="1" label="delay" units="s" minValue="-1" increment="1">Delay (-1 to enable time-of-day fields)</param>
         <param index="2" label="hour" minValue="-1" maxValue="23" increment="1">hour (24h format, UTC, -1 to ignore)</param>
@@ -981,7 +981,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="94" name="MAV_CMD_NAV_PAYLOAD_PLACE">
+      <entry value="94" name="MAV_CMD_NAV_PAYLOAD_PLACE" hasLocation="true" isDestination="true">
         <description>Descend and place payload. Vehicle moves to specified location, descends until it detects a hanging payload has reached the ground, and then releases the payload. If ground is not detected before the reaching the maximum descent value (param1), the command will complete without releasing the payload.</description>
         <param index="1" label="Max Descent" minValue="0" units="m">Maximum distance to descend.</param>
         <param index="2">Empty</param>
@@ -991,7 +991,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="95" name="MAV_CMD_NAV_LAST">
+      <entry value="95" name="MAV_CMD_NAV_LAST" hasLocation="false" isDestination="false">
         <description>NOP - This command is only used to mark the upper limit of the NAV/ACTION commands in the enumeration</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
@@ -1001,7 +1001,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="112" name="MAV_CMD_CONDITION_DELAY">
+      <entry value="112" name="MAV_CMD_CONDITION_DELAY" hasLocation="false" isDestination="false">
         <description>Delay mission state machine.</description>
         <param index="1">Delay in seconds (decimal)</param>
         <param index="2">Empty</param>
@@ -1011,7 +1011,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="113" name="MAV_CMD_CONDITION_CHANGE_ALT">
+      <entry value="113" name="MAV_CMD_CONDITION_CHANGE_ALT" hasLocation="false" isDestination="true">
         <description>Ascend/descend at rate.  Delay mission state machine until desired altitude reached.</description>
         <param index="1" label="rate" units="m/s">Descent / Ascend rate.</param>
         <param index="2">Empty</param>
@@ -1021,7 +1021,7 @@
         <param index="6">Empty</param>
         <param index="7">Finish Altitude</param>
       </entry>
-      <entry value="114" name="MAV_CMD_CONDITION_DISTANCE">
+      <entry value="114" name="MAV_CMD_CONDITION_DISTANCE" hasLocation="false" isDestination="false">
         <description>Delay mission state machine until within desired distance of next NAV point.</description>
         <param index="1" label="distance" units="m" minValue="0">Distance.</param>
         <param index="2">Empty</param>
@@ -1031,7 +1031,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="115" name="MAV_CMD_CONDITION_YAW">
+      <entry value="115" name="MAV_CMD_CONDITION_YAW" hasLocation="false" isDestination="false">
         <description>Reach a certain target angle.</description>
         <param index="1" label="angle" units="deg" minValue="0" maxValue="360">target angle, 0 is north</param>
         <param index="2" label="angular speed" units="deg/s">angular speed</param>
@@ -1041,7 +1041,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="159" name="MAV_CMD_CONDITION_LAST">
+      <entry value="159" name="MAV_CMD_CONDITION_LAST" hasLocation="false" isDestination="false">
         <description>NOP - This command is only used to mark the upper limit of the CONDITION commands in the enumeration</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
@@ -1051,7 +1051,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="176" name="MAV_CMD_DO_SET_MODE">
+      <entry value="176" name="MAV_CMD_DO_SET_MODE" hasLocation="false" isDestination="false">
         <description>Set system mode.</description>
         <param index="1" label="mode" enum="MAV_MODE">Mode</param>
         <param index="2" label="custom mode">Custom mode - this is system specific, please refer to the individual autopilot specifications for details.</param>
@@ -1061,7 +1061,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="177" name="MAV_CMD_DO_JUMP">
+      <entry value="177" name="MAV_CMD_DO_JUMP" hasLocation="false" isDestination="false">
         <description>Jump to the desired command in the mission list.  Repeat this action only the specified number of times</description>
         <param index="1" label="number" minValue="0" increment="1">Sequence number</param>
         <param index="2" label="repeat" minValue="0" increment="1">Repeat count</param>
@@ -1071,7 +1071,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="178" name="MAV_CMD_DO_CHANGE_SPEED">
+      <entry value="178" name="MAV_CMD_DO_CHANGE_SPEED" hasLocation="false" isDestination="false">
         <description>Change speed and/or throttle set points.</description>
         <param index="1" label="speed type" minValue="0" maxValue="3" increment="1">Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)</param>
         <param index="2" label="speed" units="m/s" minValue="-1">Speed (-1 indicates no change)</param>
@@ -1081,7 +1081,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="179" name="MAV_CMD_DO_SET_HOME">
+      <entry value="179" name="MAV_CMD_DO_SET_HOME" hasLocation="true" isDestination="false">
         <description>Changes the home location either to the current location or a specified location.</description>
         <param index="1" label="use current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
         <param index="2">Empty</param>
@@ -1091,7 +1091,7 @@
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="180" name="MAV_CMD_DO_SET_PARAMETER">
+      <entry value="180" name="MAV_CMD_DO_SET_PARAMETER" hasLocation="false" isDestination="false">
         <description>Set a system parameter.  Caution!  Use of this command requires knowledge of the numeric enumeration value of the parameter.</description>
         <param index="1" label="number" minValue="0" increment="1">Parameter number</param>
         <param index="2" label="value">Parameter value</param>
@@ -1101,7 +1101,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="181" name="MAV_CMD_DO_SET_RELAY">
+      <entry value="181" name="MAV_CMD_DO_SET_RELAY" hasLocation="false" isDestination="false">
         <description>Set a relay to a condition.</description>
         <param index="1" label="number" minValue="0" increment="1">Relay number</param>
         <param index="2" label="setting">Setting (1=on, 0=off, others possible depending on system hardware)</param>
@@ -1111,7 +1111,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="182" name="MAV_CMD_DO_REPEAT_RELAY">
+      <entry value="182" name="MAV_CMD_DO_REPEAT_RELAY" hasLocation="false" isDestination="false">
         <description>Cycle a relay on and off for a desired number of cycles with a desired period.</description>
         <param index="1" label="number" minValue="0" increment="1">Relay number.</param>
         <param index="2" label="count" minValue="1" increment="1">Cycle count.</param>
@@ -1121,7 +1121,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="183" name="MAV_CMD_DO_SET_SERVO">
+      <entry value="183" name="MAV_CMD_DO_SET_SERVO" hasLocation="false" isDestination="false">
         <description>Set a servo to a desired PWM value.</description>
         <param index="1" label="number" minValue="0" increment="1">Servo number</param>
         <param index="2" label="PWM" units="us" minValue="1000" maxValue="2000" increment="1">PWM</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -757,10 +757,10 @@
...
       <entry value="16" name="MAV_CMD_NAV_WAYPOINT">
         <description>Navigate to waypoint.</description>
-        <param index="1">Hold time in decimal seconds. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
-        <param index="2">Acceptance radius in meters (if the sphere with this radius is hit, the waypoint counts as reached)</param>
-        <param index="3">0 to pass through the WP, if &gt; 0 radius in meters to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
-        <param index="4">Desired yaw angle at waypoint (rotary wing). NaN for unchanged.</param>
+        <param index="1" label="hold" units="s" minValue="0">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
```
hamishwillee: @DonLakeFlyer Not the label here is "hold". Do you have any letter-case requirements on these labels for the GUI - ie should this be "Hold" or "Hold time"?

message_definitions/v1.0/common.xml
```diff
@@ -769,27 +769,27 @@
...
+        <param index="3" label="radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
+        <param index="4" label="yaw" units="deg">Desired yaw angle.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
       <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS">
         <description>Loiter around this waypoint for X turns</description>
-        <param index="1">Turns</param>
+        <param index="1" label="turns">Turns</param>
```
hamishwillee: Does it need a minValue?

message_definitions/v1.0/common.xml
```diff
@@ -817,83 +817,83 @@
...
       <entry value="23" name="MAV_CMD_NAV_LAND_LOCAL">
         <description>Land at local position (local frame only)</description>
-        <param index="1">Landing target number (if available)</param>
-        <param index="2">Maximum accepted offset from desired landing position [m] - computed magnitude from spherical coordinates: d = sqrt(x^2 + y^2 + z^2), which gives the maximum accepted distance between the desired landing position and the position where the vehicle is about to land</param>
-        <param index="3">Landing descend rate [ms^-1]</param>
-        <param index="4">Desired yaw angle [rad]</param>
-        <param index="5">Y-axis position [m]</param>
-        <param index="6">X-axis position [m]</param>
-        <param index="7">Z-axis / ground level position [m]</param>
+        <param index="1" label="target" minValue="0">Landing target number (if available)</param>
```
hamishwillee: Are landing targets numbered from 0? (I don't know)

hamishwillee: Should label be "Landing Target #"

message_definitions/v1.0/common.xml
```diff
@@ -817,83 +817,83 @@
...
+        <param index="4" label="yaw" units="rad">Desired yaw angle.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
       <entry value="30" name="MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT">
         <description>Continue on the current course and climb/descend to specified altitude.  When the altitude is reached continue to the next command (i.e., don't proceed to the next command until the desired altitude is reached.</description>
-        <param index="1">Climb or Descend (0 = Neutral, command completes when within 5m of this command's altitude, 1 = Climbing, command completes when at or above this command's altitude, 2 = Descending, command completes when at or below this command's altitude.</param>
+        <param index="1" label="action">Climb or Descend (0 = Neutral, command completes when within 5m of this command's altitude, 1 = Climbing, command completes when at or above this command's altitude, 2 = Descending, command completes when at or below this command's altitude.</param>
```
hamishwillee: We could consider an enum for this.

message_definitions/v1.0/common.xml
```diff
@@ -817,83 +817,83 @@
...
-        <param index="4">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location</param>
+        <param index="4" label="xtrack location" minValue="0" maxValue="1">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
       <entry value="32" name="MAV_CMD_DO_FOLLOW">
         <description>Being following a target</description>
-        <param index="1">System ID (the system ID of the FOLLOW_TARGET beacon). Send 0 to disable follow-me and return to the default position hold mode</param>
+        <param index="1" label="target">System ID (the system ID of the FOLLOW_TARGET beacon). Send 0 to disable follow-me and return to the default position hold mode</param>
```
hamishwillee: Do you think we should set min, max and increment for this?

message_definitions/v1.0/common.xml
```diff
@@ -817,83 +817,83 @@
...
         <param index="2">RESERVED</param>
         <param index="3">RESERVED</param>
-        <param index="4">altitude flag: 0: Keep current altitude, 1: keep altitude difference to target, 2: go to a fixed altitude above home</param>
-        <param index="5">altitude</param>
+        <param index="4" label="mode">altitude flag: 0: Keep current altitude, 1: keep altitude difference to target, 2: go to a fixed altitude above home</param>
```
hamishwillee: Shouldn't this be an enum. Not sure "mode" is right either - it is "Altitude mode" or "Altitude Flag"

message_definitions/v1.0/common.xml
```diff
@@ -817,83 +817,83 @@
...
+        <param index="5" label="altitude" units="m">altitude</param>
         <param index="6">RESERVED</param>
-        <param index="7">TTL in seconds in which the MAV should go to the default position hold mode after a message rx timeout</param>
+        <param index="7" label="TTL" units="s">TTL in seconds in which the MAV should go to the default position hold mode after a message rx timeout</param>
```
hamishwillee: Should description be: Time To Land (TTL) in which .... 
- ie expand out TTL and remove seconds from description?
- Also message RX, not "rx"

hamishwillee:
Hi @amilcarlucas 

Thanks very much for this - good subset to start with. I put some comments, but the main thing we need is guidance from @DonLakeFlyer on the capitalisation for param labels. Ie Hold or hold, Big name, or "Big Name"

message_definitions/v1.0/common.xml
amilcarlucas: Fixed

message_definitions/v1.0/common.xml
amilcarlucas: I'll wait for feedback on this one.

message_definitions/v1.0/common.xml
amilcarlucas: The message is about landing so I omited that part

message_definitions/v1.0/common.xml
amilcarlucas: There a quite a few of these. So that is a subject for another PR

message_definitions/v1.0/common.xml
amilcarlucas: Fixed

message_definitions/v1.0/common.xml
amilcarlucas: again, many of these should be enums but are not. But that is for other PR to solve.

message_definitions/v1.0/common.xml
amilcarlucas: Fixed

amilcarlucas:
@hamishwillee thanks for the review. I now did some fixes and completed all MAV_CMDs

amilcarlucas:
@LorenzMeier please review this one. It brings many changes, and it will collect a lot of merge conflicts if not merged soon.

message_definitions/v1.0/common.xml
hamishwillee: Fair enough!

message_definitions/v1.0/common.xml
hamishwillee: I'll ask @DonLakeFlyer directly and get back to you. If no reply in timely manner I am happy to merge this in a day or so.

DonLakeFlyer: QGC capatilization for UI is caps on every word.

message_definitions/v1.0/common.xml
```diff
@@ -755,47 +755,47 @@
...
-      <entry value="16" name="MAV_CMD_NAV_WAYPOINT">
+      <entry value="16" name="MAV_CMD_NAV_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint.</description>
-        <param index="1">Hold time in decimal seconds. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
-        <param index="2">Acceptance radius in meters (if the sphere with this radius is hit, the waypoint counts as reached)</param>
-        <param index="3">0 to pass through the WP, if &gt; 0 radius in meters to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
-        <param index="4">Desired yaw angle at waypoint (rotary wing). NaN for unchanged.</param>
+        <param index="1" label="hold" units="s" minValue="0" increment="1">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
```
hamishwillee: As per https://github.com/mavlink/mavlink/pull/1100#discussion_r269201027, label="Hold"

message_definitions/v1.0/common.xml
```diff
@@ -755,47 +755,47 @@
...
+        <param index="2" label="accept radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
```
hamishwillee: As per https://github.com/mavlink/mavlink/pull/1100#discussion_r269201027, label="Accept Radius"

@amilcarlucas No point me adding notes to all - can you change globally?

message_definitions/v1.0/common.xml
hamishwillee: FYI, for all of these update label to be caplitalised first letter on all words.

hamishwillee:
Also there are now conflicts to resolve :-(

message_definitions/v1.0/common.xml
amilcarlucas: Done

message_definitions/v1.0/common.xml
amilcarlucas: done

message_definitions/v1.0/common.xml
amilcarlucas: done

amilcarlucas:
Rebased for the first time.

message_definitions/v1.0/common.xml
```diff
@@ -754,47 +754,47 @@
     </enum>
     <enum name="MAV_CMD">
       <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data.</description>
-      <entry value="16" name="MAV_CMD_NAV_WAYPOINT">
+      <entry value="16" name="MAV_CMD_NAV_WAYPOINT" hasLocation="true" isDestination="true">
```
hamishwillee: hasLocation="true" isDestination="true" are the defaults, and in addition if isDestination="true"  then hasLocation="true" is implied.

Upshot you never need to specify these cases.

What we do need to specify is the "not cases". 

I think all these should be removed.

hamishwillee: PS Arguably we should allow param 5, 6, 7 to be not declared and default to lat, lon, altitude in generated code. 

I say arguably, because I don't feel that isDestination and hasLocation are particularly obvious, and this would break generators that we don't have control over. I'm OK with that for reserved values, but not particularly for these sorts of params

message_definitions/v1.0/common.xml
```diff
@@ -754,47 +754,47 @@
...
+        <param index="4" label="Yaw">Desired yaw angle.</param>
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS">
+      <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint for X turns</description>
-        <param index="1">Turns</param>
+        <param index="1" label="Turns" minValue="0" increment="1">Turns</param>
```
hamishwillee: Should remove increment as this is more restrictive than current text - ie we don't know that someone hasn't implemented this allowing partial turns. 

We can only add increment here if partial turns are nonsensical.

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
       </entry>
-      <entry value="25" name="MAV_CMD_NAV_FOLLOW">
+      <entry value="25" name="MAV_CMD_NAV_FOLLOW" hasLocation="true" isDestination="false">
         <description>Vehicle following, i.e. this waypoint represents the position of a moving vehicle</description>
-        <param index="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
-        <param index="2">Ground speed of vehicle to be followed</param>
-        <param index="3">Radius around waypoint, in meters. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4">Desired yaw angle.</param>
+        <param index="1" label="Following" increment="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
+        <param index="2" label="Ground Speed">Ground speed of vehicle to be followed</param>
```
hamishwillee: units?

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
         <param index="5">Latitude</param>
         <param index="6">Longitude</param>
         <param index="7">Altitude</param>
       </entry>
-      <entry value="30" name="MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT">
+      <entry value="30" name="MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT" hasLocation="false" isDestination="true">
         <description>Continue on the current course and climb/descend to specified altitude.  When the altitude is reached continue to the next command (i.e., don't proceed to the next command until the desired altitude is reached.</description>
-        <param index="1">Climb or Descend (0 = Neutral, command completes when within 5m of this command's altitude, 1 = Climbing, command completes when at or above this command's altitude, 2 = Descending, command completes when at or below this command's altitude.</param>
+        <param index="1" label="Action" minValue="0" maxValue="2" increment="1">Climb or Descend (0 = Neutral, command completes when within 5m of this command's altitude, 1 = Climbing, command completes when at or above this command's altitude, 2 = Descending, command completes when at or below this command's altitude.</param>
```
hamishwillee: Later PR - as enum

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
-        <param index="4">altitude flag: 0: Keep current altitude, 1: keep altitude difference to target, 2: go to a fixed altitude above home</param>
-        <param index="5">altitude</param>
+        <param index="4" label="Altitude Mode" minValue="0" maxValue="2" increment="1">Altitude mode: 0: Keep current altitude, 1: keep altitude difference to target, 2: go to a fixed altitude above home.</param>
+        <param index="5" label="Altitude">Altitude above home. (used if mode=2)</param>
         <param index="6">RESERVED</param>
-        <param index="7">TTL in seconds in which the MAV should go to the default position hold mode after a message rx timeout</param>
+        <param index="7" label="Time to Land" units="s" minValue="0">Time to land in which the MAV should go to the default position hold mode after a message RX timeout.</param>
       </entry>
-      <entry value="33" name="MAV_CMD_DO_FOLLOW_REPOSITION">
+      <entry value="33" name="MAV_CMD_DO_FOLLOW_REPOSITION" hasLocation="false" isDestination="false">
```
hamishwillee: So you could probably set hasLocation="False" here and that would cover both cases. Up to you.

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
+        <param index="4" label="Yaw" units="deg" minValue="0" maxValue="360">Yaw angle at goal</param>
         <param index="5">Latitude/X of goal</param>
         <param index="6">Longitude/Y of goal</param>
         <param index="7">Altitude/Z of goal</param>
       </entry>
-      <entry value="82" name="MAV_CMD_NAV_SPLINE_WAYPOINT">
+      <entry value="82" name="MAV_CMD_NAV_SPLINE_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint using a spline path.</description>
-        <param index="1">Hold time in decimal seconds. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
+        <param index="1" label="Hold" units="s" minValue="0" increment="1">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
```
hamishwillee: remove increment (and generally, wherever things are decimals)

message_definitions/v1.0/common.xml
```diff
@@ -1051,117 +1051,117 @@
...
       </entry>
-      <entry value="184" name="MAV_CMD_DO_REPEAT_SERVO">
+      <entry value="184" name="MAV_CMD_DO_REPEAT_SERVO" hasLocation="false" isDestination="false">
         <description>Cycle a between its nominal setting and a desired PWM for a desired number of cycles with a desired period.</description>
-        <param index="1">Servo number</param>
-        <param index="2">PWM (microseconds, 1000 to 2000 typical)</param>
-        <param index="3">Cycle count</param>
-        <param index="4">Cycle time (seconds)</param>
+        <param index="1" label="Number" minValue="0" increment="1">Servo number</param>
+        <param index="2" label="PWM" units="us" minValue="1000" maxValue="2000" increment="1">PWM</param>
```
hamishwillee: I don't think we can specify the minValue or maxValue here, as those are typical, not hard limits. minValue of 0 could be used.

message_definitions/v1.0/common.xml
```diff
@@ -1051,117 +1051,117 @@
...
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="186" name="MAV_CMD_DO_CHANGE_ALTITUDE">
+      <entry value="186" name="MAV_CMD_DO_CHANGE_ALTITUDE" hasLocation="false" isDestination="false">
         <description>Change altitude set point.</description>
-        <param index="1">Altitude in meters</param>
-        <param index="2">Mav frame of new altitude (see MAV_FRAME)</param>
+        <param index="1" label="Altitude" units="m">Altitude</param>
+        <param index="2" label="Frame" enum="MAV_FRAME">Mav frame of new altitude</param>
```
hamishwillee: Probably change description to:

>  Frame of new altitude

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
-      <entry value="200" name="MAV_CMD_DO_CONTROL_VIDEO">
+      <entry value="200" name="MAV_CMD_DO_CONTROL_VIDEO" hasLocation="false" isDestination="false">
         <description>Control onboard camera system.</description>
-        <param index="1">Camera ID (-1 for all)</param>
-        <param index="2">Transmission: 0: disabled, 1: enabled compressed, 2: enabled raw</param>
-        <param index="3">Transmission mode: 0: video stream, &gt;0: single images every n seconds (decimal)</param>
-        <param index="4">Recording: 0: disabled, 1: enabled compressed, 2: enabled raw</param>
+        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1 for all)</param>
+        <param index="2" label="Transmission" minValue="0" maxValue="2" increment="1">Transmission: 0: disabled, 1: enabled compressed, 2: enabled raw</param>
+        <param index="3" label="Interval" units="s" minValue="0" increment="1">Transmission mode: 0: video stream, &gt;0: single images every n seconds</param>
```
hamishwillee: Units on this are not seconds.

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
       </entry>
-      <entry value="201" name="MAV_CMD_DO_SET_ROI">
+      <entry value="201" name="MAV_CMD_DO_SET_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
         <description>Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicles control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
-        <param index="1">Region of interest mode. (see MAV_ROI enum)</param>
-        <param index="2">Waypoint index/ target ID. (see MAV_ROI enum)</param>
-        <param index="3">ROI index (allows a vehicle to manage multiple ROI's)</param>
+        <param index="1" label="ROI Mode" enum="MAV_ROI">Region of interest mode.</param>
+        <param index="2" label="WP Index" minValue="0" increment="1">Waypoint index/ target ID. (see MAV_ROI enum)</param>
```
hamishwillee: is "(see MAV_ROI enum)" relevant in this param or typo?

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
-        <param index="1">Modes: P, TV, AV, M, Etc</param>
-        <param index="2">Shutter speed: Divisor number for one second</param>
-        <param index="3">Aperture: F stop number</param>
-        <param index="4">ISO number e.g. 80, 100, 200, Etc</param>
-        <param index="5">Exposure type enumerator</param>
-        <param index="6">Command Identity</param>
-        <param index="7">Main engine cut-off time before camera trigger in seconds/10 (0 means no cut-off)</param>
-      </entry>
-      <entry value="203" name="MAV_CMD_DO_DIGICAM_CONTROL">
+        <param index="1" label="Mode" minValue="0" increment="1">Modes: P, TV, AV, M, Etc</param>
```
hamishwillee: Do we know how the modes map to numbers? I assume this is correct though to increment and with no max

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
+        <param index="2" label="Shutter Speed" increment="1">Shutter speed: Divisor number for one second</param>
+        <param index="3" label="Aperture">Aperture: F stop number</param>
+        <param index="4" label="ISO">ISO number e.g. 80, 100, 200, Etc</param>
```
hamishwillee: I think ISO are always positive integers.

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
-        <param index="3">Aperture: F stop number</param>
-        <param index="4">ISO number e.g. 80, 100, 200, Etc</param>
-        <param index="5">Exposure type enumerator</param>
-        <param index="6">Command Identity</param>
-        <param index="7">Main engine cut-off time before camera trigger in seconds/10 (0 means no cut-off)</param>
-      </entry>
-      <entry value="203" name="MAV_CMD_DO_DIGICAM_CONTROL">
+        <param index="1" label="Mode" minValue="0" increment="1">Modes: P, TV, AV, M, Etc</param>
+        <param index="2" label="Shutter Speed" increment="1">Shutter speed: Divisor number for one second</param>
+        <param index="3" label="Aperture">Aperture: F stop number</param>
```
hamishwillee: Fstops are always positive decimals (floats)

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
-        <param index="2">Shutter speed: Divisor number for one second</param>
-        <param index="3">Aperture: F stop number</param>
-        <param index="4">ISO number e.g. 80, 100, 200, Etc</param>
-        <param index="5">Exposure type enumerator</param>
-        <param index="6">Command Identity</param>
-        <param index="7">Main engine cut-off time before camera trigger in seconds/10 (0 means no cut-off)</param>
-      </entry>
-      <entry value="203" name="MAV_CMD_DO_DIGICAM_CONTROL">
+        <param index="1" label="Mode" minValue="0" increment="1">Modes: P, TV, AV, M, Etc</param>
+        <param index="2" label="Shutter Speed" increment="1">Shutter speed: Divisor number for one second</param>
```
hamishwillee: Shutter speed is always positive

message_definitions/v1.0/common.xml
```diff
@@ -1251,90 +1251,91 @@
...
+        <param index="3" label="Aperture">Aperture: F stop number</param>
+        <param index="4" label="ISO">ISO number e.g. 80, 100, 200, Etc</param>
+        <param index="5" label="Exposure">Exposure type enumerator</param>
```
hamishwillee: FYI Not sure of valid values, but I know both positive and negative decimals allowed.

message_definitions/v1.0/common.xml
```diff
@@ -1344,19 +1345,19 @@
...
-      <entry value="209" name="MAV_CMD_DO_MOTOR_TEST">
+      <entry value="209" name="MAV_CMD_DO_MOTOR_TEST" hasLocation="false" isDestination="false">
         <description>Mission command to perform motor test</description>
-        <param index="1">motor number (a number from 1 to max number of motors on the vehicle)</param>
-        <param index="2">throttle type (0=throttle percentage, 1=PWM, 2=pilot throttle channel pass-through. See MOTOR_TEST_THROTTLE_TYPE enum)</param>
-        <param index="3">throttle</param>
-        <param index="4">timeout (in seconds)</param>
-        <param index="5">motor count (number of motors to test to test in sequence, waiting for the timeout above between them; 0=1 motor, 1=1 motor, 2=2 motors...)</param>
-        <param index="6">motor test order (See MOTOR_TEST_ORDER enum)</param>
+        <param index="1" label="Number" minValue="1" increment="1">Motor number. (a number from 1 to max number of motors on the vehicle)</param>
```
hamishwillee: Generally you should put the full stop right at the end (i.e. the clarification is part of the sentence).

message_definitions/v1.0/common.xml
```diff
@@ -1365,70 +1366,70 @@
...
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="221" name="MAV_CMD_DO_GUIDED_MASTER">
+      <entry value="221" name="MAV_CMD_DO_GUIDED_MASTER" hasLocation="false" isDestination="false">
         <description>set id of master controller</description>
-        <param index="1">System ID</param>
-        <param index="2">Component ID</param>
+        <param index="1" label="System ID" minValue="0" increment="1">System ID</param>
```
hamishwillee: If this is a MAVLink system ID, we know the max value is 255

message_definitions/v1.0/common.xml
```diff
@@ -1365,70 +1366,70 @@
...
+        <param index="2" label="Component ID" minValue="0" increment="1">Component ID</param>
```
hamishwillee: Assuming mavlink component ID, max value is 255

message_definitions/v1.0/common.xml
```diff
@@ -1365,70 +1366,70 @@
...
-        <param index="1">0: Stop engine, 1:Start Engine</param>
-        <param index="2">0: Warm start, 1:Cold start. Controls use of choke where applicable</param>
-        <param index="3">Height delay (meters). This is for commanding engine start only after the vehicle has gained the specified height. Used in VTOL vehicles during takeoff to start engine after the aircraft is off the ground. Zero for no delay.</param>
+        <param index="1" label="Start Engine" minValue="0" maxValue="1" increment="1">0: Stop engine, 1:Start Engine</param>
+        <param index="2" label="Cold Start" minValue="0" maxValue="1" increment="1">0: Warm start, 1:Cold start. Controls use of choke where applicable</param>
+        <param index="3" label="Height Delay" units="m">Height delay. This is for commanding engine start only after the vehicle has gained the specified height. Used in VTOL vehicles during takeoff to start engine after the aircraft is off the ground. Zero for no delay.</param>
```
hamishwillee: minValue = 0

message_definitions/v1.0/common.xml
```diff
@@ -1365,70 +1366,70 @@
...
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="223" name="MAV_CMD_DO_ENGINE_CONTROL">
+      <entry value="223" name="MAV_CMD_DO_ENGINE_CONTROL" hasLocation="false" isDestination="false">
         <description>Control vehicle engine. This is interpreted by the vehicles engine controller to change the target engine state. It is intended for vehicles with internal combustion engines</description>
-        <param index="1">0: Stop engine, 1:Start Engine</param>
-        <param index="2">0: Warm start, 1:Cold start. Controls use of choke where applicable</param>
-        <param index="3">Height delay (meters). This is for commanding engine start only after the vehicle has gained the specified height. Used in VTOL vehicles during takeoff to start engine after the aircraft is off the ground. Zero for no delay.</param>
+        <param index="1" label="Start Engine" minValue="0" maxValue="1" increment="1">0: Stop engine, 1:Start Engine</param>
```
hamishwillee: Label - maybe "Stop/Start"? Not sure.

message_definitions/v1.0/common.xml
```diff
@@ -1365,70 +1366,70 @@
...
+        <param index="2" label="Cold Start" minValue="0" maxValue="1" increment="1">0: Warm start, 1:Cold start. Controls use of choke where applicable</param>
```
hamishwillee: Maybe "Cold/Warm Start". Not sure.

message_definitions/v1.0/common.xml
```diff
@@ -1447,76 +1448,76 @@
...
-        <param index="1">1: gyro calibration, 3: gyro temperature calibration</param>
-        <param index="2">1: magnetometer calibration</param>
-        <param index="3">1: ground pressure calibration</param>
-        <param index="4">1: radio RC calibration, 2: RC trim calibration</param>
-        <param index="5">1: accelerometer calibration, 2: board level calibration, 3: accelerometer temperature calibration, 4: simple accelerometer calibration</param>
-        <param index="6">1: APM: compass/motor interference calibration (PX4: airspeed calibration, deprecated), 2: airspeed calibration</param>
-        <param index="7">1: ESC calibration, 3: barometer temperature calibration</param>
-      </entry>
-      <entry value="242" name="MAV_CMD_PREFLIGHT_SET_SENSOR_OFFSETS">
+        <param index="1" label="Gyro Temperature" minValue="0" maxValue="3" increment="1">1: gyro calibration, 3: gyro temperature calibration</param>
```
hamishwillee: label - I think just "Gyro"

message_definitions/v1.0/common.xml
```diff
@@ -1447,76 +1448,76 @@
...
+        <param index="3" label="Y Offset">Y axis offset (or generic dimension 2), in the sensor's raw units</param>
+        <param index="4" label="Z Offset">Z axis offset (or generic dimension 3), in the sensor's raw units</param>
+        <param index="5" label="4th Dimension">Generic dimension 4, in the sensor's raw units</param>
+        <param index="6" label="5th Dimension">Generic dimension 5, in the sensor's raw units</param>
+        <param index="7" label="6th Dimension">Generic dimension 6, in the sensor's raw units</param>
+      </entry>
+      <entry value="243" name="MAV_CMD_PREFLIGHT_UAVCAN" hasLocation="false" isDestination="false">
         <description>Trigger UAVCAN config. This command will be only accepted if in pre-flight mode.</description>
-        <param index="1">1: Trigger actuator ID assignment and direction mapping.</param>
+        <param index="1" label="Actuator ID" minValue="0" increment="1">1: Trigger actuator ID assignment and direction mapping.</param>
```
hamishwillee: Not sure about any of these attributes. Suggest remove until can be clarified. It is certainly not clear that this is actually an ID or what "direction" means

message_definitions/v1.0/common.xml
```diff
@@ -1526,181 +1527,181 @@
         <param index="6">Reserved</param>
         <param index="7">Reserved</param>
       </entry>
-      <entry value="500" name="MAV_CMD_START_RX_PAIR">
+      <entry value="500" name="MAV_CMD_START_RX_PAIR" hasLocation="false" isDestination="false">
         <description>Starts receiver pairing</description>
-        <param index="1">0:Spektrum</param>
-        <param index="2">RC type (see RC_TYPE enum)</param>
+        <param index="1" label="Spektrum">0:Spektrum</param>
```
hamishwillee: This is a bit weird - ie spektrum here and RC type below. But no worse than before so if issue address as separate PR.

message_definitions/v1.0/common.xml
```diff
@@ -1526,181 +1527,181 @@
...
+        <param index="2" label="RC Type" enum="RC_TYPE">RC type</param>
       </entry>
-      <entry value="510" name="MAV_CMD_GET_MESSAGE_INTERVAL">
+      <entry value="510" name="MAV_CMD_GET_MESSAGE_INTERVAL" hasLocation="false" isDestination="false">
         <description>Request the interval between messages for a particular MAVLink message ID</description>
-        <param index="1">The MAVLink message ID</param>
+        <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
```
hamishwillee: I'm not sure we can specify maxValue, as this could be used in MAVLink 1.

message_definitions/v1.0/common.xml
```diff
@@ -1526,181 +1527,181 @@
...
       </entry>
-      <entry value="511" name="MAV_CMD_SET_MESSAGE_INTERVAL">
+      <entry value="511" name="MAV_CMD_SET_MESSAGE_INTERVAL" hasLocation="false" isDestination="false">
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM</description>
```
hamishwillee: @amilcarlucas **As an aside**, how about an attribute for an updated message/command that mirrors the `replaced_by` attribute on `deprecated` commands/messages on the replacement message/command. ie something that says that this message or command `replaces` or `supersedes` something else.
```
<entry value="511" name="MAV_CMD_SET_MESSAGE_INTERVAL" hasLocation="false" isDestination="false" supersedes="REQUEST_DATA_STREAM">
```
More generally, it might be good to have a softer approach than deprecation, since this appears contentious. Specifically, we could have a tag _just like_ `deprecated` but named `superseded`. This could use the same text as above on the other end.

message_definitions/v1.0/common.xml
hamishwillee: Actually this is probably fine, and better than having infinite.

message_definitions/v1.0/common.xml
```diff
@@ -1526,181 +1527,181 @@
...
-        <param index="1">The MAVLink message ID of the requested message.</param>
-        <param index="2">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
+        <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID of the requested message.</param>
+        <param index="2" label="Index ID" minValue="0" increment="1">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
       </entry>
-      <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION">
+      <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION" hasLocation="false" isDestination="false">
         <description>Request MAVLink protocol version compatibility</description>
-        <param index="1">1: Request supported protocol versions by all nodes on the network</param>
+        <param index="1" label="Protocol">1: Request supported protocol versions by all nodes on the network</param>
```
hamishwillee: Presumably this can have a value of 0 or 1?

message_definitions/v1.0/common.xml
```diff
@@ -1526,181 +1527,181 @@
...
         <param index="2">Reserved (all remaining params)</param>
       </entry>
-      <entry value="520" name="MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES">
+      <entry value="520" name="MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES" hasLocation="false" isDestination="false">
         <description>Request autopilot capabilities</description>
-        <param index="1">1: Request autopilot version</param>
+        <param index="1" label="Version">1: Request autopilot version</param>
```
hamishwillee: Presumably can have value of 0 or 1

message_definitions/v1.0/common.xml
```diff
@@ -1710,69 +1711,69 @@
...
+        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. Only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
       </entry>
-      <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST">
+      <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST" hasLocation="false" isDestination="false">
         <description>Request authorization to arm the vehicle to a external entity, the arm authorizer is responsible to request all data that is needs from the vehicle before authorize or deny the request. If approved the progress of command_ack message should be set with period of time that this authorization is valid in seconds or in case it was denied it should be set with one of the reasons in ARM_AUTH_DENIED_REASON.
         </description>
-        <param index="1">Vehicle system id, this way ground station can request arm authorization on behalf of any vehicle</param>
+        <param index="1" label="System ID" minValue="0" increment="1">Vehicle system id, this way ground station can request arm authorization on behalf of any vehicle</param>
```
hamishwillee: maxValue 255

message_definitions/v1.0/common.xml
```diff
@@ -1710,69 +1711,69 @@
...
+      <entry value="4000" name="MAV_CMD_SET_GUIDED_SUBMODE_STANDARD" hasLocation="false" isDestination="false">
         <description>This command sets the submode to standard guided when vehicle is in guided mode. The vehicle holds position and altitude and the user can input the desired velocities along all three axes.
                   </description>
       </entry>
-      <entry value="4001" name="MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE">
+      <entry value="4001" name="MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE" hasLocation="true" isDestination="false">
         <description>This command sets submode circle when vehicle is in guided mode. Vehicle flies along a circle facing the center of the circle. The user can input the velocity along the circle and change the radius. If no input is given the vehicle will hold position.
                   </description>
-        <param index="1">Radius of desired circle in CIRCLE_MODE</param>
+        <param index="1" label="Radius">Radius of desired circle in CIRCLE_MODE</param>
```
hamishwillee: units?

message_definitions/v1.0/common.xml
```diff
@@ -1853,19 +1854,19 @@
...
-        <param index="2">Desired approach vector in degrees compass heading (0..360). A negative value indicates the system can define the approach vector at will.</param>
-        <param index="3">Desired ground speed at release time. This can be overridden by the airframe in case it needs to meet minimum airspeed. A negative value indicates the system can define the ground speed at will.</param>
-        <param index="4">Minimum altitude clearance to the release position in meters. A negative value indicates the system can define the clearance at will.</param>
+        <param index="1" label="Operation Mode" minValue="0" maxValue="2" increment="1">Operation mode. 0: prepare single payload deploy (overwriting previous requests), but do not execute it. 1: execute payload deploy immediately (rejecting further deploy commands during execution, but allowing abort). 2: add payload deploy to existing deployment list.</param>
+        <param index="2" label="Approach Vector" units="deg" minValue="-1" maxValue="360">Desired approach vector in compass heading. A negative value indicates the system can define the approach vector at will.</param>
```
hamishwillee: Shouldn't minvalue be 0?

hamishwillee:
@amilcarlucas OK, I have been all the way through. Lots of minor comments/questions. Big one is that we don't need the location/destination info except on things that are NOT locations.
Also, just FYI, there are many commands that are never expected to be seen in the GCS GUI - e.g. many of the camera messages. It does no harm to add the hints, but they won't be used. 

This was a big effort to review, so I assume a massive effort to do in the first place. Thank you!

I will review again after your next set of changes. I don't think @LorenzMeier will have time to review.

message_definitions/v1.0/common.xml
amilcarlucas: fixed

message_definitions/v1.0/common.xml
amilcarlucas: That will be for a following PR

message_definitions/v1.0/common.xml
amilcarlucas: I prefer explicit labels. It makes it clearer.

message_definitions/v1.0/common.xml
amilcarlucas: The description states single images every n seconds. I interpret that as seconds

message_definitions/v1.0/common.xml
amilcarlucas: I assume fractional modes are nonsensical. I expect this should map to a custom enum, hence integer.

message_definitions/v1.0/common.xml
amilcarlucas: I went for the "active" state name, it allows the GCS to implement this as a boolean

message_definitions/v1.0/common.xml
amilcarlucas: I went for the "active" state name, it allows the GCS to implement this as a boolean

message_definitions/v1.0/common.xml
amilcarlucas: I assume this to be a bitmask:
bit 0 for gyro
bit 1 for gyro temperature

@LorenzMeier is that correct

message_definitions/v1.0/common.xml
amilcarlucas: OK

message_definitions/v1.0/common.xml
amilcarlucas: That seams redundant to me, and will lead to duplication.

message_definitions/v1.0/common.xml
amilcarlucas: It is not clear to me. Should be solved in a future PR

message_definitions/v1.0/common.xml
amilcarlucas: Again, it is not clear for me. But it feels extremely redundant to have to set this parameter to 1 to make it work.
If you do not want the version, just do not send the command.

@LorenzMeier Why is this so poorly documented?

message_definitions/v1.0/common.xml
amilcarlucas: Not sure, deferring to future PR

amilcarlucas:
I feel that we should always have the hasLocation and isDestination attributes. It makes things explicit.
The MAVLink protocol is very implicit, I do not like that, let''s too much room for false interpretations.

This entire patch is about reducing the amount of implicit information.

amilcarlucas:
Rebased for a second time.

hamishwillee:
@amilcarlucas 
Thanks for all your hard work on this! 

> I feel that we should always have the hasLocation and isDestination attributes. It makes things explicit.
> The MAVLink protocol is very implicit, I do not like that, let''s too much room for false interpretations.
> This entire patch is about reducing the amount of implicit information.

I kind of like specifying less - so for example I would like to remove all the empty/reserved params rather than explicitly declaring them. I would be OK adding a more transparent way of indicating what messages a command can be sent in. 

Anyway, while I see your point, I think we need to have broad agreement before making this change. So what you've done now is a better intermediate step.

amilcarlucas:
I agree with not having to specify empty parameters.

in common.xml:
hasLocation="false" 97 matches
hasLocation="true" 42 matches

isDestination="false" 112 matches
isDestination="true" 27 matches

So the defaults should be changes to false, regardless of implicit or explicit

hamishwillee:
>So the defaults should be changes to false, regardless of implicit or explicit

The problem here is that mission commands came first, with locations and a lot more of them than not-locations. Then command protocol came without locations. So the default assumption is and always has been that they are locations by default. 

So while I agree it makes sense now. We'd have to make sure they exist and co-ordinate assumption change over with GCS. I suspect they would argue against.

amilcarlucas:
Rebased for the third time

hamishwillee:
Hi @amilcarlucas,
We didn't get to the topic of being terse vs fully descriptive in the meeting, but I'm OK with it as it is. 
However I'm on holiday now and I think someone else should also review this and merge.

@julianoes @auturgy Can you guys sanity check this and merge if you are OK? Would be good to avoid Amilcar having to update again.

Cheers
Hamish

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
+        <param index="3" label="Descend Rate" units="m/s">Landing descend rate</param>
+        <param index="4" label="Yaw" units="rad">Desired yaw angle</param>
+        <param index="5" label="Y Position" units="m">Y-axis position</param>
+        <param index="6" label="X Position" units="m">X-axis position</param>
+        <param index="7" label="Z Position" units="m">Z-axis / ground level position</param>
+      </entry>
+      <entry value="24" name="MAV_CMD_NAV_TAKEOFF_LOCAL" hasLocation="true" isDestination="true">
         <description>Takeoff from local position (local frame only)</description>
-        <param index="1">Minimum pitch (if airspeed sensor present), desired pitch without sensor [rad]</param>
+        <param index="1" label="Pitch" units="rad">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
```
julianoes: It's really odd that we're sometimes using `rad` and sometimes `degrees`.

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
+      <entry value="25" name="MAV_CMD_NAV_FOLLOW" hasLocation="true" isDestination="false">
         <description>Vehicle following, i.e. this waypoint represents the position of a moving vehicle</description>
-        <param index="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
-        <param index="2">Ground speed of vehicle to be followed</param>
-        <param index="3">Radius around waypoint, in meters. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4">Desired yaw angle.</param>
+        <param index="1" label="Following" increment="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
+        <param index="2" label="Ground Speed">Ground speed of vehicle to be followed</param>
+        <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
+        <param index="4" label="Yaw">Desired yaw angle.</param>
```
julianoes: Unit?

message_definitions/v1.0/common.xml
```diff
@@ -804,140 +804,140 @@
...
-        <param index="1">Camera q1 (where 0 is on the ray from the camera to the tracking device)</param>
-        <param index="2">Camera q2</param>
-        <param index="3">Camera q3</param>
-        <param index="4">Camera q4</param>
-        <param index="5">altitude offset from target (m)</param>
-        <param index="6">X offset from target (m)</param>
-        <param index="7">Y offset from target (m)</param>
-      </entry>
-      <entry value="34" name="MAV_CMD_DO_ORBIT">
+        <param index="1" label="Camera Q1">Camera q1 (where 0 is on the ray from the camera to the tracking device)</param>
```
julianoes: I would probably use `q1` like in the description. I don't see why it becomes capital Q.

message_definitions/v1.0/common.xml
```diff
@@ -1365,70 +1366,70 @@
...
         <param index="7">Empty</param>
       </entry>
-      <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT">
+      <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
         <description>Mission command to control a camera or antenna mount, using a quaternion as reference.</description>
-        <param index="1">q1 - quaternion param #1, w (1 in null-rotation)</param>
-        <param index="2">q2 - quaternion param #2, x (0 in null-rotation)</param>
-        <param index="3">q3 - quaternion param #3, y (0 in null-rotation)</param>
-        <param index="4">q4 - quaternion param #4, z (0 in null-rotation)</param>
+        <param index="1" label="Q1">quaternion param q1, w (1 in null-rotation)</param>
```
julianoes: Same here. Why capital?

message_definitions/v1.0/common.xml
amilcarlucas: yes it is odd. But this PR just makes explicit a lot of stuff that has been implicit.

message_definitions/v1.0/common.xml
amilcarlucas: I left that out for a future PR, because it is not clear if it is deg, cdeg or rad.

message_definitions/v1.0/common.xml
amilcarlucas: That was a @DonLakeFlyer request. It is rule on the QGC UI.

message_definitions/v1.0/common.xml
amilcarlucas: Again, same QGC UI guidelines.

message_definitions/v1.0/common.xml
julianoes: Right makes sense!

message_definitions/v1.0/common.xml
julianoes: Ok!

message_definitions/v1.0/common.xml
julianoes: Alright.

PR #1102, opened 2019-03-25
Proper deprecate MAV_ACTION

amilcarlucas:


message_definitions/v1.0/common.xml
```diff
@@ -232,8 +232,8 @@
         <description>Mission failure.</description>
       </entry>
     </enum>
-    <!-- WARNING: MAV_ACTION Enum is no longer supported - has been removed. Please use MAV_CMD -->
```
hamishwillee: Yes, this can be deleted.

message_definitions/v1.0/common.xml
```diff
@@ -232,8 +232,8 @@
...
     <enum name="MAV_MODE_FLAG">
+      <deprecated since="2012-02" replaced_by="MAV_CMD_DO_SET_MODE">No longer supported - has been removed.</deprecated>
```
hamishwillee: I think this deprecation is probably wrong. It nothing to do with MAV_ACTION.
MAV_CMD_DO_SET_MODE replaced SET_MODE, but a message can't replace a flag "MAV_MODE_FLAG". This flag is also used in a number of messages, which are not deprecated.

message_definitions/v1.0/common.xml
amilcarlucas: Looks like only this leftover needs to be removed

hamishwillee:
Thanks @amilcarlucas ! Merged :-)

PR #1105, opened 2019-03-25
Add missing parameter attributes to MAV_CMD commands (reach 100% documentation)

amilcarlucas:
I am unsure of these and did not look at the implementation to validate it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1314,7 +1314,7 @@
       </entry>
       <entry value="22" name="MAV_CMD_NAV_TAKEOFF" hasLocation="true" isDestination="true">
         <description>Takeoff from ground / hand</description>
-        <param index="1" label="Pitch">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
+        <param index="1" label="Pitch" units="deg">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4" label="Yaw">Yaw angle (if magnetometer present), ignored without magnetometer. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
@@ -1347,10 +1347,10 @@
         <param index="1" label="Following" increment="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
         <param index="2" label="Ground Speed">Ground speed of vehicle to be followed</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4" label="Yaw">Desired yaw angle.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="4" label="Yaw" units="rad">Desired yaw angle.</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="30" name="MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT" hasLocation="false" isDestination="true">
         <description>Continue on the current course and climb/descend to specified altitude.  When the altitude is reached continue to the next command (i.e., don't proceed to the next command until the desired altitude is reached.</description>
@@ -1420,7 +1420,7 @@
         <param index="1" label="Local Ctrl" minValue="0" maxValue="2" increment="1">0: Disable local obstacle avoidance / local path planning (without resetting map), 1: Enable local path planning, 2: Enable and reset local path planning</param>
         <param index="2" label="Global Ctrl" minValue="0" maxValue="3" increment="1">0: Disable full path planning (without resetting map), 1: Enable, 2: Enable and reset map/occupancy grid, 3: Enable and reset planned route, but not occupancy grid</param>
         <param index="3">Empty</param>
-        <param index="4" label="Yaw" units="deg" minValue="0" maxValue="360">Yaw angle at goal</param>
+        <param index="4" label="Yaw" units="deg">Yaw angle at goal</param>
         <param index="5">Latitude/X of goal</param>
         <param index="6">Longitude/Y of goal</param>
         <param index="7">Altitude/Z of goal</param>
@@ -1510,7 +1510,7 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="113" name="MAV_CMD_CONDITION_CHANGE_ALT" hasLocation="false" isDestination="true">
-        <description>Ascend/descend at rate.  Delay mission state machine until desired altitude reached.</description>
+        <description>Ascend/descend to target altitude at specified rate. Delay mission state machine until desired altitude reached.</description>
         <param index="1" label="Rate" units="m/s">Descent / Ascend rate.</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
@@ -1531,7 +1531,7 @@
       </entry>
       <entry value="115" name="MAV_CMD_CONDITION_YAW" hasLocation="false" isDestination="false">
         <description>Reach a certain target angle.</description>
-        <param index="1" label="Angle" units="deg" minValue="0" maxValue="360">target angle, 0 is north</param>
+        <param index="1" label="Angle" units="deg">target angle, 0 is north</param>
         <param index="2" label="Angular Speed" units="deg/s">angular speed</param>
         <param index="3" label="Direction" minValue="-1" maxValue="1" increment="2">direction: -1: counter clockwise, 1: clockwise</param>
         <param index="4" label="Relative" minValue="0" maxValue="1" increment="1">0: absolute angle, 1: relative offset</param>
@@ -1694,10 +1694,10 @@
         <param index="1" label="Speed" units="m/s" minValue="-1">Ground speed, less than 0 (-1) for default</param>
         <param index="2" label="Bitmask" enum="MAV_DO_REPOSITION_FLAGS">Bitmask of option flags.</param>
         <param index="3">Reserved</param>
-        <param index="4" label="Yaw">Yaw heading. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.). For planes indicates loiter direction (0: clockwise, 1: counter clockwise)</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude (meters)</param>
+        <param index="4" label="Yaw" units="deg">Yaw heading. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.). For planes indicates loiter direction (0: clockwise, 1: counter clockwise)</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="193" name="MAV_CMD_DO_PAUSE_CONTINUE" hasLocation="false" isDestination="false">
         <description>If in a GPS controlled position mode, hold the current position or continue.</description>
@@ -2066,12 +2066,12 @@
       </entry>
       <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION" hasLocation="false" isDestination="false">
         <description>Request MAVLink protocol version compatibility</description>
-        <param index="1" label="Protocol">1: Request supported protocol versions by all nodes on the network</param>
+        <param index="1" label="Protocol" minValue="0" maxValue="1" increment="1">1: Request supported protocol versions by all nodes on the network</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="520" name="MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES" hasLocation="false" isDestination="false">
         <description>Request autopilot capabilities. The receiver should ACK the command and then emit its capabilities in an AUTOPILOT_VERSION message</description>
-        <param index="1" label="Version">1: Request autopilot version</param>
+        <param index="1" label="Version" minValue="0" maxValue="1" increment="1">1: Request autopilot version</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="521" name="MAV_CMD_REQUEST_CAMERA_INFORMATION" hasLocation="false" isDestination="false">
@@ -2301,7 +2301,7 @@
       <entry value="4001" name="MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE" hasLocation="true" isDestination="false">
         <description>This command sets submode circle when vehicle is in guided mode. Vehicle flies along a circle facing the center of the circle. The user can input the velocity along the circle and change the radius. If no input is given the vehicle will hold position.
                   </description>
-        <param index="1" label="Radius">Radius of desired circle in CIRCLE_MODE</param>
+        <param index="1" label="Radius" units="m">Radius of desired circle in CIRCLE_MODE</param>
         <param index="2">User defined</param>
         <param index="3">User defined</param>
         <param index="4">User defined</param>


message_definitions/v1.0/common.xml
```diff
@@ -1004,7 +1004,7 @@
       </entry>
       <entry value="112" name="MAV_CMD_CONDITION_DELAY">
         <description>Delay mission state machine.</description>
-        <param index="1">Delay in seconds (decimal)</param>
+        <param index="1" label="delay" units="s" minValue="0" increment="1">Delay</param>
```
hamishwillee: @amilcarlucas Doesn't decimal correspond to "float" so fractional values allowed? If only whole seconds were allowed wouldn't this have been "Delay in seconds (integer)"?

message_definitions/v1.0/common.xml
```diff
@@ -984,7 +984,7 @@
       </entry>
       <entry value="94" name="MAV_CMD_NAV_PAYLOAD_PLACE">
         <description>Descend and place payload. Vehicle moves to specified location, descends until it detects a hanging payload has reached the ground, and then releases the payload. If ground is not detected before the reaching the maximum descent value (param1), the command will complete without releasing the payload.</description>
-        <param index="1" label="Max Descent" minValue="0" units="m">Maximum distance to descend.</param>
+        <param index="1" label="Max Descent" units="m" minValue="0">Maximum distance to descend.</param>
```
hamishwillee: Are we enforcing consistent attribute declaration order? XML does not care or provide a mechanism to test/enforce this.

message_definitions/v1.0/common.xml
```diff
@@ -1020,7 +1020,7 @@
         <param index="4">Empty</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
-        <param index="7">Finish Altitude</param>
+        <param index="7" label="altitude" units="m">Finish Altitude</param>
```
hamishwillee: 1. Would prefer description as "Target Altitude".
2. While we're here, can you change the command description to "Ascend/descend to target altitude at specified rate."
3. Can we also update param 1, which has units in description. Also, is there a default behaviour here - ie something like -1 means ignore the rate and use default?

hamishwillee:
Thanks @amilcarlucas - I've added a few notes. We might merge this following your feedback and address the questions as separate PRs.

message_definitions/v1.0/common.xml
amilcarlucas: I interpreted all occurrences of decimal as integer, non-decimal as float.
@LorenzMeier can you explain what decimal means in this context.

message_definitions/v1.0/common.xml
amilcarlucas: I enforced it here, just for consistency reasons

message_definitions/v1.0/common.xml
amilcarlucas: 1. Done
2. Done.
3. Param1 update is part or another PR. It was not here to avoid merge conflicts.

>  is there a default behaviour here - ie something like -1 means ignore the rate and use default?
I have no clue, sorry. @LorenzMeier @dagar any ideas?

message_definitions/v1.0/common.xml
amilcarlucas: Like I said I did not look at the PX4 implementation, so not sure of what decimal means here.

hamishwillee: @amilcarlucas I am pretty sure I'm right here - decimal is [by definition](https://www.smartickmethod.com/blog/math/number-operations-in-base-ten/decimals/decimal-numbers/) a mechanism for representing fractional values. Even if PX4 and ArduPilot were to restrict these to integers we shouldn't redefine what the command itself does in case another system has implemented it as the definition states.

So can you remove the "increment" everywhere?

hamishwillee: PS if it doesn't specify decimal then arguably this also decimal. ie unless specified as integer it should be decimal.

message_definitions/v1.0/common.xml
```diff
@@ -1013,14 +1013,14 @@
...
-        <param index="1">Descent / Ascend rate (m/s)</param>
+        <description>Ascend/descend to target altitude at specified rate.</description>
+        <param index="1" label="rate" units="m/s">Descent / Ascend rate</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
-        <param index="7">Finish Altitude</param>
+        <param index="7" label="altitude" units="m">Target Altitude</param>
```
hamishwillee: As discussed, label needs first letter capitalisation on all words - so in this case "Altitude". FYI, I just added note about this into the official docs.

message_definitions/v1.0/common.xml
hamishwillee: Hope they comment, but if not, don't think we should wait on this. I'll merge once the labels are updated consistently

amilcarlucas:
Again, I did not validated any of the changes in this PR. These are pure guesswork

message_definitions/v1.0/common.xml
```diff
@@ -1003,7 +1003,7 @@
       </entry>
       <entry value="112" name="MAV_CMD_CONDITION_DELAY">
         <description>Delay mission state machine.</description>
-        <param index="1">Delay in seconds (decimal)</param>
+        <param index="1" label="Delay" units="s" minValue="0" increment="1">Delay</param>
```
hamishwillee: Delete increment

message_definitions/v1.0/common.xml
```diff
@@ -1012,14 +1012,14 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="113" name="MAV_CMD_CONDITION_CHANGE_ALT">
-        <description>Ascend/descend at rate.  Delay mission state machine until desired altitude reached.</description>
-        <param index="1">Descent / Ascend rate (m/s)</param>
+        <description>Ascend/descend to target altitude at specified rate.</description>
```
hamishwillee: Sorry, we do need to reappend this sentence too: "Delay mission state machine until desired altitude reached."

hamishwillee:
@amilcarlucas Added a few more comments sorry!

> Again, I did not validated any of the changes in this PR. These are pure guesswork

Understood. I don't think they need to be "validated" because you haven't changed the meaning/constraints of the message. Validation as such is more useful to clarify what something means when the text is unclear.

amilcarlucas:
Rebased and split into three clean commits. The third one contains stuff that I am not (yet) sure of.

DonLakeFlyer:
For values which are degrees most firmware will wrap values to 0-360 so there isn't reallyt a min/max.

For things like this: ```minValue="0" maxValue="1" increment="1"```. Shouldn't these be an enum instead?

amilcarlucas:
@DonLakeFlyer the 0-360 min max is meant also as a guideline for the users. Sometimes the code expects/wraps -180,180 sometimes 0,360. this is an attempt at unifying that.

Yes, enums could be created on a future PR. For now this is already a big step forward.

hamishwillee:
FWIW @DonLakeFlyer @amilcarlucas 

>> For values which are degrees most firmware will wrap values to 0-360 so there isn't reallyt a min/max.

> the 0-360 min max is meant also as a guideline for the users. Sometimes the code expects/wraps -180,180 sometimes 0,360. this is an attempt at unifying that.

My understanding was that these attributes are really GCS hints - so essentially we're just making it easier for QGC to present a 0-360 degree (or whatever) selector for users. The fact that the firmware will handle larger is kind of irrelevant for users.

>> For things like this: minValue="0" maxValue="1" increment="1". Shouldn't these be an enum instead?

> Yes, enums could be created on a future PR. For now this is already a big step forward.

Again, I thought the idea here was to allow GCS to provide a sensible option for users - in this case either 0 or 1 rather than "any value". 

In fact though I think in most cases the field is irritating and would be better off having been removed (if compatibility allows). The reason I say this is that we are using a param to say that we want particular information, but this is implied by the fact that we're sending the message (ie we wouldn't send the message if we didn't want the information).

E.g.

```
<param index="1" label="Version" minValue="0" maxValue="1" increment="1">1: Request autopilot version</param>
```
If they weren't so pointless I'd suggest maybe we actually need a boolean. As it is, if we use an enum we should make it very generic.

message_definitions/v1.0/common.xml
```diff
@@ -1200,7 +1200,7 @@
         <param index="1" label="Speed" units="m/s" minValue="-1">Ground speed, less than 0 (-1) for default</param>
         <param index="2" label="Bitmask" enum="MAV_DO_REPOSITION_FLAGS">Bitmask of option flags.</param>
         <param index="3">Reserved</param>
-        <param index="4" label="Yaw">Yaw heading, NaN for unchanged. For planes indicates loiter direction (0: clockwise, 1: counter clockwise)</param>
+        <param index="4" label="Yaw" units="deg">Yaw heading, NaN for unchanged. For planes indicates loiter direction (0: clockwise, 1: counter clockwise)</param>
```
hamishwillee: @DonLakeFlyer Do we know these are units degree for both ArduPilot and PX4?

message_definitions/v1.0/common.xml
```diff
@@ -773,7 +773,7 @@
         <param index="1">Empty</param>
         <param index="2">Empty</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4" label="Yaw">Desired yaw angle.</param>
+        <param index="4" label="Yaw" units="deg" minValue="0" maxValue="360">Desired yaw angle.</param>
```
hamishwillee: Are we sure degree units are used in current implementations?

message_definitions/v1.0/common.xml
```diff
@@ -820,10 +820,10 @@
       </entry>
       <entry value="22" name="MAV_CMD_NAV_TAKEOFF" hasLocation="true" isDestination="true">
         <description>Takeoff from ground / hand</description>
-        <param index="1" label="Pitch">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
+        <param index="1" label="Pitch" units="deg" minValue="0" maxValue="360">Minimum pitch (if airspeed sensor present), desired pitch without sensor</param>
```
hamishwillee: Are we sure degree units are used in current implementations?

message_definitions/v1.0/common.xml
```diff
@@ -820,10 +820,10 @@
...
         <param index="2">Empty</param>
         <param index="3">Empty</param>
-        <param index="4" label="Yaw">Yaw angle (if magnetometer present), ignored without magnetometer. NaN for unchanged.</param>
+        <param index="4" label="Yaw" units="deg">Yaw angle (if magnetometer present), ignored without magnetometer. NaN for unchanged.</param>
```
hamishwillee: Are we sure degree units are used in current implementations?

message_definitions/v1.0/common.xml
```diff
@@ -851,9 +851,9 @@
       <entry value="25" name="MAV_CMD_NAV_FOLLOW" hasLocation="true" isDestination="false">
         <description>Vehicle following, i.e. this waypoint represents the position of a moving vehicle</description>
         <param index="1" label="Following" increment="1">Following logic to use (e.g. loitering or sinusoidal following) - depends on specific autopilot implementation</param>
-        <param index="2" label="Ground Speed">Ground speed of vehicle to be followed</param>
+        <param index="2" label="Ground Speed" units="m/s">Ground speed of vehicle to be followed</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4" label="Yaw">Desired yaw angle.</param>
+        <param index="4" label="Yaw" units="rad">Desired yaw angle.</param>
```
hamishwillee: Are we sure **radian** units are used in current implementations? Why different from the others, which are degrees?

message_definitions/v1.0/common.xml
```diff
@@ -1737,8 +1737,8 @@
       </entry>
       <entry value="2800" name="MAV_CMD_PANORAMA_CREATE" hasLocation="false" isDestination="false">
         <description>Create a panorama at the current position</description>
-        <param index="1" label="Horizontal Angle" units="deg">Viewing angle horizontal of the panorama (+- 0.5 the total angle)</param>
-        <param index="2" label="Vertical Angle" units="deg">Viewing angle vertical of panorama.</param>
+        <param index="1" label="Horizontal Angle" units="deg" minValue="0" maxValue="360">Viewing angle horizontal of the panorama (+- 0.5 the total angle)</param>
```
hamishwillee: Are we sure degree units are used in current implementations?

message_definitions/v1.0/common.xml
```diff
@@ -1758,7 +1758,7 @@
       <entry value="4001" name="MAV_CMD_SET_GUIDED_SUBMODE_CIRCLE" hasLocation="true" isDestination="false">
         <description>This command sets submode circle when vehicle is in guided mode. Vehicle flies along a circle facing the center of the circle. The user can input the velocity along the circle and change the radius. If no input is given the vehicle will hold position.
                   </description>
-        <param index="1" label="Radius">Radius of desired circle in CIRCLE_MODE</param>
+        <param index="1" label="Radius" units="m">Radius of desired circle in CIRCLE_MODE</param>
```
hamishwillee: Just FYI, seems odd to have a custom mode in common.xml.

message_definitions/v1.0/common.xml
```diff
@@ -1770,7 +1770,7 @@
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Delay mission state machine until gate has been reached.</description>
         <param index="1" label="Geometry" minValue="0" increment="1">Geometry: 0: orthogonal to path between previous and next waypoint.</param>
```
hamishwillee: Do we need a maxvalue on this?

hamishwillee:
This will need to be rebased. My only concern is the units, which we have stated as degrees in most cases, and radians in one or two. We should at least confirm we're not contradicting ArduPilot and PX4 by this assumption.

amilcarlucas:
Rebased. @hamishwillee like I said in the description (very first post) I am unsure of all of this and did not (yet) validated any of these changes against the ardupilot source code.
That is the reason why this is a separated PR. The ones I validated have been merged already.

message_definitions/v1.0/common.xml
DonLakeFlyer: Don't know

DonLakeFlyer:
> My understanding was that these attributes are really GCS hints

To me they also define the spec to a mavlink implementor. Not sure how forcing to 0-360 is super helpful from a UI standpoint. For example -10 degrees may be more inituitive than 350 degrees.

hamishwillee:
@amilcarlucas 

> Rebased. @hamishwillee like I said in the description (very first post) I am unsure of all of this and did not (yet) validated any of these changes against the ardupilot source code.
That is the reason why this is a separated PR. The ones I validated have been merged already in #1100.

Yes you did - sorry, revisiting this after 3 weeks ignored that. We can't merge until the actual usage is confirmed on at least the main flight stacks.

>> My understanding was that these attributes are really GCS hints

> To me they also define the spec to a mavlink implementor. Not sure how forcing to 0-360 is super helpful from a UI standpoint. For example -10 degrees may be more inituitive than 350 degrees.

@DonLakeFlyer Agreed. So what is the solution - we just remove the max/min for these cases? We could set to -360 to +360 as a hint to the UI, but a GCS could easily assume that anyway.

DonLakeFlyer:
> we just remove the max/min for these cases

That's what I would do

hamishwillee:
>> we just remove the max/min for these cases

> That's what I would do

 @DonLakeFlyer 

I'm in two minds. Personally I think -360 to 360 would be better because it is more explicit (ie we're saying that you can't specify a > 360 degree rotation to reach a yaw) and we're making it clear that negative values are OK. You could argue that this was obvious, and you'd be right, but then I was happy with 0 -360 :-(

Also, it stops someone adding the wrong values in future.

@amilcarlucas I'm happy for you to make the decision here.

hamishwillee:
@amilcarlucas All these can be degrees for yaw (confirmed):
```
MAV_CMD_NAV_WAYPOINT
MAV_CMD_NAV_LOITER_UNLIM
MAV_CMD_NAV_LOITER_UNLIM
MAV_CMD_NAV_LOITER_TIME
MAV_CMD_NAV_LAND
MAV_CMD_NAV_TAKEOFF
```
See https://github.com/mavlink/qgroundcontrol/pull/7351#issuecomment-484098712 - basically PX4 does them as degrees, and ArduPilot does not implement the yaw param.

amilcarlucas:
I was thinking more in the lines of [-180 360[ it does allow for both intervals -180 to 180 and 0 to 360

Would that be OK for you ?

DonLakeFlyer:
> Also, it stops someone adding the wrong values in future.

As far as I know the current spec is degrees can be anything positive or negative. That is how it is implemented in firmware. Making it be something other than that changes the spec.

hamishwillee:
> As far as I know the current spec is degrees can be anything positive or negative. That is how it is implemented in firmware. Making it be something other than that changes the spec.

@amilicar GIven that we shouldn't change the spec constraints without broader approavl, just remove the max/min. We can restrict in future as a separate PR or raise in dev call.

hamishwillee:
@amilcarlucas Are you still working on this? I'm not sure what is outstanding or now. If not, propose we close this and work on the ones that we know have been agreed above?

amilcarlucas:
yes, I am still working on this, hope to get the comments handled soon.

amilcarlucas:
I moved some of the easy ones to a new PR, and rebased and updated this one

message_definitions/v1.0/common.xml
hamishwillee: Also are we sure that negative values are not supported?

amilcarlucas:
I think the comments are addressed now. I rebased this again on master

amilcarlucas:
I used github grep to search for MAV_CMD_NAV_FOLLOW in PX4_firmware. It returned no hits.
Where is that message processed?

message_definitions/v1.0/common.xml
amilcarlucas: I can not find the implementation in PX4

message_definitions/v1.0/common.xml
amilcarlucas: Not changed by this PR

message_definitions/v1.0/common.xml
hamishwillee: Stuck with master

message_definitions/v1.0/common.xml
hamishwillee: Current master now has a value. I stuck with that.

message_definitions/v1.0/common.xml
hamishwillee: Leaving for post process.

hamishwillee:
Hi @amilcarlucas . I updated to master then merged this. Open questions either resolved or ignored as "we're no worse off than before". 
Thanks for this!

amilcarlucas:
Thanks

PR #1106, opened 2019-03-26
Add SATCOM_LINK_STATUS status to ASLUAV

acfloria:
Add a message to transmit the useful information about the status of the SatCom link.

Currently this is only added to ASLUAV because we need to merge it quickly. However, if such a message would be useful for others we could add it as a v2 message to common.

### DIFF
--- a/message_definitions/v1.0/ASLUAV.xml
+++ b/message_definitions/v1.0/ASLUAV.xml
@@ -270,5 +270,16 @@
       <field type="uint8_t" name="sinr_ecio">SINR (LTE) or ECIO (WCDMA) as reported by modem (unconverted)</field>
       <field type="uint8_t" name="rsrq">RSRQ (LTE only) as reported by modem (unconverted)</field>
     </message>
+    <message id="214" name="SATCOM_LINK_STATUS">
+      <description>Status of the SatCom link</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
+      <field type="uint64_t" name="last_heartbeat" units="us">Timestamp of the last successful sbd session</field>
+      <field type="uint16_t" name="failed_sessions"> Number of failed sessions </field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@dagar @auturgy @LorenzMeier 
So we usually merge into dialects on the "this does no harm" basis. Any problem with that here?

More generally @acfloria Asked whether we want this in common? Any thoughts on this? There is a whole lot of stuff on link status going on, and it would seem better to generalise.

auturgy:
To be generally useful it needs more information, as there are a number of different methods used for satellite links: the “session” related fields in particular would need to be re-thought, as they are likely specific to the system ASL are using.
No problem going into the dialect though.

Regards,

James

acfloria:
I removed now the white spaces.

@auturgy this implementation matches how we implemented the SatCom on master in PX4 using the RockBlock/Iridium SBD system. There you always send data in packages of a few hundreds of bytes. I think the message defined would nicely cover all SBD based systems but I am open to discuss it.

auturgy:
Perhaps it would be better called SBD_LINK_STATUS then? 
My point is that as written this message doesn’t cater for a generic satcom link, and unless that is clear it will cause future confusion.
Once a message is in common it gets hard to change, so better to get it sorted out first.
As I said earlier, no issues merging to your dialect though.

Regards,

James

acfloria:
Yeah, actually SBD_LINK_STATUS might make more sense if we merge it to common.

Can we in the meantime merge it to the ASLUAV dialect while we figure out the details of the version we would put in common?

hamishwillee:
@acfloria I have merged this - generally we let dialects "do as they wish". 

Appreciate you offering this to common. Perhaps the easiest way to progress this is if you could create a PR into common that includes anything gathered from the discussion - ie the new name. 
We may well have to use a different message ID too - does this really have to be in the MAVLink 1 range?

PR #1108, opened 2019-03-31
common.xml:Loiter* MAV_CMD set NaN for unchanged yaw

hamishwillee:
Follows from Slack discussion with @DonLakeFlyer @auturgy @LorenzMeier that there is no way to set that desired yaw setpoint is unchanged in current definition. This change is consistent with WAYPOINT command.

This is a WIP while we resolve how ArduPilot handles this case.


hamishwillee:
FYI, this comes out of https://github.com/mavlink/qgroundcontrol/issues/6783#issuecomment-478299939

hamishwillee:
@DonLakeFlyer FYI

PR #1110, opened 2019-04-01
script: skip pymavlink install

khancyr:
skip useless pymavlink installation.
Pymavink already get its autotest.
Moreover this prevent failure without message when .xml aren't valid as pymavlink try to build then for install.
At least, it allow docker build to use much small image as pymavlink gcc dependency isn't need anymore (save a lot of space, so that is greener ! ;-P )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,9 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
-  
+
+cache: pip
+
 matrix:
   include:
     - python: 3.7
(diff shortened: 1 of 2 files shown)


hamishwillee:
@khancyr Hi, looks good to me, but fails Travis: "Unable to locate package python-future". I don't know why, because that should be in xenial (ie works on my Ubuntu 16.04 installation). Perhaps install from PyPi?

khancyr:
yep I will switch to pip install

khancyr:
Hello, 
I have update this. I remove pymavlink install, add cache for pip dependencies. And remove quaterion and rotmat test that are pymavlink tests... Pymavlink is already doing them and changing the message definition don't impact them, so they are useless in this repo.

amilcarlucas:
can you please rebase?

khancyr:
@amilcarlucas rebased !

LorenzMeier:
Thanks!

PR #1111, opened 2019-04-04
common.xml: Clarify that safe_return means RTL, Return, Smart RTL etc.

hamishwillee:


hamishwillee:
I'm going to self merge.

PR #1112, opened 2019-04-04
MAV_FRAME:Deprecate component-specific frames

hamishwillee:
As discussed in the Dev call, these messages are duplicates of existing frames that have the origin component in the name. This is not how MAVLink works - if origin is needed this should come from a component id (and ideally the information should be formatted so that the origin is not even relevant).

Action is to mark these as reserved and deprecated.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -589,7 +589,7 @@
         <description>Global (WGS84) coordinate frame + MSL altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="1" name="MAV_FRAME_LOCAL_NED">
-        <description>Local coordinate frame, Z-down (x: north, y: east, z: down).</description>
+        <description>Local coordinate frame, Z-down (x: North, y: East, z: Down).</description>
       </entry>
       <entry value="2" name="MAV_FRAME_MISSION">
         <description>NOT a coordinate frame, indicates a mission command.</description>
@@ -598,7 +598,7 @@
         <description>Global (WGS84) coordinate frame + altitude relative to the home position. First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
       </entry>
       <entry value="4" name="MAV_FRAME_LOCAL_ENU">
-        <description>Local coordinate frame, Z-up (x: east, y: north, z: up).</description>
+        <description>Local coordinate frame, Z-up (x: East, y: North, z: Up).</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
         <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1.0e-7, second value / y: longitude in degrees*1.0e-7, third value / z: positive altitude over mean sea level (MSL).</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -589,7 +589,7 @@
         <description>Global (WGS84) coordinate frame + MSL altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="1" name="MAV_FRAME_LOCAL_NED">
-        <description>Local coordinate frame, Z-down (x: north, y: east, z: down).</description>
+        <description>Local coordinate frame, Z-down (x: North, y: East, z: Down).</description>
```
hamishwillee: FYI only, North, East are proper nouns, so need to be capitalised. At that point for consistency you should capitalise the others.

message_definitions/v1.0/common.xml
```diff
@@ -621,29 +621,37 @@
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
         <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
-      <entry value="12" name="MAV_FRAME_BODY_FRD">
-        <description>Body fixed frame of reference, Z-down (x: forward, y: right, z: down).</description>
+      <entry value="12" name="MAV_FRAME_RESERVED_12">
```
hamishwillee: @LorenzMeier Please confirm this is the format you wanted
- Name changed to MAV_FRAME_RESERVED_XX to force compile error. 
- I put the original name in the description to make it easy for people to work out what this was previously.

message_definitions/v1.0/common.xml
```diff
@@ -621,29 +621,37 @@
...
+        <deprecated since="2019-04" replaced_by="MAV_FRAME_BODY_OFFSET_NED"/>
+        <description>MAV_FRAME_BODY_FRD - Body fixed frame of reference, Z-down (x: Forward, y: Right, z: Down).</description>
       </entry>
-      <entry value="13" name="MAV_FRAME_BODY_FLU">
-        <description>Body fixed frame of reference, Z-up (x: forward, y: left, z: up).</description>
+      <entry value="13" name="MAV_FRAME_RESERVED_13">
+        <deprecated since="2019-04" replaced_by=""/>
```
hamishwillee: What is the equivalent of MAV_FRAME_BODY_FLU in other options - ie what is this "replaced by"

hamishwillee:
@LorenzMeier @auturgy Can you guys please check that I used the correct `replaced-by` references - ie such that if people do the replacement this will not break existing code. There is one marked that I'm not sure of the correct replacement.

LorenzMeier:
Could you rebase?

hamishwillee:
Yes, and I have merged this.

FYI, most of my open questions had since been fixed in master, so this was now mergable.

TSC21:
This spec change needs a correspondent source code update. Example: https://github.com/PX4/sitl_gazebo/blob/master/src/gazebo_mavlink_interface.cpp#L1232 or https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L1328-L1375.

hamishwillee:
@auturgy @peterbarker This _may_ have a corresponding source code change in ArduPilot too. I haven't checked, but I note that PX4 just did one above ^^^^

auturgy:
I don’t think we use the ones removed.

Regards,

James

message_definitions/v1.0/common.xml
```diff
@@ -879,35 +879,37 @@
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
         <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
-      <entry value="12" name="MAV_FRAME_BODY_FRD">
```
kamilritz, line 882: @hamishwillee @LorenzMeier  Which is the frame name replacing this? BODY_NED?
There was once a discussion that BODY_NED is confusing as it suggest that it is aligned with north. Additionally the NED<-->ENU conversion is not the same as the FRD<-->FLU conversion.  BODY_FRD is in my opinion a much better name.
I have seen enough people struggling with these frame -naming / -conversion. I argue strongly for using BODY_FRD instead of BODY_NED.

hamishwillee: @kamilritz Thanks. See line below. According to that MAV_FRAME_BODY_FRD was replaced by MAV_FRAME_BODY_OFFSET_NED.

Probably best if you dislike a particular case to create a PR discussing that case. I have tried to be uninvolved in the discussion as I do not feel informed enough to "do the right thing".

kamilritz:
@hamishwillee @LorenzMeier @auturgy 
This PR was marking the `MAV_FRAME_BODY_FRD` as deprecated by the `MAV_FRAME_BODY_OFFSET_NED`. This is not valid, since both `MAV_FRAME_BODY_NED` and `MAV_FRAME_BODY_OFFSET_NED` are marked as deprecated by `MAV_FRAME_BODY_FRD` previously and still are.

@hamishwillee I would really appreciate it, if you could fix this. I would suggest you rename both `MAV_FRAME_BODY_NED` and `MAV_FRAME_BODY_OFFSET_NED` as reserved and revert the deprecation of both `MAV_FRAME_BODY_FRD` and `MAV_FRAME_BODY_FLU`.

hamishwillee:
@kamilritz Thanks. 

@julianoes This PR deprecated MAV_FRAME_BODY_FRD and MAV_FRAME_BODY_FLU. This is a problem because it creates a cyclic replaced by with MAV_FRAME_BODY_OFFSET_NED.

So the question is, are MAV_FRAME_BODY_FRD and MAV_FRAME_BODY_FLU frames that we want or not.
If we do need these then agree on approach with @kamilritz .

If we do not need these then we either need to have a genuine "replaced by" or a comment as to why they should not be used. 

Can you advise on the either-or above and who can answer if not?

kamilritz:
@hamishwillee 
I have to admit I don't fully understand the difference between MAV_FRAME_BODY_NED and MAV_FRAME_BODY_OFFSET_NED. But one of them is for sure identical to MAV_FRAME_BODY_FRD.

It would be possible to leave the MAV_FRAME_BODY_FLU away, as each part in the software could potentially easily switch from *_FRD to *_FLU.

The concept of an *offset* frame such as MAV_FRAME_LOCAL_OFFSET_NED seems very strange to me.

I checked the Firmware and the following frames are not in use:
MAV_FRAME_LOCAL_OFFSET_NED, MAV_FRAME_BODY_OFFSET_NED, MAV_FRAME_BODY_NED

hamishwillee:
I don't know if it helps or hinders, but "traditionally" the "OFFSET" meant "in the frame of the vehicle heading". A very long time ago I wrote the section [here](https://ardupilot.org/dev/docs/copter-commands-in-guided-mode.html#set-position-target-local-ned) that kind of explains it.  I believe the intent of these changes may have been to get away from the odd mix of frame for position and velocity. 

Anyway, @julianoes is my go to expert :-). He can probably confirm. I am OK with both FRD and FLU, though I get your point.

kamilritz:
@hamishwillee Thanks, this is explaining it well. Still a bit an odd concept.

julianoes:
> Anyway, @julianoes is my go to expert :-). He can probably confirm. I am OK with both FRD and FLU, though I get your point.

I'm not an expert when it comes to frames, @kamilritz is, I barely know the acronyms.

This is the first time I'm reading the `OFFSET` naming, and I don't think I'm convinced. Mostly because offset would imply to be offset from something but I'm not sure what that is.

I agree with what @kamilritz says:
> @hamishwillee I would really appreciate it, if you could fix this. I would suggest you rename both MAV_FRAME_BODY_NED and MAV_FRAME_BODY_OFFSET_NED as reserved and revert the deprecation of both MAV_FRAME_BODY_FRD and MAV_FRAME_BODY_FLU.

I'm fine with deprecating `FLU` just because I don't like multiple ways to do the same thing.

kamilritz:
> I'm fine with deprecating FLU just because I don't like multiple ways to do the same thing.

Me too.

hamishwillee:
OK, I recovered MAV_FRAME_BODY_FRD (only) in #1381. NOTE, still waiting on Lorenz to fix up DNS for MAVLink so update won't propagate to docs yet.

Thanks for following this up @kamilritz 

@julianoes FWIW  The OFFSET might be seen as LOCAL frame offset to current vehicle position. This is what is meant in the description by

> Anything expressed in this frame should be added to the current local frame position.".

@kamilritz Would it be just as accurate/more useful to say:

> Positions specified in this frame are relative to the current local frame position of the system (i.e. to the vehicle center of gravity)

Below summarises the convention.

- LOCAL 
  - positions relative to EKF origin in indicated frame [ie home position in local coordinates]
  - velocity relative to indicated frame (ie no effect on velocity)
- OFFSET
  - Positions are relative to the vehicle’s current position
  - Velocity relative to indicated frame (ie offset has no effect on velocity)
- BODY 
    - positions relative to EKF origin in indicated frame
    - velocities relative to heading. 
- BODY_OFFSET (Deprecated)
  - positions relative to vehicle heading and position
  - velocities relative to vehicle heading.

PR #1113, opened 2019-04-10
scripts: remove bash dependency on test.sh

khancyr:
we don't need bash, so I remove it.
It allow to work on default alpine docker images


amilcarlucas:
Tests do pass, merging. Thanks.

PR #1115, opened 2019-04-10
common.xml: change NAV_SET_YAW_SPEED parameter attributes

arjun7965:
as per @rmackay9's [post](https://github.com/ArduPilot/ardupilot/issues/10803)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1370,7 +1370,7 @@
         <description>Sets a desired vehicle turn angle and speed change.</description>
         <param index="1" label="Yaw" units="cdeg">Yaw angle to adjust steering by.</param>
         <param index="2" label="Speed" minValue="0" maxValue="1">speed - normalized to 0 .. 1</param>
-        <param index="3">Empty</param>
+        <param index="3" label="Angle" minValue="0" maxValue="1" increment="1">Final angle. (0=absolute, 1=relative)</param>
         <param index="4">Empty</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>


arjun7965:
fixed @amilcarlucas :+1:

arjun7965:
> Please also add increment="1" attribute.

Fixed

amilcarlucas:
@arjun7965 Thanks, now please add a new PR, that adds the first two changes that I asked you to revert.

arjun7965:
@amilcarlucas done. See #1120

message_definitions/v1.0/common.xml
```diff
@@ -1374,7 +1374,7 @@
         <description>Sets a desired vehicle turn angle and speed change.</description>
         <param index="1" label="Yaw" units="cdeg">Yaw angle to adjust steering by.</param>
         <param index="2" label="Speed" minValue="0" maxValue="1">speed - normalized to 0 .. 1</param>
-        <param index="3">Empty</param>
+        <param index="3" label="Angle" minValue="0" maxValue="1" increment="1">Final angle. (0=absolute, 1=relative)</param>
```
hamishwillee: @amilcarlucas @arjun7965 Just to be clear are you saying that:
- if index 3 = 0 the yaw param will be relative to the body frame (forward-facing-upright). 
- if index 3 = 1 the yaw will be relative to (added to) the original yaw.

IMO the label "Angle" and description "Final angle" do not convey this.

Perhaps:
- label: "Yaw reference"
- description: 1: Yaw param (1) is relative to the vehicle body frame. 2: Yaw param is relative to current/existing yaw when message received.

PR #1119, opened 2019-04-23
common.xml: add links to the microservices documentation to the messa…

amilcarlucas:
…ge descriptions


LorenzMeier:
We also need a "service" or alike XML tag that formally pulls these together. Feel free to implement that if you beat me to it.

PR #1120, opened 2019-04-24
common.xml: change NAV_SET_YAW_SPEED parameter attributes

arjun7965:
as per this [post](https://github.com/ArduPilot/ardupilot/issues/10803) 
Continuation to #1115

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1372,8 +1372,8 @@
       <!-- 211 and 212 used in dialects -->
       <entry value="213" name="MAV_CMD_NAV_SET_YAW_SPEED" hasLocation="false" isDestination="false">
         <description>Sets a desired vehicle turn angle and speed change.</description>
-        <param index="1" label="Yaw" units="cdeg">Yaw angle to adjust steering by.</param>
-        <param index="2" label="Speed" minValue="0" maxValue="1">speed - normalized to 0 .. 1</param>
+        <param index="1" label="Yaw" units="deg">Yaw angle to adjust steering by.</param>
+        <param index="2" label="Speed" units="m/s">Speed.</param>
         <param index="3" label="Angle" minValue="0" maxValue="1" increment="1">Final angle. (0=absolute, 1=relative)</param>
         <param index="4">Empty</param>
         <param index="5">Empty</param>


message_definitions/v1.0/common.xml
```diff
@@ -1372,8 +1372,8 @@
       <!-- 211 and 212 used in dialects -->
       <entry value="213" name="MAV_CMD_NAV_SET_YAW_SPEED" hasLocation="false" isDestination="false">
         <description>Sets a desired vehicle turn angle and speed change.</description>
-        <param index="1" label="Yaw" units="cdeg">Yaw angle to adjust steering by.</param>
-        <param index="2" label="Speed" minValue="0" maxValue="1">speed - normalized to 0 .. 1</param>
+        <param index="1" label="Yaw" units="deg">Yaw angle to adjust steering by.</param>
+        <param index="2" label="Speed" units="m/s">speed in m/s</param>
```
amilcarlucas: please remove the " in m/s" part. It is redundant.

arjun7965: Fixed.

message_definitions/v1.0/common.xml
```diff
@@ -1372,8 +1372,8 @@
       <!-- 211 and 212 used in dialects -->
       <entry value="213" name="MAV_CMD_NAV_SET_YAW_SPEED" hasLocation="false" isDestination="false">
         <description>Sets a desired vehicle turn angle and speed change.</description>
-        <param index="1" label="Yaw" units="cdeg">Yaw angle to adjust steering by.</param>
-        <param index="2" label="Speed" minValue="0" maxValue="1">speed - normalized to 0 .. 1</param>
+        <param index="1" label="Yaw" units="deg">Yaw angle to adjust steering by.</param>
```
hamishwillee: This is a change to the behaviour - ie other flight stacks/GCS are going to be expecting/supplying cdeg. Has this been discussed/reviewed against other stacks?

message_definitions/v1.0/common.xml
```diff
@@ -1372,8 +1372,8 @@
...
+        <param index="2" label="Speed" units="m/s">Speed.</param>
```
hamishwillee: This is a change to the behaviour - ie other flight stacks/GCS are going to be expecting/supplying a normalised value of speed. I think the change is good - but it needs to be confirmed against at least the major stacks.

message_definitions/v1.0/common.xml
arjun7965: @hamishwillee Ardupilot currently uses the speed as m/s (for missions). I did check in PX4 flight stack and I did not see them using this MAV_CMD (I may be wrong).

message_definitions/v1.0/common.xml
arjun7965: Ardupilot uses the speed in m/s (for missions) (https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mission/AP_Mission.cpp#L953)

I did check PX4 flight stack and I did not see an implementation of this message( I might be wrong)

hamishwillee:
@arjun7965 I checked as well and can't find any use in PX4 either.

OK, I am going to merge this because it is consistent with all other usage of Yaw and speed in mission commands in both ArduPilot and PX4 (ie using degrees and m/s)

PR #1125, opened 2019-04-29
Add ISBD_LINK_STATUS in common

acfloria:
Based on the comments on  [PR #1106](https://github.com/mavlink/mavlink/pull/1106) I:

- renamed the SATCOM_LINK_STATUS to ISBD_LINK_STATUS (ISBD = Iridium Short Burst Data)
- gave it an id in the MAVLink2 range

I think it would be nice to have this message in common because it would represent the status of the iridiumsbd driver which is in pixhawk master and thus potentially be helpful for other people outside ASL.
However, I am open to suggestions and if you think it should not be in common we can leave it in the ASLUAV dialect.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5101,6 +5101,17 @@
       <field type="uint16_t" name="lac">Location area code. If unknown, set to: 0</field>
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
+    <message id="335" name="ISBD_LINK_STATUS">
+      <description>Status of the Iridium SBD link</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
+      <field type="uint64_t" name="last_heartbeat" units="us">Timestamp of the last successful sbd session</field>
+      <field type="uint16_t" name="failed_sessions">Number of failed sbd sessions</field>
+      <field type="uint16_t" name="successful_sessions">Number of successful sbd sessions</field>
+      <field type="uint8_t" name="signal_quality">Signal quality</field>
+      <field type="uint8_t" name="ring_pending">Ring call pending</field>
+      <field type="uint8_t" name="tx_session_pending">Transmission session pending</field>
+      <field type="uint8_t" name="rx_session_pending">Receiving session pending</field>
+    </message>
     <message id="340" name="UTM_GLOBAL_POSITION">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5102,15 +5102,15 @@
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
     <message id="335" name="ISBD_LINK_STATUS">
-      <description>Status of the Iridium SBD link</description>
-      <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
-      <field type="uint64_t" name="last_heartbeat" units="us">Timestamp of the last successful sbd session</field>
-      <field type="uint16_t" name="failed_sessions">Number of failed sbd sessions</field>
-      <field type="uint16_t" name="successful_sessions">Number of successful sbd sessions</field>
-      <field type="uint8_t" name="signal_quality">Signal quality</field>
-      <field type="uint8_t" name="ring_pending">Ring call pending</field>
-      <field type="uint8_t" name="tx_session_pending">Transmission session pending</field>
-      <field type="uint8_t" name="rx_session_pending">Receiving session pending</field>
+      <description>Status of the Iridium SBD link.</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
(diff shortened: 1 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5057,6 +5057,17 @@
       <field type="uint16_t" name="lac">Location area code. If unknown, set to: 0</field>
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
+    <message id="335" name="ISBD_LINK_STATUS">
+      <description>Status of the Iridium SBD link</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
+      <field type="uint64_t" name="last_heartbeat" units="us">Timestamp of the last successful sbd session</field>
+      <field type="uint16_t" name="failed_sessions">Number of failed sbd sessions</field>
+      <field type="uint16_t" name="successful_sessions">Number of successful sbd sessions</field>
+      <field type="uint8_t" name="signal_quality">Signal quality</field>
```
hamishwillee: units? (added as units="")

message_definitions/v1.0/common.xml
```diff
@@ -5057,6 +5057,17 @@
       <field type="uint16_t" name="lac">Location area code. If unknown, set to: 0</field>
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
+    <message id="335" name="ISBD_LINK_STATUS">
+      <description>Status of the Iridium SBD link</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
```
hamishwillee: Since when? Typically description on these is:

> Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.

message_definitions/v1.0/common.xml
```diff
@@ -5057,6 +5057,17 @@
...
+      <field type="uint64_t" name="last_heartbeat" units="us">Timestamp of the last successful sbd session</field>
+      <field type="uint16_t" name="failed_sessions">Number of failed sbd sessions</field>
+      <field type="uint16_t" name="successful_sessions">Number of successful sbd sessions</field>
+      <field type="uint8_t" name="signal_quality">Signal quality</field>
+      <field type="uint8_t" name="ring_pending">Ring call pending</field>
```
hamishwillee: How are this and the next two used? Ie is this a count of pending calls or something else?

hamishwillee:
> Given that there are not 20 different vom sat systems deployed it makes sense.

@LorenzMeier Generally at what point do we decide future proofing by having generic message makes sense?

@auturgy @WickedShell @amilcarlucas Further thoughts?

message_definitions/v1.0/common.xml
hamishwillee: Are all these values always available? is there/does there need to be an indicator of "value not supported/value not supplied"?

message_definitions/v1.0/common.xml
acfloria: In the PX4 driver it would be the time since system boot, but I can change the description.

message_definitions/v1.0/common.xml
acfloria: The one we have does not really have units, it is on a scale between 0 to 5 where 0 indicates no signal at all and 1 to 5 indicate that a message can be sent. Would units="" indicate no units?

message_definitions/v1.0/common.xml
acfloria: Those values are boolean which just indicate if either a ring call, transmission session or receiving session is pending. We could change it to a signed integer and set -1 to "value not supported/value not supplied" but also leaving it at 0 would be fine since this then is essentially the same.

hamishwillee: Thanks. I would leave the type and just have description as follows (as you say, there is no need to differentiate 0 and not supported)

> 1: Ring call pending, 0: No call pending." (and similar)

message_definitions/v1.0/common.xml
hamishwillee: That would be great. You can copy the standard text for this. All that is important is to make clear to implementers and users what exactly the timestamp means

message_definitions/v1.0/common.xml
hamishwillee: No, just omit the units. For this one you'd just be precise in the description. Ie something like:

> Signal quality. Range is 0 to 5, where 0 indicates no signal and 5 indicates maximum signal strength.

If you can't assume that the values are evenly distributed you might add ` (values are linearly distributed)`

If the signal quality can ever not be supported by the sending system, you might add the sentence.

> UINT8_MAX indicates that sending signal quality is not supported.

More generally we _try_ to use 0 to indicate not supported if possible, otherwise max value for uints, -1 for ints, -1 or NaN for floats.

amilcarlucas:
I will review once the changes are made.

message_definitions/v1.0/common.xml
hamishwillee: Thanks!

hamishwillee:
Other than inconsistent use of full stops at end of some sentences this looks good to me.

@amilcarlucas @auturgy @LorenzMeier Anything to add?

acfloria:
What is the policy regarding the punctuation? By looking at the other messages I could determine which one would be correct.

Are there any other changes required to merge this PR?

hamishwillee:
There is no strict policy, but within a message the punctuation should be consistent - ie all descriptions have full stops or none.

It is MORE common to have no full stops for single sentences, and some people have expressed this preference (personally I prefer full stops everywhere, because if you have 2-sentence descriptions it looks odd to only have the punctuation on the first one.

As soon as you have made consistent I will merge.

hamishwillee:
@acfloria Waiting on you. I can merge once this is consistent.

acfloria:
@hamishwillee I changed it now. Sorry it took me so long. I was first at a conference and then on vacation.

PR #1127, opened 2019-05-01
MAVLink v2 C library efficiency improvements

olliw42:
although saying to be "highly optimized for resource-constrained systems with limited RAM and flash memory" the MAVLink C lib fails to be efficient in most aspects, flash, ram, stack, computation time, which can be relevant especially for resource constrained embeeded applications. This PR adds extensions which much improve the situation. It is offered just in case anyone might be interested.

It deals with only C and v2.0.

It is used in two applications, the STorM32 gimbal controller, which now makes heavy use of the new 'light' features, and the UC4H MavlinkBridge, which doesn't use them but uses the convenience functions (and thus kind of tests compatibility). That is, various, but by far not all aspects of the lib are tested, e.g. I don't use signatures at all, etc.

The additions/extensions are essentially non-intrusive, with maybe one exception.

I couldn't figure out how to properly deal with git submodules, so I simply created a new folder "pymavlink-light", into which I put all my changes, and added a mavgenerate_light.py which is to be used. For this PR I recreated all anew in order to achieve a series of commits which hopefully helps following.

The additions/extensions in detail:

**1) Receiving memory usage**

Currently the memory usage in receiving a message may be sketched as follows

user data -> payload_t -> msg_t of user -> buffer of user -> uart tx buf

There are several versions, pack, encode, send etc, which allow several things to do and skip/simplify some steps, e.g. the "msg_t of user" can be send directly to the "uart tx buf".

However, what is not possible is to put the provided message data directly into the "buffer of user". In many cases that's however what is desired since the "msg_t of user" typically has no further use. Msg_t is a quite big structure, and thus costly.

I thus added functions mavlink_msg_MSG_pack_txbuf() to the messages .h files. For instance, for the heartbeat there is now a function mavlink_msg_heartbeat_pack_txbuf(), which skips the need of "msg_t of user".

This is a totally non-intrusive addition, as it doesn't affect any of the other code.

**2) Parsing memory usage**

Currently the memory usage in parsing may be sketched as follows:

uart rx buf -> msg_t in COMM buffer -> msg_t of user -> payload_t

In many cases the double use of a msg_t structure is not needed, since the parsing could directly go into "msg_t of user". Msg_t is a quite big structure, and thus costly.

The new NULL check changes in mavlink_frame_char_buffer() allow to do that, however the whole thing is still quite back-through-the-door, e.g. the user needs to define his own mavlink_get_channel_buffer() to refer to his own msg_t variable, or keep using the COMM_BUF, etc..

I thus added a new function mavlink_parse_nextchar() which is much simpler to use, and also results in a logically structured API (i.e. mavlink_frame_char_buffer() is supposed to use now mavlink_parse_nextchar(), like mavlink_frame_char() uses mavlink_parse_nextchar(), and mavlink_parse_char() uses mavlink_frame_char()).

This easily could be made to be a totally non-intrusive addition, as it doesn't have to affect any of the other code. As is, it is intrusive however, since it also adds the handling of the msg_entry extension (see below). (is just one line of code, so easily removed if not wanted)

**3) FLASH usage and computation time**

MAVLINK_MESSAGE_CRCS is typically quite big, which has two implications, lots of flash used, and in addition the search by mavlink_get_msg_entry() takes long.

In many cases the application doesn't use all messages of a dialect, but just a relatively small subset, 10-20% or so. Having MAVLINK_MESSAGE_CRCS to only contain the entries for the used (incoming) messages can thus lower flash usage as well as make the search faster. With the lib as is this can be achieved in several ways, e.g. by doing one's own xml, or by copying the dialect's original MAVLINK_MESSAGE_CRCS and deleting all unused entries, etc. But all that is not very convenient and can add maintenance burdens.

I thus added a message_entries.h file to the respective dialect subfolder, which offers macros for each message entry. This allows the user to very simply build her own MAVLINK_MESSAGE_CRCS.

The gains can be quite substantial. E.g. for the STorM32, which has a decent MAVLink support, this saved 2760 bytes of flash. Also the computation time for mavlink_get_msg_entry() has become more than twice faster.

This is a totally non-intrusive addition, as it doesn't affect any of the other code.

**4) msg_entry_t, computation time**

Handling of incoming messages requires parsing the data stream, as well as some sort of identification if the received message should be handled (kind of the first step of routing). Currently, with the lib as is, the mavlink_get_msg_entry() is called at least twice: It is called in the parsing function, and then has to be called by the identification code to get the target system/component.

Note that mavlink_get_msg_entry() does a search which by principle has to be rated as costly (for as long as quantum computers have not yet become reality for us), and double-doing the search isn't clever.

I thus added a msg_entry field to the message_t structure as additional metadata. 

This allows to set it in the parsing function, and then later be reused whenever needed without any additional search. Since it's just a pointer it's insignificant memory wise in comparison to the size of the message_t structure.

This easily could be made to be totally non-intrusive, as it doesn't have to affect any of the other code. As is, it is intrusive however, since it adds this additional field to the message_t structure. I note that message_t has already some meta data, so this shouldn't be a shoking approach, and that I would not think that this is very troublesome (unless the user does "strange things" like relying explicitely on the length of the structure). Adding of the msg_entry to the message_t structure could have been avoided by returning it by the parse function (which would need double pointers, argh), but I just think that the approach I've taken is clean and simple.

Along with that change I've added helper functions to get the target system/component. 

I note that this improvement should be very significant also and especially for "big" applications, which use huge dialects and many MAVLink channels with lots of incoming traffic, and thus many search events. I thus think that this improvement should also be seriously considered for the c++ lib.

**5) MAVLINK_USE_CHAN_FUNCTIONS**

In some analogy to MAVLINK_USE_CONVENIENCE_FUNCTIONS I've added a MAVLINK_USE_CHAN_FUNCTIONS, which wraps all functions which use chan and/or the COMM buffers. 

It is defined when MAVLINK_COMM_NUM_BUFFERS is larger than zero. That way the user simply has to do a #define MAVLINK_COMM_NUM_BUFFERS 0 to ensure that none of all that is used. 

This is an essentially non-intrusive addition, since it normally doesn't affect any of the code.

**6) Formatting, few rearrangements, few renamings**

I've made plenty of changes which result in more properly formatted .h files. I've also rearranged some functions/structures to arrive at what I consider more logical and user friendly. For instance, I've moved the COMM related functions in mavlink_helpers.h to the top. Finally, I renamed some local variable names (only very few instances).

All of this is supposed to be totally non-intrusive, as it shouldn't affect the code behavior at all.

As said, I just share in case it might be of interest. Do what you want with it :).

Cheers, and as always, have fun, Olli


hamishwillee:
Thanks very much @olliw42 - really appreciate your contribution. Hopefully some or all of this can be integrated.

I don't have the skill to adequately evaluate this, so I've assigned to some others to check it out.

amilcarlucas:
This looks like it is mostly a pymavlink change, and therefore should be submitted upstream. Or am I missing something here?

hamishwillee:
@amilcarlucas No, they should be. This is one crap side effect of the fact that the generator is embedded in pymavlink rather than part of the main project.

@olliw42 Would it be possible to resubmit in https://github.com/ArduPilot/pymavlink ?
This is where the code that is impacted is defined. I could do this, but then of course all the discussion would come to me - which is not the right place.

olliw42:
it's probably true that this should go into pymavlink. But as said, I think I made a really serious effort to do things git-ish but failed. I of course could fork pymavlink, copy things into it, push and PR ... but I then would have no idea if that is functional code, couldn't do changes, couldn't test etc. ... unless I would spend much time with working around. Not my fault that git is as it is. Also, even if I would spend all that efforts, I'm not convinced it would be well spend since I doubt much or any of that ever will be merged. And for doing this, you to check, this here should be fully sufficient. I would be surprised if it would not just rot as so much else.

So, @hamishwillee, if you know how to do all that git thing, pl go ahead, I don't know it.

You folks know that (some of) you frequently suggest to submit code changes. And I agree it makes sense since then others could benefit. So I did.
But you also should realize that there is a limit to the burden you can expect people to take :).

have fun :)

julianoes:
@olliw42 first of all thanks for the pull request, very much appreciated!

I would be interested to see the actual changes as a diff. And I think the way to do that is to submit against https://github.com/ArduPilot/pymavlink. I'm happy to copy the files from this pull request to the other repository and create a pull request but as @hamishwillee said it would be a bit confusing for other readers where it comes from and also I don't want to "steal" your contribution but have the attribution correct.

> Also, even if I would spend all that efforts, I'm not convinced it would be well spend since I doubt much or any of that ever will be merged. And for doing this, you to check, this here should be fully sufficient. I would be surprised if it would not just rot as so much else.

I know what you mean and I see how that can be frustrating from your perspective. But you also have to understand that it is a lot of work to be a maintainer of open source projects and it's hard to follow up everywhere. That being said, from my experience, a contribution is much more likely to rot if it is just put out there but not followed up. Getting the contribution out there is one part, the other is to then work with the maintainer or community to see if and how it fits and get it in a merge-able shape.
This means I do appreciate that you published your changes in the first place but 

>  But you also should realize that there is a limit to the burden you can expect people to take :).

Yes, and that goes both ways.

amilcarlucas:
Done. See ardupilot/pymavlink#312 I will close this now, further comments should continue on that PR.

julianoes:
Thanks @amilcarlucas! Looks like you got attribution figured out :)

olliw42:
many thx, guys :)

Issue #1117, opened 2019-04-18
[Documentation] of SYS_STATUS


tuloski:
In [SYS_STATUS](https://mavlink.io/en/messages/common.html#SYS_STATUS) the field `onboard_control_sensors_health` says that 0 means disabled and 1 enabled, but probably it should be 0 --> healthy, 1-->broken (or vice versa). It's a copy paste error IMO.

And `current_battery` units is mA not cA.

hamishwillee:
I've fixed the onboard_control_sensors_health as you suggested - thanks.

> And current_battery units is mA not cA.

Really? PX4 assumes it is cA: https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_messages.cpp#L592
What am I missing?

tuloski:
Oh ok...cA? weird unit :)

PR #1128, opened 2019-05-02
SYSTEM_STATUS.onboard_control_sensors_health typo

hamishwillee:
As documented in #1117 the doc for `SYSTEM_STATUS.onboard_control_sensors_health` has a copy paste error. This makes it clear what 0 and 1 values mean for each sensor.


message_definitions/v1.0/common.xml
```diff
@@ -3225,7 +3225,7 @@
...
       <field type="uint32_t" name="onboard_control_sensors_present" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are present. Value of 0: not present. Value of 1: present.</field>
       <field type="uint32_t" name="onboard_control_sensors_enabled" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are enabled:  Value of 0: not enabled. Value of 1: enabled.</field>
-      <field type="uint32_t" name="onboard_control_sensors_health" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are operational or have an error:  Value of 0: not enabled. Value of 1: enabled.</field>
+      <field type="uint32_t" name="onboard_control_sensors_health" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are operational or have an error:  Value of 0: healthy. Value of 1: error.</field>
```
julianoes: Hm, are you sure it's this way round? Is this already used somewhere?

WickedShell:
This is wrong. 1 is healthy.

message_definitions/v1.0/common.xml
hamishwillee: Good point https://github.com/PX4/Firmware/blob/master/src/modules/commander/health_flag_helper.cpp#L62

Will change

PR #1130, opened 2019-05-06
Deprecate DO_MOUNT_CONTROL_QUAT

julianoes:
The command `DO_MOUNT_CONTROL_QUAT` was added to the spec without need and presumably without ever being implemented by neither gimbal manufacturers nor autopilot software stacks.

In my opinion it is not needed and only creates confusion.

@LorenzMeier you added this in 2014 in this commit, do you have an opinion?
https://github.com/mavlink/mavlink/commit/7d527ff474dd7e4f0ccb9982498e7954e3190597

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1391,7 +1391,8 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
-        <description>Mission command to control a camera or antenna mount, using a quaternion as reference.</description>
+        <deprecated since="2019-05" replaced_by="MAV_CMD_DO_MOUNT_CONTROL"/>
+        <description>(Deprecated) mission command to control a camera or antenna mount, using a quaternion as reference.</description>
         <param index="1" label="Q1">quaternion param q1, w (1 in null-rotation)</param>
         <param index="2" label="Q2">quaternion param q2, x (0 in null-rotation)</param>
         <param index="3" label="Q3">quaternion param q3, y (0 in null-rotation)</param>


olliw42:
Even though a bit off topic, I'd like to grab the opportunity to remind that the definition of the remaining (MAV_CMD_) DO_MOUNT_CONTROL is imprecise as regards to what the angles are supposed to mean. 

In fact, there had been significant confusion in the past, and may have lead to potentially "incorrect" implementations in e.g. ArduPilot for historical reasons. 

Correcting the situation by making the definition precise might lead to compatibility issues with legacy stuff, but that's how it would be, I guess.

have fun, Olli

hamishwillee:
I added dev call to discuss whether we can remove this because I'd like input from more than just those who have commented. 

Regarding changing MAV_CMD_DO_MOUNT_CONTROL (really any change to an existing non_WIP message) I'd suggest creating a PR with your suggested changes for discussion.

julianoes:
> absolute yaw is difficult if not impossible to achieve for a typical gimbal (they would need a full-fledged AHRS by themselves or obtain info from a full-fledged AHRS)

The gimbals I have worked with are fully capable of doing this using attitude estimation and by consuming yaw and position data of the aircraft for reference.

olliw42:
exactly as I've said, if the gimbal has either a full-fledged AHRS itself or gets such data from the aircraft, it's no problem, and you seem to explicitely confirm that. However, a "typical gimbal" to me is a gimbal which only uses gyros & accelerometers, and in that case it can't do it. This set of sensors is perfectly sufficient for doing the attitude estimation as the gimbal needs it for it's main job, to stabilize the camera. Note that this comprises most "lower-end" gimbals such as the feyus, tarots, zihuns, storm32s, etc. pp.. The STorM32 for instance can be connected to the aircraft's flight controller and can get that info, and in that case it indeed can do absolute yaw. But only then. Feyus, tarots, zihuns, etc don't have that ability.
(out of the box, some few seem to use Alexmos/Basecam, so might be modded, but...)

So, my conclusion is that when you say "The gimbals I have worked with are fully capable of doing this" you had been or are in a very fortunate situation :)

Fundamentally, a gimbal doesn't need magnetometers, gpses, and so on to do it's main job, to stabilize a camera.

I think however that generally can ask, what's the whole purpose if? 
Procedure A: The aircraft sends the gimbal a relative yaw which the aircraft has calculated taking its own absolute yaw and the desired yaw into account
Procedure B: The aircraft sends the gimbal the aircrafts own absolute yaw so that the gimbal can calculate the relative yaw from the desired absolute yaw and the aircrafts absolute yaw

except of that procedure B is just more complicate I can't see any disadvantage of A ;)

my 2 cents

julianoes:
You're right we should support both. And that is usually done with the `stabilize_x` flag. However, we should probably improve the comment around that to make it clear.

olliw42:
the "typical gimbal" I'm talking about can well be 3 axis gimbals (and indeed very nicely so). with respect to the stabilize_x flags there is no difference

you probably have meant the x_input flags, and here the "typical gimbal" indeed has limitations in that it could not do yaw_input = 2
(in most cases it also couldn't do pitch/roll_yaw input = 1 since they typically only provide angle inputs)

please note that there might be significant inconveniences from using yaw_input = 2. I think, and I understand you to agree, yaw_input = 0 should be supported in any case, for downwards compatibility. That means that if one also wants to support yaw_input = 2 that there must be some sort of mechanism to decide between the two (in the simplest case a user parameter, but other mechanisms are of course possible too). Implementing and keeping that working might add inconvenience, depending on how complicated/automated/user-friendly that mechanism is. And as said, I don't have yet seen any argument in favor of procedure (B) (except maybe that on some drawing boards it looks much cooler) . So, I would argue, all wasted time since it just adds workload without any real-world benefit (a cool drawing on some board I do not count as real-world benefit). So, I think the underlying question is: Procedure B, does it make anything easier or better?

as always just my 2 cents :)

julianoes:
> you probably have meant the x_input flags, and here the "typical gimbal" indeed has limitations in that it could not do yaw_input = 2
(in most cases it also couldn't do pitch/roll_yaw input = 1 since they typically only provide angle inputs)

Right, as you wrote above yaw in absolute frame is not possible.

What I meant with `stabilize_x` in `MAV_CMD_DO_MOUNT_CONFIGURE` is that if you set `stabilize_yaw` to 1, then it subtracts the vehicle's yaw and therefore basically achieves the same as `yaw_input = 2`. You can see we had that discussion in https://github.com/mavlink/mavlink/pull/944. I'm a bit confused by all the options as well, and wonder how it could be made clearer.

olliw42:
I see what you're saying, but I unfortunately can't agree. With respect to angles the relation is

yaw_absolute = yaw_vehicle + yaw_relative

therefore with respect to speed the relation is 

d/dt yaw_absolute = d/dt yaw_vehicle + d/dt yaw_relative

Now, your statement would be correct if d/dt yaw_vehicle would be zero, or always could be assumed as zero or as negligibly small. In general you can't, however, and especially during copter turns funny behavior could emerge, also wind could lead to funny behavior, and so on. So, also for stabilize = 1 it makes sense to distinguish between absolute and relative.
(it might well be that a gimbal controller might have a hard time to achieve the one or the other, but this has nothing to do with the message and it's fields)

Note: I obviously have used a linearized language and not expressed it in terms of the non-linear rotations as it would be more correct, but the point is clear, and would be equally so in a proper expression. Similarly, the point applies not only to yaw but all axes.

IMHO, the description for MAV_CMD_DO_MOUNT_CONFIGURE is quite clear. The description of MAV_CMD_DO_MOUNT_CONTROL in contrast could need clarification.
(besides the sign issue, also a statement like "pitch depending on mount mode" is dodgy, gladly it is explained in the bracket, yet "pitch depending on mount configuration" or similar would be better, a bit nitpicking but the sum of such blurrinesses can make things confusing)

julianoes:
> Now, your statement would be correct if d/dt yaw_vehicle would be zero, or always could be assumed as zero or as negligibly small.

You're right, and I have never really tried that in practice.

> (besides the sign issue, also a statement like "pitch depending on mount mode" is dodgy, gladly it is explained in the bracket, yet "pitch depending on mount configuration" or similar would be better, a bit nitpicking but the sum of such blurrinesses can make things confusing)

Thanks. I'll try to make a PR to fix this as well as the sign.

julianoes:
> (besides the sign issue, also a statement like "pitch depending on mount mode" is dodgy, gladly it is explained in the bracket, yet "pitch depending on mount configuration" or similar would be better, a bit nitpicking but the sum of such blurrinesses can make things confusing)

But `param7` is really called `Mount mode.`, so I'm not sure how that is confusing.

julianoes:
@olliw42 https://github.com/mavlink/mavlink/pull/1135

olliw42:
> But param7 is really called Mount mode., so I'm not sure how that is confusing.

yes, sure, but in some way it actually does depend more on e.g stabilize than on mount mode. In all mount cases pitch is pitch and always has the same meaning, while the MOUNT_CONFIGURATION fields (referred to in the bracket) in fact affect the INTERPRETATION of the value of the field. So, it would be fair to say that if there is anything the gimbal needs to know then it's the configuration settings, and the mode only in second place. 
Now, it might be that you want the interpretation of the angle fields to depend on the mode in a significant way. For instance that in RETRACT the angle fields should be ignored and the gimbal decides what to do. Or you could want that in GPS_POINT the angle fields to not be ignored but mean a speed with which the gimbal is trying to reach the POI if input = 1. And so on. In such cases there would indeed be a significant dependence on the mode. BUT ALL THAT WOULD HAVE TO BE SPELLED OUT AND DEFINED in a way the reader can understand.
With the description as is one needs to use common sense to determine what most likely is meant. :) That's all what I'm saying :)

julianoes:
> BUT ALL THAT WOULD HAVE TO BE SPELLED OUT AND DEFINED in a way the reader can understand.

Ok agreed. Let's try to make that happen then.

julianoes:
@olliw42 ok I should come back to this, sorry for the silence. Is there a chance I could talk to you 1:1 about gimbal messages some time to figure this out? If yes, send me an email to myfirstname@oes.ch, thx!

olliw42:
@julianoes sure, if you could send me a phone number to waldmanns (at) gmx (dot) de, I should have time this afternoon

olliw42:
@julianoes 
very nice discussion today, thx. Not sure it's appropriately placed here, but I'd like to write down, and if it's just to help my memory, what I think conclusions were

* clearly separate communication gcs-autopilot and autopilot-gimbal, both by logic and by different command/message sets. A GCS is not supposed to send control commands/messages to a gimbal.

what we forget a bit to discuss is the reverse, messages from the gimbal. I suggest that they also mostly should be restricted to autopilot-gimbal, e.g. all those which are control/mode related. Exceptions would be info messages which don't imply a action or response, and which often would be broadcast anyhow; these could/should be routed by the autopilot and thus possibly reach a GCS. Typical example: mount_status, mount_orientation.

* implicit to this clear separation is that modes such as RC targeting, MAVLink targeting, GPS/ROI targeting would normally be the autopilot's job. The command/message set of course should be rich/flexible enough to also support different scenarios (such as GPS/ROI handling by the gimbal, and so).

* there of course needs to be a mechanism to set the modes, but how to cast in messages was left open

* new set of commands/messages for the channel autopilot-gimbal

* also new set of messages/commands for the channel gcs-autopilot, to weed out the current strangenesses and also for strategic reason to reach a stable and consistent standard. Gimbals should ignore them and ideally never see them.

* existing DO_MOUNT commands/messages: document status quo, but mark as to not be used in new design. Docu should promote new sets and discourage using the old set (it simply should be clear what the standard is)

Comment: I promised to look up what ArduPilot is doing, i.e., if it digests the mode field in DO_MOUNT_CONTROL and if DO_MOUNT_CONFIGURE could be killed: You are right, I'm wrong, ArduPilot does correctly handle param7: https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mount/AP_Mount.cpp#L576-L586

* all angles, not only pitch&yaw (we identified at least two uses cases for roll control)

* quaternions vs Euler: That's a practical but also a strategic discussion which you guys want to discuss carefully in order to get it right for the next decade

* new commands/messages shouldn't be too big, logically distinct data shouldn't be bundled together, e.g. angles and location don't belong into one.

* the angle-related command/message(s) should allow for both angles AND rates

* relative yaw and absolute yaw: conclusion was that both are needed, depending on the use case/situation, but how to cast it in messages was left open

* basic gimbals only understand relative yaw

* targeting/pointing precision is mostly determined by whether gimbal is in follow/pan mode or hold mode, not by use of relative or absolute yaw. For gimbals in follow/yaw mode using relative yaw or absolute yaw are equally good. Using absolute yaw however implies a gimbal understanding absolute yaw. Relative yaw, i.e. having the autopilot doing the calculus, is thus more generic. For gimbals in hold mode, which implies that it does understand absolute yaw, it is more logical/practical to use absolute yaw, 

I mentioned, and I like to reinforce that point, that so far (to the best of my knowledge), targeting/pointing is done with gimbals in follow/pan mode, but much better precision - and thus potentially all sorts of new applications! - would be possible with them in hold mode, and I'd like to encourage you guys to actively push forward. Especially in combination with an angle message which also includes rates.

To cast it in buzz: It would make gimbal control much more robotic.

* commands vs messages: from my outsider's/bystander's view a consistent logic for command or message isn't apparent, and I suggested to clarify that. To me it could make sense to have both commands and messages 

(in my, potentially simplistic interpretation: Commands can be used in missions and through ACK ensure reliable message transport. Messages are best effort, are potentially smaller, and also easier to expand. Messages can include a target, no difference here. So, commands and messages seem to serve very different purposes and I don't see that one can be generally substituted by the other)

* MAV_TYPE vs componentID for component identification: agreement was reached that the component is identified via MAV_TYPE (it thus must be set properly mandatorily), while the specification for the componentID values are the suggested default values. 
There would have to be a means to allow more specific identification also for components, e.g., GIMBAL_STORM32, GIMBAL_ALEXMOS, GIMBAL_GREEMSY, QGROUNDCONTROL, MISSIONPLANNER, ... (similar to autopilots for which also  different implementations can be identified)

Comment: Currently MAV_TYPE serves more like a super class, while MAV_AUTPILOT identifies the specific implemntation. If fear is that the 255 ranges may be too quickly exhausted, one could do some sort of mapping, i.e. the autopilot field is filled with MAV_AUTOPILOT_XXX when it's of MAV_TYPE autopilot, but with e.g. MAV_GIMBAL_XXX when it's of MAV_TYPE gimbal, with e.g. MAV_CAMERA_XXX when it's of MAV_TYPE camera, and so on.

I've probably forgot some relevant points which we discussed or touched.

:)

julianoes:
Thanks a lot @olliw42. Your notes are a lot more complete than what I wrote down. I agree with everything except:


>commands vs messages: from my outsider's/bystander's view a consistent logic for command or message isn't apparent, and I suggested to clarify that. To me it could make sense to have both commands and messages

I'd vote for commands for GCS -> autopilot (or as part of missions) and messages for autopilot <-> gimbal.

> MAV_TYPE vs componentID for component identification: agreement was reached that the component is identified via MAV_TYPE (it thus must be set properly mandatorily), while the specification for the componentID values are the suggested default values.
There would have to be a means to allow more specific identification also for components, e.g., GIMBAL_STORM32, GIMBAL_ALEXMOS, GIMBAL_GREEMSY, QGROUNDCONTROL, MISSIONPLANNER, ... (similar to autopilots for which also different implementations can be identified)

I don't think it makes sense to add the specific type/brand/vendor into the component ID. As you said someone might want two gimbals on one drone, and the specific gimbal component would not solve the problem if the two gimbals have the same component id because they are the same or similar. The component id is something that an integrator has to take care of manually beforehand or the components have to do some auto-setup to resolve conflicts. E.g. a gimbal could listen for other heartbeats and choose something else. The only way around such conflicts AFAIK is probably a big enough id (such as a MAC adress) to prevent collisions.

olliw42:
> I'd vote for commands for GCS -> autopilot (or as part of missions) and messages for autopilot <-> gimbal.

good point, makes sense

> In this case the component info with a vendor and model string would be the way to go, I'd say.

this then wouldn't be in heartbeat I guess
my thoughts aimed at extending in some way what's already there, i.e., to do what is already there for autopilots  - which can be identified down to the implementation !! - also for components. The heartbeat allows to tell if it's a PX4 or ArduPilot, but not if it's a GQC or MP.
I understand that there are cases where there wouldn't be a clear distinction between "implementation" and "vendor" like it is for PX4 and ArduPilot (which are not vendors, obviously), and that in some cases the identification label would be identical to the vendor (like what I've mention for the gimbals), but I note that this distinction is also already blurred for the autopilots ... I mean, MAV_AUTOPILOT_ARMAZILA looks equally vendor-ish to me than MAV_GIMBAL_GREEMSY. I'm not sure this justifies adding the additional layer of request-response communication required for a vendor/model message exchange. Enums are simple and compact, and I don't think there will be 255 mavlink gimbal vendor/models anytime soon ...

however, that's certainly not the most pressing point and we haven't discussed that in much detail, so it's sort of my soup, I probably shouldn't have added that part. What would be more important is to resolve the inconsistency as regards MAV_TYPE and the componentID, and I think we had been in agreement here.

:)

julianoes:
Closing in favor of #1174.

PR #1135, opened 2019-05-13
Document signs for DO_MOUNT_CONTROL

julianoes:
It was not specified what the signs of the gimbal attitude controls were. This is an attempt to document it. The angles should be consistent with the setpoints used for aircraft in NED as common in MAVLink.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -700,21 +700,21 @@
     </enum>
     <!-- Camera Mount mode enumeration -->
     <enum name="MAV_MOUNT_MODE">
-      <description>Enumeration of possible mount operation modes</description>
+      <description>Enumeration of camera mount operation modes.</description>
       <entry value="0" name="MAV_MOUNT_MODE_RETRACT">
-        <description>Load and keep safe position (Roll, Pitch, Yaw) from permanent memory and stop stabilization</description>
+        <description>Retract to safe position and stop stabilization. Use Roll, Pitch, Yaw from MAV_CMD_DO_MOUNT_CONTROL message (if NaN use gimbal defaults).</description>
       </entry>
       <entry value="1" name="MAV_MOUNT_MODE_NEUTRAL">
-        <description>Load and keep neutral position (Roll, Pitch, Yaw) from permanent memory</description>
+        <description>Move to neutral position and stop stabilization. Use Roll, Pitch, Yaw from MAV_CMD_DO_MOUNT_CONTROL message (if NaN use gimbal defaults)</description>
       </entry>
       <entry value="2" name="MAV_MOUNT_MODE_MAVLINK_TARGETING">
         <description>Load neutral position and start MAVLink Roll, Pitch, Yaw control with stabilization</description>
       </entry>
       <entry value="3" name="MAV_MOUNT_MODE_RC_TARGETING">
-        <description>Load neutral position and start RC Roll, Pitch, Yaw control with stabilization</description>
+        <description>Move to neutral position and start RC Roll, Pitch, Yaw control with stabilization</description>
       </entry>
       <entry value="4" name="MAV_MOUNT_MODE_GPS_POINT">
-        <description>Load neutral position and start to point to Lat, Lon, Alt</description>
+        <description>Move to neutral position and start to point to Lat, Lon, Alt</description>
       </entry>
     </enum>
     <!-- UAVCAN node health enumeration -->
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1311,9 +1311,9 @@
       <!-- this one is messed up! altitude should be param 7, not param4 -->
       <entry value="205" name="MAV_CMD_DO_MOUNT_CONTROL" hasLocation="false" isDestination="false">
         <description>Mission command to control a camera or antenna mount</description>
-        <param index="1">pitch depending on mount mode (degrees or degrees/second depending on pitch input).</param>
-        <param index="2">roll depending on mount mode (degrees or degrees/second depending on roll input).</param>
-        <param index="3">yaw depending on mount mode (degrees or degrees/second depending on yaw input).</param>
+        <param index="1">pitch depending on mount mode (degrees or degrees/second depending on pitch input, positive to point up, negative to point down).</param>
+        <param index="2">roll depending on mount mode (degrees or degrees/second depending on roll input, positive to tilt to the right).</param>
+        <param index="3">yaw depending on mount mode (degrees or degrees/second depending on yaw input, positive to turn to the right).</param>
         <param index="4" label="Altitude" units="m">altitude depending on mount mode.</param>
         <param index="5">latitude in degrees * 1E7, set if appropriate mount mode.</param>
         <param index="6">longitude in degrees * 1E7, set if appropriate mount mode.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1310,13 +1310,13 @@
       </entry>
       <!-- this one is messed up! altitude should be param 7, not param4 -->
       <entry value="205" name="MAV_CMD_DO_MOUNT_CONTROL" hasLocation="false" isDestination="false">
-        <description>Mission command to control a camera or antenna mount</description>
-        <param index="1">Pitch in degrees or degrees/second depending on pitch input (positive to point up, negative to point down), use according to mount mode.</param>
-        <param index="2">Roll in degrees or degrees/second depending on roll input (positive to tilt to the right), use according to mount mode.</param>
-        <param index="3">Yaw in degrees or degrees/second depending on yaw input (positive to turn to the right), use according to mount mode.</param>
-        <param index="4" label="Altitude" units="m">Altitude use according to mount mode.</param>
-        <param index="5">Latitude in degrees * 1E7, set if appropriate mount mode.</param>
-        <param index="6">Longitude in degrees * 1E7, set if appropriate mount mode.</param>
+        <description>Mission command to control a camera or antenna mount. How/whether params 1 to 6 are used is defined by the mount mode (param7). </description>
+        <param index="1" label="Pitch">Pitch (positive to point up, negative to point down). The units and frame depend on the mount configuration pitch input setting, MAV_CMD_MOUNT_CONFIGURE.param6: degrees for body and absolute angle frame, and degrees/second for angular rate.</param>
+        <param index="2" label="Roll">Roll (positive to tilt to the right, negative to tilt left). The units and frame depend on the mount configuration roll input setting, MAV_CMD_MOUNT_CONFIGURE.param5: degrees for body and absolute angle frame, and degrees/second for angular rate.</param>
+        <param index="3" label="Yaw">Yaw (positive to turn to the right, negative to turn left). The units and frame depend on the mount configuration roll input setting, MAV_CMD_MOUNT_CONFIGURE.param7: degrees for body and absolute angle frame, and degrees/second for angular rate.</param>
+        <param index="4" label="Altitude" units="m">Altitude.</param>
+        <param index="5" label="Latitude" units="degE7">Latitude.</param>
+        <param index="6"  label="Longitude" units="degE7">Longitude.</param>
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -698,23 +698,23 @@
         <description>Breached fence boundary</description>
       </entry>
     </enum>
-    <!-- Camera Mount mode Enumeration -->
+    <!-- Camera Mount mode enumeration -->
     <enum name="MAV_MOUNT_MODE">
       <description>Enumeration of possible mount operation modes</description>
       <entry value="0" name="MAV_MOUNT_MODE_RETRACT">
-        <description>Load and keep safe position (Roll,Pitch,Yaw) from permant memory and stop stabilization</description>
+        <description>Load and keep safe position (Roll, Pitch, Yaw) from permanent memory and stop stabilization</description>
       </entry>
       <entry value="1" name="MAV_MOUNT_MODE_NEUTRAL">
-        <description>Load and keep neutral position (Roll,Pitch,Yaw) from permanent memory.</description>
+        <description>Load and keep neutral position (Roll, Pitch, Yaw) from permanent memory</description>
       </entry>
       <entry value="2" name="MAV_MOUNT_MODE_MAVLINK_TARGETING">
-        <description>Load neutral position and start MAVLink Roll,Pitch,Yaw control with stabilization</description>
+        <description>Load neutral position and start MAVLink Roll, Pitch, Yaw control with stabilization</description>
       </entry>
       <entry value="3" name="MAV_MOUNT_MODE_RC_TARGETING">
-        <description>Load neutral position and start RC Roll,Pitch,Yaw control with stabilization</description>
+        <description>Load neutral position and start RC Roll, Pitch, Yaw control with stabilization</description>
       </entry>
       <entry value="4" name="MAV_MOUNT_MODE_GPS_POINT">
-        <description>Load neutral position and start to point to Lat,Lon,Alt</description>
+        <description>Load neutral position and start to point to Lat, Lon, Alt</description>
       </entry>
     </enum>
     <!-- UAVCAN node health enumeration -->
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1311,12 +1311,12 @@
       <!-- this one is messed up! altitude should be param 7, not param4 -->
       <entry value="205" name="MAV_CMD_DO_MOUNT_CONTROL" hasLocation="false" isDestination="false">
         <description>Mission command to control a camera or antenna mount</description>
-        <param index="1">pitch depending on mount mode (degrees or degrees/second depending on pitch input, positive to point up, negative to point down).</param>
-        <param index="2">roll depending on mount mode (degrees or degrees/second depending on roll input, positive to tilt to the right).</param>
-        <param index="3">yaw depending on mount mode (degrees or degrees/second depending on yaw input, positive to turn to the right).</param>
-        <param index="4" label="Altitude" units="m">altitude depending on mount mode.</param>
-        <param index="5">latitude in degrees * 1E7, set if appropriate mount mode.</param>
-        <param index="6">longitude in degrees * 1E7, set if appropriate mount mode.</param>
+        <param index="1">Pitch in degrees or degrees/second depending on pitch input (positive to point up, negative to point down), use according to mount mode.</param>
+        <param index="2">Roll in degrees or degrees/second depending on roll input (positive to tilt to the right), use according to mount mode.</param>
+        <param index="3">Yaw in degrees or degrees/second depending on yaw input (positive to turn to the right), use according to mount mode.</param>
+        <param index="4" label="Altitude" units="m">Altitude use according to mount mode.</param>
+        <param index="5">Latitude in degrees * 1E7, set if appropriate mount mode.</param>
+        <param index="6">Longitude in degrees * 1E7, set if appropriate mount mode.</param>
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">


olliw42:
not claiming to be perfect, but this might give ideas

```
pitch in degrees or degrees/second depending on pitch input (positive to point up, negative to point down), use according to mount mode.
roll in degrees or degrees/second depending on roll input (positive to tilt to the right), use according to mount mode.
yaw in degrees or degrees/second depending on yaw input (positive to turn to the right), use according to mount mode.
altitude, use according to mount mode.
latitude in degrees * 1E7, set if appropriate mount mode.
longitude in degrees * 1E7, set if appropriate mount mode.

```

julianoes:
Thanks @olliw42 that's good, I can't think of something better. Maybe @hamishwillee has some better idea.

message_definitions/v1.0/common.xml
```diff
@@ -1311,12 +1311,12 @@
       <!-- this one is messed up! altitude should be param 7, not param4 -->
       <entry value="205" name="MAV_CMD_DO_MOUNT_CONTROL" hasLocation="false" isDestination="false">
         <description>Mission command to control a camera or antenna mount</description>
-        <param index="1">pitch depending on mount mode (degrees or degrees/second depending on pitch input).</param>
-        <param index="2">roll depending on mount mode (degrees or degrees/second depending on roll input).</param>
-        <param index="3">yaw depending on mount mode (degrees or degrees/second depending on yaw input).</param>
-        <param index="4" label="Altitude" units="m">altitude depending on mount mode.</param>
-        <param index="5">latitude in degrees * 1E7, set if appropriate mount mode.</param>
-        <param index="6">longitude in degrees * 1E7, set if appropriate mount mode.</param>
+        <param index="1">Pitch in degrees or degrees/second depending on pitch input (positive to point up, negative to point down), use according to mount mode.</param>
```
hamishwillee: FWIW it is irritating to have params/fields that can take multiple types of units, because then you can't use the units="" attribute.

message_definitions/v1.0/common.xml
```diff
@@ -1311,12 +1311,12 @@
...
+        <param index="2">Roll in degrees or degrees/second depending on roll input (positive to tilt to the right), use according to mount mode.</param>
+        <param index="3">Yaw in degrees or degrees/second depending on yaw input (positive to turn to the right), use according to mount mode.</param>
+        <param index="4" label="Altitude" units="m">Altitude use according to mount mode.</param>
```
hamishwillee: 4, 5, 6 just set when using MAV_MOUNT_MODE_GPS_POINT right?
And the 1, 2, 3 (roll, pitch yaw) used in the other modes right?


 right?

hamishwillee:
Consider:

> Roll in degrees or degrees/second depending on roll input (positive to tilt to the right), use according to mount mode.

1. What you mean by "depending on roll input" ? 
2. "use according to mount mode." - Does this *really* mean: "Use this param if  mode is anything other than [MAV_MOUNT_MODE_GPS_POINT](https://mavlink.io/en/messages/common.html#MAV_MOUNT_MODE_GPS_POINT), otherwise if it is `MAV_MOUNT_MODE_GPS_POINT` ignore the param?

Can't finish until I understand the above, but might word something like:

> Roll in degrees or degrees/second depending on roll input (positive: tilt clockwise, negative: counter-clockwise).  Used in mount modes that depend on roll and ignored otherwise.

olliw42:
thx to look that critically at this, I really appreciate this and the fine differences of the meaning of words can make a big difference
I'm not a native English speaker so these fine differences I might not catch

1. good catch. I guess it means "Roll in degrees or degrees/second as specified by the "roll input" field in MAV_CMD_MOUNT_CONFIGURE".

2. This is actually a part I puzzled myself a bit. It's not clear in which circumstances pitch,roll,yaw are supposed to be use. E.g. in RC_RETRACT I could see several scenarios (i) the gimbal uses the angles as specified in the message (which when must be degrees), (ii) it uses a position which is hardcoded into the gimbal, i.e. ignores these values, (iii) it uses a position as set by some parameters in the flight controller, i.e. ignores these values, (iv) .... The problem: I don't think that one wants or is able to define this in such detail at this point in time (if at all in some point in time), i.e., I guess the statement will end up to be somewhat unclear as regards to what precisely the usage is. 
I btw also was not sure if the subtle but significant difference between "use according to mount mode" and "set if appropriate mount mode" catches things correctly. E.g. for altitude, should it be also "set if appropriate mount mode"?

difficult to be precise in few words

olliw42:
I have seen that some issues got DevCall labels. I'd like to remind that the explicit specification of the signs of the angles very likely will have some impact in that some legacy hardware might not work properly anymore. It will affect ArduPilot users, I would guess that this also might affect users of the Basecam/Alexmos gimbal controllers, etc. pp. So maybe you want to discuss implications at a devcall too.

hamishwillee:
Thanks for that. Being terse is desirable, but second priority to being clear and accurate. Ie we can write as much as we need to here.

Yes the mount modes are poorly worded too. In particular MAV_MOUNT_MODE_RETRACT does seem to imply (via use of "permanent memory") that the safe retraction position is hardcoded/in device, which in turn implies that all of the other params are probably ignored. 

And yes, we will need to discuss in dev meeting. But we should get this as close to what we think reasonable as possible before the meeting. I'd rather write this "as I think it should be" and try get agreement from the dev team.

For now, proposed doc on roll (others similar pattern):

> Roll (positive: clockwise, negative: counter-clockwise). The units and frame of the roll depend on the mount configuration roll input setting (MAV_CMD_MOUNT_CONFIGURE.param5) - i.e. degrees if roll input is the body frame or absolute angle frame, and degrees/second if the roll input is angular rage.
  The roll is used as defined in by the mount mode (param7).

And for MAV_MOUNT_MODE_RETRACT I'd do:

> Retract mount to safe position and stop stabilization. Angle/position information in MAV_CMD_DO_MOUNT_CONTROL messages is ignored.

message_definitions/v1.0/common.xml
julianoes: Right but I don't know how we could improve that for this case.

message_definitions/v1.0/common.xml
julianoes: Right.

olliw42:
> we should get this as close to what we think reasonable as possible before the meeting. 

fair enough

That's super precise. Maybe one could slim down a bit:
>Roll (XXX). The units and frame depend on the mount configuration roll input setting, MAV_CMD_MOUNT_CONFIGURE.param5: degrees for body and absolute angle frame, and degrees/second for angular rate. The roll is used as defined in by the mount mode (param7).

as regards the XXX, I liked the wording of julianoes MUCH better. 
I remember that in the early days of the STorM32 there had been massive user confusion about what clockwise means. It appeared that many more people were happy however with "tilt to the left/right", maybe because it's more along the speak of flying vehicles. I changed language and the confusion went away. Similar for yaw. If there would be an election I would vote for the exact wording of julianoes :)

I'm surely overly pedantic but to just have mentioned it, MAV_MOUNT_MODE_NEUTRAL should be changed similarly to retract.

@hamishwillee : on a totally unrelated matter, but I really don't know how to contact you or any of the dronecode discuss admins: Since few weeks I can't log in to dronecode discuss anymore, I always get a window saying "Sorry, there was an error authorizing your account. Perhaps you did not approve authorization?" (I log in with Github). I also can't open a new account to log in per email, since my email is already registered. And the ContactUs link on the About page seems to be a dead end. So, I misuse this place here out of a desperation. Sorry for that. But could you/anyone have a look at this and tell me what I would need to do to get my account working again (my displayed user name there is olliw). Thx, and sorry.

julianoes:
> Since few weeks I can't log in to dronecode discuss anymore, I always get a window saying "Sorry, there was an error authorizing your account. Perhaps you did not approve authorization?" (I log in with Github). 

Thanks for telling me. I added letsencrypt to the discuss forum and the URLs are now wrong/different which is why it stopped working. I'm aware of this but I need @LorenzMeier's help due to permissions. I'm on it.

olliw42:
oh, very quick response, thx soo much. Thx for the clarification, so I will just patiently wait :). I'll also delete that part of the text in my previous post, to keep things cleaner (pl feel free to also delete this one here if that is possible :))

message_definitions/v1.0/common.xml
hamishwillee: We can't really.

hamishwillee:
I'm happy with the slimmed version, and I'm happy enough with tilt too - would be foolish to ignore your experience of what works. So proposal is:

>Roll (positive to tilt to the right, negative to tilt left). The units and frame depend on the mount configuration roll input setting, MAV_CMD_MOUNT_CONFIGURE.param5: degrees for body and absolute angle frame, and degrees/second for angular rate. The roll is used as defined by the mount mode (param7).

OK?

> Btw, you know that ArduPilot has MNT_RETRACT_X/Y/Z and MNT_NEUTRAL_X/Y/Z parameters, to set the gimbal position in case of RETRACT/NEUTRAL? I just mention as the proposed specification would make these params to not work, "force" ArduPilot to not comply with the spec (I just double checked their code, and their "MAVLink gimbal mount" aka MountSTorM32 is using these parameters), or requires them to do some changes.

No I didn't. Can you show me what command is being used in https://mavlink.io/en/messages/ardupilotmega.html

Without seeing that, I'd guess that you could either have a separate mount mode for retract where the x, y, z are obeyed OR you could have a not used indicator value like NAN for the roll, pitch, yaw. 

> I'm surely overly pedantic but to just have mentioned it, MAV_MOUNT_MODE_NEUTRAL should be changed similarly to retract.

OK, we should do a review of all of them. I didn't want to go far with these until you responded :-)

olliw42:
> OK?

fantastic :)

> No I didn't. Can you show me what command is being used in https://mavlink.io/en/messages/ardupilotmega.html

ähm, MAV_CMD_DO_MOUNT_CONTROL from commons.xml, it calls this 
https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mount/AP_Mount_SToRM32.cpp#L24-L70
where in lines 29 and 39 it asks for the parameter values

> Without seeing that, I'd guess that you could either have a separate mount mode for retract where the x, y, z are obeyed OR you could have a not used indicator value like NAN for the roll, pitch, yaw.

yeah, as mentioned before there could be several use cases which all might make sense. This are two, I think I saw three:
* ignore angles and use gimbal internals
* ignore angles and use some parameters which allow users to specify
* take the given angles

The simplest solution probably would be to REQUIRE that the angles are set to appropriate values even for the RETRACT and NEUTRAL modes, i.e., the sender should expect a mount to use these values. Could be NAN, 0, or whatever. The mount can then decide itself.

:)

julianoes:
I've looked at QGC and I found that the command `DO_MOUNT_CONTROL` requires manual input by a user, so it's up to the user to enter `-90` for the pitch, so it looks very much unspecified at this point:
Selection_025

The only reference in code I found was:
https://github.com/mavlink/qgroundcontrol/blob/8b3f37668f7b8066d2f9985d38b3cc64b7af0c51/src/MissionManager/MissionController.cc#L477-L480

@auturgy do you know any instances where this is already implemented in a certain way in code? E.g. in MissionPlanner?

julianoes:
offtopic: @olliw42 
> oh, very quick response, thx soo much. Thx for the clarification, so I will just patiently wait :).

It's fixed.

olliw42:
thx so much, sir :+1:

hamishwillee:
@olliw42 Re https://github.com/mavlink/mavlink/pull/1135#issuecomment-492527835

Yes, so I think you're taking the option that RETRACT and NEUTRAL will expect a mount to supply a value, and if that value is NAN it will use its own defaults. 

That is different than what the docs say, but I agree it makes sense. I am not yet sure how the whole update these messages or add new ones, but for the moment I'll update these in the PR so discussion not lost.

message_definitions/v1.0/common.xml
```diff
@@ -698,23 +698,23 @@
...
-        <description>Load and keep safe position (Roll,Pitch,Yaw) from permant memory and stop stabilization</description>
+        <description>Retract to safe position and stop stabilization. Use Roll, Pitch, Yaw from MAV_CMD_DO_MOUNT_CONTROL message (if NaN use gimbal defaults).</description>
       </entry>
       <entry value="1" name="MAV_MOUNT_MODE_NEUTRAL">
-        <description>Load and keep neutral position (Roll,Pitch,Yaw) from permanent memory.</description>
+        <description>Move to neutral position and stop stabilization. Use Roll, Pitch, Yaw from MAV_CMD_DO_MOUNT_CONTROL message (if NaN use gimbal defaults)</description>
       </entry>
       <entry value="2" name="MAV_MOUNT_MODE_MAVLINK_TARGETING">
-        <description>Load neutral position and start MAVLink Roll,Pitch,Yaw control with stabilization</description>
+        <description>Load neutral position and start MAVLink Roll, Pitch, Yaw control with stabilization</description>
```
hamishwillee: @olliw42 @julianoes 
1. What is difference between MAV_MOUNT_MODE_MAVLINK_TARGETING and MAV_MOUNT_MODE_RC_TARGETING - is it just that in the first the target comes from the control message, while in the second you're put into the right mode and yaw/pitch/roll come from the RC controller?
2. For MAV_MOUNT_MODE_MAVLINK_TARGETING , do we always *require* roll, pitch, yaw? Ie what if one of these was set as NaN. Is this like retract - assume there is a hardware default?

message_definitions/v1.0/common.xml
```diff
@@ -698,23 +698,23 @@
...
       </entry>
       <entry value="3" name="MAV_MOUNT_MODE_RC_TARGETING">
-        <description>Load neutral position and start RC Roll,Pitch,Yaw control with stabilization</description>
+        <description>Move to neutral position and start RC Roll, Pitch, Yaw control with stabilization</description>
       </entry>
       <entry value="4" name="MAV_MOUNT_MODE_GPS_POINT">
-        <description>Load neutral position and start to point to Lat,Lon,Alt</description>
+        <description>Move to neutral position and start to point to Lat, Lon, Alt</description>
```
hamishwillee: @julianoes @olliw42 What if Lat, Lon, Alt were NaN? My assumption is that we require that these all be set or return an error result?

hamishwillee:
Note, this is failing a format test. We can fix this when we're ready to merge.

message_definitions/v1.0/common.xml
julianoes: 1. That's also my guess. No idea.
2. Sounds reasonable.

message_definitions/v1.0/common.xml
julianoes: Correct, at least Lat/Lon. If we also want it to pitch accordingly we need to give it an Alt too.

olliw42:
> Yes, so I think you're taking the option that RETRACT and NEUTRAL will expect a mount to supply a value, and if that value is NAN it will use its own defaults.

well, I wouldn't say that it's an option I'm taking, I "just" try to describe and piece together, but yes, that's what it boils down to.

> That is different than what the docs say

yes. 

> but I agree it makes sense for the point of view of how the API is already used

yes. That's how it is used. Note that I'm not saying that therefore the spec MUST be adapted accordingly, the spec could and maybe should decide by itself what it thinks the correct usage should be.  

> also because even though I would expect a gimbal would know its retract position (or that a GCS shouldn't have to) some might not.

yes and no. For NEUTRAL one could argue it should know it itself. For RETRACT I wouldn't agree. The RETRACT operation typically would be accomplished by a piece of hardware which connects the gimbal to the vehicle, and thus wouldn't be something the gimbal would do itself, but would be controlled/operated by the vehicle. For RETRACT it therefore makes sense to me that one might not want the gimbal to sue whatever it thinks itself it should do. 

It's actually more complicated that: In my statement I was implicitly assuming that the retract mechanism is also controlled via the MOUNT message, i,.e. that the retract mechanism also monitors the MOUNT message and moves accordingly, or is controlled by something else which monitors the MOUNT message. However, one also could take the point that the position of the retract system should be controlled by a different or own set of messages, in which case RETRACT in MOUNT would mean something different to the gimbal. I'm assuming that the latter is not what the intention is. But what do I know.

Note also that there can be reasons/cases where one wants a gimbal to assume different positions as response to RETRACT, and that RETRACT might not always mean that a gimbal is moved inside and outside, but that it is put into a "safe" state. E.g. gimbals on planes often have the problem that the gimbal screws up during take off and landing because of the rattling/high accelerations. So one would want the gimbal to also "hard clamp" it's camera position (i.e. use a different motor control algorithm and work more like servos)(that's possible with encodered gimbals). The hard clamped position might  however be wanted to be different for takeoff and landing. And so on. 

I'm always for specs which - if possible (!) - are enablers and not restrictors, i.e. allow for also unforeseen but reasonable use cases. Along that line I think that one should allow the position to have some sense also in RETRACT (and then why not also in NEUTRAL). It is true that this implies that there might be additional parameters which the user needs to handle, like with ArduPilot, but that's maybe not a specs concern.

ups, long response for such a seemingly little topic :)

message_definitions/v1.0/common.xml
olliw42: These are actually two questions which are much more difficult to answer than it might seem, I thus put it in a post below.

message_definitions/v1.0/common.xml
olliw42: this IMHO also requiers more than a short sentence, hence also in an extra post below

olliw42:
> 1.  What is difference between MAV_MOUNT_MODE_MAVLINK_TARGETING and MAV_MOUNT_MODE_RC_TARGETING - is it just that in the first the target comes from the control message, while in the second you're put into the right mode and yaw/pitch/roll come from the RC controller?

yes, that's how it is used by ArduPilot. For MAVLINK_TARGETING it takes the angles from the message and sends them to the gimbal, and for RC_TARGETING it forwards the angles it is extracting from its RC/joystick system.

Note that there is a tiny but all crucial subtlety in here, there are two use cases for the MOUNT message:

It can be digested and operated on by the flightcontroller, and it can be digest and operated on by the gimbal. And issues obviously can emerge then this is not correctly "synchronized" and the gimbal gets conflicting information.
(I think that this is the main root cause for the issues with the Solo, there MOUNTs are coming from the companion computer as well as the cube, as well as to their fancy "stitching issue" they observe since Chibios).

Maybe I should explain by being specific.
Case A: a non-mavlink gimbal is connected to ArduPilot. That's simple, only ArduPilot digests the MOUNT message, and sends e.g. PWM data to the gimbal according to the angles. MAVLINK_TARGETING and RC_TARGETING work exactly as described in the above.
Case B: a mavlink speaking gimbal is attached to ArduPilot: Not simple. Then a MOUNT message reaches ArduPilot two things can happen, (1) the message can be digested by ArduPilot and a new MOUNT message being send out, (2) the message is routed forward to the gimbal, and is digesting by the gimbal. Note that depending on the targetIDs all sort of combinations of the two can happen, and that there is (at leaat one) targetIDs set which makes both (1) and (2) to happen for one and the same MOUNT message coming from a GCS/companion. 

The relevance to the above is that if one is talking about what MAVLINK_TARGETING and RC_TARGETING are doing, one FIRST needs to specify if one is talking about the flightcontroller or the gimbal as digester of the message. The answer can be different depending on which of the two one is talking about.

So, to be clear, in the above we were talking about the flightcontroller, about what ArduPilot is doing. We were not speaking about the gimbal. 

In the case of the STorM32, there is actually no difference between MAVLINK_TARGETING and RC_TARGETING, i.e., in both cases it uses the angles from the MOUNT message and orients the camera accordingly. For RC_TARGETING the STorM32 does NOT use its own RC/joystick input system to target the camera, even though this could be a reasonable response from the perspective of the gimbal too. I want to STRESS that I did this so only as a result of a best-guess, to get it all functioning.

One can imagine different ways of how all of this should work, some of which might even be better. All the above is just what has emerged as the result of history,

> 2.  For MAV_MOUNT_MODE_MAVLINK_TARGETING , do we always require roll, pitch, yaw? Ie what if one of these was set as NaN. Is this like retract - assume there is a hardware default?

I think this is actually a deeper question, related to the philosophy of how MAVLink wants to do some things. The point is, I'm always for following one and the same philosophy for as much as possible (which helps users because once they see the pattern they can assume that pattern, and also documenters since they don't have to describe things for each message but only once).

The "philosophy" I see is to whether use NANs,-1s, and such or to use enums/statusfields. Both have advantages and disadvantages, and IMHO are more about taste and preference, i.e. philosophy, and not hard arguments.

In this context I'd like to recall the recent PR "Adding an enum to storage info." #1137, in which the initial suggestion to decode states into NANs was replaced by an enum.  So, here the philosophy was to better use enums/status fields. One also could apply this philosophy to the case here, instead of NANs. 

What I'm trying to say: 
I think it would be highly useful, from the perspective of achieving a consistent, well designed and easy to use spec, if you would open up your mind and make a decision for either the NAN or enum philosophy, and henceforth would try to follow it as much as possible.

With that decision been taken, also 2. is largely answered

my 2 cents, even though they might be worth 0 :)

olliw42:
I guess I deserve a keyboard rest now LOL

julianoes:
That's very interesting points, thanks @olliw42.

Regarding `MAV_MOUNT_MODE_RC_TARGETING` vs `MAV_MOUNT_MODE_MAVLINK_TARGETING` the question is who mixes between RC and MAVLink commands. In the cases I've been involved we "mixed" the two in the autopilot since the autopilot had the RC input signals as well as the MAVLink commands either directly from a ground station or SDK or from an onboard mission.

The mixing was done in the way to always use the latest input which works fine for MAVLink commands which only tend to arrive every now and then but does not work for RC input which is always sent and published at some rate (e.g. 20 Hz). In this case we looked out for "considerable" changes in the RC gimbal channels to detect if a user was actually moving the knobs (more than noise or bumping) and therefore wanted the gimbal to follow. I think this was generally somewhat intuitive but can still lead to accidental gimbal movements during a mission when it shouldn't move.

hamishwillee:
Re https://github.com/mavlink/mavlink/pull/1135#issuecomment-492965722

> ups, long response for such a seemingly little topic :)

The only problem I have with such a long response is that I might well miss key points. Otherwise it is much appreciated - I have a lot to learn.

The specs are created with "best intent". But they might have not covered all cases, and they certainly didn't make it unambigous, or we wouldn't be having this discussion.

Good point re RETRACT "options". I think all we can say for sure is that they are intended to go to a gimbal and define safe positions. We can't say what the command is supposed to do for a mount that also retracts inside the body. How we approach this depends on whether we end up having new messages/what @julian decides. However my gut feeling is that we could easily add a new mode that also specified retraction into the vehicle and make it clear that this is for gimbal and mount, not just gimbal.

For now I believe the current set of docs for RETRACT and the control command support the current usage for PX4 and ArduPilot. They don't restrict any particular usage. The only thing I might change is "(if NaN use gimbal defaults)." TO "(if NaN use mount defaults)." - as this allows users to assume that retraction into vehicle is also something that might happen for the same command.

Hope that makes sense.

hamishwillee:
Regarding MAV_MOUNT_MODE_RC_TARGETING vs MAV_MOUNT_MODE_MAVLINK_TARGETING just to be very clear, I think you are saying that from the perspective of the usage/params they are the same. It is just that they indicate a message from a different source. It isn't obvious to me why the source isn't conveyed/managed by component ID.

Re use of NaNs or otherwise is an implementation detail. What I am asking is whether the message needs to support the concept of a default latitude, longitude, yaw, pitch, roll or that real values must always be sent.

> I think it would be highly useful, from the perspective of achieving a consistent, well designed and easy to use spec, if you guys would open up your mind and make a decision for either the NAN or enum philosophy, and henceforth would try to follow it as much as possible.

We use NaN here because it is the only number you can put into a float field (yaw say) that isn't a valid value and can hence be unambiguously used to mean "use a default for this value" (or in other cases "this value is not supplied by the sending system"). If the valid range for the field was positive numbers we might equally use -1. Ideally we use 0 if at all possible, because mavlink 2 message truncation can in some cases allow a shorter message to be sent over the wire.

I think you are suggesting that instead we might use a _separate_ field to indicate whether the yaw value field should be ignored/default values used? Or perhaps a param that is a bitmap to indicate a number of fields?

I think we are pretty consistent in the NaN, -1, 0 usage.

Personally I'd find using enums harder. I like being able to understand what the value of a field is in the field itself. I think we'd run into many cases where there wouldn't be spare fields for this enum/bitmap.

Enums are used for modes and other things. I don't particularly see the benefit in using them for this case, but perhaps I'm being closed minded. Maybe some examples?

olliw42:
@julianoes 
> Regarding MAV_MOUNT_MODE_RC_TARGETING vs MAV_MOUNT_MODE_MAVLINK_TARGETING the question is who mixes between RC and MAVLink commands.

YES :) 
and indeed, in some cool and desirable applications/situations the distinction between the two evaporates as a useful concept

@hamishwillee 

>The specs are created with "best intent". But they might have not covered all cases, and they certainly didn't make it unambiguous, or we wouldn't be having this discussion.

oh, I'm sorry. I know that by times I may come across as critical, but believe me it is not "critical" in the sense of criticism, but in the sense of "critical analysis". Personally I find them super useful since they really help me a lot, and that's my sole intention here.  

>For now I believe the current set of docs for RETRACT and the control command support the current usage for PX4 and ArduPilot. They don't restrict any particular usage. 

Yes, I largely agree! I don't agree in some details :)

Note: In the description of the yaw field in `MAV_CMD_DO_MOUNT_CONTROL` it should read "yaw input setting" and not "roll input setting".

Also, IMHO, in `MAV_MOUNT_MODE_NEUTRAL` it should be "Move to neutral position with stabilization".

> The only thing I might change is "(if NaN use gimbal defaults)." TO "(if NaN use mount defaults)." - as this allows users to assume that retraction into vehicle is also something that might happen for the same command.

It's here there we have a different view. IMHO a mount should be allowed to use its defaults also if angles are set.  It's not a great sentence, but content-wise I rather think along these lines

"Roll, Pitch, Yaw from MAV_CMD_DO_MOUNT_CONTROL message should contain useful data (or NaN), but can be ignored by the mount if the mount desires so."

--------------
not so important

>Regarding MAV_MOUNT_MODE_RC_TARGETING vs MAV_MOUNT_MODE_MAVLINK_TARGETING just to be very clear, I think you are saying that from the perspective of the usage/params they are the same. 

not exactly. My point was rather that an autopilot may want to make a different use of them than a gimbal (current practice). 

> It is just that they indicate a message from a different source. It isn't obvious to me why the source isn't conveyed/managed by component ID.

Consider this: You listen on the phone to two people and both are constantly babling, and even worse they emit partially conflicting info. You are easily able to distinguish them by their voice (= different componentIDs). This is what you suggest as concept. In contrast, I would think that the two bablers should somehow synchronize with each other so that you just get a single meaningful voice stream. It might be just preference, but to me it's clear that I vote for the 2nd concept. 
Btw, this is not a completely unrealistic example et all, it's a companion+autopilot+gimbal. And I think that this "conflicting info" thing is what is happening with Solo.

> Re use of NaNs or otherwise

I think we have different opinions here. 

NaNs have the advantage to at first come across as simple solutions, but have the disadvantage that they can be used up only once. E.g. in some hypothetical past one could have thought that it's useful to define the NaN to mean "no stabilization", as in this ancient time this might have avoided a MOUNT_CONFIGURATION message and extra stabilize fields. Then, one now couldn't use the NaN for meaning "use mount defaults". And who knows what comes in 3 years.

> Personally I'd find using enums harder. I like being able to understand what the value of a field is in the field itself.

I feel the same. But NaNs have its disadvantages as a general, extensible concept.

>I think you are suggesting that instead we might use a separate field

I was NOT voting for the one or the other, but for a certain visible common approach in the spec. I e.g. do not think that the usage with respect to enums vs NaN/-1 is consistent.

> I don't particularly see the benefit in using them for this case

You are correct, there is no benefit for this case. There can be a benefit in some future, since, as said, enums have the advantage of being extensible.
It's largely a design decision. My point only was that it would be great if the spec would not be a mix of many different design decisions but would follow a certain common visible philosophy.

:)

amilcarlucas:
This is for a camera OR antenna mount. Yes, you can have an antenna pointing at your GCS.
I do not like this message at all, there are too many options on the units side of things, param 4 and 7 are not standard..... etc
Can we please create new messages with only deg, or only deg/s or only lat/lon/alt ?

amilcarlucas:
I volunteer to do add Ardupilot support for the new clean messages.

olliw42:
:+1: 

also in view of the last mavlink devcall creating a new set of messages would probably the most sensible approach. This would combine both the positions of LorenzM and JulianO, and would avoid any backwards incompatibility. 

in the devcall I was planting the idea that quaternion based messages could be the argumentative vehicle, but Euler vs quaternions is certainly a matter of taste&perception too (quaternions are not as counter-intuitive as many may say, btw)

all over all I'm kind of convinced now too that a set of new messages would be the best way to go (DO_MOUNT could then be properly documented but also declared as deprecated)

I'd like to mention that ideally the new set of messages would have complementing messages for reporting, i.e. things like MOUNT_STATUS and MOUNT_ORIENTATION probably should be revisited too

I "volunteer" to add support of the new clean set of messages to the STorM32 gimbal controller. ;)

hamishwillee:
This is a big thread now :-) @olliw42 Just splitting out a couple of points into own responses:

Re NaNs vs enums you're quite right that a consistent approach would be awesome. I'm just can't see how NaNs can fill the roll of enums in the "general case". I can see how you might use an enum to do what NaNs do for us, but generally not without paying an efficiency cost of extra fields.

So I agree with the goal but I don't see the solution. I may be that I'm just being dumb - has happened before.

Do you think you could prototype a message or two showing how a common solution might work using either approach? So I can see how the enum replaces the NaN or visa versa?

> Consider this: You listen on the phone to two people and both are constantly babling, and even worse they emit partially conflicting info. You are easily able to distinguish them by their voice (= different componentIDs). This is what you suggest as concept. In contrast, I would think that the two bablers should somehow synchronize with each other so that you just get a single meaningful voice stream. It might be just preference, but to me it's clear that I vote for the 2nd concept.

You are saying that the decision to have separate messages is about two things: being able to differentiate streams and being able to synchronise them.

I think you are saying that you could have a single message and distinguish by component ID if all you want to do is distinguish sources/streams.

I think (?) you are further saying that if you want to synchronise streams you need to have separate messages (rather than one message with different source). Is that correct?

hamishwillee:
@julianoes I don't know how closely you're following the thread, but @amilcarlucas and @olliw42 both support new messages for this and have offered to integrate changes into ArduPilot and Storm Gimbal for said new messages
(see https://github.com/mavlink/mavlink/pull/1135#issuecomment-494304680 and https://github.com/mavlink/mavlink/pull/1135#issuecomment-494309524).

Anyway, as I understand it the decision for how to proceed this was left with you to research, so ...

FWIW:
- Personally I think this is a messy interface and the "right" thing to do is properly redefine it without the ambiguity. Even if we just add clarification on the direction, there are lots of other "holes". We could fill them with definitions, but not without potentially breaking someone.
- I do understand that uptake time of new messages is quite slow.
- @olliw42 Might also be willing to work with @amilcarlucas on the first draft. I don't know enough about gimbals and mounts to offer much useful until review stage.
- I wonder if we could involve other gimbal manufacturers so that we end up with something that gets more immediate traction.

julianoes:
> Personally I think this is a messy interface and the "right" thing to do is properly redefine it without the ambiguity. Even if we just add clarification on the direction, there are lots of other "holes". We could fill them with definitions, but not without potentially breaking someone.

In this case we would probably use different commands instead of using this mode parameter which makes things tricky. Note that we need to use commands if we want the capability to set a gimbal as part of a mission.

> I do understand that uptake time of new messages is quite slow.

I think standalone gimbals are mostly used in videography and therefore handled manually, historically controlled by RC channels, and these days by the same PWM values even if transmitted digitally.

And if tighter integration is necessary, a manufacturer typically controls both autopilot and gimbal and can make the protocol up and does not need to standardize.

Additionally the commands are called `MOUNT_` instead of gimbal, so no one that is looking for gimbal messages finds them. That's probably also why [MAV_CMD_DO_SET_ROI_LOCATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION) was added which is basically the same as the [MAV_CMD_DO_MOUNT_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONTROL) with [MAV_MOUNT_MODE_GPS_POINT](https://mavlink.io/en/messages/common.html#MAV_MOUNT_MODE_GPS_POINT).

I looked at where `DO_MOUNT_CONTROL` is implemented so far:

- QGroundControl: a mission item can be generically set to any command and the params adapted accordingly. Usability is 0 and this was probably seldom used (maybe only by me).
- MissionPlanner: not sure, I only found this: https://github.com/ArduPilot/MissionPlanner/blob/4e281f26cbbfd3f4c616e271c11563c20942b7c8/GCSViews/FlightData.cs#L3844-L3863
- Dronecode SDK: [used and working](https://github.com/Dronecode/DronecodeSDK/blob/a74bb92ab6dcd1acb8e2d63eb9b8a712ae0620e4/plugins/gimbal/gimbal_impl.cpp#L26-L39).
- ArduPilot: looks like it's used for [SToRM32](https://github.com/ArduPilot/ardupilot/blob/9c26b5bb9c50f3baeafe1e0d4e83d71bb9d3ccc4/libraries/AP_Mount/AP_Mount_SToRM32.cpp#L73-L75) and [ArduSub](https://github.com/ArduPilot/ardupilot/blob/706a75ad0356b003a411ff83644b314ec217c1ad/ArduSub/commands_logic.cpp#L793), both times just the angle mode, and [mission commands are forwarded](https://github.com/ArduPilot/ardupilot/blob/791b6effddb7d5bf06ac7b9e4c0ec13c494b617e/ArduCopter/mode_auto.cpp#L489-L492).
- PX4: [Mission commands are forwarded](https://github.com/PX4/Firmware/blob/6627c60e5bf6f63c2624ade4fe07500d3e7720d2/src/modules/vmount/output_mavlink.cpp#L107-L113)

I wonder if it would make sense to keep the [MAV_CMD_DO_MOUNT_CONTROL](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONTROL) message slimmed down for just `MAVLINK_TARGETING`. use [MAV_CMD_DO_SET_ROI_LOCATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION) for GPS (or other) points, and use `CONFIGURE` for things like `RETRACT`.
This way we would not break existing implementation but still clean it up a bit.

hamishwillee:
> I wonder if it would make sense to keep the MAV_CMD_DO_MOUNT_CONTROL message slimmed down for just MAVLINK_TARGETING. use MAV_CMD_DO_SET_ROI_LOCATION for GPS (or other) points, and use CONFIGURE for things like RETRACT.
This way we would not break existing implementation but still clean it up a bit.

So just to clarify, you mean modify MAV_CMD_DO_MOUNT_CONTROL  so that it is just for/assumes targetting mode? 

Then modify MAV_CMD_DO_MOUNT_CONFIGURE  as a new command for retracting etc?

olliw42:
indeed, this has become a long and rich thread. So I too will split my responses into individual posts :)

@julianoes 

IMHO, there are only exactly two paths to proceed:
(1) Keep the mount messages as they are but document better, and create new replacement messages.
or
(2) Modify the mount messages, and break existing implementations.

I can't see how your proposal would not break the existing implementations. IMHO there is no such thing as modifying them and not breaking anything. For instance, ArduPilot makes use of the various mode cases, which immediately means that restricting it to just targeting will break ArduPilot's functionality. Similar holds true for the gps/roi position fields.

So, IMHO, either (1) or (2), with nothing in between

julianoes:
> For instance, ArduPilot makes use of the various mode cases, which immediately means that restricting it to just targeting will break ArduPilot's functionality. 

Does it? I couldn't see that in code.

olliw42:
> the commands are called MOUNT_ instead of gimbal, so no one that is looking for gimbal messages finds them

I don't see a problem with that, "mount" is supposed to be a superclass to gimbals but might include other "pointing" devices. I think it's ok to have a more generic thing here. Users shouldn't have issues to understand that concept. 
What might be somewhat inconsistent however is that there is GIMBAL component id and type, but no mount id or type, but that's a minor IMHO.

olliw42:
> Does it? I couldn't see that in code.

https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Mount/AP_Mount_SToRM32.cpp#L24-L70

note the very different calculus for the RC, MAVLINK, and GPS cases

olliw42:
> I looked at where DO_MOUNT_CONTROL is implemented so far:

Solo's smart shoots make heavy use of it
not sure if that still counts as relevant however, given that Solo is dying out

olliw42:
@hamishwillee

>* @olliw42 Might also be willing to work with @amilcarlucas on the first draft. I don't know enough about gimbals and mounts to offer much useful until review stage.
>* I wonder if we could involve other gimbal manufacturers so that we end up with something that gets more immediate traction.

I would certainly be willing to comment on drafts for new messages. However I would find it more difficult to work on drafting a first draft. Reason is your second point. I would think that it would be more important for you guys to get opinions of other gimbal manufactures. E.g., I believe to know that Yuneec communicates with their gimbal in e.g. the H500 type things with Mavlink. Yuneec is a Dronecode member the last time I looked ... I'm not LOL. There is Greemsy and Basecam, and who knows I don't know ... all potentially important to you.

My perspective is necessarily very limited. 

The general approach I would take is to move towards the more "robotics-type" of messages also for gimbals/mount as much as one can see it for the flight control messages. For instance, a "mount control" would have angle plus speed fields.

olliw42:
> NaNs vs enums 

I believe that this was rather brought up by you and not me, I just added my sauce to it. 

My general position on NANs is, that any (float) field should be allowed to be set to NAN and result in a defined behavior, which however only can be "ignore it", but not be used for any sort of state encoding.   
E.g. if the gimbal is in a state there it consumes angle fields but the angles are NAN then the gimbal should not freak out but ignore the new data. NANs should however not be used to make the gimbal to switch the mode, e.g. one should not use NAN angles and valid pos fields to indicate GPS_POINT mode.
I note that in almost all cases it is pretty obvious what "ignore" is supposed to mean.

olliw42:
@julianoes @hamishwillee

>Note that there is a tiny but all crucial subtlety in here, there are two use cases for the MOUNT message:
It can be digested and operated on by the flightcontroller, and it can be digest and operated on by the gimbal. And issues obviously can emerge then this is not correctly "synchronized" and the gimbal gets conflicting information.

I have a bit a feeling that this isn't yet fully entered your thoughts

Another (ArduPilot) example: When a companion sends a MOUNT with e.g. MAVLINK_TARGETING to the flight controller, when normally (= has_pan is set to zero) the yaw information is used by the flight controller to rotate the copter, and the pitch/roll angles are send to the gimbal to tilt the camera. Once ArduPilot was "broken" and one had to send a set-yaw-angle command (forgot the correct msg) to rotate the copter in addition to the MOUNT to tilt the camera.
The point I'm trying to make is that - logically - situations exist in which one needs to send info to the copter as well as to the gimbal in order to accomplish a task - and a logical approach would be to have two messages, one for the FC and one for the gimbal. The disadvantage of two messages are however that it's less efficient and consumes higher bandwidth with no apparent benefit except of better logic.

Further similar dichotomies exists, since one may want a specific task to be done by the FC or by the gimbal. For instance ROI. The calculation of the gimbal angles from the GPS locations can be done inside the FC, which then sends angles to the gimbal (that's how ArduPilot is doing it), or request the gimbal to do the calculation and send location information to the gimbal. One probably can argue forever which way is the best without reaching a conclusion. The point, here too it makes logically sense to have two different messages, one to be consumed by the FC and one to be consumed by the mount. The "who-is-doing-the-mixing" problem mentioned in the above is another example along the same lines.

So, there could be reason to have a MOUNT_CONTROL and a SET_ROI. But there are also reasons for "combined" messages which carry info for both the FC and the  gimbal.

This duality/ambiguity should not be forgotten. But all this makes it difficult to come up with a really good set of messages.

olliw42:
@julianoes

>And if tighter integration is necessary, a manufacturer typically controls both autopilot and gimbal

YES. This is probably the most important statement. :+1: 

The MOUNT messages obviously allow some sort of "tighter integration" ... and that's, I presume, why they look so mixed up as the do

>And if tighter integration is necessary, a manufacturer ... can make the protocol up and does not need to standardize.

I think that Mavlink should provide a standardized way to allow for "non-standardized tighter integration". It's actually a point I'm fighting with myself for long and have tried several approaches now, none fully satisfying. I guess this your sentence has motivated me to maybe come up with a proposal.

If one would think this your thought to the end, the result would be only one simple positioning Mavlink message for positioning the gimbal, maybe including angles and speed fields, but nothing else, and whatever sets of commands which are required to operate on the copter (likely already all existing). Any "tighter integration" would then be non-standardized
(or use a standardized method for doing non-standardized messages)

Should this be the conclusion?

hamishwillee:
> NaNs vs enums

Yes. I *think* that is how NaNs are consistently used (i.e. "value unchanged", "value default" or "value not defined").

hamishwillee:
> Should this be the conclusion?

I'll leave that to Julian because I think I got lost in the nuances. Completely agree that it makes sense to involve industry in proposal, if we can get our head around what we think might work.

FWIW the commands that a _GCS_ needs are mostly high level. You probably just want to be able to set an ROI and tell the gimbal/mount to "make safe"/"get ready for use". You don't care how it does those things (e.g. move to particular angle and retract). Maybe you might also want a message that sends back "telemetry" from some gimbals on their current state.

olliw42:
ok, to not always just say what needs to be considered, let me try to be constructive , so, here a proposal:

1) All fields in DO_MOUNT_CONTROL (and MOUNT_STATUS) remain as they are, including the signs of the angles. Things will have to be properly documented. 

2) The entanglement inherent to DO_MOUNT_CONTROL and DO_MOUNT_CONFIGURE (entangled in that they could be meant for the autopilot or the gimbal as described in the above), shall be dis-entangled by specifying that these mount commands are ONLY for the autopilot, but NOT for the gimbal. That is, a gimbal must not react to any received DO_MOUNT_CONTROL or DO_MOUNT_CONFIGURE.

Rational: It's based on these observations: (a) The problems due to the entanglement I've described. (b) A MAVLink-speaking gimbal is canonically attached to an autopilot, which is so-to-say the "master". There might be practical use cases where there is no such "master", but I suggest to not draft for these rather exceptional cases but for the canonical case. (c) A gimbal can't normally do a thing like ROI. "Advanced" gimbals might be able to do it, but we always can ask the autopilot/master to do the calculus, and by (b) we assumed the presence of such a master, so there is no loss in functionality by dropping this. (d) Also, while RC TARGETING could have some meaning to a gimbal in non-canonical cases, this mode is relevant only to the autopilot in the canonical case, since it's the autopilot which receives and handles the RC signals. So, again there is no loss in functionality.

Point (2) also reflects the reality as found in ArduPilot.

3) A new message to control the orientation of the gimbal is introduced, as well as a new message for the gimbal to report it's orientation. I specifically refer to "gimbal" and not "mount" here, since there is already a MAV_TYPE_GIMBAL and MAV_COMPONENT_ID_GIMBAL but none of these for mounts, so it's at least language-wise consistent to name these new messages also by "gimbal".

I suggest a GIMBAL_CONTROL message with these fields:

- float tilt,roll,yaw: desired angles in degree in proper frame convention, can be NAN if to be ignored. 
- float tilt_rate,roll_rate,yaw_rate: desired rotation speeds in degree/sec in proper frame convention, can be NAN if to be ignored. 

Since it's to a gimbal, there can't be any confusion about that yaw and yaw_rate are relative.

Similar I suggest a GIMBAL_ORIENTATION message with the analogous fields
- timestamp
- float tilt,roll,yaw: actual angles in degree in proper frame convention, cannot be NAN, must be reported
- float tilt_rate,roll_rate,yaw_rate: actual rotation speeds in degree/sec in proper frame convention, can be NAN if not known. 

It could make some sense to add a status or mode or whatever named enum/buitmask field to GIMBAL_CONTROL, to make it a bit more future proof, but for the moment it wouldn't be needed. One also could think about a GIMBAL_CONFIGURE, but here too, for the moment it wouldn't be needed. 


As a corollary to (2) the proposal means, that the DO_MOUNT_CONTROL/CONFIGURE commands remain, but are now understood only as messages to the autopilot. That is, the proposal in a nutshell is that the MOUNT commands go to the autopilot, which does what it thinks is appropriate doing with them, and sends out GIMBAL_CONTROLs to the gimbal. 

As a further corollary it follows that a GCS or companion should not send out GIMBAL_CONTROL, but send MOUNT_CONTROL.

And yet another corollary is that everything "outside" the autopilot, such as companions or GCSes do not need to be changed, as regards these commands (they might have to be changed to consume the new GIMBAL_ORIENTATION if they desire so). The autopilot code in contrast needs to be changed, but minimally so. For ArduPilot. e.g., the code in AP_Mount_STorM32 would have to be changed in only one instances (as far as I can see), i.e. instead of sending a DO_MOUNT_CONTROL to the gimbal it should send a GIMBAL_CONTROL.

This proposal does NOT touch the question of whether the commands/messages intended for the autopilot should see a rehaul too! There would be good reasons for that too, e.g., the strange sign convention in DO_MOUNT_CONTROL still remains, and the inconvenience of having both angles and ROI values in the commands, and so on. So, one may want to discuss this. This discussion however should be much simpler now, since it's now only about the autopilot, and not the gimbal anymore.

have fun, Olli

julianoes:
Closing in favor of #1174.

PR #1136, opened 2019-05-14
common: add mission changed broadcast

### LINKED ISSUE #1134: Feature request: Mission changed broadcast
It would be nice if there was some kind of "mission changed broadcast" in order to tell all ground stations, SDKs etc. when a mission (or geofence) has changed and needs to be re-downloaded.

The message should support all mission types, so mission, geofence, and safe points.

FYI: @7200rpm @sanderux @JonasVautherin

julianoes:
This adds a mission changed broadcast which a vehicle should send whenever the mission (or geofence or safe points) have changed.

Closes #1134.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3621,6 +3621,14 @@
       <field type="float" name="p2y" units="m">y position 2 / Longitude 2</field>
       <field type="float" name="p2z" units="m">z position 2 / Altitude 2</field>
     </message>
+    <message id="52" name="MISSION_CHANGED">
+      <description>A broadcast message to notify any ground station or SDK if a mission, geofence or safe points have changed on the vehicle.</description>
+      <field type="int16_t" name="start_index">Start index for partial mission change (-1 for all items).</field>
+      <field type="int16_t" name="end_index">End index of a partial mission change (-1 for all items, or to end if start_index is set).</field>
+      <field type="uint8_t" name="origin_sysid">Sysid of the author of the new mission.</field>
+      <field type="uint8_t" name="origin_compid">Compid of the author of the new mission.</field>
+      <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
+    </message>
     <message id="55" name="SAFETY_ALLOWED_AREA">
       <description>Read out the safety zone the MAV currently assumes.</description>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame. Can be either global, GPS, right-handed with Z axis up or local, right handed, Z axis down.</field>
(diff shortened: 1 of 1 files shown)


julianoes:
> Just FMI, is this something you might also reasonably instead use the proposed events interface for?

I would say it shouldn't be part of the events interface. It's easier like this and can be included with the mission protocol docs.

message_definitions/v1.0/common.xml
```diff
@@ -3606,6 +3606,16 @@
...
       <field type="float" name="p2z" units="m">z position 2 / Altitude 2</field>
     </message>
+    <message id="52" name="MISSION_CHANGED">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>A broadcast message to notify any ground station or SDK if a mission, geofence or safe points have changed on the vehicle.</description>
+      <field type="int16_t" name="start_index">Start index for partial mission change (-1 for all items).</field>
+      <field type="int16_t" name="end_index">End index of a partial mission change (-1 for all items, or to end if start_index is set).</field>
+      <field type="uint8_t" name="origin_sysid">Sysid of the author of the new mission.</field>
+      <field type="uint8_t" name="origin_compid">Compid of the author of the new mission.</field>
```
hamishwillee: Should this be an enum of type MAV_COMPONENT ?

In the descriptions for these origin_ fields can we use System ID and Component ID rather than Sysid and Compid?

hamishwillee:
I was going to merge, but added some additional suggestions above.

If you fix those you can self merge :-)

message_definitions/v1.0/common.xml
julianoes: Yes to both.

message_definitions/v1.0/common.xml
```diff
@@ -3609,6 +3609,16 @@
       <extensions/>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
+    <message id="52" name="MISSION_CHANGED">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>A broadcast message to notify any ground station or SDK if a mission, geofence or safe points have changed on the vehicle.</description>
+      <field type="int16_t" name="start_index">Start index for partial mission change (-1 for all items).</field>
+      <field type="int16_t" name="end_index">End index of a partial mission change (-1 for all items, or to end if start_index is set).</field>
```
hamishwillee: Re 
> End index of a partial mission change (-1 for all items, or to end if start_index is set)
 
Do you mean:

> End index of a partial mission change. -1 indicates is a synonym for the last mission item (i.e. selects all items from start_index). Field ignored if start_index=-1.

julianoes: Yes that's what I mean.

hamishwillee:
OK, I updated it. Merging now. It's WIP, so you can modify again if needed. @DonLakeFlyer FYI.

PR #1137, opened 2019-05-15
Adding an enum to storage info.

dogmaphobic:
Following a [conversation here](https://github.com/mavlink/qgroundcontrol/issues/7434), I've added an enum to indicate the status of the (camera) storage. One more status was added: ~**None**~ **Not Supported**. That indicates that storage should be ignored (the camera is capable of functioning without it).

I've also added a description telling this message is not only sent in response to a request but also whenever the status of the storage changes. This is for notifying the GS of events such as microSD card inserted/removed, storage formatted, etc.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -763,8 +763,8 @@
       <entry value="2" name="STORAGE_STATUS_READY">
         <description>Storage present and ready.</description>
       </entry>
-      <entry value="3" name="STORAGE_STATUS_NONE">
-        <description>Storage should be ignored.</description>
+      <entry value="3" name="STORAGE_STATUS_NOT_SUPPORTED">
+        <description>Camera does not supply storage status information. Capacity information in STORAGE_INFORMATION fields will be ignored.</description>
       </entry>
     </enum>
     <!-- The MAV_CMD enum entries describe either: -->
@@ -4824,12 +4824,12 @@
       <field type="float" name="focusLevel">Current focus level (0.0 to 100.0, NaN if not known)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
-      <description>Information about a storage medium. This message is sent in response to a request or whenever the status of the storage changes.</description>
+      <description>Information about a storage medium. This message is sent in response to a request and whenever the status of the storage changes (STORAGE_STATUS).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="storage_id">Storage ID (1 for first, 2 for second, etc.)</field>
       <field type="uint8_t" name="storage_count">Number of storage devices</field>
       <field type="uint8_t" name="status" enum="STORAGE_STATUS">Status of storage</field>
-      <field type="float" name="total_capacity" units="MiB">Total capacity.</field>
+      <field type="float" name="total_capacity" units="MiB">Total capacity. If storage is not ready (STORAGE_STATUS_READY) value will be ignored.</field>
       <field type="float" name="used_capacity" units="MiB">Used capacity.</field>
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
       <field type="float" name="read_speed" units="MiB/s">Read speed.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4809,11 +4824,11 @@
...
     </message>
     <message id="261" name="STORAGE_INFORMATION">
-      <description>Information about a storage medium.</description>
+      <description>Information about a storage medium. This message is sent in response to a request or whenever the status of the storage changes.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="storage_id">Storage ID (1 for first, 2 for second, etc.)</field>
       <field type="uint8_t" name="storage_count">Number of storage devices</field>
-      <field type="uint8_t" name="status">Status of storage (0 not available, 1 unformatted, 2 formatted)</field>
+      <field type="uint8_t" name="status" enum="STORAGE_STATUS">Status of storage</field>
       <field type="float" name="total_capacity" units="MiB">Total capacity.</field>
```
hamishwillee: @dogmaphobic We now need to report for **each** of the capacity fields how they are affected by the `STORAGE_STATUS` for example:
```
<field type="float" name="total_capacity" units="MiB">Total capacity. If storage is not ready (STORAGE_STATUS_READY) value should be ignored.</field>
```
Or you might want to enforce a value - ie  ... If storage is not ready (STORAGE_STATUS_READY) value should be set to -1.

message_definitions/v1.0/common.xml
```diff
@@ -4809,11 +4824,11 @@
       <field type="float" name="focusLevel">Current focus level (0.0 to 100.0, NaN if not known)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
-      <description>Information about a storage medium.</description>
+      <description>Information about a storage medium. This message is sent in response to a request or whenever the status of the storage changes.</description>
```
hamishwillee: Maybe: "request AND whenever"

We could also reference the field/enum

> Information about a storage medium. This message is sent in response to a request and whenever the storage status changes (STORAGE_STATUS).

Up to you

message_definitions/v1.0/common.xml
```diff
@@ -752,6 +752,21 @@
...
+      <entry value="0" name="STORAGE_STATUS_EMPTY">
+        <description>Storage is missing (no microSD card loaded for example.)</description>
+      </entry>
+      <entry value="1" name="STORAGE_STATUS_UNFORMATTED">
+        <description>Storage present but unformatted.</description>
+      </entry>
+      <entry value="2" name="STORAGE_STATUS_READY">
+        <description>Storage present and ready.</description>
+      </entry>
+      <entry value="3" name="STORAGE_STATUS_NONE">
```
hamishwillee: Absent the prior discussion I would not know if 
- STORAGE_STATUS_NONE means "there is no storage" or "We can't report on storage" 
- "Storage should be ignored." is not very precise.

Is this what you mean?:

```

      <entry value="3" name="STORAGE_STATUS_NOT_SUPPORTED">
        <description>Camera does not supply storage status information. Capacity information in STORAGE_INFORMATION fields must be XXXX.</description>
      </entry>
```

where XXX might "ignored" or "set to -1" or whatever.

message_definitions/v1.0/common.xml
olliw42: AND

message_definitions/v1.0/common.xml
olliw42: are ignored

message_definitions/v1.0/common.xml
dogmaphobic: Changed to:
`Total capacity. If storage is not ready (STORAGE_STATUS_READY) value will be ignored.`
Stating the value **will** be ignore makes it certain I don't care what value is put here if storage is not in ready state.

message_definitions/v1.0/common.xml

message_definitions/v1.0/common.xml
dogmaphobic: Changed it to:
`Camera does not supply storage status information. Capacity information in STORAGE_INFORMATION fields will be ignored.`

hamishwillee:
@dogmaphobic Very happy with thta, but you will also have to update the other capacity fields with same pattern as total capacity - ie used_capacity, available_capacity.

If you're sure these messages are only used by you, then after that I don't see any reason not to self-merge.

dogmaphobic:
> @dogmaphobic Very happy with that, but you will also have to update the other capacity fields with same pattern as total capacity - ie used_capacity, available_capacity.
> 
> If you're sure these messages are only used by you, then after that I don't see any reason not to self-merge.

Done.

Nothing really changed. It was just never really defined. The issue came up with a camera that does not report storage, yet takes pictures. The values in `total_capacity` were replaced with a new enum but they hold the original numerical values. Only this new "ignored" type was added. I will merge once this passes. I then need to update QGC to reflect this (it doesn't even pay attention to any of it yet).

hamishwillee:
Works for me! Nice turnaround time.

PR #1138, opened 2019-05-18
common: add ACCELERATION message

okalachev:
Common messages set totally lacks the vehicle acceleration data report. And the only message consisting it, `LOCAL_POSITION_NED_COV` was [refused](https://github.com/PX4/Firmware/pull/11081#issuecomment-449159099) by PX4, as its large size, as one of the reasons.

But this data is useful for the estimator debugging. It’s important to know, what the estimator thinks of the vehicle accelerations in the local coordinate frame, to figure out, why the estimated velocity and position acts one or other way.

This PR proposes adding very small and simple message with the accelerations data (as they were computed/considered by the estimator).

@mhkabir, @TSC21

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5036,6 +5036,13 @@
       <field type="uint16_t" name="lac">Location area code. If unknown, set to: 0</field>
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
+    <message id="335" name="ACCELERATION">
+      <description>Linear acceleration in the local coordinate system.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="float" name="ax" units="m/s/s">X acceleration</field>
+      <field type="float" name="ay" units="m/s/s">Y acceleration</field>
+      <field type="float" name="az" units="m/s/s">Z acceleration</field>
+    </message>
     <message id="340" name="UTM_GLOBAL_POSITION">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->


TSC21:
Please create an RFC first under https://github.com/mavlink/rfcs. Then bring the discussion to a dev call. Only then it makes sense to create a PR.

okalachev:
@TSC21 

I'm not sure I get the point of this.

What is the point of copying the text above to RFCs repository and having this discussion there, while here exists pretty the same functionality? While keeping the discussion here, in this PR, has a big advantage: everyone later can easily find it though the blame mechanism.

TSC21:
You started from the end basically. Please check: https://mavlink.io/en/contributing/contributing.html

mhkabir:
I'm not in favour of an RFC for small changes like this one.

TSC21:
> I'm not in favour of an RFC for small changes like this one.

This is not a small change - a small change in my view is extending a message with new fields or change the field specification. This is adding a new message to the protocol, and makes sense to have a  proper discussion first on its purpose, content and structure before actually bringing it to a PR.

I am ok to discuss this change if the normal process gets followed. Otherwise, I leave to other to review it.

hamishwillee:
> I am ok to discuss this change if the normal process gets followed. 

@TSC21 The normal process for this size of change is currently to use a PR - so please review.

I will follow up in the dev call to confirm what is desired, and then clarify the contributing section. Bug here: https://github.com/mavlink/mavlink-devguide/issues/185

message_definitions/v1.0/common.xml
```diff
@@ -5036,6 +5036,13 @@
       <field type="uint16_t" name="lac">Location area code. If unknown, set to: 0</field>
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
+    <message id="335" name="ACCELERATION">
+      <description>Linear acceleration in the local coordinate system.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="float" name="ax" units="m">X acceleration</field>
```
hamishwillee: Should units be metres per second squared?

message_definitions/v1.0/common.xml
```diff
@@ -5036,6 +5036,13 @@
       <field type="uint16_t" name="lac">Location area code. If unknown, set to: 0</field>
       <field type="uint32_t" name="cid">Cell ID. If unknown, set to: UINT32_MAX</field>
     </message>
+    <message id="335" name="ACCELERATION">
+      <description>Linear acceleration in the local coordinate system.</description>
```
hamishwillee: By "local coordinate system" do you mean body frame as below?


Is there an unambiguous agreed terminology for that system?

message_definitions/v1.0/common.xml
okalachev: Yeah, my bad!

message_definitions/v1.0/common.xml
okalachev: I mean the local coordinate system used by the estimator, like in the `LOCAL_POSITION_NED` message. This should help understanding and debugging its velocities and position estimations.

Actually, `frame` field may be added to the message, which can clarify this issue, although `LOCAL_POSITION_NED` lacks (but, for example, `ODOMETRY` does have it). But this would increase the size of the message and make it less simple.

TSC21:
> @TSC21 The normal process for this size of change is currently to use a PR - so please review.

Alright then, thanks @hamishwillee.

okalachev:
@TSC21 

By debugging I meant the wide sense of this word. After all, all of the messages purposes are debugging in some way. 

This is not debug in its literal sense, I’m not proposing to add a message, which exposes the `i` variable value in some random cycle. Vehicle acceleration is the common data. And this is important information of the system.

I'm not against adding additional acceleration fields to `ODOMETRY` though. The problem is `ODOMETRY` is not widely adopted for now. A lot of existing systems use `LOCAL_POSITION_NED`. As an example, ArduPilot [doesn't have](https://github.com/ArduPilot/ardupilot/blob/e801821f0cda28d2fc115ad8915e0fee678fcf9c/libraries/GCS_MAVLink/GCS_Common.cpp#L2199) `ODOMETRY` support at all.

And if the system is not using `ODOMETRY` pipeline (with conditional `ODOMETRY_WITH_ACCEL`) `LOCAL_POSITION_NED` + `ATTITUDE_QUATERNION` + `ACCELERATION` might help.

@mhkabir, @LorenzMeier, can you please share some opinion on this?

hamishwillee:
I'm going to be hands off further review until those with more knowledge comment.

FWIW I do lean towards @TSC21 point of view. Probably because I have a preference for fewer "complete" messages. Unless of course this info should be sent at a vastly different rate than the rest of the message.

LorenzMeier:
So what you essentially need is LOCAL_POSITION_NED_COV - I think we can easily re-introduce this in PX4 and it probably makes more sense than introducing a completely new message. Does that sound fair?

TSC21:
> So what you essentially need is LOCAL_POSITION_NED_COV - I think we can easily re-introduce this in PX4 and it probably makes more sense than introducing a completely new message. Does that sound fair?

@LorenzMeier what would be the advantage of re-introduce `LOCAL_POSITION_NED_COV` instead of extending `ODOMETRY` with covariance fields?

mhkabir:
I agree with Lorenz. `ODOMETRY` doesn't have any more message space to fit accelerations.

LorenzMeier:
Odometry is for a completely different use case. One is used to report position, the other one encodes sensor / VIO input.

TSC21:
> Odometry is for a completely different use case. One is used to report position, the other one encodes sensor / VIO input.

The message definition and PX4 upstream current implementation of the message actually supports forward and backward data flowing. You have input on `ODOMETRY` format and output the same way.

LorenzMeier:
That usage doesn't make it better / more correct. Odometry is by definition drifting, while the local position is assumed to be absolute, but not registered to a global frame. It is not the same thing and *should not be used interchangeably*.

TSC21:
> Odometry is by definition drifting, while the local position is assumed to be absolute, but not registered to a global frame. It is not the same thing and should not be used interchangeably.

That's correct. Though, in systems where you need back propagation of the onboard estimator for mapping for example, that would still be odometry. It really depends on how you really interpret and use that data.

Just for a reference, `LOCAL_POSITON_NED_COV` was removed because of flash fitting issues. I am not against adding it again, just considering it would make sense to explore the current message architecture.

okalachev:
Bringing back `LOCAL_POSITION_NED_COV`, generally, solves the issue.

I'll create a PR.

PR #1142, opened 2019-05-21
OBSTACLE_DISTANCE: add higher angle resolution and angle offset

magicrub:
This MAVLink v2 extension packet allows for flexible obstacle detection from multiple sensors of varying resolution while still being backwards compatible.

increment_f (to override increment):
When forced to send exactly 72 samples, having an integer spacing just didn't make sense. Needed to be float so that the correct angular range was understood

angle_offset:
Same as, and adds to, param "[PRX_YAW_CORR](http://ardupilot.org/rover/docs/parameters.html#prx-yaw-corr-proximity-sensor-yaw-correction)" in ArduPilot. However, if you have two sensors this woudl now allow you to send multiple messages covering different areas (at varying resolutions). For example, a 360 degree spinning lidar + long-range forward radar/lidar + wide angle sonar(s).

With multiple messages, we can aggregate a better picture and down-scale the resolution to whatever the flight controller can handle. And that will improve as we get fancier flight controllers with more memory. Meanwhile, the data would no longer limited by the comms link, just on how to store it once received.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5016,10 +5016,13 @@
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the UAV with index 0 corresponding to local North. A value of 0 means that the obstacle is right in front of the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
-      <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
+      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the UAV with index 0 corresponding to local forward + angle_offset. A value of 0 means that the obstacle is right in front of the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
+      <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element. (Ignored if increment_f greater than 0).</field>
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure.</field>
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure.</field>
+      <extensions/>
+      <field type="float" name="increment_f" units="deg">Angular width in degrees of each array element as a float. If greater than 0 then this value is used instead of the uint8_t increment field.</field>
+      <field type="float" name="angle_offset" units="deg">Relative angle offset of the 0-index element in the distances array. Value of 0 corresponds to forward. Positive values are offsets to the right.</field>
     </message>
     <message id="331" name="ODOMETRY">
       <description>Odometry message to communicate odometry information with an external interface. Fits ROS REP 147 standard for aerial vehicles (http://www.ros.org/reps/rep-0147.html).</description>


hamishwillee:
@magicrub So before we had at most 72 elements with integer-spaced angles and starting with element 0 pointing north. 
With this (compatible) change you can instead specify spacing as an integer and the starting angle. Now you can specify any number of sensors and spacing (by sending multiple messages).

I like it.
- Would there be any benefit to instead have a new message?
- Message description probably should be updated.

@mrivi @baumanta Any questions/thoughts/suggestions? 

Is there anyone else we know uses this message regularly and might offer useful feedback?

message_definitions/v1.0/common.xml
```diff
@@ -5020,6 +5020,9 @@
       <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure.</field>
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure.</field>
+      <extensions/>
+      <field type="float" name="increment_f" units="deg">Angular width in degrees of each array element as a float. If >0 then this value is used instead of uint8_t increment.</field>
```
hamishwillee: Maybe:
> If >0 then this value is used instead of uint8_t increment.

TO

If >0 then this value is used instead of the uint8_t increment field.

message_definitions/v1.0/common.xml
```diff
@@ -5020,6 +5020,9 @@
...
+      <field type="float" name="angle_offset" units="deg">Relative angle offset of first distance with index 0. Value of 0 corresponding to local North/forward and matches existing map and otherwise adds an offset. Positive is to the right.</field>
```
hamishwillee: Maybe:

> Relative angle offset of first distance with index 0. Value of 0 corresponding to local North/forward and matches existing map and otherwise adds an offset. Positive is to the right.

TO

> Relative angle offset of the 0-index element in the distances array. Value of 0 corresponds to local North/forward. Positive values are offsets to the right.

- Note, not sure about "forward" in "North/forward" above. The old values only mention North.
- Removed "and matches existing map and otherwise adds an offset" because after this addition the "matching existing map" won't necessarily mean anything to a reader.

message_definitions/v1.0/common.xml
```diff
@@ -5020,6 +5020,9 @@
       <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element.</field>
```
hamishwillee: Do we need to add "(Ignored if increment_f >0)".?

hamishwillee: Also add not for the distances description above, as 0 index no longer corresponds to north by default.

magicrub:
The advantage of a new packet would be to use a larger array at the end of the msg and allow 0 to be invalid. That would allow large msgs for high resolution and also short messages with the zero truncation on MAVLink 2.

magicrub:
That being said, I think this packet extension is still useful. We can still add a new packet later on if we wish.

hamishwillee:
Just to be pedantic, zero truncation starts with the smallest non-extension fields first, so that would only be a benefit if we could ensure that the other fields are at least the same size as items in the array. With current field sizes the array would not be truncated, and to make it so, we would have to increase all other fields to use uint16_t or greater.

magicrub:
So, the benefit decreases. Even more of a reason to  use the existing packet.

WickedShell:
> zero truncation starts with the smallest non-extension fields first

That's not correct. It starts from the back of the payload, which will always be the extension fields in the order declared if they are present. So as soon as there was an extension to this that is used you can't benefit from trimming on the array.

magicrub:
@WickedShell I'm talking about the new-packet scenario. I understand that the current packet will have an extension and will have non-zero data and this never shortened. However, it's a bit of a mute point because that array of 72 shouldn't ever be filled with 0 data anyway - should be filled with UINT16

hamishwillee:
@WickedShell As @magicrub said - I do fully under the truncation rules. Just clarifying if any benefit to a new message "in general", and in particular if the tradeoffs for truncation are worth it.

You could still argue a new message is a cleaner approach but I don't think in this case.

Other than my inline comments about the descriptions I like this. However I would like others who use the message to comment because I have no practical experience of the domain. If not before, I will make sure that happens on Dev Call.

WickedShell:
@hamishwillee @magicrub Sorry, I hadn't fully tracked the context on the previous messages, the comment about zero padding just caught my eye when skimming e-mail. I took it out of the context you meant, sorry about that!

mrivi:
This seems to be an improvement for increments smaller than 5deg. I am okay with the change.

message_definitions/v1.0/common.xml
magicrub: fixed

message_definitions/v1.0/common.xml
magicrub: I don't think anyone really uses it as "north" Doesn't really make sens that way. ROS doesn't use it like that. I just removed "North" and changed it to forward.

magicrub:
@mrivi Thanks for the review. @hamishwillee does this really deserve to be a dev call topic? Seems minor.

magicrub:
Not sure why Travis is failing, seems unrelated. Does anyone know about that?

hamishwillee:
> Not sure why Travis is failing, seems unrelated. Does anyone know about that

Yes, there is some formatting error - could be as simple as EOL being "wrong" or a space where not desired in message. You need to run `format_xml.sh`.
```
./message_definitions/v1.0/common.xml needs formatting - run ./scripts/format_xml.sh ./message_definitions/v1.0/common.xml
```

hamishwillee:
> does this really deserve to be a dev call topic? Seems minor.

Looks like a no-brainer improvement to me, but what do I know about the domain? Perhaps this extension might reduce the utility of further extensions in future. Perhaps someone in the dev call will have a strong reason to argue for a new message.

So generally my take is that I try to get rubber stamp from the dev call, to ensure that at least some people with experience commented. 

@LorenzMeier @auturgy @WickedShell Speak now if you think this is a bad idea/have suggestions.

Otherwise, if you really (?) need this urgently I will merge.

magicrub:
well, the format_xml.sh script didn't do anything. I'm on Windows 10 but have git config set to LF. I just double checked and rebased and pushed the branch again to hopefully force a LF EOL. Meanwhile, since that particular EOL Travis error is totally bypassed by your git commands then it needs to be fixed in git by adding a [.gitattributes](https://git-scm.com/docs/gitattributes) file with an entry of 

`*.xml text eol=lf`
or 
`*text eol=lf`

hamishwillee:
@magicrub I ran the format_xml (in a Linux vm) and the problem is the ">" in your descriptions. These need to be changed to &gt;

Can you create the PR for gitattributes if still needed? (I can do this, but if someone questions the change, better it be defended by the person who has a deeper understanding of git than me)

magicrub:
@hamishwillee I changed ">" to "greater than"

hamishwillee:
Thanks. 

> That's a pretty weird error that travis/BashScript should be a bit more clear on

Yes and no. Yes, because it would be nice to get an error at this point so you understand for next time. No, because running the format script does indeed fix the problem without you having to know about it. 

Yes. All Travis does is run the format script, which in turn just runs the xmllint. So if you felt strongly you could create a PR that echos any issues it generates: https://github.com/mavlink/mavlink/blob/master/scripts/format_xml.sh#L37

hamishwillee:
Thanks @magicrub for your patience. Dev call agreed is good so merged.

magicrub:
great, thank you for the help!

PR #1143, opened 2019-05-22
common.xml: PARAM_EXT message clarifications

### LINKED ISSUE #1140: Responses to PARAM_EXT messages not fully specified
From the current spec one may conclude that the [PARAM_EXT_REQUEST_READ](https://mavlink.io/en/messages/common.html#PARAM_EXT_REQUEST_READ) and [PARAM_EXT_REQUEST_LIST](https://mavlink.io/en/messages/common.html#PARAM_EXT_REQUEST_LIST) messages should return [PARAM_EXT_VALUE](https://mavlink.io/en/messages/common.html#PARAM_EXT_VALUE) messages, but this could be clearly spelled out (similarly as it is done for [PARAM_EXT_SET](https://mavlink.io/en/messages/common.html#PARAM_EXT_SET) and [PARAM_EXT_ACK](https://mavlink.io/en/messages/common.html#PARAM_EXT_ACK)).

I would suggest to change the descriptions, e.g.

`PARAM_EXT_REQUEST_READ`: (Mavlink2) Request to read the value of a parameter with either the param_id string id or param_index. Responds with a PARAM_VALUE message if a valid parameter was indexed.

`PARAM_EXT_REQUEST_LIST`: (Mavlink2) Request all parameters of this component. After this request, all parameters are emitted via PARAM_VALUE messages.

`PARAM_EXT_VALUE`: (MAVLink 2) Response from a PARAM_EXT_REQUEST_READ or PARAM_EXT_REQUEST_LIST message, emiting the value of a parameter. The inclusion of param_count and param_index in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout.

:)

Btw, a statement like "If the current state is PARAM_ACK_IN_PROGRESS, you will accordingly receive a PARAM_ACK_IN_PROGRESS in response." doesn't read very meaningful to me, reads like "If the car is yellow, then the car is yellow" :)

hamishwillee:
Fixes #1140. 

Provides information about the responses to various PARAM_EXT messages

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4978,19 +4978,19 @@
       <field type="uint32_t" name="sw_vcs_commit">Version control system (VCS) revision identifier (e.g. git short commit hash). Zero if unknown.</field>
     </message>
     <message id="320" name="PARAM_EXT_REQUEST_READ">
-      <description>Request to read the value of a parameter with the either the param_id string id or param_index.</description>
+      <description>Request to read the value of a parameter with either the param_id string id or param_index. Response should be a PARAM_VALUE message containing the requested parameter.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="int16_t" name="param_index">Parameter index. Set to -1 to use the Parameter ID field as identifier (else param_id will be ignored)</field>
     </message>
     <message id="321" name="PARAM_EXT_REQUEST_LIST">
-      <description>Request all parameters of this component. After this request, all parameters are emitted.</description>
+      <description>Request all parameters of this component. The component should respond by emitting all parameters (in PARAM_VALUE messages).</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
     </message>
     <message id="322" name="PARAM_EXT_VALUE">
-      <description>Emit the value of a parameter. The inclusion of param_count and param_index in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout.</description>
+      <description>Emit the value of a parameter in response to a PARAM_EXT_REQUEST_READ or PARAM_EXT_REQUEST_LIST message. The inclusion of param_count and param_index in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout.</description>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="char[128]" name="param_value">Parameter value</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
@@ -4998,7 +4998,7 @@
       <field type="uint16_t" name="param_index">Index of this parameter</field>
     </message>
     <message id="323" name="PARAM_EXT_SET">
-      <description>Set a parameter value. In order to deal with message loss (and retransmission of PARAM_EXT_SET), when setting a parameter value and the new value is the same as the current value, you will immediately get a PARAM_ACK_ACCEPTED response. If the current state is PARAM_ACK_IN_PROGRESS, you will accordingly receive a PARAM_ACK_IN_PROGRESS in response.</description>
+      <description>Set a parameter value. In order to deal with message loss (and retransmission of PARAM_EXT_SET), when setting a parameter value and the new value is the same as the current value, you will immediately get a PARAM_ACK_ACCEPTED response. If the current state is PARAM_ACK_IN_PROGRESS, you will accordingly receive a PARAM_ACK_IN_PROGRESS response.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>


message_definitions/v1.0/common.xml
```diff
@@ -4978,27 +4978,27 @@
       <field type="uint32_t" name="sw_vcs_commit">Version control system (VCS) revision identifier (e.g. git short commit hash). Zero if unknown.</field>
     </message>
     <message id="320" name="PARAM_EXT_REQUEST_READ">
-      <description>Request to read the value of a parameter with the either the param_id string id or param_index.</description>
+      <description>Request to read the value of a parameter with either the param_id string id or param_index. Response should be a PARAM_VALUE message containing the requested parameter.</description>
```
hamishwillee: @dogmaphobic Obviously this is the response to a valid id or index. What happens if the id is invalid?

message_definitions/v1.0/common.xml
```diff
@@ -4978,27 +4978,27 @@
...
       <field type="char[128]" name="param_value">Parameter value</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint16_t" name="param_count">Total number of parameters</field>
       <field type="uint16_t" name="param_index">Index of this parameter</field>
     </message>
     <message id="323" name="PARAM_EXT_SET">
-      <description>Set a parameter value. In order to deal with message loss (and retransmission of PARAM_EXT_SET), when setting a parameter value and the new value is the same as the current value, you will immediately get a PARAM_ACK_ACCEPTED response. If the current state is PARAM_ACK_IN_PROGRESS, you will accordingly receive a PARAM_ACK_IN_PROGRESS in response.</description>
+      <description>Set a parameter value. In order to deal with message loss (and retransmission of PARAM_EXT_SET), when setting a parameter value and the new value is the same as the current value, you will immediately get a PARAM_ACK_ACCEPTED response. If the current state is PARAM_ACK_IN_PROGRESS, you will accordingly receive a PARAM_ACK_IN_PROGRESS response.</description>
```
hamishwillee: https://github.com/mavlink/mavlink/issues/1140#issue-446519713 makes the point that this is not helpful:

> If the current state is PARAM_ACK_IN_PROGRESS, you will accordingly receive a PARAM_ACK_IN_PROGRESS response.

Generally the internal current states are not necessarily documented or exposed. More useful to say something like:

> If a system receives a request to change a particular parameter while an existing request is in progress, it will return a PARAM_EXT_ACK message with result PARAM_ACK_IN_PROGRESS.

Is that what you mean? Or will it return that message/result if ANY param is being set?

olliw42: it seems you are addressing me, I unfortunately have no useful answer though, all I can say is:

Your modified statement does make sense to me, and what it describes also makes sense to me.

If it however precisely describes what the real intention is, I can't say. Some MAVLink expert would have to jump in.

hamishwillee:
Thanks @olliw42 . I added you as the person with the question, and @dogmaphobic who authored the protocol. 
It looks much more robust than the ordinary param protocol. We'll have to check if it covers the 0 params case.

olliw42:
I think with the change suggested in the comment https://github.com/mavlink/mavlink/pull/1143#discussion_r286335432, this would be good, isn't it?

the 0 params case I think was answered before elsewhere by means of a timeout

message_definitions/v1.0/common.xml
julianoes: @hamishwillee I think you misunderstood what `ACK_IN_PROGRESS` is for. If you get an `ACK_IN_PROGRESS` it means that the set param has been accepted but is not in effect yet. This means you should wait longer / raise the timeout. It is needed for instance to change some camera setting which takes a while to propagate.

olliw42: ah... many thx for the clarification
could you pl also describe what comes next after the component having responded with an ACK_IN_PROGRESS? The GCS will timeout as usual and resend a request, to which the component can answer as usual?

julianoes: When "progress" has been made and the camera is finished with the long task it shall answer with [PARAM_ACK_ACCEPTED](https://mavlink.io/en/messages/common.html#PARAM_ACK_ACCEPTED).

olliw42: so, it should NOT wait for the GCS to re-request, but do as follows:
* decide if the response will take long or short
* if it will take short respond with an ACK_ACCEPTED
* it it will take long, respond with an ACK_IN_PROGRESS, wait until done, and then send out a ACK_ACCEPTED

??

Doesn't this incur timing issues in two places: First, in the decision if it's "short" or "long" it needs to have an idea of the timeout time of the GCS, since if it knows that it will be sufficiently faster than the GCS timeout it would not go the more complicated route. In any case, if a decision is made on "long" or "short" some established criteria must exist. Second, what is the GCS doing if it receives an ACK_IN_PROGRESS, but no follow-up ACK_ACCEPTED? To not hang, there must be again a timeout mechanism, which again leads to a timing question for the component.

??

julianoes: Correct.

Now for the timing that's up to discussion. My recommendation would be to have a timeout of 1 second. If something takes more than 0.1 second, so much longer than the ping response then send an ack.
And if something takes even longer than a timeout of 1 second, then you should just answer with `ACK_IN_PROGRESS` every 0.5 seconds until done.

If no final `ACK_ACCEPTED` comes through, it's a timeout and a fail.

olliw42:
>No, there should be no responses sent without a corresponding request. The GCS sends a command and starts its normal timeout. When it gets a ACK_IN_PROGRESS response, it extends the timeout and resets the try count. It will continue sending the request as usual until it receives a definitive answer (or really times out and/or exceeded the retry count).

great. That's what I was thinking after julianoes' first comment, but his 2nd comment seemed to not align. Many thx for these details, all clear now I guess.

message_definitions/v1.0/common.xml
olliw42: answered below by @dogmaphobic

olliw42: ähm ... you guys are doing a good job in confusing me ... this seems to contradict what @dogmaphobic was just saying, doesn't it !?!?!?

olliw42:
> For the current implementation you get nothing. No response. That's based on the assumption the code doing the request is only able to do so if it has a list of parameters provided by the camera. There is no other way to request a parameter unless you received the list from the camera.

so, PARAM_EXT_REQUEST_READ is actually totally non functional ???

I don't think this is what it is

olliw42:
yeah, I'm glad to hear it has a function :)
could you then please be so kind to explain again what you were meaning with

```
<message id="320" name="PARAM_EXT_REQUEST_READ">
<description>Request to read the value of a parameter with either the param_id string id or param_index. Response should be a PARAM_VALUE message containing the requested parameter.</description>
```

> For the current implementation you get nothing. No response. That's based on the assumption the code doing the request is only able to do so if it has a list of parameters provided by the camera. There is no other way to request a parameter unless you received the list from the camera.

I really fail to get the point, sorry for it

btw, I would have wanted to respond directly in the above to https://github.com/mavlink/mavlink/pull/1143#discussion_r286454651, which would have helped with following the thread, but somehow I couldn't figure out how to do it, I e.g. don't see a Reply... field to reply, unlike to the other comments in the above there I do get the reply field. Sorry for that inconvenience.

olliw42:
ah, I see, the context got shuffled ... happens :)

I think you just gave a nice summary in https://github.com/mavlink/mavlink/issues/1181#issuecomment-525110717

I still find these two statements concerning PARAM_ACK_IN_PROGRESS contradicting though :

>No, there should be no responses sent without a corresponding request. The GCS sends a command and starts its normal timeout. When it gets a ACK_IN_PROGRESS response, it extends the timeout and resets the try count. It will continue sending the request as usual until it receives a definitive answer (or really times out and/or exceeded the retry count).

> The idea behind PARAM_ACK_IN_PROGRESS is that if something takes too long to set, the camera will respond with this (and param_value will contain the current value it is set to) and once the set operation is completed, it would send a new PARAM_ACK message with a proper conclusion (following one of the three options above).

the first seems to say that the followup PARAM_ACK response should be only send upon the next request, while the second seems to say that the followup PARAM_ACK response is send then the component determines it is done

olliw42:
ok, great, thx for clarifying
I have no preference either
the discussion shows however that some better docu is in need :)

julianoes:
@hamishwillee what should we do about this?

julianoes:
I just tried to resolve the conflicts here and it looks like the fixes here have also been made in #1442, so this can get closed.

hamishwillee:
Thanks, saved me the review task too :-)

PR #1146, opened 2019-05-27
Add a constant to MAV_AUTOPILOT for a COEX Pelican charging station

dvornikov-aa:
COEX develops a MAVLink-compatible charging station which is configurable through QGroundControl. For a better integration we need a MAV_AUTOPILOT constant for our project instead of using MAV_AUTOPILOT_GENERIC.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -65,6 +65,9 @@
       <entry value="19" name="MAV_AUTOPILOT_AIRRAILS">
         <description>AirRails - http://uaventure.com</description>
       </entry>
+      <entry value="20" name="MAV_AUTOPILOT_COEX_CS">
+        <description>COEX Pelican charging station - http://copterexpress.com/pelican</description>
+      </entry>
     </enum>
     <enum name="MAV_TYPE">
       <description>MAVLINK system type. All components in a system should report this type in their HEARTBEAT.</description>


hamishwillee:
@dvornikov-aa Can't you get sufficient differentiation by setting the `type` to [MAV_TYPE_CHARGING_STATION](https://mavlink.io/en/messages/common.html#MAV_TYPE_CHARGING_STATION). The correct autopilot for a charging station is [MAV_AUTOPILOT_INVALID](https://mavlink.io/en/messages/common.html#MAV_AUTOPILOT_INVALID).

dvornikov-aa:
@hamishwillee We use MAV_TYPE_CHARGING_STATION in FirmwarePlugin, but, unfortunately, that's not enough to provide a full integration with QGroundControl. QGC uses AutopilotPlugin for a vehicle components setup, also MAV_AUTOPILOT is used to form a parameters cache file name in QGC. 

Conceptually, I agree that it looks strange, but we need a way to distinguish a different software for the charging stations from different developers.

hamishwillee:
@dvornikov-aa It does look strange. I'll raise it with the dev call tonight to see if there are any alternatives.

LorenzMeier:
Hey - couldn't the charging station interface be made generic? Could you join the dev call so we could have an informed discussion around?

dvornikov-aa:
@LorenzMeier Charging station is not as simple as it seems. It is important for us to support a MAVLink interface for drone-station communication (station control, station position and status data). Our charging station is modular (different customers want different configuration). Most of the modules need configuration and we using QGroundControl for that.

I suppose we can join Dev Call but we need time to arrange it. @okalachev @korigod @sfalexrog

hamishwillee:
FYI The call was last night, about when @LorenzMeier made [his post above](https://github.com/mavlink/mavlink/pull/1146#issuecomment-496816265). Next call is in two weeks at about the same time. 

@LorenzMeier might be able to find some other time to discuss with you.

hamishwillee:
I understand that you guys were able to talk to Lorenz. What was the resolution with respect to this PR?

LorenzMeier:
We need to come up with a vehicle classification to resolve it.

dvornikov-aa:
@hamishwillee We are going to present our charging station description for the further discussion in this PR soon.

DonLakeFlyer:
An important thing to note here is that the QGC architecture is not really set up to create a new "type" of firmware plugin. Especially one which isn't actually a vehicle. It is only coded to support Generic/PX4/ArduPilot and it expects the things it is talking to on the other side to be a vehicle. The FirmwarePlugin interface is only meant to allow you to extend the existing plugin implementations for custom builds for specific commercial vehicles. The creation of a whole new plugin type is likely going to run into all sorts of problems since it was not meant to be used that way and isn't supported.

dvornikov-aa:
@LorenzMeier @hamishwillee [The short description of our charging station](https://docs.google.com/document/d/1l4dCsAkRf1zg9mnzLgc3NiToZqaobu-DkUqmG-5_1IQ/edit?usp=sharing).

hamishwillee:
@LorenzMeier This is in your hands as I have not been involved with your further dicussion. @DonLakeFlyer FYI, see https://github.com/mavlink/mavlink/pull/1146#issuecomment-499137027

hamishwillee:
@LorenzMeier I still don't understand why this shouldn't be done using MAV_TYPE_CHARGING_STATION with autopilot type MAV_AUTOPILOT_INVALID.

I'm proposing to reject this next time I come to this issue - unless I can get a clear explanation of why it is needed (doesn't sound like it will work from Don's post). Alternatively you can of course accept it if you understand the ramifications.

dvornikov-aa:
@hamishwillee Drone charging station is not just a battery charger and a lid controller. There is a lot of additional equipment installed on it, like a weather station, GPS RTK, etc. I kindly doubt that it's possible to create a standard for such type of device. In this case we need a way to determine a vendor (or better a software type) of the charging station and MAV_AUTOPILOT is a simple solution to do that.

hamishwillee:
Thanks @dvornikov-aa. You may end up with more problems than you solve, simply because you'll go from the case where GCSs know _something_ about your autopilot to one where they they know nothing.

The arguments about "other stuff" on the station don't particularly convince me; they could be components on the bigger system. I'm not saying you're wrong, but since I only merge stuff I am comfortable with I have deferred this to @LorenzMeier or @julianoes.

DonLakeFlyer:
> You may end up with more problems than you solve, simply because you'll go from the case where GCSs know _something_ about your autopilot to one where they they know nothing.

That is a good point. QGC won't connect to something which reports like that. Whereas mavlink generic will at least get you parameter editing support at a minimum. But maybe that's not really valuable since the other stuff is going to end up being quite strange.

dvornikov-aa:
@DonLakeFlyer We have a working prototype of the charging station and a QGC branch with charging station support. Charging station implements the approach I [described above](https://github.com/mavlink/mavlink/pull/1146#issuecomment-499137027). We have been testing this setup for a year in the field and have mostly positive experience.

hamishwillee:
Following discussion last night in DevCall we have decided to close this.

From a MAVLink perspective the right way to do this is to use [MAV_TYPE_CHARGING_STATION](https://mavlink.io/en/messages/common.html#MAV_TYPE_CHARGING_STATION) (with autopilot type [MAV_AUTOPILOT_GENERIC](https://mavlink.io/en/messages/common.html#MAV_AUTOPILOT_GENERIC)).

We appreciate that the original reason for this approach was to address limitations in how QGroundControl architecture handles the autopilot types. However the right way to address that is to patch QGroundControl.

PR #1147, opened 2019-05-29
Manual Control add 4 more joystick axes (channels 5-8)

m1kelyons:
Proposal to extend the manual_control MAVLink message for extra channels. See [RFC 0015 - Mavlink MANUAL_CONTROL message extension PR](https://github.com/mavlink/rfcs/pull/16)


hamishwillee:
@m1kelyons 

Thanks for attending the meeting last night. Here is summary of concerns:
  - This turns the MANUAL_CONTROL message which has unambiguous purpose independent of flight stack into one that requires separate parameter support/interpretation for flight stacks - ie it is more like RC_OVERRIDE
  - Also there are other messages that do the things this is intended for, so we can use those, and if needed add missing cases.
  - We would consider an extension where the purpose of each channel is defined in an enum (ie so there is no flight-stack specific params etc), but leaning towards extra messages to reduce the cost for cases where this is not needed.

The discussion faded out a bit for me at this point, but I believe that the action was to close this and for you to take some more time to think about it. Is that your understanding Mike?

LorenzMeier:
@m1kelyons Any further input here? The more defined we can make it, the more sense it makes to pull it in.

hamishwillee:
@m1kelyons  ^^^ ?

hamishwillee:
@LorenzMeier I propose we close this unless @m1kelyons has further comment?

hamishwillee:
Closing. Inactive.

PR #1154, opened 2019-06-04
[WIP] Common.xml: Clarify Protocol version protocol

hamishwillee:
The service for getting protocol is not clearly defined

Problems this tries to fix
1. PROTOCOL_VERSION marked as WIP
2. PROTOCOL_VERSION indicates it is a response to REQUEST_PROTOCOL_VERSION message, which is not in common.xml
3. There is MAV_CMD_REQUEST_PROTOCOL_VERSION which would seem like the correct response type BUT this is a command, and commands are supposed to be addressed to a particular target. See below for request for clarification.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1566,7 +1566,7 @@
         <param index="2" label="Index ID" minValue="0" increment="1">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
       </entry>
       <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION" hasLocation="false" isDestination="false">
-        <description>Request MAVLink protocol version compatibility</description>
+        <description>Request MAVLink protocol version compatibility. Systems should ACK request and then broadcast the PROTOCOL_VERSION message.</description>
         <param index="1" label="Protocol">1: Request supported protocol versions by all nodes on the network</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
@@ -4962,9 +4962,7 @@
       <field type="char[64]" name="password">Password. Leave it blank for an open AP.</field>
     </message>
     <message id="300" name="PROTOCOL_VERSION">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Version and capability of protocol version. This message is the response to REQUEST_PROTOCOL_VERSION and is used as part of the handshaking to establish which MAVLink version should be used on the network. Every node should respond to REQUEST_PROTOCOL_VERSION to enable the handshaking. Library implementers should consider adding this into the default decoding state machine to allow the protocol core to respond directly.</description>
+      <description>Version and capability of protocol version. This message is the response to MAV_CMD_REQUEST_PROTOCOL_VERSION and is used as part of the handshaking to establish which MAVLink version should be used on the network. Every node should respond to MAV_CMD_REQUEST_PROTOCOL_VERSION to enable the handshaking. Library implementers should consider adding this into the default decoding state machine to allow the protocol core to respond directly.</description>
       <field type="uint16_t" name="version">Currently active MAVLink version number * 100: v1.0 is 100, v2.0 is 200, etc.</field>
       <field type="uint16_t" name="min_version">Minimum MAVLink version supported</field>
       <field type="uint16_t" name="max_version">Maximum MAVLink version supported (set to the same value as version by default)</field>


message_definitions/v1.0/common.xml
```diff
@@ -4962,9 +4962,7 @@
...
     </message>
     <message id="300" name="PROTOCOL_VERSION">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Version and capability of protocol version. This message is the response to REQUEST_PROTOCOL_VERSION and is used as part of the handshaking to establish which MAVLink version should be used on the network. Every node should respond to REQUEST_PROTOCOL_VERSION to enable the handshaking. Library implementers should consider adding this into the default decoding state machine to allow the protocol core to respond directly.</description>
+      <description>Version and capability of protocol version. This message is the response to MAV_CMD_REQUEST_PROTOCOL_VERSION and is used as part of the handshaking to establish which MAVLink version should be used on the network. Every node should respond to MAV_CMD_REQUEST_PROTOCOL_VERSION to enable the handshaking. Library implementers should consider adding this into the default decoding state machine to allow the protocol core to respond directly.</description>
```
hamishwillee: @LorenzMeier @julianoes The above change removes wip and changes message name from REQUEST_PROTOCOL_VERSION (not defined in common.xml) to MAV_CMD_REQUEST_PROTOCOL_VERSION (is defined). But MAV_CMDs are not generally broadcast messages - so if we put a broadcast target ID in this then the system will be bombarded with ACK responses as well as the versions.

OK, so how is this supposed to work? There are two ways I can see:
1. There is a missing message REQUEST_PROTOCOL_VERSION that is supposed to be used for getting all responses
2. MAV_CMD_REQUEST_PROTOCOL_VERSION is the correct message for getting all responses and it can choose just to ignore ACKs
3. There are both REQUEST_PROTOCOL_VERSION  and MAV_CMD_REQUEST_PROTOCOL_VERSION  but the later is just for querying a single system.

julianoes: I think 2. makes sense. Ignoring ACKs is easy.

olliw42: if I may add my 2 cents
if "ignore ACKs" means to not send ACKs, then I wouldn't be fond of 2., as it kind of breaks the general pattern of how to respond to cmds. I started to believe to see a pattern here, and much of the docs make sense only if that pattern can be assumed. I like patterns ;)

hamishwillee: @olliw42 Not quite. Basically the pattern for commands is that they are "really" intended to go to one recipient which must return an ACK (either immediately or after doing something), and may then do other things. 
The problem here is that this is a broadcast message - which is outside normal use of the command protocol. 

What should happen is that every system returns an ACK. But this sux because I suspect normal command handling might be a bit confused by getting back multiple acks from a single request. Also it is a lot of unnecessary traffic. 

So for this case, I would actually expect a specific message be used rather than a command. 

See https://github.com/mavlink/mavlink/pull/1155#issuecomment-498618967 for an alternative - @julianoes thoughts? We introduced [MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE) for this purpose, but actually `REQUEST_MESSAGE` as proposed is more fit for purpose.

julianoes: @hamishwillee oh so `MAV_CMD_REQUEST_MESSAGE` is the generic variant. I didn't know that. It makes sense to use that one more and more than and slowly deprecate individual ones, I suppose.

olliw42: I think I understood what should happen and the pain points ("sux"). I would consider it a sign of a poorly designed spec if there were per-command exceptions. Patterns are good, since they imply a concept. Maybe one could restore at least some sort of pattern = generality = concept by saying that non-broadcast commands should send an ACK while broadcast commands should not send an ACK.
I think your argument is not yet totally complete, since it's not only about broadcast/non-broadcast but also about ensured/non-ensured transport, which I believe the ACKs are for. I would link that to whether a component is required to support PROTOCOL_VERSION or not, if required I would expect an ensured transmission, i.e. ACKs, if voluntary when not. 
You too seem to not favor 2., so we seem to be on the same page. I guess if I would have to vote I would vote for a REQUEST_PROTOCOL_VERSION message with target fields, i.e. exactly what ArduPilot has, and I would think that support of this should be voluntary.

auturgy:
Suggest closing.

hamishwillee:
Done. I think more thought could go into it. But as this particular message is deprecated :-)

PR #1155, opened 2019-06-04
Common.xml: Clarify MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES

hamishwillee:
Clarifies/cross links MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES and AUTOPILOT_VERSION.

This is what PX4 appears to do, and is also logical.

FYI @auturgy

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1571,7 +1571,7 @@
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="520" name="MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES" hasLocation="false" isDestination="false">
-        <description>Request autopilot capabilities</description>
+        <description>Request autopilot capabilities. The receiver should ACK the command and then emit its capabilities in an AUTOPILOT_VERSION message</description>
         <param index="1" label="Version">1: Request autopilot version</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Also @auturgy - if you're OK with this update, you might want to consider removing the ArduPilot message AUTOPILOT_VERSION_REQUEST  that does the same thing but not as a command.

I'm not strictly convince that a command is the "right approach". After all, the returned capabilities are their own response. Also, as discussed (somewhere?) I'd much rather we had a generic requester function that you put in the message ID in and it caused the recipient to broadcast that message (ie one off interval).

auturgy:
It’s not clear to me that the ArduPilot implementation isn’t better.
Perhaps pulling the ArduPilot message into common would be the right approach.

Regards,

James

hamishwillee:
@auturgy Personally I think the ArduPilot message makes more sense and saves an unnecessary ACK. It may be there are other concerns here to discuss. 

I'm leaning towards deprecating both and moving towards a  _general_ requestor (e.g. like [MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE) but without the command)

So replace both messages with something like :

```xml
    <message id="183" name="REQUEST_MESSAGE">
      <description>Request a specified message, by id, from a particular system/component (or all components using broadcast id).</description>
      <field type="uint32_t" name="message_id">Message ID.</field>
      <field type="uint8_t" name="target_system">System ID.</field>
      <field type="uint8_t" name="target_component">Component ID.</field>
    </message>
```

Either way, the MAV_CMD should still be changed as I've indicated because it can't just be deprecated.

hamishwillee:
I'd be happy to merge this as is - ie it is IMO an accurate description of what is supposed to happen. 

HOwever leaving this open so we can discuss getting a common behaviour in next dev call.

LorenzMeier:
What is missing in the discussion is that commands have in-built retransmission.

PR #1161, opened 2019-06-11
add waypoint type extension to TRAJECTORY_REPRESENTATION_WAYPOINT

mrivi:
This PR adds the ability to add the type associated with each waypoint e.g. land, takeoff.... 
This information is available also in the mission item however the two messages aren't synchronized in ROS.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5077,6 +5077,7 @@
       <field type="float[5]" name="acc_z" units="m/s/s">Z-acceleration of waypoint, set to NaN if not being used</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw angle, set to NaN if not being used</field>
       <field type="float[5]" name="vel_yaw" units="rad/s">Yaw rate, set to NaN if not being used</field>
+      <field type="uint16_t[5]" name="command" enum="MAV_CMD">Scheduled action for each waypoint, UINT16_MAX if not being used.</field>
     </message>
     <message id="333" name="TRAJECTORY_REPRESENTATION_BEZIER">
       <wip/>


julianoes:
Could you specify which enum this would use?

I assume this would be [MAV_CMD](https://mavlink.io/en/messages/common.html#MAV_CMD), so:
```
enum="MAV_CMD"
```
And it looks like it should be an `uint16_t` for this.

Like here: https://github.com/mavlink/mavlink/blob/06030fb4fd8e5e304e3fdd152957d0673842dbb5/message_definitions/v1.0/common.xml#L3526

mrivi:
@julianoes can we get this in?

julianoes:
I will review again and then we might want a review from someone else before merging.

message_definitions/v1.0/common.xml
```diff
@@ -5077,6 +5077,8 @@
       <field type="float[5]" name="acc_z" units="m/s/s">Z-acceleration of waypoint, set to NaN if not being used</field>
       <field type="float[5]" name="pos_yaw" units="rad">Yaw angle, set to NaN if not being used</field>
       <field type="float[5]" name="vel_yaw" units="rad/s">Yaw rate, set to NaN if not being used</field>
+      <extensions/>
+      <field type="uint16_t[5]" name="type" enum="MAV_CMD">Waypoint type, -1 if unknown.</field>
```
julianoes: An `unsigned` can't really be -1, I guess this needs to be `UINT16_MAX` then for unknown.

mrivi:
@julianoes done, who can do another review?

mrivi:
@amilcarlucas I adressed your comment

mrivi:
@amilcarlucas done

mrivi:
@julianoes @amilcarlucas @hamishwillee could we get this in?

amilcarlucas:
I would like to better understand the use case of this change. How does this improve ROS integration ?

mrivi:
It enables to have syncronization. Right now I have to subscribe to the list of mission items but know the command associated with the waypoint e.g. when it's a land waypoint and the trajectory_representation_waypoint to have the waypoint in local coordinates. However it can happen that the waypoint update happens before the command or vice versa and it complicates things on the ROS side.

hamishwillee:
@mrivi Why have you made it an `extension` - this message is still `<wip>` so you are allowed to break compatibility, and IMO you should do so.

Otherwise this makes perfect sense to me.

amilcarlucas:
When you do the change requested be @hamishwillee pleas also replace "unused" with "not being used". For cosistency reasons :)

mrivi:
@hamishwillee @amilcarlucas done :)

amilcarlucas:
Thanks

PR #1163, opened 2019-06-19
Fix typo in MAV_CMD_DO_FOLLOW description

mortenfyhn:
Found a minor typo while looking through some message definitions: I assume "_Being_ following a target" should be "_Begin_ following a target". (https://mavlink.io/en/messages/common.html#MAV_CMD_DO_FOLLOW)

I felt it would be overkill to open an issue for this so there is no connected issue to this PR.


hamishwillee:
Thanks @mortenfyhn - much appreciated.

Issue #1171, opened 2019-07-03
Mission checksum package


holmnikolaj:
Instead of downloading all the mission from the drone to check that they are in sync it would be much faster if the mission started with a checksum item. As far as I can see there is no such thing at the moment but this would be very helpful for setups with long missions and poor link quality.

dagar:
Very interested in this as well. 

@antiheavy @DonLakeFlyer

Antiheavy:
Yes!  And helpful with knowing if the flight plan on the vehicle is the same or different than what is onscreen (eg QGC).

hamishwillee:
PX4 and QGC do something similar for parameters. It would be nice if we could use the 0th index for this, but ArduPilot populate that with the home position.

holmnikolaj:
It could be something as simple as a mission item that is simply ignored by the autopilot. It is only the UI that should be in charge of synchronization anyway. 
Then the seven parameters (or just some of them) can contain the checksum. 

1. The UI will upload the checksum mission item  as part of the mission if it supports it.
2. The Autopilot completely ignores it.
3. The UI reads the checksum mission item from the autopilot and verifies that the mission is in sync. 

The checksum item could be placed first in the mission of second (just after takeoff) if this is more convenient.

JohannesFriis:
I guess it could be done similar to MAV_CMD_DO_LAND_START (#189 ). This acts as a marker in the mission waypoint list. I know that making it a command has a double purpose, that makes sense for land start cmd, but the idea of a marker wp in the mission list containing the mission checksum could be implemented with small changes in both mavlink and autopilot's. i agree that placing the marker early in the mission list is preferred.

dagar:
I think it needs to be handled like metadata across multiple parts of the mission protocol to solve most of the problems efficiently. I'd also like to consider the checksum itself being part of the spec.

For example, `MISSION_CURRENT` (https://mavlink.io/en/messages/common.html#MISSION_CURRENT) is what the vehicle broadcasts so that the GCS can update the active waypoint. It could be extended to also broadcast the ID (mission checksum).

We should be able to request a checksum for an entire mission, subset range, or individual mission_item. That could also enable incremental sync.

LorenzMeier:
@dagar Proposal makes total sense. @holmnikolaj do you want to extend your PR to include a full proposal?

hamishwillee:
[...]

PR #1172, opened 2019-07-04
Added mission checksum message

holmnikolaj:
fixes:
https://github.com/mavlink/mavlink/issues/1171


message_definitions/v1.0/common.xml
```diff
@@ -1532,6 +1532,18 @@
         <param index="1" label="First Item" minValue="0" increment="1">first_item: the first mission item to run</param>
         <param index="2" label="Last Item" minValue="0" increment="1">last_item:  the last mission item to run (after this item is run, the mission ends)</param>
       </entry>
+      <entry value="350" name="MAV_CMD_MISSION_CHECKSUM" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Calculate mission checksum and put in beginning of mission to easily test mission synchronisation</description>
+        <param index="1" label="Checksum bytes 1 and 2"> </param>
```
hamishwillee: Any particular reason to only pack 2 bytes per index? These are always sent in 4 byte fields. Index 1-4,7 are always packed as floats, while index 6,7 may be packed as float or int32, depending on if sent as MISSION_ITEM or MISSION_ITEM_INT.

hamishwillee:
@DonLakeFlyer @julianoes Don't know if you have been following discussion, but this is a mission item that includes a checksum to make it easier/faster to determine whether mission on vehicle matches GCS/API. See https://github.com/mavlink/mavlink/issues/1171 for broader discussion.

julianoes:
I would probably prefer a separate `MISSION_...` message that fits with the rest of the mission protocol instead of a command.

message_definitions/v1.0/common.xml
holmnikolaj: No - that was a blunder from my side. 
Also I don't think 28 bytes is necessary. Maybe just index 6 and 7 to not use floats as checksum and leave the rest open for possible future extension. Any thoughts on that?

hamishwillee: Arg, I mean index 5 and 6 can be ints (sorry, corrected above and below)

> Maybe just index 5 and 6 to not use floats as checksum and leave the rest open for possible future extension. 

I'd probably use anything other than 5 and 6, because then handling is same irrespective of what mission item type is used.

But, I think we're more likely to end up with a separate set of messages rather than a mission item like this because it is much more flexible (as per the ongoing discussion in the linked issue). You've opened a (good) can of worms :-).

DonLakeFlyer:
I would just call this thing a unique id for the mission. If the GCS is in control of generating it and there is no spec as to how to do that then it can be done using whatever means is appropriate. It could be a database key or something like that in some cases. The only responsibility of the vehicle is to store all 7 params and send then back to the gcs on download.

message_definitions/v1.0/common.xml
```diff
@@ -1532,6 +1532,18 @@
         <param index="1" label="First Item" minValue="0" increment="1">first_item: the first mission item to run</param>
         <param index="2" label="Last Item" minValue="0" increment="1">last_item:  the last mission item to run (after this item is run, the mission ends)</param>
       </entry>
+      <entry value="350" name="MAV_CMD_MISSION_UID" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Unique mission ID to be put in beginning of a mission to easily test mission synchronisation.</description>
+        <param index="1" label="First 4 bytes of the unique mission ID"> </param>
+        <param index="2" label="Last 4 bytes of the unique mission ID"> </param>
+        <param index="3">Reserved</param>
```
hamishwillee: IMO best to simply not define the reserved params: https://mavlink.io/en/guide/define_xml_element.html

This can of course wait on the end of the discussion on "design".

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3650,14 @@
       <field type="uint8_t" name="origin_compid" enum="MAV_COMPONENT">Compnent ID of the author of the new mission.</field>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
+    <message id="53" name="MISSION_UID">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Unique mission ID for the current mission. The UID is set in a mission with MAV_CMD_MISSION_UID. When requesting the MISSION_UID with MAV_CMD_REQUEST_MESSAGE then use mission_type as param 2 of that message.</description>
+      <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
+      <field type="float" name="uid_1">First 4 bytes of the unique mission ID</field>
```
amilcarlucas: Encoding 4 bytes as float might give strange results. I advise using int_32 here.
Besides that 32 bit already provides a decent Unique ID why do you need 64bits?

hamishwillee: > Encoding 4 bytes as float might give strange results. I advise using int_32 here.

@amilcarlucas Good point. Initial discussion was sending in a MAV_CMD, where float made sense. But in our own message this could even be an int64_t or uint64_t (or equivalent 32 bit storage if those were used).

IamPete1:
Any update on this? It would solve the problem I was trying to solve with #1527 if it was to also include a enum for mission type, ie mission, fence, rally ect. Seems like that would be a simple addition and then this would be a useful feature, not sure why this has not been merged already?

holmnikolaj:
It got stale because the solution we agreed upon was based on the, at the time, not yet implemented mavlink microservices protocol. I don't know the status of it but @julianoes was tasked to implement it.

julianoes:
@holmnikolaj right. And then @ItsTimmy worked on the implementation and had a first proof of concept ready to review and test:
https://github.com/PX4/PX4-Autopilot/pull/13565
https://github.com/mavlink/qgroundcontrol/pull/8065

Unfortunately, it seemed like no one cared to move it forward from there and it was abandoned and eventually closed :disappointed:.

message_definitions/v1.0/common.xml
hamishwillee: @holmnikolaj Thoughts on this?
- Do we need 64 bit ids or would 32 bits be fine?
- What is easiest for an implementer? What is important is that this is being set by the send in a float - I assume you're making a float output so that at the end of a round trip they can just do a comparison?

hamishwillee:
@IamPete1 @holmnikolaj Let's do an iteration and get this in. It has been a while, so let's summarise the thinking and confirm we're happy. 

The proposal as I recall (please confirm):
- Mission includes a new MAV_CMD for the id.
- This gets set by GCS on mission upload. The method used to determine the UID value is entirely up to the GCS : TBD 
- GCS can query current UID value using generic requestor method (UID emitted in a new message)
- There is a bit to indicate support for this feature.

Notes/Discussion:
- Should the MISSION_UID should also be emitted by the flight stack when the mission changes (for any reason). This gives interested systems the chance to know that a mission has changed sooner.
- The UID needs to change every time the mission changes - including partial updates. How do we expect this to be handled? 
  - One suggestion is that if a system supports this mechanism and gets a mission without a UID, then internally the UID should be set by the to a UID that means "invalid" (by convention). Similarly if there was a partial update.
- The GCS can set whatever UID it likes. Perhaps we should make it clear that this shouldn't just be an iterated value - otherwise two systems are likely to clash.

Originally we discussed having the UID as a hash, calculated by the ground station and not done as a mission item at all. Flight stack generates mission hash on every mission change and emits a new mission UID as part of the `MISSION_ACK` (also emits on request). Can anyone remember why we decided against?

hamishwillee:
Further, under new governance approach, we're trying to only include PRs where there is going to be an actual flight stack implementation and commitment to get into another stack. So if we do this, is anyone committed to getting this into PX4/ArduPilot?

LorenzMeier:
For effective implementation: I would propose to hash these elements and to apply the same logic as on the generator:

- frame
- command
- param1
- param2
- param3
- param4
- param5
- param6
- param7

holmnikolaj:
@hamishwillee That summarizes it pretty well. As I recall we didn't agree on how to handle partial upload, but agreed that the proposed solution would not make it more difficult to support it in the future. 

I will not be committed to the issue as we (Sky-Watch) opted for another solution we could implement right away.

IamPete1:
I agree with  Lorenz, it should be a hash of the whole command, and it should be possible to re-hash on the flight controller, then partial upload and scripting in AP will also work.

I think it should definitely be extended for fence and rally points.

From the AP side I can take on the implementation.

hamishwillee:
Discussed in the devcall. We'd like this to use a hash for the id; which implies that the best place for this to be generated is the flight stack.

Proposal is then:
- Flight stack generates _separate_ hash on any change to the mission, fence, or rally points.
  - Things to be part of that hash are (for all items in the mission, fence or rally point list) the frame, command and parameter values as per https://github.com/mavlink/mavlink/pull/1172#issuecomment-733533758
- The generated hash is broadcast in a `MISSION_UID` following completion of an [update to a on-vehicle mission](https://mavlink.io/en/services/mission.html#uploading_mission) - ie just after the ACK.
- The generated hash may also be emitted in response to `MAV_CMD_REQUEST_MESSAGE`
- Note that following this change no explicit mission item added to mission.

The a `MISSION_UID` would look something like (TBD):
  ```
    <message id="53" name="MISSION_UID">
      <description>Unique hash for the current mission, fence, or rally points definition. This is emitted by the flight stack after any change to a mission, fence, or rally point definition, typically after the` MISSION_ACK` is sent, and on request using MAV_CMD_REQUEST_MESSAGE.</description>
      <field type="uint8_t" name="plan_type" enum="MAV_MISSION_TYPE">Plan type for which the hash applies.</field>
      <field type="uint32_t" name="uid">Unique hash of items in specified plan_type.</field>
    </message>
   ```

Points for discussion/questions:
- Is this reasonable?
- Above I have _assumed_ that hash result is uint32_t - is that reasonable - or should it be a 64 bit number?
- Is there a better name for this - ie would MISSION_HASH, or PLAN_HASH be better?
- What is hashing function should be used? SHA-1, MD5 etc etc? If we use one with greater hash length, how do we truncate to fit message?
- Should we also include the hash in the MISSION_ACK message from drone following upload? 
  - If we do this then the GCS would not need to generate the hash, just store whatever it was at upload time.

DonLakeFlyer:
Generating the hash only on the vehicle makes the whole thing a nightmare from a standpoint of how a GCS would use it. A gcs needs to be able to save the uid in the persisted file format. You shouldn't need to roundtrip a mission through the vehicle to be able to do that. That leads to all sort of strange things with respect to having to go back and re-write the persisted format to add the uid after it was already saved. Either the hash generation needs to be spec'ed such that it can be done on both ends. This would be best and is just like the parameter cache spec. Or if that isn't possible then the gcs should be in control but that leads to other problems like the hash is now gcs specific. But that is still better than the vehicle being in control.

IamPete1:
I don't think there was any suggestion that the hash should only be on the vehicle, just that it should be possible on the vehicle. I guess we would just clearly define the method for hashing such that all implementations, both GCS and vehicle, give the same result.

hamishwillee:
> I don't think there was any suggestion that the hash should only be on the vehicle, just that it should be possible on the vehicle. I guess we would just clearly define the method for hashing such that all implementations, both GCS and vehicle, give the same result.

@DonLakeFlyer @IamPete1 
1. Why does the uid need to persist in the plan? If the GCS can reliably calculate it once, it can reliably calculate it as many times as it needs to. I guess it does provide an easy way to compare if two plans are the same.
2. I don't care where the UID is generated as long as it can reliably be used for comparison.
3. I think that a UID calculated on the vehicle is the only really robust way to know the correct hash IF partial updates end up being supported.

message_definitions/v1.0/common.xml
hamishwillee: I'll put in 64 bits and we can argue.

hamishwillee:
@holmnikolaj  Re "I will not be committed to the issue as we (Sky-Watch) opted for another solution we could implement right away." Thanks very much for commenting. Sorry we've been slow on this.

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3638,19 @@
...
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
+    <message id="53" name="MISSION_UID">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Hash for the current mission, rally points or geofence plan (a GCS can use this hash to determine if it has a matching plan defintion).
+        This message must be broadcast following any change to a plan (immediately after the MISSION_ACK that completes the plan upload sequence).
+        It may also be requested using MAV_CMD_REQUEST_MESSAGE, where param 2 indicates the plan type for which the hash is required.
+        The hash must be calculated on the autopilot, but may also be calculated by the GCS.
+        The hash algorithm is XXXX. It is run over each item in the plan in seq order (excluding the home location, if stored), and covers fields:
```
hamishwillee, line 3648: @julianoes Can you please help me word Lorenz' suggestion from [here](https://github.com/mavlink/mavlink/pull/1172#issuecomment-733533758) "hash these elements and to apply the same logic as on the generator: ..." in a way that is unambiguous?

Would you interpret this to be a suggestion to use the CRC-16/MCRF4XX checksum as done on the message name and type, but on each of the field values instead?  I was assuming a SHA or similar?

hamishwillee:
@DonLakeFlyer @IamPete1 OK, so I've updated the wording on this "in source". Essentially as discussed - autopilot emits this following any update, or it can be requested. The autopilot must generate the hash, but the GCS may do so also. I've identified the fields in the hash and the order, and excluded the home position (which might be included in ArduPilot, and which should not really be part of a plan definition). 
I haven't put the name of the actual hash algorithm in there. 

The GCS can generate its hash at any point. It would compare whenever it receives a MISSION_UID. Workflow otherwise looks like
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG4gICAgR0NTLT4-RHJvbmU6IE1JU1NJT05fQ09VTlRcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXRcbiAgICBEcm9uZS0-PkdDUzogTUlTU0lPTl9SRVFVRVNUX0lOVCAoMClcbiAgICBEcm9uZS0-PkRyb25lOiBTdGFydCB0aW1lb3V0XG4gICAgR0NTLS0-PkRyb25lOiBNSVNTSU9OX0lURU1fSU5UICgwKVxuICAgIE5vdGUgb3ZlciBHQ1MsRHJvbmU6IC4uLiBpdGVyYXRlIHRocm91Z2ggaXRlbXMgLi4uXG4gICAgRHJvbmUtPj5HQ1M6IE1JU1NJT05fUkVRVUVTVF9JTlQgKGNvdW50LTEpXG4gICAgRHJvbmUtPj5Ecm9uZTogU3RhcnQgdGltZW91dFxuICAgIEdDUy0tPj5Ecm9uZTogTUlTU0lPTl9JVEVNX0lOVCAoY291bnQtMSlcbiAgICBEcm9uZS0-PkdDUzogTUlTU0lPTl9BQ0tcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBpbW1lZGlhdGVseSBlbWl0IGhhc2ggZm9yIHVwZGF0ZWQgcGxhblxuICAgIERyb25lLT4-R0NTOiBNSVNTSU9OX1VJRChoYXNoLCBwbGFuX3R5cGUpXG5cbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBBdCBzb21lIHBvaW50IGxhdGVyLCB5b3UgbWlnaHQgZ2V0XG4gICAgR0NTLS0-PkRyb25lOiBNQVZfQ01EX1JFUVVFU1RfTUVTU0FHRSAobWlzc191aWQsIHR5cGUpXG4gICAgRHJvbmUtPj5HQ1M6IE1JU1NJT05fVUlEKGhhc2gsIHBsYW5fdHlwZSkiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)

Other than the actual hash algorithm, any suggestions, comments?

julianoes:
I suggest we re-use crc32 that we also use for MAVLink FTP and params.

The CRC32 can be described best as @olliw42 put it:
> The CRC32 calculation used by MAVLink FTP is similar to but different from the ISO 3309 standard based on the polygon 0x04C11DB7, which is also referred to the CRC32 based on Gary Brown's work.

hamishwillee:
Thanks @julianoes - updated.

@auturgy @DonLakeFlyer @holmnikolaj @IamPete1 Any comments. Source now describes how this is used, including algorithm. I am pretty sure this is good to go!

holmnikolaj:
@hamishwillee looks good to me.

IamPete1:
Looks good to me, might be nice to provide a example mission and the expected hash, that is more of a documentation thing tho

hamishwillee:
@auturgy @julianoes YOu OK for this to merge?

Thanks @holmnikolaj @IamPete1 - Yes, it will be nice to have docs of algorithm. I am kind of hoping someone can do the algorithm and I can then copy code for docs :-0. If what is written here isn't sufficient to make a non-ambiguous implementation though, then we do need to keep on working.

hamishwillee:
@holmnikolaj Could you please squash and rebase? I'd do this, but I don't think I'm "allowed" on your branch.

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3638,20 @@
...
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Hash for the current mission, rally points or geofence plan (a GCS can use this hash to determine if it has a matching plan defintion).
+        This message must be broadcast following any change to a plan (immediately after the MISSION_ACK that completes the plan upload sequence).
+        It may also be requested using MAV_CMD_REQUEST_MESSAGE, where param 2 indicates the plan type for which the hash is required.
+        The hash must be calculated on the autopilot, but may also be calculated by the GCS.
+        The hash uses the same CRC32 algorithm as MAVLink FTP (https://mavlink.io/en/services/ftp.html#crc32-implementation).
+        It is run over each item in the plan in seq order (excluding the home location if present in the plan), and covers the following fields (in order):
+        frame, command, autocontinue, param1, param2, param3, param4, param5, param6, param7.
+      </description>
+      <field type="uint8_t" name="plan_type" enum="MAV_MISSION_TYPE">Mission type.</field>
```
julianoes, line 3652: This is called `mission_type in other places.

message_definitions/v1.0/common.xml
```diff
@@ -2556,6 +2556,9 @@
       <entry value="65536" name="MAV_PROTOCOL_CAPABILITY_FLIGHT_INFORMATION">
         <description>Autopilot supports the flight information protocol.</description>
       </entry>
+      <entry value="131072‬" name="MAV_PROTOCOL_CAPABILITY_MISSION_UID">
+        <description>Autopilot supports unique mission ID in the mission.</description>
+      </entry>
```
julianoes, line 2561: What about? 

```suggestion
      <entry value="131072‬" name="MAV_PROTOCOL_CAPABILITY_MISSION_CHECKSUM">
        <description>Autopilot supports mission checksum.</description>
      </entry>
```

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3638,20 @@
       <field type="uint8_t" name="origin_compid" enum="MAV_COMPONENT">Compnent ID of the author of the new mission.</field>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
+    <message id="53" name="MISSION_UID">
```
julianoes, line 3641:
```suggestion
    <message id="53" name="MISSION_CHECKSUM">
```

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3638,20 @@
...
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Hash for the current mission, rally points or geofence plan (a GCS can use this hash to determine if it has a matching plan defintion).
```
julianoes, line 3644:
```suggestion
      <description>Checksum for the current mission, rally points or geofence plan (a GCS can use this checksum to determine if it has a matching plan definition).
```

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3638,20 @@
...
+        This message must be broadcast following any change to a plan (immediately after the MISSION_ACK that completes the plan upload sequence).
+        It may also be requested using MAV_CMD_REQUEST_MESSAGE, where param 2 indicates the plan type for which the hash is required.
+        The hash must be calculated on the autopilot, but may also be calculated by the GCS.
+        The hash uses the same CRC32 algorithm as MAVLink FTP (https://mavlink.io/en/services/ftp.html#crc32-implementation).
+        It is run over each item in the plan in seq order (excluding the home location if present in the plan), and covers the following fields (in order):
```
julianoes, line 3649:
```suggestion
        The checksum must be calculated on the autopilot, but may also be calculated by the GCS.
        The checksum uses the same CRC32 algorithm as MAVLink FTP (https://mavlink.io/en/services/ftp.html#crc32-implementation).
        It is run over each item in the plan in seq order (excluding the home location if present in the plan), and covers the following fields (in order):
```

message_definitions/v1.0/common.xml
```diff
@@ -3635,6 +3638,20 @@
...
+        frame, command, autocontinue, param1, param2, param3, param4, param5, param6, param7.
+      </description>
+      <field type="uint8_t" name="plan_type" enum="MAV_MISSION_TYPE">Mission type.</field>
+      <field type="uint32_t" name="hash">Hash of current plan for specified type.</field>
```
julianoes, line 3653:
```suggestion
      <field type="uint32_t" name="checksum">CRC32 checksum of current plan for specified type.</field>
```

julianoes (changes requested):
I consider this more of a checksum than a uid. I would use a uid for something that survives over a longer time and you want to be able to identify it again.

On the other hand, a checksum I would consider something that you use to check a transfer, or in this case if transfer is required.

Also, we are using a CRC which is a cyclic redundancy *check* and not a hash function, and they supposedly have somewhat different meaning, at least when I read https://eklitzke.org/crcs-vs-hash-functions.

This is not to say I disagree with the concept or implementation, I just think the wording would be better calling it a checksum rather than a uid and hash.

hamishwillee:
Thanks @julianoes . I take your point re `MISSION_CHECKSUM` vs `MISSION_UID`. I believe Don and co do want to includes this in the plan as well. But personally I think this is better because it clear what it really is. 

I'll look at this next week. Hoping others will also review.

holmnikolaj:
I agree with @julianoes on the checksum vs. UID but I thought the general opinion was that UID was preferred.?

DonLakeFlyer:
Naming it `uid` or 'checksum` doesn't matter much to me. That said I would not mix the wording of 'uid' and 'hash' if that is what will be used. Pick one name and stick with it. The 'COMPONENT_INFORMATION' spec uses 'uid' for the similar usage pattern of comparison. But it's slightly different in that the generation is fully vehicle side.

hamishwillee:
@holmnikolaj I think this is good to go. Can you squash all the commits and rebase? (I don't think I can do this on your banch)

holmnikolaj:
image

hamishwillee:
@holmnikolaj Thanks for that. 
The error shown is 
```
./message_definitions/v1.0/common.xml needs formatting - run ./scripts/format_xml.sh ./message_definitions/v1.0/common.xml
```
Basically this means what it says. Can you please rebase then run `./scripts/format_xml.sh ./message_definitions/v1.0/common.xml` (I do this on ubuntu). It is basically a linter for whitespace and other minor issues.

If this is a problem for you let me know and I'll push a PR to your branch with the required updates updates. Note though that probably won't be for a couple of weeks because I am out of office.

@amilcarlucas @IamPete1 Can you please do final sanity check of PR. I am going to push it in once there is a build that passes tests.

message_definitions/v1.0/common.xml
```diff
@@ -2786,6 +2786,9 @@
       <entry value="65536" name="MAV_PROTOCOL_CAPABILITY_FLIGHT_INFORMATION">
         <description>Autopilot supports the flight information protocol.</description>
       </entry>
+      <entry value="131072‬" name="MAV_PROTOCOL_CAPABILITY_MISSION_CHECKSUM">
```
peterbarker, line 2789: Not really sure of the utility of this bit.

Why not just request a `MISSION_CHECKSUM` message and see what answer you get back?

peterbarker (commented):
Onboard cripted changes to missions (like @davidbuzz 's ) might lead to a *lot* of checksumming and network traffic.

message_definitions/v1.0/common.xml
hamishwillee: Fair enough! The original design was different and this made more sense. I will remove.

hamishwillee:
> Onboard scripted changes to missions (like @davidbuzz 's ) might lead to a lot of checksumming and network traffic.

@peterbarker Yes. What do you suggest?

holmnikolaj:
@hamishwillee It seems to pass now, and the script didn't make changes - perhaps it was the protocol bit part?

davidbuzz:
I'd just drop a note here that i have operationally used AUTO and a dynamic mission (such as lua) that is changed repeatedly (a companion computer can also be used to repeatedly push a  trivial  'next target'  mission, i've done that pre-lua), as often as 1hz.    The main value for us was being in AUTO was the pre-canned fully auto takeoff, fully auto landing, and the fact that when an aircraft gets to the end of its auto/mission it "rtls" without operator interaction, which for us brought it back, totally hands-off, still in auto, if for any reason the companion-computer (or lua) failed to keep the dynamic mission up-to-date.   so although i suppose a lua script that changes it to guided-does-stuff-then-changes-back is probably very similar, for us, it was the 'fall off the end of the mission = rtl'  feature that was an important failsafe for when either the CC or lua updating the mission stops for any reason.

dagar:
Wouldn't this be helpful as extension fields to existing mission messages (MISSION_CURRENT, MISSION_COUNT, etc)?

http://mavlink.io/en/messages/common.html#MISSION_CURRENT
http://mavlink.io/en/messages/common.html#MISSION_COUNT

hamishwillee:
> Wouldn't this be helpful as extension fields to existing mission messages (MISSION_CURRENT, MISSION_COUNT, etc)?
> [MISSION_CURRENT](http://mavlink.io/en/messages/common.html#MISSION_CURRENT), [MISSION_COUNT](http://mavlink.io/en/messages/common.html#MISSION_COUNT)

@dagar  I was thinking is that it might be useful in `MISSION_ACK`. This would ensure that on completion of an upload the GCS gets the precise value of the hash that the drone has for this upload and doesn't have to calculate the hash if it does not want. The drone would still need to emit the hash separately though because no other systems will be looking at this. Thoughts?

For other messages
- If it is included in additional messages this is an increase in network consumption. What's the additional benefit and is it worth that cost (for each case)? 
- MISSION_CURRENT is emitted every time the current mission number changes. I don't particularly see the benefit of this over GCS requesting when needed or getting a notification when it changes. I guess just that it gets emitted regularly but you'd still need the other emit cases.
- MISSION_COUNT is sent to kick off upload of mission to drone, and to tell GCS how many items will be sent in download from drone. Current thinking is that the drone hash is "canonical" so in the download from drone case it is fine, but during upload what should drone do with this info?

hamishwillee:
Thanks @davidbuzz - that's what we needed to know. I completely "get" why you would use this approach;  most of this can be done in other ways, but the end of script failsafe is pretty convenient.

@peterbarker So on this basis you're looking at an extra message being emitted at 1Hz. Perfectly acceptable in and of itself - but it might not end there; in theory every GCS on the network could then decide it needs to update the onboard mission. I guess we could provide guidance that a GCS should mark current mission as invalid, but only update when explicitly requested or after a timeout without changed hash messages?

Otherwise I'm open to proposals/suggestions for improvement but do not consider this "blocking".

peterbarker:
On Mon, 28 Dec 2020, Hamish Willee wrote:

> @peterbarker So on this basis you're looking at an extra message being emitted at 1Hz. Perfectly acceptable in and of itself - but it might not end there;
> in theory every GCS on the network could then decide it needs to update the onboard mission. I guess we could provide guidance that a GCS should mark
> current mission as invalid, but only update when explicitly requested or after a timeout without changed hash messages?

Probably removing the requirement that all updates to the mission be 
notified and leave it to the autopilot to determine when the hash should 
be sent would be a good idea.

hamishwillee:
> Probably removing the requirement that all updates to the mission be notified and leave it to the autopilot to determine when the hash should be sent would be a good idea.

I'm pretty sure it isn't a good idea; for this to work all systems either have to be able to rely on it, or know that they can't rely on it. 

But it would be fine to add back the protocol bit, and have an autopilot have a parameter to enable/disable support. That way anyone can check whether it is currently supported by an autopilot.

In theory we might also have a command set the expected behaviour in the spec (i.e. a command that says notify all the time or notify when requested). But I think that is perhaps an unnecessary complication.

IamPete1:
I'm still planning on doing a AP implementation, have not had a chance to get to it yet tho. (@auturgy )

hamishwillee:
This was discussed in the [Devcall](https://github.com/mavlink/mavlink/wiki/20210120-Dev-Meeting). Current policy is that we don't merge into common without a working implementation and at least intent to have in another implementation.

That said, the current design is considered acceptable, so you can take it and move forward (i.e. we will iterate based on experience with the prototype). OK by you @IamPete1 and @holmnikolaj ?

> **Note** For example, if the requirement to send out hash on every update does swamp links, then we might make the requirement for auto emission "optional" (I'm kind of OK with that because even though it is desirable, the fact is any system could miss this update so no system can absolutely rely on it).

holmnikolaj:
> 
> 
> This was discussed in the [Devcall](https://github.com/mavlink/mavlink/wiki/20210120-Dev-Meeting). Current policy is that we don't merge into common without a working implementation and at least intent to have in another implementation.
> 
> That said, the current design is considered acceptable, so you can take it and move forward (i.e. we will iterate based on experience with the prototype). OK by you @IamPete1 and @holmnikolaj ?

Sure!

julianoes:
@hamishwillee do you remember why this stalled? @dagar brought it up in todays QGC call.

When I look at the PR I would say that we merge it and try implementing it? Anyone against that?

peterbarker:
> @hamishwillee do you remember why this stalled? @dagar brought it up in todays QGC call.
> 
> When I look at the PR I would say that we merge it and try implementing it? Anyone against that?

`common.xml` shouldn't be a sandbox.

As this spec is written, the checksum ArduPilot returns won't be the same over time, even with no GCS interaction - at least while we're disarmed.  ArduPilot considers its home location to be part of the mission, and returns it as mission item 0.  We drift the home location until the vehicle is armed (exact behaviour varies by supported AP vehicle type).

item-0 could be excluded from the checksum by AP - but then a GCS will need to know to exclude it from its own checksum (and my nefarious plan to use a capability bit to indicate mission-slot-0 handling is still only a plan).

julianoes:
> `common.xml` shouldn't be a sandbox.

Right, I get that but I doubt we have done the plumbing work (aka `px4.xml`, or `development.xml`) to enable that sort of sandbox, right? In that case we have a chicken and egg problem where we're told we have to implement it first but at the same time, there is no way to do so. (One could of course generate the headers locally to start implementing it but it would not work once you try to push it and make a PR to actually proof that there is an implementation.)

Note that I'm out of the loop of the governance discussion, so I might be missing something.

> item-0 could be excluded from the checksum by AP - but then a GCS will need to know to exclude it from its own checksum (and my nefarious plan to use a capability bit to indicate mission-slot-0 handling is still only a plan).

Ok, so that's the other thing that needs to be resolved first...

hamishwillee:
@peterbarker @julianoes 

This is waiting on an implementation before merging. Yes this is a pain for developers, and I hope we can move to having a development.xml soon that can be used instead to ease things going forward. @IamPete1 can you give us an update on your progress?

The issue with home is already addressed. The algorithm for calculating the has excludes the home location The GCS knows that, because they looked at the documentation in the message in this PR :-)

julianoes:
> This is waiting on an implementation before merging.

I don't think we should change and enforce the process without having the tools in place.

hamishwillee:
>> This is waiting on an implementation before merging.

> I don't think we should change and enforce the process without having the tools in place.

I have no strong opinion, other than the last time something like this came up @LorenzMeier suggested that it isn't all that hard to branch mavlink for your changes while developing, which I took to mean we should modify the process. Certainly very strong push not to merge anything without a prototype.

julianoes:
@hamishwillee that's what I wrote further above:

> (One could of course generate the headers locally to start implementing it but it would not work once you try to push it and make a PR to actually proof that there is an implementation.)

hamishwillee:
@julianoes Then let's discuss that in the dev call - [added to agenda](https://github.com/mavlink/mavlink/wiki/20210317-Dev-Meeting)

hamishwillee:
Superceded by #1611

@IamPete1 @holmnikolaj Once that goes in, implementators should be able to use the development.xml rather than common.xml (or ideally standard.xml) while prototyping.

Issue #1175, opened 2019-07-09
Update docs related to AUTOPILOT_VERSION


olliw42:
so that #1156 isn't lost (I couldn't reopen it)

julianoes:
I have no opinion on that one, I'd need to read through it all.

hamishwillee:
Thanks for "reopening" @olliw42 

@julianoes The important bit is in https://github.com/mavlink/mavlink/issues/1156#issuecomment-506602488

When I started working on MAVLink the first thing I naively documented all the messages/ways of getting version info: https://mavlink.io/en/guide/mavlink_version.html#determining-protocolmessage-version

The problem is that not all the methods are implemented everywhere so you can't rely on them. 
- You can rely on HEARTRBEAT for the version number of the library (as it is auto-generated) and the  start marker byte for the protocol version of the current message (but not whether other versions are supported by the library). 
- You must be able to rely on PROTOCOL_VERSION.
- It is unclear from the docs whether you can rely on AUTOPILOT_VERSION

The action here are to:
1. Move PROTOCOL_VERSION to minimal.xml, and ideally as a follow on make common.xml split  off minimal.xml. 
2. Add note to AUTOPILOT_VERSION and to https://mavlink.io/en/guide/mavlink_version.html to make it clear that this cannot be relied on.

julianoes:
> ArduPilot has an AUTOPILOT_VERSION_REQUEST message in it’s dialect, which is used by MissionPlanner to query for AUTOPILOT_VERSION, and MissionPlanner kind of requires a component to respond to this request in order to work properly. This isn't so for QGroundControl.

I can't find AUTOPILOT_VERSION_REQUEST. I think you meant  [MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES) which is actually what we use in QGC and MAVSDK as well to connect to an autopilot.

MAVSDK currently does not implement [PROTOCOL_VERSION](https://mavlink.io/en/messages/common.html#PROTOCOL_VERSION) but probably should.
(The only problem there is that it's marked WIP and that you should not use it in production. If the message is actually required to correctly connect then a change to this message would break everything. Given it's already in QGC the WIP tag probably needs to be removed and we have to stabilize it.)

[...]

PR #1174, opened 2019-07-08
New gimbal messages

julianoes:
This is the follow up to the [discussion on the dev call](https://github.com/mavlink/mavlink/wiki/20190515-Dev-Meeting#call-1) where the overall opinion was to deprecate the existing mount/gimbal messages and come up with a new/consistent set.

I also had further helpful discussions with @olliw42, for details see #1130 and #1135.

**This PR/proposal has evolved quite a bit over time. There is a [document for the high level description](https://docs.google.com/document/d/16pekKRXLN2FhlL9YNFP983cjfBKAsDwN0gOSks8USo4/edit?usp=sharing).**

Issue #1156, opened 2019-06-04
Pl clarify if a component is required to support AUTOPILOT_VERSION


olliw42:
this is in reference to https://discuss.px4.io/t/mavlink-support-for-the-storm32-gimbal-controller-how-to-make-it-compatible/10723/8, last point.

ArduPilot has an AUTOPILOT_VERSION_REQUEST message in it’s dialect, which is used by MissionPlanner to query for AUTOPILOT_VERSION, and MissionPlanner kind of requires a component to respond to this request in order to work properly. This isn't so for QGroundControl.

This thus raises the question if a component should be required to support AUTOPILOT_VERSION for proper functioning, or not, and if so this would need to be clarified by the spec.

Comment: For PROTOCOL_VERSION the specification unambiguously states that every component should support it. However, AUTOPILOT_VERSION and PROTOCOL_VERSION kind of serve a similar purpose in that they both can be used to determine the MAVLink version. Requiring support of both thus appears to be somewhat superfluous. However, not requiring AUTOPILOT_VERSION would kind of break compatibility with current MissionPlanner.

amilcarlucas:
I guess this boils down to: which messages should be moved to minimal.xml.

By moving them would make it 100% clear what are minimal set of messages that a system needs to understand in order to be able to say that it is mavlink compatible !!

hamishwillee:
My take is that implementing AUTOPILOT_VERSION is not mandatory. Everything needs to be able to confirm it's version, but all the other information in AUTOPILOT_VERSION is only necessary for an autopilot node. I'm sure we could find arguments either way, but let's try keep minimal.xml as minimal as possible :-)

@amilcarlucas Good point re minimal.xml - we should move PROTOCOL_VERSION in there too. What we need to do before that is make minimal.xml an include file for common.xml. I had intended to do this some time ago, but was waiting for deeper nesting of includes to be supported. I worry this will break things. Will put back on my list

hamishwillee:
PS. Closing. I believe this is answered.

olliw42:
many thx for the clarification

could this in some way be unambiguously indicated in the docs? 

The point is that MissionPlanner/ArduPilot use this message and MissonPlanner in fact relies on every component to support it. I doubt that MissionPlanner/ArduPilot will replace it by PROTOCOL_VERSION, but it should be at least made clear (to them ;)) that this is not according to spec, and MissionPlanner might be motivated by this to actually adapt it's behavior.

:)

[...]

Issue #1153, opened 2019-06-03
Pl clarify MAV_PROTOCOL_CAPABILITY_PARAM_FLOAT and MAV_PROTOCOL_CAPABILITY_PARAM_UNION


olliw42:
I've pointed to this in the px4 discuss forum (https://discuss.px4.io/t/mavlink-support-for-the-storm32-gimbal-controller-how-to-make-it-compatible/10723) but it seems it hasn't much frequency, so pl allow me to put this up here:

According to the docu the MAV_PROTOCOL_CAPABILITY bit mask has the flags

>  MAV_PROTOCOL_CAPABILITY_PARAM_FLOAT: Autopilot supports the new param float message type.
  
>  MAV_PROTOCOL_CAPABILITY_PARAM_UNION: Autopilot supports the new param union message type.

Since these appear to be related to the parameter handling, which is certainly one of the most fundamental MAVLink features, I suspect that these are important, but I failed to find any information. either direct or indirect, on what they possible could mean or indicate. 

So, could you please add few statements somewhere to clarify their meaning/usage, e.g. in the parameter protocol section? (if they're WIP pl indicate as WIP)

Many thx, Olli

olliw42:
@hamishwillee 
since you thankfully responded in the said post, which I take as indication to continue the discussion there, I'm closing this issue

many thx

PR #1176, opened 2019-07-10
COMPONENT_INFORMATION

olliw42:
- MAV_CMD_REQUEST_COMPONENT_INFORMATION added, mirrors MAV_CMD_REQUEST_CAMERA_INFORMATION
- COMPONENT_INFORMATION message added, mirrors CAMERA_INFORMATION to a large extend, but has also a hardware version field. COMPONENT_CAP_FLAGS are currently not defined.
- both were given the tentative ids 700

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3227,8 +3227,8 @@
     </enum>
     <enum name="COMPONENT_CAP_FLAGS">
       <description>Component capability flags (Bitmap)</description>
-      <entry value="0" name="DUMMY">
-        <description>Dummy to make generator happy.</description>
+      <entry value="0" name="COMPONENT_CAP_FLAGS_NONE">
+        <description>No flags set.</description>
       </entry>
     </enum>
   </enums>
@@ -5219,10 +5219,10 @@
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
-    <message id="700" name="COMPONENT_INFORMATION">
+    <message id="385" name="COMPONENT_INFORMATION">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
-      <description>Information about a component. For camera components use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
+      <description>Information about a component. For camera components instead use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the component model</field>
@@ -5230,7 +5230,7 @@
       <field type="uint32_t" name="hardware_version">Version of the component hardware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
       <field type="uint32_t" name="capability_flags" enum="COMPONENT_CAP_FLAGS">Bitmap of component capability flags.</field>
       <field type="uint16_t" name="component_definition_version">Component definition version (iteration)</field>
-      <field type="char[140]" name="component_definition_uri">Component definition URI (if any, otherwise only basic functions will be available).</field>
+      <field type="char[140]" name="component_definition_uri">Component definition URI (if any, otherwise only basic functions will be available). The XML format is not yet specified and work in progress. </field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1645,13 +1645,6 @@
         <param index="1" label="Tag" minValue="0" increment="1">Target tag to jump to.</param>
         <param index="2" label="Repeat" minValue="0" increment="1">Repeat count.</param>
       </entry>
-      <entry value="700" name="MAV_CMD_REQUEST_COMPONENT_INFORMATION" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Request component information (COMPONENT_INFORMATION). For camera components use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
-        <param index="1" label="Capabilities" minValue="0" maxValue="1" increment="1">0: No action 1: Request component information</param>
-        <param index="2">Reserved (all remaining params)</param>
-      </entry>      
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
@@ -3231,6 +3224,9 @@
       <entry value="2" name="PARACHUTE_RELEASE">
         <description>Release parachute.</description>
       </entry>
+    </enum>
+    <enum name="COMPONENT_CAP_FLAGS">
+      <description>Component capability flags (Bitmap)</description>
     </enum>
   </enums>
   <messages>
@@ -5232,7 +5228,7 @@
       <field type="uint32_t" name="capability_flags" enum="COMPONENT_CAP_FLAGS">Bitmap of component capability flags.</field>
       <field type="uint16_t" name="component_definition_version">Component definition version (iteration)</field>
       <field type="char[140]" name="component_definition_uri">Component definition URI (if any, otherwise only basic functions will be available).</field>
-    </message>    
+    </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
       <description>Cumulative distance traveled for each reported wheel.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3735,9 +3735,6 @@
       <entry value="2" name="COMPONENT_CAP_FLAGS_PARAM_EXT">
         <description>Component has parameters, and supports the extended parameter protocol (PARAM_EXT messages).</description>
       </entry>
-      <entry value="4" name="COMPONENT_CAP_FLAGS_FTP">
-        <description>Component supports MAVLinkFTP for downloading the component definition file.</description>
-      </entry>
     </enum>
   </enums>
   <messages>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3227,6 +3227,9 @@
     </enum>
     <enum name="COMPONENT_CAP_FLAGS">
       <description>Component capability flags (Bitmap)</description>
+      <entry value="0" name="DUMMY">
+        <description>Dummy to make generator happy.</description>
+      </entry>
     </enum>
   </enums>
   <messages>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3735,6 +3735,9 @@
       <entry value="2" name="COMPONENT_CAP_FLAGS_PARAM_EXT">
         <description>Component has parameters, and supports the extended parameter protocol (PARAM_EXT messages).</description>
       </entry>
+      <entry value="4" name="COMPONENT_CAP_FLAGS_FTP">
+        <description>Component supports MAVLinkFTP for downloading the component definition file.</description>
+      </entry>
     </enum>
   </enums>
   <messages>
@@ -5766,7 +5769,7 @@
     <message id="395" name="COMPONENT_INFORMATION">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
-      <description>Information about a component. For camera components instead use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION. Components including GCSes must support requests of this message via MAV_CMD_REQUEST_MESSAGE.</description>
+      <description>Information about a component. For camera components instead use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION. Components including GCSes should consider supporting requests of this message via MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the component model</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3229,7 +3229,7 @@
     <enum name="COMPONENT_CAP_FLAGS">
       <description>Component capability flags (Bitmap)</description>
       <entry value="0" name="COMPONENT_CAP_FLAGS_NONE">
-        <description>No flags set.</description>
+        <description>No flags set. (do not use this flag, it will eventually be removed)</description>
       </entry>
     </enum>
   </enums>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3703,8 +3703,11 @@
     </enum>
     <enum name="COMPONENT_CAP_FLAGS">
       <description>Component capability flags (Bitmap)</description>
-      <entry value="0" name="COMPONENT_CAP_FLAGS_NONE">
-        <description>No flags set. (do not use this flag, it will eventually be removed)</description>
+      <entry value="1" name="COMPONENT_CAP_FLAGS_PARAM">
+        <description>Component has parameters, and supports the parameter protocol (PARAM messages).</description>
+      </entry>
+      <entry value="2" name="COMPONENT_CAP_FLAGS_PARAM_EXT">
+        <description>Component has parameters, and supports the extended parameter protocol (PARAM_EXT messages).</description>
       </entry>
     </enum>
   </enums>
@@ -5711,7 +5714,7 @@
     <message id="390" name="COMPONENT_INFORMATION">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
-      <description>Information about a component. For camera components instead use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
+      <description>Information about a component. For camera components instead use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION. Components including GCSes must support requests of this message via MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
       <field type="uint8_t[32]" name="model_name">Name of the component model</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1645,6 +1645,13 @@
         <param index="1" label="Tag" minValue="0" increment="1">Target tag to jump to.</param>
         <param index="2" label="Repeat" minValue="0" increment="1">Repeat count.</param>
       </entry>
+      <entry value="700" name="MAV_CMD_REQUEST_COMPONENT_INFORMATION" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Request component information (COMPONENT_INFORMATION). For camera components use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
+        <param index="1" label="Capabilities" minValue="0" maxValue="1" increment="1">0: No action 1: Request component information</param>
+        <param index="2">Reserved (all remaining params)</param>
+      </entry>      
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
@@ -5213,6 +5220,19 @@
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
+    <message id="700" name="COMPONENT_INFORMATION">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Information about a component. For camera components use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
+      <field type="uint8_t[32]" name="model_name">Name of the component model</field>
+      <field type="uint32_t" name="firmware_version">Version of the component firmware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
+      <field type="uint32_t" name="hardware_version">Version of the component hardware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
+      <field type="uint32_t" name="capability_flags" enum="COMPONENT_CAP_FLAGS">Bitmap of component capability flags.</field>
+      <field type="uint16_t" name="component_definition_version">Component definition version (iteration)</field>
+      <field type="char[140]" name="component_definition_uri">Component definition URI (if any, otherwise only basic functions will be available).</field>
+    </message>    
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
       <description>Cumulative distance traveled for each reported wheel.</description>


auturgy:
@peterbarker this might be useful for APSync?

message_definitions/v1.0/common.xml
```diff
@@ -1645,6 +1645,13 @@
         <param index="1" label="Tag" minValue="0" increment="1">Target tag to jump to.</param>
         <param index="2" label="Repeat" minValue="0" increment="1">Repeat count.</param>
       </entry>
+      <entry value="700" name="MAV_CMD_REQUEST_COMPONENT_INFORMATION" hasLocation="false" isDestination="false">
```
hamishwillee: Use https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE for this

message_definitions/v1.0/common.xml
```diff
@@ -5213,6 +5220,19 @@
...
+    <message id="700" name="COMPONENT_INFORMATION">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Information about a component. For camera components use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
+      <field type="uint8_t[32]" name="model_name">Name of the component model</field>
+      <field type="uint32_t" name="firmware_version">Version of the component firmware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
+      <field type="uint32_t" name="hardware_version">Version of the component hardware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
+      <field type="uint32_t" name="capability_flags" enum="COMPONENT_CAP_FLAGS">Bitmap of component capability flags.</field>
```
julianoes: I wonder if 64 bits for the flag is safer. What do you think?

julianoes: Also: this enum would have to be defined above with the other enums.

julianoes:
@olliw42 travis complains:
```
./message_definitions/v1.0/common.xml needs formatting - run ./scripts/format_xml.sh
```

message_definitions/v1.0/common.xml
olliw42: well, as said, I took CAMERA_INFORMATION as a guide/template ...

... I would think that if to use MAV_CMD_REQUEST_MESSAGE or not is a deeper more general question, which was raised somewhere else but not yet finally settled I believe. I'm always for being systematic, and I would find it most strange if that would be done now here but not at the same time all the other REQUEST commands be marked as deprecated ...
... but at the end of the day I really don't care :D, so pl decide among you guys :)

message_definitions/v1.0/common.xml
olliw42: well, as said, I took CAMERA_INFORMATION as a guide/template ... (LOL)

I have no opinion as regards u32 or u64 ... given that we don't have a single bit in use so far (and I even don't have one in mind so far), u32 sounds like a lot
maybe, if it ever should exhaust, it would be acceptable to extend the message by a 2nd flag

olliw42:
great that travis is so kind to complain, but sadly it complains to someone who doesn't understand what it wants ... something with the formatting ... huh ??? in the file view (https://github.com/mavlink/mavlink/blob/0a1c9fd9daa20669a9938cf5bf931bca135c10c4/message_definitions/v1.0/common.xml) I can't notice anything, and I would think that the git diff view also should show what the problem is (and I didn't notice anything in Notepad++ when I did the change in the first place ...)

I'm sorry to be clueless here

olliw42:
700 ??

message_definitions/v1.0/common.xml
julianoes: Yes, the decision on today's call was to use MAV_CMD_REQUEST_MESSAGE in the future as much as possible. Given that, you're right in that we should start to mark the old way as deprecated.

julianoes:
> 700 ??

How would I know. Yea?

message_definitions/v1.0/common.xml
olliw42: is it possible to define an enum without any value? 

i.e like so
```
<enum name="COMPONENT_CAP_FLAG">
<description>blabla</description>
</enum>
```

olliw42:
>   700 ??
> How would I know. Yea?

I chose 700 based on no reasony ... I have no idea how you guys do decide ids, so please suggest

message_definitions/v1.0/common.xml
julianoes: I'd hope so, otherwise create one dummy entry.

julianoes:
@olliw42 travis tells you to run:
```
./scripts/format_xml.sh
```

olliw42:
> @olliw42 travis tells you to run:
./scripts/format_xml.sh

that's now light years above my head ... never ever ran a .sh and frankly I'm not eager at all to change that now ... I'm sorry
if it knows this needs to be run why isn't it doing it

I'll do the changes suggested in the above, and then will try to make it as clean as possible with the editors I have at hand, and then let's see ...

message_definitions/v1.0/common.xml
olliw42: ok, so I'll just remove

good to have that decission

olliw42:
removed MAV_CMD_REQUEST_COMPONENT_INFORMATION
and added COMPONENT_CAP_FLAGS enum, with no entry - the py generator at least didn't complained about this
also converted a tab into spaces

olliw42:
while thinking about the capability flags for components I stumbled across an issue I'd like to discuss:

The component capability flags, are they supposed to be generic and valid for all components, or could they be depending on the MAV_TYPE?

example: 
It could be useful to give a gimbal the capability "supports ROI message" (as you also asked in #1174), and the question would be if it would make sense to use the capability flag in COMPONENT_INFORMATION for that purpose.
Which would raise the next question. This would bring the issue that u32 or even u64 can be very quickly exhausted, since who knows how many components want to add how many flag bits in future. One obvious resolution would be to make the meaning of the bits dependent on the type of the component, so that each type would have the full u32 at its disposal. That is for a MAV_TYPE_GIMBAL there could be the definition HAS_GPS_TRACKING  0x0001, but for a MAV_TYPE_CHARGING_STATION there could be the definition HAS_WHATEVERITWANTSTOHAVE  0x0001, both at the same bit location.
If such a thing were acceptable it would simplify things, and avoid exhaustion, but I have no idea if you guys find that acceptable.

If acceptable, one also could split the bit range into an area which is generic for all components and an area which is specific for the type. Either by bit ranging or by e.g. actually having two capability flag fields in the message.

I would find it appealing if there would be one mechanism for getting the component specific capabilities, and not to need x mechanisms for x component types.

What do you guys say?

message_definitions/v1.0/common.xml
olliw42: the tactics is quite confusing though ... just realized that a MAV_CMD_REQUEST_PROTOCOL_VERSION was added just few hours ago, and to minimal.xml actually ... :confused: 
anyway, removed it

hamishwillee: MAV_CMD_REQUEST_PROTOCOL_VERSION was **already defined** in common.xml. All this change did was duplicate it in minimal.xml  (as we discussed this should contain the minimal set of things you need to set up a connection) and validate a link. 

You may be wondering why the duplication - why not just include minimal.xml in common then? Well, that's the next step. But I realised our generator still doesn't handle this properly following testing, so that is on hold.

At some point though there should be just one MAV_CMD_REQUEST_PROTOCOL_VERSION and that should be deprecated in favour of MAV_CMD_REQUEST_MESSAGE

hamishwillee: or maybe MAV_CMD_REQUEST_MESSAGE will have to go in minimal. Bit of a can of worms.

olliw42: many thx
that's what I would have kind of "expected" in light of the decision, that at least also MAV_CMD_REQUEST_MESSAGE shows up in minimal, and maybe even that MAV_CMD_REQUEST_PROTOCOL_VERSION be marked as deprecated

as regards the generator, you're saying the issue is that includes such as <include>common.xml</include> in the .xml files are not yet properly handled when they are recursive? If that's what it is I might be motivated to sort this out

hamishwillee: I can't deprecate a message when no one has implemented it. So this will happen over time.

> as regards the generator, you're saying the issue is that includes such as common.xml in the .xml files are not yet properly handled when they are recursive? If that's what it is I might be motivated to sort this out

Yes. More specifically, we support multiple levels of inclusion but I screwed up the MAVLINK_MESSAGE_CRCS calculation so it only supports inclusion of immediate descendents, not all of them: https://github.com/ArduPilot/pymavlink/blob/master/generator/mavgen.py#L91

The very last discussion on slack is about this. 

If you have a cunning way to sort out the generator then great. I'm planning on looking at it in the coming weeks.

olliw42: but if no one has implemented it, and it's going to be deprecated, it never would have to be added in the first place, and/or could be removed immediately there it was, right?

generator: yeah, I just was looking at this function too ... I guess the approach I would have attempted is to simply build up a complete .xml file (in storage) before parsing. The code seems to first parse each file and when tries to get the includes right. I would reverse that. C preprocessor like. So would be a change in processing logic and from what you write I'm not sure it would be appreciated, and if you plan to do it anyhow ... I maybe shouldn't get my fingers dirty LOL.

olliw42: @hamishwillee 
as regards the generator try this, I think it's doing it: www.olliw.eu/drop/4hamiswillee-minimal.zip (copy the files in the respective places)
:)

olliw42:
yippieee ... travis seems to be happy now LOL ... and this without any .sh or understanding what I did, stupidity won !!!

message_definitions/v1.0/common.xml
hamishwillee: As you can see, I missed this.. I guess two implementations are better than one??? :-) Leave it to Peter to work out how he wants to assess the options. 

> but if no one has implemented it, and it's going to be deprecated, it never would have to be added in the first place, and/or could be removed immediately there it was, right? Also, since it was already in common.xml as you emphasized, and even more it is even in the docs, how can you then know that no one has implemented it? Puzzling.

The message MAV_CMD_REQUEST_PROTOCOL_VERSION has been implemented and is in production. The message  MAV_CMD_REQUEST_MESSAGE was added later, but no flight stack has yet wrtten code that will a)accept this for any message, b) accept it to get a response MAV_CMD_REQUEST_PROTOCOL_VERSION

So if you deprecate MAV_CMD_REQUEST_PROTOCOL_VERSION and point to MAV_CMD_REQUEST_MESSAGE that doesn't work anywhere, then we are not helping people.

peterbarker:
On Thu, 11 Jul 2019, Hamish Willee wrote:

> The message MAV_CMD_REQUEST_PROTOCOL_VERSION has been implemented and is in production. The message MAV_CMD_REQUEST_MESSAGE was added later, but no flight
> stack has yet wrtten code that will a)accept this for any message, b) accept it to get a response MAV_CMD_REQUEST_PROTOCOL_VERSION

ArduPilot does accept this for any message it supports streaming of.  So 
it won't support it for STATUSTEXT, for example.

ArduPilot does not support PROTOCOL_VERSION and last time it came up there 
was some resistance to the concept of supporting it.

> So if you deprecate MAV_CMD_REQUEST_PROTOCOL_VERSION and point to MAV_CMD_REQUEST_MESSAGE that doesn't work anywhere, then we are not helping people.

deprecating is not the same as removing - but it probably be a bit rude to 
put such a new message into a "minimal" message set.  Additionally, that 
specific message has been used in documentation as to how mavlink should 
be used - so it should probably not go anywhere.

olliw42:
> ArduPilot does accept this for any message it supports streaming of. So it won't support it for STATUSTEXT, for example.

yeah, pointed that out also already here: https://github.com/mavlink/mavlink/pull/1180. I think the message came across :)

> ArduPilot does not support PROTOCOL_VERSION and last time it came up there was some resistance to the concept of supporting it.

Thx for mentioning this. This a different can of worms however, and I guess it would be better placed here: https://github.com/mavlink/mavlink/issues/1175. Maybe you could chime in there, since that issue was cerated exactly to finally sort out the PROTOCOL_VERSION, AUTOPILOT_VERSION, Mavlink1/2 versioning mayhay.

>So if you deprecate MAV_CMD_REQUEST_PROTOCOL_VERSION and point to MAV_CMD_REQUEST_MESSAGE that doesn't work anywhere, then we are not helping people.

> Additionally, that specific message has been used in documentation as to how mavlink should be used - so it should probably not go anywhere.

this is so however for many of the REQUEST commands, e.g. also the many related to the camera microservice.  https://github.com/mavlink/mavlink/pull/1180 is maybe a better place to raise that concern.

sorry if you find me a bit directive here, but my feeling is the discussion of the many lose ends has become a bit entangled, and disentangling into different issues/prs would help stay clear (at least me).

message_definitions/v1.0/common.xml
olliw42: @hamishwillee 
as regards the generator, I think your version is NOT correct. Your main subfolder may come out correctly, but I think you miss to see that your include subfolders are going to be incorrect! E.g., for ardupilot.xml, the files in the ardupilot subfolder may be correct, but I don't think the files in the other subfolders will be identical to then you would start your generator with them as root!!! But this is how it has to be, anything else is a bug. Please carefully compare the result of my version to the result of your version as regards ALL generated files, compared to when the an included file is used in the generator.
cheers

hamishwillee: I'll address in https://github.com/ArduPilot/pymavlink/pull/339

olliw42:
> I think this is good to go, right?

maybe yes, maybe no ... 

IMHO there is a more general point in here which one might want to address before. I have mentioned it in few other places, e.g. also in the new gimbal message PR thread, but not here. It was triggered by the recent suggestion of a COMPONENT_STATUS, but the thoughts there also applies to INFORMATION type messages.

The basic question is if one wants a COMPONENT_INFORMATION for all components, or XXX_INFORMATION per MAV_TYPE  (gimbal, servo, gps, ...), or both (in which case one would have to decide which information goes where). It arises since some information is indeed generic for each component while other information might depend on the MAV_TYPE. 

The component definition file stuff e.g. would be certainly generic for all components. However, the capabilities flag may not. In the above I thus was asking if it would be conceivable to map a range of the bitmask to flags per MAV_TYPE. The alternative would be to have the capabilities flag in a XXX_INFORMATION message.

For instance, in the new gimbal message thread you suggested a GIMBAL_INFORMATION. So we would have then 

CAMERA_INFORMATION
COMPONENT_INFORMATION
GIMBAL_INFORMATION

with a very unfortunate (IMHO) asymmetry in the distribution of fields and flags. CAMERA_INFORMATION carries the definition file and the capabilities, while these info is in COMPOENT_INFORMATION for all non-cameras with potentially MAV_TYPE dependent capability flags, while GIMBAL_INFORMATION has none of that ... frankly, what a MESS

so, if one would not stop now for a second to think twice one can already see that we would do exactly what I kind of criticize about how things were sometimes done in the past ... and layed the basis for some mess ... 

the above looming "mess" could (!) be resolved e.g. by having only MAV_TYPE INFORMATION messages, which each carry the general fields but also the specific ones. That is, there would NOT be a COMPONENT_INFORMATION, but there would be a GIMBAL_INFOMATION with the same fields as of the currently suggested COMPONENT_INFORMATION plus whatever additional fields one likes for gimbal types. For a servo there would be then a SERVO_INFORMATION with again  the fields of the current COMPONENT_INFORMATION plus whatever additional fields one likes for servo types. And so on. This would be kind of the most symmetric in view of the existing CAMERA_INFORMATION.

an other approach could (!) be to accept CAMERA_INFORMATION as outlier, and to decide to have both a COMPONENT_INFORMATION and MAV_TYPE_INFORMATION messages. In which case one would have to decide which info goes where.

and as third approach one could (!) go ahead with the current COMPONENT_INFORMATION and make the capabilities flag MAV_TYPE dependent. One still could then have MAV_TYPE_INFORMATION messages but these must not have then capabilities flags.

and there are probably more possible approaches one could go ...

I don't have clear favorite, I just clearly think it would be worthwhile to think that through before making decisions ...

message_definitions/v1.0/common.xml
```diff
@@ -5213,6 +5219,19 @@
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
+    <message id="700" name="COMPONENT_INFORMATION">
```
hamishwillee: Suggest use id=385. Basically the [advice](https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges) is to use next available, but it is a good idea to leave a couple free to allow additional messages or iterations of the message if needed. 

I am assuming that this isn't needed by MAVLink 1 -only systems.

message_definitions/v1.0/common.xml
```diff
@@ -5213,6 +5219,19 @@
...
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Information about a component. For camera components use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION.</description>
```
hamishwillee: Might be worth saying "For camera components *instead* ...

message_definitions/v1.0/common.xml
```diff
@@ -5213,6 +5219,19 @@
...
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
```
hamishwillee: We are somewhat tied to `time_usec` as a standard timestamp for "timestamp since Unix Epoch", but we're not tied to that format otherwise. So perhaps we should do this right and remove the units from this one.

So either `time_boot` or `uptime` (argue among yourselves)

message_definitions/v1.0/common.xml
```diff
@@ -5213,6 +5219,19 @@
...
+      <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
+      <field type="uint8_t[32]" name="model_name">Name of the component model</field>
+      <field type="uint32_t" name="firmware_version">Version of the component firmware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
+      <field type="uint32_t" name="hardware_version">Version of the component hardware (v &lt;&lt; 24 &amp; 0xff = Dev, v &lt;&lt; 16 &amp; 0xff = Patch, v &lt;&lt; 8 &amp; 0xff = Minor, v &amp; 0xff = Major)</field>
+      <field type="uint32_t" name="capability_flags" enum="COMPONENT_CAP_FLAGS">Bitmap of component capability flags.</field>
+      <field type="uint16_t" name="component_definition_version">Component definition version (iteration)</field>
```
hamishwillee: What is this for?

message_definitions/v1.0/common.xml
```diff
@@ -5213,6 +5219,19 @@
...
+      <field type="char[140]" name="component_definition_uri">Component definition URI (if any, otherwise only basic functions will be available).</field>
```
hamishwillee: Do we have a format fully or partially defined for this? If so then link to it, if not, then state what is allowed (e.g. "XML format is component specific and must be agreed separately with the MAVLink devteam ... or similar)

message_definitions/v1.0/common.xml
```diff
@@ -3225,6 +3225,12 @@
         <description>Release parachute.</description>
       </entry>
     </enum>
+    <enum name="COMPONENT_CAP_FLAGS">
+      <description>Component capability flags (Bitmap)</description>
+      <entry value="0" name="DUMMY">
```
hamishwillee: While this is strictly accurate, names should always follow the convention enum_name_x. 
So perhaps make `name="COMPONENT_CAP_FLAGS_NONE" and description: "No flags set."

hamishwillee: Or even better, think of a good valid first flag (though I suspect you would have done this if it were easy.

message_definitions/v1.0/common.xml
olliw42: thx. done.

message_definitions/v1.0/common.xml
olliw42: thx. Done.

message_definitions/v1.0/common.xml
olliw42: Good question. I obviously took this line exactly as is from CAMERA_INFORMATION. And I already wondered there what exactly this is for, how it should be used. I have no definite answer. 

I believe (and this is only what I believe, not what I know for sure) that in the case of the CAMERA_INFORMATION it is used by QGC to determine if it needs to download the xml again or can use a cached version. This is a bit odd since in the camera definition xml there is also a <version> which appears to not be used at all. 

I agree that this could need some clarification, but that's above my pay grade.

message_definitions/v1.0/common.xml
olliw42: I did the change.

However, IMHO, the current entry should actually be replaced once a first flag is defined. Note that a value of 0 is not exactly a bit flag. I've thus added a respective comment. Pl tell if you like that or if I remove and just do what you suggested.

I agree it would be better to have a first flag. On the quick I again could imagine one, but I will do some more thinking. There must be one, right LOL

message_definitions/v1.0/common.xml
olliw42: Excellent question. IMHO, the xml format for sure should be identical for all components, and not component specific. After all, the purpose is to give GCSes a generic way to draw a UI for components. 

Having said this, the stumbling stone is also clear, it needs input from and working together with the GCS guys, or whatever higher entity than me. Currently this has not yet happened or started. So, at this point in time the xml format is still to be developed.

I mention that it was decided in a previous dev call that I should come up with a message (and an implementation in STorM32, which I did) to actually trigger exactly this. So, in short, in my understanding, this message should go in to facilitate working out exactly what you named as weakness(es). 

For your convenience let me add https://github.com/mavlink/qgroundcontrol/issues/7503

message_definitions/v1.0/common.xml
olliw42: oookkkkaaaaayyyy .... 
(I btw have seen the similar discussion in the COMPONENT_STATUS PR)

In that matter I would not want having to argue but get a very specific call what to do.

I do have, however, a clear opinion on the matter (LOL). I first note that there are already MANY different and partially even contradicting uses of times in MAVLink. One just has to look up and down from there the COMPONENT_INFORMATION is to find (from bottom towards top):
* `<field type="uint64_t" name="time_usec" units="us">Timestamp (synced to UNIX time or since system boot).</field>`
* `<field type="uint64_t" name="time_usec" units="us">Timestamp (since system boot).</field>`
* `<field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>`
* `<field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>`
* `<field type="uint64_t" name="time" units="us">Time of applicability of position (microseconds since UNIX epoch).</field>`
* `<field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>`

and so on. E.g. `time_usec` isn't consistently used for UNIX Epoch but most often for time since boot. `uptime` is only used by the two UAVCAN messages in addition to a `time_usec`, so I would rule out this. time, time_usec, timestamp, time_boot_ms, time_boot, uptime, ....

I also note that there is an overwhelming majority of uses of `<field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot).</field>` and of `<field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>`. So, if one would go by the majority this is what it should be.

I further note that 
* `<field type="uint64_t" name="time_unix_usec" units="us">Timestamp (UNIX epoch time).</field>`
* `<field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>`
* `<field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>`
appear to have been the historically oldest (at least from their message id) - at this time there seems to have been sort of a clear idea of how to use times (with only few exceptions like few `timestamp`). Sad this vision went away.

The `time_boot_ms` thing I used here I obviously have taken from `CAMERA_INFORMATION`.

That's the current status. My conclusions would be.
1. You definitely want to decide soon - and once for all !!!! - what time things you want to use. I would expect that all WIP messages are changed to the new scheme (they are WIP, right, so don't argue with breaking implementations, no one should complain).
2. If I were you I would avoid increasing the mess by not using old time field names but adding in new ones, e.g. new ones without a _usec or _ms. 
3. I would definitely keep the _usec or _ms. I understand you made the decision to not have units anymore in the name, but for the time field I would clearly make an exception, simply because it can be ms or us ... and it is really annoying (and error prove) to always have to look up what it is. Note that this would somewhat help with (2).

Personally I like the historical `time_unix_usec`, `time_boot_ms`, `time_usec` the most ... very clear and unambiguous .. and since they are already used most often using them would minimize the chaos.

My opinion. But I repeat: In that matter I would not want having to argue but get a very specific call what to do. :D:D:D

message_definitions/v1.0/common.xml
julianoes: Yes, this is to determine if the cached file is valid or needs to be re-downloaded. It's also an entry in the xml file.

message_definitions/v1.0/common.xml
julianoes: What about "component supports automatic system ID assignment" by listening to an autopilot (compid 1) and copying the sysid.

message_definitions/v1.0/common.xml
julianoes: I agree, except:

> I would expect that all WIP messages are changed to the new scheme (they are WIP, right, so don't argue with breaking implementations, no one should complain

They are nevertheless somehwat implemented and changing them creates churn and troubles, so I'd avoid it if possible.

message_definitions/v1.0/common.xml
olliw42: > It's also an entry in the xml file.

so, this reads to me as if the value in `component_definition_version` should match the version value in the xml file ... which however brings in a real issue, i.e., is IMHO totally unrealistic.

The content of the xml file can be maintained by the definition file owner, however, the value emitted in `component_definition_version` is not easily accessible to that maintainer, since it's in the component's code. And note: The component won't usually be able to read the xml file!.

IMHO the concept is actually somewhat of a flaw. You know, I'm using this since a while for the CAMERA, and have had repeated cases there I changed my camera definition xml files ... and the only real practical solution I found was to erase the cached file from QGC's folder...

The only way I see to get this working as is, is to have a parameter in the component, to have the user find and read the xml file to get the xml's version value, and then to have the user enter this value into the component's parameter ... that's really anything than user friendly ...

Alternatively, a GCS could be forced to not cache ... in which case the field `component_definition_version` would be obsolete.

I wonder how many users/applications are there which use the camera microservice ... these issues must then have been noted before

message_definitions/v1.0/common.xml
olliw42: ah ... very interesting idea ... 

is this a capability which needs to be announced to anyone? My first though is that it rather is specific to the component. I mean, in which sense would it affect any other component in its behavior if that flag would be set or not? 

also: so far there is only one official MAV_COMP_ID_AUTOPILOT id. I would find it more natural that the component would identify an autopilot by means of MAV_AUTOPILOT/MAV_TYPE, but these rules - as much as I understand it - have not yet finally be settled nor agreed on. IMHO this would be a strong use case for what I was proposing in the MAV_AUTOPILOT/MAV_TYPE/COMPONENT_ID context. Note that if such an autodetection based on componentId would be implemented now it would fail e.g. once the range of autopilot componentIDs would be expanded. Or do I miss the point and it doesn't make any sense to have more than one autopilot componentId?

Can it be ruled out that there could be heartbeats from two autopilots reaching the gimbal, e.g. in multi drone setups?

julianoes: That's true, it doesn't have to announce it. It's up to the integrator to take care of it (or not if not needed).

> that the component would identify an autopilot by means of MAV_AUTOPILOT/MAV_TYPE

Oh yes, you're correct. But let's not open this can of worms here.

> Can it be ruled out that there could be heartbeats from two autopilots reaching the gimbal, e.g. in multi drone setups? 

I hope that never happens. This would open a huge bucket of worms.

olliw42: even though it's maybe not a capability, it's a good idea, I will consider adding it to STorM32 (maybe like then the SystemID parameter is set to 0 that it then autodetermines) THX!

message_definitions/v1.0/common.xml
hamishwillee: OK, so we were supposed to standardise on this format:
```
```
I did change a heap of these to match, where it was clear nothing would break. 

The theory being that this single definition can reflect both boot time and epoch time; the receiver can work out which based on the type.  I'm not completely convinced that it is not better to state an expectation of either epoch or timestamp - generally you do know which will/should be sent, and it would better for implementer not to have to check the the size ... but this is what got agreed.

So
- For this case my _preference_ is to use the "standard". 
  - Yes it is a mess, but it won't get less so by adding more variants.
  - In this case you get to use the units extension without any question :-)
- If there are technical reasons why using ms and specifying "since boot" is better then @julianoes in your hands to provide canonical direction to @olliw42 
- We should ask about the WIP cases.

message_definitions/v1.0/common.xml
hamishwillee: @olliw42  I have separated out the discussion in https://github.com/mavlink/mavlink/issues/1198 - let's see what Gus says. 

My thought is that unless we can assume that component definition files can always transparently replace the previous version (without need for change to component firmware) you're always going to have to have a different URL for every version - or the new definition will break old un-updated firmware. IN that case GCS could use the URL for versioning, and the component would have to be updated with the new URL.

hamishwillee: ... but perhaps continue discussion in 1198 :-)

message_definitions/v1.0/common.xml
olliw42: I believe that in many cases it will not matter if its time since boot or unix epoch since only differences will be checked, i.e. the time of the current message in relation to the last received

most components will have a hard time to send unix epoch

ms is often (?) sufficient resolution and it's just 4 bytes over 8 bytes, which our 32bit microprocessors can handle without overhead

julianoes: > most components will have a hard time to send unix epoch

True.

> ms is often (?) sufficient resolution and it's just 4 bytes over 8 bytes, which our 32bit microprocessors can handle without overhead

Also true and probably the reason why it is also used. And that might explain why it's not consistent. For some messages ms is enough, for others us might potentially be required.

message_definitions/v1.0/common.xml
julianoes: More standardization on the xml format is definitely something to stride for. I think it should mostly live in the docs around it:
https://mavlink.io/en/services/camera_def.html

hamishwillee: > ....this message should go in to facilitate working out exactly what you named as weakness(es).

That was my understand too. Let's get this in, then see whether gimbal XML can be generalised. In any case, it is not the end of the world to have component-specific XML DTDs.

olliw42: > Let's get this in

to correctly understand this, pl let me ask: Is this based on a decision that a general COMPONENT_INFORMATION is preferred over MAV_TYPE specific xxx_INFORMATION messages, or is this rather the same argument, let's start with something and then see?

btw, I have updated the files to latest master, and changed to id 390. I have left time_boot_ms, since it's still not clear to me what the official position is, e.g. in https://github.com/mavlink/mavlink/pull/1186 its currently `uptime`. I however suspect that time_boot_ms for the above reasons might prevail. I also could change to `time_usec`, which seems to be decided. As said, I tried to stay similar to the existing CAMERA_INFORMATION.

olliw42:
I'd like to throw this in:

the heartbeat's `MAV_TYPE` and `MAV_AUTOPILOT` methodology has been significantly improved lately, however what is - IMHO - still missing is a possibility to more precisely tell what type a component is. E.g., for an autopilot it is known if it's a PX4, ArduPilot and so on, which is often a very relevant piece of  information! For components this is not possible, e.g. it's not known if a GCS is QGC or MP, if a gimbal is a Gremsy or STorM32, and so on, yet also this can often be a very relevant piece of  information. While it would have been cute to have this somehow in heartbeat too, this apparently isn't "easy" to do. So - and this is the point - one could think of using `COMPONENT_INFORMATION` for that purpose. At least, there would be some logic to it, given that it already has the `vendor_name`, which provides exactly this info. 

The implication however: For this to reasonably work, essentially all components should be required to support `COMPONENT_INFORMATION`, including the common GCSes! That is it should be possible to send a `MAV_CMD_REQUEST_MESSAGE` to e.g. QGC and it should respond with a `COMPONENT_INFORMATION`.

I'm not sure if that implication is liked, hence this food for discussion.

message_definitions/v1.0/common.xml
olliw42: what about flags to indicate support of params?
(1) not all components might have params, yet might not be totally passive (like some OSDs), i.e. emit e.g. a heartbeat to activate routing
(2) one could have flags for old param method, for ext param method, and the looming new 3rd param method, so everyone interested could easily figure which one to use
(3) in principle (1) and (2) of course could also be figured out by trying and timeout, but there could be advantages if not everyone in the system has to do that, and also not having to try for each param method. Avoids spilling the network

??

julianoes: I was thinking that's already covered by https://mavlink.io/en/messages/common.html#MAV_PROTOCOL_CAPABILITY_PARAM_FLOAT, however, these are autopilot capabilities flags and maybe don't apply to components. Therefore, it would make total sense to add it here. There probably needs to be a flag for "normal" params and param_ext.

julianoes:
> all components should be required to support COMPONENT_INFORMATION, including the common GCSes

I think that sounds sensible.

message_definitions/v1.0/common.xml
olliw42: ah, interesting point ... so let me cite https://github.com/mavlink/mavlink/issues/1153 (which leads to https://discuss.px4.io/t/mavlink-support-for-the-storm32-gimbal-controller-how-to-make-it-compatible/10723/5 and ff) where I was asking for the actual meaning of these fields, and it was concluded that they are unclear and I should ignore them ...

olliw42: note also that the MAV_PROTOCOL_CAPABILITY flags are communicated via the AUTOPILOT_VERSION message (somewhat ironically), which isn't mandatory for components to be supported, see https://github.com/mavlink/mavlink/issues/1156

(note that the docs are not yet accordingly updated, see https://github.com/mavlink/mavlink/issues/1175)

olliw42:
to move this forward, I took julianoes' responses as "decision", and changed this
- added COMPONENT_CAP_FLAGS_PARAM and COMPONENT_CAP_FLAGS_PARAM_EXT
- specified that support of COMPONENT_INFORMATION should be mandatory

edwinhayes:
This seems to rather assume that each component is a separate physical device, but that is not always the case; multiple different bits of software running on one computer may present as separate logical components.

1. Fields like 'firmware version' and 'hardware version' are not appropriate in these cases.
2. Requiring components to support the COMPONENT_INFORMATION request substantially increases the minimum complexity of a MAVLink component.

olliw42:
> This seems to rather assume that each component is a separate physical device, but that is not always the case; multiple different bits of software running on one computer may present as separate logical components.
>
> 1. Fields like 'firmware version' and 'hardware version' are not appropriate in these cases.

this comment appllies to ANYTHING in a MAVLink network, so anything in the spec would be vulnerable to it and according to you objectable

the spec is such that each component has it's individual component ID, by which also the type of the component - if it is a gimbal, a servo, a gps, a gcs, .... - is determined. It certainly makes sense to talk about such fields for each such MAVLink individuum, except in the weirdest cases. 

This doesn't contradict in no wayl with having several of them in one physical device, e.g. a camera and a gimbal, or a GSP and an IMU, and so on.

If these fields don't apply, set them to "".

I don't think this is a valid objection, and don't think it should uphold moving forward.

> 2. Requiring components to support the COMPONENT_INFORMATION request substantially increases the minimum complexity of a MAVLink component.

yes, certainly, but maybe you want to research the reasons for why this requirement was added - it's being discussed in quite some depth across several issues/PRs lately. There is an issue which needs a solution. I can tell you that part of the answer is that other suggestions to solve the underlying issue were found to be not acceptable. 

I would have hoped that you also would have come forward with what you think is a better solution to the issue then.

edwinhayes:
> this comment appllies to ANYTHING in a MAVLink network, so anything in the spec would be vulnerable to it and according to you objectable

Not really: AFAIK, at present the only message which Components are obligated to transmit is `HEARTBEAT`, so that rather limits the number of fields which could be subject to this objection.

> If these fields don't apply, set them to "".

Is this really best practice?

> I can tell you that part of the answer is that other suggestions to solve the underlying issue were found to be not acceptable.

What underlying issue?  To OEMs today who have products with perfectly functional MAVLink implementations, there is no issue.  This change is adding new functionality.  Which is great; but as proposed, it would be a breaking change; any existing components which were unable to implement the proposed micro-service would be rendered non-compliant.

> I would have hoped that you also would have come forward with what you think is a better solution to the issue then.

Well, as a matter of principle, no: as this would be a breaking change, the onus should be on the proposer(s) to address any issues.

But regardless; yes, I have a solution: don't make this mandatory.

olliw42:
frankly, if this sort of argumentation and mindset would prevail MAVLink never could improve. It's 2019, not 2009.

there had been stronger breaking changes in the past

it's IMHO, as matter of principle, the duty of "criticisers" to know about the underlying issues and reasons for suggestions, otherwise a meaningful discussion isn't possible

the proposer's suggestion had been called sensible by a dev team member 

just my opinions. I'm not deciding anyway. I just wish that MAVLink is advancing. I stay with my first statement.

:)

olliw42:
I've update to latest master, and whereby also did
* changed id to 395
* removed the controversial "must support" and replace with "consider to support" (I got the impression by now that the issue of how to determine the type/brand of a component won't find any non-contentious solution in any near future, and this shouldn't uphold this here)
* added a FTP flag, jumping the gun in view of #1218

edwinhayes:
Seems legit.

How about `COMPONENT_CAP_FLAGS_LOGICAL` (or 'virtual' or something similar) to allow distinguishing logical components which don't represent (by themselves) a single physical device (and thus whether to ignore the hardware version field)?   Also `COMPONENT_CAP_FLAGS_PING`; might useful to know if the component supports the ping protocol?

olliw42:
great to hear that

maybe one should wait few days (whatever this means) to see if #1218 maybe comes to a quick conclusion. The same issue applies to the component definition URI. Hamishwille came up with some good alternative suggestion than what I was jumping to. May save us one iteration.

olliw42:
issue #1218 got resolved, and I've accordingly removed the FTP capability flag
so, I think, now this holds
> I would say this is good enough for WIP and we should get it in.

:)

olliw42:
I'd like to mention that I recently made a first attempt of adding COMPONENT_INFORMATION support to QGroundControl: https://github.com/mavlink/qgroundcontrol/pull/8007

It's not yet fully worked out of course, but working to a point that it's functional.

A major pain point currently however is the difficulty to get the COMPONENT_INFORMATION message into the flight stack and QGC. I thus would very much like to bring this PR back to consideration.  It really would easy things if it would be in the mavlink repo.

The id of 395 seems to be still good. if I should change it pl tell and I certainly will :)

I'd like to emphasize again how cool this IMHO is. I really like the GUI improvements possible through this a lot. :):):)

Cheers, olli

olliw42:
@hamishwillee @julianoes 
to me it looks that all discussion points were resolved, and that the PR as is was kind of found acceptable. 
However, if merging this needs more discussion I'll be happy to join the dev call this Wednesday if it would help. 
I have seen that this PR has a devcall label, but looking through the past dev call notes this by itself appears not to ensure a topic to make it on the list. So, if you think it should be discussed and I should attend I would appreciate it to be added to the list of topics.
thx, Olli

hamishwillee:
@olliw42 Yes, did not get covered. I'm OK with this going in as WIP. @julianoes any objection to merging?

olliw42:
@hamishwillee I would briefly show up at the dev call in order to briefly discuss this here, if that's ok.

hamishwillee:
Hi @olliw42 ,
Ah, we saw the phone number, but no sound! Sorry we didn't get a chance to talk.

@julianoes Can you please comment on above?

olliw42:
@julianoes 
please !

for your help, in https://github.com/mavlink/mavlink/pull/1176#pullrequestreview-279511401 you said: 
> I would say this is good enough for WIP and we should get it in.

:)

olliw42:
@DonLakeFlyer asked me to hold off pushing it: https://github.com/mavlink/qgroundcontrol/pull/8007#issuecomment-554023386

:) :) :)

julianoes:
Yes I think it would be nice to have this in as WIP. The only confusing thing is that there is a change of the pymavlink submodule in this PR. Is that on purpose?

olliw42:
no, this is not on purpose
I noted this but have no clue what has happened. Since I just did what I did all the time so far I assumed that's somehow how it is (this also happened in some other PRs of mine which I updated ... I always do git submodule update --recursive). Sorry if that is not how it is supposed to be.
If not, any hints how I could correct?

julianoes:
Looks fine now, it was probably just a merge or rebase were it was accidentally changed.

olliw42:
the pymavlink submodule changed, and I forgot that a submodule update alone doesn't do it, but that one also has to cd into the submodule and fetch and checkout the desired commit ... I really wish git would have gotten the submodule thing right, it's not clear to me why a submodule update doesn't really do an update ... anyway :) :)

hamishwillee:
@julianoes @olliw42 I lost track of this. What is it waiting on?

olliw42:
I ask this myself :)

since the project is not really moving on with lightning speed I was pondering the last week or so about the situation. I was going to ask if one could put it on the next dev call's TOP, to discuss ways forward. Generally, the challenge with such a project is that it needs good coordination points in several places, mavlink, a component/flightstack, a ground control station, but there seem to be some only in the first two places. No idea what can be done. I wonder if it's a dead horse.

on my workbench the situation is as such: The STorM32 supports the COMPONENT_INFORMATION message (albeit still at the old 700 id). It returns a xml file, either an internet URL, but it also supports MAVFtp, i.e., the file can be downloaded from it via MAVftp. This implies a first draft of an xml design is also existing (I actually investigated both QGC and MP for what they currently want to make it more likely it's acceptable to both). I have mocked ArduPilot so that it can route the message. And there is a first working implementation of support in QGC, which does bring some significant improvements to how the parameters are presented. So, there is a working backbone. It should not be difficult to recreate it on anyone's workbench. 

Clearly, once a backbone is agreed on and merged, it is IMHO only a matter of creativity of what to improve further. I would have plenty of ideas LOL. But I'm actually currently not very hopeful these will ever materialize, pl see 2nd paragraph.

my 2 cents  :)

auturgy:
We actually did discuss this briefly on the call today.
I support it going in.

Regards,

James

olliw42:
oh ... I'm out of schedule ... I thought the next is next week LOL

hamishwillee:
OK. Thanks. @julianoes You happy to merge this? It is WIP.

hamishwillee:
OK. This is merged. Thanks @julianoes @olliw42 et al. Interested to see what magic you make out of it :-)

olliw42:
GREAT :)
thx

question: how long will it take for this to appear in QGC?

(I've seen that there appears to be a PX4BuildBot which on Wed was doing an update of the mavlink submodule, but it seems it is not triggered by any commit to the mavlink submodule, hence my question)

hamishwillee:
I don't know how QGC handles mavlink updates.

auturgy:
From memory usually @donlakeflyer or @dagar bump the submodule hash

olliw42:
thx for the feedback

so, let's hope for the best :)

PR #1178, opened 2019-07-10
Add PROTOCOL_VERSION and friends to minimal.xml

hamishwillee:
As discussed in https://github.com/mavlink/mavlink/issues/1175 , [PROTOCOL_VERSION](https://mavlink.io/en/messages/common.html#PROTOCOL_VERSION)  (and hence [MAV_CMD_REQUEST_PROTOCOL_VERSION](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_PROTOCOL_VERSION)) are mandatory to implement for a MAVLINK2 system.

Therefore they should be in minimal.xml. 

This PR is created to test this assumption. 

Notes/questions
- PROTOCOL_VERSION is marked wip in common - should this be removed?
- that the next step would be to strip the messages out of common.xml and include dialect in common. Any reason why not to do this?
- Would it be better to put the protocol methods in protocol_version.xml and include that?

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -381,6 +381,14 @@
         <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
       </entry>
     </enum>
+    <enum name="MAV_CMD">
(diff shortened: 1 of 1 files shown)


hamishwillee:
This was discussed in [20190710-Dev-Meeting](https://github.com/mavlink/mavlink/wiki/20190710-Dev-Meeting):
- Yes, moving this here makes sense.
- WIP can be removed. 
- This can be merged. 

This question was not addressed "Would it be better to put the protocol methods in protocol_version.xml and include that?"

PR #1180, opened 2019-07-11
Deprecate MAV_CMD_REQUEST_* in favor of MAV_CMD_REQUEST_MESSAGE

olliw42:
the title says it
takes up https://github.com/mavlink/mavlink/pull/1176#discussion_r302010705

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2076,7 +2076,7 @@
       </entry>
       <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION" hasLocation="false" isDestination="false">
         <deprecated since="2019-08" replaced_by="MAV_CMD_REQUEST_MESSAGE"/>
-        <description>Request MAVLink protocol version compatibility</description>
+        <description>Request MAVLink protocol version compatibility. All receivers should ACK the command and then emit their capabilities in an PROTOCOL_VERSION message</description>
         <param index="1" label="Protocol">1: Request supported protocol versions by all nodes on the network</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
@@ -2216,8 +2216,8 @@
         <deprecated since="2019-08" replaced_by="MAV_CMD_REQUEST_MESSAGE"/>
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Re-request a CAMERA_IMAGE_CAPTURE message.</description>
-        <param index="1" label="Number" minValue="0" increment="1">Sequence number for missing CAMERA_IMAGE_CAPTURE message</param>
+        <description>Re-request a CAMERA_IMAGE_CAPTURED message.</description>
+        <param index="1" label="Number" minValue="0" increment="1">Sequence number for missing CAMERA_IMAGE_CAPTURED message</param>
         <param index="2" reserved="True" default="NaN"/>
         <param index="3" reserved="True" default="NaN"/>
         <param index="4" reserved="True" default="NaN"/>
@@ -6031,7 +6031,7 @@
       <field type="float" name="focusLevel">Current focus level (0.0 to 100.0, NaN if not known)</field>
     </message>
     <message id="261" name="STORAGE_INFORMATION">
-      <description>Information about a storage medium. This message is sent in response to a request with MAV_CMD_REQUEST_MESSAGE and whenever the status of the storage changes (STORAGE_STATUS).</description>
+      <description>Information about a storage medium. This message is sent in response to a request with MAV_CMD_REQUEST_MESSAGE and whenever the status of the storage changes (STORAGE_STATUS). Use MAV_CMD_REQUEST_MESSAGE.param2 to indicate the index/id of requested storage: 0 for all, 1 for first, 2 for second, etc.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="storage_id">Storage ID (1 for first, 2 for second, etc.)</field>
       <field type="uint8_t" name="storage_count">Number of storage devices</field>
@@ -6052,7 +6052,7 @@
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
     </message>
     <message id="263" name="CAMERA_IMAGE_CAPTURED">
-      <description>Information about a captured image</description>
+      <description>Information about a captured image. This is emitted every time a message is captured. It may be re-requested using MAV_CMD_REQUEST_MESSAGE, using param2 to indicate the sequence number for the missing image.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint64_t" name="time_utc" units="us">Timestamp (time since UNIX epoch) in UTC. 0 for unknown.</field>
       <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
@@ -6111,7 +6111,7 @@
     <message id="269" name="VIDEO_STREAM_INFORMATION">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Information about video stream</description>
+      <description>Information about video stream. It may be requested using MAV_CMD_REQUEST_MESSAGE, where param2 indicates the video stream id: 0 for all streams, 1 for first, 2 for second, etc.</description>
       <field type="uint8_t" name="stream_id">Video Stream ID (1 for first, 2 for second, etc.)</field>
       <field type="uint8_t" name="count">Number of streams available.</field>
       <field type="uint8_t" name="type" enum="VIDEO_STREAM_TYPE">Type of stream.</field>
@@ -6128,7 +6128,7 @@
     <message id="270" name="VIDEO_STREAM_STATUS">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Information about the status of a video stream.</description>
+      <description>Information about the status of a video stream. It may be requested using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint8_t" name="stream_id">Video Stream ID (1 for first, 2 for second, etc.)</field>
       <field type="uint16_t" name="flags" enum="VIDEO_STREAM_STATUS_FLAGS">Bitmap of stream status flags</field>
       <field type="float" name="framerate" units="Hz">Frame rate</field>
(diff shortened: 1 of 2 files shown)


hamishwillee:
I fully support moving to this message, but I am not sure that it can be deprecated if there is no implementation or plan for implementation on some of our core flight stacks.

If nothing else this may encourage people to start at least supporting the new message type.

Let's see what the Devcall says. 

FYI @auturgy @julianoes

olliw42:
I continue to be highly puzzled, I fail to piece things together to a consistent story

I mean, if there was a decision as reported in https://github.com/mavlink/mavlink/pull/1176#discussion_r302010705 (which I have no reasons to doubt), then this here would be the logical consequence and it's puzzling that there is some hesitation and need for further devcall discussions. It comes across more like "yeah we want, but not really actually" ...

btw, if there is indeed no support yet of the MAV_CMD_REQUEST_MESSAGE in PX4, this should be really most easy to add, it really can take only few minutes for someone somewhat familiar with the code, i.e. this shouldn't be an argument. Btw, in ArduPilot there is at least already something (https://github.com/ArduPilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L3993-L3995), and also STorM32 uses it to some extend since a while already.

I understand that there needs to be a transitional state, but this I only can assume must have been part of the decision making process having lead to the above decision. 
I also understand that it's not so nice that it especially affects the micro services, and that's why my proposal would have been different, but again, I only can assume that this must have been part of the decission making process having lead to the above decission.
I also understand that it sometimes can be difficult to find a good decission, the gimbal message case is clearly such an example, but that's the pre-decission state while here a decision is reported to have been made.

Hopefully my confusion is at least comprehensible. I find it somewhat inconvenient to say the least that it appears to be so difficult to get answer to simple things. I mean, at the end of the day I just want to know what I should implement in STorM32, and not having to do a new best guess with each new version...

But as you say, let's see.

It was a good git training for me to do this PR, so at least I learned something LOL

hamishwillee:
Good to see ArduPilot are on their toes. I don't think you will find the same in PX4 for this message.

I'm sorry you're confused. The simple reason is that the process for deprecating a message is that we need buy-in from the dev call. This follows some rancour about messages that were prematurely deprecated in the past when some stacks did not update.

Raising the PR is a good thing. This change should happen. It may be that the discussion will be the push needed for stacks to update.

amilcarlucas:
Any plans to include minimal.xml in the other xml files and remove the duplicated messages in common.xml ???

hamishwillee:
> Any plans to include minimal.xml in the other xml files and remove the duplicated messages in common.xml ???

@amilcarlucas Yes, but when I tried it I discovered the (my) code to support multiple inclusion was flawed. There are now two PRs in ArduPilot/pymavlink to fix that (one from me, the other from another developer). They achieve the same thing in different ways - I'm hoping PeterB can decide what he likes most. 

When either goes in, the minimal.xml can be included in standard.xml, and standard.xml in common.xml.

hamishwillee:
@olliw42 Would you like to try put a PR to add support for this in PX4? If not, I can ask Julian or try do it myself.

olliw42:
@hamishwillee 
thx for your confidence in my capabilities, but I'm pretty unfamiliar with PX4 ... 
since it's supposed to be a simple code thing I could make a suggestion and PR it, if that would be considered to be of any help, but I would NOT be able to compile and thus not able to test it ... so would be total guesswork ... so I doubt it would useful if I were to PR ...

so, pl just tell, if it makes sense if I would PR under these conditions

btw, I just looked at it, and I guess it would be replacing these lines https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L460-L487 with

```
if (cmd_mavlink.command == MAV_CMD_REQUEST_MESSAGE) {
	/* send requested message, param1 holds the requested message ID */
          /* currently only replaces other existing request type commands */
        
        switch ((int)cmd_mavlink.param1) {
            
        case MAVLINK_MSG_ID_AUTOPILOT_VERSION:
            _mavlink->send_autopilot_capabilites();
            break;
            
        case MAVLINK_MSG_ID_MESSAGE_INTERVAL:
            get_message_interval((int)cmd_mavlink.param2); // param2: message ID for which the interval is requested
            break;
            
        case MAVLINK_MSG_ID_STORAGE_INFORMATION:
			send_storage_information((int)(cmd_mavlink.param2 + 0.5f)); // param2: Storage ID
            break;
            
        case MAVLINK_MSG_ID_FLIGHT_INFORMATION:
            send_flight_information();
            break;

        case MAVLINK_MSG_ID_PROTOCOL_VERSION:
            _mavlink->send_protocol_version();
            break;

        default:
		break;
        }            

} else if (cmd_mavlink.command == MAV_CMD_REQUEST_PROTOCOL_VERSION) {
		/* send protocol version message */
		_mavlink->send_protocol_version();

} else if (cmd_mavlink.command == MAV_CMD_REQUEST_AUTOPILOT_CAPABILITIES) {
		/* send autopilot version message */
		_mavlink->send_autopilot_capabilites();
        
} else if (cmd_mavlink.command == MAV_CMD_REQUEST_STORAGE_INFORMATION) {
		if ((int)(cmd_mavlink.param2 + 0.5f) == 1) {
			send_storage_information(cmd_mavlink.param1 + 0.5f);
		}
        
} else if (cmd_mavlink.command == MAV_CMD_REQUEST_FLIGHT_INFORMATION) {
		send_flight_information();
        
} else if (cmd_mavlink.command == MAV_CMD_GET_HOME_POSITION) {
		_mavlink->configure_stream_threadsafe("HOME_POSITION", 0.5f);

} else if (cmd_mavlink.command == MAV_CMD_SET_MESSAGE_INTERVAL) {
		if (set_message_interval((int)(cmd_mavlink.param1 + 0.5f), cmd_mavlink.param2, cmd_mavlink.param3)) {
			result = vehicle_command_ack_s::VEHICLE_RESULT_FAILED;
		}

} else if (cmd_mavlink.command == MAV_CMD_GET_MESSAGE_INTERVAL) {
		get_message_interval((int)cmd_mavlink.param1);

} else {

```

If that's close to what you were seeking, pl don't hesitate to grab and PR it :)

hamishwillee:
@olliw42 Thanks very much - I am sure this will help me get started if I do the work (can't accept a PR that is untested, but I can probably do so with this starting point).

olliw42:
this wasn't the clear yes/no :) but I guess you're agreeing I shouldn't PR
if the snippet is helpful, great.

I know I should spend more time with PX4, but so far I unfortunately didn't yet had good luck with it (my failure for sure). Vacation time is nearing, maybe ...

hamishwillee:
Hi, @olliw42 . Yes, a clear "no" on the PR until you're able to put time into build and test :-)
Who knows why you have problems. I have months of no problems at all and then get stumped. But that's all software!

olliw42:
thx for the clear no :)

>Who knows why you have problems.

I would say I just don't understand the parameters, maybe I should eventually ask in the forum, I guess, but that's OT

have fun :)

amilcarlucas:
From a Ardupilot point of view, we can deprecate it. We do support the new MAV_CMD_REQUEST_MESSAGE

hamishwillee:
Thanks @amilcarlucas. This is also supported in PX4.

Originally this is absolutely what we intended. However subsequently I've been considering that this might be better as an _alternative_ rather than _replacement_ for specific messages. From a user perspective it is better to just say "use this message to request any message". 

But we've talked a lot about service design, and how it might make sense to split services into into individual xml files etc. If you do this, then the API definition is much more clear if you have a specific message for just the service call.

@julianoes Am I just overthinking it?

julianoes:
@hamishwillee I like `MAV_CMD_REQUEST_MESSAGE`, it makes a lot of sense to me, and separate messages for each "service" would be a pain.

hamishwillee:
OK, gave this a final review. I updated some of the requested messages to reflect that the original requestor message has parameters - these will be requested with `MAV_CMD_REQUEST_MESSAGE.param2` - see https://github.com/mavlink/mavlink/pull/1180/commits/3e5ba0574f6d03b65768bd2f0148288cbe09adcf

Merging now.

hamishwillee:
@olliw42 Thanks for this. Sorry it took so long to get around to merging. I am aware of your other posts, but will not get to do almost anything on my inflow in the next two weeks. 

@julianoes Can you please take in hand managing any changes needed for MAVSDK?

@dogmaphobic FYI, this deprecates specific messages the camera API uses for requesting an information message (e.g. MAV_CMD_STORAGE_INFORMATION) in favour of the generic requestor message `MAV_CMD_REQUEST_MESSAGE`. This essentially just returns any message based on the message ID specified in `MAV_CMD_REQUEST_MESSAGE.param1`. Where the original requestor message takes params, these are assigned to param2+ and that information is defined in *requested* message.

Nothing stopping you using the deprecated messages for now, but would be great to update and one day delete those. In particular the WIP ones.

dogmaphobic:
> Nothing stopping you using the deprecated messages for now, but would be great to update and one day delete those. In particular the WIP ones.

Yup, this new system makes a whole lot of sense. With that said, there is a great deal of code out in the wild using the existing commands. From a receiving point of view, it's simple to support both as described by @olliw42 above. It will be tricky from a sender's perspective though. I need to pick one or the other. Otherwise I have to send two commands each time I want to request something. If I switch QGC to use the new MAV_CMD_REQUEST_MESSAGE, existing systems will not send me back anything. How do we solve that?

hamishwillee:
> If I switch QGC to use the new MAV_CMD_REQUEST_MESSAGE, existing systems will not send me back anything. How do we solve that?

@dogmaphobic Both ArduPilot and PX4 should send you back the message right now, provided your original requester message has no meaningful parameters.  Most don't -the only method that might fail AFAIK is using it for CAMERA_IMAGE_CAPTURED - because MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE takes a sequence , and you have to specifically code for what that index means.

But I guess you are talking about other systems than these two? 
Do we have any idea who?

dogmaphobic:
What I mean is that almost all those request messages marked for deprecation are being actively used everywhere. Out of all of them only two are not used by QGC:

- MAV_CMD_REQUEST_FLIGHT_INFORMATION
- MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE

What should QGC do? If I switch to the new, generic `MAV_CMD_REQUEST_MESSAGE`, existing cameras/devices will not respond until they too are updated. That will break all existing camera and video handling.

I'm already running into issues with the Herelink where they are using since deprecated (WIP) mavlink messages that makes it a glorified paper weight if you want to create your own ground station app for it. This will only make it worse. Don't get me wrong, I do agree that this is a much more logical approach to these requests. I just don't quite get how to handle the existing commands.

hamishwillee:
@dogmaphobic All good points. This sort of thing has always been a problem for us. 

MAVLink has two answers this, neither of which are "live", so the upshot is, stick with what you're doing and we'll have to update later.

The two things we're looking at are microservices versioning and component information querying. The first basically says "we support camera protocol v2" which would be this new approach, or we support v1, which would be the old approach.
The second says we support this set of MAV_CMDs (and I think we'll need to test the message).

PR #1182, opened 2019-07-19
Clarify use of multiple batteries.

hamishwillee:
This attempts to address issues from https://github.com/PX4/Firmware/pull/12034#issuecomment-512361816

1. How BATTERY_STATUS.id should be allocated.
   - Note, sequentially from 0. There is no concept of "main" or other associated with the index.
   - I pushed this out to SMART_BATTERY_ messages too.
2. How should SYS_STATUS report when it has multiple batteries? - Not done.


hamishwillee:
[SYS_STATUS](https://mavlink.io/en/messages/common.html#SYS_STATUS) reports three battery related values: `voltage_battery`, `current_battery`, `battery_remaining`. Similarly [HIGH_LATENCY2.battery](https://mavlink.io/en/messages/common.html#HIGH_LATENCY2) reports percentage charge.

It is not clear which battery(ies)  the value(s) reflects on a multi-battery system:
- `voltage_battery`, `current_battery` can only be for one battery.
- `battery_remaining` (or `HIGH_LATENCY2.battery`) could reflect a combined capacity.
- They could all be for just one battery, but which one?:
  - The battery with id=0 (or highest id?)
  - The battery with the most or least capacity
  - The battery with the lowest or highest voltage

In https://github.com/PX4/Firmware/pull/12034 they were being populated with the data from the connected battery with the lowest percentage remaining. This is for safety: If any legacy system is only looking at the status of one battery (i.e., using SYS_STATUS instead of BATTERY_STATUS), it should see the lowest battery. If one battery is full and the other almost empty, a system should not see the full battery and assume everything is fine, unless it is explicitly handling multiple batteries.

How should SYS_STATUS report when it has multiple batteries?

@auturgy @DonLakeFlyer @julianoes @bkueng

hamishwillee:
Can we assume that all batteries reported power the system as a whole, or do we have to assume that one might power motors, and another other systems? If we can assume that batteries cumulatively power the system then I'd be tempted to make the capacity fields reflect the cumulative charge for all batteries and deprecate the voltage/current in SYS_STATUS.

IF we can't make any assumptions, then there is no "cumulative value". In that case we should either deprecate the fields altogether, or report the minimum battery.

Thoughts?

WickedShell:
ArduPilot will report the "primary" (the first or index 0) battery on SYS_STATUS. Without an ID on it I find the PX4 one to be problematic as you get unclear discrepancy when it switches what battery is being reported, whats worse is if it's based purely on a percentage remaining then from a display perspective it's not obvious at all to the user they are seeing different information.

The assumption that all batteries power the system is definitely not a valid one. I've seen (or use) vehicles where you have seperate battery packs for forward vs vertical flight, separate packs for payloads, and due to a limitation in MAVLink reporting actually monitor power rails through out the aircraft as a battery input even though it technically isn't a battery (for example the servo power supply when adding current monitoring to it).

auturgy:
Not to mention fuel flow and level also reported via battery messages (to hook into existing failsafe mechanisms), and the virtual battery that sums from esc telemetry.


Regards,

James

hamishwillee:
@WickedShell @auturgy Thanks very much. So in summary, you'd prefer SYS_STATUS battery information be for battery id=0?

FWIW I'd personally be happy with that, but also mark the fields as "do not use" in the description so people understand this is not something that can reliably be used for battery status of the whole system. 

What about for `HIGH_LATENCY2.battery`, which is still WIP?

WickedShell:
> So in summary, you'd prefer SYS_STATUS battery information be for battery id=0?

Yes, or whatever a vehicle determines to be it's primary battery if that is not battery 0 for any reason. IE if you had a parameter on the vehicle that said "battery 2 is my primary battery" then I could see sending that down in SYS_STATUS. But unless your idea of primary changes in flight (which while not impossible is very rare), there doesn't seem to be a lot of reasons you couldn't just swap your setup such that the primary was id 0 again.

> What about for HIGH_LATENCY2.battery, which is still WIP?

Presumably the primary battery again, but if I'm being honest I don't feel very strongly about high latency2 except that having WIP stuff like that forever seems counter productive, but that's not a useful comment on the battery part :)

hamishwillee:
Thanks. I'm a little unsure of the idea of primary battery though. Some vehicles have 2 peer batteries.

I guess we could document as something like "if your vehicle has a primary battery, id=0 should map to that battery?

What do you think of the idea of also (effectively) deprecating the battery fields in SYS_STATUS?

DonLakeFlyer:
> So in summary, you'd prefer SYS_STATUS battery information be for battery id=0?

That makes a lot of sense to me as well since it maintains current behavior. QGC has a single battery icon on the toolbar which goes from yellow to red based on remaining battery. The question from a ground station side of things is when running two batteries whether it is still possible to show such a display. Should a ground station base that on SYS_STATUS?

hamishwillee:
> QGC has a single battery icon on the toolbar which goes from yellow to red based on remaining battery. The question from a ground station side of things is when running two batteries whether it is still possible to show such a display. Should a ground station base that on SYS_STATUS?

@DonLakeFlyer 

I'll make sure we discuss this in the dev call this week and get you some clarity.

FWIW IMO it is possible to show such a display, I'm not sure you would want to. 

I see three options:
- Display each of the batteries individually based on `BATTERY_STATUS` and `SMART_BATTERY_` messages. This is conceptually easy, and copes with cases where there are batteries for different purposes (the message includes battery purpose).
- Display some sort of cumulative battery status that aggregates each of the battery messages (that looks like what we have now). 
  - I like this idea but it is more complicated because batteries have different types, and I'm not sure how easy it is to sensibly aggregate - i.e. separate battery on FC and main motors can't just be added by "cumulative capacity". 
  - This also highlights flaws in our battery warning/safety thinking, since in a multi battery world you could have plenty of power in one battery type but not another, and a single voltage warning level isn't going to necessarily be sufficient.
- Display just the primary battery as we do now. I think we can do better.

> Should a ground station base that on SYS_STATUS?

As above, I think not. This is also why I'd like to mark the battery information in `SYS_STATUS` as deprecated so it is clear that you shouldn't use this information.

olliw42:
I would suggest a totally different view on the data in SYS_STATUS

In the above the spirit seems to make the data relatable to any of the batteries in some way or the other, i.e. to define them through the sources of information. 

In contrast I would suggest to define them rather by the potential actions the values may induce or consequences they may have. 

Actions/consequences could be failsafes, warnings, displays, and such.

In that interpretation it would be up to the flight controller implementation to specify what the "system status" as regards the power system is. For instance, if a system has two batteries it would be up to the flight stack to determine if it's a "simple" two battery setup and send the lowest voltage or if it's a smarter two battery setup and send the larger voltage if e.g. an auto switchover would have disabled the low-voltage battery, and so on. Whatever the systems hardware configuration is and whatever the flight controller implementation supports, SYS_STATUS would be the flight controller's means to inform the world about what it thinks the best and most representative data would be, and on which data it want's e.g. GCSes and so on to act, e.g. emit a speech warning, display an icon, and so.

The battery messages in contrast would indeed be specific to whatever sources there are.

There would be in general no reasonable way for externals to determine how the data in SYS_STATUS is composed of the individual battery data, unless one would have exact knowledge of what the flight controller stack is doing. That's not a flaw, but indeed the purpose, and it's in fact not really relevant to the external in order to make informed decisions on failsafes etc. 

So, I would specify SYS_STATUS by saying that these are the values decisions are being made on by externals (and that it's the flight stack's duty to provide sensible data to allow reasonable decisions being made by externals, for whatever battery configurations it supports).

IMHO that's the only reasonable way to make sense of it. After all, it's called SYS_STATUS.

DonLakeFlyer:
The description of SYS_STATUS from @olliw42 makes a lot of sense to me. Vehicle is in control as to the single display of "System Status".

hamishwillee:
@olliw42 @DonLakeFlyer 
1. I'd assumed that it would be hard for flight stack to provide a single view of system battery status  - so supplying individual battery information and letting GCS aggregate would make sense. Thanks for the correction.
2. I guess if a flight stack has to provide a representation of a complex battery system where voltage or current don't make sense, it simply wouldn't supply them? Does that mean we need to allow `SYS_STATUS.voltage_battery` to have a -1 value?

DonLakeFlyer:
QGC assumptions:
* SYS_STATUS.voltage_battery = UINT16_MAX means unavailable
* SYS_STATUS.current_battery = -1 means unavailable

hamishwillee:
Thanks. This now captures all the discussions above - @DonLakeFlyer @olliw42 et al, please re-review.

hamishwillee:
Discussed in the meeting last night. There was some confusion about the proposal and whether it is the correct approach/whether it can in practise be implemented by an autopilot.

The proposal as captured by PR "today" is that the `SYS_STATUS` battery fields reflect a _notional_ battery state from the autopilot that a user can use to determine whether the vehicle is going to fail due to out of battery. Ie it is just a single warning indicator with capacity, either from the capacity field or guessed from the voltage/current info supplied.
In a simple one battery system this would be just the values for the single main battery. For a more complex system it would be whatever the autopilot thinks is a good indicator for the user. That could be as complex or as simple as the autopilot wants to make the calculation.

Again, the theory is 
- not that this is a single battery, but a "notional" overall power status. 
- that an autopilot is more capable of calculating the power status than GCS or end user. 

@auturgy [please excuse me if I misrepresented you here!]
- was not sure how this would be implemented in code and would like some examples. 
- Wanted to understand how this is better than letting the user determine what battery is displayed as ArduPilot does.

I'll hand over to @olliw42 et al to discuss, but I think that @auturgy has a point. 

The problem is that an autopilot can have any combination of batteries, and expecting the autopilot to be able to work out the notional "correct" notional status from an arbitrary set is either going to be impossible to code, or will be overly simple resulting in a  deceptive estimation. Neither good for the user.

As an end user I would probably prefer that the GCS display every battery, and individually allow me, as the user, to set failsafes on those (or get them from airframe config for fixed-config vehicles). For complex systems I would certainly prefer that to presenting just one battery as ardupilot does. I'd be OK in short term with making the SYS_STATUS battery be just one battery, and indicate that this might be the battery with id=0 or user configurable. Longer term though, we could suggest "do not use".

auturgy:
I agree the intent, but practically doing this is just providing numbers without context.  As SYS_STATUS doesn't support MAV_BATTERY_FUNCTION or BATTERY_STATUS.id filling it with values based on autopilot logic doesn't necessarily aid an external decision (and the autopilot should be making the decision internally anyhow), or give greater awareness to the operator. How does the user know which battery the data in SYS_STATUS represents?  How do they know if the source changes?  Without knowing those things, how do you make a decision / determine an action?
ArduPilot supports multi battery systems via BATTERY_STATUS (and BATT2 for older systems), so the issue is how much info SYS_STATUS can be expected to provide.  I think sending data with a known context (ie battery 0, as we do presently) at least ensures that the data sent is useful.  If this is documented appropriately, I think it more generally "safe" than just sending lowest voltage.
An alternative might be something like TIME_REMAINING estimates (which also covers non-battery vehicles), or replacing the current field with a context identifier (battery id or function).

DonLakeFlyer:
Hmm, complicated and messy. 

But that said I think the ability for an end user to understand **simply** whether they are about to fall out of the sky is a requirement. To me that means a single indicator, whatever that may be: percentage indicator, color indicator, time indicator. To me it's a hard requirement that it exists and is a single thing. You can't for example make an end user of a commerical vehicle look at multiple pieces of information to determine if they are about to crash. I hear this sort of thing from commerical vehicle folks who user QGC custom builds for their vehicles all the time. It has to be simple. Don't expect user to pay attention to complex things.

That said, doesn't the failsafe settings associated with each battery tell you the information you need related to flight status? Those should be set to the truth with respect to things like forced RTL and so forth. If you know those and understand rates of decline for each then you should also be able to provide some sort of information as to how far away from the failsafe you are.

Sure you could create a non-deterministic setup with respect to power. Which maybe has one power source used for something other than flight. In that case you might get random rates of decline which would cause the single reported status to fluctuate up/down. But I would contend that is a crap hardware setup and you should't fly that way.

Don't know if I'm making any sense with that. But to me if you have the concept of failsafes (which relate to flight termination), then you should be pretty close to having the information you need to figure something out and provide a single point of information.

Or to put it more simply: If a power user can stare at displays of multiple power sources and figure out for themselves whether they are in trouble or not. Then whatever that power user is doing can be codifed into a set of rules which can be coded to a single piece of information. If the system is so complex that is not possible I would contend you shouldn't be flying it.

hamishwillee:
@DonLakeFlyer @auturgy To summarise choices:
1. SYS_STATUS battery values are for a particular predictable real battery. 
   - either hardcoded to be whatever is is `BATTERY_STATUS.id=0` or set as a user parameters. 
  - This may not reflect an overall power status but is predictable.
  - Essentially it is giving up on the autopilot providing any sort composite view of a complex power system.
1. SYS_STATUS battery values contain nothing.
   - We don't send anything useful in SYS_STATUS and we prevent people from making assumptions that it is possible to reflect the overall battery status from this.
1. SYS_STATUS battery values some sort of autopilot notion of the composite battery state.
   - This is OK if it can be done in such a way that the user can make sensible decisions.
   - I'm not sure an autopilot can do this properly for complex systems. Using lowest battery level and median battery level are both likely to give deceptive results. 

I don't know what is best but I think it is _important_ for the end user to be able to see battery indicators and set failsafes for every single battery. This information should be accessible from the main battery status. If there is a failsafe trigger in one of the batteries this should be reflected in the top level.

DonLakeFlyer:
FYI: After thinking longer on my ramblings I think I'm mostly wrong and @auturgy is correct. Specically on trying to come up with a single status item for the multi-case. I think there is still something smaller missing but let me think on it a bit more and I'll write something up.

message_definitions/v1.0/common.xml
```diff
@@ -3237,14 +3237,14 @@
...
       <field type="uint32_t" name="onboard_control_sensors_health" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors have an error (or are operational). Value of 0: error. Value of 1: healthy.</field>
       <field type="uint16_t" name="load" units="d%">Maximum usage in percent of the mainloop time. Values: [0-1000] - should always be below 1000</field>
-      <field type="uint16_t" name="voltage_battery" units="mV">Battery voltage</field>
-      <field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
-      <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy, -1: autopilot estimate the remaining battery</field>
+      <field type="uint16_t" name="voltage_battery" units="mV">Battery voltage, UINT16_MAX: Voltage not sent by autopilot</field>
+      <field type="int16_t" name="current_battery" units="cA">Battery current, -1: Current not sent by autopilot</field>
```
amilcarlucas: Why not use INT16_MIN when current not sent by autopilot here?
That way negative values (charging the battery) might still be usefull.

message_definitions/v1.0/common.xml
```diff
@@ -3237,14 +3237,14 @@
...
+      <field type="int8_t" name="battery_remaining" units="%">Battery energy remaining, -1: Battery remaining energy not sent by autopilot</field>
```
amilcarlucas: Again INT8_MIN might be a better value here.

message_definitions/v1.0/common.xml
```diff
@@ -4681,7 +4681,7 @@
       <field type="uint8_t" name="epv" units="dm">Maximum error vertical position since last message</field>
       <field type="int8_t" name="temperature_air" units="degC">Air temperature from airspeed sensor</field>
       <field type="int8_t" name="climb_rate" units="dm/s">Maximum climb rate magnitude since last message</field>
-      <field type="int8_t" name="battery" units="%">Battery (percentage, -1 for DNU)</field>
+      <field type="int8_t" name="battery" units="%">Battery (percentage, -1 for DNU). For multi-battery systems this should reflect the power status of the whole vehicle, sufficient for an external system to be able to make decisions about continuing to fly safely (and ideally to estimate remaining flight time)</field>
```
amilcarlucas: What is DNU ?? Is it do not use ? I prefer the full text

DonLakeFlyer:
My opinion would be:
* SYS_STATUS returns information from battery 0. Keep it simple.
* BATTERY_STATUS.id = Ids are sequential starting at 0. So if 3 batteries: 0, 1, 2. Never: 0,2,3, or 1,2,3.

message_definitions/v1.0/common.xml
DonLakeFlyer: We shouldn't be changing an existing detailed spec. Things are already coded to that standard.

hamishwillee: Agreed. Except for cases where the spec is unclear - specifically, does anyone know what DNU below is? https://github.com/mavlink/mavlink/pull/1182#discussion_r307674856

hamishwillee: PS @amilcarlucas FYI probably unclear, but I didn't change anything in terms of behaviour here. This was clarifications against what PX4 and ArduPilot already do.

message_definitions/v1.0/common.xml
hamishwillee: @auturgy Any idea?

hamishwillee:
@DonLakeFlyer https://github.com/mavlink/mavlink/pull/1182#issuecomment-515510972 works for me.

I assume you have some idea of how you'll present the multi-battery case to users.

message_definitions/v1.0/common.xml
WickedShell: `Do Not Use`, basically indicates we don't have a battery percentage available.

DonLakeFlyer:
> I assume you have some idea of how you'll present the multi-battery case to users.

Once this all gets locked down this pull will get updated to reflect the descisions: https://github.com/mavlink/qgroundcontrol/pull/7575

edwinhayes:
> SYS_STATUS returns information from battery 0. Keep it simple.

This would then be inconsistent with HIGH_LATENCY messages, which need to represent the aggregate battery status in as simple a form as possible.

> BATTERY_STATUS.id = Ids are sequential starting at 0. So if 3 batteries: 0, 1, 2. Never: 0,2,3, or 1,2,3.

Does that imply that if a system has a battery which isn't always present, the other batteries could be reporting with a different ID depending on whether or not the battery was present?

julianoes:
> But that said I think the ability for an end user to understand simply whether they are about to fall out of the sky is a requirement. To me that means a single indicator, whatever that may be: percentage indicator, color indicator, time indicator. To me it's a hard requirement that it exists and is a single thing. 

I agree with this notion. And I would argue that the default should be as safe as possible. This means that the information transmitted in SYS_STATUS should be what matters mostly regarding safety for a given aircraft configuration.

Now when it comes to the implementation in a flight stack I would treat it this way:
- 1 battery -> easy
- 2 identical batteries on both power modules -> select the lower one by default because it's the safest option
- 3 more involved setup -> configuration paramaters or more involved logic implemented by the integrator given the knowledge of the specific battery setup

This means you can get false positives (battery low for something which does not matter for flight) but that's what is expected if you have multiple batteries and you don't add the logic required for it. And it's better to have false positives than to miss a critical battery because it's not visible.

DonLakeFlyer:
My take on this is that the battery status stuff really needs a ground up rethink. Trying to shoe horn it into existing BATTERY_STATUS and SYS_STATUS isn't going to work. There is too much missing information in those messages to provide truly valuable feedback. Hence the first step is make what's there now work as well as it can be made to work. Then start over.

Some quick examples of missing information:
* What is the battery used for: Flight power or Auxiliary powered item. Something like this is required for a ground station to provide decent ui indicators.
* What should be the warnings level to the user (for each battery) in order to provide a simplified green/yellow/red style indication of status. Currently for example QGC has these limits hardwired on the GCS side. This is wrong.
* If there is an overall power status indicator which specific battery (ies?) is currently driving the indicator?
* May even require the concept of knowing whether for flight batteries a single battery can drive status. Or if it's the combination of the two which does. Or simpler: If one battery goes red you should, RTL. Or if only both batteries go red you should RTL. All depends on how they are wired up doesn't it?

hamishwillee:
@julianoes Even for just three batteries, I'd rather see them all in the GCS and configure their relationships that have the autopilot just use the lowest. Then I have the choice and can get rid of those false positives. 

We do have https://mavlink.io/en/messages/common.html#MAV_BATTERY_FUNCTION which tells us something about how important the battery is and what it does. The GCS could use this to infer the default importance of batteries.

As  @DonLakeFlyer says this needs more thought. 
Do we know anyone who has real experience with complex battery systems who can help? Are there other domains where they have to present complex battery systems to users that we can learn from?

auturgy:
@wickedshell has a reasonably complex setup.

julianoes:
I'm all for options to configure this to avoid false positives. My point is just that needs to be a safe default for a user that plugs in two batteries and doesn't provide any config. That would unblock the single indicator but leave everything open for future improvements.

DonLakeFlyer:
@anitheavy Todd, have you worked with systems with multiple batteries?

hamishwillee:
> I'm all for options to configure this to avoid false positives. My point is just that needs to be a safe default for a user that plugs in two batteries and doesn't provide any config. That would unblock the single indicator but leave everything open for future improvements.

For me that would be using BATTERY_STATUS and displaying both batteries, with their own failsafes.

hamishwillee:
Just had a chat to @WickedShell about what he does/his opinion. Basically he and I are in agreement. Yes, I do feel smug.

The summary of what his GUI does:
1. Uses BATTERY_STATUS (only) for all battery display. Each battery has its own set of failsafes etc.
   >... GCS can't give you sane percentages/warnings without a lot more side channel information about what the min/max/failsafes are for each battery
   > ... we already need to have a side channel that tells us how to display this to a user, which means that displaying this on a per source of failure should be easy

2. Uses the SYS_STATUS battery fields only for log review purposes when the link is really bad.
   - IF we use these fields then they should reflect an actual battery.
3. Dislikes the idea of using lowest battery in SYS_STATUS, because it really is only a useful default for a single setup - one where your multiple batteries all start up and drain at the same level. For any other setup they are very deceptive. 
4. UI does a one-time setup when vehicle is connected where the user decides which battery(ies) should appear in the top bar, and whether the icon is just an icon or reflects an associated battery status. Clicking the icon takes you to information on all batteries.

Note, I think we can do a bit more automated than this - i.e. GCS accumulates batteries on connection to a vehicle. If it sees 1 or 2 it displays them individually. If it sees more (or appearance of a new battery for first time) then it prompts user to decide which ones to display in top bar. Either way, the battery can be added/removed using its detail page, and can be accessed below the top level.

hamishwillee:
Status:
- Everyone is OK with battery ids being indexed sequentially from 0.
- No consensus on SYS_STATUS  or HIGH_LATENCY2 battery values, but the suggestion that they should reflect a single real battery appears much stronger. I have modified it that way. (Also quite a lot of support for idea that battery info should accumulate from BATTERY_STATUS or SMART_BATTERY_STATUS.

Lets discuss in devcall again. If we can't agree consensus I will just merge the battery id changes. Perhaps we leave handling of SYS_STATUS to individual systems (cop out).

hamishwillee:
@LorenzMeier @dagar After way too much discussion we've reached consensus for handling of multiple batteries:
- BATTERY_STATUS, SMART_BATTERY_STATUS ids sequentially increment from 0
- SYS_STATUS and HIGH_LATENCY2 battery information must reflect a single real battery, where the selected battery is dependent on the autopilot, but we expect will be set by a parameter or similar. 
  - The point is that this always reflects a single real system battery; it isn't an autopilot "summary" of battery state that can be used for failsafe triggering. 
  - the implication of this is that GCS will have to work out how to present multiple batteries and failsafes in a way that allows users to make sensible decisions.

Any last comments before merging? @auturgy @julianoes feel free to tweak the wording.

edwinhayes:
@hamishwillee Can you summarise the reasoning regarding the decision to have SYS_STATUS reflect a single real battery?  It seems like nonsense to me.

hamishwillee:
@edwinhayes There was a range of opinions, for which you'd have to read the thread. In essence though :
- no one thought that we could do a great job of reflecting an arbitrarily complex battery system  in SYS_STATUS battery fields - we can do better using BATTERY_STATUS.
- SYS_STATUS battery fields then are then always going to trade off between flight safety and underreporting power. For many setups, in particular those with the concept of a single primary battery, better to let the user decide which one it is.

I missed your comment on HIGH_LATENCY messages. Perhaps that case should be different - ie you'd be happy to, say, vastly under-report battery life? 

Or to put it another way, what do you think the rules should be?

>> BATTERY_STATUS.id = Ids are sequential starting at 0. So if 3 batteries: 0, 1, 2. Never: 0,2,3, or 1,2,3.

> Does that imply that if a system has a battery which isn't always present, the other batteries could be reporting with a different ID depending on whether or not the battery was present?

Bugger, that is a good question. I don't know :-( @auturgy @WickedShell @julianoes ?

My first thought is that with multiple batteries we need multiple fail-safes. Which implies that we need a way to link battery and failsafe levels. The obvious way would be for batteries to always have the same ID published, meaning that a removable battery would result in an empty "slot" in the ids. Not sure whether the empty battery should be published/indicated.

Back to the drawing board.

edwinhayes:
@hamishwillee - Yeah sorry; I've followed this thread, I just wondered if there were some compelling explanations for the proposed behaviour put forward in the devcall or some other channel.

Basically:

* I agree the primary mechanism for software to obtain an understanding of the power state of a system should be `BATTERY_STATUS`.
* Forcing `SYS_STATUS` to be a particular single battery would essentially render it useless on any aircraft with multiple batteries, since its value would be meaningless without also parsing all the `BATTERY_STATUS` messages, so in that case why have the field at all?
* There are cases where a single scalar representation of the aircraft's power state is useful, so may as well use the `SYS_STATUS` and `HIGH_LATENCY` values for that.
* Therefore, I think that the field in `SYS_STATUS` should be denoted as an implementation specific summary value.
* Sure, it may be difficult for open source flight controller implementations to guarantee they provide a fully representative summary value for all possible hobbyist configurations, but that shouldn't preclude people from trying.
* That way the mainline Ardupilot and Pixhawk builds can just report the battery one value and state their docs that is what they are doing.
* And commercial implementations can include whatever fancy machine learning aggregation algorithm they might want to fiddle around with.

hamishwillee:
@edwinhayes Thank you. Doing the best we can - including trying to pull in expertise. 

> Forcing SYS_STATUS to be a particular single battery would essentially render it useless on any aircraft with multiple batteries, since its value would be meaningless without also parsing all the BATTERY_STATUS messages, so in that case why have the field at all?

That was my personal opinion too. I wanted to mark it deprecated in the description and lead towards it not being used. Others see advantage in providing additional logging, or perhaps useful for systems with just one battery. Or one battery that is "main". We're a committee ....

> There are cases where a single scalar representation of the aircraft's power state is useful, so may as well use the SYS_STATUS and HIGH_LATENCY values for that.
> Therefore, I think that the field in SYS_STATUS should be denoted as an implementation specific summary value.

For SYS_STATUS values I strongly disagree - for reasons already given. If we can't give a reliable estimate for complex systems then we shouldn't pretend to.

HIGH_LATENCY is a bit different though - we don't have the alternative of sending individual battery info so it is kind of "deceptive or nothing".  My concern is safety - ie implementation specific seems OK until that implementation results in a crash or someone whinging that you landed with a spare battery 80% full.  For this case I'd have preferred the field to be just an enum - "OK", "Risky battery"

FMI Just how "latent" are messages?

edwinhayes:
> If we can't give a reliable estimate for complex systems then we shouldn't pretend to.

But that isn't for the MAVLink protocol to decide.  I **can** give a reliable estimate for my aircraft; this change would say I'm not _allowed_ to.

How about just providing a mechanism to denote that the estimate is no good?  Have NaN or -1 indicate that the a summary value isn't available.

edwinhayes:
(Sorry, that's not very clear given that's basically what the fields are marked as in the PR; I mean, why wouldn't that be acceptable in the case that the "these fields must be a real battery (which is autopilot dependent)" rule is just replaced with "these fields are autopilot dependent"?)

edwinhayes:
Also, I note that in the case of the `HIGH_LATENCY` message, having a good vs not-good enum seems like to could be pretty difficult in practice: the typical use of that message is for long distance BVLOS links, and for those types of operations it will be pretty difficult to easily identify what constitutes good/risky: it's not like just buzzing around at the park where you wait until the flight controller is screaming "battery critical" and then put it down wherever!

hamishwillee:
@edwinhayes The MAVLink protocol does allow you to  provide "invalid data" flag for the battery fields. And yes, IMO a multi battery system that can't provide that information should do so. 

So yes, we could say that a system that can provide reliable summary information about system overall charge level can do so or otherwise must put in NA for each of the fields. That is not a bad compromise; a GCS can then use it or not, knowing that (a) it can be trusted at least in theory (b) there is not necessarily a direct mapping to the underlying power system, so users may not be able to understand where it comes from when compared to the actual batteries they can see reported.

Not sure about the HIGH_LATENCY case. If you can't provide an OK vs Not OK then how is battery level going to be any use to anyone?

edwinhayes:
@hamishwillee - Regarding OK vs Not OK: yeah, it's tough.  We have to cater to "treat it like it's a Phantom4, upload the mission, don't bother reading it back, auto mode, arm and takeoff" customers.  But also to CPL pilots who _can_ compare the reported battery voltage against graphs of the planned mission profile and replan their route if the power consumption isn't going to leave sufficient reserve.  So in general, I need to both make things simpler, but still leave the capability for sophisticated users to do cool stuff.

julianoes:
> But that isn't for the MAVLink protocol to decide. I can give a reliable estimate for my aircraft; this change would say I'm not allowed to.

I agree and my point was to go for best (or safest) effort for anyone that doesn't have the logic for the reliable estimate.

hamishwillee:
Re battery status, we can see if the arguments change anyone's mind in the dev call. 

I am happy with any solution that a GCS can reliably trust. A compromise proposal could be _something_ like:

> The SYS_STATUS battery fields should either reliably report the overall vehicle power state or all values must be set as invalid, where reliable is taken to mean that critical level battery failsafe can be set and trusted by the user, and that the battery levels are not under-reported by more than 20%.

Note that personally I'd still rather just make all the fields deprecated for sys_status and use the above approach for high latency. I base that on KISS (keep it simple, stupid) rule.

hamishwillee:
OK, so any thoughts on sequential battery ids and vehicles that might have removable batteries ...

>>> BATTERY_STATUS.id = Ids are sequential starting at 0. So if 3 batteries: 0, 1, 2. Never: 0,2,3, or 1,2,3.

>> Does that imply that if a system has a battery which isn't always present, the other batteries could be reporting with a different ID depending on whether or not the battery was present?

> My first thought is that with multiple batteries we need multiple fail-safes. Which implies that we need a way to link battery and failsafe levels. The obvious way would be for batteries to always have the same ID published, meaning that a removable battery would result in an empty "slot" in the ids. Not sure whether the empty battery should be published/indicated.

edwinhayes:
I agree; I think for each aircraft 'configuration' the battery IDs should be static, with missing batteries having an empty slot.  That way it's feasible to have a table of failsafe values mapped to IDs.

DonLakeFlyer:
So the really important part of the battery ids is how they relate to the parameters which set up the power modules associated with the specific battery. In order to provide voltage/current multiplier calibration ui from a ground station it needs to know which voltage/current value to read from telemetry. So for example if a user is setting up Battery 2 from a gcs ui, which telemetry values correspond to that battery?

edwinhayes:
Not sure what it's like in PX4, but an AP there are individually named parameters BATT*, BATT2* ... BATT9*.  So whatever the IDs are, you have named parameters which are directly mapped to those IDs.
Then telemetry values are delivered inside the `BATTERY_STATUS` message though, so have the same ID as the battery to which they correspond?

ItsTimmy:
> So the really important part of the battery ids is how they relate to the parameters which set up the power modules associated with the specific battery. In order to provide voltage/current multiplier calibration ui from a ground station it needs to know which voltage/current value to read from telemetry. So for example if a user is setting up Battery 2 from a gcs ui, which telemetry values correspond to that battery?

For PX4, it will be done the same way as Ardupilot: Separate parameters for each battery. This is currently in PR PX4/Firmware#12551, and will hopefully be merged soon.

ItsTimmy:
I have a very minor concern about the IDs of batteries. Based on everything user-facing (parameter names, labels on autopilot hardware, etc.), the battery IDs start at 1. So I think it would be wise for the ID field to be 1-indexed everywhere, for consistency. It would get confusing if, for example, the battery with ID `1` is configured by `BATT2*` or `BAT2_*` parameters.

Other than that, I agree that the IDs should not need to be sequential, and that there should not necessarily be a pattern to the IDs, other than that they are constant for a given battery slot on a given vehicle.

DonLakeFlyer:
> So I think it would be wise for the ID field to be 1-indexed everywhere, for consistency.

The field is already implemented in existing firmwares as 0-based with single battery systems reporting a single battery on index 0. Ground stations are coded to this standard. It's too late to change that.

hamishwillee:
@edwinhayes @ItsTimmy . OK, so for the battery ids, I propose the following description. Note that it changes the id from "battery id" to "battery slot id" and keeps the need for sequential allocation (of slots-  not ids). Thoughts?

> Battery slot ID. Slot ids are indexed sequentially from zero, and must refer to the same slot for the lifetime of a particular vehicle. Note that battery parameter names associated with a particular battery slot are indexed from one).

edwinhayes:
@hamishwillee To confirm, battery ID is removed, and battery slot ID is added?  This seems sensible to me.  (Though I note that it would be proper to replace "Note that battery parameter names associated with..." with "Note that in most flight controller implementations, battery parameter names associated with...")

Unrelated, sigh, I just noticed that the array of voltages only allows packs to 10S. :-(

hamishwillee:
@edwinhayes  

> To confirm, battery ID is removed, and battery slot ID is added? This seems sensible to me. 

Yes. In single battery systems it is the same thing. But this makes clear that batteries themselves are potentially removable. 

> (Though I note that it would be proper to replace "Note that battery parameter names associated with..." with "Note that in most flight controller implementations, battery parameter names associated with...")

OK, so I really want to "hard code" this. If we allow flexibility then a GCS has to support multiple implementations.

> Unrelated, sigh, I just noticed that the array of voltages only allows packs to 10S. :-(

Yes. For smart battery message we added a cell offset so you could send multiple messages to get larger cell counts. 
I'd be open to a new message or extension that did the same thing for battery_status (though if we use an extension we'd have to accept that a system that has not updated might misinterpret offsets > 0.

edwinhayes:
@hamishwillee - Is there already existing precedent for the MAVLink specification ruling what the names of any underlying parameters must be?  For practical purposes, I wouldn't really care, though on principle my preference would be that if anything the specification insisted that the existing implementations renamed their parameters to be zero indexed!

> I'd be open to a new message or extension that did the same thing for battery_status

I'll defer dealing with this for now; we'll be looking at smart batteries later in the year.

hamishwillee:
> Is there already existing precedent for the MAVLink specification ruling what the names of any underlying parameters must be? For practical purposes, I wouldn't really care, though on principle my preference would be that if anything the specification insisted that the existing implementations renamed their parameters to be zero indexed!

No, and you could reasonably argue that while parameter protocol is within MAVLink scope, the naming of actual parameters is definitely not.

In practise though, the main users are PX4 and ArduPilot, so if we to all agree to do this then it would happen.

The obvious alternative is to have "yet another message set" such that a GCS can explicitly set/request the values against a battery id, and these get mapped to parameters. Seems overly complicated.

> my preference would be that if anything the specification insisted that the existing implementations renamed their parameters to be zero indexed!

Frankly, me too. The irritant here is breaks to firmware following such a rename.

hamishwillee:
@edwinhayes @DonLakeFlyer Et al, OK, so the battery ids were updated to:

> Battery slot ID. Slot ids are indexed sequentially from zero, and must refer to the same slot for the lifetime of a particular vehicle. Note that battery parameter names associated with a particular battery slot are typically indexed from one).

This essentially ensures that the battery id always refers to the same thing, which is a precondition for mapping a battery to its associated parameters. 

We discussed in dev call, and stopped short of "forcing" parameters to follow any sort of naming pattern (though we did note the normal index relationship). 

We would still like to have a way to make it easy for QGC to associate a battery and its information without having to know anything about the flight stack. Ideas welcome???

The obvious way is to have a message that QGC can use to get/set that information. We're just not sure if it is possible, and what information would be needed that could be common for ArduPilot, PX4, and still work with most other systems.

Off the top of my head we'd need
- id - to map to particular battery,
- failsafe levels % (at least 2 - warning and critical)
- failsafe actions for each level - with options none, notify/warn, hold, land, rtl, kill. 
  - Perhaps option to select an arbitrary mode?
  - Perhaps some enums for system specific options?
  - Not sure what we do for modes that are not supported on a system - NAK error for invalid settings?
- Full voltages per cell
- Empty voltage per cell
- Voltage divider
- Amps per volt

Where it gets complicated are the flight stack specific methods of doing various compensations - e.g. the params in http://docs.px4.io/master/en/config/battery.html#load_compensation

hamishwillee:
Devcall discussed concerns that only reporting a single real battery in SYS_STATUS explicitly prevents someone from reliably reporting an overall system status if they are able. Also the proposal of doing something like the [idea here](https://github.com/mavlink/mavlink/pull/1182#issuecomment-520660936) that battery fields should either reliably report the overall vehicle power state or all values must be set as invalid.

The takeaway points of discussion were that:
- The "provide trustable values or don't send anything model" might work if used properly. We think that most people will, if only because the vast majority of vehicles have just one main battery so they will just report tat battery
- We don't trust that it won't be abused by multi-battery users. 
- We can't say "don't use at all" because there are systems like OSD that are never going to update, and will lose useful information for the single battery case.
- ArduPilot are not going to change what they are doing (one battery set by user).
- We aren't going to say what should be done in HIGH_LATENCY2. Not enough users for it to matter.

Anyway, we decided to allow "provide trustable values or don't send anything model" wording. Generally though we hope GCS will move away from displaying/using this info and flight stacks from sending it. 

Thoughts?

edwinhayes:
@hamishwillee - I am interested to understand what the scenarios are which people are concerned about with regards to the system being "abused by multi-battery users"?  But to me, this seems good.

hamishwillee:
> I am interested to understand what the scenarios are which people are concerned about with regards to the system being "abused by multi-battery users"? But to me, this seems good.

The concern is that people are going to just export whatever values they like, without concern for the consumer relying on them. Essentially laziness.

edwinhayes:
> The concern is that people are going to just export whatever values they like, without concern for the consumer relying on them.

Sure, but pretty much all the examples I can think of fall into the category of "play stupid games, win stupid prizes"!

hamishwillee:
Actually, the bigger problem here is how the gcs should map a voltage or current to a level. that's not a real problem with a real battery because you have a voltage divider.

edwinhayes:
> to a level.

You mean to some kind of % measurement?

hamishwillee:
>> to a level.

> You mean to some kind of % measurement (of 'fullness')?

Yes. Without a way to map the voltage or current to a level/fullness they are pointless.

edwinhayes:
Is interpolating between some configured 'full' and 'empty' values not a valid solution for something that is not 'a real battery'?  Real batteries don't have linear voltage curves either.

hamishwillee:
> Is interpolating between some configured 'full' and 'empty' values not a valid solution for something that is not 'a real battery'? Real batteries don't have linear voltage curves either.

Exactly what are the full and empty battery values that you wish to interpolate against? 
This is why we wanted a real battery. If you say this is the battery with id=0 or any known battery then you know the flight stack has parameters to do that comparison. If it is an imaginary battery we'd have to require them to have non-imaginary parameters for that purpose. If we only allow imaginary batteries to output a level, then we remove that requirement. 

... and if it is really some sort of magic rendering of power state you'd expect the autopilot to be able to do that.

edwinhayes:
Is that even true though?  What parameters does ArduCopter have to support this?  There is `BATT_LOW_VOLT` which tells you the lower bound, but I don't think there is anything which captures the upper bound?

hamishwillee:
I _think_ the question is irrelevant. A GCS should be able to work out what the power state is from the message without having to know and do custom code for the stack. I don't see any way that this is possible for the imaginary battery case other than for the % charge field.

edwinhayes:
But I don't see how it's possible for _real_ batteries either right now, so saying you don't like how imaginary batteries are handled is also irrelevant.

edwinhayes:
Put it this way, I think that, unless you're only ever going to report % for _all_ batteries, there needs to be a mechanism by which the GCS can fetch the configuration values for a given battery metric.  I don't think that exists at present.

edwinhayes:
Something like a message `BATTERY_THRESHOLDS` which the GCS can request perhaps.

hamishwillee:
OK, so these are separate questions that really should not be joined. 

The first question is how a battery gets its associated parameters. YOu're right, that should be standardised so a GCS does not need to customise on a per-autopilot basis. See  https://github.com/mavlink/mavlink/pull/1182#issuecomment-523746942

hamishwillee:
Putting an imaginary battery in SYS_STATUS is a different thing. There are existing parameters in both ArduPilot and PX4 that handle real batteries based on the information provided. They map to parameters that are expected to exist for single battery.
There is no similar mapping for an additional imaginary battery, and you can't reuse the existing parameters for that imaginary battery - because then what will the real batteries do for their calculations?

If an autopilot is making up a composite power state then it can certainly output that as a level, and remove any additional work for the GCS to interpret what the voltage and current might mean.

edwinhayes:
Ah, yeah sorry, forgot about that.  Too many threads.  Yeah, exactly that, except I'd say that the ID field should also be able to cater to setting the thresholds for the composite state. (Aside - having battery ID=0 refer to the composite state could arguably have had some merit; it would address the flight controller parameters indexing from 1 for starters!)

hamishwillee:
Yes, having a message though is a chance for us to map indexes to batteries consistently. I hate the idea of the composite power state, but if there is generally strong support for using "0" for that then not my call.

auturgy:
fwiw ArduPilot does support “imaginary batteries” to a degree, via the “sum” battery type, which aggregates current and averages voltage. Intended for setups with esc telemetry but no dedicated battery sensor.

Regards,

James

edwinhayes:
> fwiw ArduPilot does support “imaginary batteries” to a degree, via the “sum” battery type, which aggregates current and averages voltage. Intended for setups with esc telemetry but no dedicated battery sensor.

From my point of view, this approach seems fairly practical for addressing the concerns about `SYS_STATUS`.

hamishwillee:
Cobblers. No. It. Is. Not. Because there is no mapping of any type to the SYS_STATUS battery.

edwinhayes:
Was it just that Ardupilot had a parameter to control the mapping?

hamishwillee:
It is probably  OK with me if it is very easy for a GCS to work out how to interpret the values without having to know anything about the flight stack. The composite battery type is represented as a single BATTERY_STATUS with a unique id that gets published.

So if this can be added to common.xml https://mavlink.io/en/messages/common.html#MAV_BATTERY_TYPE and we can revert the text to say something like "the SYS_STATUS battery fields must map to one battery output by SYS_STATUS ...

hamishwillee:
Closing this. The bits that were non controversial were captured. Everything else too opinionated to progress.

PR #1183, opened 2019-07-20
common.xml CAMERA_INFORMATION

robinhilliard:
Set display="bitmask" for flags field.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4845,7 +4845,7 @@
       <field type="uint16_t" name="resolution_h" units="pix">Horizontal image resolution</field>
       <field type="uint16_t" name="resolution_v" units="pix">Vertical image resolution</field>
       <field type="uint8_t" name="lens_id">Reserved for a lens ID</field>
-      <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS">Bitmap of camera capability flags.</field>
+      <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration)</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available).</field>
     </message>


hamishwillee:
The `display=` tells a GCS _editor_ that the value should be displayed as checkboxes. Obviously if the value is not being edited, then this isn't needed. I suspect that this is the case for this message.

@dogmaphobic Can you comment on whether this attribute is needed for this message?

julianoes:
I think that makes sense. Thanks.

robinhilliard:
Hi Hamish, 

Just to explain, I’m writing an Elixir Mavlink library and to make messages more Elixir idiomatic I’ve been translating enums into ‘atoms’ - these also enable static type checking of code using the library. I was using display=“bitmask” to do this mapping but found that in ardupilotmega and its includes there were only two messages where a field that was obviously a bitmask did not have this set (this one and ekf status flags). In a way I am using it to change the way the bitmask field is presented to the developer using the Elixir library. I have a work around where I detect enum values increasing in powers of two and the generator prints a warning saying that it’s assuming they’re bitmasks, but it’s a little hacky. Anyway hope that provides some context for your consideration.

hamishwillee:
@robinhilliard Thanks for the explanation. I don't think there is any reason we can't change this, but I'm concerned that your implementation will just get broken next time someone adds a bitmask. 

We can perhaps reduce the risk by suggesting that all bitmasks use this tag (but we can't easily enforce that via automated validation). 

@DonLakeFlyer Have you any problem with requiring `display="bitmask"` is used on messages that are unlikely to ever be displayed?

amilcarlucas:
LGTM

DonLakeFlyer:
I'm not really up on all these tags. I assume this is a new tag? If so, doesn't seem like a problem.

hamishwillee:
@DonLakeFlyer No `display="bitmask"` is an old flag, which I assumed you had introduced. AFAIK this one just lets QGC know that it can render the bitmask as a set of checkboxes. 
I'm pretty sure it is OK to set this on any bitmask field.

DonLakeFlyer:
If it's supposed to work like the "bitmask" metdata in parameters then it should also include the values like this (using param meta data example): ```<field name="Bitmask">0:Roll,1:Pitch,2:Yaw</field>```. Otherwise you have no idea what the bits mean.

amilcarlucas:
@DonLakeFlyer No, it does not need that.

It is just a hint to let GCSs know that the data can be intrepreted as a bitmask.

All fields that have a 'display="bitmask"' attribut must also provide an 'enum=' attribute. That needs to be enforced by the schema at some point.

But otherwise this PR is sane.

DonLakeFlyer:
> All fields that have a 'display="bitmask"' attribut must also provide an 'enum=' attribute.

Ah, so that accomplishes the same thing. So this is fine

PR #1184, opened 2019-07-22
Reserving component IDs for private components

hamishwillee:
@edwinhayes @auturgy This replaces #1077 with an actual range of components in both common.xml and minimal.xml. Using individual IDs ensures that the range can't be used by accident in the same library build as common.xml or minimal.xml.

I'm OK to merge this. Any final thoughts?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -377,6 +377,232 @@
       <entry value="1" name="MAV_COMP_ID_AUTOPILOT1">
         <description>System flight controller component ("autopilot"). Only one autopilot is expected in a particular system.</description>
       </entry>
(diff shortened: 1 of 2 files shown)


auturgy:
I’m guessing the intent is to revert this from common when minimal is added as an include?

Regards,

James

hamishwillee:
@auturgy Yes. I'm trying to keep them in sync for now, and hoping Peter will find time to merge the fixes to file inclusion sooner rather than later.

hamishwillee:
Merged, as discussed/approved in dev call yesterday. 

@edwinhayes, it has been a long road. Thanks for your patience!

edwinhayes:
Awesome, thanks @hamishwillee

PR #1186, opened 2019-07-26
[WIP] add a message for component status reports

baumanta:
we would like to add a mavlink message for hardware status of system components. Mainly this would concern the companion computer. This message would enable monitoring the cpu usage, temperature etc. on the system. On the ground station side, such a message would allow to print warnings if the companion reaches a critical state. On the autopilot side the corresponding failsafe actions could be implemented. The message was named quite generically on purpose, such that it could also be used for other components, e.g. for a camera to report the storage situation or similar.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5702,11 +5702,13 @@
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
       <field type="uint32_t" name="uptime" units="ms">Time since system boot.</field>
       <field type="uint8_t" name="type">Type of the onboard computer: 0: Mission computer primary, 1: Mission computer backup 1, 2: Mission computer backup 2, 3: Compute node, 4-5: Compute spares, 6-9: Payload computers.</field>
-      <field type="uint8_t[8]" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
-      <field type="uint8_t[2]" name="gpu_usage">GPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="uint8_t[8]" name="cpu_cores">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="uint8_t[10]" name="cpu_combined">Combined CPU usage as the last 10 slices of 100 MS (a histogram). This allows to identify spikes in load that max out the system, but only for a short amount of time. A value of UINT8_MAX implies the field is unused.</field>
+      <field type="uint8_t[4]" name="gpu_cores">GPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="uint8_t[10]" name="gpu_combined">Combined GPU usage as the last 10 slices of 100 MS (a histogram). This allows to identify spikes in load that max out the system, but only for a short amount of time. A value of UINT8_MAX implies the field is unused.</field>
       <field type="int8_t" name="temperature_board" units="degC">Temperature of the board. A value of INT8_MAX implies the field is unused.</field>
       <field type="int8_t[8]" name="temperature_core" units="degC">Temperature of the CPU core. A value of INT8_MAX implies the field is unused.</field>
-      <field type="int16_t[4]" name="fan_speed" units="RPM">Fan speeds. A value of INT16_MAX implies the field is unused.</field>
+      <field type="int16_t[4]" name="fan_speed" units="rpm">Fan speeds. A value of INT16_MAX implies the field is unused.</field>
       <field type="uint32_t" name="ram_usage" units="MiB">Amount of used RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t" name="ram_total" units="MiB">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t[4]" name="storage_type">Storage type: 0: HDD, 1: SSD, 2: EMMC, 3: SD card (non-removable), 4: SD card (removable). A value of UINT32_MAX implies the field is unused.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5214,19 +5214,20 @@
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
-    <message id="390" name="COMPONENT_STATUS">
+    <message id="390" name="COMPANION_SYSTEM_STATUS">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Hardware status sent by a component</description>
       <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
       <field type="int8_t" name="temperature" units="degC">Temperature. A value of INT8_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="ram_free" units="MiB">Amount of free RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="ram_usage" units="MiB">Amount of used RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t" name="ram_total" units="MiB">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t" name="storage_free" units="MiB">Amount of free storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t" name="storage_total" units="MiB">Total amount of storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t" name="network_send_rate" units="KiB/s">Network traffic from the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t" name="network_receive_rate" units="KiB/s">Network traffic to the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint32_t" name="time_usec" units="us">Message timestamp in sender timebase.</field>
+      <field type="uint32_t" name="uptime" units="ms">Time since system boot.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5218,14 +5218,14 @@
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Hardware status sent by a component</description>
-      <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle)</field>
-      <field type="int8_t" name="temperature" units="celsius">Temperature on the component</field>
-      <field type="uint32_t" name="memory_free" units="megabytes">Amount of free memory on the component system</field>
-      <field type="uint32_t" name="memory_total" units="megabytes">Total amount of memory on the component system</field>
-      <field type="uint32_t" name="storage_free" units="megabytes">Amount of free storage space on the component system</field>
-      <field type="uint32_t" name="storage_total" units="megabytes">Total amount of storage space on the component system</field>
-      <field type="uint32_t" name="network_send" units="kilobytes/s">Network traffic from the component system</field>
-      <field type="uint32_t" name="network_receive" units="kilobytes/s">Network traffic to the component system</field>
+      <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="int8_t" name="temperature" units="celsius">Temperature. A value of INT8_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="RAM_free" units="megabytes">Amount of free RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="RAM_total" units="megabytes">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_free" units="megabytes">Amount of free storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_total" units="megabytes">Total amount of storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="network_send_rate" units="kilobytes/s">Network traffic from the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="network_receive_rate" units="kilobytes/s">Network traffic to the component system. A value of UINT32_MAX implies the field is unused.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5219,13 +5219,13 @@
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Hardware status sent by a component</description>
       <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
-      <field type="int8_t" name="temperature" units="celsius">Temperature. A value of INT8_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="RAM_free" units="megabytes">Amount of free RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="RAM_total" units="megabytes">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="storage_free" units="megabytes">Amount of free storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="storage_total" units="megabytes">Total amount of storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="network_send_rate" units="kilobytes/s">Network traffic from the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t" name="network_receive_rate" units="kilobytes/s">Network traffic to the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="int8_t" name="temperature" units="degC">Temperature. A value of INT8_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="ram_free" units="MiB">Amount of free RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="ram_total" units="MiB">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_free" units="MiB">Amount of free storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_total" units="MiB">Total amount of storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="network_send_rate" units="KiB/s">Network traffic from the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="network_receive_rate" units="KiB/s">Network traffic to the component system. A value of UINT32_MAX implies the field is unused.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5695,20 +5695,28 @@
       <field type="uint16_t" name="payload_type" enum="MAV_TUNNEL_PAYLOAD_TYPE">A code that identifies the content of the payload (0 for unknown, which is the default). If this code is less than 32768, it is a 'registered' payload type and the corresponding code should be added to the MAV_TUNNEL_PAYLOAD_TYPE enum, and the entry possibly to https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. Software creators can register blocks of types as needed. Codes greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
       <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
     </message>
-    <message id="390" name="COMPANION_SYSTEM_STATUS">
+    <message id="390" name="ONBOARD_COMPUTER_STATUS">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
-      <description>Hardware status sent by a component</description>
+      <description>Hardware status sent by an onboard computer.</description>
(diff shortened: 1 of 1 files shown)


baumanta:
fyi @julianoes , @hamishwillee , @jkflying

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5214,19 @@
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
+    <message id="390" name="COMPONENT_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Hardware status sent by a component</description>
```
hamishwillee: Are all fields mandatory? If not, then document a mechanism for sender to indicate "not supported" on each field (for uints this is ideally 0, but more normally the max int value.

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5214,19 @@
...
+      <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle)</field>
+      <field type="int8_t" name="temperature" units="celsius">Temperature on the component</field>
```
hamishwillee: How will you indicate not supported? Description would probably be "Temperature" (but if you want to spell it out, "use "of", not "on)

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5214,19 @@
...
+      <field type="uint32_t" name="memory_free" units="megabytes">Amount of free memory on the component system</field>
```
hamishwillee: I presume you mean program memory as opposed to storage memory? Is there a good way of saying that - can we still say "RAM?"

message_definitions/v1.0/common.xml
hamishwillee: For all fields, what should the recipient do with this/how should the information be used?

Where I am going with this is that in some cases you want the value of a field because the recipient needs to actually know the field, and in others you're actually interested in a warning status, and the value doesn't give you that. So this question is about "is this really the information you need"? 

I.e. An autopilot probably wants to know the camera available storage space (and gets that through the camera APIs). However at this level/without knowing the detail of the camera it won't know if 1 MB or 30Mb is a risky level. You could be asking because you want a general mechanism for finding storage, or you could be asking because actually you want a field that generally says "hey, my component is about to fail due to out of storage", or you might want both. If you want a warning though, then the approach is wrong.

hamishwillee:
@baumanta Thanks. I think it is a good idea to start off making it a generic message (we can review if it makes sense).
In theory it is obviously very cool to know the status of all our components. 
I'm personally most interested in what fields you need from a companion component and what you plan on doing with them - i.e. what will you do in the autopilot if you detect high or low network traffic from a companion computer, and how you will configure what high and low actually mean. Can an autopilot know enough about a component "generically" to do anything useful, or would it be better off getting warning flags - on the assumption that the companion knows it is in trouble? Or if the companion is in trouble, are things already too late ... blahdy blah blah :-)

@olliew42 What of these could/would a gimbal reasonably send? Is there any other general information that it could send the might be useful for "health" that the autopilot isn't already sending?

@auturgy FYI, any thoughts on this. Have ArduPilot got any messages or special ways of monitoring companion or component health?

message_definitions/v1.0/common.xml
baumanta: No, I think it makes sense to only use the fields applicable to the component sending the message. I don't know what makes more sense to send 0 or max int for disabled fields? I'll add that to the message description

hamishwillee: YOu can only use 0 if that is an "invalid value" for the type of field (ie in most cases here 0 would be a reasonable value). 

But if it can be 0, that is the preference (because then we may be able to truncate the empty fields in cases where they are not used, thereby reducing traffic).

message_definitions/v1.0/common.xml
julianoes: RAM is clearer than memory.

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5214,19 @@
...
+      <field type="uint32_t" name="memory_total" units="megabytes">Total amount of memory on the component system</field>
+      <field type="uint32_t" name="storage_free" units="megabytes">Amount of free storage space on the component system</field>
+      <field type="uint32_t" name="storage_total" units="megabytes">Total amount of storage space on the component system</field>
+      <field type="uint32_t" name="network_send" units="kilobytes/s">Network traffic from the component system</field>
```
julianoes: `send` and `receive` is not clear. What about `network_send_rate`?

message_definitions/v1.0/common.xml
```diff
@@ -5218,14 +5218,14 @@
...
-      <field type="uint32_t" name="memory_total" units="megabytes">Total amount of memory on the component system</field>
-      <field type="uint32_t" name="storage_free" units="megabytes">Amount of free storage space on the component system</field>
-      <field type="uint32_t" name="storage_total" units="megabytes">Total amount of storage space on the component system</field>
-      <field type="uint32_t" name="network_send" units="kilobytes/s">Network traffic from the component system</field>
-      <field type="uint32_t" name="network_receive" units="kilobytes/s">Network traffic to the component system</field>
+      <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="int8_t" name="temperature" units="celsius">Temperature. A value of INT8_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="RAM_free" units="megabytes">Amount of free RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
```
julianoes: All underscore, I think. so `ram_free`.

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5214,19 @@
...
     </message>
+    <message id="390" name="COMPONENT_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Hardware status sent by a component</description>
+      <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="int8_t" name="temperature" units="celsius">Temperature. A value of INT8_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="RAM_free" units="megabytes">Amount of free RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="RAM_total" units="megabytes">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_free" units="megabytes">Amount of free storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
```
julianoes: `megabytes` is not accurate enough. I suggest to use `MiB` like in https://mavlink.io/en/messages/common.html#STORAGE_INFORMATION.

hamishwillee: Actually, you **have to** use units from the supported set: https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L66

julianoes: Oh, I didn't know. And `MiB` are there, so that's good.

baumanta:
@julianoes with `time_boot_ms` do you mean for how long it has been running or the time it took it to boot? Would `uptime_ms` be more clear?

julianoes:
> @julianoes with time_boot_ms do you mean for how long it has been running or the time it took it to boot? Would uptime_ms be more clear?

`time_boot_ms` is just added to most of these messages by default so I think it makes sense to add it here as well.

On the other hand, uptime is a good addition, I'd like that.

baumanta:
ok, I can add both of them, I was just wondering what `boot_time_ms` means. What would you put in there?

baumanta:
I just looked at other message definitions. I see that this field exists in many of them, so I guess we stick with the naming. It is described as `Timestamp (time since system boot)` so I guess uptime becomes obsolete

julianoes:
> so I guess uptime becomes obsolete

I guess you're right.

hamishwillee:
@baumanta FWIW and FYI only I quite like the naming of uptime :-)

But
- many of these timestamps serve dual use - ie it is up to the autopilot to decide if they are time since boot or linux epoch time, or whatever.
- Changing messages is a pain because it breaks the receiver CRC to change the name of a field
- Consistency across all messages makes things easier if you're more than just an occasional user.

message_definitions/v1.0/common.xml
baumanta: I think this message is important of course to give warnings for the most critical values, e.g see directly in you groundstation if cpu usage goes above 80%, get a warning if ram is less than 10% etc. On the autopilot side, this would enable us to take failsafes as well, e.g. if avoidance is running and companion cpu load is above 80% go to hold. But for other stuff such as network traffic, I think logging is the most important point. If anything does not work as expected, we can go back to the log and we will be able to look at those states and see if they contain reasonable values. Right now, we have no insight on what is happening on the companion computer at all. So if we have a bad flight it is always guessing what could have caused delays or drops. I hope this message would give us a tool for debugging as well

baumanta:
I think it might make sense to add the field `type`. The  MAV_TYPE enum already contains onboard controller, gimbal, camera. So we could identify the sending object over the MAV_TYPE instead of the component id. Would that make sense?

julianoes:
`MAV_TYPE` should be transmitted via the heartbeat message as far as I know.

hamishwillee:
> I think it might make sense to add the field type. The MAV_TYPE enum already contains onboard controller, gimbal, camera. So we could identify the sending object over the MAV_TYPE instead of the component id. Would that make sense?

FWIW I don't think we should just add fields to identify sources when we already have mechanisms for that. The recognised and expected mechanism for routing remains the component id. 

The use of the `HEARTBEAT.type` to identify a component type is "proposed" but not accepted - currently the type is the system type, even for a component. But assuming it changes, the type will be sent in the heartbeat already - should not just get added to another message. 

The field is primarily intended for UI layout in a GCS. I guess you can use it to decide how to handle messages from the system if you like. If you need to handle message based on a specific instance (e.g. "camera 3" ) you would still use component ID.

baumanta:
Ok I see so lets stick to component ID. Mavros sends out all messages with component id MAV_COMP_ID_UART_BRIDGE (241) by default. We could either treat that as equal to the companion computer and assume that no other component sends the COMPONENT_STATUS over Mavros. Then we could leave both PRs as they are. 

Or if there are other components talking ROS (e.g. a camera, which would also like to report a COMPONENT_STATUS over ROS -> mavros -> mavlink) then we would have to add a component ID field to the mavros message and make mavros output the COMPONENT_STATUS message with the respective component id such that we can distinguish them.

Do you think option 2 is necessary?

julianoes:
So component IDs can collide if you integrate a system and it's up to the system integrator to de-conflict them. Therefore, mavros should allow that you can specify the component ID that is using. Without that option it would mean that you can only ever run one mavros instance in a "mavlink network", otherwise there are conflicts.

baumanta:
ok I fixed that in mavros. Now the mavros message contains a component id, such that we can distinguish between ROS messages from different components and forward them with the correct component ID

olliw42:
> @olliew42 What of these could/would a gimbal reasonably send? Is there any other general information that it could send the might be useful for "health" that the autopilot isn't already sending?

@hamishwillee  thx for asking :)  (it didn't saw it earlier since my name was spelled incorrectly :))

I think that this message touches a sensible point which probably could need broader discussion. 

My answer to your actual questions might make it clear:

> What of these could/would a gimbal reasonably send? 

None. (except of timestamp)

> Is there any other general information that it could send the might be useful for "health" that the autopilot isn't already sending?

Sure, plenty ... but they would be specific to that it's a gimbal. And I think that if one would ask the very same questions with just having "gimbal" replaced by anything else the answers would nearly always be the same.

This make me actually to not like that this message carries "COMPONENT" and "STATUS" in it's name. And it makes me think that - as a general concept - it probably would be more reasonable to have status messages per MAV_TYPE. 

I mean, the fields are - obviously (?!) - very much related to what is needed for the companion, but it's really far from obvious to me what the use of network_xxx or cpu_usage for a servo component should be ... "COMPONENT" in the name gives the message a generality which it never will reasonably fulfill. Also "STATUS" is a bit problematic, since so many info could be called "status". 

So, I guess the proposal I would derive from this proposal here is to call this message ONBOARD_CONTROLLER_XXX (expecting it's MAV_TYPE is MAV_TYPE_ONBOARD_CONTROLLER), and to be prepared or set the stages for GIMBAL_XXX, CAMERA_XXX, and so on, messages, which each can be designed according to what is most reasonable for the respective type.

Note that this kind of would be an extension of what is there e.g. for camera, just put on a common conceptually consistent ground (which might need some few modifications to fully realize).

Using "STATUS" for the "XXX" might not be convenient since there could be potential conflict with existing "STATUS" messages, for which one might not want to create a MAV_TYPE. I unfortunately have no good suggestion for an alternative - "STATE" might confuse - "CONDITION" ???.

Assuming "XXX" = "CONDITION" for the moment, I guess I would rename the message to ONBOARD_CONTROLLER_CONDITION

IMHO. :)

hamishwillee:
Thanks @olliw42 (sorry about previous typo!). 

@baumanta So Olli's point https://github.com/mavlink/mavlink/pull/1186#issuecomment-517711844 that none of this information is useful for gimbal is indicative that making this a general status info message is not a good idea. 

Quickly thinking about other possible components - servos, optical flow devices, osds, flarm, none of them would find this particular set useful. So my leaning would be to make it particular for a companion computer, or some other name for general computing platform.

message_definitions/v1.0/common.xml
hamishwillee: @baumanta Thanks! Sounds reasonable for a companion, where you know what sort of behaviour you consider OK :-) Not sure the same logic would be quite so clear for the camera or other components.

baumanta:
@hamishwillee @olliw42 @julianoes @jkflying

I just want to quickly summarize the proposed changes to be sure I got everyone right. So we will make this message specific to the companion computer instead of a generic message because no other component would benefit from it. Maybe call it COMPANION_SYSTEM_STATUS to be compliant with COMPANION_PROCESS_STATUS? 

@mhkabir made a point that `time_boot_ms` is used aa a timestamp in other messages. Should we rename to `uptime_ms` after all?

@Jaeyoung-Lim would like to have `ram_usage` and `ram_total` instead of  `ram_free` and `ram_total`

Did I get this right? Any objections?

mhkabir:
To be clear - uptime isn't redundant, since a `time_usec` timestamp from a companion isn't necessarily time from boot. For all current messages from mavros for example, it is the time since Linux epoch.

hamishwillee:
> Add an uptime_ms field for monitoring how long the system has been up.

Though note you'd just call it `uptime`. Except for legacy fields like time_usec  we don't put units in names anymore. 

Otherwise makes sense to me

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5214,21 @@
...
+      <field type="uint32_t" name="ram_usage" units="MiB">Amount of used RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="ram_total" units="MiB">Total amount of RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_free" units="MiB">Amount of free storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="storage_total" units="MiB">Total amount of storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="network_send_rate" units="KiB/s">Network traffic from the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="network_receive_rate" units="KiB/s">Network traffic to the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="time_usec" units="us">Message timestamp in sender timebase.</field>
```
mhkabir: This should be a uint64_t

olliw42:
pending the outcome of this discussion https://github.com/mavlink/mavlink/pull/1176#discussion_r311301146 it probably also should be changed to
 
`<field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>`

message_definitions/v1.0/common.xml
julianoes: And probably the first one as usual.

LorenzMeier:
Needs to be squashed once all checks pass.

message_definitions/v1.0/common.xml
baumanta: should uptime then be a uint64_t as well?

message_definitions/v1.0/common.xml
```diff
@@ -5695,6 +5695,21 @@
       <field type="uint16_t" name="payload_type" enum="MAV_TUNNEL_PAYLOAD_TYPE">A code that identifies the content of the payload (0 for unknown, which is the default). If this code is less than 32768, it is a 'registered' payload type and the corresponding code should be added to the MAV_TUNNEL_PAYLOAD_TYPE enum, and the entry possibly to https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. Software creators can register blocks of types as needed. Codes greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
       <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
     </message>
+    <message id="390" name="COMPANION_SYSTEM_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Hardware status sent by a component</description>
```
julianoes: The description says "hardware status sent by a component" but this is about a companion, not any component, and it's also the software state so "system status" would probably be better (the same as the name implies).

message_definitions/v1.0/common.xml
```diff
@@ -5695,6 +5695,21 @@
...
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint8_t" name="cpu_usage">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
```
julianoes: Should we also have a field for GPU usage? That's potentially a thing.

message_definitions/v1.0/common.xml
```diff
@@ -5695,6 +5695,21 @@
...
+      <field type="int8_t" name="temperature" units="degC">Temperature. A value of INT8_MAX implies the field is unused.</field>
+      <field type="uint32_t" name="ram_usage" units="MiB">Amount of used RAM on the component system. A value of UINT32_MAX implies the field is unused.</field>
```
julianoes: I don't love the fact that RAM has usage and storage has free. That's not consistent.

LorenzMeier: I've resolved that.

message_definitions/v1.0/common.xml
LorenzMeier: I've resolved that.

message_definitions/v1.0/common.xml
LorenzMeier: I've fixed it accordingly.

LorenzMeier:
@baumanta @julianoes I've polished the message and changed it quite significantly. A couple of notes why:

- It is a low frequency message, so we don't need to save the last byte
- You need to know for every CPU core what the load is, otherwise you're never finding out that your non-multithreaded app is maxing core 1 while core 2 is idle and your CV pipeline executes slow, despite having "only" 50% CPU load
- Same for temperatures - most mainboards read them at multiple locations and you need to know the different locations.
- You have for sure more than one network link
- You have almost certainly more than one harddrive
- You will have more than one onboard computer in a number of systems, so we need to know what we're looking at

olliw42:
is `uptime` now officially the new `time_boot_ms`?

(https://github.com/mavlink/mavlink/pull/1176#discussion_r311381749)

message_definitions/v1.0/common.xml
```diff
@@ -5695,6 +5695,31 @@
...
       <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
     </message>
+    <message id="390" name="ONBOARD_COMPUTER_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Hardware status sent by an onboard computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint32_t" name="uptime" units="ms">Time since system boot.</field>
+      <field type="uint8_t" name="type">Type of the onboard computer: 0: Mission computer primary, 1: Mission computer backup 1, 2: Mission computer backup 2, 3: Compute node, 4-5: Compute spares, 6-9: Payload computers.</field>
```
hamishwillee: @LorenzMeier Would this be better as either an enum, or enum for type and id for whether primary or some backup?

Either
```
      <field type="uint8_t" name="type" enum="MAV_CONTROLLER">Type of the onboard computer.</field>
      <field type="uint8_t" name="id"">Id. Each computer  of a particular type (e.g. MAV_CONTROLLER_MISSION) is numbered sequentially from 0. Where appropriate, the 0 id indicates a primary controller (higher ids indicate backups).</field>

    <enum name="MAV_CONTROLLER">
      <description>Onboard computer type/purpose.</description>
      <entry value="0" name="MAV_CONTROLLER_MISSION">
        <description>Mission computer.</description>
      </entry>
      <entry value="1" name="MAV_CONTROLLER_COMPUTE_NODE">
        <description>Compute node.</description>
      </entry>
      <entry value="2" name="MAV_CONTROLLER_PAYLOAD">
        <description>Payload computer.</description>
      </entry>
    </enum>
```

OR

```
    <enum name="MAV_CONTROLLER">
      <description>Onboard computer type/purpose.</description>
      <entry value="0" name="MAV_CONTROLLER_MISSION_PRIMARY">
        <description>Primary mission computer.</description>
      </entry>
      <entry value="1" name="MAV_CONTROLLER_MISSION_BACKUP1">
        <description>Mission computer backup 1.</description>
      </entry>
      ....
    </enum>
```

hamishwillee:
Can/should we provide direction on how often this should be emitted?

hamishwillee:
@baumanta @julianoes @LorenzMeier How do you feel about my comments above https://github.com/mavlink/mavlink/pull/1186#discussion_r312764782 regarding using enum/id for controller identity.

Also @baumanta what rate do you plan on sending this? Can we provide guidance for others?

baumanta:
@hamishwillee I would not send it at a very high rate. Maybe around 0.2 - 0.5Hz

message_definitions/v1.0/common.xml
julianoes: I think this type and ID should come out of `MAV_TYPE`, For multiple, there is already a component ID in the message anyway in case there is more than one.

hamishwillee: Good point. Though you would also need to add the instances to the MAV_COMPONENT right?

julianoes: If you need more than one, yes.

hamishwillee: Works for me, but I don't know if @LorenzMeier is attached to his design or would agree this is better.

LorenzMeier: I wouldn't do it as ENUM until we've tested this message in the field. I did on purpose stay away from it since we really don't know what we are doing.

LorenzMeier: To be clear: We need 4-5 new ENUMs for all the fields and I think it is not fair to introduce the massive churn in the spec (and others starting to use those) before the values have settled.

baumanta:
I think most comments have been addressed. I anybody opposed to merging this?
@julianoes @LorenzMeier @hamishwillee

message_definitions/v1.0/common.xml
hamishwillee: That is up to you. Action on @baumanta to decide what is most useful for the recipient :-)

Note that if you want to make this change, you can still so after merging as this is WIP.

message_definitions/v1.0/common.xml
hamishwillee: @LorenzMeier Thanks for explanation. 

@baumanta When you're happy this is working we will eventually need to remove the WIP. Action on you to, at that point confirm the values so we can do final iteration on this.

hamishwillee:
I am merging as this is WIP and it makes sense to iterate in the "real world" before finalising.

edwinhayes:
@hamishwillee Now this has been merged as WIP; what is the correct way to continue discussion about this merge?

I object to the use of the term "component system" in the descriptions: the words Component and System have specific meaning in the MAVLink spec, so this phrase doesn't make sense.

hamishwillee:
@edwinhayes The best way is to create a PR which does what you want, and invite others to discuss.

Issue #1152, opened 2019-06-03
Pl confirm/clarify heartbeat type for components


olliw42:
I've pointed to this in the px4 discuss forum (https://discuss.px4.io/t/mavlink-support-for-the-storm32-gimbal-controller-how-to-make-it-compatible/10723) but it seems it hasn't much frequency, so pl allow me to put this up here:

There is a MAV_TYPE_GIMBAL value in the heartbeat enum (https://mavlink.io/en/messages/common.html#MAV_TYPE_GIMBAL), and it appears obvious (at least to me) that for a gimbal this should be used. However, the docu for the heartbeat type field says that 

>type: Type of the system (quadrotor, helicopter, etc.). Components use the same type as their associated system.

I must admit I fail to understand the rational, and would also see real issues associated to this. So:

1) I wonder if that is indeed correct and if it's not "just" the wording which is misleading, and thus would like to ask to check this.

2) If this is indeed how it's supposed to be, I would find it useful if there could be few statements in the docu explaining the rational/intention/purpose, e.g. in the heartbeat microservice section.

Many thx, Olli

hamishwillee:
Hi @olliw42 

So `MAV_TYPE` is the *system* type, where a system is made up of components. If the gimbal was a complete "standalone" system, somehow separate of an autopilot (ie not addressed via the autopilot), it would have the type [MAV_TYPE_GIMBAL](https://mavlink.io/en/messages/common.html#MAV_TYPE_GIMBAL).

However if the gimbal is a component of a system (as I believe is the case) then system MAV_TYPE is that of the bigger system - ie autopilot, and the component is derived from the component ID in the mavlink package - ie it has id [MAV_COMP_ID_GIMBAL](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_GIMBAL).

Does that make sense?

And yes, that should absolutely be captured in https://mavlink.io/en/services/heartbeat.html , but I will await your response.

olliw42:
thx, sir

this is how I understand what is written in the docs

I can't say though that it makes sense to me. A steering wheel doesn't become a car because it is mounted in a car, it's still steering wheel.

Also, I can't see any reason for why one would want a gimbal to report itself as of type PX4 when it is connected to a PX4. I do see though issues with requiring this. So, if what you describe is indeed what is desired I would find it helpful to better understand the reason for doing so, e.g. what advantage it brings or what practical problem it whelps to solve.

hamishwillee:
[...]

PR #1188, opened 2019-08-01
Clarify MAV_TYPE usage

hamishwillee:
This clarifies that `MAV_TYPE` (`HEARTBEAT.type`) is the vehicle type for an FC and the component type for other components. 

This makes it easier for components, as they don't have to know at firmware build time what system they are used in. This will have no impact on QGC, and "should" not adversely impact other GCS.

Discussion here: https://github.com/mavlink/mavlink/issues/1152

Assuming this goes in, there is follow on work in the docs for [heartbeat protocol](https://mavlink.io/en/services/heartbeat.html).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -67,9 +67,9 @@
       </entry>
     </enum>
     <enum name="MAV_TYPE">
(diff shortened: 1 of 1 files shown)


hamishwillee:
Merged, as discussed in dev call.

PR #1189, opened 2019-08-01
OPTICAL_FLOW.flow_comp_m_(x/y) units to m/s

### LINKED ISSUE #1187: Clarify units in OPTICAL_FLOW
The units in OPTICAL_FLOW are off. The message now reads:
```

Field Name | Type | Units | Description
-- | -- | -- | --
time_usec | uint64_t | us | Timestamp  (UNIX Epoch time or time since system boot). The receiving end can  infer timestamp format (since 1.1.1970 or since system boot) by checking  for the magnitude the number.
sensor_id | uint8_t |   | Sensor ID
flow_x | int16_t | dpix | Flow in x-sensor direction
flow_y | int16_t | dpix | Flow in y-sensor direction
flow_comp_m_x | float | m | Flow in x-sensor direction, angular-speed compensated
flow_comp_m_y | float | m | Flow in y-sensor direction, angular-speed compensated
quality | uint8_t |   | Optical flow quality / confidence. 0: bad, 255: maximum quality
ground_distance | float | m | Ground distance. Positive value: distance known. Negative value: Unknown distance
flow_rate_x ** | float | rad/s | Flow rate about X axis
flow_rate_y ** | float | rad/s | Flow rate about Y axis
```

The issue is that `flow_comp_m_x` and `flow_comp_m_y` should be in m/s and not in m.

Optical flow is usually expressed in rad/s:
- The raw OF measurement is in pixels/frame (what is sent in `flow_x` and `flow_y`) but that is not really useful alone.
- Once corrected with frame rate, we obtain OF in pixels/s, and once corrected with focal length we obtain OF in rad/s (what is sent in `flow_rate_x` and `flow_rate_y`)
- For visual odometry, optical flow in rad/s is not useful either. It needs to be scaled using a distance sensor to obtain a speed measurement in m/s.

In PX4Flow Firmware, `flow_comp_m_x/y` contains values in m/s.

Possible solutions:
- modify the unit in the message description
- add an integration time-frame field (similar to OPTICAL_FLOW_RAD)

hamishwillee:
Fixes #1187

As stated by @jlecoeur the `OPTICAL_FLOW.flow_comp_m_x` and `flow_comp_m_y` must either be velocity units or an integration time unit is required so that flow can be determined from these. Given there is no such integration unit it is very likely that people are assuming m/s OR that it is not being used. Either way, the "best" fix is to clarify m/s.

Still waiting for confirmation on linked item from @auturgy.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4026,8 +4026,8 @@
       <field type="uint8_t" name="sensor_id">Sensor ID</field>
       <field type="int16_t" name="flow_x" units="dpix">Flow in x-sensor direction</field>
       <field type="int16_t" name="flow_y" units="dpix">Flow in y-sensor direction</field>
-      <field type="float" name="flow_comp_m_x" units="m">Flow in x-sensor direction, angular-speed compensated</field>
-      <field type="float" name="flow_comp_m_y" units="m">Flow in y-sensor direction, angular-speed compensated</field>
+      <field type="float" name="flow_comp_m_x" units="m/s">Flow in x-sensor direction, angular-speed compensated</field>
+      <field type="float" name="flow_comp_m_y" units="m/s">Flow in y-sensor direction, angular-speed compensated</field>
       <field type="uint8_t" name="quality">Optical flow quality / confidence. 0: bad, 255: maximum quality</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Discussed in dev call. Agreed that code for implementation of this message is velocity. Can be merged.

PR #1192, opened 2019-08-02
MAV_CMDs - put param labels next to index

hamishwillee:
This moves the param lablels next to the index, if they exist. This is much more readable, and helps users see what the index is for. 

image

The old way is below. In this case all "GCS display hints" are displayed together. 

image


hamishwillee:
Self merging. I think this is a clear improvement.

PR #1193, opened 2019-08-05
common.xml: Set MAV_AUTOPILOT_INVALID for non-autopilots

hamishwillee:
A non-flight controller component should have `HEARTBEAT.autopilot` and `HIGH_LATENCY2.autopilot` set to MAV_AUTOPILOT_INVALID.


hamishwillee:
Self merging (change spells out what should be obvious).

Issue #1159, opened 2019-06-05
Proposal for an opaque, tunnel message


olliw42:
The proposal is to establish a MAVlink message which allows a component to send specifc, non-standardized data to another component. 

Currently this is not possible in a MAVLink-ish way, and even adding one owns message to one owns dialect to both components typically won't work when the data transmission involves some "router", which is frequently the case. For instance: GUI to autopilot to component. The "router" or autopilot won't be able to forward since it wouldn't know about the new message and disregard it. A standardized message for transmitting non-standardized data would resolve this.

I'm not sure I should present arguments and/or examples for why such a capability is needed. I suspect that such a discussion could quickly become long and less useful. So, for briefness I assume that a majority will agree with the need, and enhance the point by mentioning that MAVLink actually already has this sort of message in other context:

An obvious one is `FILE_TRANSFER_PROTOCOL`, which is very explicitly described as opaque. In fact, the above task could be achieved with `FILE_TRANSFER_PROTOCOL`. However, I feel that this would be a bit of a hack and dirty misuse, given that even if it is opaque it serves the specific purpose of FTP. Also messages such as `GPS_INJECT_DATA` or `SERIAL_CONTROL` should be considered part of this class. 

Finally, there is `V2_EXTENSION`. In fact, this message and its accompanying specification for message_type is, except of the target_network field, essentially exactly what I would have suggested! However, it feels strange. First, the name. Second, the description "Message implementing parts of the V2 payload specs in V1 frames for transitional support". Doesn't sound like long-living and general purpose. Third, I've tried to find the extension-message-ids.xml, but failed, which makes me further wonder what the status of the `V2_EXTENSION` message is.

Hence this proposal.

I propose a message, maybe named `TUNNEL`, with target fields, a protocol field which is exactly handled like the message_type field of `V2_EXTENSION`, and an opaque data array. I would not have added a target_network field, simply because its use is beyond my imagination, but if the consensus would be that it might bring some advantages it of course should be added, in which case the new message would be essentially identical to `V2_EXTENSION`. 

The outcome of this proposal of course also could be to solidify `V2_EXTENSION` and establish this message for the said purpose, and to add its use as opaque tunnel message to the message description, or change the message description completely in this sense.

Cheers, Olli

hamishwillee:
A couple of points
[...]

PR #1194, opened 2019-08-05
Opaque Tunnel message

olliw42:
MAVlink message which allows a component to send specific, non-standardized data to another component. Addresses #1159 (Proposal for an opaque, tunnel message). 

In comparison to the proposal and/or the V2_EXTENSION message, I've made some changes:

* broadcast is not forbidden, but should be discouraged and not the normal use case
* I removed the target_network field, simply because I don't know what its use would be. If there is some use, it of course could/should be added,.
* I've changed the handling of the code for the encoding, i.e. payload_type, somewhat. Fundamentally only a registered number is required, so I've made it an enum. This also makes coding for implementers simpler since the enum is then already in the header ready to use. Since there is no example it isn't totally clear to me how the .xml was planned to work. I suspect that the intention was to not only use it for registering a value but also to have some xml markup to describe the structure of the data. This however would require specifying the markup, which I think would be better done at a later stage, since it's not really fundamental to bringing in the tunnel message. Similar so for the exact procedures of accepting/adding new payload types; can be discussed any time later. I suggest a friendly non-strict acceptance procedure, which minimally could consist of accepting a new code to the enum. The associated .xml is then more the "code-holder"'s responsibility.
* I've reduced the payload size to 128 bytes. I do not really see any advantage of a large payload size, and the smaller size can help implementations on not so powerful microprocessors. For comparison, the UAVCAN tunnel message has a payload size of 60, which works very well as much as I can tell from experience. (so, if you ask me the payload size could also be reduced to 64...).

One point which isn't clear to me is if there should be some payload_length field or not. The approach here is taken from `V2_EXTENSION` and `FILE_TRANSFER_PROTOCOL`, but I note that there are several messages in common.xml  with non-string variable-length data which DO have an explicit length field, e.g. `LOGGING_DATA`, `LOGGING_DATA_ACKED` (which are also MAvlink2). Hopefully someone could clarify.

This is of course open for discussion. Not sure I should have marked it as WIP.

BTW, I chose a message id of 701 without any reason. Please suggest what a more reasonable choice would be.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -666,11 +666,6 @@
         <description/>
       </entry>
     </enum>
-    <enum name="MAV_TUNNEL_PAYLOAD_TYPE">
-      <entry value="0" name="MAV_TUNNEL_PAYLOAD_TYPE_UNKNOWN">
-        <description>Encoding of payload unknwon.</description>
-      </entry>
-    </enum>
     <!-- fenced mode enums -->
     <enum name="FENCE_ACTION">
       <entry value="0" name="FENCE_ACTION_NONE">
@@ -3229,6 +3224,11 @@
       </entry>
       <entry value="2" name="PARACHUTE_RELEASE">
         <description>Release parachute.</description>
+      </entry>
+    </enum>
+    <enum name="MAV_TUNNEL_PAYLOAD_TYPE">
+      <entry value="0" name="MAV_TUNNEL_PAYLOAD_TYPE_UNKNOWN">
+        <description>Encoding of payload unknown.</description>
       </entry>
     </enum>
   </enums>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5219,7 +5219,9 @@
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
     </message>
-    <message id="701" name="TUNNEL">
+    <message id="385" name="TUNNEL">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Message for transporting "arbitrary" variable-length data from one component to another (broadcast is not forbidden, but discouraged). The encoding of the data is usually extension specific, i.e. determined by the source, and is usually not documented as part of the mavlink specification.</description>
       <field type="uint8_t" name="target_system">System ID (can be 0 for broadcast, but this is discouraged)</field>
       <field type="uint8_t" name="target_component">Component ID (can be 0 for broadcast, but this is discouraged)</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -666,6 +666,11 @@
         <description/>
       </entry>
     </enum>
+    <enum name="MAV_TUNNEL_PAYLOAD_TYPE">
+      <entry value="0" name="MAV_TUNNEL_PAYLOAD_TYPE_UNKNOWN">
+        <description>Encoding of payload unknwon.</description>
+      </entry>
+    </enum>
     <!-- fenced mode enums -->
     <enum name="FENCE_ACTION">
       <entry value="0" name="FENCE_ACTION_NONE">
@@ -5213,6 +5218,13 @@
       <field type="int32_t" name="mission_next_item" units="s">Estimated time for reaching/completing the currently active mission item. -1 means no time estimate available.</field>
       <field type="int32_t" name="mission_end" units="s">Estimated time for completing the current mission. -1 means no mission active and/or no estimate available.</field>
       <field type="int32_t" name="commanded_action" units="s">Estimated time for completing the current commanded action (i.e. Go To, Takeoff, Land, etc.). -1 means no action active and/or no estimate available.</field>
+    </message>
+    <message id="701" name="TUNNEL">
+      <description>Message for transporting "arbitrary" variable-length data from one component to another (broadcast is not forbidden, but discouraged). The encoding of the data is usually extension specific, i.e. determined by the source, and is usually not documented as part of the mavlink specification.</description>
+      <field type="uint8_t" name="target_system">System ID (can be 0 for broadcast, but this is discouraged)</field>
+      <field type="uint8_t" name="target_component">Component ID (can be 0 for broadcast, but this is discouraged)</field>
+      <field type="uint16_t" name="payload_type" enum="MAV_TUNNEL_PAYLOAD_TYPE">A code that identifies the content of the payload (0 for unknown, which is the default). If this code is less than 32768, it is a 'registered' payload type and the corresponding code should be added to the MAV_TUNNEL_PAYLOAD_TYPE enum, and the entry possibly to https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. Software creators can register blocks of types as needed. Codes greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
+      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">


message_definitions/v1.0/common.xml
```diff
@@ -666,6 +666,11 @@
         <description/>
       </entry>
     </enum>
+    <enum name="MAV_TUNNEL_PAYLOAD_TYPE">
+      <entry value="0" name="MAV_TUNNEL_PAYLOAD_TYPE_UNKNOWN">
+        <description>Encoding of payload unknwon.</description>
```
hamishwillee: Typo unknown

olliw42: oje, thx. corrected.

message_definitions/v1.0/common.xml
```diff
@@ -5214,6 +5219,13 @@
...
+      <field type="uint8_t" name="target_system">System ID (can be 0 for broadcast, but this is discouraged)</field>
+      <field type="uint8_t" name="target_component">Component ID (can be 0 for broadcast, but this is discouraged)</field>
+      <field type="uint16_t" name="payload_type" enum="MAV_TUNNEL_PAYLOAD_TYPE">A code that identifies the content of the payload (0 for unknown, which is the default). If this code is less than 32768, it is a 'registered' payload type and the corresponding code should be added to the MAV_TUNNEL_PAYLOAD_TYPE enum, and the entry possibly to https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. Software creators can register blocks of types as needed. Codes greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
+      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
```
hamishwillee: Suggest you use the full available payload size. Reason is to benefit from [empty byte truncation](https://mavlink.io/en/guide/serialization.html). 
Basically this allows you to fill whatever fraction of the message you want - any bytes that are 0 at the end don't get sent.

hamishwillee:
Is there any need for very large messages or resends? Ie are you ever going to want to compose larger messages from smaller messages, and if so do you need support in this packet or will that be handled by the protocol that handles your opaque data?

hamishwillee:
@LorenzMeier I don't have any view on how you envisage this working. Can you review please.

message_definitions/v1.0/common.xml
olliw42: I fully understand your point. Yet I think it might be discussed a bit further. As said, I don't really see a significant benefit and for "smaller" microcontrollers it can be a pain. I suggest to let's hear further voices :).

hamishwillee: Probably worth thinking about some examples of real data and real microcontrollers that will need this.
But very happy to leave this discussion with you and @LorenzMeier.

olliw42: a real example would be the STorM32. The MAVLink stuff is by now the largest RAM consumer, increasing the RAM consumption by nearly 50% compared to what it would need without MAVLink (and I believe I have a very efficient implementation, see my light PR). 128bytes by itself is really not much, but it all adds up. And if there is no real benefit ...

olliw42:
> Is there any need for very large messages or resends? Ie are you ever going to want to compose larger messages from smaller messages, and if so do you need support in this packet or will that be handled by the protocol that handles your opaque data?

I would suggest to leave that to the payload-type holder, as it can depend and several mechanisms are possible, and none of that has to do with MAVLink. E.g., if the specific use case doesn't have long data messages, a simple send-and-digest can do it. In other cases the specific use case might want to put additional meta data into the payload, similar to how it is done for the MAVLink FTP. In yet other cases the packeting and reassembly can be done based on time and simple buffering.

The case of the UAVCAN tunnel message (which only has 60 bytes payload) would be an example of the last. magicrub offered an implementation where a timeout is used (https://github.com/ArduPilot/ardupilot/pull/8508), I offered a more elaborate approach using a "blank" detector (https://github.com/olliw42/ardupilot/tree/BetaCopter-3.6.7-v020u/libraries/AP_UAVCAN). These implementations work like transparent links, i.e. you wouldn't really notice them, independent on any packet sizes of the "original" protocol (like MAVLink messages). A video of this working in reality you have here: https://www.youtube.com/watch?v=Sw24wAIoemA. So, the bottom line, there are already real-world working implementations of this kind of message. I don't forsee any issue in this regards.

julianoes:
@olliw42 I have two questions on this. Sorry if you have already answered them and I missed it.

1. Have you considered [ENCAPSULATED_DATA](https://mavlink.io/en/messages/common.html#ENCAPSULATED_DATA)? If yes, why does it not work?

2. Why does the tunnel message not include a sequence number? Wouldn't that be required for retransmission in case it gets lost?

olliw42:
1. I haven't considered ENCAPSULATED_DATA. First, by its description it is linked to images and a specific protocol. So even if it could be used I would feel it would be a misuse. Note that a similar argument applied to the many other messages which could be mis-used, such as FILE_TRANSFER_PROTOCOL, SERIAL_CONTROL, and especially V2_EXTENSION. I also think it misses the target fields, since IMHO broadcasting opague data should not be the default but rather be discouraged I would consider this a significant miss. 

If one wants to use one of the existing messages - which in principle would be fine with me if the spec would allow it to be used so - I would rather prefer to go with V2_EXTENSION, or maybe FILE_TRANSFER_PROTOCOL

2. My position is that this can be decided by whoever wants to use that message for whatever purpose. If for a particular purpose it is concluded that a sequence number is needed and that the two ends talking to each other need to retransmit, then this seq number can be included into the payload (and anyone would know due to the code in payload_type). Note that this is exactly the position take for MAVLinkFTP. Note also that a seq number not only requires adding a seq number, but specifying some some sort of additional communication to tell the other end that a retransmission is required. In many applications all this might not be needed however and would be really overdoing it. I look at it like TCP vs UDP or non-loss vs lossy transmission in general. There are obviously many applications which work fine with loss. To be specific, for the STorM32 and anything else I have done so far I can't imagine any situation where I would want to bother with a sequence number and the additional complexity brought in by it.

If you think a seq number should be added even if it is not used, fine. I would not like however if the spec would require that this seq number MUST be used and always a non-lossy communication be implemented.

julianoes:
> since IMHO broadcasting opague data should not be the default but rather be discouraged I would consider this a significant miss.

That's a good point.

> I look at it like TCP vs UDP or non-loss vs lossy transmission in general. There are obviously many applications which work fine with loss.

Fair point.

Ok given your explanations I'm fine with the current suggestion then.

olliw42:
cool

I guess I should change the id to 390 then ?? Or 385 ??
(as 385 is likely to be occupied by COMPONENT_INFORMATION if that is going to be, or I change this here to 835 if it's better to always go with which goes first)

hamishwillee:
@olliw42 I'll push this in first, so please make it 385 :-)

We discussed this in dev call. Can you please add the `<wip />` tag into your message - just copy the pattern of another message.  (or see [here](https://mavlink.io/en/guide/xml_schema.html#wip)). I'll then push it in. We can remove the tag later after this has been tested in anger - but what it means is that no one can complain if you decide on a better design while you are testing.

olliw42:
thx
both done
(although I'm quite convinced that the message is ok, a wip doesn't hurt)

hamishwillee:
Awesome. Merged! Keen to see how it goes for your use cases.

olliw42:
COOL!

My first use case will be to establish a passthrough between gimbal GUI and gimbal ... this is currently possible with BetaCopter (https://www.youtube.com/watch?v=lz_0BdGGuxc, https://www.youtube.com/watch?v=IF1_38fVppU), but the mechanism is really "tricky" and anything MAVLink-ish ... the cool thing now will be that it would work with any flight controller, not only with my BetaCopter fork, but also PX4 if you wish ... isn't this how it is supposed to be? I might not be fast however with getting this done since it will need lot's of rewrite on both the GUI and STorM32 fimware side ... but I'm very happy to have this now on my plate of things to do LOL

hamishwillee:
>  isn't this how it is supposed to be?

Yes, if this is effectively an internal interface for a closed use case, then the fact you can do what you need to without impacting anyone else is fantastic.

Generally we're trying to create clear interfaces for common behaviours so that anyone can implement to them and GCS/endpoints can rely on the behaviour. Part of the reason we have been a bit nervous of this message.

hamishwillee:
^^^^ But you know that :-)

olliw42:
yes, I know that you folks are a bit concerned about what you call misuse of this message.

(I'm not so much, and I think it's better to have that message than people using other messages such as V2_EXTENSION or ENCAPSULATED_DATA out of desperation)

the pass-through thing is IMHO an excellent use case. I'm a bit confused now as to what you are saying: You consider it a good use case too, or not?

hamishwillee:
I think it is a good use case.

olliw42:
uff ... I was a bit worried now that it would be me who sets the first example of how it should not be used ...

thx. Have fun.

olliw42:
@hamishwillee 

> Keen to see how it goes for your use cases.

here for your entertainment a demo of the MAVLink tunnel passthrough (and that it works as intended): https://youtu.be/FRyZ7VZ5lvY

hamishwillee:
@olliw42 

Yes, it does work well. I guess one reason you don't need a bigger payload is the use case - easy access to the gimbal via the autopilot - is always going to be over a fast link where you the overhead of MAVLink packet itself makes little difference.

And presumably if PX4 routes the packets, then you can do exactly the same things on that flight stack too. 

Thanks very much for putting that together. Often don't get to see the end result with such a short lag :-)

olliw42:
> I guess one reason you don't need a bigger payload is the use case - easy access to the gimbal via the autopilot - is always going to be over a fast link where you the overhead of MAVLink packet itself makes little difference.

actually, slow links are my main reason/concern for why I use 64 bytes instead of 128 bytes.
The overhead is - IMHO - really negligible: 

64 bytes: message length is 12 bytes header/crc + 5 bytes tunnel header +64 bytes payload = 81 bytes

128 bytes:  message length is 12 bytes header/crc + 5 bytes tunnel header +128 bytes payload = 145 bytes

thus 2 x 64 messages is 162 bytes compared to 145 bytes for 1 x 128 message => 11% overhead

for 128 vs 250 the overhead would be just 5%.

However, the time for one packet for even a "fast" 115200 bps link is:

64 bytes message = 81 bytes = ca 7 ms
128 bytes message = 145 bytes = ca 14 ms

14 ms is really a long time for which the link is saturated (!!) (and it obviously would be worse for 250 bytes payload messages). I think smaller junks are really better as they help keeping the message flow smooth.

So, I do not agree that it's just the use case. :) 128 bytes is a good compromise.

And note that transporting massive amounts of data won't (shouldn't) be the most typical use case - as it's with any slow link ...

> And presumably if PX4 routes the packets, then you can do exactly the same things on that flight stack too.

YEAH!!! That's the idea, that it now works for any flight stack! And it really should, I can't see any reason why it wouldn't.
Sadly I can't test myself since I'm still not on PX4 ... I need to change that at some point LOL 
(if there would be a default PX4 setup configuration for the Solo...)

Thx fo all.

PR #1196, opened 2019-08-07
MESSAGE_INTERVAL interface, improve description of its usage

olliw42:
at least I found the description of the MESSAGE_INTERVAL interface unclear (and in fact had implemented it incorrectly, in comparison to how it is used in ArduPilot). This PR attempts to make the description more clear.

If the suggested improved descriptions do not correctly describe how that interface is supposed to work, please advice.


hamishwillee:
Yes, this is correct, and a clear improvement. Merging. Thanks very much!

Issue #1179, opened 2019-07-10
Message SCALED_IMU - magnetic field


MrTypename:
According documentation message [SCALED_IMU](https://mavlink.io/en/messages/common.html#SCALED_IMU), magnetic field values (xmag, ymag, zmag) represent the magnetic field in [mT]. According to [wikipedia](https://en.wikipedia.org/wiki/Earth%27s_magnetic_field) the earth magnetic field is, in worst case, something around 65[uT]. Which is equal to 0.065[mT]. So casting 0.065[mT] to (int) results always in 0. It seems that value should be expressed in [uT] not [mT] ?

hamishwillee:
Yes, that sounds like a screw up. @auturgy @julianoes You got any thoughts on thsi?

auturgy:
ArduPilot uses mG internally. I’m not sure if there’s a 10^-4 somewhere to convert or we send mG not mT.


Regards,

James

hamishwillee:
Hmm. PX4 [source docs](https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_messages.cpp#L862) think it is dealing in mT and there is no obvious scaling.

auturgy:
(I should clarify that ArduPilot doesn’t send SCALED_IMU, but does send RAW_IMU)

julianoes:
I had a look in PX4 and it looks like we send the "raw" sensor values, so not "scaled" at all but scaled by whatever is acquired from the sensors.
I checked the often used hmc5883 and it looks like the range is set to +/-2 Ga mapping to 12 bits -2048..2047. Of course that might differ for other sensors.

Also, PX4 doesn't seem to implement `RAW_IMU`. 

So it looks like we should send what is currently in `SCALED_IMU` in `RAW_IMU` and implement SCALED_IMU correctly (to whatever spec makes sense).

https://github.com/PX4/Firmware/blob/2fbb70d9ca94f442de1345221be18ba738cda952/src/modules/mavlink/mavlink_messages.cpp#L957-L959

https://github.com/PX4/Firmware/blob/2fbb70d9ca94f442de1345221be18ba738cda952/msg/sensor_mag.msg#L13-L15

https://github.com/PX4/Firmware/blob/2fbb70d9ca94f442de1345221be18ba738cda952/src/drivers/magnetometer/hmc5883/hmc5883.cpp#L891-L894

julianoes:
[...]

PR #1200, opened 2019-08-08
RAW_IMU: Add id extension to support multiple IMU

hamishwillee:
As discussed in #1179 and dev call, it is useful to be able to report multiple RAW_IMU. Rather than having separate messages (as for SCALED_IMUn) this adds the identity as an extension field.

Systems that do not understand the field will interpret a message as being for IMU1 even if it is from another IMU. We're expecting systems to update before this becomes a problem. 

@auturgy @julianoes As discussed in dev call. Is this OK to merge?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3636,7 +3636,7 @@
       <field type="int16_t" name="ymag">Y Magnetic field (raw)</field>
       <field type="int16_t" name="zmag">Z Magnetic field (raw)</field>
       <extensions/>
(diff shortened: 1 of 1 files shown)


LorenzMeier:
RAW IMU is intended to be really raw (otherwise it has a loss of precision) and so what people really want is HIGHRES_IMU - I've added the field to that one as well and removed the incorrect unit. Please review.

LorenzMeier:
Just FYI: I want all multi-commit message changes to be squashed - hence the commit message. There should NOT be EVER incremental commits changing a single message or message set.

hamishwillee:
Change looks good thanks @LorenzMeier . Default is "squash and merge", and I don't change that.

PR #1201, opened 2019-08-08
common.xml:GPS_GLOBAL_ORIGIN clarify when emitted

hamishwillee:
This improves wording of GPS_GLOBAL_ORIGIN and SET_GPS_GLOBAL_ORIGIN  (which I found a little confusing).

It also states what happens if setting the origin might fail - i.e. GPS_GLOBAL_ORIGIN should be sent even if setting a local origin failed (for whatever reason). This might not be agreeable to all users, but it is generally good practise. I created it so that we can remove ambiguity.

@auturgy @julianoes @nsteele

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4065,7 +4065,7 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="48" name="SET_GPS_GLOBAL_ORIGIN">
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes Yes, @nsteele is creating an implementation, which is one reason I'm trying to make sure that requirements are clear.

@auturgy Any objection to me merging?

auturgy:
LGTM

hamishwillee:
Merging, as discussed in mavlink dev call.

PR #1204, opened 2019-08-12
Correct MAV_TYPE entry for servo

olliw42:
does two simple things
1. corrects the MAV_TYPE enum entry for a servo type to MAV_TYPE_SERVO (was SERVO, which looks like an obvious mistake to me)
2. grabs the occasion and makes the MAV_TYPE enum entries in minimal.xml match those in common.xml

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -168,7 +168,7 @@
       <entry value="32" name="MAV_TYPE_FLARM">
         <description>FLARM collision avoidance system</description>
       </entry>
-      <entry value="33" name="SERVO">
+      <entry value="33" name="MAV_TYPE_SERVO">
         <description>Servo</description>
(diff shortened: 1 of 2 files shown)


hamishwillee:
Thanks! SERVO fix great.  Minimal.xml should match what is in common.xml precisely - so that when we are able to #include minimal there is no question about correctness.

olliw42:
am I supposed to do anything here?

hamishwillee:
@olliw42 Yes please. Remove the changes to minimal.xml or change them to match common.xml precisely (whichever you prefer).

olliw42:
ähm, I'm under the impression that I exactly carried over the changes I made to common.xml to minimal.xml ...

hamishwillee:
Thanks. You did - not sure why I thought there was a mismatch.

PR #1205, opened 2019-08-13
Allow IMU's to convey their temperature

WickedShell:
This is useful information for the GCS and safety reasons. The main question I have is do we want a DNU value for this? (IE `INT16_MIN`) The downside to using a DNU is that if you don't have the extension you get 0, which is inband. We could also just declare 0 to be the DNU, and if you send a value of 0, then just add 1 to it, and accept the slight error.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3877,6 +3877,8 @@
       <field type="int16_t" name="xmag" units="mgauss">X Magnetic field</field>
       <field type="int16_t" name="ymag" units="mgauss">Y Magnetic field</field>
       <field type="int16_t" name="zmag" units="mgauss">Z Magnetic field</field>
+      <extensions/>
+      <field type="int16_t" name="temperature" units="cdegC">Temperature, 0 indicates that the IMU does not provide the temperature, if the IMU is actually at 0C, then a value of 1 should be sent.</field>
     </message>
     <message id="27" name="RAW_IMU">
       <description>The RAW IMU readings for a 9DOF sensor, which is identified by the id (default IMU1). This message should always contain the true raw values without any scaling to allow data capture and system debugging.</description>
@@ -3892,6 +3894,7 @@
       <field type="int16_t" name="zmag">Z Magnetic field (raw)</field>
       <extensions/>
       <field type="uint8_t" name="id">Id. Ids are numbered from 0 and map to IMUs numbered from 1 (e.g. IMU1 will have a message with id=0)</field>
+      <field type="int16_t" name="temperature" units="cdegC">Temperature, 0 indicates that the IMU does not provide the temperature, if the IMU is actually at 0C, then a value of 1 should be sent.</field>
     </message>
     <message id="28" name="RAW_PRESSURE">
       <description>The RAW pressure readings for the typical setup of one absolute pressure and one differential pressure sensor. The sensor values should be the raw, UNSCALED ADC values.</description>
@@ -4752,6 +4755,8 @@
       <field type="int16_t" name="xmag" units="mgauss">X Magnetic field</field>
       <field type="int16_t" name="ymag" units="mgauss">Y Magnetic field</field>
       <field type="int16_t" name="zmag" units="mgauss">Z Magnetic field</field>
+      <extensions/>
+      <field type="int16_t" name="temperature" units="cdegC">Temperature, 0 indicates that the IMU does not provide the temperature, if the IMU is actually at 0C, then a value of 1 should be sent.</field>
     </message>
     <message id="117" name="LOG_REQUEST_LIST">
       <description>Request a list of available logs. On some systems calling this may stop on-board logging until LOG_REQUEST_END is called.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
My thought is that we do want DNU. FWIW My preference is to use INT16_MAX - yes this will be risky initially, but over a relatively short timeframe systems should adapt.

Is there any value in also updating the RAW_IMU with a temperature value?

hamishwillee:
Would there be any value in having a new message SCALED_IMU_V2 which includes both this field and an id value - ie it can be used for arbitrary IMU messages? We've proposed id as an extension for RAW_IMU, so this would align the messages going forward.

WickedShell:
The problem with the DNU is you can't get it if it was sent MAVLink 1 and you have a MAVLink2 aware parser, unless your application knows it was a MAVLink 1 encoded, or your parser drops fields if they were not set. You also have a lot of legacy problems with a DNU of not 0.

> Is there any value in also updating the RAW_IMU with a temperature value?

That's the first line of change? This does RAW_IMU, SCALED_IMU2, SCALED_IMU3. I don't see how an ID field helps scaled2, and it causes problems for scaled3

hamishwillee:
1. I fully understand the problem with DNU. I am proposing we ignore them. How serious would that be in reality? Less or more irritating that using 0 as DNU?

>> Is there any value in also updating the RAW_IMU with a temperature value?

> That's the first line of change? This does RAW_IMU, SCALED_IMU2, SCALED_IMU3. I don't see how an ID field helps scaled2, and it causes problems for scaled3

Ah I actually missed the fact you'd done RAW_IMU.

In another PR we have a proposal to add `id` as extension to RAW_IMU, allowing RAW_IMU to be used for _instances_ of RAW_IMU (currently there is just one message, and it can be used for just one IMU). We are aware that this potentially breaks MAVLink 1, and any system that hasn't updated to the latest message set. 

So what I was suggesting was that we might use a similar approach for SCALED_IMU - ie add another message with an id (for instances) and the temperature. Because we're creating a new message we don't have to worry about the other issues. Though of course people may need to migrate to a new message. 

It might be a bad idea, just figured I'd throw it in.

WickedShell:
> In another PR we have a proposal to add id as extension to RAW_IMU, allowing RAW_IMU to be used for instances of RAW_IMU 

This was already merged.

> We are aware that this potentially breaks MAVLink 1, and any system that hasn't updated to the latest message set.

That's not a potentially breaks, it does break it if you send it and the end system isn't aware of that/decoding MAVLink 2. It's a rather terrible idea to try and retrofit that IMO.

> So what I was suggesting was that we might use a similar approach for SCALED_IMU - ie add another message with an id (for instances) and the temperature. Because we're creating a new message we don't have to worry about the other issues. Though of course people may need to migrate to a new message.
>
>It might be a bad idea, just figured I'd throw it in.

This is an entirely independent conversation to adding the temperature, and I'd rather see it discussed on a separate issue, rather then drag this down.

hamishwillee:
The time to decide you hate a change is before it goes in :-). If you feel strongly enough it has only been a week, raise an issue, we can revert and create a new message altogether for this. 

Anyway, for temperature, there is no good solution, so suggest you put in your preference as text and we all argue it.

hamishwillee:
Suggestion is use INT16_MAX if you want to use an extension - that's the standard way of doing things. If that isn't "safe enough" for you, then probably best way is to declare a new message.

magicrub:
I'm in favor of this change, and also using INT16_MAX as invalid.

hamishwillee:
@WickedShell This is waiting on you - as above.

WickedShell:
@hamishwillee I think INT16_MAX doesn't work because you can't distinguish it when you get sent a 0 from a vehicle if it's old or current. And I don't want to make yet another new message as we'd be stuck sending & supporting both for quite awhile with the mixture of GCS versions out there.

hamishwillee:
@WickedShell True. We know. Is the problem with doing this right (ie using INT16_MAX) that you are unable to identify when you're talking to an older version of the flight stack?

@auturgy @magicrub @julianoes The other option is to use 0 as _not supported_ and either say 1/-1 are first valid values (live with not having zero) or use INT16_MAX for 0. Personally I prefer a new message or an extension as discussed in the dev call.

hamishwillee:
The special case of 0 bothers me because it is unintuitive and inconsistent. Specs work much better if people can generally expect things to work the same way/there are few special cases. FWIW I think I'd slightly prefer INT16_MAX for zero if we went this route.

Unless @auturgy @magicrub @julianoes say they are happy with an alternative, dev call position was extension or new message or nothing. So waiting on them to see if they have a different opinion now.

tridge:
the special for 0 isn't really a big deal, if we cared we could report 0.01C for zero, but we really don't care that much

tridge:
> The special case of 0 bothers me because it is unintuitive and inconsistent. Specs work much better if people can generally expect things to work the same way/there are few special cases. FWIW I think I'd slightly prefer INT16_MAX for zero if we went this route.

no, INT16_MAX would make graphing horrible, using '1' is better, giving 0.01C

hamishwillee:
Thanks @tridge. @WickedShell Document it this way then. @julianoes Any objection?

julianoes:
> @julianoes Any objection?

Fine with me!

message_definitions/v1.0/common.xml
```diff
@@ -4874,6 +4879,8 @@
       <field type="int16_t" name="xmag" units="mgauss">X Magnetic field</field>
       <field type="int16_t" name="ymag" units="mgauss">Y Magnetic field</field>
       <field type="int16_t" name="zmag" units="mgauss">Z Magnetic field</field>
+      <extensions/>
+      <field type="int16_t" name="temperature" units="cdegC">Temperature, 0 indicates that the IMU does not provide the temperature, if the IMU is actually at 0C, then a value of 1 should be sent.</field>
```
hamishwillee: Looks good to me. Optional rewording.

> Temperature, 0: IMU does not provide temperature values. If the IMU is at 0C it must send 1 (0.01C).

PR #1206, opened 2019-08-14
Ardupilot downstream changes

amilcarlucas:


message_definitions/v1.0/ardupilotmega.xml
```diff
@@ -231,6 +231,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="42651" name="MAV_CMD_BATTERY_RESET" hasLocation="false" isDestination="false">
+        <description>Reset battery capacity for batteries that accumulate consumed battery via integration.</description>
+        <param index="1" label="battery mask">Bitmask of batteries to reset. Least significant bit is for the first battery.</param>
+        <param index="2" label="percentage" minValue="0" maxValue="100" increment="1">Battery percentage remaining to set.</param>
+        <param index="3">Reserved, send 0</param>
```
hamishwillee: FYI only, default is reserved send 0, so these could be omitted.

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5102,8 @@
       <field type="float" name="horiz_accuracy" units="m">GPS horizontal accuracy</field>
       <field type="float" name="vert_accuracy" units="m">GPS vertical accuracy</field>
       <field type="uint8_t" name="satellites_visible">Number of satellites visible.</field>
+      <extensions/>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw of vehicle, zero means not available, use 36000 for north</field>
```
hamishwillee: Changes to common should be submitted to common. @auturgy @julianoes got any opinion on this change?

auturgy:
The gps yaw extension was implemented for a specific use case (testing ekf performance with a mocap system capable of providing either mocap or synthesised gps messages), but is also generally useful: there are gps units that can provide yaw. In general I agree that changes to common pass through upstream, but I think it should be considered normal that some will be prototyped in forks then pushed up, as has happened here.

hamishwillee:
@auturgy I'm not so concerned about the origin/direction of the push as the fact that it is mixed with other changes. But sure.

Yes it does look useful.

hamishwillee:
@julianoes Are you OK with this?

message_definitions/v1.0/common.xml
LorenzMeier: Given everything else in this message is CI units the heading should be degrees as well. And this is not clearly specified as heading or as course over ground (which we already have), so it warrants a discussion and much better documentation.

amilcarlucas: @tridge this is vehicle yaw (heading) and not course over ground right ?

message_definitions/v1.0/common.xml
```diff
@@ -5508,9 +5510,9 @@
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="float" name="roll" units="deg">Roll in global frame (set to NaN for invalid).</field>
       <field type="float" name="pitch" units="deg">Pitch in global frame (set to NaN for invalid).</field>
-      <field type="float" name="yaw" units="deg">Yaw relative to vehicle(set to NaN for invalid).</field>
+      <field type="float" name="yaw" units="deg">Yaw relative to vehicle (set to NaN for invalid).</field>
       <extensions/>
-      <field type="float" name="yaw_absolute" units="deg">Yaw in absolute frame, North is 0 (set to NaN for invalid).</field>
+      <field type="float" name="yaw_absolute" units="deg">Yaw in absolute frame relative to Earth's north, north is 0 (set to NaN for invalid).</field>
```
hamishwillee: North is a proper noun. Must be capitalised.

hamishwillee: That is true on other changes too.

message_definitions/v1.0/common.xml
```diff
@@ -5642,14 +5644,14 @@
...
       <field type="uint16_t" name="max_distance" units="cm">Maximum distance the sensor can measure.</field>
       <extensions/>
       <field type="float" name="increment_f" units="deg">Angular width in degrees of each array element as a float. If non-zero then this value is used instead of the uint8_t increment field. Positive is clockwise direction, negative is counter-clockwise.</field>
       <field type="float" name="angle_offset" units="deg">Relative angle offset of the 0-index element in the distances array. Value of 0 corresponds to forward. Positive is clockwise direction, negative is counter-clockwise.</field>
-      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame of reference for the yaw rotation and offset of the sensor data. Defaults to MAV_FRAME_GLOBAL, which is North aligned. For body-mounted sensors use MAV_FRAME_BODY_FRD, which is vehicle front aligned.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame of reference for the yaw rotation and offset of the sensor data. Defaults to MAV_FRAME_GLOBAL, which is north aligned. For body-mounted sensors use MAV_FRAME_BODY_FRD, which is vehicle front aligned.</field>
```
hamishwillee: Why change. North was correct.

message_definitions/v1.0/common.xml
```diff
@@ -5642,14 +5644,14 @@
...
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the vehicle with index 0 corresponding to North + angle_offset, unless otherwise specified in the frame. A value of 0 is valid and means that the obstacle is practically touching the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
+      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the vehicle with index 0 corresponding to north + angle_offset, unless otherwise specified in the frame. A value of 0 is valid and means that the obstacle is practically touching the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
```
hamishwillee: Same. Why change?

message_definitions/v1.0/common.xml
hamishwillee: > This is already used in ardupilot. changing it to degrees float will cause the existing SW to break. There have been recent commits also using cdeg.

And that's why the process is to go into this repo first.

hamishwillee:
@amilcarlucas Thanks for making the change. Modulo the spelling of North, I'm OK with this. 

@julianoes Are you OK with this "overall" the specific blocker

message_definitions/v1.0/common.xml
amilcarlucas: According to the spelling in wikipedia north should not be capitalized. I followed that.
But you`re a native speaker, so I'll capitalize all North, East, West and South on the common.xml for consistency.

hamishwillee: Hmmm. Don't change it everywhere.
It gets capitalised if it is used as a proper noun but not otherwise. What is hard here is whether  this usage makes it a proper noun. I *think* so (certainly it looks wrong to me as written). When you say North-aligned I think you're saying alighted with the North pole, so the "thing".

amilcarlucas: Aligned to the north direction, that is how I read it. I went again and fixed it.

hamishwillee: I'm not certain, but OK with this.

hamishwillee:
Thanks @amilcarlucas . I think this is good. Merging.

PR #1211, opened 2019-08-20
msg_obstacle_distance: increment_f allows negative values

magicrub:
increment_f allows negative values

also clarified some of the other fields

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5546,14 +5546,14 @@
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the UAV with index 0 corresponding to local forward + angle_offset. A value of 0 means that the obstacle is right in front of the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
-      <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element. (Ignored if increment_f greater than 0).</field>
+      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the vehicle with index 0 corresponding to vehicle forward + angle_offset. A value of 0 is valid and means that the obstacle is practically touching the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
+      <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element. Increment direction is clockwise. This field is ignored if increment_f is non-zero.</field>
       <field type="uint16_t" name="min_distance" units="cm">Minimum distance the sensor can measure.</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Makes sense. I doubt anyone would have implemented anything different at this point. Merging

message_definitions/v1.0/common.xml
```diff
@@ -5546,13 +5546,13 @@
...
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the UAV with index 0 corresponding to local forward + angle_offset. A value of 0 means that the obstacle is right in front of the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
-      <field type="uint8_t" name="increment" units="deg">Angular width in degrees of each array element. (Ignored if increment_f greater than 0).</field>
+      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the vehicle with index 0 corresponding to vehicle forward + angle_offset. A value of 0 is valid and means that the obstacle is practically touching the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
```
jkflying: IMO this isn't backwards compatible. Before it was local forward, which is the local frame, ie. North aligned. Now it is vehicle forward, which is body frame, which is not north aligned, but vehicle yaw aligned.

Unless I misunderstood what was meant by local forward? But this is how it is implemented in PX4 at least (North aligned).
@mrivi

mrivi: The original description before the last change was `Distance of obstacles around the UAV with index 0 corresponding to local North`. See diff https://github.com/mavlink/mavlink/pull/1142/files
PX4 implementation follows this definition. This change completely breaks the current implementation

hamishwillee: @jkflying @magicrub @mrivi Sorry; clearly I pushed the button too fast. 

So my understanding is that to revert this while still keeping the improvements., we would just need to change 

> Distance of obstacles around the vehicle with index 0 corresponding to vehicle forward + angle_offset

back to 

> Distance of obstacles around the vehicle with index 0 corresponding to **local** forward + angle_offset

Correct? We should do that then argue about body vs local.

jkflying: Correct. I think AP actually implemented it in body frame, though, so we need to figure out a path forward here.

Depending on how 'smart' the sensor is, it might be a lot easier to send messages in body frame. Maybe make an extension to specify that it is in body instead of local?

hamishwillee: I would imagine  it would be easier to display the collisions in body frame too. 

I will revert in https://github.com/mavlink/mavlink/pull/1213

Extension sounds good. Let's see what @magicrub says.

magicrub:
A sensor has no idea what North is so using North make no sense. How does px4 do that? How do you handle multiple sensors? With the offset it makes local frame much easier. The angle offset does not break compatibility.

jkflying:
In PX4 it typically isn't published directly by a sensor, but by the [Local Planner](https://github.com/PX4/avoidance). This does everything in local coordinates. Similarly, the flight controller's collision prevention operates in manual position mode which is in local coordinates, so velocities that get constrained all work directly in local coordinates too. Otherwise everything would have to go to body frame and back at each side.

I think the best way to make this work for everyone is to add an extension with a [MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME) enum. Then 'dumb' sensors can publish in body frame, 'smart' sensors/SLAM algorithms can publish in local frame, and the flight controller can translate into whatever it needs for its own internal representation.

hamishwillee:
Let's move discussion to #1214 (see problem there - "setting a default".

PR #1214, opened 2019-08-21
Add extension to OBSTACLE_DISTANCE specifying MAV_FRAME

jkflying:
This should resolve issues brought up in https://github.com/mavlink/mavlink/pull/1211

@hamishwillee @magicrub @mrivi

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5553,6 +5553,7 @@
       <extensions/>
       <field type="float" name="increment_f" units="deg">Angular width in degrees of each array element as a float. If non-zero then this value is used instead of the uint8_t increment field. Positive is clockwise direction, negative is counter-clockwise.</field>
       <field type="float" name="angle_offset" units="deg">Relative angle offset of the 0-index element in the distances array. Value of 0 corresponds to forward. Positive is clockwise direction, negative is counter-clockwise.</field>
+      <field type="uint8_t" name="frame_id" enum="MAV_FRAME">Coordinate frame of reference for the yaw rotation and offset of the sensor data. Typical choices are MAV_FRAME_LOCAL_NED and MAV_FRAME_BODY_FRD.</field>
     </message>
     <message id="331" name="ODOMETRY">
       <description>Odometry message to communicate odometry information with an external interface. Fits ROS REP 147 standard for aerial vehicles (http://www.ros.org/reps/rep-0147.html).</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5553,6 +5553,7 @@
       <extensions/>
       <field type="float" name="increment_f" units="deg">Angular width in degrees of each array element as a float. If non-zero then this value is used instead of the uint8_t increment field. Positive is clockwise direction, negative is counter-clockwise.</field>
       <field type="float" name="angle_offset" units="deg">Relative angle offset of the 0-index element in the distances array. Value of 0 corresponds to forward. Positive is clockwise direction, negative is counter-clockwise.</field>
+      <field type="uint8_t" name="frame_id" enum="MAV_FRAME">Coordinate frame of reference for the yaw rotation and offset of the sensor data. Typical choices are MAV_FRAME_LOCAL_NED and MAV_FRAME_BODY_FRD.</field>
```
hamishwillee: 1. I'd name it `frame` - that is used consistently except for in the ODOMETRY message. 
2. A reciever that gets this message/field from a system that has not been updated to use the extension will get "0" (MAV_FRAME_GLOBAL) so in effect the default is MAV_FRAME_GLOBAL. What does that mean?
3. In spect you can define a  numerical default for a field, but I've never seen it used because there is no way to reliably push that default through to generated libraries. Where possible default value is mapped to zero.

jkflying: 1. sure, will do
2. Since we're really concerned about the rotation offsets, not x/y coordinates, global has the same as local since both are North aligned. So this would be backwards compatible with the existing spec.
3. I think (2) solves this?

hamishwillee: Yes. Great.
I'm OK with this; @magicrub - does this work for you.

hamishwillee: @magicrub ^^^^

rmackay9: ok so the idea is that the frame could either be "body frame" or NE frame?  That seems reasonable to me although I haven't personally seen a device that cam provide it in NE frame.  I agree with Hamish that the field should just be called "frame" instead of "frame_id" because "frame" is consistent with existing messages (like this one: https://mavlink.io/en/messages/common.html#MISSION_ITEM)

jkflying: @rmackay9 The message was originally added by @mrivi for interfacing with SLAM systems, which usually hold their internal representations in NE. The idea back then was that distance sensors would be directly attached to the flight controller, rather than using MAVLink.

Since then this message has found good use for interfacing (multiple) external distance sensors, but this frame thing slipped through the cracks. So we're trying to make sure everyone is speaking the same language by adding this extension.

magicrub:
there's a merge conflict, hopefully it's just with https://github.com/mavlink/mavlink/pull/1211 .

hamishwillee:
Merged! Thanks all.

PR #1215, opened 2019-08-22
common.xml: Add FENCE_LIMIT to FENCE_BREACH enum

IamPete1:
This adds a new value to the FENCE_BREACH enum, 'FENCE_LIMIT' = 4. 

This is to report when the autopilot is having to take action to prevent a fence breach, typically by limiting the velocity component in the direction of the fence.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -956,6 +956,16 @@
         <description>Breached fence boundary</description>
       </entry>
     </enum>
+    <enum name="FENCE_MITIGATE">
+      <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Actions being taken to mitigate/prevent fence breach</description>
+      <entry value="0" name="FENCE_MITIGATE_NONE">
+        <description>No actions being taken</description>
+      </entry>
+      <entry value="1" name="FENCE_MITIGATE_VEL_LIMIT">
+        <description>Velocity limiting active to prevent breach</description>
+      </entry>
+    </enum>
     <!-- Camera Mount mode Enumeration -->
     <enum name="MAV_MOUNT_MODE">
       <description>Enumeration of possible mount operation modes</description>
@@ -5103,6 +5113,8 @@
       <field type="uint16_t" name="breach_count">Number of fence breaches.</field>
       <field type="uint8_t" name="breach_type" enum="FENCE_BREACH">Last breach type.</field>
       <field type="uint32_t" name="breach_time" units="ms">Time (since boot) of last breach.</field>
+      <extensions/>
+      <field type="uint8_t" name="breach_mitigation" enum="FENCE_MITIGATE">Active action to prevent fence breach</field>
     </message>
     <!-- MESSAGE IDs 180 - 229: Space for custom messages in individual projectname_messages.xml files -->
     <message id="230" name="ESTIMATOR_STATUS">


message_definitions/v1.0/common.xml
```diff
@@ -940,6 +940,9 @@
       <entry value="3" name="FENCE_BREACH_BOUNDARY">
         <description>Breached fence boundary</description>
       </entry>
+      <entry value="4" name="FENCE_LIMIT">
```
hamishwillee: @IamPete1 I don't think this is the right place - all the other values in this enum reflect that a breach has happened and what is was. In the message that uses it, the enum is described as "breach_type" - "Last breach type.". Can you add that "Last breach type" as an enum description as part of this?

Looking at the message that uses this, the only existing field that this might make sense in is `breach_status`. But even that won't be very "clean" since "actively limiting velocity" and "currently inside fence" match.
```
    <message id="162" name="FENCE_STATUS">
      <description>Status of geo-fencing. Sent in extended status stream when fencing enabled.</description>
      <field type="uint8_t" name="breach_status">Breach status (0 if currently inside fence, 1 if outside).</field>
      <field type="uint16_t" name="breach_count">Number of fence breaches.</field>
      <field type="uint8_t" name="breach_type" enum="FENCE_BREACH">Last breach type.</field>
      <field type="uint32_t" name="breach_time" units="ms">Time (since boot) of last breach.</field>
    </message>
```

So, how about an extension (something) like this:
```
      <extensions/>
      <field type="uint8_t" name="breach_mitigation" enum="FENCE_MITIGATE">Active action to prevent fence breach</field>
```

With enum
```
    <enum name="FENCE_MITIGATE">
       <description>Actions being taken to mitigate/prevent fence breach</description>
       <entry value="0" name="FENCE_MITIGATE_NONE">
         <description>No actions being taken</description>
      </entry>
       <entry value="0" name="FENCE_MITIGATE_LIMIT">
         <description>Velocity limiting active to prevent breach</description>
      </entry>
```

hamishwillee:
@IamPete1 Did my response make sense?

IamPete1:
@hamishwillee sorry i have been busy with other stuff, I had breach_status in mind. I think your extension makes sense, It fixed the issue I was having, namely the vehicle stopping for no apparent reason, this is not so bad for a polygon fence as it would typically be displayed by the GCS, but I got quite confused by a circle fence when the rover just stopped.

hamishwillee:
Thanks @IamPete1 ! Can you update the PR appropriately so we can get it discussed?

IamPete1:
I have updated this to the extension and enum suggested by @hamishwillee, thanks!, and rebased

hamishwillee:
Thanks @IamPete1 - looks good to me.

@auturgy @amilcarlucas @MaEtUgR Any comments?

message_definitions/v1.0/common.xml
```diff
@@ -956,6 +956,17 @@
         <description>Breached fence boundary</description>
       </entry>
     </enum>
+    <enum name="FENCE_MITIGATE">
+      <wip/>
```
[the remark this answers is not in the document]
IamPete1: fixed, its now passing travis

hamishwillee:
Thanks @IamPete1 . IN the absence of other comments, and because this looks like a good idea to me. Merging.

message_definitions/v1.0/common.xml
```diff
@@ -956,6 +956,16 @@
         <description>Breached fence boundary</description>
       </entry>
     </enum>
+    <enum name="FENCE_MITIGATE">
+      <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
```
hamishwillee: @IamPete1 Can you please let me know when you have some working code for this. Then we'll know this fix was OK and can remove the WIP note. Though I guess a little pointless having it really given the field addition is in.

Issue #898, opened 2018-05-01
MAVLink set_home_position in local frame coordinate system


alexx600:
Hello,
I would like to ask if it is possible to change home position using local NED x,y not global lattitude, longitude coordinate?

According to http://mavlink.org/messages/common#SET_HOME_POSITION “The global and local positions encode the position in the respective coordinate frames”.
What shoud i send as a lattitude and longitute if i would like to move home position in local frame (for example 2 m forwad)?
I tried set lattituede and longitude to zero, but this caused reseting home position. Is it possible to change home position in local system?

shubham9436:
Commenting to get the notification

julianoes:
Is this a question for the MAVLink protocol or PX4 or ArduPilot?

hamishwillee:
@julianoes FWIW I don't think PX4 handles this message at all. But that said, it isn't clear (to me) how this should work - and that has to be clarified.

What are X, Y, Z? I think they are NED of home position relative to current vehicle position - correct? In that case we would minimally need to clarify whether Z is positive up.

Can you just specify X, Y, Z or just lat, lon? It looks like you have to specify everything, which makes this a bit useless for indoor use.

Last of all, this should note that the value of HOME can be changed underneath you by the system.

hamishwillee:
@julianoes  How do we progress this. Shubham appears to have lost interest.

I would still quite like to understand how the message is used: https://github.com/mavlink/mavlink/issues/898#issuecomment-514027281 - 
IN particular, related to local frames, and whether both frames need to be set (unlikely). Local frame is particularly unclear because I don't know "what" local frame - quite a lot have been defined e.g. https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_FRD. Perhaps it doesn't matter, or that is something that is defined elsewhere?

julianoes:
Wouldn't this position handled like any other position with frame in `COMMAND_INT`?

hamishwillee:
COMMAND_INT allows you to specify a frame; this doesn't. Or are you saying something else?

julianoes:
Oh, nevermind. Could we add the frame as an extension field?

hamishwillee:
> Oh, nevermind. Could we add the frame as an extension field?

Yes, and I have done that in https://github.com/mavlink/mavlink/pull/1216

That still doesn't fix everything though. I added my additional question in above PR.

julianoes:
[...]

PR #1216, opened 2019-08-23
SET_POSITION_HOME: Add frame for local coordinates

hamishwillee:
This is an attempt to clarify how [SET_POSITION_HOME](https://mavlink.io/en/messages/common.html#SET_HOME_POSITION) should work, as discussed in 
https://github.com/mavlink/mavlink/issues/898#issuecomment-514027281

The (first) specific fix is to add field to specify frame for local coordinate values - since this could be any frame. [Note, the intention could be that a system has the concept of a default frame - effectively I opened this to find out what the thinking is]

**Question**: what should user do if they don't want to specify a local frame or global frame. Is it that if you're working in global system (ie outside) the global home will be used, and if you're working in a local system (e.g. from a VIO system) that the local system will be used - so you can and should specify both, and the irrelevant one will be ignored?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5234,6 +5234,7 @@
       <field type="float" name="approach_z" units="m">Local Z position of the end of the approach vector. Multicopters should set this position based on their takeoff path. Grass-landing fixed wing aircraft should set it the same way as multicopters. Runway-landing fixed wing aircraft should set it to the opposite direction of the takeoff, assuming the takeoff happened from the threshold / touchdown zone.</field>
       <extensions/>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the local frame fields: x, y, z.</field>
     </message>
     <message id="244" name="MESSAGE_INTERVAL">
       <description>The interval between messages for a particular MAVLink message ID. This message is the response to the MAV_CMD_GET_MESSAGE_INTERVAL command. This interface replaces DATA_STREAM.</description>


hamishwillee:
Reposted from @julianoes 

> Probably only one global and one local frame will be supported at first. If this was a command you could just nack the request with UNSUPPORTED. By having this as a separate message, I'm not sure how it will work.

The global frame isn't a problem because the docs hardcode that (e.g. to WGS84). 

Good point re the local frame. It almost guarantees that the assumption is that a system has only one local frame that is a "default" or they are assuming only one.

hamishwillee:
@TSC21 @auturgy Do either of you know anything about [SET_POSITION_HOME](https://mavlink.io/en/messages/common.html#SET_HOME_POSITION). 

It doesn't appear to be supported on PX4 or ArduPilot and I can't see how it would work - in particular with respect to the local frame - what local frame is expected to mean.

TSC21:
@hamishwillee I think that the use should just send the LLA or the XYZ. Sending the LLA is quite trivial to handle on the autopilot side. XYZ though need a reprojection depending on the current local position. So if the the home position was set at a certain LLA, if you send XYZ, then it means it needs to reproject the new home position depending the the vehicle current position.

I don't think we need a field to set up the frame, but rather make sure that if we send LLA, we use it, if we send XYZ we also use it, though requiring a reprojection. Sending both LLA and XYZ would represent an issue IMHO because the LLA we are setting may not match the XYZ we are setting since the origin is not known at hand (only the autopilot know it). I would rather make the choice between LLA and XYZ exclusive: if we set one, we can't set the other.

hamishwillee:
@TSC21 Thanks for taking the time to reply. I'm not very familiar with frames/frame conversions, so please excuse probably stupid clarifications. 

So in summary you would either specify the global or XYZ, but not both. So what needs to change is a note to that effect in the description, and also a note in each of the fields to indicate how they are marked as "invalid" - e.g. for floats NaN and for Int32 a max value. Correct?

You're also indicating we do not add a frame from XYZ. 
I still don't really get "why"? Generally I imagine a scenario where I'm flying around inside, with no GPS, and an existing local frame (with 0,0,0 known by the autopilot) and possibly and existing home position. I then decide to set a new home position where I am currently. So I get the current position - how? I don't know the local frame that is being used, then I set it using this message?

hamishwillee:
This has gone nowhere, and no one seems interested. I guess it is not a problem. Closing.

Issue #1220, opened 2019-08-26
Pl explain how "payload length is defined by remaining message length" is working


olliw42:
messages such as [V2_EXTENSION](https://mavlink.io/en/messages/common.html#V2_EXTENSION) and the recently added [TUNNEL](https://mavlink.io/en/messages/common.html#TUNNEL) carry opaque data in a payload field of variable length, and as regards the length determination it is said that 

> Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields.

However, I really don't see and/or can't figure out how this should or can work, and hence would highly appreciate an explanation of how to do this in practice.

The issue - IMHO - and which I don't get around without fooling - is that payloads may have valid zeros at the end, but the MAVLink2's empty byte payload truncation chops them off and modifies the message size length accordingly. I can't see any reasonable method to retrieve the original valid payload size from the received message. So, I argue, the procedure as described in the spec fails, and a payload_length data field is needed to completely and correctly retrieve the intended payload.

I note that in #1194 I had suspicion that this might not work and a payload_length field be needed, but assumed by the precedence of V2_EXTENSION and that it wasn't commented on further that I'm wrong and it's not needed. Maybe my fault if I wasn't insisting enough in this point.

thx for any insight.

hamishwillee:
I had nothing to do with defining either of these messages so I could be wrong about the intention but ....
if the payload format is opaque (entirely up to you), then you can define that it must terminate with some marker that isn't zero to prevent truncation, or you can include a payload length as part of your private protocol.

olliw42:
sure, there are workarounds (I'm actually doing your first method)

BUT

1. The statements as such are - IMHO - simply incorrect, and should (must) thus - IMHO - be corrected (independent on any workaround).

2. The payload length is - IMHO - quite different to things like a sequence number and such since there is no useful application of the message without knowing the payload length, that is ANY application will have to have this information! So I consider it a very basic requirement for the working of the message, which I think is justification to add such a field outside of the opaque structure. 

You also can look at it this way: Anyone will have to consume one byte of the payload for that purpose, and it thus doesn't make much sense to not put it outside and canonicalize the payload length handling.

as I said, it was probably my fault that I wasn't more pressing this point and more insistingly expressing my concerns that the described method might not actually work.

hamishwillee:
I have no strong opinion. Suggest you create a PR that does what you want and we can review.

olliw42:
[...]

PR #1222, opened 2019-08-29
Tunnel message: add length field

olliw42:
addresses #1220

EDITED (HamishW). The tunnel message is WIP, and Ollie is the person who championed it and is first implementer. This change makes sense and is not affecting stable messages.

EDIT (OlliW): Thx for the edit! I should have realized that it's important info that changing the message now won't do damage.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5727,7 +5727,8 @@
       <field type="uint8_t" name="target_system">System ID (can be 0 for broadcast, but this is discouraged)</field>
       <field type="uint8_t" name="target_component">Component ID (can be 0 for broadcast, but this is discouraged)</field>
       <field type="uint16_t" name="payload_type" enum="MAV_TUNNEL_PAYLOAD_TYPE">A code that identifies the content of the payload (0 for unknown, which is the default). If this code is less than 32768, it is a 'registered' payload type and the corresponding code should be added to the MAV_TUNNEL_PAYLOAD_TYPE enum, and the entry possibly to https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. Software creators can register blocks of types as needed. Codes greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
-      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
+      <field type="uint8_t" name="payload_len">Length of the data transported in payload</field>
+      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by payload_len. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
     </message>
     <message id="390" name="ONBOARD_COMPUTER_STATUS">
       <wip/>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5727,7 +5727,8 @@
...
-      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
+      <field type="uint8_t" name="payload_len">Length of the data transported in payload</field>
+      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by payload_len. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
```
hamishwillee: Is it worth revisiting the length of this array at the same time - e.g. making it max available?

message_definitions/v1.0/common.xml
```diff
@@ -5727,7 +5727,8 @@
...
       <field type="uint8_t" name="target_component">Component ID (can be 0 for broadcast, but this is discouraged)</field>
       <field type="uint16_t" name="payload_type" enum="MAV_TUNNEL_PAYLOAD_TYPE">A code that identifies the content of the payload (0 for unknown, which is the default). If this code is less than 32768, it is a 'registered' payload type and the corresponding code should be added to the MAV_TUNNEL_PAYLOAD_TYPE enum, and the entry possibly to https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. Software creators can register blocks of types as needed. Codes greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
-      <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
+      <field type="uint8_t" name="payload_len">Length of the data transported in payload</field>
```
hamishwillee: I'd slightly prefer `payload_length` as name.

message_definitions/v1.0/common.xml
olliw42: I personally (still) don't see a need for it
and for the same argument as before of keeping resources lower I (still) prefer 128 bytes max

message_definitions/v1.0/common.xml
olliw42: yeah ... I checked with the other messages which have a length/size field, and the naming isn't really consistent ... payload_length is fine to me, would you maybe prefer payload_size? I really have no opinion.

message_definitions/v1.0/common.xml
mhkabir: What is the argument behind not making it the max available? In practice, won't the truncation cut out whatever is zero at the end?

message_definitions/v1.0/common.xml
olliw42: ok, so changed to payload_length :)

message_definitions/v1.0/common.xml
hamishwillee: @mhkabir Truncation affects what is sent OTA, but not the resources allocated to handle the message either end. Ollie is working on gimbals, so smaller is better for him.

hamishwillee:
Thanks @olliw42 . Merged.

PR #1224, opened 2019-09-01
FOCUS_TYPE_RANGE: Typo

hamishwillee:
@dogmaphobic:
I think this was a typo. Can you confirm?

In addition, in https://groups.google.com/forum/#!msg/mavlink/9KrPARAMe1c/LRW6IPBHBAAJ there is a question:

> I would like to pass the focus distance in meters to the camera. But would this be focus type number 2, or maybe a new one?

My take is that we don't support this. Do we want to?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3300,7 +3300,7 @@
         <description>Continuous focus up/down until stopped (-1 for focusing in, 1 for focusing out towards infinity, 0 to stop focusing)</description>
       </entry>
       <entry value="2" name="FOCUS_TYPE_RANGE">
-        <description>Zoom value as proportion of full camera range (a value between 0.0 and 100.0)</description>
+        <description>Focus value as proportion of full camera focus range (a value between 0.0 and 100.0)</description>
       </entry>
     </enum>
     <enum name="PARAM_ACK">


dogmaphobic:
Yes, it's a typo. The result of copying and pasting it from the zoom command. It's not that we don't support it, this has never been implemented and I'm not aware of any camera that allows you to define the focus through an API (in meters, percent of focus range, etc.) Meaning, if he's talking about a specific camera that does accept setting the focus in meters, we can either change the description of the message to follow that camera (it makes a whole lot more sense to set the focus distance rather than range percentage) or add something that allows such a thing. The only camera we support focus with is really, really dumb. It only has focus + and - with no indication of the actual focus position.

hamishwillee:
@dogmaphobic Thanks. I'll give Jeroen Tavernea few days to come back to us with information about the camera. If not, then I'll just merge to remove the typo.

hamishwillee:
I'm merging this. We can revisit if we get more feedback.

jeroentaverne:
Thanks.
The focus distance of the camera hardware can be set in milimeters. As the MAV_CMD_SET_CAMERA_FOCUS uses float, it might be good to use meters for the communication.

hamishwillee:
@dogmaphobic So @jeroentaverne got back to us above. The camera is a "Phase one" (I'm guessing this thing https://www.phaseone.com/en/) which will take a focus distance in milimetres.

Propose we just add a new FOCUS_TYPE: `FOCUS_TYPE_METRES` | Focus value in metres. 
Do we need to add some value for infinity? Ie -1 = infinity?










https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/mavlink/9KrPARAMe1c/JoQJcBcLAgAJ

dogmaphobic:
>Would it make sense to add or is it just not worth it given you can do this in the definition file?

Good question. In cases where there are specific MAVLink commands to change a value, it would be helpful to have a way to get the range of valid values that can be set. This applies to Zoom, Focus, and maybe others. The new gimbal interface has the same issue. How do you know the valid pitch/yaw range?

We need however, to make sure to not try to fit everything in these messages. Things get specialized and full of exceptions very quickly and the only way to handle that is through definition files and parameters. For example what if the zoom range is X-Y when in Photo Mode and X-Z when in Video Mode? You get the idea.

>Devils advocate - we add the values just so we have a good place to document that you actually can't use them without knowing the ranges.

Yes, even though QGC cannot use it, we should not prevent custom implementations, ones that know exactly what the camera does, from working.

hamishwillee:
Thanks @dogmaphobic - I created PR for your review in https://github.com/mavlink/mavlink/pull/1235 of JUST the enums for the API as it is now.

PR #1226, opened 2019-09-02
V2_EXTENSION: Correct payload length determination

olliw42:
as regards the determination of the payload length it currently says 

> Variable length payload. The payload length is defined by the remaining message length when subtracting the header and other fields.

which for the reasons discussed in #1220 doesn't work. This PR corrects the description, by also mentioning two methods suggested in #1220.


message_definitions/v1.0/common.xml
```diff
@@ -5291,8 +5291,8 @@
...
+      <field type="uint16_t" name="message_type">A code that identifies the software component that understands this message (analogous to USB device classes or mime type strings). If this code is less than 32768, it is considered a 'registered' protocol extension and the corresponding entry should be added to https://github.com/mavlink/mavlink/extension-message-ids.xml. Software creators can register blocks of message IDs as needed (useful for GCS specific metadata, etc...). Message_types greater than 32767 are considered local experiments and should not be checked in to any widely distributed codebase.</field>
+      <field type="uint8_t[249]" name="payload">Variable length payload. The length must be handled separately, e.g. by terminating the data in payload with some marker that is not zero to prevent truncation, or by including the payload length as part of the payload. The entire content of this block is opaque unless you understand any the encoding message_type. The particular encoding used can be extension specific and might not always be documented as part of the mavlink specification.</field>
```
hamishwillee: Thanks for this. Proposal is to explain the problem with truncation in a little more detail. How about:

> Variable length payload. The length must be handled separately, e.g. by terminating the data in payload with some marker that is not zero to prevent truncation, or by including the payload length as part of the payload.

TO

> Variable length payload. The length must be encoded in the payload as part of the message_type protocol, e.g. by including the length as payload data, or by terminating the payload data with a non-zero marker. This is required in order to reconstruct zero-terminated payloads that are (or otherwise would be) trimmed by MAVLink 2 empty-byte truncation.

olliw42: fantastic. That's why you are the writer and I'm just the geek. Adopted. Plus some few text/spelling.

hamishwillee: Well, we're both a little bit pedantic, so it's a happy ending :-)

olliw42: LOL 
you really made me laugh now ... great start of a day, I'd say :)

hamishwillee:
@auturgy FYI, this got discussed before and not sure if you saw it. 

This is a description/semantic _clarification_ to the V2_EXTENSION message. 

Essentially if payload data could validly end in zeros they will be trimmed by empty byte truncation, and there is no way to reconstruct the payload unless either the truncation is prevented at the end of the data or the payload includes the length information.

PR #1228, opened 2019-09-04
common.xml: add AIS message

IamPete1:
This adds a AIS message and the relevant enums. This message is an attempt to cover the two most common AIS messages. Position reports and Static and Voyage Related Data, see: https://gpsd.gitlab.io/gpsd/AIVDM.html 

I have attempted to only include the information from these messages that would be useful, there is quite a bit left out, such as vessel destination and ETA. I have, for the most part left the values in the units they arrive in over AIS, ie knots not m/s (could easily be persuaded to change this). I have added bits to the flags enum for magic values. For example in AIS turn rate is quite complex: 
- 0 = not turning
- 1…126 = turning right at up to 708 degrees per minute or higher
- 1…-126 = turning left at up to 708 degrees per minute or higher
- 127 = turning right at more than 5deg/30s (No TI available)
- -127 = turning left at more than 5deg/30s (No TI available)
- 128 (80 hex) indicates no turn information available (default)

I have replaced the magic values with bits in the AIS_FLAGS bit-mask. 

Also the type and navigation status enums are copyed directly from the AIS standards, not sure if there is a better way?

I have also added a nmea pass-through message, this passes the NMEA sentence as received. Note that they are a maximum of 74 characters long, could be shorter, I'm not sure if i should have a length field also?

The idea of this is that we only partly cover 2 AIS messages out of the 27 with our dedicated message. This allows other messages to be passed through if desired, in quite a lot of cases they would be messages that happen very infrequently, for example you can broadcast plain text messages or get weather reports. There is even a stay-out zone message. It seems silly to duplicate all these in mavlink so i think its better to just allow pass-through directly. This extends to all NMEA stuff, not just AIS

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3834,6 +3834,182 @@
         <description>Format is Modern Music Markup Language (MML): https://en.wikipedia.org/wiki/Music_Macro_Language#Modern_MML.</description>
       </entry>
     </enum>
+    <!-- AIS related enums-->
+    <enum name="AIS_TYPE">
+      <description>Type of AIS vessel, enum duplicated from AIS standard, https://gpsd.gitlab.io/gpsd/AIVDM.html</description>
+      <entry value="0" name="AIS_TYPE_UNKNOWN">
+        <description>Not available (default).</description>
+      </entry>
+      <entry value="1" name="AIS_TYPE_RESERVED_1"/>
+      <entry value="2" name="AIS_TYPE_RESERVED_2"/>
+      <entry value="3" name="AIS_TYPE_RESERVED_3"/>
+      <entry value="4" name="AIS_TYPE_RESERVED_4"/>
+      <entry value="5" name="AIS_TYPE_RESERVED_5"/>
+      <entry value="6" name="AIS_TYPE_RESERVED_6"/>
+      <entry value="7" name="AIS_TYPE_RESERVED_7"/>
+      <entry value="8" name="AIS_TYPE_RESERVED_8"/>
+      <entry value="9" name="AIS_TYPE_RESERVED_9"/>
+      <entry value="10" name="AIS_TYPE_RESERVED_10"/>
+      <entry value="11" name="AIS_TYPE_RESERVED_11"/>
+      <entry value="12" name="AIS_TYPE_RESERVED_12"/>
+      <entry value="13" name="AIS_TYPE_RESERVED_13"/>
+      <entry value="14" name="AIS_TYPE_RESERVED_14"/>
+      <entry value="15" name="AIS_TYPE_RESERVED_15"/>
+      <entry value="16" name="AIS_TYPE_RESERVED_16"/>
+      <entry value="17" name="AIS_TYPE_RESERVED_17"/>
+      <entry value="18" name="AIS_TYPE_RESERVED_18"/>
+      <entry value="19" name="AIS_TYPE_RESERVED_19"/>
+      <entry value="20" name="AIS_TYPE_WIG">
+        <description>Wing In Ground effect.</description>
+      </entry>
+      <entry value="21" name="AIS_TYPE_WIG_HAZARDOUS_A"/>
+      <entry value="22" name="AIS_TYPE_WIG_HAZARDOUS_B"/>
+      <entry value="23" name="AIS_TYPE_WIG_HAZARDOUS_C"/>
+      <entry value="24" name="AIS_TYPE_WIG_HAZARDOUS_D"/>
+      <entry value="25" name="AIS_TYPE_WIG_RESERVED_1"/>
+      <entry value="26" name="AIS_TYPE_WIG_RESERVED_2"/>
+      <entry value="27" name="AIS_TYPE_WIG_RESERVED_3"/>
+      <entry value="28" name="AIS_TYPE_WIG_RESERVED_4"/>
+      <entry value="29" name="AIS_TYPE_WIG_RESERVED_5"/>
+      <entry value="30" name="AIS_TYPE_FISHING"/>
+      <entry value="31" name="AIS_TYPE_TOWING"/>
+      <entry value="32" name="AIS_TYPE_TOWING_LARGE">
+        <description>Towing: length exceeds 200m or breadth exceeds 25m.</description>
+      </entry>
+      <entry value="33" name="AIS_TYPE_DREDGING">
+        <description>Dredging or other underwater ops.</description>
+      </entry>
+      <entry value="34" name="AIS_TYPE_DIVING"/>
+      <entry value="35" name="AIS_TYPE_MILITARY"/>
+      <entry value="36" name="AIS_TYPE_SAILING"/>
+      <entry value="37" name="AIS_TYPE_PLEASURE"/>
+      <entry value="38" name="AIS_TYPE_RESERVED_20"/>
+      <entry value="39" name="AIS_TYPE_RESERVED_21"/>
+      <entry value="40" name="AIS_TYPE_HSC">
+        <description>High Speed Craft.</description>
+      </entry>
+      <entry value="41" name="AIS_TYPE_HSC_HAZARDOUS_A"/>
+      <entry value="42" name="AIS_TYPE_HSC_HAZARDOUS_B"/>
+      <entry value="43" name="AIS_TYPE_HSC_HAZARDOUS_C"/>
+      <entry value="44" name="AIS_TYPE_HSC_HAZARDOUS_D"/>
+      <entry value="45" name="AIS_TYPE_HSC_RESERVED_1"/>
+      <entry value="46" name="AIS_TYPE_HSC_RESERVED_2"/>
+      <entry value="47" name="AIS_TYPE_HSC_RESERVED_3"/>
+      <entry value="48" name="AIS_TYPE_HSC_RESERVED_4"/>
+      <entry value="49" name="AIS_TYPE_HSC_UNKNOWN"/>
+      <entry value="50" name="AIS_TYPE_PILOT"/>
+      <entry value="51" name="AIS_TYPE_SAR">
+        <description>Search And Rescue vessel.</description>
+      </entry>
+      <entry value="52" name="AIS_TYPE_TUG"/>
+      <entry value="53" name="AIS_TYPE_PORT_TENDER"/>
+      <entry value="54" name="AIS_TYPE_ANTI_POLLUTION">
+        <description>Anti-pollution equipment.</description>
+      </entry>
+      <entry value="55" name="AIS_TYPE_LAW_ENFORCEMENT"/>
+      <entry value="56" name="AIS_TYPE_SPARE_LOCAL_1"/>
+      <entry value="57" name="AIS_TYPE_SPARE_LOCAL_2"/>
+      <entry value="58" name="AIS_TYPE_MEDICAL_TRANSPORT"/>
+      <entry value="59" name="AIS_TYPE_NONECOMBATANT">
+        <description>Noncombatant ship according to RR Resolution No. 18.</description>
+      </entry>
+      <entry value="60" name="AIS_TYPE_PASSENGER"/>
+      <entry value="61" name="AIS_TYPE_PASSENGER_HAZARDOUS_A"/>
+      <entry value="62" name="AIS_TYPE_PASSENGER_HAZARDOUS_B"/>
+      <entry value="63" name="AIS_TYPE_AIS_TYPE_PASSENGER_HAZARDOUS_C"/>
+      <entry value="64" name="AIS_TYPE_PASSENGER_HAZARDOUS_D"/>
+      <entry value="65" name="AIS_TYPE_PASSENGER_RESERVED_1"/>
+      <entry value="66" name="AIS_TYPE_PASSENGER_RESERVED_2"/>
+      <entry value="67" name="AIS_TYPE_PASSENGER_RESERVED_3"/>
+      <entry value="68" name="AIS_TYPE_AIS_TYPE_PASSENGER_RESERVED_4"/>
+      <entry value="69" name="AIS_TYPE_PASSENGER_UNKNOWN"/>
+      <entry value="70" name="AIS_TYPE_CARGO"/>
+      <entry value="71" name="AIS_TYPE_CARGO_HAZARDOUS_A"/>
+      <entry value="72" name="AIS_TYPE_CARGO_HAZARDOUS_B"/>
+      <entry value="73" name="AIS_TYPE_CARGO_HAZARDOUS_C"/>
+      <entry value="74" name="AIS_TYPE_CARGO_HAZARDOUS_D"/>
+      <entry value="75" name="AIS_TYPE_CARGO_RESERVED_1"/>
+      <entry value="76" name="AIS_TYPE_CARGO_RESERVED_2"/>
+      <entry value="77" name="AIS_TYPE_CARGO_RESERVED_3"/>
+      <entry value="78" name="AIS_TYPE_CARGO_RESERVED_4"/>
+      <entry value="79" name="AIS_TYPE_CARGO_UNKNOWN"/>
+      <entry value="80" name="AIS_TYPE_TANKER"/>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -3742,6 +3742,181 @@
...
+      <entry value="14" name="AIS_TYPE_RESERVED_14"></entry>
+      <entry value="15" name="AIS_TYPE_RESERVED_15"></entry>
+      <entry value="16" name="AIS_TYPE_RESERVED_16"></entry>
+      <entry value="17" name="AIS_TYPE_RESERVED_17"></entry>
+      <entry value="18" name="AIS_TYPE_RESERVED_18"></entry>
+      <entry value="19" name="AIS_TYPE_RESERVED_19"></entry>
+      <entry value="20" name="AIS_TYPE_WIG">
+        <description>Wing In Ground effect.</description>
+      </entry>
+      <entry value="21" name="AIS_TYPE_WIG_HAZARDOUS_A"></entry>
```
hamishwillee: FWIW I prefer descriptions for all enums, with possible exception of reserved.

message_definitions/v1.0/common.xml
```diff
@@ -3742,6 +3742,181 @@
...
+      <entry value="14" name="AIS_NAV_AIS_SART">
+        <description>Search And Rescue Transponder.</description>
+      </entry>
+      <entry value="15" name="AIS_NAV_UNKNOWN">
+        <description>Not available (default).</description>
+      </entry>
+    </enum>
+    <enum name="AIS_FLAGS">
+      <description>These flags indicate status such as data validity of each data source. Set = data valid</description>
+      <entry value="1" name="AIS_FLAGS_POSSION_ACCURACY">
```
hamishwillee: What is this? Position (description looks like typ too)

message_definitions/v1.0/common.xml
```diff
@@ -5507,6 +5682,30 @@
...
+      <field type="uint32_t" name="MMSI">Mobile Marine Service Identifier, 9 decimal digits</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude</field>
+      <field type="uint16_t" name="COG" units="cdeg">Course over ground</field>
+      <field type="uint16_t" name="heading" units="cdeg">true heading</field>
+      <field type="uint16_t" name="velocity" units="cKnots">Speed over ground</field>
+      <field type="int8_t" name="turn_rate" units="cdeg/s">turn rate</field>
+      <field type="uint8_t" name="navigational_status" enum="AIS_NAV_STATUS">Navigational status</field>
+      <field type="uint8_t" name="type" enum="AIS_TYPE">Type of vessels</field>
+      <field type="uint16_t" name="dimension_Bow" units="m">Distance from location to bow</field>
```
hamishwillee: inconsistent case on name

message_definitions/v1.0/common.xml
```diff
@@ -5507,6 +5682,30 @@
...
+      <field type="uint16_t" name="dimension_Stern" units="m">Distance from location to stern</field>
```
hamishwillee: inconsistent case on name

message_definitions/v1.0/common.xml
```diff
@@ -5507,6 +5682,30 @@
       <field type="uint8_t[8]" name="spec_version_hash">The first 8 bytes (not characters printed in hex!) of the git hash.</field>
       <field type="uint8_t[8]" name="library_version_hash">The first 8 bytes (not characters printed in hex!) of the git hash.</field>
     </message>
+    <message id="301" name="AIS_VESSEL">
+      <description>The location and information of an AIS vessel</description>
+      <field type="uint32_t" name="MMSI">Mobile Marine Service Identifier, 9 decimal digits</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude</field>
+      <field type="uint16_t" name="COG" units="cdeg">Course over ground</field>
+      <field type="uint16_t" name="heading" units="cdeg">true heading</field>
```
hamishwillee: Inconsistent capitalisation ("true")

message_definitions/v1.0/common.xml
```diff
@@ -5507,6 +5682,30 @@
...
+      <field type="uint16_t" name="velocity" units="cKnots">Speed over ground</field>
+      <field type="int8_t" name="turn_rate" units="cdeg/s">turn rate</field>
+      <field type="uint8_t" name="navigational_status" enum="AIS_NAV_STATUS">Navigational status</field>
+      <field type="uint8_t" name="type" enum="AIS_TYPE">Type of vessels</field>
```
hamishwillee: Type of vessel (not vessels)

message_definitions/v1.0/common.xml
```diff
@@ -3742,6 +3742,181 @@
         <description>The location of the remote pilot is a fixed location.</description>
       </entry>
     </enum>
+    <!-- AIS related enums-->
+    <enum name="AIS_TYPE">
```
hamishwillee: Suggest add [wip](https://mavlink.io/en/guide/xml_schema.html#wip) to messages and enums. Unless these aren't going to change.

message_definitions/v1.0/common.xml
hamishwillee: When you say "location" you mean values of lat/lon fields?

message_definitions/v1.0/common.xml
```diff
@@ -3742,6 +3742,181 @@
...
+      <entry value="13" name="AIS_NAV_RESERVED_3"></entry>
+      <entry value="14" name="AIS_NAV_AIS_SART">
+        <description>Search And Rescue Transponder.</description>
+      </entry>
+      <entry value="15" name="AIS_NAV_UNKNOWN">
+        <description>Not available (default).</description>
+      </entry>
+    </enum>
+    <enum name="AIS_FLAGS">
+      <description>These flags indicate status such as data validity of each data source. Set = data valid</description>
```
hamishwillee: Are they really used for anything other than data validity?

Perhaps

> These flags are used in the AIS_VESSEL.fields bitmask to indicate validity of data in the other message fields. When set, the data is valid.

message_definitions/v1.0/common.xml
```diff
@@ -3742,6 +3742,181 @@
...
+      <entry value="8" name="AIS_FLAGS_HIGH_VELOCITY">
+        <description>1 = velocity over 102.2 knots</description>
+      </entry>
+      <entry value="16" name="AIS_FLAGS_VALID_TURN_RATE"></entry>
+      <entry value="32" name="AIS_FLAGS_TURN_RATE_SIGN_ONLY">
+        <description>the only the sign of the returned turn rate value is valid, either greather than 5deg/30s or less than -5deg/30s</description>
+      </entry>
+      <entry value="64" name="AIS_FLAGS_VALID_DIMENSIONS"></entry>
+      <entry value="128" name="AIS_FLAGS_LARGE_BOW_DIMENSION">
+        <description>distance to bow is larger than 511m</description>
```
hamishwillee: For all, perhaps Distance to ... (in any case, consistency across messages and enums)

message_definitions/v1.0/common.xml
```diff
@@ -5507,6 +5682,30 @@
...
+      <field type="uint16_t" name="dimension_Bow" units="m">Distance from location to bow</field>
+      <field type="uint16_t" name="dimension_Stern" units="m">Distance from location to stern</field>
+      <field type="uint8_t" name="dimension_port" units="m">Distance from location to port side</field>
+      <field type="uint8_t" name="dimension_starboard" units="m">Distance from location to starboard side</field>
+      <field type="char[7]" name="callsign">The vessel callsign</field>
+      <field type="char[20]" name="name">The vessel name</field>
+      <field type="uint16_t" name="tslc" units="s">Time since last communication in seconds</field>
+      <field type="uint16_t" name="flags" enum="AIS_FLAGS" display="bitmask">Bitmask to indicate various statuses including valid data fields</field>
+    </message>
+    <message id="302" name="NMEA_PASSTHROUGH">
```
hamishwillee: Can you explain how this will be used?

In addition, can we assume that no strings will be terminated with valid "0" data? Ie how is the string terminated?

message_definitions/v1.0/common.xml
```diff
@@ -5507,6 +5682,30 @@
...
       <field type="uint8_t[8]" name="library_version_hash">The first 8 bytes (not characters printed in hex!) of the git hash.</field>
     </message>
+    <message id="301" name="AIS_VESSEL">
+      <description>The location and information of an AIS vessel</description>
+      <field type="uint32_t" name="MMSI">Mobile Marine Service Identifier, 9 decimal digits</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude</field>
+      <field type="uint16_t" name="COG" units="cdeg">Course over ground</field>
+      <field type="uint16_t" name="heading" units="cdeg">true heading</field>
+      <field type="uint16_t" name="velocity" units="cKnots">Speed over ground</field>
```
hamishwillee: This won't validate because Knots/cKnots are not part of the schema: https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd

_Generally_ we use SI units where possible.

hamishwillee:
@IamPete1 Thanks for this. I know nothing about matters nautical, so deferring to @auturgy 

A few general comments though, speaking from ignorance of the domain.

FWIW  I think the approach looks reasonable. I'm not certain on the passthrough. That's a hole that makes sense if you really just want to pass through systems. If you want to use the data it usually makes sense to have a message. 

> I have attempted to only include the information from these messages that would be useful, there is quite a bit left out, such as vessel destination and ETA. 

That information sounds useful for me. E.g. to calculate likely collisions.

> I have, for the most part left the values in the units they arrive in over AIS, ie knots not m/s (could easily be persuaded to change this).

If you want knots they will need to be added to the schema. I prefer SI units.

auturgy:
Re knots: it’s standard nautical speed, and what is provided by AIS, but as the consumer will almost always need to convert anyway, it doesn’t really matter. Unfortunately there will be a loss of precision somewhere in any case.
The NMEA0183 pass through: I’m on the fence with that, as degrading mavlink to a transport tunnel isn’t really in the spirit of standardisation, but is already done elsewhere. For prototyping it does have significant benefits. If it’s a problem for common.xml we can look at Ardupilotmega initially and perhaps pull it up later. I’d suggest breaking that commit out into a separate PR so that we can talk around in circles for a month or two without holding AIS up.

Re destination and eta: these don’t directly relate to position, course and speed, so aren’t as useful for avoidance as might be assumed. @iampete has made the right call on that imo.

Regards,

James

hamishwillee:
@auturgy 

1. Sure re destination/eta. Thanks.
2. My take is that we do convert to SI units. I'm guessing that this will be easier for calculations etc and presentation in GCS.
3. I'm a little opposed to passthrough messages "just in case". Could an alternative like https://mavlink.io/en/messages/common.html#V2_EXTENSION  be used in the meantime and it only be created if it proves useful.

message_definitions/v1.0/common.xml
IamPete1: yep, typo, fixed

message_definitions/v1.0/common.xml
IamPete1: yeah, i have added this to the description

message_definitions/v1.0/common.xml
IamPete1: fixed

message_definitions/v1.0/common.xml
IamPete1: fixed

message_definitions/v1.0/common.xml
IamPete1: I have added wip to the message and flags enum, the AIS_TYPE and AIS_NAV_STATUS are direct copy's of the AIS enums so should not change

message_definitions/v1.0/common.xml
IamPete1: fixed

message_definitions/v1.0/common.xml
IamPete1: fixed

message_definitions/v1.0/common.xml
IamPete1: I have split this bit to a new PR

message_definitions/v1.0/common.xml
IamPete1: changed to cm/s

IamPete1:
I have corrected lots of typos and dropped the NMEA passthrough stuff to a new PR, thanks for the review

message_definitions/v1.0/common.xml
IamPete1: fixed

hamishwillee:
Thanks @IamPete1 . Looks good to me. I've raised for discussion on Wednesday meeting. Just to get a few more eyes on it.

message_definitions/v1.0/common.xml
hamishwillee: Thanks, as per https://github.com/mavlink/mavlink/pull/1231#issuecomment-532916771 we're opposed to the passthrough message. If missing data proves useful lets add MAVLink messages to support it in a future iteration.

hamishwillee:
@IamPete1 James (@auturgy) has agreed to do a thorough review of this PR and convince himself of the approach. When he says he is basically happy we'll merge.

IamPete1:
@hamishwillee @auturgy any update on this?

hamishwillee:
@IamPete1 Sorry for the delay. @auturgy agreed to go through these in the dev meeting a couple of weeks ago. He was unable to attend the meeting last week due to other work commitments. I THINK he will not be available for anything for at least another week.

Upshot, I don't know whether he has started review, and I don't think we will get a response for a couple of weeks. Many apologies.

Can you wait until he return or is this seriously impacting your work? I'd much prefer wait until he gets back because I have no background on these systems personally.

IamPete1:
@hamishwillee Thanks for the update, no massive rush, just wanted to make sure it hadn't been forgotten

hamishwillee:
FYI, had a chat to auturgy. He is still away, but still aware of this and *hoping* to look at it soon (within 2 weeks).

auturgy:
Needs a rebase, but otherwise I'm happy.

IamPete1:
great!, I will try and get this passing checks this eve

IamPete1:
I have rebased this and done some formatting fixes, the current issue seems to be that cdeg/s is not a valid unit, I think if we have cdeg cdeg/s ought to be allowed. The other fix of course is to go to a float and use deg/s.

hamishwillee:
@IamPete1 I see no reason why not, especially if that is the format coming from the AIS system. This will require a corresponding change in the XSD, which I have made here: https://github.com/ArduPilot/pymavlink/pull/368

@auturgy If you can ping Peter in a few days to get that in would be appreciated.

message_definitions/v1.0/common.xml
```diff
@@ -3834,6 +3834,182 @@
...
+        <description>Towing: length exceeds 200m or breadth exceeds 25m.</description>
+      </entry>
+      <entry value="33" name="AIS_TYPE_DREDGING">
+        <description>Dredging or other underwater ops.</description>
+      </entry>
+      <entry value="34" name="AIS_TYPE_DIVING"/>
+      <entry value="35" name="AIS_TYPE_MILITARY"/>
+      <entry value="36" name="AIS_TYPE_SAILING"/>
+      <entry value="37" name="AIS_TYPE_PLEASURE"/>
+      <entry value="38" name="AIS_TYPE_RESERVED_18"/>
```
hamishwillee: @IamPete1 This is a duplicate. There might be more. This is why  value names are usually extensions of enum name - so AIS_TYPE_DREDGING_RSV1.

IamPete1: fixed

IamPete1:
Its finally passing checks!!, Thanks @hamishwillee @auturgy for reviewing

hamishwillee:
Awesome. Merged :-)

PR #1229, opened 2019-09-05
SYS_STATUS - clarify value invalid voltage

hamishwillee:
This integrates the non-controversial changes to clarify the terms without breaking any semantics. Ie the "bug fixes" in #1182.

@julianoes @auturgy As discussed in dev meeting yesterday.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3759,9 +3759,9 @@
       <field type="uint32_t" name="onboard_control_sensors_enabled" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are enabled:  Value of 0: not enabled. Value of 1: enabled.</field>
       <field type="uint32_t" name="onboard_control_sensors_health" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors have an error (or are operational). Value of 0: error. Value of 1: healthy.</field>
       <field type="uint16_t" name="load" units="d%">Maximum usage in percent of the mainloop time. Values: [0-1000] - should always be below 1000</field>
-      <field type="uint16_t" name="voltage_battery" units="mV">Battery voltage</field>
-      <field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
-      <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy, -1: autopilot estimate the remaining battery</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -3759,9 +3759,9 @@
       <field type="uint32_t" name="onboard_control_sensors_enabled" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are enabled:  Value of 0: not enabled. Value of 1: enabled.</field>
       <field type="uint32_t" name="onboard_control_sensors_health" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors have an error (or are operational). Value of 0: error. Value of 1: healthy.</field>
       <field type="uint16_t" name="load" units="d%">Maximum usage in percent of the mainloop time. Values: [0-1000] - should always be below 1000</field>
-      <field type="uint16_t" name="voltage_battery" units="mV">Battery voltage</field>
-      <field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
-      <field type="int8_t" name="battery_remaining" units="%">Remaining battery energy, -1: autopilot estimate the remaining battery</field>
+      <field type="uint16_t" name="voltage_battery" units="mV">Battery voltage, UINT16_MAX: Voltage not sent by autopilot</field>
```
hamishwillee: Note, would love to set to -1 for consistency but type of field means that would be a breaking change.

message_definitions/v1.0/common.xml
```diff
@@ -3759,9 +3759,9 @@
...
+      <field type="int16_t" name="current_battery" units="cA">Battery current, -1: Current not sent by autopilot</field>
```
hamishwillee: Note:
1. This is a slight change in that -1 could now indicate an autopilot that measures the current but chooses not to send it. In effect not a change for a user, who just needs to know they don't have a value. Also this is a more "standard" way of working it.
2. If we could go back in time would use an int max for this, allowing negative currents.

message_definitions/v1.0/common.xml
```diff
@@ -3759,9 +3759,9 @@
...
+      <field type="int8_t" name="battery_remaining" units="%">Battery energy remaining, -1: Battery remaining energy not sent by autopilot</field>
```
hamishwillee: Note, the original version is probably a copy-paste error. What it means to say is that the remaining battery must be estimated from the voltage/current. Either way, it is more clear to just say the information is not sent/provided, because that is what a user cares about.

PR #1231, opened 2019-09-06
common.xml: NMEA 0183 passthrough message

IamPete1:
This is the NMEA bit broken out from https://github.com/mavlink/mavlink/pull/1228.

I have added a nmea pass-through message, this passes the NMEA sentence as received. NMEA sentences are terminated with a * then two check sum characters, from this it will be possible to detect the end of the message.

The idea of this is that we only partly cover 2 AIS messages out of the 27 with our dedicated message. This allows other messages to be passed through if desired, in quite a lot of cases they would be messages that happen very infrequently, for example you can broadcast plain text messages or get weather reports. There is even a stay-out zone message. It seems silly to duplicate all these in mavlink so i think its better to just allow pass-through directly. This extends to all NMEA stuff, not just AIS

tldr: Over AIS we sometimes get messages that are useful for operators but not for vehicles, plain text messages and weather reports for example. It seems redundant (and a lot of work) to add duplicate mavlink messages for each AIS one. This allows us to send on to the operator as is.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5507,6 +5507,10 @@
       <field type="uint8_t[8]" name="spec_version_hash">The first 8 bytes (not characters printed in hex!) of the git hash.</field>
       <field type="uint8_t[8]" name="library_version_hash">The first 8 bytes (not characters printed in hex!) of the git hash.</field>
     </message>
+    <message id="302" name="NMEA_PASSTHROUGH">
+      <description>Passthrough for NMEA 0183</description>
+      <field type="char[74]" name="sentence">The NMEA sentence without start character, line feed and carriage return</field>
+    </message>
     <!-- UAVCAN related messages. Please keep the range [310, 320) reserved for UAVCAN. -->
     <message id="310" name="UAVCAN_NODE_STATUS">
       <description>General status information of an UAVCAN node. Please refer to the definition of the UAVCAN message "uavcan.protocol.NodeStatus" for the background information. The UAVCAN specification is available at http://uavcan.org.</description>


olliw42:
this is then essentially a tunnel message dedicated to the purpose NMEA, isn't it?

is there no need to better identify the source of the content in the message? 

NEMA 0183 is used also for e.g GPSes, right, so the message name allows/implies this being used here too?

IamPete1:
The idea was that any message not understood by the flight controller could be forwarded on the the ground station. In my aplication this would be for AIS messages but it could be used for any others such as gps ect.

olliw42:
this would further the point that this is essentially a tunnel message, and that it might need a better identification of the source of the content

IamPete1:
The NMEA messages themselves give the scorce, i was trying to avoid having to duplicate the NMEA standard in mavlink. I guess we could cut the talker id and message type off the NMEA sentence and incode as two enums but i think this just it more complex at both ends for little benifit

olliw42:
I'm sorry, I didn't actually wanted to get myself into a deeper discussion here, I clearly lack knowledge and background. I just wondered and found it sufficiently important to drop the points. It is e.g. not clear to me how the NEMA message would distinguish between two GPSes on one autopilot (and you just confirmed it could be used for GPSes). But you guys will sort this out :)

hamishwillee:
@IamPete1 I don't know if the source is relevant if you have multiple GPS on a system, or alternatively, whether this will be part of the payload.

Are all messages 74 char long? If not, and if 0 values at the end of the string are valid, you would also need to encode the length of the message as a field or add some sort of non-zero terminator (otherwise mavlink zero-byte truncation will lose you some data that you can't recover).

You could of course use the `TUNNEL` message for this as well, but I'd be tempted to just say you won't support any messages that aren't explicitly defined - and if people need them they can add in their dialects or suggest push into common. The point being either they are important enough to have messages or they aren't worth sending at all. That is of course a point for discussion :-)

olliw42:
> If not, and if 0 values at the end of the string are valid, you would also need to encode the length of the message as a field or add some sort of non-zero terminator (otherwise mavlink zero-byte truncation will lose you some data that you can't recover).

this is no issue here, since NMEA is ASCII and hence 0 doesn't occur in valid data packets and it has it's unique end-marker CRLF

hamishwillee:
Thanks @olliw42. @IamPete1 has explicitly omitted the The start character, line feed and carriage return.

The bigger discussion in dev call is whether we should push towards not enabling this at all, or using this message, or using a tunnel.

IamPete1:
I guess another option for my AIS use-case would be a AIS pass-through message. It would be essentially the same but for only the AIVDM NMEA message payload.

olliw42:
>  @IamPete1 has explicitly omitted the The start character, line feed and carriage return.

ah, overlooked that .... but it still would be then all ASCI, and 79 and not 74 as he said that the ids should not be extracted in fields  :dancer:

hamishwillee:
@IamPete1 Discussed this in the Dev Call and there was strong opposition to this message. As a point of principle if the data is worth getting then it is worth having a message for. There is nothing to stop you adding new messages for that data if needed.

More broadly, we would be very supportive though of a set of messages to map NMEA into MAVLink.

I'm therefore closing this.

Issue #1225, opened 2019-09-02
MAVLink Tunnel: Two points to discuss


olliw42:
first off, the `TUNNEL` message is working great, as intended. A brief demo is here: https://youtu.be/FRyZ7VZ5lvY.

The following two points are thus not issues MAVLink necessarily has to be concerned about, but it would be nice if MAVLink would provide some help to ease things.

**1. Component ID, GUI heartbeat**

Let's exemplarily consider a system GUI <-serial-> AutoPilot <-serial-> STorM32.

Since the `TUNNEL` message should not be broadcasted, but the target mechanism be used, the GUI obviously must be assigned a component ID. In addition, the GUI needs to emit a heartbeat by istelf, so that the AutoPilot can open the channel and route the messages between GUI and STorM32.

Currently I'm using for the GUI:
component ID = MAV_COMP_ID_USER42 (= 66)
heartbeat.type = MAV_TYPE_GCS
heartbeat.autopilot = MAV_AUTOPILOT_INVALID

This, as said, works fine. However, it would be great and make things easier if - like with other components - a "tunnel server" could be assigned a `MAV_TYPE` and a default component ID. The idea here is somewhat similar to the `MAV_COMP_ID_UDP_BRIDGE` and `MAV_COMP_ID_UART_BRIDGE` components, where it obviously also made sense to provide default component IDs. Since the type of a component is first determined by a `MAV_TYPE` and secondarily by the component ID, the concept implies that also a `MAV_TYPE` should be defined, which also would clarify the heartbeat.

So, my proposal would be to add to the spec a component ID and `MAV_TYPE`:

MAV_COMP_ID_TUNNEL_SERVER = 242
MAV_TYPE_TUNNEL_SERVER = 34

I chose the name "TUNNEL_SERVER" only as example, any other which you guys would find more appropriate would be fine. IMHO one default component ID would be totally sufficient, like with the bridge IDs.

**2. Procedure(s) for registering codes for payload_type**

The TUNNEL message has a field payload_type, and it suggested that software creators can register blocks of types as needed, and involves a file https://github.com/mavlink/mavlink/tunnel-message-payload-types.xml. The procedures and mechanism for doing this were however explicitly left open so far.

Currently I'm using
MAV_TUNNEL_PAYLOAD_TYPE_STORM32_IN = 33333
MAV_TUNNEL_PAYLOAD_TYPE_STORM32_OUT = 33334

[...]

PR #1232, opened 2019-09-07
Add component ID for a TUNNEL node

olliw42:
adds a default component ID for a tunnel communication partner node.

For the rational pl see https://github.com/mavlink/mavlink/issues/1225#issuecomment-529074833.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -736,6 +736,10 @@
       <entry value="241" name="MAV_COMP_ID_UART_BRIDGE">
         <description>Component to bridge to UART (i.e. from UDP).</description>
       </entry>
+      </entry>
+      <entry value="242" name="MAV_COMP_ID_TUNNEL_NODE">
+        <description>Component handling TUNNEL messages (e.g. vendor specific GUI of a component).</description>
+      </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
         <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID.</deprecated>
         <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -736,7 +736,6 @@
       <entry value="241" name="MAV_COMP_ID_UART_BRIDGE">
         <description>Component to bridge to UART (i.e. from UDP).</description>
       </entry>
-      </entry>
       <entry value="242" name="MAV_COMP_ID_TUNNEL_NODE">
         <description>Component handling TUNNEL messages (e.g. vendor specific GUI of a component).</description>
       </entry>


hamishwillee:
This makes sense to me. Will we ever need more than one?

olliw42:
> Will we ever need more than one?

very good question

maybe, potentially yes, why not.

I had given it some thought before, my conclusion in #1225 was that IMHO one default component ID would be totally sufficient, like with the bridge IDs. 

I mean, it's not like that there are not plenty of IDs available for use, these IDs here are just the suggested default IDs, to simplify configuration work. If more than one is needed one can do it, and some level of configuration is then needed anyhow.

Note also that such a default tunnel ID is needed only for "stand-alone tunnel nodes". Many nodes which may use the tunnel message might have however already an ID, e.g. the gimbal does have a gimbal ID.

So, I concluded, one might be sufficient. However, I have to clearly say that it of course can happen that at some point in future one finds that one isn't sufficient ... there are too many unknown variables for me to predict.

I'm sorry that I don't have a clear answer, and just can share my thoughts.

hamishwillee:
Works for me. @auturgy Any thoughts?

olliw42:
I think this are the sort of "fears" the devs had from day one with the tunnel message. And I can't really argue against it, it's a valid point, there is of course a risk of misuse of this message. 

My main arguments were that (1) if folks misuse it in their own (non-public) implementation it doesn't harm but may give them opportunities, (2) I doubt it would really reduce the efforts to standardize (I mean, those not interested to help to standardize wouldn't be helping anyway), and (3) it has also good use cases, for which it in fact would be difficult to establish a meaning full standard.

I hope/believe that with my first application of this message for MAVLink conform pass-through configuration of mounted components I have presented a good example.

> For instance, if using this for a gimbal, whilst the gimbal “speaks mavlink”, it isn’t a mavlink gimbal and would still require a bespoke driver.

the STorM32 has become quite MAVLink-ish recently, so I guess it's fair to call it a mavlink gimbal, yet I'm happy - and I blieve many users would be too - that one now can do this in a fully MAVLink conform way: https://youtu.be/FRyZ7VZ5lvY

a "bespoke driver" is only needed in the GUI, but that's I think fine since this doesn't affect any other party in the MAVLink game.

auturgy:
Understood: I’m not blocking this, just making sure that we document consideration of the broader context.

Thanks for your efforts!

olliw42:
THX

in case you haven't seen, hamishwillee already did a great effort in documentation: https://github.com/mavlink/mavlink-devguide/pull/207. 
In https://github.com/mavlink/mavlink/issues/1225#issuecomment-529794194 he pointed out that this docu is a draft not submitted soon to get a deeper review. Maybe a good place to ensure that the docu gives the proper broader context.

hamishwillee:
@auturgy I would appreciate your https://github.com/mavlink/mavlink-devguide/pull/207
If only to say "Yes/No, that is not a strong enough explanation of when this should/should not be used."
I have pitched this as losing the benefits of MAVLink, but suitable for cases where MAVLink is purely a transport.

hamishwillee:
Dev call. chatted about this, and the general thought was that a close analogy is a router - you need this to provide heartbeat for your tunnel node, when no other component makes sense. Correct?

olliw42:
I don't understand the sentence

in the demonstrated application, the GUI needs a component ID to
* allow the GUI emit a heartbeat, which is needed to enable routing (at least in AP it is such, that it forwards messages only to links where it has seen a heartbeat)
* set the the target fields properly, to avoid broadcasting

hamishwillee:
Thanks! Merged.

olliw42:
THX!

PR #1233, opened 2019-09-07
Register 10 tunnel payload codes for STorM32

olliw42:
kindly asks to register 10 tunnel payload codes for the STorM32 gimbal controller project.

Follows the suggested rules for registration outlined here: https://github.com/mavlink/mavlink/issues/1225#issuecomment-529075395

<strike>In contrast to the suggested rules, which suggest a default block of 25 codes, I here ask for only 10 codes. Main reason is that I don't want to blow up the documentation web page with a that long enum. </strike> EDIT: default size was settled to 10

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3500,6 +3500,36 @@
       <entry value="0" name="MAV_TUNNEL_PAYLOAD_TYPE_UNKNOWN">
         <description>Encoding of payload unknown.</description>
       </entry>
+      <entry value="200" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED0">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="201" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED1">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="202" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED2">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="203" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED3">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="204" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED4">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="205" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED5">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="206" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED6">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="207" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED7">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="208" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED8">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
+      <entry value="209" name="MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED9">
+        <description>Registered for STorM32 gimbal controller.</description>
+      </entry>
     </enum>
     <enum name="MAV_ODID_IDTYPE">
       <entry value="0" name="MAV_ODID_IDTYPE_NONE">


hamishwillee:
Hi @olliw42,

> if I should change this PR to include the new location https://github.com/mavlink/mavlink/definition_files/tunnel-message-payload-types.xml, or not

I don't think so. We can change that anytime, and I would suggest we wait until we have someone who wants to define their payload.

I'd prefer to wait until someone else has commented on https://github.com/mavlink/mavlink-devguide/pull/207 ... but if I get no feedback by MAVLink meeting I'll merge anyway as this feels like common sense.

Cheers
Hamish

olliw42:
great, many thx for the feedback

I'm certainly happy to wait until Wednesday LOL

I would not expect that you get as much response to the devguide docs as one might would want to have by Wednesday. So, it's good that this wouldn't uphold this PR here.

thx, Olli

julianoes:
I don't understand why tunnel message specs should end up in the MAVLink specs. To me it's a tunnel, and MAVLink should not be concerned with any of its details. Otherwise you break the promise that you can do whatever through the tunnel and for every new tunnel implementation you need to talk to "the MAVLink people" again to get your addition approved.

hamishwillee:
> I don't understand why tunnel message specs should end up in the MAVLink specs...

`TUNNEL` copied the pattern of the [V2_EXTENSION](https://mavlink.io/en/messages/common.html#V2_EXTENSION) which had this idea - that was never built out so I don't know how much thought went into this aspect originally or into `TUNNEL`; probably just enough to say "maybe we will want it one day so better future proof". 

Everything still WIP so we can remove this possibility and simply state that all we do is reserve blocks for specific purposes to reduce chance of clashes. I don't have a strong opinion, other than if we start making tunnels seem OK perhaps they will get more use than we want.

olliw42:
@julianoes 

first off, I think that the idea of a payload type and of having a mechanism of registering values for avoiding collisions is a very good idea (not sure you contested it)

then, I fully agree that the values, registration, and potential documentation shouldn't be part of MAVLink in the sense of a spec

however, I also think that it would be most convenient for all to put this in a place hosted by MAVLink. I see that a bit like the MAVLink code generator. It obviously isn't part of the MAVLink spec (that's at least my interpretation of the fact that it isn't hosted in the mavlink/mavlink repo). However, I consider it most inconvenient that it is not hosted here, but somewhere else, and in addition is just a small part of this somewhere else. In the very same sense I would consider it convenient that the discussed mechanisms would be put under the wings of MAVLink.

hamishwillee:
@julianoes @olliw42 Re https://github.com/mavlink/mavlink/pull/1233#issuecomment-532547131

Discussed in the dev meeting last night (@julianoes not present). We do need to provide mechanism for reserving blocks/avoiding collisions. No one was particularly concerned how this was done - ie by explicitly allocating each value as done here or by just having a comment in the enum. Upshot though, is that at least for now this is fine. @julianoes if you have no objection I will merge this PR.

The point of contention is whether we should provide the mechanism for allowing someone to document their encoding spec - ie the new XML definition file referred to in the message. This isn't implemented and one might argue that if we don't really want tunnels, we really don't want to support/enable documenting them.

My take is that we should remove that altogether as an option "for now". If someone wants this later on, they can provide a compelling use case. What it does though is remove the dangling expectation that this is something we support. THoughts?

olliw42:
> The point of contention is whether we should provide the mechanism for allowing someone to document their encoding spec

> My take is that we should remove that altogether as an option "for now".

as you mentioned, the xml part was strongly motivated by the V2_EXTENSION message, and it makes sense when one reads it, but I can understand that there can be cons. 

from my side the documentation part isn't really needed. It would be very convenient, but at the end of the day it's just about convenience. So, if you guys decide to drop it, perfectly fine with me.

The "rules" 5. and 6. may need to be revisited, but this discussion can or should be moved to https://github.com/mavlink/mavlink-devguide/pull/207.

A possible alternative would be that as part of the registration of the block of values one can or should add a reference or contact address to the enum description, so that other folks have at least a good chance to find info or a contact in case they ever would bother.

> What it does though is remove the dangling expectation that this is something we support.

When I was reading the description for the V2_EXTENSION message the first time, and was reading that

_"The particular encoding used can be extension specific and might not always be documented as part of the MAVLink specification."_

I read it exactly as to mean that this part is not supported by MAVLink. In the description for the TUNNEL message this has been made even stronger. I take statements in specs seriously, and I think you should expect any user of the MAVLink spec to do so too. In short, I can't see where the worry of a dangling expectation of support comes from. :)

but as said, perfectly fine with me to remove the xml part

olliw42:
@hamishwillee 

just to have mentioned it: I'm not sure but got an impression from some other enums that maybe writing e.g. `MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED_0` instead of `MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED0` might be more conform to the general MAVLink style.

julianoes:
I have my opinion but I can live with whatever.

olliw42:
so, what would be your suggestion?

julianoes:
I think whoever needs the tunnel should implement everything outside of mavlink. If you decide to do something without mavlink why would you then rely on mavlink to spec out things? Isn't it supposed to be transparent to mavlink? If you start speccing what is in the tunnel outside of the tunnel, then you might as well not call it a tunnel but spec it in mavlink itself.

I guess I still don't understand it, so just ignore my opinion and do what works for you.

olliw42:
I understand the point of view you're expressing (and actually largely agree), but I don't understand your conclusions of that.

are you saying there shouldn't have been a payload_type field in the TUNNEL message?

IMHO, if one agrees with a payload_type field, it follows that one needs some sort of registration mechanism to avoid conflicts. This mechanism is certainly outside of the MAVLink spec, and thus could be hosted anywhere. But IMHO a payload_type field implies the existence of a registration mechanism. Are you saying that said registration mechanism should be located somewhere else, e.g. in a new repo mavlink/tunnel-stuff or mavlink/non-spec-stuff or whatever? 

IMHO, if one agrees with a payload_type field, it follows that it is an MAVLink enum. Where should the definition of the enum in your opinion be located, if not in common.xml?

julianoes:
Aha. Yes if there is a  paylod_type field it should be documented as you suggest.

olliw42:
> if there is a paylod_type field it should be documented as you suggest.

and since there is such a payload_type field, this implies you agree to a registration mechanism, there the enum in common.xml is extended via PRs to mavlink/mavlink, like in this PR here and suggested elsewhere, right? 

(this is just about the enum, not the contentious xml file thing)

hamishwillee:
> just to have mentioned it: I'm not sure but got an impression from some other enums that maybe writing e.g. MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED_0 instead of MAV_TUNNEL_PAYLOAD_TYPE_STORM32_RESERVED0 might be more conform to the general MAVLink style.

I'm OK with what you have and I'm going to merge.

I'm also going to remove the info about the XML file. It can be added retrospectively if needed without any issue.

I think we're all in agreement now re the ids for tunnel payload format. There was one other alternative of course - deciding that either there can only be one tunnel in a network at a time OR that the payload itself contains the format and must self determine valid packets. I think what we have makes more sense than that.

olliw42:
THX!

PR #1235, opened 2019-09-16
FOCUS_TYPE_METRES - add

hamishwillee:
Add FOCUS_TYPE_METERS and ZOOM_TYPE_FOCAL_LENGTH to respective enums, as discussed in 
https://github.com/mavlink/mavlink/pull/1224#issuecomment-531091300

This does not add messages to get the focal range/zoom/etc from a camera. IMO that probably should be done - because otherwise IMO the mechanisms for setting these outside of parameters may not be worth having at all. And while I understand this could quickly get complicated, adding ranges for video and camera does not seem too scary (and addresses the 80% of likely use).

If you agree can you add range-getting messages either in their own PR or this one?


hamishwillee:
PS I'll fix the build errors when you say the docs are good and/or have added the new messages.

PR #1238, opened 2019-09-20
TUNNEL: Remove option to publish tunnel format

hamishwillee:
As discussed in dev call, we don't see need to publish opaque tunnel format, as we wish to discourage its use. This removes that option. If it proved necessary could be added back without harm/cost.


hamishwillee:
This was discussed in the devcall. Essentially there is no great need to publish tunnel encoding formats. This removes that suggestion. If it is needed later it can be re-added.

PR #1239, opened 2019-09-26
Update OpenDroneId messages to match version 0.8 of the specification

friissoren:
Multiple changes to the message fields and new messages were added in version 0.8 of the OpenDroneID specification. This PR aligns the Mavlink messages.

The OpenDroneID specification is basically frozen but you never really know so let's keep the WIP tags for the time being.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3768,8 +3768,17 @@
       <entry value="0" name="MAV_ODID_AUTH_TYPE_NONE">
         <description>No authentication type is specified.</description>
       </entry>
-      <entry value="1" name="MAV_ODID_AUTH_TYPE_MPUID">
-        <description>Manufacturer Programmed Unique ID.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Note, merged as still in WIP and you are currently sole user. When comes time to remove WIP we'd like more docs on how this works/is expected to be implemented by senders and receivers. The assumption currently is that this is a direct mapping to open ID messages so that on the transmitter end you'll have some layer that converts the MAVLink message payload directly to format expected by the transponder.

friissoren:
Indeed. I will work on some more documentation, when I find the time and we know that things truly won't change anymore.
About conversion to "compressed" OpenDroneID format for transmission on Bluetooth, WiFi, Internet (and conversion back again to Mavlink messages, in case someone needs that). This part already exists and I am in the middle of updating it to be in line with this recent Mavlink message update:
https://github.com/opendroneid/opendroneid-core-c/tree/master/libmav2odid

Issue #1244, opened 2019-10-03
Pl clarify compid = 0 (MAV_COMP_ID_ALL) in the message header


olliw42:
the current docu is somewhat confusing on the matter, but it seems that it supports using MAV_COMP_ID_ALL (= 0) for the component ID in the message header.

This to me looks weird, since naively I would have thought that any component in a MAVLink system should have a unique component ID, and thus send out messages only with the component ID field in the header be populated accordingly. Having a component sending out messages with component ID = MAV_COMP_ID_ALL = 0 in the header doesn't make sense to me. 

Note that this is not about having target component ID = 0, which means broadcasting and of course makes lots of sense. Note also that in the routing rules there is no mention whatsoever of a message component ID = 0.

I've stumbled across this since QGC appears to (occasionally !?) send out messages with component ID = 0 in the message header, specifically e.g. the requestCameraInformation. 

My naive expectation effectively means that there can't be messages with msg.compid = 0, but if I add code to STorM32 which rejects all messages with msg.compid = 0 based on my naive expectation I find that QGC's requestCameraInformation command is rejected and the MAVLink camera service be broken. My debug output also explicitly shows that QGC's requestCameraInformation command has msg.compid = 0.

So, the above issue appears to not only be academic, but be of some relevance. 

If component ID = MAV_COMP_ID_ALL = 0 in the header does make some sense, I would wish that the docu could be improved to better explain what the usage is. If it doesn't and indeed is effectively forbidden, i.e. allowed only for target, the docu could say this.

thx for all your work on MAVLink

hamishwillee:
IMO your "naive" impression is correct. System and component IDs are used for addressing, and 0 is used to indicate broadcast. Obviously this means that if a system or component identifies itself as having an address of 0 in the sender component id this means that the component is not separately addressable. [Again note, this is sender id in the packet fields not target id in the payload]

This is clearly a mis-use but is it a big problem?
- For receiving broadcast messages, clearly not.
- If receiving a message or command then all components on the receiving system will receive the message. This could have unintended consequences depending on the command.
- If sending a command the ACK will be seen by everything on the commanding system. Probably not a problem.

I don't think semantically a component should ever need to identify itself as MAV_COMP_ID_ALL and this would break things. 

In terms of docs, we would need to clarify this in https://mavlink.io/en/guide/serialization.html - which states 0 is valid. 

[...]

PR #1245, opened 2019-10-06
MAV_COMPONENT: Clarify broadcast ID only used for targets

hamishwillee:
Follows on from https://github.com/mavlink/mavlink/issues/1244

0 ids are use to indicate broadcast address and should not be used as the source address of a component. If you did this a component could not be separately addressed as all systems would think the message was intended for them. In addition, if such a component sent a command the ACKs would go to all other systems.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -375,7 +375,7 @@
       Components must use the appropriate ID in their source address when sending messages. Components can also use IDs to determine if they are the intended recipient of an incoming message. The MAV_COMP_ID_ALL value is used to indicate messages that must be processed by all components.
       When creating new entries, components that can have multiple instances (e.g. cameras, servos etc.) should be allocated sequential values. An appropriate number of values should be left free after these components to allow the number of instances to be expanded.</description>
       <entry value="0" name="MAV_COMP_ID_ALL">
-        <description>Used to broadcast messages to all components of the receiving system. Components should attempt to process messages with this component ID and forward to components on any other interfaces.</description>
+        <description>Target id (target_component) used to broadcast messages to all components of the receiving system. Components should attempt to process messages with this component ID and forward to components on any other interfaces. Note: This is not a valid *source* component id for a message.</description>
       </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -375,7 +375,7 @@
       Components must use the appropriate ID in their source address when sending messages. Components can also use IDs to determine if they are the intended recipient of an incoming message. The MAV_COMP_ID_ALL value is used to indicate messages that must be processed by all components.
       When creating new entries, components that can have multiple instances (e.g. cameras, servos etc.) should be allocated sequential values. An appropriate number of values should be left free after these components to allow the number of instances to be expanded.</description>
       <entry value="0" name="MAV_COMP_ID_ALL">
-        <description>Used to broadcast messages to all components of the receiving system. Components should attempt to process messages with this component ID and forward to components on any other interfaces.</description>
+        <description>Target id (target_component) used to broadcast messages to all components of the receiving system. Components should attempt to process messages with this component ID and forward to components on any other interfaces. Note: This is not a valid *source* component id for a message.</description>
```
olliw42: 0 can be used for BOTH a target system id and target component id, not only component id

maybe something like this: Used to broadcast messages with a target system to all components of the receiving system. Components should process and forward such messages according to the routing rules (link).  Note: This is not a valid Id for the source system and component ids in the message header.

julianoes: `MAV_COMP_ID_ALL` can only be used for target component id, strictly speaking. If it had a real type (instead of `uint8_t` it would only match component IDs.

To address all system IDs we should have something like `MAV_SYS_ID_ALL`, also 0.

olliw42: ah ... true ... the name ... MAV_COMP LOL

sorry, I confused myself by that it's 0 in both cases

hamishwillee:
FYI, I also updated the serialisation docs to note that the value must be non zero in the packet https://github.com/mavlink/mavlink-devguide/pull/215

I have not stated what a recipient should do if it gets such a value. Probably the right thing to do would be to drop the packet, so that the problem with the sender is picked up during testing.

PR #1247, opened 2019-10-07
Extend ATTITUDE_QUATERNION for correct display of attitude with tailsitters

jlecoeur:
[edited]:

The new field is a rotation offset by which the attitude quaternion and angular speed vector should be rotated for user display (quaternion with [w, x, y, z] order, zero-rotation is [1, 0, 0, 0], send [0, 0, 0, 0] if uneeded).

This field is intended for systems in which the reference attitude may change during flight. For example, tailsitters VTOLs rotate their reference attitude by 90 degrees between hover mode and fixed wing mode, thus repr_offset_q is equal to [1, 0, 0, 0] in hover mode and equal to [0.7071, 0, 0.7071, 0] in fixed wing mode.

Related issues:
- https://github.com/mavlink/qgroundcontrol/issues/3849
- https://github.com/PX4/Firmware/issues/5291 
- https://github.com/PX4/Firmware/issues/9179
- https://github.com/PX4/Firmware/issues/10405

Matching PR for PX4 and QGC:
- https://github.com/mavlink/qgroundcontrol/pull/7881
- https://github.com/PX4/Firmware/pull/13114

FYI @RomanBapst @dagar @thomasgubler 

[Previous description]:
> This PR extends the message VFR_HUD with attitude fields meant for user display.
It is especially useful for drones that have different attitude reference in rotary-wing and fixed-wing modes.  
For example, tailsitter VTOLs rotate around the pitch axis by 90 degrees during transitions. As a consequence, the attitude displayed on the HUD is currently unusable when in fixed-wing mode (see screenshots in https://github.com/PX4/Firmware/pull/9440).
The added fields in VFR_HUD contain the attitude rotated so as to be useful for a human operator.
The messages ATTITUDE and ATTITUDE_QUATERNION  are untouched, which ensures that no offboard/companion workflow is broken.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4006,6 +4006,8 @@
       <field type="float" name="rollspeed" units="rad/s">Roll angular speed</field>
       <field type="float" name="pitchspeed" units="rad/s">Pitch angular speed</field>
       <field type="float" name="yawspeed" units="rad/s">Yaw angular speed</field>
+      <extensions/>
+      <field type="float[4]" name="repr_offset_q">Rotation offset by which the attitude quaternion and angular speed vector should be rotated for user display (quaternion with [w, x, y, z] order, zero-rotation is [1, 0, 0, 0], send [0, 0, 0, 0] if uneeded). This field is intended for systems in which the reference attitude may change during flight. For example, tailsitters VTOLs rotate their reference attitude by 90 degrees between hover mode and fixed wing mode, thus repr_offset_q is equal to [1, 0, 0, 0] in hover mode and equal to [0.7071, 0, 0.7071, 0] in fixed wing mode.</field>
     </message>
     <message id="32" name="LOCAL_POSITION_NED">
       <description>The filtered local position (e.g. fused computer vision and accelerometers). Coordinate frame is right-handed, Z-axis down (aeronautical frame, NED / north-east-down convention)</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4007,7 +4007,7 @@
       <field type="float" name="pitchspeed" units="rad/s">Pitch angular speed</field>
       <field type="float" name="yawspeed" units="rad/s">Yaw angular speed</field>
       <extensions/>
-      <field type="float[4]" name="repr_offset_q">Rotation offset by which the attitude quaternion and angular speed vector should be rotated for user display (quaternion with [w, x, y, z] order, zero-rotation is [1, 0, 0, 0], send [0, 0, 0, 0] if uneeded). This field is intended for systems in which the reference attitude may change during flight. For example, tailsitters VTOLs rotate their reference attitude by 90 degrees between hover mode and fixed wing mode, thus repr_offset_q is equal to [1, 0, 0, 0] in hover mode and equal to [0.7071, 0, 0.7071, 0] in fixed wing mode.</field>
+      <field type="float[4]" name="repr_offset_q">Rotation offset by which the attitude quaternion and angular speed vector should be rotated for user display (quaternion with [w, x, y, z] order, zero-rotation is [1, 0, 0, 0], send [0, 0, 0, 0] if field not supported). This field is intended for systems in which the reference attitude may change during flight. For example, tailsitters VTOLs rotate their reference attitude by 90 degrees between hover mode and fixed wing mode, thus repr_offset_q is equal to [1, 0, 0, 0] in hover mode and equal to [0.7071, 0, 0.7071, 0] in fixed wing mode.</field>
     </message>
     <message id="32" name="LOCAL_POSITION_NED">
       <description>The filtered local position (e.g. fused computer vision and accelerometers). Coordinate frame is right-handed, Z-axis down (aeronautical frame, NED / north-east-down convention)</description>


WickedShell:
This is a lot of data to the VFR_HUD message. I'd probably rather see a companion message added instead, as this is adding a lot of data to a commonly used message that doesn't have value in a lot of use cases.

jlecoeur:
@WickedShell fair enough. Although I do not agree that it does not have value in most cases. What this PR does is making VFR_HUD the primary source of information for operation by human, while ATTITUDE_QUATERNION is preferred for machine to machine communication. 
Also, we already have two messages for attitude (ATTITUDE and ATTITUDE_QUATERNION) , so I am not sure that adding a third one will make things any clearer. VFR_HUD, on the other hand, is meant to be understood by a human so it seems like a good fit.
Strawman solution, we can somehow reduce the size of VFR_HUD by removing the fields rollspeed, pitch speed and yawspeed.

Let's see if someone has another idea/optinion

hamishwillee:
I'm a human. The new fields q1-q4 mean nothing to me as a human. By comparison rollspeed, pitchspeed, yawspeed and all the existing fields are pretty clear. How are they presented in the HUD? What do you do with them?

Lorenz has been pushing for MAVLink to use "natural" units like radians in MAVLink rather than convenient units like degrees. FWIW  I somewhat agree for setters, but for information designed to be consumed by real people it seems like adding more work for little value. 

On systems that don't understand the fields the new values will not be sent but a receiver that has been updated will get 0 - if that is a valid value you need to consider what will happen in this case. Generally speaking you should also define a value meaning "not sent, which for floats is commonly NaN".

We have the options:
1. Extensions like this. 
1. New HUD message that contains all the HUD fields we like. 
1. New message for just the new fields (though in this case sounds like these messages already exist)

My first take on this is that the existing message has been designed to make it very easy for some HUD system (e.g. OSD) to present relevant information without any extra processing or thinking. It groups all the relevant information and uses the form that it is to be presented. I like message consistency so I'd be tempted to extend it in the same way. If we still want to group relevant info but use radians etc then I'd be tempted to have an entirely new message.

jlecoeur:
These are  good points @hamishwillee. Degrees are way more readable than radians. Also I agree that a quaternion is not readable by a human, euler angles (roll pitch yaw) would be better. 
 In fact, I started by adding euler angles, but then used quaternion for the reason you mention: the case where it is not sent. With quaternion, I can check the norm, and ignore it if the norm is 0 (see the QGC PR). Is there a way to tell, from the receiver, if extension fields were set? If yes, then I am all in favor of using euler angles in HUD.

hamishwillee:
Most of what I know about extensions is here: https://mavlink.io/en/guide/define_xml_element.html

Essentially though,
- sender doesn't know about extensions, and the receiver does, the receiver will get zero filled values. 
- sender does know about extensions and receiver doesn't, receiver won't get the fields at all. 

What this means is that the 0 value of an extension has to mean "not supported" or "invalid" or it doesn't matter that the value is misinterpreted. 

The most common workaround is to make zero mean "not supported", and either use some other value to represent zero (e.g. Infinite) or have a recipient treat a small non zero as zero (0.1).  These both offend the purist in me, though the second is less irritating. 

If neither are acceptable then the remaining choice is usually to either have a new message or accept that the message won't work right for recipients (I would not do that in this case, as things like OSD that might use this never update).

jlecoeur:
Thanks for that. It is a pity that NaN is not used for "invalid" float fields. Maybe that is something to fix with MAVLink 3?

In this case, a quaternion equal to [0 0 0 0] is not a valid attitude quaternion, so 0-filled extension can be detected. 
It would unfortunately not be possible to do the same with euler angles because [0 0 0] is perfectly valid.

jlecoeur:
Options to move forward:
1. Message
- extend VFR_HUD (this PR) 
- create new VFR_HUD_ATTITUDE

2. Units
- radians (this PR) 
- degrees

3. Content
- quaternion + rates (this PR)
- quaternion
- euler angles + rates
- euler angles
- same as above + validity flag

---
My opinion:
1. Extend VFR_HUD not to add more clutter.
2. No strong feeling, I am OK with degrees as it is more readable, although less consistent with other messages
3. Rates are not absolutely needed. Quaternion is nice because the receiver can detect its validity. Euler angles would be more human friendly and not require additional processing on the receiver side. Based on the comments so far, I would say euler angles in degrees + a validity flag is the way to go

hamishwillee:
@LorenzMeier Thanks for taking the time to explain your thinking.

An argument that humans can't necessarily understand quaternions a certainly not a good reason  not to send them. But an argument that the binary data isn't human readable isn't a good reason choose _not_ to send human readable data.

Arguments for not using a human readable format might be that they don't contain the necessary information or precision required by the recipient. Or that both ends are going to be natively processing them in a non human readable format so converting is pointless.

The argument I might use for a human readable format in this case is that a recipient like an OSD is pretty resource constrained. They need to be able to decode the message but they will not likely need to implement any microservice other than the connection. The aren't going to use the MAVSDK. They aren't in the flight stack business - does it make sense to force them to understand quaternions to ultimately display a value in degrees/second?

hamishwillee:
@jlecoeurMy understanding of @LorenzMeier point is that the machine readability or otherwise of the message is irrelevant. Does this mean we could reasonably use ATTITUDE and ATTITUDE_QUATERNION?

If a concern is data marshalling  - ie presenting all the HUD information together we might address that in documentation of the message.

The other point is that we should use radians for reason that some recipients will actually need the greater precision.

> Thanks for that. It is a pity that NaN is not used for "invalid" float fields. Maybe that is something to fix with MAVLink 3?

 A message sent by a system that does not understand the extensions is just the same as one that has had zero-byte truncation. The recipient zero-fills the receive buffer of the right size then populates the fields it did get values for. This is very efficient. 
It is impossible to make the  zero-filled floats into NaN because they could actually have been validly sent as zero values - we just don't know.

LorenzMeier:
@hamishwillee A legacy OSD is increasingly a fringe use case for MAVLink. It would be like arguing that HTML can't be extended because of legacy browsers like Mosaic. Those legacy systems can constrain themselves to MAVLink 1. And more compute is naturally becoming available.

It is well known that attitude has a gimbal lock if not using a quaternion and it seems unreasonable, given the general purpose nature of the protocol, to use a flawed data representation.

LorenzMeier:
For anyone needing a rotation matrix or Euler angles:
https://github.com/mavlink/c_library_v2/blob/master/mavlink_conversions.h#L103

jlecoeur:
> @jlecoeur My understanding of @LorenzMeier point is that the machine readability or otherwise of the message is irrelevant. Does this mean we could reasonably use ATTITUDE and ATTITUDE_QUATERNION?

jlecoeur:
> @jlecoeur My understanding of @LorenzMeier point is that the machine readability or otherwise of the message is irrelevant. Does this mean we could reasonably use ATTITUDE and ATTITUDE_QUATERNION?

@hamishwillee Yes we could do this, i.e. do the rotation autopilot side and use the existing attitude messages. This is what Ardupilot does as far as I know. This is also what I tried a while back, see https://github.com/PX4/Firmware/pull/9440.
The potential problem with this approach is the use case of offboard mode, or more generally machine-based operation (SDK), because the rotated attitude is not the actual system attitude, has a discontinuity during transition, and is not coherent with other messages (IMU, etc).

hamishwillee:
@julianoes @WickedShell @dagar Can you comment on the concerns @jlecoeur raises here: https://github.com/mavlink/mavlink/pull/1247#issuecomment-539856754

My thought would have been that if ArudPilot can do it why not PX4. But MAVLink should make sensible decisions that work for a broad range of flight stack designs.

julianoes:
> because the rotated attitude is not the actual system attitude, has a discontinuity during transition, and is not coherent with other messages

I agree with this. I would not mess with ATTITUDE/ATTITUDE_QUATERNION. At the end of the day something needs to "speak the truth". I would rather do the proposed addition or add a separate message depending on the impact on OSDs, etc.

**Taking a step back:**
I'm wondering what this indicator actually means for a tailsitter VTOL, especially during the transition. At this point the HUD becomes a representation of the offset against the setpoint, right? For example, as a tailsitter transitions from vertical to horizontal flight the setpoint changes gradually while the VTOL pitches down. If well controlled, the horizon would always be around 0, if not you would see pitch movements. Alternatively, you could make it switch instantly which means you go from tilting down (forward) to suddenly tilting up until you reach horizontal flight.

Therefore, wouldn't it make more sense to have a visualization of the plane in the right attitude instead of the HUD?

jlecoeur:
> At this point the HUD becomes a representation of the offset against the setpoint, right?

against the "neutral point" yes. In multirotor mode, the neutral point is when rotors are aligned with the horizon, while in fixed-wing mode the neutral point is when wings are aligned with the horizon.  

>  If well controlled, the horizon would always be around 0, if not you would see pitch movements

Agreed. That can be added later on PX4-side if this PR goes through.

Another step back, a HUD is meant to give you the information you need to control the drone only from the instruments, in a first-person view. Arguably this PR would make it possible to control a tailsiteer in fixed wing drone BVLOS, while it is impossible now.

> Therefore, wouldn't it make more sense to have a visualization of the plane in the right attitude instead of the HUD?

Do you mean a third person 3D view of the aircraft? Or maybe better, 3 views (rear, side, north-aligned-top) displaying a representation of the aircraft based on its type (rotated by roll, pitch and yaw angles, respectively), and also dashed lines indicating the current setpoint. 
I would certainly use it. As an RC pilot, I sometimes find the HUD confusing. For example most visual features rotate leftward during rightward roll. A third person view would be more intuitive for me. However, many people are very happy with a HUD, so it still has to display useful information.

julianoes:
> Another step back, a HUD is meant to give you the information you need to control the drone only from the instruments, in a first-person view. Arguably this PR would make it possible to control a tailsiteer in fixed wing drone BVLOS, while it is impossible now.

Yes, you're right. It would make it possible to "control" it, however, I would think, it's very unlikely you would ever control a BVLOS drone by looking at attitude. You would look at the map, state and altitude and plan mission or a landing spot. The only exception might be an emergency landing with FPV with a HUD overlay. Ok so that would be one use case.

> Do you mean a third person 3D view of the aircraft?

Yes, that's what I mean.

jlecoeur:
3rd person view would great indeed.

> it's very unlikely you would ever control a BVLOS drone by looking at attitude. You would look at the map,

That is right. Note that the map is also affected by the issue. Let me copy paste the screenshot from https://github.com/PX4/Firmware/pull/9440 for reference
- Current: image
- Fixed: image

hamishwillee:
So did this discussion result in a conclusion?

IF you put the quaternions at the end (perhaps in an array) they can be zero filled and auto truncated for systems that don't want to use them (you'd indicate that 0 = not supported in the doc). You also mentioned dropping the euler rates. If you have to have these then I quite like the idea of using a byte to indicate that the data is valid.

julianoes:
As said I'm fine with a new message or a VFR_HUD addition.

jlecoeur:
@hamishwillee Quaternion as array would be better, no doubt. I fear that we will never be able to update ATTITUDE_QUATERNION to use an array. Now, SET_ATTITUDE_TARGET and ATTITUDE_TARGET use arrays, which makes for a terrible experience when comparing 1-indexed data from ATTITUDE_QUATERNION and 0-indexed data from ATTITUDE_TARGET !
Do we prefer to stay consistent, or to move forward with a more convenient array format?

@julianoes How would you name the new message? ATTITUDE_QUATERNION_VIRTUAL ? ( or ATTITUDE_QUATERNION_VTOL, ATTITUDE_QUATERNION_TAILSITTER, etc)

I feel like a new message will be the more acceptable option.

Let's do a quick poll:

### Option A: new message
```xml
<message id="400" name="ATTITUDE_QUATERNION_VIRTUAL">
      <description>The attitude in the aeronautical frame (right-handed, Z-down, X-front, Y-right), expressed as quaternion. Quaternion order is w, x, y, z and a zero rotation would be expressed as (1 0 0 0). 
                   This message complements ATTITUDE_QUATERNION for systems in which the reference attitude may change during flight. Tailsitters VTOLs, for example, rotate their reference attitude by 90 degrees between hover mode and fixed wing mode, thus making ATTITUDE_QUATERNION difficult to interpret by a human operator. 
                   In such case, the system is expected to send ATTITUDE_QUATERNION with non-rotated (true) attitude, and to send ATTITUDE_QUATERNION_VIRTUAL with the attitude rotated accordingly to the current flight mode.</description>
      <field type="float[4]" name="q">Attitude quaternion (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
      <field type="float[3]" name="body_rates" units="rad/s">Angular speed about X, Y, Z body axis.</field>
</message>
```

### Option B: extend VFR_HUD
```xml
<message id="74" name="VFR_HUD">
      <description>Metrics typically displayed on a HUD for fixed wing aircraft.</description>
      <field type="float" name="airspeed" units="m/s">Current indicated airspeed (IAS).</field>
      <field type="float" name="groundspeed" units="m/s">Current ground speed.</field>
      <field type="int16_t" name="heading" units="deg">Current heading in compass units (0-360, 0=north).</field>
      <field type="uint16_t" name="throttle" units="%">Current throttle setting (0 to 100).</field>
      <field type="float" name="alt" units="m">Current altitude (MSL).</field>
      <field type="float" name="climb" units="m/s">Current climb rate.</field>
      <extensions/>
      <field type="uint8_t" name="attitude_valid">Boolean indicating whether the attitude fields (q and body_rates) contain valid information (valid: 1, invalid: 0). Default is 0 (invalid).</field>
      <field type="float[4]" name="q">Attitude quaternion (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
      <field type="float[3]" name="body_rates" units="rad/s">Angular speed about X, Y, Z body axis.</field>
</message>
```

@hamishwillee can we add polling to this repo (ex https://github.com/apps/polls)?

hamishwillee:
@jlecoeur I have requested a poll, but I don't have rights to approve it. 

1. We don't need to be consistent with old messages - use the format that is best for the user and consider on a per-data basis (i.e. you may decide quaternions are an array but not rates). If that is an array then go for it. I have no strong "general" position on when you use an array or not.
2. If these are intended for HUD then good to link them to HUD message in some way. ie you might consider having VFR_HUD_V2 that includes the fields. YOu'd prefer to use an extension if you can add the fields such that they are truly optional (e.g. zero filled if someone doesn't want to use them)
3. The new message does not indicate how you plan to indicate "information not supplied by sender".
4. MAVLink reorders non-extension fields to be smallest last ([size of fields here](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavparse.py#L48)) (otherwise holding declaration order) and then extensions are added to the end in declaration order. Any zero filled bytes at the end are stripped for sending (reducing the payload). Upshot is that I know that you can zero fill the quaternion to indicate that it is not used - to benefit from the above behaviour it needs to be last in order of declaration. If you can mark both rates and quaternion as 0 to indicate not used then order of declaration doesn't matter so much.
5. And for the new message, same logic applies - unless you know messages are send on high bandwidth links you'd prefer to declare messages such that the most data possible does not get sent.

hamishwillee:
PS. I'll try get clear direction for you in the devcall on Wednesday.

hamishwillee:
Polls should be good to go now.

hamishwillee:
@jlecoeur Dev call had more questions before direction can be provided:
1. Why are rates useful in HUD? Attitude might reasonably be displayed, but what would user do with the rates?
2. Is your intention/need entirely for HUD use, or is there some other purpose for which you require this information?

jlecoeur:
1. Rates are not used in QGC's HUD if that is your question. However, if the user reads rotated attitude in HUD and non-rotated rates in another message, it will be confusing.
2. My intention is to pass information that makes sense to the user, extending the HUD message was a way to avoid the creation of a third attitude message. In the end I feel like a new message like ATTITUDE_QUATERNION_VIRTUAL is superior in both clarity of the intent and bandwidth because we can stream only from the few systems that need it.

hamishwillee:
> ATTITUDE_QUATERNION_VIRTUAL is superior in both clarity of the intent and bandwidth because we can stream only from the few systems that need it.

I'm sorry if I'm being dim, but I'm a bit confused by "thus making ATTITUDE_QUATERNION difficult to interpret by a human operator. ". Wasn't Lorenz point that the protocol shouldn't be trying to satisfy a human? It should transfer the best data and let the GCS do the right thing. 

Can a GCS do what it needs for the human operator with the information in [ATTITUDE_QUATERNION](https://mavlink.io/en/messages/common.html#ATTITUDE_QUATERNION)?

If the answer is yes, then IMO we don't need a new message. If the answer is no, would a better option be to add frame information to `ATTITUDE_QUATERNION` as an extension (now that we're not adding to the HUD message).

jlecoeur:
@hamishwillee I do not think you are being dim :-) I must be using the wrong vocabulary. Lorenz and you are right that binary and quaternions representations are not human readable. My point was that for tailsitters, even when the quaternion is processed and displayed to the user, it is difficult to interpret, see my first screenshot above. The tailsitter was flying in fixed with mode, doing a coordinated turn, but you cannot see it from the displayed attitude (both in the map and the HUD).

> Can a GCS do what it needs for the human operator with the information in ATTITUDE_QUATERNION?

In theory yes. The GCS, knowing that the current system is a tailsitter in fixed-wing mode, could rotate the attitude quaternion accordingly. However, this is not wanted by the QGC project because it would add too much airframe-specific code. For this reason, things should happen on the autopilot side.

> would a better option be to add frame information to ATTITUDE_QUATERNION as an extension

Yes :+1: . It is a good solution. The autopilot always sends "true" attitude, plus an added quaternion by which the attitude has to be rotated for user display. When this additional rotation is not needed, it can be set to [0, 0, 0, 0], so there is no bandwidth cost. From the GCS perspective, this is little additional work and should be acceptable. Let me push an update.

hamishwillee:
Or if you wanted to send it in two messages you could provide a frame field [MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME) and use the same values.

jlecoeur:
> provide a frame field MAV_FRAME and use the same fields to send the quaternion in two different messages

I do not understand this, can you explain more or give an example?

hamishwillee:
So lots of messages can take a frame field - ie see [POSITION_TARGET_LOCAL_NED.coordinate_frame](https://mavlink.io/en/messages/common.html#POSITION_TARGET_LOCAL_NED) and [SET_POSITION_TARGET_LOCAL_NED.coordinate_frame](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED). The meaning of the positional fields x,y,z etc is then relative to the frame of reference specified ([MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME)). What this allows is for you to have the same fields be used for different co-ordinate systems where they make sense. 

Now I have no idea if one of the frames can be used to specify the current default frame for ATTITUDE_QUATERNION and the new frame that you want to use for the new fields. But if one can, this is a scalable solution because it means that if later on you wanted to have another frame, you'd just modify the enum to support it. But it does mean that if you need the data in both frames, you have to send the message twice. All tradeoffs. 

Does that make sense?

jlecoeur:
I understand now. It makes sense, this is basically sending an enum value vs sending the quaternion. 

MAV_FRAME would not work, but [MAV_SENSOR_ORIENTATION](https://mavlink.io/en/messages/common.html#MAV_SENSOR_ORIENTATION) may do. I see 3 drawback. 
- There is a slight ambiguity: does the enum value represent a rotation relative to the board frame (what we want to express), or relative to the sensor frame (what MAV_SENSOR_ORIENTATION is now used for)? 
- The enum values are discrete, so it is not possible to implement smooth transition from one frame to the other: as @julianoes said, if well controlled the attitude should remain nearly horizontal throughout the tailsitter transition.
- if we send ATTITUDE_QUATERNION with different data in different frames, then the extension *has* to be interpreted by the receiver, otherwise it would receive flickering attitude. For this one I would instead send always the same attitude, with the frame field describing an additional rotation, equal to MAV_SENSOR_ROTATION_NONE (=0) by default. In the worst case the receiver does not interpret the extension field and there is the same problem as today.

That said, if an enum is the preferred solution, I will do it.

message_definitions/v1.0/common.xml
```diff
@@ -4006,6 +4006,8 @@
       <field type="float" name="rollspeed" units="rad/s">Roll angular speed</field>
       <field type="float" name="pitchspeed" units="rad/s">Pitch angular speed</field>
       <field type="float" name="yawspeed" units="rad/s">Yaw angular speed</field>
+      <extensions/>
+      <field type="float[4]" name="repr_offset_q">Rotation offset by which the attitude quaternion and angular speed vector should be rotated for user display (quaternion with [w, x, y, z] order, zero-rotation is [1, 0, 0, 0], send [0, 0, 0, 0] if uneeded). This field is intended for systems in which the reference attitude may change during flight. For example, tailsitters VTOLs rotate their reference attitude by 90 degrees between hover mode and fixed wing mode, thus repr_offset_q is equal to [1, 0, 0, 0] in hover mode and equal to [0.7071, 0, 0.7071, 0] in fixed wing mode.</field>
```
hamishwillee: uneeded > unneeded or even better, "if field not supported"

hamishwillee:
@jlecoeur Valid points. You've convinced me.

@julianoes @WickedShell @auturgy This proposal has now been updated (see source) to add a new extension field to [ATTITUDE_QUATERNION](https://mavlink.io/en/messages/common.html#ATTITUDE_QUATERNION) for the quaternion offset for user display - rather than in VFR_HUD.

The field default is zero/not supported, so nothing is broken on systems that don't want to send this or haven't been updated (addresses your point Michael). We also discussed alternative of including frame field to control interpretation, but I think there are good reasons against in this case: https://github.com/mavlink/mavlink/pull/1247#issuecomment-544480818

Anyone object to this or have suggestions for improvement.

message_definitions/v1.0/common.xml
jlecoeur: fixed

hamishwillee:
I believe this solution makes sense, or at least is not breaking/costs nothing for solutions that don't want to use it. Therefore, as everyone has had more than enough time to comment, I am merging.

Thanks @julianoes for your review, and @jlecoeur for your patience.

PR #1249, opened 2019-10-16
Clarify target address for requested message/stream

### LINKED ISSUE #1248: Pl clarify targets in stream/request messages
a component can be made to emit a message (one-shot or streaming) with `MAV_CMD_REQUEST_MESSAGE` or `MAV_CMD_SET_MESSAGE_INTERVAL`. However, if a message is requested which has a target system/component, it is not clear how the target should be set. 

Contradicting lines of arguments come to mind. One straight-forward "answer" could be that the target should be non-broadcast, i.e., set to the requester's source system/component IDs. Another straight-forward "answer" could be that the behavior should be similar to that of the non-target messages, i.e., be broadcast. One also could argue that for a `MAV_CMD_REQUEST_MESSAGE` the target should be the source while for a `MAV_CMD_SET_MESSAGE_INTERVAL` it should be broadcasting. One generally could envision situations where, especially for a stream, either a targeted or a broadcasting response would be more preferred and thus would want to have the option. And so on.

Currently the spec is unclear as to this point. I think it should not be left unspecified however, since different implementations could then behave very differently.

As food for discussion, three simple solutions:
* it is specified that the target must be set to the source
* it is specified that the target must be broadcasting
* a param is used to hold a enum which allows to choose between options

Note that this is not totally academic: For instance, the ArduPilot specific message https://mavlink.io/en/messages/ardupilotmega.html#MOUNT_STATUS has a target, and is (at least currently) a main incredient of the gimbal support. It is not clear what the expected behavior should be.

hamishwillee:
Fixes #1248.

As discussed in #1248 many messages are intended to be broadcast and if you request them, they will be broadcast. However some messages have target address in them so can either be broadcast or sent to a specific target.

This update repurposes parameters such that a user can explicitly specify that the target address of the response is broadcast or the address of the component that requested the message/stream. 

By default the message will be sent with 0, which is "whatever the flight stack wants to do".

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1831,11 +1831,13 @@
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM.</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
         <param index="2" label="Interval" units="us" minValue="-1" increment="1">The interval between two messages. Set to -1 to disable and 0 to request default rate.</param>
+        <param index="4" label="Response Target" minValue="0" maxValue="2" increment="1">Target address of message stream (if message has target address fields). 0: Flight-stack default (recommended), 1: address of requestor, 2: broadcast.</param>
       </entry>
       <entry value="512" name="MAV_CMD_REQUEST_MESSAGE" hasLocation="false" isDestination="false">
         <description>Request the target system(s) emit a single instance of a specified message (i.e. a "one-shot" version of MAV_CMD_SET_MESSAGE_INTERVAL).</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID of the requested message.</param>
         <param index="2" label="Index ID" minValue="0" increment="1">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
+        <param index="4" label="Response Target" minValue="0" maxValue="2" increment="1">Target address for requested message (if message has target address fields). 0: Flight-stack default, 1: address of requestor, 2: broadcast.</param>
       </entry>
       <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION" hasLocation="false" isDestination="false">
         <description>Request MAVLink protocol version compatibility</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1831,13 +1831,13 @@
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM.</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
         <param index="2" label="Interval" units="us" minValue="-1" increment="1">The interval between two messages. Set to -1 to disable and 0 to request default rate.</param>
-        <param index="4" label="Response Target" minValue="0" maxValue="2" increment="1">Target address of message stream (if message has target address fields). 0: Flight-stack default (recommended), 1: address of requestor, 2: broadcast.</param>
+        <param index="7" label="Response Target" minValue="0" maxValue="2" increment="1">Target address of message stream (if message has target address fields). 0: Flight-stack default (recommended), 1: address of requestor, 2: broadcast.</param>
       </entry>
       <entry value="512" name="MAV_CMD_REQUEST_MESSAGE" hasLocation="false" isDestination="false">
         <description>Request the target system(s) emit a single instance of a specified message (i.e. a "one-shot" version of MAV_CMD_SET_MESSAGE_INTERVAL).</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID of the requested message.</param>
         <param index="2" label="Index ID" minValue="0" increment="1">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
-        <param index="4" label="Response Target" minValue="0" maxValue="2" increment="1">Target address for requested message (if message has target address fields). 0: Flight-stack default, 1: address of requestor, 2: broadcast.</param>
+        <param index="7" label="Response Target" minValue="0" maxValue="2" increment="1">Target address for requested message (if message has target address fields). 0: Flight-stack default, 1: address of requestor, 2: broadcast.</param>
       </entry>
       <entry value="519" name="MAV_CMD_REQUEST_PROTOCOL_VERSION" hasLocation="false" isDestination="false">
         <description>Request MAVLink protocol version compatibility</description>


message_definitions/v1.0/common.xml
```diff
@@ -1831,11 +1831,13 @@
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM.</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
         <param index="2" label="Interval" units="us" minValue="-1" increment="1">The interval between two messages. Set to -1 to disable and 0 to request default rate.</param>
+        <param index="4" label="Response Target" minValue="0" maxValue="2" increment="1">Target address of message stream (if message has target address fields). 0: Flight-stack default (recommended), 1: address of requestor, 2: broadcast.</param>
```
hamishwillee: @julianoes 
1. I used index 4 rather than 3 to give space for another user-specified value. Why not 6 - dunno!
2. I used numbers rather than an enum because it didn't seem worth creating one for what is probably not going to be a user-selectable value.

olliw42:
please provide 4 options

0: Flight-stack default
1: system and component ID of requestor
2: system ID of requestor and component ID broadcasting 
3: both system and component ID broadcasting

I also would not indicate 0 as recommended. The most recommended values would be 1 or 2. 0 is there "only" for backwards compatibility.

I personally would prefer an enum, but I see your reasoning.

thx for taking the action.

message_definitions/v1.0/common.xml
olliw42: please provide 4 options

0: Flight-stack default
1: system and component ID of requestor
2: system ID of requestor and component ID broadcasting
3: both system and component ID broadcasting

I also would not indicate 0 as recommended. The most recommended values would be 1 or 2. 0 is there "only" for backwards compatibility.

I personally would prefer an enum, but I see your reasoning.

btw: not 6 in order to have a better chance to take advantage of trailing zeros (I know, there is the target and confirmation, yet...)

hamishwillee: OK, how about (modified in line with above comments). Note, not recommendation, though IMO in practice you'd be fine to accept the flight stack default in 99.9% of cases.

> Target system/component of message stream (only if message has target fields). 0: Flight-stack default, 1: Requestor system/component ID, 2: Requestor system ID and broadcast component ID, 3: Broadcast system and component ID.

olliw42: looks great

I think it mostly will depend on when flight stacks would start to support the option ...

message_definitions/v1.0/common.xml
```diff
@@ -1831,11 +1831,13 @@
...
       </entry>
       <entry value="512" name="MAV_CMD_REQUEST_MESSAGE" hasLocation="false" isDestination="false">
         <description>Request the target system(s) emit a single instance of a specified message (i.e. a "one-shot" version of MAV_CMD_SET_MESSAGE_INTERVAL).</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID of the requested message.</param>
         <param index="2" label="Index ID" minValue="0" increment="1">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
+        <param index="4" label="Response Target" minValue="0" maxValue="2" increment="1">Target address for requested message (if message has target address fields). 0: Flight-stack default, 1: address of requestor, 2: broadcast.</param>
```
hamishwillee: @julianoes @olliw42 I'm thinking about moving this to the final index and defining all these index values as "optional". My reasoning being that I have a message that has 3 possible params :-0. All  would be zero by default.

OK?

hamishwillee:
@julianoes It also occurs to me that I haven't properly got my head around the expected [COMMAND_ACK](https://mavlink.io/en/messages/common.html#COMMAND_ACK). 

The success case is fine - as usual we just return [MAV_RESULT_ACCEPTED](https://mavlink.io/en/messages/common.html#MAV_RESULT_ACCEPTED). 

I'm just not sure what we return if `MAV_CMD_REQUEST_MESSAGE` is supported but the particular message we want to act on is not. i.e. I'd like to return `MAV_RESULT_UNSUPPORTED` but normally this would be be interpreted as `MAV_CMD_REQUEST_MESSAGE` is not supported. Or am I over-thinking this?

message_definitions/v1.0/common.xml
julianoes: I think it would be weird to have it in the last param. Shouldn't it just be at index="3"?

julianoes:
> The success case is fine - as usual we just return MAV_RESULT_ACCEPTED.

> I'm just not sure what we return if MAV_CMD_REQUEST_MESSAGE is supported but the particular message we want to act on is not. i.e. I'd like to return MAV_RESULT_UNSUPPORTED but normally this would be be interpreted as MAV_CMD_REQUEST_MESSAGE is not supported. Or am I over-thinking this?

Good question. I would say.
`MAV_RESULT_ACCEPTED`: default, all is good.
`MAV_RESULT_UNSUPPORTED`: according to spec "Command UNKNOWN/UNSUPPORTED"
`MAV_RESULT_DENIED`: not possible for this message because there are reasons
`MAV_RESULT_FAILED`: not possible because something went wrong

hamishwillee:
@julianoes So what do we return if MAV_CMD_REQUEST_MESSAGE  is supported but someone requests a message be streamed?

message_definitions/v1.0/common.xml
hamishwillee: @julianoes Perhaps I wasn't very clear. 

So we currently have an optional configurable parameter (2). In future we might have multiple optional configurable parameters and having them surrounding param 3 is even more odd. For example consider the command I suggested in the Microservices versioning, which has 3 params - service, max version, min version. It would seem odd to have param 2, 4, 5, as configurable and 3 as the response target. 

The "nice" think to do here would be to make param 2 the response target, and make the optional ones from 3. But given 3 has been defined for a while I was thinking we might explicitly define param 3, 4,5 6, as configurable params, and just have 7 as the response target. 

Does that make sense?

olliw42:
very good question ... 

NOTE: This situation doesn't only apply/occur for this command but in fact is quite general in nature, i.e., a similar situation can happen for other cmds too. That is, is could and should be addressed by a generally valid specification.  !!

As suggestion, here the pattern I decided to follow in my implementation:

* MAV_RESULT_ACCEPTED: Command ACCEPTED and EXECUTED
* MAV_RESULT_TEMPORARILY_REJECTED: Command was accepted, but could currently not be executed. Implementations supports this command, so it can make sense to retry later.
* MAV_RESULT_DENIED: Command was accepted, but cannot be executed, e.g. because of inappropriate params. It doesn't make sense to retry as outcome will be the same.
* MAV_RESULT_UNSUPPORTED | Command UNKNOWN/UNSUPPORTED
* MAV_RESULT_FAILED | Command executed, but failed. It can make sense to retry.
* MAV_RESULT_IN_PROGRESS

message_definitions/v1.0/common.xml
olliw42: I think it depends on how many more commands there are for which such a need of specifying the target of a returned message is needed. From scratch I only see the two commands under discussion.
If it is really just few it's maybe overkill to wrap a general strategy around them, and I would think any place might be considered as good as any other. My natural feeling would be to just fill from bottom to top. But anything would IMHO be equally good.

julianoes:
> So what do we return if MAV_CMD_REQUEST_MESSAGE is supported but someone requests a message that can't be sent?

MAV_RESULT_DENIED I'd say.

hamishwillee:
@olliw42 @julianoes You're both right. I'll try clarify the specification here: https://github.com/mavlink/mavlink/pull/1253 and `MAV_RESULT_DENIED` would be used for a request to send back a message that hasn't been set up for this purpose.

message_definitions/v1.0/common.xml
hamishwillee: I'm going to do it as I proposed and see if the dev call or anyone else objects strongly.

olliw42:
the MAV_RESULT_XXX topic has been settled, so this doesn't need to be on hold anymore and could be settled too, right?

hamishwillee:
I changed the param for the response target to 7.  Otherwise IMO this is good to merge. @julianoes Any last comments?

hamishwillee:
Done. Thanks all.

PR #1250, opened 2019-10-16
Add setpoint types for position setpoint [DO NOT MERGE]

hamishwillee:
It is useful to specify the setpoint type (e.g. land, takeoff) in calls like `SET_POSITION_TARGET_LOCAL_NED` so that you can define additional behaviour - e.g. deploy landing gear, if you intend to land.

This adds the information as an enum in an extension field to all the position target messages. Default is "normal" and 0, so this change will not affect existing implementations.

**EDITED** Original version showed PX4 implementation - which silently/privately added unused values to the `type_mask`  (as discussed [here](https://github.com/PX4/Firmware/pull/13199#issuecomment-542569388)). The new approach is cleaner, and avoids the inconsistency that type_mask is a bitmask (flags) while setpoints are mutually exclusive values

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3432,7 +3432,7 @@
         <description>Ignore yaw rate</description>
       </entry>
       <entry value="4096" name="POSITION_TARGET_TYPEMASK_TAKEOFF_SETPOINT">
-        <description>Position setpoint is a target setpoint.</description>
+        <description>Position setpoint is a takeoff setpoint.</description>
       </entry>
       <entry value="8192" name="POSITION_TARGET_TYPEMASK_LAND_SETPOINT">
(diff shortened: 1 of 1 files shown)


hamishwillee:
FYI,

julianoes:
I find the `type_mask` difficult enough as such and would therefore vote not to clutter it with additional information. I agree that in that case an extension might be more appropriate.

hamishwillee:
@julianoes Thanks. So in next revision of this PR I'll create an extension.

Why is knowing the type of setpoint useful?

hamishwillee:
@julianoes @auturgy OK, updated this to include the position setpoint type in a new enum. Any objections?

hamishwillee:
Closing. Stale. Probably not needed.

PR #1255, opened 2019-10-22
PLAY_TUNE: Specify default format

### LINKED ISSUE #1251: PLAY_TUNE message should be non-ambiguous
I am not so convinced about the [PLAY_TUNE](https://mavlink.io/en/messages/common.html#PLAY_TUNE) message in the specs. It reads:

```
target_system    | uint8_t   | System ID
--               | --        | --
target_component | uint8_t   | Component ID
tune             | char[30]  | tune in board specific format
tune2 **         | char[200] | tune extension (appended to tune)
```

My point here is that `tune in board specific format` is a "non-spec": it says "we have a message that says that there is no standard way to do that".

What PX4 implements is using the [QBasic Play](https://www.qbasic.net/en/reference/qb11/Statement/PLAY-006.htm) syntax:

```
Octave and tone commands:
 Ooctave    Sets the current octave (0 - 6).
 < or >     Moves up or down one octave.
 A - G      Plays the specified note in the current octave.
 Nnote      Plays a specified note (0 - 84) in the seven-octave
            range (0 is a rest).

Duration and tempo commands:
 Llength    Sets the length of each note (1 - 64). L1 is whole note,
            L2 is a half note, etc.
 ML         Sets music legato.
 MN         Sets music normal.
 MS         Sets music staccato.
 Ppause     Specifies a pause (1 - 64). P1 is a whole-note pause,
            P2 is a half-note pause, etc.
 Ttempo     Sets the tempo in quarter notes per minute (32 - 255).

Mode commands:
 MF          Plays music in foreground.
 MB          Plays music in background.

Suffix commands:
 # or +      Turns preceding note into a sharp.
 -           Turns preceding note into a flat.
 .           Plays the preceding note 3/2 as long as specified.
```

And it makes perfect sense to me. I believe that the specs should read: "tune in the qbasic play format". That would make it unambiguous.

hamishwillee:
Fixes #1251

This adds a field that specifies the tune format, and sets the default value (0) as QBasic (the format used by both ArduPilot and PX4).

This should not break existing systems that use another format (although they will be non-compliant). For example, ArduPilot supports another format, but gets the matching firmware version out of band.

The supported formats can be extended by adding new format field values. The cost would be that the tune buffer would no longer be truncated.
If a new format is added we might consider creating a new tune message (along with enum for the formats) and a mechanism to query the format).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5416,12 +5416,13 @@
       <field type="uint8_t" name="state" display="bitmask">Bitmap for state of buttons.</field>
     </message>
     <message id="258" name="PLAY_TUNE">
-      <description>Control vehicle tone generation (buzzer)</description>
+      <description>Control vehicle tone generation (buzzer). Default format is QBasic 1.1 Play Format: https://www.qbasic.net/en/reference/qb11/Statement/PLAY-006.htm. </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="char[30]" name="tune">tune in board specific format</field>
+      <field type="char[30]" name="tune">Tune in format specified in the field: format</field>
       <extensions/>
-      <field type="char[200]" name="tune2">tune extension (appended to tune)</field>
+      <field type="char[200]" name="tune2">Tune extension (appended to tune field)</field>
+      <field type="uint8_t" name="format">Tune format. 0 (default) indicates QBasic 1.1 Play Format.</field>
     </message>
     <message id="259" name="CAMERA_INFORMATION">
       <description>Information about a camera</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@JonasVautherin Compromise in #1251. 
At the point people actually decide they _really_ need another tune format, we might reasonably create a new message(s)

JonasVautherin:
I don't understand what the format extension brings. For autopilots that already exist and use a different format, they will keep on not using the format extension, and that won't break their system.

Couldn't we live without the "format" extension, and think about adding it if one day somebody comes with a reason for not using the same format as APM and PX4?

hamishwillee:
It doesn't really bring anything - the right thing to do if you break something is to add a new and sane message. But it made me feel better.

JonasVautherin:
Closing following discussions in #1251.

PR #1256, opened 2019-10-24
PLAY_TUNE_V2: New tune messages

hamishwillee:
This adds new definitions:
- play a tune: PLAY_TUNE_V2.
- query what tune formats are supported: SUPPORTED_TUNES
- enum for known tuneformats QBasic 1.1 and Modern MML
It also deprecates PLAY_TUNE

With this a GCS/SDK can query a to work out what tune/tunes. Default format is assumed to be QBasic 1.1.

Fixes https://github.com/mavlink/mavlink/issues/1251.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3802,6 +3802,15 @@
         <description>CAA (Civil Aviation Authority) registered operator ID.</description>
       </entry>
     </enum>
+    <enum name="TUNE_FORMAT">
+      <description>Tune formats (used for vehicle buzzer/tone generation).</description>
+      <entry value="0" name="TUNE_FORMAT_QBASIC1_1">
+        <description>Format is QBasic 1.1 Play: https://www.qbasic.net/en/reference/qb11/Statement/PLAY-006.htm.</description>
+      </entry>
+      <entry value="0" name="TUNE_FORMAT_MML_MODERN">
+        <description>Format is Modern Music Markup Language (MML): https://en.wikipedia.org/wiki/Music_Macro_Language#Modern_MML.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="0" name="HEARTBEAT">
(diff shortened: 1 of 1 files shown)


hamishwillee:
FYI, @auturgy says he is OK with this.

message_definitions/v1.0/common.xml
```diff
@@ -5842,6 +5852,19 @@
...
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t" name="format" enum="TUNE_FORMAT">Tune format</field>
+      <field type="char[252]" name="tune">Tune definition</field>
+    </message>
+    <message id="401" name="SUPPORTED_TUNES">
+      <description>Tune formats supported by vehicle. This should be emitted as response to MAV_CMD_REQUEST_MESSAGE.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t[200]" name="format" enum="TUNE_FORMAT">Array of supported tune formats</field>
```
julianoes: Why would this be 200? That means when you want to extend this message, you need to send 200 empty bytes. 10 would have been enough.

julianoes:
Why was this merged so quickly? I had not even reviewed it.

LorenzMeier:
Because I've reviewed it and because James was happy with it. That gives it coverage from the maintainer and from all affected constituencies.

julianoes:
Fine, I'll make a follow up PR then.

message_definitions/v1.0/common.xml
hamishwillee: Yes. Indeed right now this is really intended only for emission by an autopilot, which will probably only support 1. Also this is missing an easy way to know the number of tunes in the array - you can't even assume 0 is the last one, since 0 is valid. We should put a field in for that too. Given this is early PR can break IMO.

message_definitions/v1.0/common.xml
```diff
@@ -5842,6 +5852,19 @@
       <field type="uint32_t[6]" name="link_tx_max" units="KiB/s">Network capacity from the component system. A value of UINT32_MAX implies the field is unused.</field>
       <field type="uint32_t[6]" name="link_rx_max" units="KiB/s">Network capacity to the component system. A value of UINT32_MAX implies the field is unused.</field>
     </message>
+    <message id="400" name="PLAY_TUNE_V2">
+      <description>Play vehicle tone/tune (buzzer). Default format is QBasic 1.1. Supersedes message PLAY_TUNE.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t" name="format" enum="TUNE_FORMAT">Tune format</field>
+      <field type="char[252]" name="tune">Tune definition</field>
```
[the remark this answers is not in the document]
hamishwillee: There should be a mechanism of specifying the end of this if 0 at the end is potentially a valid value (which it may be). Continue discussion on PR.

hamishwillee:
Apologies all. I pushed too hard for this - it is important we do the right thing, but by no means urgent. I have created #1258 to fix these.

Issue #919, opened 2018-05-22
Add versioning support in the extensions mechanism


hamishwillee:
The extensions tag in XML allows you to specify that a message is for MAVLink 2, and hence should not be included in the code generation for MAVLink 1. 

What about if we iterate to MAVLink 3 etc? Does it make sense to use this tag for more fine-grained versioning, and allowing multiple extension levels?

E.g. 

```
<extensions/> //in from MAVLink 2, inclusive
... messages ...
<extensions in="2.0" out="4.0"  />`   ///generated into output from 2.0 to all versions before 4.0
... Some messages ...
<extensions in="3.0" out="4.0"  /> 
... Some messages ...
```
I tend to think in this case we might be better of grouping using start and end tags.

hamishwillee:
Arguably, as suggested [here](https://github.com/mavlink/mavlink/issues/795#issue-273674681) you might better add extension info to each field
```
<message id="152" name="MEMINFO">
    <description>state of APM memory</description>
    <field name="brkval" type="uint16_t">heap top</field>
    <field name="freemem" type="uint16_t">free memory</field>
    <field name="extrafield1" type="uint8_t" extension_start="mav2">extension1</field>
    <field name="extrafield2" type="uint32_t" extension_start="mav2 extension_end="mav4">extension2</field>
</message>
```
or 
```
<message id="152" name="MEMINFO">
    <description>state of APM memory</description>
    <field name="brkval" type="uint16_t">heap top</field>
    <field name="freemem" type="uint16_t">free memory</field>
    <field name="extrafield1" type="uint8_t" extension="true">extension1</field>
    <field name="extrafield2" type="uint32_t" extension="true">extension2</field>
</message>
```
If you wanted to go one step further, it might be nice to add an attribute to the <extensions> node indicating the minimum version for which the extensions are applicable (i.e. <extensions version="2">) and then using multiple <extensions> nodes to group child <field>s by minimum version – or use the extension attribute of the <field> node to contain that information.

Just my two cents – hoping to save someone else this headache! Thanks for all y

hamishwillee:
Closing. Too many things to do to address possible fixes to possible future problems.

Issue #795, opened 2017-11-14
Protocol definition XML format structural enhancement


kenkeiter:
Hi folks –
Given the random/broken state of the Go language implementations of MAVLink, I began work on a new one that would "catch up" (so to speak) with the current state of the protocol, do things somewhat more idiomatically, provide better documentation, generally _work_, etc. 

Things were going _mostly_ swimmingly until I went to implement a particular item on my checklist:  MAVLink 2 extended messages. The problem is as follows: the XML format currently used to identify extended message parameters does not annotate the extended messages in any meaningful way _unless you are parsing them sequentially._ Most XML parsers allow you to use something like XPath (or in Go's case, define `structs` with XPath-ish annotations, and parse data into them reflectively) to select nodes by their attributes, children, ancestors, etc. By injecting the `<extensions/>` node as a sibling of the `<field>` nodes, the parser must maintain state in order to identify a `<field>` as being an extension. While it's not a show-stopper for me (I got around it by rewriting it with a regex), it's probably going to limit others that wish to follow in my footsteps, and it's not semantically-valid XML.

Currently, the XML definition looks like this (pulled from [here](https://github.com/mavlink/mavlink/blob/master/doc/MAVLink2.md)):

```xml
  <message id="152" name="MEMINFO">
    <description>state of APM memory</description>
    <field name="brkval" type="uint16_t">heap top</field>
    <field name="freemem" type="uint16_t">free memory</field>
    <extensions/>
    <field name="extrafield1" type="uint8_t">extension1</field>
    <field name="extrafield2" type="uint32_t">extension2</field>
  </message>
```

I would suggest moving to a format that annotates the fields either directly or hierarchically. A hierarchical annotation might look like the following:

```xml
<message id="152" name="MEMINFO">
    <description>state of APM memory</description>
    <field name="brkval" type="uint16_t">heap top</field>
    <field name="freemem" type="uint16_t">free memory</field>
    <extensions>
        <field name="extrafield1" type="uint8_t">extension1</field>
        <field name="extrafield2" type="uint32_t">extension2</field>
    </extensions>
</message>
```

while a direct annotation could simply be the addition of an attribute to the field, like so:

```
<message id="152" name="MEMINFO">
    <description>state of APM memory</description>
    <field name="brkval" type="uint16_t">heap top</field>
    <field name="freemem" type="uint16_t">free memory</field>
    <field name="extrafield1" type="uint8_t" extension="true">extension1</field>
[...]

PR #1257, opened 2019-10-25
Created XSD - Improved XML for consistency

joe-sonrichard:
## Problem:
Was trying to import the common.xml into a node database and discovered that there are inconsistencies within the document. For example, most enum entries specify the value. However, a few don't. 

## Solution:
I auto-generated XSD for common.xml via https://www.freeformatter.com/ and made it stricter to spot abnormalities. I then modified the xml files to bring a greater level of consistency. I did not intend to make any significant changes to meaning, only to make the implied meanings more explicit as in the case of enums with implied but undefined values.

### common.xml
1. Specified all unspecified enum values assuming a 0 index.
2. Changed `<wip/> <!-- some comment -->` to `<wip> some comment <wip/>`.
3. Removed `<extension/>` because (I don't think) it is meaningful as used. (Do we want to make the fields following `<extension/>` into sub-elements? - didn't do that because I didn't want to assume that's what it meant.)

### All other xml files
1. Made all of the above changes as necessary to bring into alignment with XSD.
2. Added version and dialect fields where missing.

## Value Added
1. Existing XML documents are more consistent reducing number of cases to address in code-generation and thereby reducing potential causes for bugs.
2. Future XML additions (new files or changes to existing) may be checked against XSD for conformity. (Or XSD may be improved)


amilcarlucas:
There is already an hand written xsd file, and xml changes are validated against it already.
The file exists in https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd

I understand your confusion regarding the extensions. It was a bad decision to implement it like that and there are two issues that address this: https://github.com/mavlink/mavlink/issues/795 and https://github.com/mavlink/mavlink/issues/919

Changing readable hexadecimal numbers into strange decimal numbers is not an improvement.

But adding missing `version` and `dialect` information is

hamishwillee:
Thanks - any effort to make things more consistent and robust welcome. This doesn't all work though. Please keep the enum and wip changes. Dump the rest. 

The final test of a change like this is to run the test code on the output XSD and also to make sure that the generated libraries are identical. This breaks any message that has an extension. It also breaks the versioning (though IMO that isn't particularly useful anyway). Detail below. 


> Specified all unspecified enum values assuming a 0 index.

The existing generators auto-assign from 0. I like this change because it is _possible_ to break compatibility accidentally the way things are currently. 

> Changed <wip/>  to <wip> some comment <wip/>.

No strong opinion. I wouldn't have bothered, but if it is a scripted change and doesn't break the XSD then great. 

> Removed <extension/> because (I don't think) it is meaningful as used. (Do we want to make the fields following <extension/> into sub-elements? - didn't do that because I didn't want to assume that's what it meant.)

**REVERT**. The [toolchain](https://mavlink.io/en/guide/define_xml_element.html#message_extensions) needs this information in order to control how messages are packaged for serial transmission, and to ensure they are not included in CRC calculation.

> Added version and dialect fields where missing.

REVERT the version field. The generators use the last included version for heartbeat version information. For a dialect based off common.xml there shouldn't be any version in the dialect - the dialect should iterate with common.
Dialects that don't include command can do what they like. 

For dialect tag it doesn't matter. As far as I can tell the information is not used by any toolchain. But if they are assigned they must be unique.

> Changing readable hexadecimal numbers into strange decimal numbers is not an improvement.

FYI the numbers are in hex/binary usually because to make binary flags more obvious.

hamishwillee:
@joe-sonrichard Thanks for this, as discussed, it can't be merged as is. I'm going to close it, but would be happy for you to reopen and update this.

PR #1258, opened 2019-10-27
PLAY_TUNE_V2 SUPPORTED_TUNES - WIP and minor fixes

hamishwillee:
Apologies. This should have been WIP until first implementation, and needs some changes.

As this went in only on Friday I'm making changes in breaking way. 

* [x] SUPPORTED_TUNES 
  - Mark as WIP
  - Include array length
  - Shrink array to more realistic length to allow extension if needed. 
* [x] PLAY_TUNE_V2
  - indicate end of tune in char array
  - Mark as WIP

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5858,7 +5858,7 @@
       <description>Play vehicle tone/tune (buzzer). Supersedes message PLAY_TUNE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint16_t" name="format" enum="TUNE_FORMAT">Tune format</field>
+      <field type="uint32_t" name="format" enum="TUNE_FORMAT">Tune format</field>
       <field type="char[252]" name="tune">Tune definition as a NULL-terminated string.</field>
     </message>
     <message id="401" name="SUPPORTED_TUNES">
@@ -5867,7 +5867,7 @@
       <description>Tune formats supported by vehicle. This should be emitted as response to MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint16_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
+      <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5858,7 +5858,7 @@
       <description>Play vehicle tone/tune (buzzer). Supersedes message PLAY_TUNE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint8_t" name="format" enum="TUNE_FORMAT">Tune format</field>
+      <field type="uint16_t" name="format" enum="TUNE_FORMAT">Tune format</field>
       <field type="char[252]" name="tune">Tune definition as a NULL-terminated string.</field>
     </message>
     <message id="401" name="SUPPORTED_TUNES">
@@ -5867,7 +5867,7 @@
       <description>Tune formats supported by vehicle. This should be emitted as response to MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint8_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
+      <field type="uint16_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5859,7 +5859,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="format" enum="TUNE_FORMAT">Tune format</field>
-      <field type="char[252]" name="tune">Tune definition.</field>
+      <field type="char[252]" name="tune">Tune definition as a NULL-terminated string.</field>
     </message>
     <message id="401" name="SUPPORTED_TUNES">
       <wip />
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3804,10 +3804,10 @@
     </enum>
     <enum name="TUNE_FORMAT">
       <description>Tune formats (used for vehicle buzzer/tone generation).</description>
-      <entry value="0" name="TUNE_FORMAT_QBASIC1_1">
+      <entry value="1" name="TUNE_FORMAT_QBASIC1_1">
         <description>Format is QBasic 1.1 Play: https://www.qbasic.net/en/reference/qb11/Statement/PLAY-006.htm.</description>
       </entry>
-      <entry value="1" name="TUNE_FORMAT_MML_MODERN">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5853,17 +5853,20 @@
       <field type="uint32_t[6]" name="link_rx_max" units="KiB/s">Network capacity to the component system. A value of UINT32_MAX implies the field is unused.</field>
     </message>
     <message id="400" name="PLAY_TUNE_V2">
+      <wip />
       <description>Play vehicle tone/tune (buzzer). Default format is QBasic 1.1. Supersedes message PLAY_TUNE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="format" enum="TUNE_FORMAT">Tune format</field>
-      <field type="char[252]" name="tune">Tune definition</field>
+      <field type="char[252]" name="tune">Tune definition.</field>
```
hamishwillee: Re https://github.com/mavlink/mavlink/pull/1256#discussion_r339370049 @julianoes What's the best way to mark the end of the string, in particular when 0 might potentially be a valid end character in some string formats?  
Can you update this directly rather than tell me and waiting on response?

message_definitions/v1.0/common.xml
```diff
@@ -5853,17 +5853,20 @@
...
     </message>
     <message id="401" name="SUPPORTED_TUNES">
+      <wip />
       <description>Tune formats supported by vehicle. This should be emitted as response to MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint8_t[200]" name="format" enum="TUNE_FORMAT">Array of supported tune formats</field>
+      <field type="uint8_t" name="formats_supported">Number of entries in the array tune array. 0: Only the default tune format is supported (TUNE_FORMAT_QBASIC1_1).</field>
```
hamishwillee: @julianoes This is most efficient way for the default to be 0 and for caller not to have to potentially iterate whole array.

message_definitions/v1.0/common.xml
hamishwillee: My thinking is that the easiest way is just to have a length field.

julianoes: Good point. If 0 is a valid character of the format then we can't discuss `\0` char termination.

You are implying that the format might not be a string/ascii format but could be binary as well, however in that case it should be `uint8_t[252` because `char` usually refers to strings.

message_definitions/v1.0/common.xml
julianoes: I think that's fine like that, yes. And 10 is more than enough. If we need more 3 ever I'd already be very surprised :smile:.

message_definitions/v1.0/common.xml
hamishwillee: I guess it might be binary. But I'd rather keep it char because both our formats that exist are char. If someone wants to do binary format, they could pack it in char too. 

Are you ok with field for the length?

message_definitions/v1.0/common.xml
hamishwillee: Me too :-)

olliw42: it really doesn't do any harm to have an array bigger than 10, thx to M2
so I would go way beyond any imaginable, e.g. 32 or 64 or so
maybe some day someone does a mavlink-tune-player component supporting all tunes ... one never knows LOL

hamishwillee: I don't know what M2 is. But zero byte truncation only happens if the array is last, so if it is very big and we ever have to extend it we're wasting more bandwidth. Either way, when we get to a system that wants to have more tunes I'm happy to extend.

julianoes: Exactly @hamishwillee

message_definitions/v1.0/common.xml
julianoes: A length doesn't really solve it because if you put length to 252 the user of the API still needs to copy the buffer in order to add a zero termination at the end of it.

I would suggest that it needs to be zero terminated and that way you can just assume that `tune[251]` can be set to `\0`. I know this means you're wasting one byte in the transfer but it means you don't need 253 bytes of allocation on the API user side.

message_definitions/v1.0/common.xml
olliw42: M2 was meant to indicate MAVlink2 :)

the argument is of course correct, but it applies to ANY mavlink message with long arrays, so bites quite generally. Also, this is a very low frequency message, so bandwidth might not be the greatest concern, especially as it is still a relatively short message in comparison to many other. 

Also note that argumentation is inconsistent: Argument 1 says that the array should be kept small since the message might be extended, and Argument 2 says that if the array ever needs to be extended one happily can do so. However, if the message is extended one also can't extend the size of the array anymore; the very moment one extends the message one also has set the array size in stone.

I actually wonder why one isn't using a bit flag instead of this array. Could support 32 tones with just a uint32. And also would avoid having to have the length field. 

So, if one is concerned about bandwidth ... :):):)

julianoes: Of course a bitflag! That would be way nicer :man_facepalming:.

message_definitions/v1.0/common.xml
```diff
@@ -5853,17 +5853,20 @@
       <field type="uint32_t[6]" name="link_rx_max" units="KiB/s">Network capacity to the component system. A value of UINT32_MAX implies the field is unused.</field>
     </message>
     <message id="400" name="PLAY_TUNE_V2">
+      <wip />
```
amilcarlucas: Please add the XML comment that expaind what wip is

message_definitions/v1.0/common.xml
```diff
@@ -5853,17 +5853,20 @@
...
       <description>Play vehicle tone/tune (buzzer). Default format is QBasic 1.1. Supersedes message PLAY_TUNE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="format" enum="TUNE_FORMAT">Tune format</field>
-      <field type="char[252]" name="tune">Tune definition</field>
+      <field type="char[252]" name="tune">Tune definition.</field>
     </message>
     <message id="401" name="SUPPORTED_TUNES">
+      <wip />
```
amilcarlucas: Same as above

message_definitions/v1.0/common.xml
hamishwillee: Thanks. Done.

message_definitions/v1.0/common.xml
hamishwillee: Sure. Changed to "Tune definition as a NULL-terminated string."

hamishwillee:
Good idea on the bitmask @olliw42 . Decided 16 enough, but we can revisit until this is implemented if there really is a use case.

@julianoes Looks great. The [field reordering](https://mavlink.io/en/guide/serialization.html#field_reordering) means that the tune definition is reordered to the end, so we don't need to worry about it not being truncated. 

I'll merge as discussed once it passes checks.

olliw42:
I'm confused.

Why are the enums now starting with 1, and not with 0 as before?

This change would have been needed if one would have had an array of entries an 0 would be the end marker. With the bit mask none of that is needed, and 0 indicates the first bit, like (1 << 0).

due to field reordering extending to 32 bits would be a breaking change, i.e., can't be done easily if one would want to do it later. I'm not sure I understand the effort to keep the message size ultra-dense for this ultra-low frequency message given how wasteful many other messages are designed.

In PLAY_TUNE_V2 the format can and should be uint8 as before. It is not a bit mask but a enum value. And since the bitmask is limited to 16 (or 32) the max value can only be 16 (or 32), so uint8 is really plenty.

IMHO

julianoes:
@olliw42 the enum value is the bitmask value.

- QBASIC1.1 is bit 0, so 1.
- MML Modern is bit 1, so 2.
- The next gen polyphonic ringtone pro is bit 2, so 4.

olliw42:
I never have used bit masks that way, and I never have seen bit masks being used that way.
As I said, (1<<0) is bit 1, and the natural usage of the enum field to check the mask would be (1 << TUNE_FORMAT_QBASIC1_1),  and not (1 << (TUNE_FORMAT_QBASIC1_1-1))

I have to admit though that I have not double checked with the other MAVLink messages where a bit mask is used. Let's see ...

hamishwillee:
Changed to 32 bits. I really don't think needed, but not concerned about these bytes for this message either. Will merge as soon as status pass.

PR #1263, opened 2019-11-05
[WIP] MAV_BATTERY_FUNCTION: Clarify values

hamishwillee:
This is created for _discussion_ of [BATTERY_STATUS.battery_function](https://mavlink.io/en/messages/common.html#BATTERY_STATUS) and more specifically of [MAV_BATTERY_FUNCTION](https://mavlink.io/en/messages/common.html#MAV_BATTERY_FUNCTION).

>**Note** Also "function" appears to be omitted from [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO). Should this not be added?

This discussion arises from https://github.com/mavlink/mavlink-devguide/pull/218#issuecomment-549240327

The design assumption is that `BATTERY_STATUS` is emitted by an **autopilot** for each battery that is part of the system. So a vehicle might have one battery for everything which has `MAV_BATTERY_FUNCTION_ALL` or a different battery for supporting different operations - e.g. payload, avionics, propulsion. 

But what if you have a component that has its own battery, for which it emits its own battery status - e,g, a gopro camera.
1. The main battery for the whole system is no longer _really_ `MAV_BATTERY_FUNCTION_ALL` because the gopro has its own battery.
2. From the autopilot perspective the battery would be `MAV_BATTERY_TYPE_PAYLOAD`  but from the camera perspective (which is emitting the type) the battery powers everything - e.g. `MAV_BATTERY_FUNCTION_ALL` .
   - An implication of this is that a camera might need to emit `MAV_BATTERY_TYPE_PAYLOAD` if in a vehicle and `MAV_BATTERY_FUNCTION_ALL` if standalone. 
3. For a component sending the message the function is redundant - a GCS can infer the function from the component sending the message.

Obvious options:
- Specify that a component other than autopilot should specify `MAV_BATTERY_FUNCTION_ALL` (this is kind of what the PR does at the moment).
- Add a type `MAV_BATTERY_FUNCTION_CAMERA` for a autopilot to emit for a camera or for a camera to emit for itself? 
- Also we should specify that `MAV_BATTERY_FUNCTION_ALL` is "all functions other than ones covered by an explicit battery type for the same system"

What else could we do?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3364,7 +3364,7 @@
         <description>Battery function is unknown</description>
       </entry>
       <entry value="1" name="MAV_BATTERY_FUNCTION_ALL">
-        <description>Battery supports all flight systems</description>
+        <description>Battery supports all systems</description>
       </entry>
       <entry value="2" name="MAV_BATTERY_FUNCTION_PROPULSION">
         <description>Battery for the propulsion system</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3364,7 +3364,7 @@
         <description>Battery function is unknown</description>
       </entry>
       <entry value="1" name="MAV_BATTERY_FUNCTION_ALL">
-        <description>Battery supports all systems</description>
+        <description>Battery supports all functions (of vehicle if sent by autopilot, otherwise the current component).</description>
       </entry>
       <entry value="2" name="MAV_BATTERY_FUNCTION_PROPULSION">
         <description>Battery for the propulsion system</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3364,7 +3364,7 @@
         <description>Battery function is unknown</description>
       </entry>
       <entry value="1" name="MAV_BATTERY_FUNCTION_ALL">
-        <description>Battery supports all functions (of vehicle if sent by autopilot, otherwise the current component).</description>
+        <description>Battery supports all functions of the current system/vehicle. Should not be set by components with individual batteries, such as cameras.</description>
       </entry>
       <entry value="2" name="MAV_BATTERY_FUNCTION_PROPULSION">
         <description>Battery for the propulsion system</description>


olliw42:
many thx for putting this up.

I unfortunately do not have a good opinion (except of that the current MAV_BATTERY_FUNCTION enum isn't really well designed LOL).

I think the fundamental question to be answered first is on whom's perspective the info decoded into MAV_BATTERY_FUNCTION should be based on. I see two situations: system-wide or component centric.
System-wide would mean that each battery tells it's role in the overall system, comprising of the autopilot and it's components. This would make it relatively easy to recognize the role of a battery, but may need configuration.
Component-centric would mean from the perspective of the component, which is unique given that any component can be identified by it's ID. This would make it harder to recognize the role of a battery, since one in addition has to take into account the nature of a component as revealed by it's id and MAV_TYPE, but may need less configuration, i.e. be somewhat more generic in that sense.
Both are essentially mutually exclusive, so one can't easily have it both ways, and thus a A-or-B decision be made. Once decided, everything else will fall in place easily, I would believe.

Btw, you may also want to e.g. consider the case that a camera/gimbal/etc is powered from the main battery (which isn't unrealistic), yet may emit a battery_status. So technically it shouldn't emit it, but this again may involve configuration work which needs to be done. In that situation the battery status would be more of a power status report. 

...

difficult topic to get right

:)

hamishwillee:
Yes. The other consideration is that this is implemented by lots of systems already, so care needs to be taken to manage any move from system to component centric.

message_definitions/v1.0/common.xml
```diff
@@ -3353,7 +3353,7 @@
         <description>Battery function is unknown</description>
       </entry>
       <entry value="1" name="MAV_BATTERY_FUNCTION_ALL">
-        <description>Battery supports all flight systems</description>
+        <description>Battery supports all functions (of vehicle if sent by autopilot, otherwise the current component).</description>
```
```suggestion
        <description>Battery supports all functions of the current system/vehicle. Should not be set by components with individual batteries, such as cameras.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -3353,7 +3353,7 @@
         <description>Battery function is unknown</description>
       </entry>
       <entry value="1" name="MAV_BATTERY_FUNCTION_ALL">
-        <description>Battery supports all flight systems</description>
+        <description>Battery supports all functions of the current system/vehicle. Should not be set by components with individual batteries, such as cameras.</description>
```
hamishwillee, line 3367: This is a vehicle-centric enum for a battery and assumes that a vehicle might have one battery powering everything, or separate batteries for controlling the motors vs the flight controller (avionics) and some payload.  

What it doesn't consider is that the battery might belong to another mavlink component such as a camera, remote id or gimbal. In that case, what value should be used for those components, and what should the GCS display for the function of the battery?

My take is that those batteries should always set one of the other values - i.e. payload for camera/gimbal. Avionics presumably for remoteid? The GCS should display a more precise function from the MAV_TYPE

What this change does is make that explicit.

@peterbarker Does that make sense to you? We could instead add more explicit payloads here to make it generic.

hamishwillee:
Closing as the _intent_ is clear enough by inference - i.e. the enum assumes a system either has one battery for everything or separate batteries for avionics etc. Clarifying that this doesn't cover a camera or whatever won't change anything.

This is as useful as it is going to get for the main cases. A recipient probably should use this info if it gets the message from a flight stack, and otherwise use the MAV_TYPE to determine the purpose.

PR #1266, opened 2019-11-09
common: MAV_CMD_DO_ORBIT: add new option for heading behaviour (param index #3)

TSC21:
Extends param index 3 of `MAV_CMD_DO_ORBIT` with a new option for the heading behaviour in Orbit mode that allows the vehicle to keep its heading forward while executing the flight following the orbit circle. Will follow a PR to PX4 with support for it. QGC to come later.

@dayjaby FYI.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1038,6 +1038,24 @@
         <description>Camera does not supply storage status information. Capacity information in STORAGE_INFORMATION fields will be ignored.</description>
       </entry>
     </enum>
+    <enum name="ORBIT_YAW_BEHAVIOUR">
+      <description>Yaw behaviour during orbit flight.</description>
+      <entry value="0" name="ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER">
+        <description>Vehicle front points to the center (default).</description>
+      </entry>
+      <entry value="1" name="ORBIT_YAW_BEHAVIOUR_HOLD_INITIAL_HEADING">
+        <description>Vehicle front holds heading when message received.</description>
+      </entry>
+      <entry value="2" name="ORBIT_YAW_BEHAVIOUR_UNCONTROLLED">
+        <description>Yaw uncontrolled.</description>
+      </entry>
+      <entry value="3" name="ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TANGENT_TO_CIRCLE">
+        <description>Vehicle front follows flight path (tangential to circle).</description>
+      </entry>
+      <entry value="4" name="ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED">
+        <description>Yaw controlled by RC input.</description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM and MISSION_ITEM_INT messages) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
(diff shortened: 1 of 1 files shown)


TSC21:
@MaEtUgR @Stifael FYI.

message_definitions/v1.0/common.xml
```diff
@@ -1190,7 +1190,7 @@
         <description>Start orbiting on the circumference of a circle defined by the parameters. Setting any value NaN results in using defaults.</description>
         <param index="1" label="Radius" units="m">Radius of the circle. positive: Orbit clockwise. negative: Orbit counter-clockwise.</param>
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Vehicle configuration default.</param>
-        <param index="3" label="Yaw Behavior" minValue="0" maxValue="2" increment="1">Yaw behavior of the vehicle. 0: vehicle front points to the center (default). 1: Hold last heading. 2: Leave yaw uncontrolled.</param>
+        <param index="3" label="Yaw Behavior" minValue="0" maxValue="3" increment="1">Yaw behavior of the vehicle. 0: vehicle front points to the center (default). 1: Hold last heading. 2: Leave yaw uncontrolled. 3: vehicle follows the flight direction with forward heading (90 degrees with respect to the circle center).</param>
```
MaEtUgR:
```suggestion
        <param index="3" label="Yaw Behavior" minValue="0" maxValue="3" increment="1">Yaw behavior of the vehicle. 0: vehicle front points to the center (default). 1: Hold last heading. 2: Leave yaw uncontrolled. 3: vehicle front points into tangential flight direction</param>
```

hamishwillee: Thoughts
1. You could make the options an enum. Up to you.
2. Is _hold last heading_ the heading when the message was received or the heading when it entered the orbit path?
3. What does "leave yaw controlled" mean. Essentially whatever? When would you use this? Is it, for example to allow yaw to be separately controlled from the sticks? If not, should there be such an option explicitly?

Without knowing the answers to the above. 
```
<param index="3" label="Yaw Behavior" minValue="0" maxValue="3" increment="1">Yaw behavior of the vehicle. 0: Vehicle front points to the center (default). 1: Vehicle front holds heading when message received. 2: Yaw uncontrolled. 3: Vehicle front follows flight path (tangential to circle).</param>
```

TSC21: @hamishwillee, `<param index="3" label="Yaw Behavior" minValue="0" maxValue="3" increment="1">Yaw behavior of the vehicle. 0: Vehicle front points to the center (default). 1: Vehicle front holds heading when message received. 2: Yaw uncontrolled. 3: Vehicle front follows flight path (tangential to circle).</param>` is right on spot. Will change to that.

hamishwillee: @TSC21 Great. I think I can merge when you've done that. 

Do you think that it is worth adding a case for Yaw controlled by RC at this point?

TSC21: We can but then that needs to be added in https://github.com/PX4/Firmware/pull/13426.
@dayjaby can you do that on your end?

dayjaby: @hamishwillee @TSC21 Sure! This sounds like an awesome feature, especially because the orbit flight mode is already processing RC input for controlling altitude/radius/flight direction.

TSC21: @dayjaby I added an enum for the modes and also the value `4` for yaw controlled by RC. This needs to be reproduced also in PX4/Firmware#13426, including the switch cases, which should use the Mavlink enumerations instead of the direct integers for better readability.

TSC21:
@hamishwillee @MaEtUgR since we now have 5 different options for the yaw heading, it looked more reasonable to add an enumeration instead of having all options on the description. Can you please recheck again? @hamishwillee please feel free to merge if all's good.

TSC21:
@hamishwillee thanks for merging!

PR #1268, opened 2019-11-13
common.xml: update extended fence status message

IamPete1:
This is a change to the new fence mitigation enum to include Unknown mitigation action, this came about after a discussion on the ArduPilot dev-call about my ArduPilot PR to make use of the new extension.

https://github.com/ArduPilot/ardupilot/pull/12536

https://github.com/ArduPilot/mavlink/pull/108

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -962,10 +962,13 @@
     <enum name="FENCE_MITIGATE">
       <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Actions being taken to mitigate/prevent fence breach</description>
-      <entry value="0" name="FENCE_MITIGATE_NONE">
+      <entry value="0" name="FENCE_MITIGATE_UNKNOWN">
+        <description>Unknown</description>
+      </entry>
+      <entry value="1" name="FENCE_MITIGATE_NONE">
         <description>No actions being taken</description>
       </entry>
-      <entry value="1" name="FENCE_MITIGATE_VEL_LIMIT">
+      <entry value="2" name="FENCE_MITIGATE_VEL_LIMIT">
         <description>Velocity limiting active to prevent breach</description>
       </entry>
     </enum>


message_definitions/v1.0/common.xml
```diff
@@ -962,10 +962,13 @@
     <enum name="FENCE_MITIGATE">
       <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Actions being taken to mitigate/prevent fence breach</description>
-      <entry value="0" name="FENCE_MITIGATE_NONE">
+      <entry value="0" name="FENCE_MITIGATE_UNKNOWN">
```
hamishwillee: So does this indicate "you know that you're taking a mitigation operation but you don't know what it is" or does it indicate "you don't know if you are taking a mitigation operation".

My assumption is that you always know you're doing mitigation or not (ie this is the first case above). If so you need to swap these around because the default (0) should reflect the system state of a system that knows nothing about fences.

IamPete1: This would be if we don't know if you are taking a mitigation operation or not, for example if fences are not enabled. If we were to report none then the GCS might assume that we checked the fence and that were fine. I think that was the general argument in any-case.

hamishwillee: This no longer works for the GCS end. Previously getting 0 means no mitigation. This change would force the GCS to assume there might be mitigation even for systems that don't support it. So now it presumably has to display an icon indicating "maybe I am maybe I am not mitigating". A bit silly.

Also the argument "for" is spurious.  A system should always know if it is mitigating or not - e.g. by checking if fences are enabled. 

So I'm inclined to reject.

IamPete1: You first point, surely this is the point of a WIP enum? So we can change it?

Your second point I can see either way, @peterbarker maybe you can argue the case more eloquently than I?

hamishwillee: Yes you can certainly change the enum to something else that makes sense. You can't change it to something daft :-).

Let's see what @peterbarker says. Perhaps there are other considerations.

peterbarker: @hamishwillee Sadly, "we have no idea what the vehicle is doing" is actually exactly what a value of 0 must mean, as we can't tell the difference between a zero coming in because a mavlink client really does mean 0 or because it doesn't understand that extension.  Zero-stripping means we can't tell how many extensions were understood by the sender.

This aspect is actually a real problem for mavlink2 extensions; you have to choose a range type for your value based on the magic-zero problem.

hamishwillee: @peterbarker Thanks. I do "intimately" understand the problem with extensions. You're right that it doesn't know. 

My thought was that this isn't a helpful distinction - because the GCS action would only an icon when it is actively mitigating and you know it is. But I see that with this you can have a UI where the mitigating button is on/off and either hidden altogether or greyed out if you don't know.

So yes. Thanks (and to @IamPete1) for clarification.

message_definitions/v1.0/common.xml
```diff
@@ -962,10 +962,13 @@
...
+        <description>Unkown</description>
```
hamishwillee: Typo > Unknown.

IamPete1: oops, fixed, Thanks

hamishwillee:
Merged. Thanks for your patience @IamPete1 .

PR #1272, opened 2019-11-19
Add illuminator message

nicovanduijn:
This PR introduces a new Mavlink command to set the RGB values of an illuminator.
Potential use cases include illuminating something in the dark or light shows.

Not sure if this is the right place to add it, I'm open to moving it somewhere else.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4010,36 +4010,6 @@
       <entry value="2048" name="AIS_FLAGS_VALID_CALLSIGN"/>
       <entry value="4096" name="AIS_FLAGS_VALID_NAME"/>
     </enum>
-    <enum name="MAV_LED_COLOR">
-      <description>Defines the color of an RGB LED.</description>
-      <entry value="0" name="OFF">
-        <description>Do not use the defined color, but the raw RGB values instead.</description>
-      </entry>
-      <entry value="1" name="RED">
-        <description>Set the LED to red.</description>
-      </entry>
-      <entry value="2" name="GREEN">
-        <description>Set the LED to green.</description>
-      </entry>
-      <entry value="3" name="BLUE">
-        <description>Set the LED to blue.</description>
-      </entry>
-      <entry value="4" name="YELLOW">
-        <description>Set the LED to yellow.</description>
-      </entry>
-      <entry value="5" name="PURPLE">
-        <description>Set the LED to purple.</description>
-      </entry>
-      <entry value="6" name="AMBER">
-        <description>Set the LED to amber.</description>
-      </entry>
-      <entry value="7" name="CYAN">
-        <description>Set the LED to cyan.</description>
-      </entry>
-      <entry value="8" name="WHITE">
-        <description>Set the LED to white.</description>
-      </entry>
-    </enum>
     <enum name="MAV_LED_MODE">
       <description>Defines the mode of operation of an RGB LED.</description>
       <entry value="0" name="OFF">
@@ -6155,10 +6125,9 @@
       <field type="uint8_t[32]" name="G">G values for the selected LEDs</field>
       <field type="uint8_t[32]" name="B">B values for the selected LEDs</field>
       <field type="uint8_t[32]" name="W">W values for the selected LEDs. Ignored if LED system does not support W values.</field>
-      <field type="uint8_t" name="color" enum="MAV_LED_COLOR">Color of the LEDs. Overrides RGB values.</field>
+      <field type="uint8_t" name="color_all">Color all LEDs to R, G, B field 0th index values (ignore other RGB array values). 1: Enable, 0: Disable (use array values).</field>
       <field type="uint8_t" name="mode" enum="MAV_LED_MODE">Mode of LED operation.</field>
       <field type="uint8_t" name="num_blinks">How many times to blink (number of on-off cycles if mode is one of MAV_LED_MODE::BLINK_*). Set to 0 for infinite.</field>
-      <field type="uint8_t" name="priority">Priority. Higher-priority events will override lower-priority events.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6121,13 +6121,13 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint16_t" name="led_mask">Bitmask to select up to 16 LED strips/channels</field>
+      <field type="uint8_t" name="color_all">Color all LEDs to R, G, B, W field 0th index values (ignore other RGBW array values). 1: Enable, 0: Disable (use array values).</field>
+      <field type="uint8_t" name="mode" enum="MAV_LED_MODE">Mode of LED operation.</field>
+      <field type="uint8_t" name="num_blinks">How many times to blink (number of on-off cycles if mode is one of MAV_LED_MODE::BLINK_*). Set to 0 for infinite.</field>
       <field type="uint8_t[32]" name="R">R values for the selected LEDs</field>
       <field type="uint8_t[32]" name="G">G values for the selected LEDs</field>
       <field type="uint8_t[32]" name="B">B values for the selected LEDs</field>
       <field type="uint8_t[32]" name="W">W values for the selected LEDs. Ignored if LED system does not support W values.</field>
-      <field type="uint8_t" name="color_all">Color all LEDs to R, G, B field 0th index values (ignore other RGB array values). 1: Enable, 0: Disable (use array values).</field>
-      <field type="uint8_t" name="mode" enum="MAV_LED_MODE">Mode of LED operation.</field>
-      <field type="uint8_t" name="num_blinks">How many times to blink (number of on-off cycles if mode is one of MAV_LED_MODE::BLINK_*). Set to 0 for infinite.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4010,6 +4010,60 @@
       <entry value="2048" name="AIS_FLAGS_VALID_CALLSIGN"/>
       <entry value="4096" name="AIS_FLAGS_VALID_NAME"/>
     </enum>
+    <enum name="MAV_LED_COLOR">
+      <description>Defines the color of an RGB LED.</description>
+      <entry value="0" name="OFF">
+        <description>Do not use the defined color, but the raw RGB values instead.</description>
+      </entry>
+      <entry value="1" name="RED">
+        <description>Set the LED to red.</description>
+      </entry>
+      <entry value="2" name="GREEN">
+        <description>Set the LED to green.</description>
+      </entry>
+      <entry value="3" name="BLUE">
+        <description>Set the LED to blue.</description>
+      </entry>
+      <entry value="4" name="YELLOW">
+        <description>Set the LED to yellow.</description>
+      </entry>
+      <entry value="5" name="PURPLE">
+        <description>Set the LED to purple.</description>
+      </entry>
+      <entry value="6" name="AMBER">
+        <description>Set the LED to amber.</description>
+      </entry>
+      <entry value="7" name="CYAN">
+        <description>Set the LED to cyan.</description>
+      </entry>
+      <entry value="8" name="WHITE">
+        <description>Set the LED to white.</description>
+      </entry>
+    </enum>
+    <enum name="MAV_LED_MODE">
+      <description>Defines the mode of operation of an RGB LED.</description>
+      <entry value="0" name="OFF">
+        <description>Turns the LED off.</description>
+      </entry>
+      <entry value="1" name="ON">
+        <description>Turns the LED on.</description>
+      </entry>
+      <entry value="2" name="DISABLED">
+        <description>Disables the current priority mode of LED. (Switch to lower priority task)</description>
+      </entry>
+      <entry value="3" name="BLINK_SLOW">
+        <description>Start blinking the LED slowly.</description>
+      </entry>
+      <entry value="4" name="BLINK_FAST">
+        <description>Start blinking the LED quickly.</description>
+      </entry>
+      <entry value="5" name="BREATHE">
+        <description>Set the LED to continuously increase and decrease in brightness.</description>
+      </entry>
+      <entry value="6" name="FLASH">
+        <description>Flash the LED.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="0" name="HEARTBEAT">
@@ -6091,6 +6145,20 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
+    </message>
+    <message id="410" name="RGBW_LED_CONTROL">
+      <description>Set the RGBW values a set of LEDs.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint16_t" name="led_mask">Bitmask to select up to 8 LEDs</field>
+      <field type="uint8_t[32]" name="R">R values for the selected LEDs</field>
+      <field type="uint8_t[32]" name="G">G values for the selected LEDs</field>
+      <field type="uint8_t[32]" name="B">B values for the selected LEDs</field>
+      <field type="uint8_t[32]" name="W">W values for the selected LEDs</field>
+      <field type="uint8_t" name="color" enum="MAV_LED_COLOR">Color of the LEDs. Overrides RGB values.</field>
+      <field type="uint8_t" name="mode" enum="MAV_LED_MODE">Mode of LED operation.</field>
+      <field type="uint8_t" name="num_blinks">How many times to blink (number of on-off cycles if mode is one of MAV_LED_MODE::BLINK_*). Set to 0 for infinite.</field>
+      <field type="uint8_t" name="priority">Priority. Higher-priority events will override lower-priority events.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6150,11 +6150,11 @@
       <description>Set the RGBW values a set of LEDs.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint16_t" name="led_mask">Bitmask to select up to 8 LEDs</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Hi Nico, 
This is a good place for the message. I know nothing about illuminators, but my first thoughts were:
1. Will there ever be more than one on a vehicle? ie is index required?
2. Are they always on? ie do we need on/off?
3. Do they have any other characteristics - e.g. brightness?
4. Do we ever need to query the values?
5. Do we need acknowledgement? Ie why a command rather than a message?

hamishwillee:
PS We would often add as <wip /> and then remove after first implementation.

nicovanduijn:
Regarding your questions:
1) Fair point, I can add an index
2) I made the command take RGB values, e.g. turning it off would be equivalent to sending 0/0/0
3) Same for brightness, it is implicitly part of the RGB values. E.g. For a purely white LED, the brightness could just be set by sending it anything between 0/0/0 (off) to 255/255/255 (full power).
However, there are fairly common RGBW LEDs on the market, so I'm open to including a fourth value if you think I should
4) I don't see a use case for querying these values
5) An ack isn't strictly required, although if we wish to add something like a "light switch" to QGC, it may be nice to know if the message went through

hamishwillee:
> 1) Fair point, I can add an index

Great.

> 2) I made the command take RGB values, e.g. turning it off would be equivalent to sending 0/0/0
> 3) Same for brightness, it is implicitly part of the RGB values. E.g. For a purely white LED, the brightness could just be set by sending it anything between 0/0/0 (off) to 255/255/255 (full power).

Assume people won't realise the above and add note in description:

Set the RGB values of the illuminator. Values range from 0,0,0 (off) to  255/255/255 (white, full power).

> However, there are fairly common RGBW LEDs on the market, so I'm open to including a fourth value if you think I should

I don't know. If the there is ever a "need" to independently control the W as well then yes.  But in this case you would include note "on RGB LEDs the W LED should always be set to 0"

> 4) I don't see a use case for querying these values

In the general case you would do this if you don't have the ack and want to know if turn on succeeded, OR something else can change the LEDs outside of your control. So for example a companion computer is controlling the LEDs or there are multiple GCS and you still want your light switch on the UI. 

> 5) An ack isn't strictly required, although if we wish to add something like a "light switch" to QGC, it may be nice to know if the message went through

True. This works for that case, but perhaps not so much for the multi-system case.

Generally for multi system cases the paradigm is a setter message that causes recipient to emit a broadcast with current state. Another sender can also query that state by requesting the message be emitted again.

I don't know how likely that is, so "just sayin'" :-)

hamishwillee:
@julianoes You got any advice on this too?

julianoes:
1. I would probably not add an index but rather a mask which allows you to address a single LED or multiple together.

2. / 3. At first I immediately thought that brightness is missing but I guess you're right that it's implicit in RGB.

We should check what else is needed from what we're already using PX4 internally:
https://github.com/PX4/Firmware/blob/master/msg/led_control.msg
For instance I wonder if we should add the blinking/breathing preset/mode.

nicovanduijn:
Thanks for the feedback! I actually wasn't aware of https://github.com/PX4/Firmware/blob/master/msg/led_control.msg.

I did my best to make the message "mappable" to that uORB in PX4 and yet keep all the functionality there.

message_definitions/v1.0/common.xml
```diff
@@ -1749,6 +1749,17 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="225" name="MAV_CMD_DO_SET_ILLUMINATOR" hasLocation="false" isDestination="false">
+        <description>Set the RGB values of the illuminator.</description>
+        <param index="1" label="LED Mask" minValue="0" maxValue="255" increment="1">Bitmask to select up to 8 LEDs.</param>
+        <param index="2" label="R" minValue="0" maxValue="255" increment="1">Red color intensity of the RGB illuminator.</param>
+        <param index="3" label="G" minValue="0" maxValue="255" increment="1">Green color intensity of the RGB illuminator.</param>
+        <param index="4" label="B" minValue="0" maxValue="255" increment="1">Blue color intensity of the RGB illuminator.</param>
+        <param index="5" label="color" minValue="0" maxValue="8" increment="1">Choose a pre-set color. 0:OFF (RGB values are used), 1:RED, 2:GREEN, 3:BLUE, 4:YELLOW, 5:PURPLE, 6:AMBER, 7:CYAN, 8:WHITE.</param>
```
hamishwillee: Should state that this supersedes param 2, 3, 4 values. Worth doing an enum for these?

message_definitions/v1.0/common.xml
```diff
@@ -1749,6 +1749,17 @@
...
+        <param index="6" label="mode" minValue="0" maxValue="8" increment="1">LED mode. 0:OFF, 1:ON, 2:DISABLED, 3:BLINK_SLOW, 4:BLINK_NORMAL, 5:BLINK_FAST, 6:BREATHE, 7:FLASH.</param>
```
hamishwillee: Should state that this supersedes param 2, 3, 4 values. Worth doing an enum for these?

message_definitions/v1.0/common.xml
```diff
@@ -1749,6 +1749,17 @@
...
+        <param index="7" label="num_blinks" minValue="0" maxValue="255" increment="1">How many times to blink (number of on-off cycles if mode is one of MODE_BLINK_*). Set to 0 for infinite.</param>
```
hamishwillee: Any particular reason to limit maxvalue?

message_definitions/v1.0/common.xml
```diff
@@ -1749,6 +1749,17 @@
...
+        <param index="8" label="priority" minValue="0" maxValue="2" increment="1">Priority. Higher-priority events will override lower-priority events.</param>
```
hamishwillee: Is this so different systems can send commands to the illuminators? I imagine that if one system sent a message you'd expect the new command to be obeyed irrespective of the priorities?

message_definitions/v1.0/common.xml
```diff
@@ -1749,6 +1749,17 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="225" name="MAV_CMD_DO_SET_ILLUMINATOR" hasLocation="false" isDestination="false">
+        <description>Set the RGB values of the illuminator.</description>
+        <param index="1" label="LED Mask" minValue="0" maxValue="255" increment="1">Bitmask to select up to 8 LEDs.</param>
```
hamishwillee: Any particular reason to limit to 8 LEDs? Might be easier to do as an enum.

hamishwillee:
@julianoes Do you think you might ever want to be able to check the illuminator status? ie read what it is doing?

olliw42:
I guess I would add a brightness as it, as it was mentioned, would allow to independently set it.
The rgb might come out of some whatever fancy code part, which might not allow adjusting the overall brightness, but users may want to adjust it depending on their preference or power of their actual illuminator installed.

I didn't knew a CMD can have more than 7 entries

:)

message_definitions/v1.0/common.xml
olliw42: it is float size so could easily support more than 8

nicovanduijn: A float for a bitmask? Can someone explain to me how that should work?

32-bit floating point numbers typically follow the IEEE 754 floating point representation, with the first bit for the sign, 8 bits for the exponent (biased by 127) and 23 bits for the normalized mantissa. Please tell me which floating point number you would need to send to set LEDs 1,4,5,8,21,22,31 to be on and the others off. (The answer is 6.712116E7)

I can only assume you meant by 'float size' that it is 32 bits. I still think it's a bit extreme to put 32 individually controllable RGB LEDs on a MAV, at some point it's just called a display. The reason I stopped at 8 is because PX4's uORB message does so. Of course there's no real reason to extend this limitation to mavlink, but it would make parsing easier

message_definitions/v1.0/common.xml
nicovanduijn: Only reason was that the PX4's uORB message has this limitation, and it would make parsing easier if we don't have to handle the overflow case on that side. Technically no, there's no real reason to limit the number of blinks to 255

message_definitions/v1.0/common.xml
nicovanduijn: Beats me. I just followed the PX4 message

nicovanduijn:
I can make it a message then

message_definitions/v1.0/common.xml
```diff
@@ -5894,6 +5948,15 @@
...
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
+    <message id="410" name="RGBW_LED_CONTROL">
+      <description>Set the RGBW values a set of LEDs.</description>
+      <field type="uint8_t" name="led_mask">Bitmask to select up to 8 LEDs</field>
+      <field type="uint8_t[4]" name="rgbw">RGBW values for the selected LEDs</field>
+      <field type="uint8_t" name="color" enum="MAV_LED_COLOR">Color of the LEDs. Overrides RGB values.</field>
+      <field type="uint8_t" name="mode" enum="MAV_LED_MODE">Mode of LED operation.</field>
+      <field type="uint8_t" name="num_blinks">How many times to blink (number of on-off cycles if mode is one of MAV_LED_MODE::BLINK_*). Set to 0 for infinite.</field>
+      <field type="uint8_t" name="priority">Priority. Higher-priority events will override lower-priority events.</field>
```
julianoes: If it's a message it needs a `target_system` and `target_component`.

nicovanduijn: done

message_definitions/v1.0/common.xml
```diff
@@ -4010,6 +4010,60 @@
       <entry value="2048" name="AIS_FLAGS_VALID_CALLSIGN"/>
       <entry value="4096" name="AIS_FLAGS_VALID_NAME"/>
     </enum>
+    <enum name="MAV_LED_COLOR">
+      <description>Defines the color of an RGB LED.</description>
```
hamishwillee: Optionally (i.e. I quite like to see where things are used, but it is easy enough to search)

> Defines the color of an RGB LED (used in RGBW_LED_CONTROL).

message_definitions/v1.0/common.xml
```diff
@@ -6092,6 +6146,17 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
+    <message id="410" name="RGBW_LED_CONTROL">
+      <description>Set the RGBW values a set of LEDs.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t" name="led_mask">Bitmask to select up to 8 LEDs</field>
```
hamishwillee: @nicovanduijn Are you sure you're only going to need max 8 LEDs? I'm pretty sure someone somewhere will need more. I'd rather waste a bit of space now than have to do it later. Of course now we'll argue how many is enough. I'm going to suggest 32.

message_definitions/v1.0/common.xml
```diff
@@ -6092,6 +6146,17 @@
...
+      <field type="uint8_t[4]" name="rgbw">RGBW values for the selected LEDs</field>
+      <field type="uint8_t" name="color" enum="MAV_LED_COLOR">Color of the LEDs. Overrides RGB values.</field>
+      <field type="uint8_t" name="mode" enum="MAV_LED_MODE">Mode of LED operation.</field>
+      <field type="uint8_t" name="num_blinks">How many times to blink (number of on-off cycles if mode is one of MAV_LED_MODE::BLINK_*). Set to 0 for infinite.</field>
+      <field type="uint8_t" name="priority">Priority. Higher-priority events will override lower-priority events.</field>
```
hamishwillee: Can you explain a scenario when you expect these would be used, and how? 

Minimally we should also clarify what equal priority events do, and that lower priority events are ignored.

hamishwillee:
@nicovanduijn looks good to me module comments on priority and whether limiting number of illuminators to 8 is reasonable. 

We can add a message to get illuminator state later if needed.

@auturgy Any comments?

auturgy:
ArduPilot supports up to 512 individually addressable LED’s (16 channels, 32 per channel). The API is currently exposed via the Lua scripting engine, not mavlink, though.
See https://youtu.be/4py2rp74lJQ 


Regards,

James

hamishwillee:
Thanks @auturgy 

@nicovanduijn We could map this similarly:
```
<field type="uint16_t" name="index">Select LED channel/index. 0-indexed.</field>
<field type="uint32_t" name="led_mask">Bitmask to select up to 32 LEDs  (for specified index.</field>
```
Of course we could use bigger for both or adjust the sizes relatively. Up to you, but we should support at least the sizes we know are supported by MAVLink flight stacks. Personally I'd suggest we make 32 channels too.

olliw42:
hm ... I think the point of 16 channels and 32 per channel are the LED strips, i.e., you can have 16 LED strips with 32 LEDs each, and address each strip and each LED on each strip. The up to 32 values for one strip need to kind of streamed always at once, and a bit map for the 32 LEDs doesn't then make much sense (it could be a bit map for the 16 channels).

This let's me think that - if one wants to mirror that - an array of up to 32 rgb values would make more sense. The channel would be then the illuminator, and the 32 rgb values the light pattern this illuminator displays. 

It however also let's me think that a support of more sophisticated patterns than just on/off/blink, like pulsing, knight rider, etc., will be soon desired ... which suggests some more sophisticated fields, like a pattern enum and some variable.

I'm not sure if one wants to make such a mavlink message however, which let's me think KIS.

Which here could mean to have 16 channels, and that the rgb value of that message would map to setting all 32 LEDs of a strip to the same color.

:)

auturgy:
That makes some sense.
Or an index/instance for the channel (bitmask?), and 32 bit array for individual LED’s.
Basically, 8 isn’t enough. 16 would be better, 32 plus a channel mask better still (imho).

Regards,

James

hamishwillee:
Hmm, so something like?:
```
<field type="uint16_t" name="mask">Select LED channel - up to 16.</field>
<field type="uint8_t[32]" name="R">R values for the selected LEDs</field>
<field type="uint8_t[32]" name="G">G values for the selected LEDs</field>
<field type="uint8_t[32]" name="B">B values for the selected LEDs</field>
<field type="uint8_t[32]" name="W">W values for the selected LEDs</field>
```
And presumably a mask rather than an index so you can apply the setting to all the illuminators at once if you want?

Much bigger message now, and not much trimming.

nicovanduijn:
Ok, this is getting way out of hand and far beyond the use case I'm looking for myself.

I want to turn on an illuminator.

I thought, to make it general to other use cases, let's add RGB.

Now we're talking about sending 128 bytes of data plus fluff, where I really only need one single bit. I'm working on a system where mavlink bandwidth is already a big issue, as it is being sent over the same link as a video stream. There is no way I can use the message you guys suggest for the very purpose I originally designed it.

I'm dropping this PR for anyone else to pick it up.

nicovanduijn:
Increased the message to address 32 RGBW LEDs at once now

hamishwillee:
@auturgy With Nico's changes, would this be adequate for representing the ArduPilot LED system? 

My main concern with this is the priority field. I can guess how it might work, but not how it could ever work well.

auturgy:
I'm unsure about the priority field, and also the MAV_LED_COLOR enum.  I'm not convinced this shouldn't the the color hex code or 8 bit R, G and B values (which are basically the same thing), rather than an enum.
The 8 bit values directly align with the ArduPilot API, but I don't want to hijack this any further.  Up to you guys.

hamishwillee:
@auturgy Thanks. Nico has stepped back from this as his use case actually orthogonal to this message - he can use [MAV_CMD_ILLUMINATOR_ON_OFF](https://mavlink.io/en/messages/common.html#MAV_CMD_ILLUMINATOR_ON_OFF).

Upshot, we should do the right thing here. IMO that is:
1. Remove the priority field. It makes no sense.

I slightly lean towards the idea of hex codes for the enum because it would allow colours to be defined and extended unambigously - ie we could support any colour we liked. Is it worth the cost of having to represent this as a UINT32? 

Not sure how you would do this with separate RGB values ...

auturgy:
RGB would be 3 uint8’s, one for each.
The hex code is actually just a different representation of exactly the same thing (six hex digits, first two R, then G, and B, 0-255 for each)

Regards,

James

hamishwillee:
So, just to be very precise, you're suggesting no enum? Just something like
```
<field type="uint8_t[3]" name="color">Color of all the LEDs. Overrides other RGB values.</field>
```
Alternatively, in line with this we could just do
```
<field type="uint8_t" name="color all">Color of all the LEDs to same value - the 0 index values of the R, G, B fields (ignoring other array elements) . 0 disabled, 1 enabled .</field>
```

I have no problem in theory, other than I figure the enum was designed to make it easier just to pick a colour without thinking about it too hard.

So for example:
```
      <entry value="32678" name="GREEN">
        <description>Set the LED to green.</description>
      </entry>
```
and
```
      <field type="uint32_t" name="color" enum="MAV_LED_COLOR">Color of the LEDs. Overrides RGB values.</field>
```

auturgy:
Yes.
I understand the intent behind the enum, and it’s attraction, but I think the abstraction/simplification should be handled by the UI (or even SDK), to retain full value in the message.

Regards,

James

hamishwillee:
Thanks @auturgy . OK, so removed priority.

Re colour enums, sure - makes sense. What I have done is deleted the enum, and added a field that, if set to 1, sets the colour of all LEDs to be the 0th array element of the RGBW fields. I think this is still useful, because it potentially allows a slightly smaller message to be sent.

If you disagree then let me know, and I'll remove that field.

Any other concerns?

hamishwillee:
@auturgy This what you meant? ^^^^

nicovanduijn:
This hasn't moved in a while - despite my main problem being taken care of I would still like to see this go in. What's holding back the merge?

hamishwillee:
@nicovanduijn Intertia. If a PR loses a stakeholder then normally we close it until we get someone committing to the work. That way we don't end up with common messages that are "theoretical".

@auturgy Does this deliver what would be needed by ArduPilot (as per your comment https://github.com/mavlink/mavlink/pull/1272#issuecomment-559634879). If so, do you think anyone will implement it on that flight stack? I doubt it will be a priority for anyone to implement in PX4 right now.

IF not, I propose to close this until we get a new stakeholder.

nicovanduijn:
I would be willing to pick it up again on the PX4 side, if that helps

hamishwillee:
@nicovanduijn It would. YOu have hardware for this? Does the message do everything you think it should? If so I can merge as WIP and we can progress from there (OK?)

nicovanduijn:
I'm not quite there yet, I'm currently developing for it.

it's for a customer-specific implementation so only parts of it can go upstream. The hooks to listen for the message will go upstream, but the actual driver cannot. But it enables anyone to write a simple LED driver. Is that good enough?

hamishwillee:
> Is that good enough?

Yes. The point is that we know the message meets the main use cases for real devices. Ideally of course there would be a driver for this.

hamishwillee:
@nicovanduijn Thanks for this. The dev call decide that we should close it unless there is proved to be pull from the community - ie someone who needs this and can champion it. If that is the case they can reopen or recreate. Otherwise this sort of thing sits around forever.

Thanks for taking the time to discuss and generalise this when all you wanted originally was the torch: https://mavlink.io/en/messages/common.html#MAV_CMD_ILLUMINATOR_ON_OFF

PR #1274, opened 2019-11-29
Add illuminator on/off command

nicovanduijn:
This PR replaces the RGB version proposed in #1272  that got way out of hand for the use-case I'm targeting.

This is supposed to be a light-weight command to turn a light on board of an MAV on or off. No more, no less


hamishwillee:
This is OK by me. There is the other PR for a fully functional LED controller, but I see a lightweight turn lights on and off message as fine in parallel. @julianoes any objection?

hamishwillee:
Merging. This makes sense to me, but as WIP can still be modified if needed.

PR #1276, opened 2019-12-02
MAV_CMD_REQUEST_MESSAGE: Define that parameters can be used

hamishwillee:
This allows us to request messages using MAV_CMD_REQUEST_MESSAGE where the value of the returned message can be conditional on values passed to the target system.

In most cases message don't depend on parameters and the new params should be set to zero. In cases where parameters are expected, the values of the params must be specified in the message that is requested.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1864,6 +1864,10 @@
         <description>Request the target system(s) emit a single instance of a specified message (i.e. a "one-shot" version of MAV_CMD_SET_MESSAGE_INTERVAL).</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID of the requested message.</param>
         <param index="2" label="Index ID" minValue="0" increment="1">Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.</param>
+        <param index="3">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="4">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="5">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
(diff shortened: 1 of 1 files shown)


julianoes:
I don't understand what this is used for.

hamishwillee:
An example is requesting the [service version](https://docs.google.com/document/d/1zK_JcJbhtbRA3W0MgnXSNWMUxOcBu6vXocrjZVMXki4/edit?ts=5de097ca#heading=h.q82gs8qw8s1). The information you get back depends on what you put in.

I don't see this as particularly confusing, but I do see that it isn't quite so encapsulated a design as having (in this case) a separate message.

julianoes:
@hamishwillee so where do we document param3 and param4 which only matter for `MAVLINK_SERVICE_VERSION`?

hamishwillee:
As it says in the MAV_CMD_REQUEST_MESSAGE definition here "The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0)."

So, in `MAVLINK_SERVICE_VERSION` you would have a description that says something like

> This message is emitted in response to MAV_CMD_REQUEST_MESSAGE , where param 3 is X and param 4 is Y (other params should be set to zero)

julianoes:
Ok fine. I don't have a better suggestion.

hamishwillee:
@julianoes I like the idea of this message; it is convenient to be able to request whatever we like without having to create a new message. But perhaps this is a "step to far."

The alternative is to simply decide that this can only be used to request messages that have no "configuration". The implication is that you couldn't use this for cases like the microservices versioning. 

I actually don't have a particular problem with that; a microservice having a dedicated messages feels more "encapsulated".

But if you say nothing in the next couple of days I propose to merge this. Doesn't mean we shouldn't still have dedicated messages for this microservice case though.

julianoes:
I think we should still have a dedicated command for the microservice handshake.

hamishwillee:
I agree. I'll merge this anyway though.

PR #1277, opened 2019-12-03
Update ADS-B flags

WickedShell:
This adds a couple of additional ADS-B flags. These have been shipping with uAvionix hardware for awhile, so this meaning is already present and deployed in the field, and the validity flags have a lot of value to be receiving either way.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3064,6 +3064,9 @@
       <entry value="16" name="ADSB_FLAGS_VALID_CALLSIGN"/>
       <entry value="32" name="ADSB_FLAGS_VALID_SQUAWK"/>
       <entry value="64" name="ADSB_FLAGS_SIMULATED"/>
+      <entry value="128" name="ADSB_FLAGS_VERTICAL_VELOCITY_VALID"/>
+      <entry value="256" name="ADSB_FLAGS_BARO_VALID"/>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -3064,6 +3064,9 @@
       <entry value="16" name="ADSB_FLAGS_VALID_CALLSIGN"/>
       <entry value="32" name="ADSB_FLAGS_VALID_SQUAWK"/>
       <entry value="64" name="ADSB_FLAGS_SIMULATED"/>
+      <entry value="128" name="ADSB_FLAGS_VERTICAL_VELOCITY_VALID"/>
+      <entry value="256" name="ADSB_FLAGS_BARO_VALID"/>
+      <entry value="32768" name="ADSB_FLAGS_SOURCE_UAT"/>
```
hamishwillee: FMI How come this was added out of sequence?

WickedShell: I have no clue, just that it was, and it shipped :)

hamishwillee: Sorry, I mean "why did you use leave a bunch of empty flags in the middle - 256 jumps to 32768

WickedShell: Oh I understand the question, I just don't have the answer. I'm just trying to update the docs to match what has actually shipped as the firmware for a bunch of hardware.

hamishwillee:
Thanks @WickedShell - much appreciated. Merged.

PR #1284, opened 2019-12-16
common.xml: Added lat, lon alt labels

amilcarlucas:
add missing yaw units

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1258,19 +1258,19 @@
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
         <param index="3" label="Pass Radius" units="m">0 to pass through the WP, if &gt; 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
         <param index="4" label="Yaw" units="deg">Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="17" name="MAV_CMD_NAV_LOITER_UNLIM" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint an unlimited amount of time</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4" label="Yaw">Desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1260,19 +1260,19 @@
         <param index="2" label="Transition Heading" enum="VTOL_TRANSITION_HEADING">Front transition heading.</param>
         <param index="3">Empty</param>
         <param index="4" label="Yaw Angle" units="deg">Yaw angle. NaN for unchanged.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="5" label="Latitude">Latitude</param>
```
hamishwillee: Do these have any meaning fora VTOL transition?

message_definitions/v1.0/common.xml
```diff
@@ -1260,19 +1260,19 @@
...
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="85" name="MAV_CMD_NAV_VTOL_LAND" hasLocation="true" isDestination="true">
         <description>Land using VTOL mode</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
-        <param index="3" label="Altitude">Approach altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
+        <param index="3" label="Aproach Altitude" units="m">Approach altitude (with the same reference as the Altitude field). NaN if unspecified.</param>
```
hamishwillee: Approach

message_definitions/v1.0/common.xml
```diff
@@ -1484,8 +1484,8 @@
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
+        <param index="5" label="Latitude">Latitude</param>
```
hamishwillee: Might be worth pulling in the comment from above for Lat and Lone - ie "Optional - set to 0 if not needed."

message_definitions/v1.0/common.xml
```diff
@@ -1077,39 +1077,39 @@
...
       <entry value="17" name="MAV_CMD_NAV_LOITER_UNLIM" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint an unlimited amount of time</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4" label="Yaw">Desired yaw angle. NaN for unchanged.</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="4" label="Yaw" units="deg" minValue="0" maxValue="360">Desired yaw angle. NaN for unchanged.</param>
```
hamishwillee: Are negative yaws allowed?  I know they are the same thing, but ground station might accept negative, and it might use the value to determine which direction it turns to reach the yaw (or it might just go whichever direction is fastest).

hamishwillee:
I guess it is nice to be precise, especially as some of the fields in some cases are optional.

amilcarlucas:
I'll try to address these things soon. This should go in before #1318

message_definitions/v1.0/common.xml
BazookaJoe1900: So what to do here? How to call param 2, and how to call param 7?
I prefer that param 7 will keep being `Altitude` and that param 2 will be renamed to `gate altitude`
Agree?

message_definitions/v1.0/common.xml
amilcarlucas: I do not know. That clarification can be done in another PR

message_definitions/v1.0/common.xml
amilcarlucas: done, thanks

message_definitions/v1.0/common.xml
amilcarlucas: OK, removed

amilcarlucas:
Comments addressed now. Will push as soon as tests pass.

PR #1290, opened 2020-01-13
common: added MAV_SYS_STATUS_PREARM_CHECK

dagar:
Cherry-picked ardupilot/mavlink - https://github.com/ArduPilot/mavlink/pull/119

Is there a process for keeping these in sync? I was adding something similar before I stumbled across this.

FYI @MaEtUgR

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -833,6 +833,9 @@
       <entry value="134217728" name="MAV_SYS_STATUS_SENSOR_SATCOM">
         <description>0x8000000 Satellite Communication </description>
       </entry>
+      <entry value="268435456" name="MAV_SYS_STATUS_PREARM_CHECK">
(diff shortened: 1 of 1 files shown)


auturgy:
@amilcarlucas syncs the xml’s in both directions semi-regularly.
More broadly, the fortnightly mavlink devcall aims to achieve alignment, but frankly I’m not sure it’s worth my time.

hamishwillee:
@dagar There is no process for checking whether contributions in ArduPilot or any other dialect should be in common. It would be great if people had time for this, or even better if dialects wanted to put commonly useful features into the common.xml. What the mavlink meeting tends to do is ensure that if something does go into common it makes sense for the main stakeholders.

What @amilcarlucas does is keep the mavlink/mavlink and ArduPilot/mavlink XML variants in sync with each other.

PR #1291, opened 2020-01-21
Conan package

galikrosentalik:
this enhancement gives the possibility to create conan package from the mavlink repository.
https://conan.io/

known issues:
* get_version method in conanfile.py requires deeper understanding in conan to work flawlessly.
* Submodules are currently not supported in conan architecture, we are using git clone in order to bypass the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ include/
 *.pyc
 include/mavlink/v1.0
 .idea
+cmake-build-debug
--- a/conanfile.py
+++ b/conanfile.py
@@ -30,10 +30,10 @@ class MavlinkConan(ConanFile):
     description = "micro air vehicle communication protocol"
     topics = ("mavlink", "communication")
     exports_sources = "*"
-    options = {"msg_definitions": ["ardupilotmega", "ASLUAV", "autoquad", "common", "icarous", "matrixpilot",
+    options = {"msg_dialect": ["ardupilotmega", "ASLUAV", "autoquad", "common", "icarous", "matrixpilot",
                                     "minimal", "paparazzi", "python_array_test", "slugs", "standard", "test",
                                    "ualberta", "uAvionix"]}
-    default_options = ("msg_definitions = ardupilotmega")
+    default_options = ("msg_dialect = ardupilotmega")
     settings = "os", "compiler", "build_type", "arch"
 
     def source(self):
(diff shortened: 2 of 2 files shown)


hamishwillee:
Hi @galikrosentalik 
Thanks very much. My concern with this contribution is that we don't know how to maintain it and to someone coming to the repo it isn't clear what it is for. 

So, for example, it isn't clear if we have to do anything each time we make a new release and how we keep this up to date. 

So cautiously interested in taking this, but with more documentation in the file on usage. Might it also make sense to have this in a "package_manager" subdirectory with a README describing the content?

@auturgy Have you any thoughts on this?

auturgy:
We need to have a think about package management more generally: the only “managed” package is pymavlink, but there are maven and nuget examples also floating around that should be gripped up.
My suggestion would be to automate the package generation based on mavlink releases, but we’d need to actually do mavlink releases....

Regards,

James

hamishwillee:
@galikrosentalik Also, this has `license = "Tevel"`. What does that mean?

@auturgy Thanks & yes. Unfortunately that's a bigger project that we both know won't progress quickly. What I'm trying for is a holding pattern solution that leaves us no worse off and means we're better prepared to organise similar contributions in future.

hamishwillee:
@galikrosentalik Thanks for license update. Can you give us your thoughts on docs for this script?

galikrosentalik:
hey hamishwillee & auturgy, 
I appreciate very much your comments.
It does seem that the issues regarding package management should be considered cautiously.
I think that in order to promote the issue, we need to start somewhere.. 
Conan seems like a good position as an industry standard for managing c++ packages , supported by jfrog.
I sure agree that a well documentation is needed(this will be included in my next commit).

galikrosentalik:
hey hamishwillee & auturgy,
we had some discussion in my office regarding this pull request, after considering your comments.
the controversy is regarding the right place for the conan management, is it in the mavlink or in the mavlink generator. 
if we displace it to the pymavlink repository, so additionally in creating the cpp header files you could ask it to generate the conanfile as well.
I would appreciate your opinion.

hamishwillee:
@galikrosentalik Re https://github.com/mavlink/mavlink/pull/1291#issuecomment-582280906 While I tend to think "here" is the right place I don't have a strong opinion.

How about you raise the question in as an issue in ArduPilot/Pymavlink (cross link here). The generator maintainers mostly live there. They may have a view.

Worse case, once this is documented with clear instructions on how it might be maintained then we can take it.

hamishwillee:
Closing. Thanks for this, but not clear to me that this is needed or can be maintained.

PR #1292, opened 2020-01-24
add avoidance sys_status

mrivi:
This change would allow to report the healthiness and use of different avoidance components from the fight controller to the ground station with a single message. For example in PX4 we have the Collision Prevention working in manual modes and Obstacle Avoidance working in auto modes. This message would report to the ground station if any of the avoidance systems is functional in the flight mode the vehicle currently is. The ground station can simply check the SYS_STATUS message to display the information to the user.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -836,8 +836,8 @@
       <entry value="268435456" name="MAV_SYS_STATUS_PREARM_CHECK">
         <description>0x10000000 pre-arm check status. Always healthy when armed</description>
       </entry>
-      <entry value="536870912" name="MAV_SYS_STATUS_VISION_SAFETY_SYSTEMS">
-        <description>0x20000000 Vision based safety systems status </description>
+      <entry value="536870912" name="MAV_SYS_STATUS_OBSTACLE_AVOIDANCE">
+        <description>0x20000000 Avoidance/collision prevention</description>
       </entry>
     </enum>
     <enum name="MAV_FRAME">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -836,6 +836,9 @@
       <entry value="268435456" name="MAV_SYS_STATUS_PREARM_CHECK">
         <description>0x10000000 pre-arm check status. Always healthy when armed</description>
       </entry>
+      <entry value="536870912" name="MAV_SYS_STATUS_VISION_SAFETY_SYSTEMS">
+        <description>0x20000000 Vision based safety systems status </description>
+      </entry>
     </enum>
     <enum name="MAV_FRAME">
       <entry value="0" name="MAV_FRAME_GLOBAL">


LorenzMeier:
To me this is a generic enough category of sensor to make sense.

hamishwillee:
Would this be reported for ADS-B or FLARM? If not, then IMO the name (or description) is not sufficiently general.

hamishwillee:
@mrivi Thanks. To my mind this is better.

@auturgy @julianoes This is status for the vision based safety systems like collision prevention/obstacle avoidance. I think it is specific enough to differentiate with other safety systems.

Any comment? If nothing then I hope to merge in day or so.

auturgy:
This is a misuse of “safety system” as a term.

hamishwillee:
> This is a misuse of “safety system” as a term.

@auturgy That was my suggestion :-(. We're trying to capture that obstacle avoidance, collision prevention, safe landing and other computer-vision based avoidance and protection software is healthy (@mrivi I am assuming healthy, not necessarily "actively protecting the vehicle").

@auturgy @mrivi How about we drop "safety" - that can be implied. So below we're just taking about the health of our vision system
```
      <entry value="536870912" name="MAV_SYS_STATUS_VISION_SYSTEMS">
        <description>0x20000000 Vision-based systems (e.g. for object avoidance, collision prevention, etc.)</description>
```
If not - got a better idea?

message_definitions/v1.0/common.xml
```diff
@@ -836,6 +836,9 @@
       <entry value="268435456" name="MAV_SYS_STATUS_PREARM_CHECK">
         <description>0x10000000 pre-arm check status. Always healthy when armed</description>
       </entry>
+      <entry value="536870912" name="MAV_SYS_STATUS_VISION_SAFETY_SYSTEMS">
```
julianoes: We only have 4 items left. Could we make this one more generic? E.g. VISION_SYSTEM, SAFETY_SYSTEM, OBSTACLE_AVOIDANCE system?

auturgy:
Unless there’s some specific reason why this needs to refer to a vision based system, I’d go with Julian’s OBSTACLE_AVOIDANCE suggestion  as it generalises better (could capture lidar, ultrasonic, radar based systems too).

jkflying:
I like OBSTACLE_AVOIDANCE as well.

mrivi:
My initial submission was MAV_SYS_STATUS_AVOIDANCE. I changed it because of Hamish comment. If @hamishwillee is okay with  MAV_SYS_STATUS_OBSTACLE_AVOIDANCE, I'll make the change

hamishwillee:
@julianoes @jkflying @auturgy The problem with obstacle avoidance as a name is what should a user assume it indicates/does not indicate? Does it include collision prevention? Does it include Flarm or other transponder based avoidance systems? What about Geofence?
 
I have made this avoidance again (much like @mrivi original suggestion). If there are suggestions to improve this/capture the above points then please add by COP. Otherwise I will assume you're all happy with the current definition and merge.

hamishwillee:
Merged!

julianoes:
>  The problem with obstacle avoidance as a name is what should a user assume it indicates/does not indicate? Does it include collision prevention? Does it include Flarm or other transponder based avoidance systems? 

I would say it does include collision prevention but not Flarm/transponder based avoidance systems. However, by saying vision based your excluding solutions based on sonar or lidar.

> What about Geofence?

That's something else, I would say.

PR #1293, opened 2020-01-24
common: Add missing bitmask in PLAY_TUNE_V2

patrickelectric:
bitmask exist in SUPPORTED_TUNES but not in PLAY_TUNE_V2

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6159,7 +6159,7 @@
       <description>Play vehicle tone/tune (buzzer). Supersedes message PLAY_TUNE.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint32_t" name="format" enum="TUNE_FORMAT">Tune format</field>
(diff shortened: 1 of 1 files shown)


LorenzMeier:
Thanks!

hamishwillee:
@LorenzMeier @patrickelectric I think this change is incorrect. The bitmask attribute allows a GCS to understand that the values can be treated as bitmasks - ie you can have multiple bits set and OR the result. That makes no sense for PLAY_TUNE_V2 because you can only ask a system to play a tune in one format at a time.

It is OK (but not particularly useful) for SUPPORTED_TUNES because a system can in theory support multiple tune formats and tell you all of them.

patrickelectric:
@hamishwillee I thought that was the original idea behind the message and that's why I asked for you review.
My problem was related to https://github.com/mavlink/rust-mavlink where such message was breaking our code generator, it was the only one using a bitmask variable type without the bitmask xml tag. Be free to revert if necessary.

hamishwillee:
Hi @patrickelectric ,
I don't particularly feel the need to revert, but can you please update your code generator to cope with the case that the variable might be missing?

PR #1296, opened 2020-01-29
common: Missions obey system yaw behaviour on Nan

hamishwillee:
Follows on from https://github.com/PX4/px4_user_guide/pull/642

Waypoints yaw to specified param4 value. Previously docs stated that NaN means unchanged, but in practice flight stacks actually have a typical behaviour and NaN reverts to it. This clarifies that the NaN behaviour is system dependent.


hamishwillee:
THanks. Me to. Merging.

PR #1297, opened 2020-01-29
common: Set mission yaw mode command

hamishwillee:
During missions, if yaw is not specified for a waypoint (or other mission item) then NaN is sent, indicating "use the system yaw mode". This might be, for example, to always face the waypoint, or to always face home [note, this applies to multirotor vehicles, and anything that can have independent heading and yww].

Flight stacks typically set the yaw behaviour using a parameter. This PR:
- Creates a MAV_CMD to allow the default behaviour to be set by a GCS. The initial mode set is based on PX4 [MPC_YAW_MODE](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#MPC_YAW_MODE)
- The same MAV_CMD can be used as a mission item to over-ride the system default.
- Creates a message that can be queried by the GCS to get the system yaw mode default for all missions.

This follows on from discussion [here](https://github.com/PX4/px4_user_guide/pull/642#issuecomment-579104971).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7349,6 +7349,16 @@
       <field type="uint8_t" name="count">Number of wheels reported.</field>
       <field type="double[16]" name="distance" units="m">Distance reported by individual wheel encoders. Forward rotations increase values, reverse rotations decrease them. Not all wheels will necessarily have wheel encoders; the mapping of encoders to wheel positions must be agreed/understood by the endpoints.</field>
     </message>
+    <message id="405" name="MISSION_DEFAULT_YAW">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>
+        The system default mission yaw.
+        This can be requested using MAV_CMD_REQUEST_MESSAGE.
+        Note that the yaw is the default value for all missions if yaw is not explicitly specified, and may be set using AV_CMD_SET_DEFAULT_YAW_MODE.
+      </description>
+      <field type="uint8_t" name="connection_type" enum="MISSION_YAW_MODE">Default mission yaw mode (e.g. towards next waypoint, towards home, etc).</field>
+    </message>
     <message id="9005" name="WINCH_STATUS">
       <description>Winch status.</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (synced to UNIX time or since system boot).</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -772,6 +772,25 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
     </enum>
+    <enum name="MISSION_YAW_MODE">
+      <description>
+        Default vehicle yaw.
+        This is used in MAV_CMD_SET_DEFAULT_YAW_MODE to set the default yaw used in missions (if the yaw is not explicitly set).
+        The yaw mode applies to vehicles that allow independent control over yaw and direction of travel (e.g. multirotors).
+      </description>
+      <entry value="0" name="MISSION_YAW_MODE_WAYPOINT">
+        <description>Vehicle front points towards next waypoint.</description>
+      </entry>
+      <entry value="1" name="MISSION_YAW_MODE_HOME">
+        <description>Vehicle front points towards home.</description>
+      </entry>
+      <entry value="2" name="MISSION_YAW_MODE_AWAY_HOME">
+        <description>Vehicle front points away from home (rear faces home location).</description>
+      </entry>
+      <entry value="3" name="MISSION_YAW_MODE_TRAJECTORY">
+        <description>Vehicle front points along vehicle trajectory.</description>
+      </entry>
+    </enum>
     <enum name="WIFI_CONFIG_AP_RESPONSE">
       <description>Possible responses from a WIFI_CONFIG_AP message.</description>
       <entry value="0" name="WIFI_CONFIG_AP_RESPONSE_UNDEFINED">
@@ -1032,7 +1051,7 @@
         <param index="1" label="Hold" units="s" minValue="0">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
         <param index="3" label="Pass Radius" units="m">0 to pass through the WP, if &gt; 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
-        <param index="4" label="Yaw" units="deg">Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
+        <param index="4" label="Yaw" units="deg">Desired yaw angle at waypoint (multicopter). NaN to use the current system or mission-specific yaw heading mode as set using MAV_CMD_SET_DEFAULT_YAW_MODE (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
         <param index="7" label="Altitude" units="m">Altitude</param>
@@ -1891,6 +1910,15 @@
         <param index="1" label="Failure unit" enum="FAILURE_UNIT">The unit which is affected by the failure.</param>
         <param index="2" label="Failure type" enum="FAILURE_TYPE">The type how the failure manifests itself.</param>
         <param index="3" label="Instance">Instance affected by failure (0 to signal all).</param>
+      </entry>
+      <entry value="421" name="MAV_CMD_SET_DEFAULT_YAW_MODE" hasLocation="false" isDestination="false">
+        <description>
+          Sets the default mission yaw mode (e.g. yaw: towards next waypoint, towards home, etc.).
+          Sets the system-default mission yaw if sent via the command protocol.
+          When used it a mission it over-rides the system-default mission yaw for the current mission (from the point(s) at which it is inserted).
+          Note that yaw modes only apply on vehicles that support independent control of the direction of movement and the direction that the vehicle is facing (e.g. multirotor).
+        </description>
+        <param index="1" label="Yaw Mode" enum="MISSION_YAW_MODE">Default yaw behaviour in mission mode.</param>
       </entry>
       <entry value="500" name="MAV_CMD_START_RX_PAIR" hasLocation="false" isDestination="false">
         <description>Starts receiver pairing.</description>


hamishwillee:
@julianoes Do you happen to know how [MPC_YAW_MODE](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#MPC_YAW_MODE)  option `4: towards waypoint (yaw first)` differs from option 0?

hamishwillee:
@auturgy The ones I'd need to include from ArduPilot are these, right? : https://ardupilot.org/copter/docs/parameters-Copter-stable-V3.6.9.html?#wp-yaw-behavior-yaw-behaviour-during-missions

julianoes:
> @julianoes Do you happen to know how MPC_YAW_MODE option 4: towards waypoint (yaw first) differs from option 0?

No idea, I would ask @bresch.

bresch:
In both modes 0 and 4 the vehicle is oriented towards the next waypoint but in mode
- 0: the vehicle moves horizontally and yaws at the same time (usually there is a simultaneous motion in the first few meters). This creates a smooth and faster mission.
- 4: the vehicle yaws first and when roughly aligned, starts to move horizontally. This is important for vehicles that need to fly forward only (e.g.: VTOL plane or forward-facing avoidance camera).

hamishwillee:
@bresch @julianoes That sounds like a distinction that a MAVLink user should never need to know about - ie the vehicle should know if it is VTOL or has avoidance camera. 

Is there value in adding this, or can I assume that the vehicle will "do the right thing"?

hamishwillee:
@auturgy Further to https://github.com/mavlink/mavlink/pull/1297#issuecomment-579559161 ...

I propose the following mapping for ardupilot yaw modes:
* 0 - Never change yaw - add mode
  ```
      <entry value="4" name="MISSION_YAW_MODE_FIXED">
        <description>Vehicle yaw does not change.</description>
      </entry>
  ```
* 1 -  Face next waypoint =  already proposed mode
  ```
      <entry value="0" name="MISSION_YAW_MODE_WAYPOINT">
        <description>Vehicle front points towards next waypoint.</description>
      </entry>
   ```
* 2 - Face next waypoint except RTL -
  - Where does it point during RTL? If it should point towards the RTL item destination then I'd expect this to be identical to MISSION_YAW_MODE_WAYPOINT.
* 3 -  Face along GPS course = already proposed mode
  ```
      <entry value="3" name="MISSION_YAW_MODE_TRAJECTORY">
        <description>Vehicle front points along vehicle trajectory.</description>
      </entry>
  ```

hamishwillee:
@auturgy @julianoes can you please sanity check this?

James, please see my comment https://github.com/mavlink/mavlink/pull/1297#issuecomment-580508523. I believe that this captures the ardupilot modes, except possibly 2.

Should we make this WIP for now?

amilcarlucas:
I need some time to review this. Ardupilot already has something similar in place, would be good to make it partialy compatible with it.

hamishwillee:
@amilcarlucas I attempted to be compatible with ArduPilot - see https://github.com/mavlink/mavlink/pull/1297#issuecomment-580508523

There was an open question on the mode "2" in that post (ie the ArduPilot definition is ambiguous). I think it is covered.

message_definitions/v1.0/common.xml
```diff
@@ -1861,6 +1881,11 @@
         <param index="6">Reserved</param>
         <param index="7">Reserved</param>
       </entry>
+      <entry value="420" name="MAV_CMD_SET_YAW_MODE" hasLocation="false" isDestination="false">
+        <description>Set the default mission yaw mode (e.g. yaw: towards next waypoint, towards home, etc.). This is the yaw mode that will be used if yaw is not explicitly specified in a mission item (e.g. MAV_CMD_NAV_WAYPOINT.param4=Nan). Yaw modes only apply on vehicles that support independent control over direction of movement and the direction the vehicle is facing (e.g. multirotor).</description>
+        <param index="1" label="Yaw Mode" enum="MISSION_YAW_MODE">System default yaw behaviour in mission mode.</param>
+        <param index="2" label="Persistence" enum="MISSION_YAW_MODE_PERSISTENCE">Persistence of the commanded mode setting. By default the setting is permanent until explicitly changed.</param>
```
julianoes: I really disagree with making this persistent by default because it is unexpected (for me).

To me this can be part of the mission setting in the same way that "finish with RTL after mission" is. This way it can be uploaded with the mission and hence does not need to stick.
If it really needs to stick for an airframe in general, then a param can be used.

message_definitions/v1.0/common.xml
```diff
@@ -1071,6 +1071,26 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
     </enum>
+    <enum name="MISSION_YAW_MODE">
+      <description>Default vehicle yaw in missions (used when mission item yaw is not explicitly set/is set to NaN). Applies to multirotors (and other vehicles that allow independent control over yaw and direction of travel).</description>
```
julianoes: I would add this command to the description of [MAV_CMD_NAV_WAYPOINT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_WAYPOINT) so that it's clear in both directions.

hamishwillee: Good point. Have updated all affected commands.

message_definitions/v1.0/common.xml
hamishwillee: @julianoes My original thinking was just a command protocol "command" to directly replace the parameter setting - so that a GCS can set and display the behaviour without having to know the parameters. 

Then this changed in discussion so that we could allow it in a mission. Do we really need that? I.e. will you really want to be able to change the default yaw mode just within a mission? I'm not sure its a needed use case. At which point we could perhaps argue that we just need to make it clear this should not be used in missions. Thoughts?

If we do allow this in missions, I get your point. I expect a command in a mission to apply until cancelled/replaced OR to last until next waypoint. But I don't expect it to last across missions. So having it as a default is a problem.

What is the solution? What happens if the default is something that isn't supported? Do we also need a message for the system to display what yaw modes it actually supports?

julianoes (approved):

message_definitions/v1.0/common.xml
hamishwillee: @julianoes Thanks for approving, but would really love your opinion on ^^^^. "Is there a solution".

The easiest option is to explicitly state that this is a command and should not be used in missions. Then this is just an alias for the parameter as originally intended, and we can use normal command protocol mechanisms to indicate a particular option is not supported.

We can later decide it can be used in a mission. FMI what should a system do if it can't execute a mission item - ie if it supports the mav_cmd but can't execute the particular variant? AFAIK this is undefined but happens quite regularly.

julianoes: To me the obvious place where to use this command would be as part of a mission. The reason is that there are already a few "settings" that can passed to each mission to set the behavior like [DO_CHANGE_SPEED](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CHANGE_SPEED).

I don't find it intuitive to have commands for settings. If the point is to standardize certain settings then we should standardize the params and not create a (duplicate) command in my opinion.

@LorenzMeier 's point was that we already have other commands which change settings but I don't remember his examples.

I liked through the list and what I found is:
- Camera/gimbal configuration commands (e.g. [MAV_CMD_DO_MOUNT_CONFIGURE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_MOUNT_CONFIGURE)): to be used directly or as part of a mission.
- Trigger configuration (e.g. [MAV_CMD_DO_SET_CAM_TRIGG_DIST](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_CAM_TRIGG_DIST)): to be used for a flight (not permanently).
- Fence enable (is that even used? [MAV_CMD_DO_FENCE_ENABLE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_FENCE_ENABLE))
- Calibration: that's clearly an action.

The advantage of having this param as part of a mission is that it will be saved with the .plan (or .mission) file and therefore stick to the mission. It means that it is a mission setting and not a vehicle setting, and it could even be changed during a mission if required.

hamishwillee: @julianoes Yes,  `MAV_CMD_DO_FENCE_ENABLE` is an example of a command that changes a setting and is not expected to be used in a mission. Looking at the definition there is no way to tell that, which makes it very hard to use. 

Standardising parameters is another can of worms because the whole philosophy of params is that you can do what you like.

But I get your point on the value in a mission.

The easiest way to spec this might be to use documentation to state that:
- if it is used as a command it affects the system defaults. 
- If it is used a mission it affects only the current waypoint. It will override any yaw setting in the current mission item (either specific or NaN). We could instead pair it with a disable explicit yaw setting command, but I quite like the idea that if you want to explicitly define this you do it on every single item [the QGC UI could abstract this pretty easily to the To>From caase]

The reason I like that is that we can have it as a command for now, and change it to a mission item later. 

Thoughts?

julianoes: I still don't like it but I guess I'll have to live with it :smile: .

message_definitions/v1.0/common.xml
```diff
@@ -2047,6 +2061,10 @@
         <param index="2" name="Failure type" enum="FAILURE_TYPE">The type how the failure manifests itself.</param>
         <param index="3" name="Instance">Instance affected by failure (0 to signal all).</param>
       </entry>
+      <entry value="421" name="MAV_CMD_SET_DEFAULT_YAW_MODE" hasLocation="false" isDestination="false">
```
hamishwillee, line 2064: @julianoes OK, thought about it some more, and I *hope* this approach seems more intuitive. 

The message is now <code>MAV_CMD_SET_<strong><i>DEFAULT</i></strong>_YAW_MODE</code>.  
- If used in a **command** it sets the system default mode. This makes sense - what else could it set :-)
- If used in a mission it sets the mission default yaw mode. This over-rides the system default mode for the duration of the mission, or until it is set again in the same mission. 

Mission nav points (e.g. waypoints) behave just as currently with respect to yaw
- if the yaw is explicitly set then they will use that. 
- if yaw is set to NaN is set then they will use the default yaw mode. BUT they will get the current default from a mission if one has been set.

The other/original design was to have a message that explicitly set the mission yaw for a waypoint (over-riding any yaw setting there). That has some benefits, but then we'd have to argue the persistence of the message, and I think it might be harder for a GCS to work out the rules.

Does this work better for you?

If so we can see what Don thinks. GCS probably needs a message to get the current  mission and system yaw mode.

DonLakeFlyer: > GCS probably needs a message to get the current mission and system yaw mode.

I'm not sure one-offing this specific case is a good idea. The concept of the gcs wanted to know this value is just the tip of the iceberg in the vehicle controlled stuff which would be useful. Another simple example are the default mission flight speeds for multi-rotor and fixed wing modes. A gcs needs those to calc mission flight times. Battery information to calculate how many batteries a mission will take. Does this vehicle even support a specific mission command. On and on... I'd rather see some consistent way to handle all of this which in the end may also include a single message to request a value. But I think it needs to be thought through at a higher level.

Which then ties back to MAV_CMD_SET_DEFAULT_YAW_MODE which is on one those vehicle things which are traditionally deal with in an opaque parameter. I would contend that would also be better thought through at a higher level. It may turn out that when thought through to take into account other cases as well that this still is the way to go. But rather get that understanding correct from the start before adding new messages which will just be deprecated further down the road when more things are taken into account.

hamishwillee: @DonLakeFlyer The question of whether any command is supported is being dealt with by another mechanism under proposal: https://docs.google.com/document/d/1wETCft5N7IN4iR9NhfG9cI_fMvlG1PvCXXOPwxSEwGE/edit

I accept it is worth doing a broad sweep on all the information needed by a GCS. 
- You're pretty much the only person who can compile that information. Can you help with this. Otherwise this will just stall and we will do nothing.
- What did you have in mind as a possible "alternative"? Our toolkit is pretty much - opaque parameters (not good for standardization), messages/commands, definition files in XML or whatever.

Your list so far
- default mission flight speeds for multi-rotor and fixed wing modes. - Yes.
- Battery information to calculate how many batteries a mission will take - don't you get that from BATTERY_STATUS?

amilcarlucas:
Just for documentation sake: ArduCopter used the [WP_YAW_BEHAVIOR|https://ardupilot.org/copter/docs/parameters.html#wp-yaw-behavior-yaw-behaviour-during-missions] parameter to control this

hamishwillee:
> Just for documentation sake: ArduCopter used the WP_YAW_BEHAVIOR parameter to control this

Thanks @amilcarlucas - that's captured in the description! The proposal includes the ArduPilot options that differ.

FYI all, the current status of this is that it is "on hold". It is likely we will need it more or less as is, but we're looking at what can be done using component information metadata so that we use the common mechanism for this sort of thing, where possible.

hamishwillee:
keeping open

hamishwillee:
This PR on setting default yaw in missions got stalled by @DonLakeFlyer comment in https://github.com/mavlink/mavlink/pull/1297#discussion_r402548425 

He suggests that there are many bits of vehicle specific information currently encoded in parameters that it would be useful for a GCS to be able to get without knowing the vehicle type, and that we might want to consider consider if this can be done in a standard way rather than just create new getters messages and setter commands every time.

An example he gave was mission default flight speed. I'm guessing many of the [PX4 mission params](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#mission) would also be things to consider - "maximum distance between waypoings".

The alternatives I see are:
- **Standarised parameters** with fixed names and meanings that a GCS can get/set for these kinds of things.
  - GCS can know that if these exist they are supported with given meaning, and read them using normal mechanisms
  - GCS can use normal mechanisms to detect change - e.g. emission of PARAM_VALUE.
  - This is contrary to meaning of parameters as fully "flight stack controlled".  But:
    - There is some "prior art" for this - we do it for camera API params that we might want to pull out in the UI
    - This is how DroneCan may do standardisation of some node behaviours (anecdotally)
  - The set of supported "public defaults" parameters might be delivered as component metadata too.
- **Metadata parameter mapping**: We define component metadata that maps keys for the defaults to parameters
  - This approach uses metadata to identify the parameters standardized parameters. So here you can keep your old parameter names - all we're doing is saying "this is the parameter that is used for this function on this flight stack". We might also specify the units and param type, or we might expect these to be worked out from param metadata.
  - After parsing the metadata things as before - you get/set via the param. 
- **Metadata with hidden mapping:** Metadata just includes the values of each default. The values that can change must be defined as dynamic metadata in component metadata, and be re-emitted if they change. 
- **Messages for each data type** what is proposed here. Not particularly scalable.

Any of these could work to get/set the system default for the yaw mode use case. For the in-mission yaw override we would still need a MAV_CMD

hamishwillee:
Closing. There is no stakeholder.

Anyone seeing this, there is an intersting discussion above in https://github.com/mavlink/mavlink/pull/1297#issuecomment-1118138344 about how we might better expose metadata about system defaults in a generic way.

PR #1299, opened 2020-01-31
Edit documentation to handle case where cell voltages are not measured

ItsTimmy:
Currently, there are two ways to report battery voltage over MAVLink: `SYS_STATUS` and `BATTERY_STATUS`. `SYS_STATUS` only has one field for each battery value, so if there are multiple batteries in a system, there is ambiguity. There is discussion about this in https://github.com/mavlink/mavlink/pull/1182.

To solve this ambiguity, a vehicle can send multiple instances of `BATTERY_STATUS`, with different values for `id`. This is how PX4 currently does it.

In `BATTERY_STATUS`, there is the `voltages` field for sending individual voltages of each cell of the battery. This is useful for 'smart batteries' and other power monitors that can monitor individual cells. However, many power monitors cannot do this; they can only measure the overall voltage of the battery. 

This PR updates the documentation of `BATTERY_STATUS` to handle this case. It works as follows:

 - Set `voltages[0]` to the total battery voltage, in mV. Essentially, treat the battery as if it is one big cell.
 - If the total battery voltage is greater than or equal to `UINT16_MAX`, then set `voltages[0]` to `UINT16_MAX - 1`, and set `voltages[1]` to the remaining voltage.
 - This can be extended to multiple cells if the voltage is greater than `2 * (UINT16_MAX - 1)`.

In this way, the total voltage of the battery can still always be calculated by simply summing every value in `voltages` which is not `UINT16_MAX`.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5388,7 +5388,7 @@
       <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
       <field type="uint8_t" name="type" enum="MAV_BATTERY_TYPE">Type (chemistry) of the battery</field>
       <field type="int16_t" name="temperature" units="cdegC">Temperature of the battery. INT16_MAX for unknown temperature.</field>
-      <field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells. Cells above the valid cell count for this battery should have the UINT16_MAX value.</field>
+      <field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells. Cells above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).</field>
       <field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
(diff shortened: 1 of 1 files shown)


WickedShell:
The approach used in ArduPilot is in the case of a non smart battery to provide the pack voltage as the first cell (and if it exceed 65.534 volts we set the first sell to this, and store the rest in the next cell). It's not amazing, but in the general case it matches the best information we have on the battery (IE that it's a single cell pack as far as we can observe it), and doesn't require any extra bandwidth, and works on a MAVLink 1 link as well.

WickedShell:
@hamishwillee you just add up all the cell voltages that aren't UINT16_MAX the same as you normally would. You don't have to set the next cell to 0, you just set all the cells that don't contain a valid value of less then UINT16_MAX to the no data value of UINT16_MAX. 0 is actually a very valid value for a cell, so you wouldn't set that.

The PR in ArduPilot that did this was: https://github.com/ArduPilot/ardupilot/pull/8999

hamishwillee:
Thanks @WickedShell. Got it (misread 65.534 as UINT16_MAX rather than UINT16_MAX -1).

ItsTimmy:
I think the ardupilot solution sounds good. I have updated this PR accordingly.

hamishwillee:
@julianoes I think this is good. Any comments? (just squash-merge if you are OK with this)

hamishwillee:
Hi @WickedShell 
Further to this, assume we have a 70V battery - this will be a 65.534 in first slot and 4.whatever in the second slot. How does the GCS determine that this is a single cell battery of 70 v or a two-cell battery with values above? 
Does it assume if you get 65.534 in the first cell this is a one cell battery?

WickedShell:
@hamishwillee The GCS notionally sums the voltages to show the overall pack voltage, then would drop a dialog that showed 65.534 and the remainder in the second cell. You could however make the GCS smart enough to know that any pack summing that involves 65.534 from a low cell is not actual cell voltages and stick to displaying a single voltage and hide the cells interface.

PR #1300, opened 2020-02-01
PROJECT_SOURCE_DIR for MAVLINK_SOURCE_DIR

jacobschloss:
Hi,

Can we set MAVLINK_SOURCE_DIR to PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR so that building as a subdirectory works? Otherwise MAVLINK_SOURCE_DIR incorrectly points to the top level project directory instead of the mavlink directory and cmake fails.

- Jacob

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 project (mavlink)
 
 if (NOT DEFINED MAVLINK_SOURCE_DIR)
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @jacobschloss . It seems we can :-). Thank you.

PR #1301, opened 2020-02-05
Remove STATUSTEXT_LONG (unused)

LorenzMeier:
A better solution has been proposed and merged in https://github.com/mavlink/mavlink/pull/1295.


hamishwillee:
@DonLakeFlyer FYI, I assume QGC didn't do anything with this, but just in case, note that the STATUSTEXT_LONG message was removed (ie not just deprecated).

DonLakeFlyer:
Ah, ok. Interesting.

PR #1303, opened 2020-02-14
common: NAN and INT32_MAX for default param values

### LINKED ISSUE #1286: Proposal: Support for optional values in MISSION_ITEM_INT.[x|y]
As the spec currently stands the x and y values of MISSION_ITEM_INT have no mechanism to indicate an optional value. With MISSION_ITEM this is commonly done by setting the value to a NaN. Without this it limits the parameter usage for mission items which are not location based and use x/y for other things than a lat/lon. Even given that I would contend that the ability to support am optional value in lat/lon can be useful as well. For example a TAKEOFF item that has optional location with it. 

We can accomplish this by spec'ing a max int value in x/y indicates the lack of an optional value.

The reason to spec that at the MISSION_ITEM_INT level as opposed to each individual message is to give a gcs enough information to handle this internally. At least for QGC it stores mission items with doubles internally and converts them as they come on. In order to do that correctly it needs to know that for any MISSION_ITEM_INT a max int value in x/y means optional. And it should convert it to a NaN on the way in.

hamishwillee:
Fixes #1286 

This attempts to address #1286 by clarifying that `NaN` and `INT32_MAX` may be used in MAV_CMD params to indicate optional/default values - e.g. a taking yaw from vehicle position rather than the specified value. 

I cover `MISSION_ITEM`, `MISSION_ITEM_INT` and `MAV_CMD`.

I don't cover the implication that if you explicitly specify INT32_MAX for param 5 (say) you are mandating the command is only use in an integer-based message. 

Do we need to update COMMAND and COMMAND_INT?

@auturgy FYI

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1073,7 +1073,7 @@
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
     <!-- ALL the entries in the MAV_CMD enum have a maximum of 7 parameters -->
     <enum name="MAV_CMD">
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes This shouldn't change anything "out of the box", because all it does is recognise that at an INT32_MAX "can be" used for an optional/default value. For this to actually be a default value that would have to be recorded in the specific message. Ditto for NaN (though there are already cases of this).

@DonLakeFlyer Does this capture what you wanted?

DonLakeFlyer:
> Does this capture what you wanted?

Yup.

hamishwillee:
All merged then!

PR #1309, opened 2020-02-18
COMMAND_INT.autocontinue - clarify not used

hamishwillee:
I'd like to also mark these as not used, or at least confirm that someone uses them and how:
- `COMMAND_INT.current` - false:0, true:1
- `COMMAND_LONG.confirmation` - 0: First transmission of this command. 1-255: Confirmation transmissions (e.g. for kill command)

According to @julianoes the `confirmation` is not used https://github.com/mavlink/mavlink-devguide/pull/217#issuecomment-547283478 by PX4, and I suspect not by ArduPilot. 

@auturgy Do you know anything about these two?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4816,7 +4816,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the COMMAND.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the mission item.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">autocontinue to next wp</field>
+      <field type="uint8_t" name="autocontinue">Not used (set 0).</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>


julianoes:
- `autocontinue`: I don't know if it is used but it is generally expected to be set to 1 during a mission so that we go to the next waypoint.
- `COMMAND_INT.current`: This is required in order to upload a mission where you don't want to start with number 0 but some other index. Also, during a mission you can select which one is active, hence that one "becomes" current.
- `COMMAND_LONG.confirmation`: Yes, that one is a mystery to me. Sometimes it's used to indicate whether an ack is required or not, however, `COMMAND_INT` doesn't have it anyway, so you can't rely on it.

dayjaby:
I use autocontinue in a lot of my code. When reaching certain waypoints, you might want to stay there for a while until the user continues so you'd set it false.

If a command is sent via PX4, it usually sends a burst of 4 commands with "confirmation" values of 0, 1, 2, 3 unless the command is acknowledged. That way you can detect package losses, if you only receive commands with confirmation values higher than 0.

hamishwillee:
Thanks very much @dayjaby !

> I use autocontinue in a lot of my code. When reaching certain waypoints, you might want to stay there for a while until the user continues so you'd set it false.

I understand the use for a mission, but this is for command protocol - where there is no "next waypoint". Can you please give me a more detailed explanation?

> If a command is sent via PX4, it usually sends a burst of 4 commands with "confirmation" values of 0, 1, 2, 3 unless the command is acknowledged. That way you can detect package losses, if you only receive commands with confirmation values higher than 0.

Ah, actually this is what is [documented](https://mavlink.io/en/services/command.html#sequences) - good to have confirmation. 

Do you know about how `COMMAND_INT.current` works/should be used. My guess is that it is similar to `confirmation` - you set to 1 if resending a command after missing an ACK. A recipient of this knows not to treat as a new command. In other words it is an improvement of `confirmation` which recogises that your handling of 1,2,3 is always the same.

hamishwillee:
@julianoes Can you approve this change? Logically autocontinue makes no sense in a command (though it does in a mission) and it is not used in PX4 or Ardupilot (checked https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L383 and ardupilot.

hamishwillee:
Self merging. This makes no sense.

Issue #1302, opened 2020-02-06
Feature:  Offboard Altitude + Attitude + Yaw-rate


Stifael:
Note: This issue is a feature discussion

### Goal
Find a solution that is future proof and therefore no hack solution is required at this stage.

### Feature Definition
Being able to control the drone via offboard (companion computer, desktop computer...) through `attitude`-setpoint, `altitude`-setpoints and `yaw-rate` setpoints.

The `attitude`-setpoint can either be `quaternion`, `roll/pitch/yaw` or just `roll/pitch` or all of them. In principle only roll/pitch is required because heading will be controlled through yaw-rate-setpoints.

Altitude control should support altitude-commands relative to takeoff position and velocity-setpoints in z-direction. 

### Possible options
I already discussed possible options, but we did not really find a consent. 

#### Option A: Use two mavlink messages
One idea was to use two existing messages such as [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) and [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED). 

*Advantage*: No need to implement a new message

*Disadvantage*: These messages currently have no dependencies. This means that the autopilot, who has to process the messages for validity, cannot decide if certain fields are missing or not. This means that quite some changes has to be done on the autopilot side against the current implementation where each offboard-mavlink-message contains the full information. 

#### Option B: Create a new mavlink message
The new message would contain all the information required to successfully run attitude-altitude-yaw-rate control. This option aligns with the current offboard-mavlink-implementation, where each mavlink-message contains all the information required to successfully execute the setpoints. 

*Advantage*: Easy to implement
*Disadvantage*: requires a new message and might not cover other combination of setpoints (are there any which actually would make sense?)

#### Option C: virtual sticks
`Virtual sticks` is similar to what DJI has (please don't be hang up on the name `Virtual sticks`). The idea is that the [MANUAL_CONTROL](https://mavlink.io/en/messages/common.html#MANUAL_CONTROL) message can be mimicked via an offboard-message. This means that depending on the desired-flight mode, x,y,z and r have different meanings, exactly as it is done with the existing `MANUAL_CONTROL` message. 

[...]

PR #1310, opened 2020-02-19
Extend SET_POSITION_TARGET_LOCAL_NED 

Stifael:
with roll/pitch/hold_altitude for Offboard Altitdue-mode support

This PR follows the discussion https://github.com/mavlink/mavlink/issues/1302

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3513,6 +3513,12 @@
       <entry value="2048" name="POSITION_TARGET_TYPEMASK_YAW_RATE_IGNORE">
         <description>Ignore yaw rate</description>
       </entry>
+      <entry value="4096" name="POSITION_TARGET_TYPEMASK_ROLL_IGNORE">
+        <description>Ignore roll</description>
+      </entry>
+      <entry value="8192" name="POSITION_TARGET_TYPEMASK_PITCH_IGNORE">
+        <description>Ignore pitch</description>
+      </entry>
     </enum>
     <enum name="UTM_FLIGHT_STATE">
       <description>Airborne status of UAS.</description>
@@ -4758,6 +4764,10 @@
       <field type="float" name="afz" units="m/s/s">Z acceleration or force (if bit 10 of type_mask is set) in NED frame in meter / s^2 or N</field>
       <field type="float" name="yaw" units="rad">yaw setpoint</field>
       <field type="float" name="yaw_rate" units="rad/s">yaw rate setpoint</field>
+      <extensions/>
+      <field type="float" name="roll" units="rad">roll setpoint</field>
+      <field type="float" name="pitch" units="rad">pitch setpoint</field>
+      <field type="uint8_t" name="hold_altitude">If hold_altitude set to true and vz set to 0, the altitude will be locked to current altitude. Otherwise normal velocity control will be applied in the D-direction.</field>
     </message>
     <message id="85" name="POSITION_TARGET_LOCAL_NED">
       <description>Reports the current commanded vehicle position, velocity, and acceleration as specified by the autopilot. This should match the commands sent in SET_POSITION_TARGET_LOCAL_NED if the vehicle is being controlled this way.</description>


Stifael:
@hamishwillee @julianoes please have a look for further discussion

message_definitions/v1.0/common.xml
```diff
@@ -4758,6 +4764,10 @@
       <field type="float" name="afz" units="m/s/s">Z acceleration or force (if bit 10 of type_mask is set) in NED frame in meter / s^2 or N</field>
       <field type="float" name="yaw" units="rad">yaw setpoint</field>
       <field type="float" name="yaw_rate" units="rad/s">yaw rate setpoint</field>
+      <extensions/>
+      <field type="float" name="roll" units="rad">roll setpoint</field>
+      <field type="float" name="pitch" units="rad">pitch setpoint</field>
+      <field type="uint8_t" name="hold_altitude">If hold_altitude set to true and vz set to 0, the altitude will be locked to current altitude. Otherwise normal velocity control will be applied in the D-direction.</field>
```
Jaeyoung-Lim: Why do we need a separate `hold_alittude` field? wouldn't it be better to set it in `POSITION_TARGET_TYPEMASK`?

Stifael:
> This PR breaks this representation by adding roll, pitch to the message.
> For me it is a bit wierd that we have roll, pitch setpoint being sent for SET_TARGET_LOCAL_NED and then we also have a SET_ATTITUDE_TARGET message. Shouldn't roll, pitch fit more in the SET_ATTITUDE_TARGET message

The setpoints of this PR is a mix between position-control (altitude-control) and attitude-control. I don't have a strong opinion about where the additional setpoints should be added to (`SET_ATTITUDE_TARGET` vs `SET_POSITION_TARGET_LOCAL_NED`). First I also was planing to add it to `SET_ATTITUDE_TARGET`, but eventually I decided against it. However, I really don't mind to add it to `SET_ATTITUDE_TARGET`, given the fact that extending an existing mavlink-msg is not my preferred solution (I would go for option C in #1302). The main reason why I added it to `SET_POSITION_TARGET_LOCAL_NED` is because the new offboard Altitude-feature will be handled in the same PX4 code scope as the existing offboard Position-control, which means one uORB message is enough to cover offboard Altitude and Position control (both offboard types will enter FlightTask). In contrast,  `SET_ATTITUDE_TARGET` is handled differently in PX4 where this message gets mapped to uORB message that is not handled by FlightTask. Hence, from a control-pipeline perspective, Altitude-control is more alike Position-control than  Attitude-control. But again, I am fine with the other implementation as well. 

---

Let me summarize what the objective is and then let's go from there:
The **objective** is to support Manual Altitude Control via offboard. The setpoint structure should contain following setpoints:
* `roll`-setpoint
* `pitch`-setpoint
* `yaw_rate`-setpoint
* `velocity_z`-setpoint

In addition, it should be possible to tell the autopilot that 0-velocity in z-direction should be interpreted with an altitude-lock or not. For that, another flag is required:
* `hold_alt`

---

Extending the `SET_ATTITUDE_TARGET` mavlink-msg would require following changes:
* extend struct with `roll` and `pitch`
* either extend struct with velocity or use bit 4 (preserved) of `type_mask` to define if field `thrust` should be interpreted as `thrust` or as `velocity_z`
* use bit 5 (preserved) of `type_mask` to indicate if `roll/pitch` should be ignored
* use bit 6 (preserved) to indicate if altitude_lock

---

Extending `SET_POSITION_TARGET_LOCAL_NED`: see [commit](https://github.com/mavlink/mavlink/pull/1310/commits/b416ca54d734a0e0b277497bbbc62b7b84dd8a03)

---

About your other questions:
>  what would be the use case for specifying roll, pitch compared to quaternions?

As reference for attitude, there are basically two options if **no** external state-estimation source has been added :
1. attitude-setpoint (quaternion or euler-angles) without yaw_rate
2. roll-/pitch-setpoint + yaw_rate-setpoint

Quaternion (euler-angles) + yaw_rate makes not much sense, unless the yaw-setpoint is equal to the yaw-estimate of the vehicle. However, that means that yaw-estimate of the vehicle needs to be known. Of course this is known through telemetry data but only with a delay.

By the way, for manual attitude- and altitude-control, you apply the same logic where a attitude (quaternion) is generated out of roll/pitch- and yaw_rate setpoints. In contrast, in offboard both options are possible.

> Why do we need a separate hold_alittude field? I am not sure if it is really worth it to add a hold_altitude in it

Here I am quite sure that it is needed, because I had this conversation already in the past. Before the FlightTask refactor, the offboard-position-control was implemented in such a way that the position-lock in xy- and z-direction was applied when the velocity-setpoints were around 0, and this even without a flag, it was basically a hard-coded logic in PX4. I removed it myself during the FlightTask-refactor because the mavlink-msg [SET_POSITION_TARGET_LOCAL_NED](https://mavlink.io/en/messages/common.html#SET_POSITION_TARGET_LOCAL_NED) does not say anything about it. However, being able to lock position was still requested by people and I had to justify that decision more than once. 

Put it differently: how would you lock altitude without knowing the vehicle altitude?

Jaeyoung-Lim:
@Stifael Thought more about this, but why are we not using the acceleration message instead of using roll + pitch + altitude hold?The outcome would be quite similar, while resulting in a more clean and sensible message definition (nearly nothing to change)

Stifael:
I think you have to explain your reasoning a bit more. How do you convert acceleration into attitude without misusing it's definition? Acceleration is still a translational quantiity wheras attitude is a rotational one.

Jaeyoung-Lim:
@Stifael The way I see it is that one of the nicest properties of multirotor dynamics are that it is differentially flat, meaning the system can be fully defined through the output and the differentiation of the outputs.

In case of multirotors, this is position, velocity, acceleration and yaw. ([Lee 2010](https://arxiv.org/pdf/1109.4457.pdf)) With these "flat" outputs the attitude can be fully defined. Neglecting drag, the acceleration is fully defined with thrust and gravity. But even including drag models the dynamics are still differentially flat ([Feassler 2017](https://arxiv.org/abs/1712.02402))

The reason I am having opinions on adding roll / pitch messages to the `SET_POSITION_TARGET_LOCAL_NED` is that the current fields fit with the flat outputs perfectly and this is exactly what you want with trajectory planning / tracking applications. Adding roll / pitch in this message makes the message no longer fits clearly to this use case anymore. Moreover, roll / pitch is a redundant information since this can already be fully defined from the acceleration.

That is the reason I was suggesting maybe `SET_ATTITUDE_TARGET` might fit better to the use case for the following reasons.
- The information fits the use case better - it is setting the attitude target
- Most people using the `SET_ATTITUDE_TARGET` message struggles to do sufficient thrust control to hold altitude. Holding the altitude will make people's live much easier
-  One of the reason `SET_ATTITUDE_TARGET` as a offboard message doesn't end up as a good choice is because yaw estimates are usually not good enough to do any high acceleration control

Stifael:
> Adding roll / pitch in this message makes the message no longer fits clearly to this use case anymore.

**Please** read my comments above. I don't have a strong opinion about where to add the extension.

> One of the reason SET_ATTITUDE_TARGET as a offboard message doesn't end up as a good choice is because yaw estimates are usually not good enough to do any high acceleration control

Agreed if not external estimation has been added. Please read my comments above about `reference for attitude`. That is why there is **NO** yaw in the objective.

I think we are going here in circles. The acceleration-setpoint might be fine, and maybe you can open a new issue / feature request. This issue has be implemented in our source code based on the request from another company.  I can try to invite him to participate in this discussion. But then again, during the mavlink dev-call the plan was to decide about how to add the new message (please read #1302) and not about a different implementation strategy.

Stifael:
In addition, PX4 does not yet support acceleration-setpoint anyway (there is the ongoing work to use it as feedforward). Hence, the suggestion to use acceleration is right now really a feature request on the PX4 side. Roll/pitch/yaw-rate and velocity is basically already supported, but the mavlink-msg is still required.

message_definitions/v1.0/common.xml
```diff
@@ -4758,6 +4764,10 @@
       <field type="float" name="afz" units="m/s/s">Z acceleration or force (if bit 10 of type_mask is set) in NED frame in meter / s^2 or N</field>
       <field type="float" name="yaw" units="rad">yaw setpoint</field>
       <field type="float" name="yaw_rate" units="rad/s">yaw rate setpoint</field>
+      <extensions/>
```
hamishwillee: If sent by a system that has not been updated these values will all be received as 0. Normally 0 then is used to mean "not supported or not changed". It might be best if we use 0 to indicate "not supplied" and 2rad to indicate the 0 point? Or is that going to be a problem?

message_definitions/v1.0/common.xml
hamishwillee: Are we saying here that hold_altitude is a flag? ie a value of 1 indicates that altitude should be held?

hamishwillee:
So would a summary be that @Jaeyoung-Lim thinks it should  go in `SET_ATTITUDE_TARGET` and @Stifael doesn't particularly care (because he'd rather have a new message :-). 

I don't understand the arguments for/against. My only point would be that from a mavlink extensions point of view 0 has to be treated as the invalid/ignore value. 

@julianoes is back in a couple of days and may have further comments.

LorenzMeier:
I would strongly recommend to sort these things on the MAVLink call, rather than going back and forth in lengthy discussions.

message_definitions/v1.0/common.xml
```diff
@@ -3513,6 +3513,12 @@
       <entry value="2048" name="POSITION_TARGET_TYPEMASK_YAW_RATE_IGNORE">
         <description>Ignore yaw rate</description>
       </entry>
+      <entry value="4096" name="POSITION_TARGET_TYPEMASK_ROLL_IGNORE">
+        <description>Ignore roll</description>
+      </entry>
+      <entry value="8192" name="POSITION_TARGET_TYPEMASK_PITCH_IGNORE">
+        <description>Ignore pitch</description>
+      </entry>
```
LorenzMeier: Adding roll / pitch to a position target doesn't make any sense. This PR needs to be reworked (a clean offfboard attitude set point).

LorenzMeier:
I'm closing this. This should be changed instead to a PR that creates a new message long the lines of https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET (e.g. SET_ALTITUDE_TARGET).

It makes no sense to mix roll/pitch with position from an interface perspective.

hamishwillee:
@Stifael Sorry this ended up reverting to your original option "1"; Lorenz is better informed than us. At least you have clear direction now.
Note, for us to proceed on any message we need a level of engagement/stakeholder to commit to implementation.

PR #1311, opened 2020-02-19
Add flag to indicate simulation with lockstep

julianoes:
We would like to use a flag marked for future use to indicate that lockstep is being used. This means that simulators can be configured automatically instead of needing a config argument.

This was first proposed by @lovettchris in https://github.com/PX4/Firmware/pull/13442.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4878,7 +4878,7 @@
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
       <field type="float[16]" name="controls">Control outputs -1 .. 1. Channel assignment depends on the simulated hardware.</field>
       <field type="uint8_t" name="mode" enum="MAV_MODE_FLAG" display="bitmask">System mode. Includes arming state.</field>
-      <field type="uint64_t" name="flags" display="bitmask">Flags as bitfield, reserved for future use.</field>
+      <field type="uint64_t" name="flags" display="bitmask">Flags as bitfield, 1: indicate simulation using lockstep.</field>
     </message>
     <message id="100" name="OPTICAL_FLOW">
       <description>Optical flow from a flow sensor (e.g. optical mouse sensor)</description>


LorenzMeier:
I'm wondering if a separate simulation status message wouldn't be the better approach. I'd like to keep the system completely unchanged in its state vectors when simulation is enabled as a principle. Could you propose a PR with a simulation state message instead?

This should then also be sent by the simulator connector, NOT by the regular MAVLink instance. We need a separation of concerns here.

hamishwillee:
I'll review again when you've addressed https://github.com/mavlink/mavlink/pull/1311#issuecomment-590204923
I don't understand the arguments for simulation status message, but if we do this, what other status would usefully be sent?

julianoes:
What else would go into the status message? I literally only need one bit, and I don't think we need to create and send yet another message just for that, given we already have 64bits reserved for something like that. The lockstep information also needs to come in sync with the actuator controls in order for the simulator to know how to use it, so a separate message would make this more complicated.

LorenzMeier:
The main issue I have is the principle of making a SITL system non-transparent. That's a no-no from a simulator design perspective to me because it alters the system output.

In terms of status, here is what I think would make sense long term (starting with the realtime factor). I would send them out in MAVLink and log them separately (separate file) as simulation log.

- Realtime factor
- Lockstep status
- Wether or not SITL / HITL is enabled
- If in HITL actuators are "hot" (metal bird)
- Realtime jitter (variance of timestamps between lockstep frames on host OS)
- Timestamp jitter (variance of timestamps inside the lockstep time)
- Uptime of this simulation instance (how long is the system booted)
- Wallclock time that is being simulated (is it day / night, etc)
- A simulation run UUID (if you have multiple logs, how can you reference them to the same thing)
- A configuration UUID (using this UUID you can reproduce the exact environment)

julianoes:
Ok, I think we are not talking about the same thing.

This message is used between autopilot and simulator. It is not for a GCS. A simulator needs to know whether the simulation should happen in lockstep and it needs to wait for responses in each cycle or whether it can "freewheel".

LorenzMeier:
Cool, thanks for closing the loop - I've expanded the diff after your hint and that was the missing context. It all makes sense now, merging.

PR #1316, opened 2020-03-06
adding bitmask="true" for enums

BazookaJoe1900:
this can help decoders to decide to look for exact value, or mix of bits

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -190,7 +190,7 @@
         <description>official stable release</description>
       </entry>
     </enum>
-    <enum name="HL_FAILURE_FLAG">
+    <enum name="HL_FAILURE_FLAG" bitmask="true">
       <description>Flags to report failure cases over the high latency telemtry.</description>
       <entry value="1" name="HL_FAILURE_FLAG_GPS">
         <description>GPS failure.</description>
@@ -235,7 +235,7 @@
         <description>Mission failure.</description>
       </entry>
     </enum>
-    <enum name="MAV_MODE_FLAG">
+    <enum name="MAV_MODE_FLAG" bitmask="true">
       <description>These flags encode the MAV mode.</description>
       <entry value="128" name="MAV_MODE_FLAG_SAFETY_ARMED">
         <description>0b10000000 MAV safety set to armed. Motors are enabled / running / can start. Ready to fly. Additional note: this flag is to be ignore when sent in the command MAV_CMD_DO_SET_MODE and MAV_CMD_COMPONENT_ARM_DISARM shall be used instead. The flag can still be used to report the armed state.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @BazookaJoe1900 , but this will break the schema checks: https://mavlink.io/en/guide/xml_schema.html

The bitmask is declared when the enum is **used**. That makes sense because that is the point the user of the API is creating a UI - which is where QGC at least uses the information.
If you think this adds value over or in addition to having it at that point (and I can see arguments for that) then you would need to update the schema doc first to support this.

Thoughts?

BazookaJoe1900:
@hamishwillee, Yes I see value of this metadata. and I understand its break the current schema.
Basically every data that the XML can give, as soon as possible is better, then in later stages. I look at the XML as source code, of course that later it being translated to different languages (C, C++, C#, LUA). there is no need to spread the metadata on the ground-station (and tell on the GCS that thas bitmast is bit-field, and other is not)

hamishwillee:
@DonLakeFlyer The proposal here is to move the enum "this is a bitmask" information to the bitmask, as well as/possibly instead of the place where the enum is used.

As the GCS is the main consumer of this information, does it make any/much difference to you?

```
<enum name="HL_FAILURE_FLAG" bitmask="true">
```

FWIW I actually like this idea from an information management point of view. But not sure of the cost to existing consumers.

DonLakeFlyer:
Change makes sense. Also just FYI: At this point QGC does not currently use this information.

hamishwillee:
> Change makes sense. Also just FYI: At this point QGC does not currently use this information.

Thanks @DonLakeFlyer . I thought you used it to render enums as checkboxes. Learn something new every day :-)

hamishwillee:
@BazookaJoe1900 OK, since Don is good with this, can you please create a PR in https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd to allow this syntax and then cross link to here (since this can't go in until that merges). We'll remove the `display="bitmask"` as a later pass.

DonLakeFlyer:
> I thought you used it to render enums as checkboxes. Learn something new every day :-)

QGC has had it's own internal system for meta data on commands prior to any of this existing in mavlink. The majority of these mavlink changes mirror stuff which already exists in QGC in that internal system. At some point it will switch to using the mavlink data, but right now that's not at the top of the list to get done.

amilcarlucas:
I stll do not get why is this better than the existing display=bitmask

hamishwillee:
@amilcarlucas  If the enum is tagged as representing a bitmap then this approach ensures that this gets reflected wherever the enum is used. Essentially that is less prone to error (and less verbose) than choosing to tag wherever the bitmap is used.

DonLakeFlyer:
As I said above I don't currently use any of this in QGC. At some point in the future I will.

BazookaJoe1900:
Its sound to me like it ok to go with the change

hamishwillee:
@amilcarlucas @peterbarker Can you please merge then.

peterbarker:
Appears to require a rebase here.

BazookaJoe1900:
rebased

peterbarker:
On Wed, 25 Mar 2020, BazookaJoe1900 wrote:

> rebased

Thanks.

The pymavlink PR's been merged now, so you should be able to get this to 
pass CI now.

... might require someone with mavlink-repo privileges to push the retry 
button.

hamishwillee:
I'm confused @peterbarker - this  says isn't merged. Once it is, I could pull it upstream.

peterbarker:
@hamishwillee 

The pymavlink PR referenced on this PR has been merged.  Was there another you were thinking of?

image

hamishwillee:
@peterbarker I am an idiot. Ignore me. I will submit PR to update Pymavlink so this can merge.

hamishwillee:
Merged. Thanks @BazookaJoe1900 !

hamishwillee:
Next step - to remove the values from fields :-)

PR #1317, opened 2020-03-06
Update according to the final ASTM Remote ID standard and remove WIP

friissoren:
I did the final smaller edits needed to align the message definitions with the final ASTM Remote ID standard.

I have uploaded some documentation here:
https://github.com/mavlink/mavlink-devguide/pull/220

Once that PR is approved and merged, I assume we should be able to remove the Work in Progress marks as well?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6477,7 +6477,7 @@
     <message id="12900" name="OPEN_DRONE_ID_BASIC_ID">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change, and should NOT be used in stable production environments -->
-      <description>Data for filling the OpenDroneID Basic ID message. This and the below messages are primarily meant for feeding data to/from an OpenDroneID implementation. E.g. https://github.com/opendroneid/opendroneid-core-c. See also the ASTM Remote ID standard at https://www.astm.org/Standards/F3411.htm.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -3857,13 +3857,13 @@
       <entry value="1" name="MAV_ODID_UA_TYPE_AEROPLANE">
         <description>Aeroplane/Airplane. Fixed wing.</description>
       </entry>
-      <entry value="2" name="MAV_ODID_UA_TYPE_ROTORCRAFT">
-        <description>Rotorcraft (including Multirotor).</description>
+      <entry value="2" name="MAV_ODID_UA_TYPE_HELICOPTER_OR_MULTIROTOR">
+        <description>helicopter or Multirotor.</description>
```
hamishwillee: Helicopter or multirotor.

hamishwillee:
Generally we remove the WIP when we're sure we have a workable API, and that is the point their is an implementation. Is there a PX4, ArduPilot or other flight stack implementation - and if not, is one planned?

@auturgy @LorenzMeier @julianoes Any comments on this final revision of the messages - ie expectation is that the ASTM remote ID messages will not be "stable".

magicrub:
I could probably whip up an ArduPilot implantation.

hamishwillee:
@magicrub That would be great (though I don't know if there is a "real" module to play with). Just looking at the [docs](https://github.com/mavlink/mavlink-devguide/pull/220) we might be missing a component ID.

auturgy:
@apache405 might have hardware.
Re stability: It’s unlikely to change in the short term. Takes a long time to get through F38, so I think remove WIP.

friissoren:
Review comment fixed.

About implementations, I don't know if anyone else than me/Intel have used these new Mavlink messages. I have used them for sending data from a flight controller to a Bluetooth transmitter module in a so-far unpublished Intel drone.

PR #1318, opened 2020-03-06
some MAV_CMD metadata updates

BazookaJoe1900:
this is rain-check to see if the changes are acceptable, what was done is:
* Adding labels for lat/lon/alt params
* rename params description to unify names ('RESEREVED' to 'Reserved', 'Reserved, send 0' to 'Reserved (set to 0)', 'sysid' to 'System ID', etc)

more changes can/should be done, like:
* adding units to the metadata, 
* add labels to all of the params
* adding units, minValue, maxValue, enum, increment

and so on

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1279,7 +1279,7 @@
         <param index="1" label="Turns" minValue="0">Number of turns.</param>
         <param index="2">Empty</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
+        <param index="4" label="Xtrack Location">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
         <param index="7" label="Altitude" units="m">Altitude</param>
@@ -1289,7 +1289,7 @@
         <param index="1" label="Time" units="s" minValue="0">Loiter time.</param>
         <param index="2">Empty</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise.</param>
-        <param index="4">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle.  NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
+        <param index="4" label="Xtrack Location">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle.  NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
         <param index="7" label="Altitude" units="m">Altitude</param>
@@ -1402,9 +1402,9 @@
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Vehicle configuration default.</param>
         <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
         <param index="4">Reserved (e.g. for dynamic center beacon options)</param>
-        <param index="5">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
-        <param index="6">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
-        <param index="7">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
+        <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
+        <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
+        <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
@@ -1423,9 +1423,9 @@
         <param index="2" label="Global Ctrl" minValue="0" maxValue="3" increment="1">0: Disable full path planning (without resetting map), 1: Enable, 2: Enable and reset map/occupancy grid, 3: Enable and reset planned route, but not occupancy grid</param>
         <param index="3">Empty</param>
         <param index="4" label="Yaw" units="deg">Yaw angle at goal</param>
-        <param index="5">Latitude/X of goal</param>
-        <param index="6">Longitude/Y of goal</param>
-        <param index="7">Altitude/Z of goal</param>
+        <param index="5" label="Latitude/X">Latitude/X of goal</param>
+        <param index="6" label="Longitude/Y">Longitude/Y of goal</param>
+        <param index="7" label="Altitude/Z">Altitude/Z of goal</param>
       </entry>
       <entry value="82" name="MAV_CMD_NAV_SPLINE_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint using a spline path.</description>
@@ -1433,9 +1433,9 @@
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5">Latitude/X of goal</param>
-        <param index="6">Longitude/Y of goal</param>
-        <param index="7">Altitude/Z of goal</param>
+        <param index="5" label="Latitude/X">Latitude/X of goal</param>
+        <param index="6" label="Longitude/Y">Longitude/Y of goal</param>
+        <param index="7" label="Altitude/Z">Altitude/Z of goal</param>
       </entry>
       <entry value="84" name="MAV_CMD_NAV_VTOL_TAKEOFF" hasLocation="true" isDestination="true">
         <description>Takeoff from ground using VTOL mode, and transition to forward flight with specified heading.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
There is some overlap with https://github.com/mavlink/mavlink/pull/1284

FWIW I personally don't see this sort of change as essential and creates quite a lot of churn. Consistency is nice though. 
FYI Specifically reserved and empty are synonyms, and nowdays you could just remove the entry (the toolchain on our generated libraries auto-generates the missing elements)

I see lots of value in "adding units, minValue, maxValue, enum, increment" (where these make sense)

@amilcarlucas Thoughts? #1284 got bogged on some questions. Good to get these in or closed.

BazookaJoe1900:
@hamishwillee, thanks for the response.
* About the value of the metadata. I see it valuable. In my costum ground station the metadata of the commands is used. the commands are well structured structs, named by the label (and not param1, param2...).
I basically wanted to do that on the PX4 code too, that the commands are structured, but that might be for later work.

 * About reserved,  empty and so, do you want me to make new pr that remove them all, the common file will be much cleaner...

* Consistency on names (sysid, sytem_id.... and other), Do you want me to make deeper cleanup?

and finel question, to-do each on different pr's?

hamishwillee:
Hi @BazookaJoe1900 

Lets wait on @amilcarlucas before making any changes.

> About the value of the metadata. I see it valuable. In my costum ground station the metadata of the commands is used. the commands are well structured structs, named by the label (and not param1, param2...).
> I basically wanted to do that on the PX4 code too, that the commands are structured, but that might be for later work.

I understand the rationale. My point is that by definition the item types that have the hasLocation or isDestination flags will always be lat, lon, altitude in param 4, 5, 6. 
You could argue that it is worth adding the label anyway, or that a GCS can assume the label, reducing the size of the file, duplication, and chance of spelling errors etc.

> About reserved, empty and so, do you want me to make new pr that remove them all, the common file will be much cleaner...

No, I'll do this at some point if we have general agreement - @amilcarlucas? It is not quite as simple as I described. 

> Consistency on names (sysid, sytem_id.... and other), Do you want me to make deeper cleanup?

@amilcarlucas To me this is nice, but I wouldn't spend my time on for the consistency. Do you have preferred way you'd like to see these named?

BazookaJoe1900:
@hamishwillee. I agree that we need broad agreement for such changes. 
about the first point:
if 'hasLocation' and 'isDestination' can be the flags to tell the full information (scale, label and so on) about param 4, 5, 6 -> then I think we should remove the lines.
but note that currently this is the list of the names I found in the params for altitude and latitude:
can you assure that all means the same on all the commands? same scale? same explanation?

Altitude
Landing   altitude (ground level in current frame).
Altitude
Desired   altitude
Center   point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to   MAV_FRAME. NaN: Use current vehicle position or current center if already   orbiting.
Altitude/Z   of goal
Altitude   (ground level)
Finish   Altitude
Altitude   (meters)
Altitude   of ROI location
MAV_ROI_WPNEXT:   yaw offset from next waypoint, MAV_ROI_LOCATION: altitude
Altitude/Z   position.
Altitude   (MSL), in meters


Latitude
Center   point latitude (if no MAV_FRAME specified) / X coordinate according to   MAV_FRAME. NaN: Use current vehicle position or current center if already   orbiting.
Latitude/X   of goal
Latitude   of ROI location
MAV_ROI_WPNEXT:   pitch offset from next waypoint, MAV_ROI_LOCATION: latitude
latitude,   set if appropriate mount mode.
Latitude/X   position.
Unscaled   target latitude of center of circle in CIRCLE_MODE
Latitude   unscaled for MISSION_ITEM or in 1e7 degrees for MISSION_ITEM_INT
Latitude   unscaled

amilcarlucas:
I need some time to review this. I hope to do it next week.

message_definitions/v1.0/common.xml
```diff
@@ -2313,9 +2313,9 @@
         <param index="2" label="Altitude">Altitude: 0: ignore altitude</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5">Latitude</param>
-        <param index="6">Longitude</param>
-        <param index="7">Altitude</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude_2">Altitude</param>
```
amilcarlucas: Nope, I prefer the way I did it here.

BazookaJoe1900: @amilcarlucas, I am not sure what you meant there. 
if you meant that I added Altitude_2 as label for param7, that is usually just Altitude, I agree. This message has small issue, because it has two altitudes, and at this point I didn't want to rename the first altitude that is already labeled.
BTW, in order to evaluate the parameters its much easier to look and filter them as table, and i used [this convertor](https://conversiontools.io/convert/xml-to-excel) to change to excel table.

amilcarlucas: Take a look on how I did it on #1284

BazookaJoe1900: @amilcarlucas, So what to do here? How to call param 2, and how to call param 7?
I prefer that param 7 will keep being Altitude and that param 2 will be renamed to gate altitude
Agree?

BazookaJoe1900: @amilcarlucas ping

amilcarlucas: I merged my PR into master. Rebase your's and you will understand what I meant.

BazookaJoe1900:
@amilcarlucas, Thanks! I rebased.
your changes looks great, and about the other changes I made, they looks minor now, so feel free to merge or drop this pr.

hamishwillee:
@amilcarlucas I'll leave you to decide whether this merges or gets dropped. I have no strong feeling either way.

message_definitions/v1.0/common.xml
```diff
@@ -2009,9 +2009,9 @@
         <param index="2" label="Position" enum="MAV_GOTO">MAV_GOTO_HOLD_AT_CURRENT_POSITION: hold at current position, MAV_GOTO_HOLD_AT_SPECIFIED_POSITION: hold at specified position.</param>
         <param index="3" label="Frame" enum="MAV_FRAME">Coordinate frame of hold point.</param>
         <param index="4" label="Yaw" units="deg">Desired yaw angle.</param>
-        <param index="5">Latitude / X position.</param>
-        <param index="6">Longitude / Y position.</param>
-        <param index="7">Altitude / Z position.</param>
+        <param index="5" label="Latitude">Latitude/X position.</param>
```
amilcarlucas, line 2012: I guess the label should be "Latitude/X"

amilcarlucas:
Does these change make the use of the "Reserved" keyword consistent in the entire file?

BazookaJoe1900:
> Does these changes make the use of the "Reserved" keyword consistent in the **entire** file?
yes
but note that there are some points that can be done to even make the file more consiste. 
- there are many _"Empty"_ labels
- there are places with different reserved strings:
_"Reserved (all remaining params)"_
_"Reserved (Set to 0)"_
_"Reserved, send 0"_
_"Reserved, set to NaN"_

> Does these changes make the use of the "System ID" keyword consistent in the **entire** file?
yes, for params on MAV_CMD. not as every location that system ID might reference to.

as I sayied on the first thread, there are also commands that has no labels.

amilcarlucas:
Can you please add the missing MAV_CMD labels?

BazookaJoe1900:
@amilcarlucas sure. can you check my changes?

1. There are many "User defined" comments. I am not sure how to label them.
2. "Xtrack Location" can you please check correct label?
3. for MAV_CMD_DO_SET_ROI, I didn't labeled, because the command is deprecated, and I am not sure anyway how to label them
4. I didn't touch MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN because it WIP

BazookaJoe1900:
@amilcarlucas, ping. can you check and merge changes?

BazookaJoe1900:
rebased, @amilcarlucas. can you check the changes?

message_definitions/v1.0/common.xml
```diff
@@ -1805,20 +1805,20 @@
...
       </entry>
       <!-- this one is messed up! altitude should be param 7, not param4 -->
       <entry value="205" name="MAV_CMD_DO_MOUNT_CONTROL" hasLocation="false" isDestination="false">
         <deprecated since="2020-01" replaced_by="MAV_CMD_DO_GIMBAL_MANAGER_ATTITUDE">This message is ambiguous and inconsistent. It has been superseded by MAV_CMD_DO_GIMBAL_MANAGER_ATTITUDE and MAV_CMD_DO_SET_ROI_*. The message can still be used to communicate with legacy gimbals implementing it.</deprecated>
         <description>Mission command to control a camera or antenna mount</description>
-        <param index="1">pitch depending on mount mode (degrees or degrees/second depending on pitch input).</param>
-        <param index="2">roll depending on mount mode (degrees or degrees/second depending on roll input).</param>
-        <param index="3">yaw depending on mount mode (degrees or degrees/second depending on yaw input).</param>
+        <param index="1" label="Pitch Mode">pitch depending on mount mode (degrees or degrees/second depending on pitch input).</param>
```
hamishwillee, line 1816: This is pitch, not pitch mode. It depends on the mode.

message_definitions/v1.0/common.xml
```diff
@@ -1805,20 +1805,20 @@
...
+        <param index="2" label="Roll Mode">roll depending on mount mode (degrees or degrees/second depending on roll input).</param>
```
hamishwillee, line 1817: Roll, not Roll mode.

message_definitions/v1.0/common.xml
```diff
@@ -1805,20 +1805,20 @@
...
+        <param index="3" label="Yaw Mode">yaw depending on mount mode (degrees or degrees/second depending on yaw input).</param>
```
hamishwillee, line 1818: Yaw, not yaw mode.

hamishwillee:
@amilcarlucas I added a few comments on Roll, Pitch, Yaw mode, but otherwise I am happy with this. If you have any other issues let me know - or I will merge this after those are fixed.

@BazookaJoe1900  Yes there are a lot of other potential changes, but I'd rather be iterative. Lets get this in.

---

As an aside, with respect to https://github.com/mavlink/mavlink/pull/1318#issuecomment-608395841, usage of Empty, reserved, whatever. This is indeed inconsistent. My understanding is that we only support one concept for an unassigned parameter, and that is "reserved". So even if it says "Empty" we really do not want people to go and reuse for their own purposes without going through the process.

The expectation is that if no parameter is defined for a particular index, then the param is considered reserved with a default value of 0. If you wanted you could also write this as either of:
```
<param index="3" reserved="true" default="0"/>
<param index="3" reserved="true"/>
```
If the param needs to have a default value of "NaN" then you need to explicitly specify it like this:
```
<param index="3" reserved="true" default="NaN"/>
```
What this means is that we should be able to go through and delete all the items that say "Empty" as these will then defaul to reserved with default value of 0. Ditto for anything that currently says "Reserved" or "Reserved for now".

The only changes will be to add the default NaN case if needed, which I thnk I have already done.

The REASON I have not done that change is that more testing of the generation of docs is needed. I put a PR in to auto-create items and docs for the missing items, but I need to check that it works properly for the in-source generated documentation because PeterB partially reverted a change. Have not go around to that. On my very big list. Feel free to verify yourself if you are so inclined.

message_definitions/v1.0/common.xml
BazookaJoe1900: @hamishwillee, fixed.
I am not sure how to hadale the units here, because it can be both deg or deg/sec

BazookaJoe1900:
@hamishwillee, I think we can leave the 'reserved' thing for later. not in this pr

hamishwillee:
Just leave out the units -is is the fields that suck :-). This is good for me. @amilcarlucas OK for me to merge?

message_definitions/v1.0/common.xml
amilcarlucas: yes, this message is a mess, it's imposible to document it properly with the current xml schema.
But is is deprecated, so it will go away at some time :)

message_definitions/v1.0/common.xml
```diff
@@ -2338,8 +2338,8 @@
         <param index="2">User defined</param>
         <param index="3">User defined</param>
         <param index="4">User defined</param>
-        <param index="5">Unscaled target latitude of center of circle in CIRCLE_MODE</param>
-        <param index="6">Unscaled target longitude of center of circle in CIRCLE_MODE</param>
+        <param index="5" label="Latitude" units="deg">Unscaled target latitude of center of circle in CIRCLE_MODE</param>
+        <param index="6" label="Longitude" units="deg">Unscaled target longitude of center of circle in CIRCLE_MODE</param>
```
amilcarlucas, line 2342: Is this deg, or degE7 ?

message_definitions/v1.0/common.xml
```diff
@@ -1805,20 +1805,20 @@
...
-        <param index="1">pitch depending on mount mode (degrees or degrees/second depending on pitch input).</param>
-        <param index="2">roll depending on mount mode (degrees or degrees/second depending on roll input).</param>
-        <param index="3">yaw depending on mount mode (degrees or degrees/second depending on yaw input).</param>
+        <param index="1" label="Pitch">pitch depending on mount mode (degrees or degrees/second depending on pitch input).</param>
+        <param index="2" label="Roll">roll depending on mount mode (degrees or degrees/second depending on roll input).</param>
+        <param index="3" label="Yaw">yaw depending on mount mode (degrees or degrees/second depending on yaw input).</param>
         <param index="4" label="Altitude" units="m">altitude depending on mount mode.</param>
-        <param index="5">latitude, set if appropriate mount mode.</param>
-        <param index="6">longitude, set if appropriate mount mode.</param>
+        <param index="5" label="Latitude">latitude, set if appropriate mount mode.</param>
```
amilcarlucas, line 1820: missing units

message_definitions/v1.0/common.xml
BazookaJoe1900: @amilcarlucas, I am not sure. this command is not mantioned on PX4 code. and I couldn't find how it used in Ardu. can you check

message_definitions/v1.0/common.xml
BazookaJoe1900: @amilcarlucas , this one and on many other command params there is no units.
it might be needed another deep sweep on the XML to set unit to all the commands. I couldn't do it on this pr.
(this message is deprecated anyway, by the way)

hamishwillee: Yep. In particular this one is hard because the gimbal API is rubbish (and has been superseded). More generally, latitude/longitude are often different units depending on whether setting a local or global frame. Ignore for now.

message_definitions/v1.0/common.xml
hamishwillee: I don't think this is used in ArduPilot either. @amilcarlucas I'd propose we set to degE7?

hamishwillee: @amilcarlucas ^^^^ When you have a moment.

BazookaJoe1900: @amilcarlucas ?

hamishwillee: @amilcarlucas Actually the description says "unscaled" so deg is intentional. Whether it is the "right design" is another matter - but this is correct in the sense "does not break anyone who has followed the stated behaviour". Setting resolved.

hamishwillee:
@amilcarlucas Circle submode is not implemented on PX4. PX4 supports orbit, but this is in another mode context. Given no one uses this "that we know", I'm going to change to degE7 as more consistent design. 

In future I think we should deprecate this and replace it with a do orbit message that works for both PX4 and ArduPilot and others. That can be used to implement an orbit whether in circle mode or some other approach.

hamishwillee:
Merged. Thanks all!

PR #1319, opened 2020-03-10
Allow correlating CAMERA_IMAGE_CAPTURED with the camera command.

remster:
We fly (image) survey missions (grids, individual photos, with `MAV_CMD_IMAGE_START_CAPTURE` and `MAV_CMD_DO_SET_CAM_TRIGG_DIST`). We then process thus captured photos and therefore need to know what they are (their urls) and what command each corresponds to. We receive `CAMERA_IMAGE_CAPTURED`  - both response to `MAV_CMD_IMAGE_START_CAPTURE` and `MAV_CMD_DO_SET_CAM_TRIGG_DIST`, but not when photos were taken during the time the drone was disconnected. We also don't know what command these images correspond to. Because of that we cannot work out what photos did the drone take during a mission and don't thus know what to download.

There is the WIP [`MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE`](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE) (though its [WIP status](https://github.com/mavlink/mavlink/commit/da2f75e3937035f79b7ceb3403d07e26382ad1db) appears stale - 2 years). Provided its WIP status can be lifted, we think can use it to request missing `CAMERA_IMAGE_CAPTURED` events, but it is insufficient to allow us correlate these photos with commands. 

This PR is our first PR and the first stab in understanding this exchange. Here's how we justify these changes:

- remove unclear/unnecessary(?) caveat about the sequence number only applying to `MAV_CMD_IMAGE_START_CAPTURE` with image count of 1. We'd like to propose that the sequence number is semantically an asset number and an asset can be a group of photos resulting from one command. As a consequence we clarify that multiple `CAMERA_IMAGE_CAPTURED` can share the same sequence number. Elsewhere (QGC) we will propose to use this sequence number as a grouping folder for the associated photos. This will allow processing them automatically. 

- because `CAMERA_IMAGE_CAPTURED` is also sent in response to `MAV_CMD_DO_SET_CAM_TRIGG_DIST` allow specifying sequence there too;

- Add sequence field to `CAMERA_IMAGE_CAPTURED` event to allow correlating it with a specific `MAV_CMD_IMAGE_START_CAPTURE` or `MAV_CMD_DO_SET_CAM_TRIGG_DIST`. We don't know if there are any backwards compatibility requirements we need to obey;

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1821,7 +1821,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
+        <param index="4" label="Sequence Number" minValue="0" increment="1">Capture sequence number starting from 1  (0 indicates sequence number is not supported/used). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
@@ -2175,7 +2175,7 @@
         <param index="1">Reserved (Set to 0)</param>
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Capture Count" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
-        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
+        <param index="4" label="Sequence Number" minValue="0" increment="1">Capture sequence number starting from 1  (0 indicates sequence number is not supported/used). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
         <param index="5">Reserved (all remaining params)</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
@@ -2186,8 +2186,8 @@
       <entry value="2002" name="MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE" hasLocation="false" isDestination="false">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Re-request a CAMERA_IMAGE_CAPTURE message. Use NaN for reserved values.</description>
-        <param index="1" label="Number" minValue="0" increment="1">Sequence number for missing CAMERA_IMAGE_CAPTURE messages</param>
+        <description>Re-request all CAMERA_IMAGE_CAPTURED messages that have the specified sequence number. Use NaN for reserved values.</description>
+        <param index="1" label="Sequence Number" minValue="0" increment="1">Sequence number for missing CAMERA_IMAGE_CAPTURED messages</param>
         <param index="2">Reserved (all remaining params)</param>
       </entry>
       <entry value="2003" name="MAV_CMD_DO_TRIGGER_CONTROL" hasLocation="false" isDestination="false">
@@ -5972,7 +5972,7 @@
       <field type="int32_t" name="image_index">Zero based index of this image (image count since armed -1)</field>
       <field type="int8_t" name="capture_result">Boolean indicating success (1) or failure (0) while capturing this image.</field>
       <field type="char[205]" name="file_url">URL of image taken. Either local storage or http://foo.jpg if camera provides an HTTP interface.</field>
-      <field type="int32_t" name="sequence">If non-zero, sequence number of the MAV_CMD_IMAGE_START_CAPTURE or MAV_CMD_DO_SET_CAM_TRIGG_DIST command this image is a response to.</field>
+      <field type="int32_t" name="sequence">Sequence number of the MAV_CMD_IMAGE_START_CAPTURE or MAV_CMD_DO_SET_CAM_TRIGG_DIST command this image is a response to. 0 if the sequence number is not known or supported.</field>
     </message>
     <message id="264" name="FLIGHT_INFORMATION">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1821,7 +1821,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Idenifty this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
@@ -2175,7 +2175,7 @@
         <param index="1">Reserved (Set to 0)</param>
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Capture Count" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
-        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Idenifty this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
         <param index="5">Reserved (all remaining params)</param>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1821,7 +1821,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4">Empty</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Idenifty this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1821,7 +1821,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4">Empty</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Idenifty this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
```
remster: This message is semantically very similar (bordering redundant) to MAV_CMD_IMAGE_START_CAPTURE.

hamishwillee: Yes. Though it's been around a long time. @dogmaphobic What's the point of having both?

message_definitions/v1.0/common.xml
```diff
@@ -2175,7 +2175,7 @@
...
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Capture Count" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
-        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it.</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Idenifty this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
```
hamishwillee: @dogmaphobic Is this correct? 

Note, typo "Idenifty"

hamishwillee:
@dogmaphobic You're the camera API expert, can you please review this and suggest a way forward?

The top level problem being solved is how to get the `CAMERA_IMAGE_CAPTURED` information for messages that were missed due to out of range. 

I do see this as a problem - the sequence number is ill defined and does not have a clear mapping to what is meant by `CAMERA_IMAGE_CAPTURED.image_index`. Upshot, I'm not sure that `MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE` can work as is.

message_definitions/v1.0/common.xml
dogmaphobic: > Yes. Though it's been around a long time. @dogmaphobic What's the point of having both?

Mostly on the assumption all these old MAV_CMD_DO_DIGICAM_XXX messages were old and deprecated. They don't at all allow fine camera control. They mostly only handle cameras with single analog control (shutter trigger). That's why a whole new camera API was put in place. This particular MAV_CMD_DO_SET_CAM_TRIGG_DIST message is a bit on an exception. Missions still use it. Ideally we would add this functionality to the new camera API (with proper arguments) and deprecate this as well.

message_definitions/v1.0/common.xml
dogmaphobic: > Note, typo "Idenifty"

Yes, that's correct.

message_definitions/v1.0/common.xml
remster: :+1: 
That way I wouldn't need to add sequence number to `MAV_CMD_DO_SET_CAM_TRIGG_DIST` - but QGC would need to first migrate away from it. This is not done so easily, so I think my change needs to stay (as we also need to identify images taken with `MAV_CMD_DO_SET_CAM_TRIGG_DIST`)

message_definitions/v1.0/common.xml
remster: typo fixed

remster:
> I do see this as a problem - the sequence number is ill defined and does not have a clear mapping to what is meant by `CAMERA_IMAGE_CAPTURED.image_index`. Upshot, I'm not sure that `MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE` can work as is.

Not to  `CAMERA_IMAGE_CAPTURED.image_index`, but yes to  `CAMERA_IMAGE_CAPTURED.sequence`, which I am adding herewith.

message_definitions/v1.0/common.xml
```diff
@@ -1821,7 +1821,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4">Empty</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
```
hamishwillee: - minValue should be set to 0. It means "not valid" but the value itself is still settable - if you get my meaning?
- Please modify the bit in brackets to: (0 indicates sequence number is not supported/used). "

message_definitions/v1.0/common.xml
```diff
@@ -2175,7 +2175,7 @@
...
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Capture Count" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
-        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it.</param>
+        <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1 (use 0 to ignore it). This is to: (1) Identify this command and prevent double captures when the command is re-transmitted (2) Correlate CAMERA_IMAGE_CAPTURED events (and underlying photos taken) with this command (see CAMERA_IMAGE_CAPTURED::sequence). Note; all CAMERA_IMAGE_CAPTURED corresponding with this command will share this sequence number</param>
```
hamishwillee: See comments as above: https://github.com/mavlink/mavlink/pull/1319/files#r393346249

message_definitions/v1.0/common.xml
```diff
@@ -2187,7 +2187,7 @@
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Re-request a CAMERA_IMAGE_CAPTURE message. Use NaN for reserved values.</description>
-        <param index="1" label="Number" minValue="0" increment="1">Sequence number for missing CAMERA_IMAGE_CAPTURE message</param>
+        <param index="1" label="Number" minValue="0" increment="1">Sequence number for missing CAMERA_IMAGE_CAPTURE messages</param>
```
hamishwillee: Suggest change label to "Sequence Number". 
Should the description also change to "Re-request all CAMERA_IMAGE_CAPTURE messages that have the specified sequence number."

message_definitions/v1.0/common.xml
```diff
@@ -5972,6 +5972,7 @@
       <field type="int32_t" name="image_index">Zero based index of this image (image count since armed -1)</field>
       <field type="int8_t" name="capture_result">Boolean indicating success (1) or failure (0) while capturing this image.</field>
       <field type="char[205]" name="file_url">URL of image taken. Either local storage or http://foo.jpg if camera provides an HTTP interface.</field>
+      <field type="int32_t" name="sequence">If non-zero, sequence number of the MAV_CMD_IMAGE_START_CAPTURE or MAV_CMD_DO_SET_CAM_TRIGG_DIST command this image is a response to.</field>
```
hamishwillee: Move "If non-zero" to the end. ie. "... response to. 0 if the sequence number is not known or supported.

hamishwillee:
@remster @gus A concern with this is that the sequence number might apply to a lot of images (ie if this capture runs for a while). That means you might be missing just one image, but you're re-requesting an abitrary number of them? Thoughts?

Either way, we should in `MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE` also support requesting an image by image_index.

message_definitions/v1.0/common.xml
remster: done: https://github.com/mavlink/mavlink/pull/1319/commits/f6f945cf39071667035e94aaee8888e1f8d7e9b4

message_definitions/v1.0/common.xml
remster: done: https://github.com/mavlink/mavlink/pull/1319/commits/f6f945cf39071667035e94aaee8888e1f8d7e9b4

message_definitions/v1.0/common.xml
remster: done: https://github.com/mavlink/mavlink/pull/1319/commits/ebe6906cf7d0fba659c921919592b67517447144

But note the `minValue` of `MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE.Sequence Number` is set to 0, which means in can be invoked against an invalid sequence number. We have two options:

- set `minValue` to 1
- stipulate that if `minValue` is 0 the client is re-requesting **all** CAMERA_IMAGE_CAPTURED events. That's not putting actual images on the wire, but is it going to be too much to ask for?

message_definitions/v1.0/common.xml
remster: done: https://github.com/mavlink/mavlink/pull/1319/commits/ebe6906cf7d0fba659c921919592b67517447144

remster:
> @remster @gus A concern with this is that the sequence number might apply to a lot of images (ie if this capture runs for a while). That means you might be missing just one image, but you're re-requesting an abitrary number of them? Thoughts?
> 
> Either way, we should in `MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE` also support requesting an image by image_index.

Indeed. This is similar to the comment i left [here](https://github.com/mavlink/mavlink/pull/1319#discussion_r393660480). I don't see why adding `image_index` to `MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE` might hurt, so I am happy to add this (subject to how this chat develops). 

We rarely capture more than 100 photos in a mission so the consequence would be to have, at most 100 `CAMERA_IMAGE_CAPTURED` appearing at quick succession and occupying the link. Looking at the `CAMERA_IMAGE_CAPTURED` composition it looks like it needs around 2Kb of living space, which would result in 200K being put on the link in that worst case. It doesn't seem a lot (compared to video feed or actual photo), but I have no other experience to judge whether that's too big a footprint.

message_definitions/v1.0/common.xml
hamishwillee: I presume that this number was previously the image number, which may or may not have been 0 indexed. It is quite likely you would want to request either by image_number or sequence number, or even some number within a sequence (that we have not yet worked out).

That needs @dogmaphobic to comment. Once that is done we'll know what to do with this. But IMO we don't really want to send all images again, particularly on zero value that systems might easily set without thinking.

hamishwillee:
@dogmaphobic This is just a reminder that we'd love your attention on this!

dogmaphobic:
Sorry, I've been quite busy and it's not easy to switch contexts.

As to changing MAV_CMD_DO_SET_CAM_TRIGG_DIST.param4 from empty to something else, I'm not sure how safe this would be or what difference it would make. Why is MAV_CMD_DO_SET_CAM_TRIGG_DIST.param4 being set to anything? The firmware will ignore this. It will not change the sequence number to something other than "zero based index from arm".

@DonLakeFlyer, I don't ever use this command. Only mission does it. Can you think of any impact this might have? Does this make sense to you? I don't quite get the rationale. The CAMERA_IMAGE_CAPTURED message uses a zero based index. Essentially, the number of image captures since the vehicle was armed. Nothing will change that.

This thread got quite long and convoluted. I may be missing something as I don't quite get what is being proposed here. If I got this right, you are running both a mission (using MAV_CMD_DO_SET_CAM_TRIGG_DIST) and taking arbitrary photos (using MAV_CMD_IMAGE_START/STOP_CAPTURE) and expect to have a way to determine which images correspond to one or the other? Indeed there is currently no way to differentiate them. If anything, the change would have to be done in the CAMERA_IMAGE_CAPTURE message indicating if that capture corresponds to a MAV_CMD_DO_SET_CAM_TRIGG_DIST or a MAV_CMD_IMAGE_START/STOP_CAPTURE command. This would require changes in both firmware and all existing cameras that implement this protocol.

There is a MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE command, which in theory would request a missing CAMERA_IMAGE_CAPTURE message but this was meant for MAV_CMD_IMAGE_START_CAPTURE commands and not for MAV_CMD_DO_SET_CAM_TRIGG_DIST.

Mixing both MAV_CMD_DO_SET_CAM_TRIGG_DIST and MAV_CMD_IMAGE_START_CAPTURE won't change the sequence number. It will continue to be the image count -1.

By the way, MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE is marked as WIP because it has never been used. Messages and commands start in a WIP state. It's only changed away from WIP once it's implemented and proved to work and/or satisfy a requirement.

DonLakeFlyer:
How does CAMERA_IMAGE_CAPTURED telemetry coming back the ground station help you determine something about a photo to download? It's just an incremented index since arm. What do you do with the value from there?

remster:
> How does CAMERA_IMAGE_CAPTURED telemetry coming back the ground station help you determine something about a photo to download? It's just an incremented index since arm. What do you do with the value from there?

I think it may have escaped that I am proposing that:
CAMERA_IMAGE_CAPTURED not only carries the increment index since arm, [but also sequence number](https://github.com/mavlink/mavlink/pull/1319/files#diff-ceacc1e2390186650146a45be8fcb691R5975) originally supplied to MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE.

remster:
> @DonLakeFlyer, I don't ever use this command. Only mission does it. Can you think of any impact this might have? Does this make sense to you? I don't quite get the rationale. The CAMERA_IMAGE_CAPTURED message uses a zero based index. Essentially, the number of image captures since the vehicle was armed. Nothing will change that.

Once again, in this PR i am proposing that CAMERA_IMAGE_CAPTURED [additionally carries the sequence number](https://github.com/mavlink/mavlink/pull/1319/files#diff-ceacc1e2390186650146a45be8fcb691R5975) originally supplied to MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE. 

> This thread got quite long and convoluted. I may be missing something as I don't quite get what is being proposed here. If I got this right, you are running both a mission (using MAV_CMD_DO_SET_CAM_TRIGG_DIST) and taking arbitrary photos (using MAV_CMD_IMAGE_START/STOP_CAPTURE) and expect to have a way to determine which images correspond to one or the other? Indeed there is currently no way to differentiate them. If anything, the change would have to be done in the CAMERA_IMAGE_CAPTURE message indicating if that capture corresponds to a MAV_CMD_DO_SET_CAM_TRIGG_DIST or a MAV_CMD_IMAGE_START/STOP_CAPTURE command. This would require changes in both firmware and all existing cameras that implement this protocol.

That is almost exactly right. Almost because I don't care whether a photo was taken in response to MAV_CMD_DO_SET_CAM_TRIGG_DIST or MAV_CMD_IMAGE_START_CAPTURE (these are just _take-a-photo_ commands and i understand the former is a bit of a backwards-compatibility liability). All I care about is to know - upon reception of CAMERA_IMAGE_CAPTURED - what **instance** of the previous _take-a-photo_ command is this CAMERA_IMAGE_CAPTURED a response to. The sequence number therefore becomes more of an asset identifier. The controller needs a photo and instructs the firmware: when this photo is done, please stamp this asset identifier on it so I know it's this one,

 
> There is a MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE command, which in theory would request a missing CAMERA_IMAGE_CAPTURE message but this was meant for MAV_CMD_IMAGE_START_CAPTURE commands and not for MAV_CMD_DO_SET_CAM_TRIGG_DIST.

MAV_CMD_DO_SET_CAM_TRIGG_DIST suffers from the same problem MAV_CMD_IMAGE_START_CAPTURE does: unable to work out what photos were taken during periods of disconnection. MAV_CMD_DO_SET_CAM_TRIGG_DIST is actively used by QGC. There is a problem that we have in the field. It's a very simple one and it is, i dare to say, essential to fix one day:
To be able to work out what photos did the drone take and for each photo - what _take-a-photo_ command does it relate to. 

> 
> Mixing both MAV_CMD_DO_SET_CAM_TRIGG_DIST and MAV_CMD_IMAGE_START_CAPTURE won't change the sequence number. It will continue to be the image count -1.

This i did not understand. I also do not understand the current docu for sequence number:
"Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1). Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted. Use 0 to ignore it." 

In this PR i am giving it an "asset identifier" semantics **asking** what semantics does it enjoy now and whether I am breaking something here. Can you clarify what is sequence number today on MAVLINK? The docu does not explain it well.

DonLakeFlyer:
> To be able to work out what photos did the drone take and for each photo - what _take-a-photo_ command does it relate to.

Ok. Understand now. In order to do this to create some sort of grouping identifier I don't think modifying the various camera trigger commands to include a new grouping identifier is the way to go. This would create additional complexity in things like pattern planning which is already overly complex. I would suggest adding a completely new command. Call it whatever you want. That command would set a group id for all the subsequent images taken. Put them where you want in the mission if you need them. And then for most folks who are not going to need something like this they don't have to see it or deal with it at all.

remster:
Thanks @DonLakeFlyer - this is clear and can work, but I am concerned.
- I can't see clear purpose in the [sequence number](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2189) of the MAV_CMD_IMAGE_START_CAPTURE. @hamishwillee mentioned i think he didn't understand it either. Assuming it has no purpose, my PR gives it one and is thus a chance to remove cruft. If it does have a purpose I have to ask you to explain it here. I can edit the sequence number description to reflect this explanation.
- Adding additional message as you suggest will just remove my changes to MAV_CMD_IMAGE_START_CAPTURE and MAV_CMD_DO_SET_CAM_TRIGG_DIST, but will keep the changes related to MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE and CAMERA_IMAGE_CAPTURE. I hope this is clear.
- Adding additional message will put burden on drone vendors so i'd like to double-check that your suggestion is as close to MAVLINK community consensus as possible to solicit.

remster:
closed as continued with https://github.com/mavlink/mavlink/pull/1500

PR #1321, opened 2020-03-13
Trajectory message fix and remove wip

hamishwillee:
This fixes some terminology in the trajectory interface and removes WIP (as these are now in production use).

@jkflying Can you please carefully review these messages to confirm that they do actually reflect our expectation.

FYI @auturgy @julianoes


hamishwillee:
Thanks!

PR #1322, opened 2020-03-17
Add ESC_INFO and ESC_STATUS mavlink messages

RicardoM17:
This PR adds ESC_INFO and ESC_STATUS messages for slow changing and fast changing ESC information respectively. An ESC_CONNECTION_TYPE  and an ESC_FAILURE_FLAGS enum were also added for this purpose. Each message can house data for up to 4 ESC and by using the index field each message can be used repeatedly up to a maximum of 64 ESC.

Signed-off-by: Ricardo Marques <marques.ricardo17@gmail.com>

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1226,6 +1226,56 @@
         <description>The node is no longer available online.</description>
       </entry>
     </enum>
+    <enum name="ESC_CONNECTION_TYPE">
+      <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Indicates the ESC connection type.</description>
+      <entry value="0" name="ESC_CONNECTION_TYPE_PPM">
+        <description>Traditional PPM ESC.</description>
+      </entry>
+      <entry value="1" name="ESC_CONNECTION_TYPE_SERIAL">
+        <description>Serial Bus connected ESC.</description>
+      </entry>
+      <entry value="2" name="ESC_CONNECTION_TYPE_ONESHOT">
+        <description>One Shot PPM ESC.</description>
+      </entry>
+      <entry value="3" name="ESC_CONNECTION_TYPE_I2C">
+        <description>I2C ESC.</description>
+      </entry>
+      <entry value="4" name="ESC_CONNECTION_TYPE_CAN">
+        <description>CAN-Bus ESC.</description>
+      </entry>
+      <entry value="5" name="ESC_CONNECTION_TYPE_DSHOT">
+        <description>DShot ESC.</description>
+      </entry>
+    </enum>
+    <enum name="ESC_FAILURE_FLAGS" bitmask="true">
+      <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Flags to report ESC failures.</description>
+      <entry value="0" name="ESC_FAILURE_NONE">
+        <description>No ESC failure.</description>
+      </entry>
+      <entry value="1" name="ESC_FAILURE_OVER_CURRENT">
+        <description>Over current failure.</description>
+      </entry>
+      <entry value="2" name="ESC_FAILURE_OVER_VOLTAGE">
+        <description>Over voltage failure.</description>
+      </entry>
+      <entry value="4" name="ESC_FAILURE_OVER_TEMPERATURE">
+        <description>Over temperature failure.</description>
+      </entry>
+      <entry value="8" name="ESC_FAILURE_OVER_RPM">
+        <description>Over RPM failure.</description>
+      </entry>
+      <entry value="16" name="ESC_FAILURE_INCONSISTENT_CMD">
+        <description>Inconsistent command failure i.e. out of bounds.</description>
+      </entry>
+      <entry value="32" name="ESC_FAILURE_MOTOR_STUCK">
+        <description>Motor stuck failure.</description>
+      </entry>
+      <entry value="64" name="ESC_FAILURE_GENERIC">
+        <description>Generic ESC failure.</description>
+      </entry>
+    </enum>
     <enum name="STORAGE_STATUS">
       <description>Flags to indicate the status of camera storage.</description>
       <entry value="0" name="STORAGE_STATUS_EMPTY">
@@ -6520,6 +6570,30 @@
       <field type="float" name="pan" units="rad">Pan/yaw angle (positive: to the right, negative: to the left, NaN to be ignored).</field>
       <field type="float" name="tilt_rate" units="rad/s">Tilt/pitch angular rate (positive: up, negative: down, NaN to be ignored).</field>
       <field type="float" name="pan_rate" units="rad/s">Pan/yaw angular rate (positive: to the right, negative: to the left, NaN to be ignored).</field>
+    </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information for lower rate streaming. Recommended streaming rate 1Hz. See ESC_STATUS for higher-rate ESC data.</description>
+      <field type="uint8_t" name="index">Index of the first ESC in this message. minValue = 0, maxValue = 60, increment = 4.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Counter of data packets received.</field>
+      <field type="uint8_t" name="count">Total number of ESCs in all messages of this type. Message fields with an index higher than this should be ignored because they contain invalid data.</field>
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint16_t[4]" name="failure_flags" enum="ESC_FAILURE_FLAGS" display="bitmask">Bitmap of ESC failure flags.</field>
+      <field type="uint32_t[4]" name="error_count">Number of reported errors by each ESC since boot.</field>
+      <field type="uint8_t[4]" name="temperature" units="degC">Temperature measured by each ESC. UINT8_MAX if data not supplied by ESC.</field>
+    </message>
+    <message id="291" name="ESC_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information for higher rate streaming. Recommended streaming rate is ~10 Hz. Information that changes more slowly is sent in ESC_INFO. It should typically only be streamed on high-bandwidth links (i.e. to a companion computer).</description>
+      <field type="uint8_t" name="index">Index of the first ESC in this message. minValue = 0, maxValue = 60, increment = 4.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="int32_t[4]" name="rpm" units="rpm">Reported motor RPM from each ESC (negative for reverse rotation).</field>
+      <field type="float[4]" name="voltage" units="V">Voltage measured from each ESC.</field>
+      <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
     <message id="299" name="WIFI_CONFIG_AP">
       <description>Configure WiFi AP SSID, password, and mode. This message is re-emitted as an acknowledgement by the AP. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE</description>


dagar:
This is quite a large message. Perhaps consider breaking it up to send each ESC status separately + an id field? See BATTERY_STATUS (https://mavlink.io/en/messages/common.html#BATTERY_STATUS).

message_definitions/v1.0/common.xml
```diff
@@ -5765,6 +5765,22 @@
       <field type="int8_t" name="custom1">Field for custom payload.</field>
       <field type="int8_t" name="custom2">Field for custom payload.</field>
     </message>
+    <message id="240" name="ESC_STATUS">
+      <description>ESC status feedback</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
+      <field type="uint8_t" name="esc_count" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint8_t" name="esc_connection_type" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
```
hamishwillee: Description does not match name

hamishwillee:
I agree with https://github.com/mavlink/mavlink/pull/1322#issuecomment-600369284

Another approach you might use in parallel is to split into static and dynamic messages. The static being for things that don't change between reboots - e.g. addresses, while dynamic is for things that change regularly.

What sort of rate would this be sent at?

RicardoM17:
> This is quite a large message. Perhaps consider breaking it up to send each ESC status separately + an id field? See BATTERY_STATUS (https://mavlink.io/en/messages/common.html#BATTERY_STATUS).

Hmm so if I understand correctly the BATTERY_STATUS message will get published several times for each "loop", once for each battery with the ID at the top changing?

> I agree with [#1322 (comment)](https://github.com/mavlink/mavlink/pull/1322#issuecomment-600369284)
> 
> Another approach you might use in parallel is to split into static and dynamic messages. The static being for things that don't change between reboots - e.g. addresses, while dynamic is for things that change regularly.
> 
> What sort of rate would this be sent at?

Also another interesting approach. Is there an example that I can follow? I'm assuming you mean something similar to the SMART_BATTERY_STATUS and SMART_BATTERY_INFO messages but unfortunately those don't have the PX4 implementation yet, that I could check on.

The rate would need to be quite high to have an actual use, my ultimate goal is to have this also being parsed in MAVROS and then being used in offboard (which I've already accomplished but haven't made the PR as I'm hoping to finalize the message structure first in these 2 PRs). That being said I don't know how high we can go until it's too much so, I'm actively looking for suggestions here. For the reference the other PR can be seen here https://github.com/PX4/Firmware/pull/14408 and the value I used was 50Hz. I'm not really sure if there's an added benefit in going higher than this.

@hamishwillee @dagar should I try another approach in the message field or do you guys want to first agree on the structure and then I make the changes??

Let me know.

message_definitions/v1.0/common.xml
RicardoM17: Noted. Thanks for the comment.

I'll fix alongside the changes that are being discussed in the main comment thread.

hamishwillee:
We try to reduce the total content sent to the minimum practical amount:
- Daniel's suggestion is to send a separate message for each ESC 
  - because many systems will only have 4 ESC and the current message supports 8, resulting in a lot of empty data.
  - because then you could choose perhaps to send only when you have an esc change.
  - Generally this makes sense unless it was the case (AFAIK it is not) that ESCs are dependent on each other and separate messages would have to be aggregated by a time-synced GSC before they could be used. I.e. if you would be creating heaps of work for the recipient.
- Yes my suggestion is the approach in [SMART_BATTERY_STATUS](https://mavlink.io/en/messages/common.html#SMART_BATTERY_STATUS) (dynamic) and [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO) (static).
  - same justification - you send the static message once, and the higher rate message can be much smaller. 
- I asked about rates, not because I know what they should be, but because I think you should hint at this in the spec if you know what will be useful. I.e. if the RPM (say) is intended just for reporting then you'd have maybe at 1Hz, while if it is for controlling something in another system you'd have possibly much higher. Depending on the intention you will design the message differently - perhaps splitting out into even more messages. Hope that makes sense.
- Do you need this to work with MAVLink 1? If not, the message ID should not be in the low range as used. https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges

Personally I would use both approaches. So the static message can cover all ESCs, but single message for individual ESC.
If you think it makes sense, update PR appropriately and we can then proceed the discussion. 

 (note, below is example with notes put in it for discussion. Not checked properly so you can't just copy into the pr - it is a hint). I think the broad structure makes sense though.

```
    <message id="XXXX" name="ESC_INFO">
      <description>ESC information. See ESC_STATUS for dynamic ESC information.</description>
      <field type="uint8_t" name="esc_count" units="us">Number of ESC.</field>
      <field type="uint8_t" name="esc_connection_type" units="us">What is this? Probably and enum?.</field>
      <field type="uint8_t[8]" name="esc_address">Address of each ESC.</field>

ALSO here you might have RANGES - e.g. max/min voltage, rpm or whatever for the esc or other values below. Anything else static that might be useful for setting up the receiving system. E.g if it is a GCS it will like ranges if it is using a bar display vs a numeric display.
    </message>
```
```
    <message id="XXXX" name="ESC_STATUS">
      <description>ESC status feedback</description>
      <field type="uint8_t" name="esc_index" >The number of this ESC.</field>
      <field type="uint16_t" name="counter">Data counter.</field>
      <field type="uint8_t" name="esc_info" display="bitmask">THIS would become status for current ESC.</field>
      <field type="uint64_t" name="esc_time_usec" units="us">Dunno what this is.</field>
      <field type="uint32_t" name="esc_error_count">Number of reported errors by THIS ESC.</field>
      <field type="int32_t" name="esc_rpm" units="rpm">Reported motor RPM from THIS ESC, negative for reverse rotation.</field>
      <field type="float" name="esc_voltage" units="V">Voltage measured from THIS ESC. NOTE, check that float and units is large enough value AND great enough resolution for both current ESC and the biggest/smallest ESC you can think of. Ditto for other fields</field>
      <field type="float" name="esc_current" units="A">Current measured from THIS ESC.</field>
      <field type="uint8_t" name="esc_temperature" units="degC">Temperature measured from each ESC.</field>
      <field type="uint8_t" name="esc_state">State of THIS ESC.</field>
    </message>
```

RicardoM17:
> We try to reduce the total content sent to the minimum practical amount:
> 
> * Daniel's suggestion is to send a separate message for each ESC
>   
>   * because many systems will only have 4 ESC and the current message supports 8, resulting in a lot of empty data.
>   * because then you could choose perhaps to send only when you have an esc change.
>   * Generally this makes sense unless it was the case (AFAIK it is not) that ESCs are dependent on each other and separate messages would have to be aggregated by a time-synced GSC before they could be used. I.e. if you would be creating heaps of work for the recipient.
> * Yes my suggestion is the approach in [SMART_BATTERY_STATUS](https://mavlink.io/en/messages/common.html#SMART_BATTERY_STATUS) (dynamic) and [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO) (static).
>   
>   * same justification - you send the static message once, and the higher rate message can be much smaller.

Understood. Lots of good info thanks. I will try to make the changes for this between today and tomorrow if I find the time but it does make sense. I believe there are some issues in PX4 side with the ESC detection so this creates some additional work there (which probably would still have to be done) by I agree that it's better long term and from a MAVLink POV it does make much more sense.

> * I asked about rates, not because I know what they should be, but because I think you should hint at this in the spec if you know what will be useful. I.e. if the RPM (say) is intended just for reporting then you'd have maybe at 1Hz, while if it is for controlling something in another system you'd have possibly much higher. Depending on the intention you will design the message differently - perhaps splitting out into even more messages. Hope that makes sense.

I can see it being useful for most generic applications as a reporting at low frequency like the 1Hz that you suggested, however my own use case is to incorporate this with MAVROS so that it can be used in Offboard node with PX4. In particular the RPM, Voltage and Current data is the most important info and needs to be streamed at somewhat high frequency, I'd say 50Hz is a good value but I don't know how much overhead this causes for people who don't need this. This is to be applied for onboard regeneration with VESC. Just providing some context.

> * Do you need this to work with MAVLink 1? If not, the message ID should not be in the low range as used. https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges

I'd also take some feedback here from more experienced people. From what I've read, which unfortunately wasn't much to really make an informed decision, it seems that MAVLink 1 is being used the most? I'm not really sure if using MAVLink 2 would break anything from my side (PX4, MAVROS) but if not, and if you recommend it, I'll gladly change it. Ive been following the docs in https://dev.px4.io/master/en/ros/mavros_custom_messages.html so it looks like I've been using 2.0 from the start so it looks like the change to MAVLink 2 is the best here? Correct me if I'm wrong but it's only an ID change in this case? and if there's a specific one that I should use let me know, otherwise I will find the best one at my discretion.

> Personally I would use both approaches. So the static message can cover all ESCs, but single message for individual ESC.
> If you think it makes sense, update PR appropriately and we can then proceed the discussion.
> 
> (note, below is example with notes put in it for discussion. Not checked properly so you can't just copy into the pr - it is a hint). I think the broad structure makes sense though.
> 
> ```
>     <message id="XXXX" name="ESC_INFO">
>       <description>ESC information. See ESC_STATUS for dynamic ESC information.</description>
>       <field type="uint8_t" name="esc_count" units="us">Number of ESC.</field>
>       <field type="uint8_t" name="esc_connection_type" units="us">What is this? Probably and enum?.</field>
>       <field type="uint8_t[8]" name="esc_address">Address of each ESC.</field>
> 
> ALSO here you might have RANGES - e.g. max/min voltage, rpm or whatever for the esc or other values below. Anything else static that might be useful for setting up the receiving system. E.g if it is a GCS it will like ranges if it is using a bar display vs a numeric display.
>     </message>
> ```
> 
> ```
>     <message id="XXXX" name="ESC_STATUS">
>       <description>ESC status feedback</description>
>       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
>       <field type="uint8_t" name="esc_index" >The number of this ESC.</field>
>       <field type="uint16_t" name="counter">Data counter.</field>
>       <field type="uint8_t" name="esc_info" display="bitmask">THIS would become status for current ESC.</field>
>       <field type="uint64_t" name="esc_time_usec" units="us">Dunno what this is.</field>
>       <field type="uint32_t" name="esc_error_count">Number of reported errors by THIS ESC.</field>
>       <field type="int32_t" name="esc_rpm" units="rpm">Reported motor RPM from THIS ESC, negative for reverse rotation.</field>
>       <field type="float" name="esc_voltage" units="V">Voltage measured from THIS ESC. NOTE, check that float and units is large enough value AND great enough resolution for both current ESC and the biggest/smallest ESC you can think of. Ditto for other fields</field>
>       <field type="float" name="esc_current" units="A">Current measured from THIS ESC.</field>
>       <field type="uint8_t" name="esc_temperature" units="degC">Temperature measured from each ESC.</field>
>       <field type="uint8_t" name="esc_state">State of THIS ESC.</field>
>     </message>
> ```

Many thanks. Yes I understand it's not a direct copy-paste but this helps quite a bit.

Thanks once again @hamishwillee for all the help and patience. Much to learn on this side :sweat_smile:

RicardoM17:
Hey again @hamishwillee ,

I put some more thought into this and also took into account your many suggestions. I split the message into two as to me it also makes perfect sense. As such I currently have one message that has all the static information + other information that doesn't need to be streamed at such high right like temperature, and another message with the information that does require high frequency streaming, in this case timestamps, rpm, voltage and current.

I did also look into publishing each ESC data separately (talking about the high frequency info like RPM) and while from the publishing side this doesn't pose much of an issue, I believe this complicates things further down the "stream" aka in MAVROS/offboard control. There either we would have to wait for all messages from all the ESCs and thus introducing unwanted latency and  losing all advantages of having a high frequency message streaming, or we would parse each message individually which also doesn't make too much sense. All data would be out of sync and as such less meaningful.

Below is the draft of my current proposal. Once I get some feedback on the mavlink1/2/ID issue I will push a more prepared version.

```
    <message id="XXX" name="ESC_INFO">
      <description>ESC information. See ESC_STATUS for dynamic ESC information.</description>
      <field type="uint16_t" name="counter">Data counter.</field>
      <field type="uint8_t" name="esc_count" units="us">Number of ESC.</field>
      <field type="uint8_t" name="esc_connection_type" units="us">Connection type protocol for all ESC. (0: PPM, 1: Serial, 2: Oneshot, 3: I2C, 4: CAN, 5: DShot)</field>
      <field type="uint8_t" name="esc_info" display="bitmask">Information regarding online/offline status of each ESC.</field>
      <field type="uint8_t[8]" name="esc_address">Address of each ESC.</field>
      <field type="uint32_t[8]" name="esc_error_count">Number of reported errors by each ESC.</field>
      <field type="uint8_t[8]" name="esc_temperature" units="degC">Temperature measured by each ESC.</field>
    </message>
      <message id="XXX" name="ESC_STATUS">
      <description>ESC status dynamic feedback.</description>
      <field type="uint64_t[8]" name="esc_time_usec" units="us">Timestamp for each ESC (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
      <field type="int32_t[8]" name="esc_rpm" units="rpm">Reported motor RPM from each ESC, negative for reverse rotation.</field>
      <field type="float[8]" name="esc_voltage" units="V">Voltage measured from each ESC.</field>
      <field type="float[8]" name="esc_current" units="A">Current measured from each ESC.</field>
    </message>
```

Let me know what you think

hamishwillee:
@TSC21 Can you look at the message set above: https://github.com/mavlink/mavlink/pull/1322#issuecomment-601315363

The `ESC_STATUS` is intended to be a high-rate message (50Hz?) for sending info for up to 8 ESC. It is pretty wasteful if you're sending 4 ESC so we proposed sending ESC individually with an id and the recipient (ROS) composing.
1. Can you confirm whether or not composing the messages is reasonable or not?
2. Is the proposed rate sensible/possible?
3. Do you think it reasonable to consider splitting out just this field into its own message?

We're still in design phase of the message, so anything is possible. I'm trying to understand the limits. More context from the link above.

[ @RicardoM17 Please target MAVLink 2 and use MAVLink 2 ids - if you don't know you need MAVLink 1, then you don't. I'll review the rest after Nuno responds ]

amilcarlucas:
Ardupilot (ardupilotmega.xml) already contains a good ESC message, maybe use that as inspiration?

hamishwillee:
That's a good idea too https://github.com/mavlink/mavlink/pull/1322#issuecomment-603209377

IF they have significant overlap and/or meet the same needs we might even move it to common. @RicardoM17 can you do first evaluation?

RicardoM17:
Thanks for the suggestion @amilcarlucas I wasn't aware of this message.

@hamishwillee Having had a look here is my opinion. Their approach differs from the one here suggested as they publish 3 equal messages for ESCs 1-4, 5-8 and 9-12. There is an extra field in the total current which I'm not sure if it can be supported on the PX4 side. Also from the description it looks to be a specific message for BLHeli ESCs only.

IMO I feel like my current suggestion ( the one in https://github.com/mavlink/mavlink/pull/1322#issuecomment-601315363 , not the one in the PR) is a better approach as we have more data, while at the same time being able to stream the relevant data at a faster rate because of the split in the message into INFO and STATUS.

However if you feel like it would make things easier I can split the ESC_STATUS message into two (One for ESCs 1-4 and the other for 5-8), however with this approach unless we have some way of only enabling the second message in certain scenarios, there's really no benefit and this adds complexity to the MAVROS side. Currently for me 4 motors work but in a few months I will require 6 (just leaving some context on my particular use case. I know that all of this is public and should be generic). 

I leave below a part of one of the three ESC messages present in ardupilotmega.xml for easier referencing. 

```
    <message id="11031" name="ESC_TELEMETRY_5_TO_8">
      <description>ESC Telemetry Data for ESCs 5 to 8, matching data sent by BLHeli ESCs.</description>
      <field type="uint8_t[4]" name="temperature" units="degC">Temperature.</field>
      <field type="uint16_t[4]" name="voltage" units="cV">Voltage.</field>
      <field type="uint16_t[4]" name="current" units="cA">Current.</field>
      <field type="uint16_t[4]" name="totalcurrent" units="mAh">Total current.</field>
      <field type="uint16_t[4]" name="rpm" units="rpm">RPM (eRPM).</field>
      <field type="uint16_t[4]" name="count">count of telemetry packets received (wraps at 65535).</field>
    </message>
```

I'd like to hear @TSC21 's take on my current approach in the comment above as to me that looks the best approach but there's many possible implications that I might not be aware of.

hamishwillee:
Thanks very much @RicardoM17 

FWIW, what is good about the ardupilot approach is that it is:
- efficient: There are more Quads and VTOL quads than any other type - so in the most common case you have just one message that is fully packed. If you have an octo quad again you're fully packed. It is a bit less efficient with a hexaquad, but still, you have nowhere near the overhead of a single large message. 
- scalable - if someone comes along with a bigger quad, then they just add another message.
- note also how they have specified that counter might roll over.

Contrasting with `ESC_STATUS` as you defined it - their message will be more efficient on anything other than a hexa, and can easily scale to larger vehicles. Contrasting to the ESC_STATUS with an esc index their approach is probably more efficient due to packet overheads. I still lean towards this unless we need the ESCs delivered in sets.

W.r.t. the total current, we just need to indicate mechanism(s) for specifying that a particular field is not supported on a platform. For ints usually that is max value. 

W.r.t comment about BLHELI I'm sure that can be generalised.

I'll chase up @TSC21

TSC21:
@hamishwillee @RicardoM17 if the message sole purpose is to only provide status, and not to be used in control routines, I suggest a message contains all the fields stated plus an `esc_id` field that identifies where is the data coming from. On the ROS side (or whatever other recipient  and/or parser of this message), you just have to identify what is the source of the message and handle the data in a way you can split the streams for each of the ESC's. A good example of how to do that is the [`distance_sensor` plugin](https://github.com/mavlink/mavros/blob/master/mavros_extras/src/plugins/distance_sensor.cpp) in MAVROS, where you setup the expected streams and set a publisher per stream.

If this was to be used in control routines, it would be preferable to just send the data of all ESCs in the same message, since that way you would guarantee synchronisation.

RicardoM17:
@TSC21 @hamishwillee thanks for the feedback.

As I've stated above synchronization is important to our use case hence my suggestion (and need/desire) of keeping the same data from the multiple ESCs in the same message. To reduce the overhead of this message all the data that isn't really relevant for control routines was moved into the `ESC_INFO` message so that it can be streamed at much lower frequency.

Splitting the "high frequency message" into 2 does make it "more efficient" for quadcopters but for Hexa+Octocopters then we have the synchronization issues by having split the two messages, which makes it very sub-optimal for any type of analysis there.

I already have this working on my side in the 3 parts (PX4, MAVLink and MAVROS) so if needed I can always continue with my fork/version, which obviously fits our use case best. I'm just trying to bring this upstream. However it's likely that more development will be done in these areas which I would also like to also bring upstream for many reasons. That being said if it starts to differ much, like the splitting of the data between ESCs, it will hinder our progress and in that case it will make more sense to use our custom version, which will make it harder to share these changes. I also understand the concerns you are raising though which is why I'm trying to reach a consensus here :sweat_smile: .

If it helps, reducing the frequency of the high frequency, into something like 25Hz or 10Hz, is something that we could look into or at the least keep as a diff between our version and upstream as it's just a value change instead of a whole architecture/structure difference. By doing both versions will basically be the same.

Also I don't believe that this high-frequency stream should/will be enabled by default in PX4, but instead it only makes sense in very specific scenarios. First it needs UAVCAN or any type of motor feeback and secondly it may make sense to also only have this with offboard control. So the vast majority of the users won't be affected by this unless they want it and in that case it will probably be important for them as well to have synchronization.

cmic0:
Hey guys!
I have started doing some work about handling ESC failures in PX4, so might be worth to also monitor this PR https://github.com/PX4/Firmware/pull/14494 so that the MAVLink message can be adapted.

In my scenario failures are handled at Firmware level, the MAVLink message would be intended for notify the user only.

message_definitions/v1.0/common.xml
amilcarlucas: having an "esc_" prefix on every name attribute is redundant.

amilcarlucas: This message is huge, it will use a lot of mavlink datalink capacity. Furthermore it will not scale past 8 instances

RicardoM17: Hi @amilcarlucas . Thanks for the review. You are right the message is quite big. Currently there has been quite a lot of discussions back and forth on how to improve it. The latest "iteration" is in this comment https://github.com/mavlink/mavlink/pull/1322#issuecomment-601315363

I apologize for not having those changes pushed but because even the structure of the message is still quite far from consensus, to me it didn't make sense to do it. Also I was hoping a conclusion would be reached quite a bit faster but seems I've overestimated some things.

I'd take your input on that message however and on the other points that have been discussed in the other comments since then. To me that seems the only approach that would make sense for my use case and future ones that have been discussed already where synchronization is required.

P.s. The `esc_` prefix comment is still valid and if others agree I am OK with making that change as well. I was just going off the existing uORB message and topic present in PX4.

hamishwillee:
This got quite complicated as discussion goes. I think we should try agree a broad way forward next dev call. 

@RicardoM17 We would prefer to use and extend existing messages where at all possible. Often compatibility is more important than efficiency.
So one alternative you might want to propose is what a message set that uses  the ArduPilot messages might look like.

RicardoM17:
Apologies for the (long) delay. A big milestone deadline came up and time just vanished.

As you may or may not recall this PR was discussed in a previous MAVLINK call, I gathered all the feedback and here is the updated version up for discussion and improvements.

Main points are:

The message is again split into a more "static" message  `ESC_INFO`  and now two dynamic messages `ESC_STATUS_1_TO_4` and `ESC_STATUS_5_TO_8`. I can also extend support for a third message if necessary?  but to me this makes sense that in that case ESC_INFO the addresses, temperature and error_count fields would also go up to 12 in size.

Change to MAVLINK 2 ID as per @hamishwillee suggestions.

Removal of the `esc_` prefix based on @amilcarlucas feedback.

I checked and it seems that the timestamps are the same for all ESCs so I also reduced the timestamp on the dynamic messages to size 1 while also removing the timestamp in `ESC_INFO` (I had this written in my notes but tbh it doesn't make much sense to me, especially as it's just 1 field and this message is already low frequency by design).

Lastly I thought this through and I believe that perhaps 50Hz was too high and not necessarily needed for my purposes so perhaps 10Hz should suffice. Also it seems that for example in PX4 the ESC data only gets pooled at this rate as well. I still need to confirm this however. Perhaps with this it is once again feasible to only have 1 dynamic message for all 8 ESC? just a thought.

I would much appreciate further reviews and feedback on this.

Many thanks

message_definitions/v1.0/common.xml
```diff
@@ -6315,6 +6315,37 @@
...
       <field type="uint32_t" name="v_estimated_delay_us" units="us">Estimated delay of the speed data.</field>
       <field type="float" name="feed_forward_angular_velocity_z" units="rad/s">Feed forward Z component of angular velocity, positive is yawing to the right, NaN to be ignored. This is to indicate if the autopilot is actively yawing.</field>
+    </message>
+      <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
+      <field type="uint8_t" name="count" units="us">Number of ESC.</field>
```
amilcarlucas, line 6325: The units are wrong here

message_definitions/v1.0/common.xml
```diff
@@ -6315,6 +6315,37 @@
...
+      <field type="uint8_t" name="connection_type" units="us">Connection type protocol for all ESC. (0: PPM, 1: Serial, 2: Oneshot, 3: I2C, 4: CAN, 5: DShot)</field>
```
amilcarlucas, line 6326: The units are wrong here

message_definitions/v1.0/common.xml
```diff
@@ -6315,6 +6315,37 @@
...
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[8]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[8]" name="error_count">Number of reported errors by each ESC.</field>
+      <field type="uint8_t[8]" name="temperature" units="degC">Temperature measured by each ESC.</field>
```
amilcarlucas, line 6330: This starts to be *VERY* redundant with https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/ardupilotmega.xml#L1632

Please consider using those messages instead, and just adding one static message with the information that is not already present in those messages.

message_definitions/v1.0/common.xml
RicardoM17: Correct. I will fix. Thanks.

message_definitions/v1.0/common.xml
RicardoM17: Correct. I will fix. Thanks.

message_definitions/v1.0/common.xml
RicardoM17: I'll have to disagree here. First one of the main reasons I split this message in more than one was because of bandwidth concerns raised by other members. As such the dynamic message I'm proposing is as lean as possible, while it doesn't make sense for me to be streaming at high-ish frequencies data like temperature, totalcurrent, and count. In fact I would like to go more in the opposite way and have all 8 ESCS in one dynamic message but this is a compromise that I'm making based on suggestions. But if this is not a concern at 10Hz then I can do all 8 ESC in one dynamic message which to me is preferable.

I understand your concerns with redundancies but I also can't really use the ardupilot messages as for example the both the current and the rpm fields are unsigned meaning we get no info negative currents or rpms so we can't have info on regenerative braking or can't really use motors that are able to spin in two directions. The latter is a good thing to have for the sake of future compatibility while the former is absolutely essential to my use case (and for others in the future hence why I'm trying to get this in open source). I'm also not 100% sure on the possibility of using ardupilot messages on flightstacks like PX4 but given the reasons above turns this question irrelevant perhaps.

message_definitions/v1.0/common.xml
```diff
@@ -6342,6 +6342,37 @@
...
       <field type="float" name="feed_forward_angular_velocity_z" units="rad/s">Feed forward Z component of angular velocity, positive is yawing to the right, NaN to be ignored. This is to indicate if the autopilot is actively yawing.</field>
+    </message>
+      <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
+      <field type="uint8_t" name="count">Number of ESC.</field>
+      <field type="uint8_t" name="connection_type">Connection type protocol for all ESC. (0: PPM, 1: Serial, 2: Oneshot, 3: I2C, 4: CAN, 5: DShot)</field>
```
amilcarlucas, line 6353: Would be better if you create an enum for this, or even better, reuse an existing one if it fits.

RicardoM17: Sounds good @amilcarlucas . I will have a look around to see if I can find that I can use or something similar to check the good implementation. If you have any suggestions/links for that let me know.


Thanks for the review once again.

RicardoM17: @amilcarlucas like so?

amilcarlucas:
I would prefer if you scale this to 12 ESCs on the static message and added a third dynamic message.

RicardoM17:
@amilcarlucas Done. Let me know if there are any other changes that you require. I can also remove the [WIP] from the title if that is no longer the case.

amilcarlucas:
Can you please squash all commits into a single one, and improve the git commit message?

RicardoM17:
@amilcarlucas thanks for the feedback. I've also edited the PR description, but if you want I can ammend and add that to the commit description message as well.

amilcarlucas:
Please add this text to the commit message:
`This PR adds the ESC_INFO, ESC_STATUS_1_TO_4, ESC_STATUS_5_TO_8 and ESC_STATUS_9_TO_12 mavlink messages. An enum ESC_CONNECTION_TYPE was also created.

The first message ESC_INFO contains the more static info of all, and up to 12, ESC. On the other hand the more dynamic data is sent via 3 separate ESC_STATUS_X_TO_Y messages in groups of 4, once again up to 12 ESC.`

amilcarlucas:
@hamishwillee I still think that the ardupilot messges are very good for some situations. So I would like to merge this, but also move the ESC messages from ardupilotmega.xml to common.xml.

What do you think?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
...
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
+      <field type="uint8_t" name="count">Number of ESC.</field>
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[12]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[12]" name="error_count">Number of reported errors by each ESC.</field>
```
hamishwillee, line 6441: errors *since boot*?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
...
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
+      <field type="uint8_t" name="count">Number of ESC.</field>
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[12]" name="address">Address of each ESC.</field>
```
hamishwillee, line 6440: What is the address of and ESC? Is this UAVCAN address or something else?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
...
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
+      <field type="uint8_t" name="count">Number of ESC.</field>
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
```
hamishwillee, line 6438: How is this this information expected to be used? Ie what could you do with this?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Data counter.</field>
```
hamishwillee, line 6436: What is this counting?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
...
+      <field type="uint8_t" name="count">Number of ESC.</field>
```
hamishwillee, line 6437: Note max value of 12. This whole message will be broken if we end up with a 13 ESC system.

hamishwillee: Is the assumption that this value is used to determine what other values are valid? E.g. if this is 6 then we know we can ignore the `voltage` field for array values 7, 8 in ESC_STATUS_5_TO_8?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
...
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[12]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[12]" name="error_count">Number of reported errors by each ESC.</field>
+      <field type="uint8_t[12]" name="temperature" units="degC">Temperature measured by each ESC.</field>
+    </message>
+    <message id="291" name="ESC_STATUS_1_TO_4">
```
hamishwillee, line 6444: Why not make this ESC_STATUS and have index to indicate the value of the first batch of esc?  So you would send this out with the index value of 0, 4, 8 etc?

Then you'd only need one message, and you could have as high a count of these as you want without having to screw up the first message.

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC information. See ESC_STATUS for dynamic ESC information. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
```
hamishwillee, line 6434: Approximate rate that this should be sent at?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,46 @@
...
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[12]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[12]" name="error_count">Number of reported errors by each ESC.</field>
+      <field type="uint8_t[12]" name="temperature" units="degC">Temperature measured by each ESC.</field>
+    </message>
+    <message id="291" name="ESC_STATUS_1_TO_4">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC status dynamic feedback.</description>
```
hamishwillee, line 6447: Maybe ESC dynamic (high rate) information. Should be sent at approximately XXXX (? for you) Hz. Static information is sent in ESC_INFO.

hamishwillee:
> I still think that the ardupilot messges are very good for some situations. So I would like to merge this, but also move the ESC messages from ardupilotmega.xml to common.xml.

@amilcarlucas I'm not at all opposed to having a discussion on that, but I see it as a separate issue. The questions I'd want answered are 
- when we would use them/ what are the cases we would use these instead of the new ones
- what the value of the count and total current is.

The same kinds of questions should be asked of these messages. I hope I have. 

@RicardoM17 Sorry for the silence - I have been on other things and lost track of this. Apologies if I'm re-asking questions. If possible we should try to make it so that we have a good generalist message that could replace the ardupilot message; so we do need to understand what it offers that these do not.

hamishwillee:
@bkueng This PR adds messages for ESC information. There is a message for low rate data (largely static info) and a number of messages for higher rate data (e.g. voltage, current, rpm). 

Could you please have a look and see if there is any information that you consider missing/unnecessary in both messages, and whether they are in the right category?

message_definitions/v1.0/common.xml
RicardoM17: That sounds like a good idea. I'll implement it today.

RicardoM17:
Hey @bkueng as per @hamishwillee request I'm pinging you here. It would be good if you could have a look to see if these messages have all the required fields and configuration to work with UAVCAN messages. If I understood correctly there is are new messages coming?.

@amilcarlucas what do you think of this?

https://github.com/mavlink/mavlink/pull/1322/files#r444631673

This was quickly discussed in today's dev call with @hamishwillee and James Pattison and it came up that it would be a great idea to also do it to the ESC_INFO message (which would get shrunk to 4 ESC only). Like that it could scale indefinitely while also keeping both messages small.

Also it was briefly discussed that while for this use case it is not possible to use the current Ardupilot message (hence the new message) but the opposite doesn't seem quite true to me. In fact it seems that only the `totalcurrent` and `count` fields are missing from this implementation. If so it could also be added and then it would become an all-in-one message meaning no need to bring the ardupilot messages to common, as these could be used instead.

message_definitions/v1.0/common.xml
amilcarlucas: @hamishwillee if that is the case, that information should be added to the description:
`Number of ESCs. Message fields with an index higher than this should be ignored because they contain invalid data.`

message_definitions/v1.0/common.xml
amilcarlucas: I guess this can also be i2c addressses ?

message_definitions/v1.0/common.xml
amilcarlucas: This is *not* static information, this is almost static information!

amilcarlucas:
@hamishwillee I like your sugestions.

message_definitions/v1.0/common.xml
hamishwillee: True. We should refer to it as ESC information that updates at a low rate, or similar.

message_definitions/v1.0/common.xml
hamishwillee: I guess so. We need to make sure that whatever this is it works for the possible use cases.

Had a discussion with Ricardo. He is going to chase up what UAVCAN might need too.

hamishwillee:
Thanks @amilcarlucas . We discussed in the dev call. Ricardo has some changes to make and we're also going to chase up to see what (if anything) might additionally be desirable for UAVCAN interworking.

Ideally this will be sufficiently better that ArduPilot will not be blocked from migrating to these, and may actually find it a good idea. I'm hoping you can be the person who comments with that perspective in mind.

message_definitions/v1.0/common.xml
RicardoM17: Changed it to

`Counter of data packets received.`

Let me know if that works

message_definitions/v1.0/common.xml
RicardoM17: Hoping @bkueng can chime in here. I'm going mostly off of the current esc_report msg in PX4.

message_definitions/v1.0/common.xml
RicardoM17: I guess it's related to the address field for example. Like I said in the comments there I'm going mostly off of the current esc_status msg in this case in PX4.

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Low rate ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint8_t" name="index">Index of the first ESC in this batch. (0, 4, 8 etc)</field>
```
amilcarlucas, line 6435: How about adding `minValue="0" maxValue="60" increment="4"`  ?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
...
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[4]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[4]" name="error_count">Number of reported errors by each ESC since boot.</field>
+      <field type="uint8_t[4]" name="temperature" units="degC">Temperature measured by each ESC.</field>
+    </message>
+    <message id="291" name="ESC_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC dynamic (high rate) information. Recommended to be sent at ~10 Hz. Information that updates at a lower rate is sent in ESC_INFO. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint8_t" name="index">Index of the first ESC in this batch. (0, 4, 8 etc)</field>
```
amilcarlucas, line 6449: How about adding `minValue="0" maxValue="60" increment="4"`  ?

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Low rate ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
```
amilcarlucas, line 6434: I think you should remove `Due to bandwidth limitations it should only be used in conjunction with a companion computer.`

RicardoM17: Yes I guess it can be removed from this message and stay only on ESC_STATUS.

message_definitions/v1.0/common.xml
RicardoM17: Can be done but does the maxValue have any meaning? I guess it's probably irrelevant but it's also a bit misleading.

amilcarlucas: Yes, the mavlink users shouldn't put a value of 100000 in there. 64 ESCs is pretty futureproff :)

RicardoM17: Sounds good. I was more asking in like why 60 and not 32 but I guess it's just a high number so we're sure it will never be reached.

message_definitions/v1.0/common.xml
```diff
@@ -6431,8 +6431,8 @@
     <message id="290" name="ESC_INFO">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Low rate ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
-      <field type="uint8_t" name="index">Index of the first ESC in this batch. (0, 4, 8 etc)</field>
+      <description>Low rate ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data.</description>
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this batch.</field>
```
amilcarlucas, line 6435: Can you replace "batch" with "message"?

message_definitions/v1.0/common.xml
```diff
@@ -6446,7 +6446,7 @@
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>ESC dynamic (high rate) information. Recommended to be sent at ~10 Hz. Information that updates at a lower rate is sent in ESC_INFO. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
-      <field type="uint8_t" name="index">Index of the first ESC in this batch. (0, 4, 8 etc)</field>
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this batch.</field>
```
amilcarlucas, line 6449: Can you replace "batch" with "message"?

RicardoM17: Done. @amilcarlucas . Once you are satisfied again let me know and I can squash everything again and update the commit description.

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Low rate ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data.</description>
```
amilcarlucas, line 6434: Replace "Low rate" with:
Slow changing ESC information.

This is not abot rates, this is about how fast the information changes. If a user has a lot of errors he might decide to ask for this message at a higher rate in order to see when the error counter increments. Or he might decide to only get this once in every 10 seconds, if he does not need this info

message_definitions/v1.0/common.xml
amilcarlucas: Move the index after the counter, to allow simplified Mavlink packaging

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
...
+      <field type="uint8_t[4]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[4]" name="error_count">Number of reported errors by each ESC since boot.</field>
+      <field type="uint8_t[4]" name="temperature" units="degC">Temperature measured by each ESC.</field>
+    </message>
+    <message id="291" name="ESC_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC dynamic (high rate) information. Recommended to be sent at ~10 Hz. Information that updates at a lower rate is sent in ESC_INFO. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
```
amilcarlucas, line 6450: The time_usec should be the first field in this message

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
...
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[4]" name="address">Address of each ESC.</field>
+      <field type="uint32_t[4]" name="error_count">Number of reported errors by each ESC since boot.</field>
+      <field type="uint8_t[4]" name="temperature" units="degC">Temperature measured by each ESC.</field>
+    </message>
+    <message id="291" name="ESC_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC dynamic (high rate) information. Recommended to be sent at ~10 Hz. Information that updates at a lower rate is sent in ESC_INFO. Due to bandwidth limitations it should only be used in conjunction with a companion computer.</description>
```
amilcarlucas, line 6448: ESC fast changing, dynamic information. Recommended to be sent at ~10 Hz. Information that changes slower is sent in ESC_INFO. Due to bandwidth limitations it should typically only be used in conjunction with a companion computer.

amilcarlucas:
Please squash and add a link to https://github.com/mavlink/mavlink/pull/1322 in the git commit message.

good job. Obrigado!

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Slow changing ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Counter of data packets received.</field>
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
```
hamishwillee, line 6451: @amilcarlucas You say "Move the index after the counter, to allow simplified Mavlink packaging".

The declaration order only matters between items of the same type - all the bigger ones will be magically reordered ahead of the smaller ones. The only exception is for extension fields, and we don't have any in this message.

I'm going to suggest it goes back. IMO index should be first.

message_definitions/v1.0/common.xml
```diff
@@ -6406,6 +6428,30 @@
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
     </message>
+    <message id="290" name="ESC_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Slow changing ESC information. Recommended rate 1Hz. See ESC_STATUS for more dynamic ESC data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Counter of data packets received.</field>
```
hamishwillee, line 6450: @RicardoM17 FMI  I still don't get this what are the data packets actually are/contain, and where are they coming from? What they are for.

message_definitions/v1.0/common.xml
hamishwillee: Min value is 1 right? ie. ESC typically numbered from 1?  FMI, why 60 for max value?

message_definitions/v1.0/common.xml
```diff
@@ -6420,6 +6442,30 @@
...
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
+      <field type="uint8_t" name="count">Number of ESCs in this message. Message fields with an index higher than this should be ignored because they contain invalid data.</field>
```
hamishwillee, line 6452: Note that the recipient now has to accumulate messages to know how many ESC there are. Is this OK?

message_definitions/v1.0/common.xml
```diff
@@ -6420,6 +6442,30 @@
...
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[4]" name="address">Address of each ESC.</field>
```
hamishwillee, line 6455: @RicardoM17 Is uint8 what is used in the UORB message? This implies at most 256 ESC right?

@bkueng For the uorb message for escs, the address is allocated by us right - it isn't something relevant to the bus type itself?

message_definitions/v1.0/common.xml
```diff
@@ -6420,6 +6442,30 @@
...
+      <field type="uint32_t[4]" name="error_count">Number of reported errors by each ESC since boot.</field>
+      <field type="uint8_t[4]" name="temperature" units="degC">Temperature measured by each ESC.</field>
```
hamishwillee, line 6457: Are all of these fields always going to be supplied? For example does every single ESC have a temperature sensor?

For any field that might not be supplied by an ESC we need a way to indicate "ESC does not send this information". For a this case we might use UINT8_MAX to indicate temperature not sent. For a float we use NaN.

Are all the fields of the right size/type for the likely data. I.e. for temperature, will be ever have negative temperatures or very large temps?

Is the resolutionright? I.e. is within a degree fine for the use case.

Need to ask this question for all values.

message_definitions/v1.0/common.xml
```diff
@@ -6420,6 +6442,30 @@
...
+    </message>
+    <message id="291" name="ESC_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>ESC fast changing, dynamic information. Recommended to be sent at ~10 Hz. Information that changes slower is sent in ESC_INFO. Due to bandwidth limitations it should typically only be used in conjunction with a companion computer.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
+      <field type="int32_t[4]" name="rpm" units="rpm">Reported motor RPM from each ESC (negative for reverse rotation).</field>
+      <field type="float[4]" name="voltage" units="V">Voltage measured from each ESC.</field>
```
hamishwillee, line 6466: See https://github.com/mavlink/mavlink/pull/1322/files#r445316985

Are these always sent (is there need for a not-sent value), do they have small enough resolution, do they have a big enough range. I see they go negative, is that OK/desired?

hamishwillee:
@RicardoM17 Much more to my taste. Structure works for me now, but added some further clarifications.

Before we merge I'd like to confirm with @bkueng we're not missing anything for UAVCAN or other bus we want to integrate.

message_definitions/v1.0/common.xml
RicardoM17: @hamishwillee I'm completely neutral on the whole index as I can see both reasonings. Whatever you and @amilcarlucas agree I will change.

Regarding the index itself, other than the whole arrays start at 0, for my current use case (as an example) with VESC + UAVCAN the VESC IDs start at 1 while the UAVCAN IDs start at 0 and this is the one used by PX4 iirc.

Regarding the 60max value you will find more context here:
https://github.com/mavlink/mavlink/pull/1322#discussion_r444817027

message_definitions/v1.0/common.xml
RicardoM17: It's a good point. This splitting the message in 4 and sending them in batches threw a wrench in my machine so I'm still not sure what is the best way to implement this on both sides of the message. I was hoping to finalise things here first before moving on to avoid double work.

That being said it might make sense to change this to `Total number of ESCs in all messages of this type.` as it could also help when subscribing to this data to know in advance how many there are. Also because in the other message the index will be higher so the whole `Message fields with an index higher than this should be ignored because they contain invalid data.` still applies.

message_definitions/v1.0/common.xml
RicardoM17: @hamishwillee correct. https://github.com/PX4/Firmware/blob/master/msg/esc_report.msg#L7

message_definitions/v1.0/common.xml
RicardoM17: > Are all of these fields always going to be supplied? For example does every single ESC have a temperature sensor?
> 
> For any field that might not be supplied by an ESC we need a way to indicate "ESC does not send this information". For a this case we might use UINT8_MAX to indicate temperature not sent. For a float we use NaN.

Not sure if they all would have every single sensor. I think that is a good approach. Should that be specified in every single field or is it the proper convention?

> Are all the fields of the right size/type for the likely data. I.e. for temperature, will be ever have negative temperatures or very large temps?
> 
> Is the resolutionright? I.e. is within a degree fine for the use case.

In theory you can have negative temperatures yes, especially at boot, if you are using this in VERY cold places. In practice I would say it's unlikely. Same for over 255 degrees. But I guess it can be changed to another, signed, type with a bigger range, possibly including a higher resolution. The Ardupilot uses some fields like cA which we could adopt here for the temperature for example. That being said the temperature doesn't change that fast and so the extra precision is not quite required.

For RPM and current the types are correct imo in both range and accuracy. They both need to be signed and rpm needs to be quite high as some ESC report eRPM instead which is a multiple of RPM so these values can be quite high.

I guess the one I see that could be changed is voltage. I believe it can be unsigned and the range can be smaller but with added accuracy.

this is the current ardupilot message implementation for context:

`      <field type="uint16_t[4]" name="voltage" units="cV">Voltage.</field>
`

But imo this cV is bit non-intuitive.

message_definitions/v1.0/common.xml
RicardoM17: Replied in the other thread to keep things centralized.

bkueng:
Looks good, nice work.

Maybe we want to add a failure state (over current, over temp, over voltage, motor stuck, ...)?

RicardoM17:
Many thanks @bkueng .

Regarding the failure state. Should I create an enum for that as well? Is there already some list (for example from PX4 side) that I could look into for all the fields? And to me this would be included in the ESC_INFO if everyone agrees.

Also not sure if you want to give some feedback on the other points currently open by @hamishwillee.

bkueng:
In PX4 there are currently these: https://github.com/PX4/Firmware/blob/master/msg/esc_report.msg#L14

RicardoM17:
Thanks again @bkueng . I had a look and based my approach on that.

@hamishwillee ready for another look? Just as a reminder both messages and enums are flagged as WIP so any changes we aren't seeing at this stage can always be corrected in the future.

RicardoM17:
@hamishwillee reminder. Anything else that needs to be added here? (It's fine if you haven't had the time, just commenting as you might have forgotten).

If not just let me know and i'll rebase, push so that it can be merged.

message_definitions/v1.0/common.xml
hamishwillee: OK. Have clarified this is a flight stack reference, not a "bus address" or something from the ESC itself

message_definitions/v1.0/common.xml
```diff
@@ -6429,6 +6479,31 @@
...
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Counter of data packets received.</field>
+      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
+      <field type="uint8_t" name="count">Total number of ESCs in all messages of this type. Message fields with an index higher than this should be ignored because they contain invalid data.</field>
+      <field type="uint8_t" name="connection_type" enum="ESC_CONNECTION_TYPE">Connection type protocol for all ESC.</field>
+      <field type="uint8_t" name="info" display="bitmask">Information regarding online/offline status of each ESC.</field>
+      <field type="uint8_t[4]" name="address">Address of each ESC.</field>
+      <field type="uint8_t[4]" name="failure_flags" enum="ESC_FAILURE_FLAGS" display="bitmask">Bitmap of ESC failure flags.</field>
```
hamishwillee, line 6493: @RicardoM17 This can contain at most 8 flags, which gives us no room to expand. @bkueng Are there likely to be more?
We could double the size if we think there is a reasonable chance there will be.

message_definitions/v1.0/common.xml
hamishwillee: 1. "not applicable" should be specified for every sensor. It is a convention, but there are exceptions.
2. W.r.t. voltage
   - Un-Intuitive is not a consideration :-). The theory is that MAVLink should use the best units for machine to machine transfer and the resolution needed. 
   - That said, assuming negative voltages are needed, I think it reasonable that we use float for this. I base this on the fact that for the range of likely values of esc voltage the float resolution is much better than you would get using CV.

hamishwillee: W.r.t. temperature I'm not sure what is relevant. But my feeling is that what we have now is suitable for the reasons you gave. We can always revisit this as you say.

Upshot, fixed up the "not applicable markup" but otherwise good.

message_definitions/v1.0/common.xml
hamishwillee: 1. Please move the index before timestamp. 
2. I don't think it matters what index we use as long as we state what the first item in the array must be so that a GCS or whatever can be sure it has all ESC reported.
3. 60 for max value is fine. If we want to change it later that is a non-breaking change.

hamishwillee:
@RicardoM17 I have done review/minor updates and have just a few minor remaining comments. Sorry for delay; pinging me right thing to do.

message_definitions/v1.0/common.xml
bkueng: Yes I think 8 is a bit tight for the mavlink message. I'd use 16 bits.

RicardoM17: Thanks @bkueng I'll make the changes later.

message_definitions/v1.0/common.xml
RicardoM17: Sounds good.

Also I'm trying to wrap my mind around what you said here

> Upshot, fixed up the "not applicable markup" but otherwise good.

Most importantly if there's any action needed still.

message_definitions/v1.0/common.xml
RicardoM17: Will do.

RicardoM17:
@hamishwillee I made the small changes you requested, rebased and squashed the commits into 1. 

It should be ready to go.

message_definitions/v1.0/common.xml
hamishwillee: Yes, what I am saying is that you still need to update the temperature field to add "UINT8_MAX if not supported by ESC.". Similar for the other fields that might not be supplied by some ESC.

hamishwillee: @RicardoM17 What I meant was "add UINT8_MAX if data not supplied by ESC." for the temperature field. I have done this.

Are there any other fields that are "optional" - ie data might not be supplied?

hamishwillee:
@RicardoM17 Still not entirely sure about the index/address.
```
 <message id="290" name="ESC_INFO">
      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
      <field type="uint8_t[4]" name="address">Address of each ESC. This is flight stack reference for the ESC; typically indexed from 1.</field>
...
```
```
   <message id="291" name="ESC_STATUS">
      <field type="uint8_t" name="index" minValue="0" maxValue="60" increment="4">Index of the first ESC in this message.</field>
```

I THINK that we would be best of removing the address field so that in both cases and changing index to
```
      <field type="uint8_t" name="first_address" minValue="0" maxValue="60" increment="4">Address of the first ESC in this message. Addresses are indexed from one, and the message contains payload for 4 sequential ESC addresses/field>
```

The reason is that the first message allows index and address, so presumably you could send four arbitrary addresses. But the second message just has index. If you have to map this to address across messages it is a pain. So it makes sense that the addresses are sequential. And then there is no difference between and address and an index.

Thoughts?

message_definitions/v1.0/common.xml
amilcarlucas: What about negative temperatures. I know the global warming is making them rare, ... but it might make sense.
Usualy electronics can not work at more than 125°C
Power electronics might go all the way to 150°C though...

RicardoM17:
@hamishwillee 

I have some limited knowledge on this part but I believe that the assumption that indexes and adresses are always the same thing is not entirely correct. For example with my current setup with VESC + UAVCAN the adresses start at 1 and are sequential after that, while the indexes start at 0. Now this is a small change and I'm assuming with other UAVCAN ESC the address will also start at 0 so there they would match indeed. The questions is what happens with I2C type ESCs where they can have a much larger range.

message_definitions/v1.0/common.xml
RicardoM17: I guess we can be extra safe here and either

1. change to int16 so we have a negative range that would work for all electronics.

2. change to float so that we have the negative range + added resolution (that way we can have not just int temperatures).

hamishwillee: Other options:
3. We accept max of 127 and make this an int_8
4. We create a duplicate array for negative values, but put it at the end (it could be a flag or a value). Either way, default is zero filled, so is empty unless we go negative and gets truncated.

Anyone have an opinion? I lean towards using a float.

hamishwillee:
@RicardoM17 Beat indicated that the address is PX4-allocated - it isn't a UAVCAN address, but a handle for PX4. So it will range from 1-4 or 1-8 in all cases. So does the internal UAVCAN address matter?

message_definitions/v1.0/common.xml
```diff
@@ -6442,6 +6492,30 @@
       <field type="float" name="pan" units="rad">Pan/yaw angle (positive: to the right, negative: to the left, NaN to be ignored).</field>
       <field type="float" name="tilt_rate" units="rad/s">Tilt/pitch angular rate (positive: up, negative: down, NaN to be ignored).</field>
       <field type="float" name="pan_rate" units="rad/s">Pan/yaw angular rate (positive: to the right, negative: to the left, NaN to be ignored).</field>
+    <message id="290" name="ESC_INFO">
```
hamishwillee, line 6495: Something weird going on here. We're missing the end message tag above for the last message.

hamishwillee: Lets see if my fix makes the test pass.

RicardoM17: weird indeed. Probably that error was introduced in one my (many) rebase/squashes etc

FYI the error is regarding the index line, more consequently the increment

`/home/travis/build/mavlink/mavlink/message_definitions/v1.0/common.xml:6500:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'minValue': The attribute 'minValue' is not allowed.
/home/travis/build/mavlink/mavlink/message_definitions/v1.0/common.xml:6500:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'maxValue': The attribute 'maxValue' is not allowed.
/home/travis/build/mavlink/mavlink/message_definitions/v1.0/common.xml:6500:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'increment': The attribute 'increment' is not allowed.
/home/travis/build/mavlink/mavlink/message_definitions/v1.0/common.xml:6515:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'minValue': The attribute 'minValue' is not allowed.
/home/travis/build/mavlink/mavlink/message_definitions/v1.0/common.xml:6515:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'maxValue': The attribute 'maxValue' is not allowed.
/home/travis/build/mavlink/mavlink/message_definitions/v1.0/common.xml:6515:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'increment': The attribute 'increment' is not allowed.`

RicardoM17:
> @RicardoM17 Beat indicated that the address is PX4-allocated - it isn't a UAVCAN address, but a handle for PX4. So it will range from 1-4 or 1-8 in all cases. So does the internal UAVCAN address matter?

Not sure. I guess in that case we can remove this entirely. For my use case it's not really required, it will depend a bit on how things are handled PX4 side however for some more unusual use cases.

message_definitions/v1.0/common.xml
RicardoM17: it seems that `minValue` `maxValue` and `increment` are not allowed in message fields? All of these fields are present only in the enums in common.xml at least

amilcarlucas:
I think having the UAVCAN address information is interresting. Having an internal PX4 address information is not interresting because it is sequencial and predefined.

message_definitions/v1.0/common.xml
amilcarlucas: Probably they are only available in MAV_CMD commands

hamishwillee:
@RicardoM17 @amilcarlucas OK, so Beat clarified that the internal uorb topic WAS for uavcan information, but is not used by PX4. The dev call decided that we should remove that field for now. It can be added back as an extension if desired in future.

message_definitions/v1.0/common.xml
hamishwillee: Yes. We'll have to remove them. If useful, we add that as comments in the descriptions.

@amilcarlucas Is there a reason we don't have these in messages? i.e. should we?

amilcarlucas: There is no reason, and yes, I think we should add them.

RicardoM17: Should I then remove them from the current message, add them to the description and then change it at some point once this feature is added (which I believe should be a separate PR)?

If I can help in adding this, let me know as well. I just need to know where to look but it could be that this is a bit out of my scope.

RicardoM17:
Sounds good @hamishwillee I will make the changes.

message_definitions/v1.0/common.xml
hamishwillee: Thanks @RicardoM17 - @amilcarlucas do you want to update messages with min/max/incremented/default? Or should i do it (if me, I may hand to Ricardo :-))

RicardoM17:
Thanks @hamishwillee! that was a long ride but I feel a lot of progress was made. :sweat_smile:

PR #1323, opened 2020-03-18
Improve cellular status mavlink message

Jaeyoung-Lim:
The current definition of mavlink fails to provide sufficient information to make use of cellular communications as a datalink. This PR improves the `CELLULAR_STATUS` message to improve monitoring / diagnoising  cellular modems.

The summary of the improvements include
- Extension of enum of `CELLULAR_STATUS_FLAG` to better represent the cellular status
- Switched unit of cellular quality from -dbm to percent
- Added `failure_status` flag in message to represent failur status when `CELLULAR_STATUS_FLAG` is in `CELLULAR_STATUS_FLAG_FAILED` state
- Added field of band with enum `CELLULAR_NETWORK_BAND` to represent band usage for cellular networks 

A lot of the enumerations and definitions are inspired from the definitions of [ModemManager API](https://www.freedesktop.org/software/ModemManager/api/latest/gdbus-org.freedesktop.ModemManager1.Modem.html),  which are widely used. This also makes the interfacing with Modem Manager simpler

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3949,10 +3949,61 @@
       <entry value="3" name="CELLULAR_NETWORK_RADIO_TYPE_WCDMA"/>
       <entry value="4" name="CELLULAR_NETWORK_RADIO_TYPE_LTE"/>
     </enum>
-    <enum name="CELLULAR_NETWORK_STATUS_FLAG" bitmask="true">
+    <enum name="CELLULAR_STATUS_FLAG">
       <description>These flags encode the cellular network status</description>
-      <entry value="1" name="CELLULAR_NETWORK_STATUS_FLAG_ROAMING">
-        <description>Roaming is active</description>
+      <entry value="0" name="CELLULAR_STATUS_FLAG_FAILED">
+        <description>Modem is unusable</description>
+      </entry>
+      <entry value="1" name="CELLULAR_STATUS_FLAG_UNKNOWN">
+        <description>State unknown or not reportable.</description>
+      </entry>
+      <entry value="2" name="CELLULAR_STATUS_FLAG_INITIALIZING">
+        <description>Modem is being initialized</description>
+      </entry>
+      <entry value="3" name="CELLULAR_STATUS_FLAG_LOCKED">
+        <description>Modem is locked</description>
+      </entry>
+      <entry value="4" name="CELLULAR_STATUS_FLAG_DISABLED">
+        <description>Modem is not enabled and is powered down</description>
+      </entry>
+      <entry value="5" name="CELLULAR_STATUS_FLAG_DISABLING">
+        <description>Modem is currently transitioning to the CELLULAR_STATUS_FLAG_DISABLED state</description>
+      </entry>
+      <entry value="6" name="CELLULAR_STATUS_FLAG_ENABLING">
+        <description>Modem is currently transitioning to the CELLULAR_STATUS_FLAG_ENABLED state</description>
+      </entry>
+      <entry value="7" name="CELLULAR_STATUS_FLAG_ENABLED">
+        <description>Modem is enabled and powered on but not registered with a network provider and not available for data connections</description>
+      </entry>
+      <entry value="8" name="CELLULAR_STATUS_FLAG_SEARCHING">
+        <description>Modem is searching for a network provider to register</description>
+      </entry>
+      <entry value="9" name="CELLULAR_STATUS_FLAG_REGISTERED">
+        <description>Modem is registered with a network provider, and data connections and messaging may be available for use</description>
+      </entry>
+      <entry value="10" name="CELLULAR_STATUS_FLAG_DISCONNECTING">
+        <description>Modem is disconnecting and deactivating the last active packet data bearer. This state will not be entered if more than one packet data bearer is active and one of the active bearers is deactivated</description>
+      </entry>
+      <entry value="11" name="CELLULAR_STATUS_FLAG_CONNECTING">
+        <description>Modem is activating and connecting the first packet data bearer. Subsequent bearer activations when another bearer is already active do not cause this state to be entered</description>
+      </entry>
+      <entry value="12" name="CELLULAR_STATUS_FLAG_CONNECTED">
+        <description>One or more packet data bearers is active and connected</description>
+      </entry>
+    </enum>
+    <enum name="CELLULAR_NETWORK_FAILED_REASON">
+      <description>These flags are used to diagnose the failure state of CELLULAR_STATUS</description>
+      <entry value="0" name="CELLULAR_NETWORK_FAILED_REASON_NONE">
+        <description>No error</description>
+      </entry>
+      <entry value="1" name="CELLULAR_NETWORK_FAILED_REASON_UNKNOWN">
+        <description>Error state is unknown</description>
+      </entry>
+      <entry value="2" name="CELLULAR_NETWORK_FAILED_REASON_SIM_MISSING">
+        <description>SIM is required for the modem but missing</description>
+      </entry>
+      <entry value="3" name="CELLULAR_NETWORK_FAILED_REASON_SIM_ERROR">
+        <description>SIM is available, but not usuable for connection</description>
       </entry>
     </enum>
     <enum name="PRECISION_LAND_MODE">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6296,13 +6473,14 @@
     <message id="334" name="CELLULAR_STATUS">
       <wip/>
       <description>Report current used cellular network status</description>
```
LorenzMeier: This looks all good, but what about a continuous quality metric? E.g. drop rate / error rate per second, etc? Is there anything available we could use?

Jaeyoung-Lim: Hmm that is a good point, and would certainly be useful. But looking at [modemmanager](https://www.freedesktop.org/software/ModemManager/api/latest/) i don't see any metrics that give continuous quality metrics. This might be a dumb question - The information represented here are mostly focused on monitoring / diagnosing the status of the modem, and not the network between two endpoints. Where do we expect these error rates to be monitored?  how do we define such drop / error rates if we have more than one endpoint?

message_definitions/v1.0/common.xml
```diff
@@ -3771,10 +3771,187 @@
...
+      <entry value="119" name="CELLULAR_NETWORK_BAND_UTRAN_25"/>
+      <entry value="120" name="CELLULAR_NETWORK_BAND_UTRAN_26"/>
+      <entry value="121" name="CELLULAR_NETWORK_BAND_UTRAN_32"/>
+      <entry value="122" name="CELLULAR_NETWORK_BAND_ANY"/>
+    </enum>
+    <enum name="CELLULAR_STATUS_FLAG">
       <description>These flags encode the cellular network status</description>
-      <entry value="1" name="CELLULAR_NETWORK_STATUS_FLAG_ROAMING">
-        <description>Roaming is active</description>
+      <entry value="0" name="CELLULAR_STATUS_FLAG_FAILED">
```
julianoes, line 3902: I would consider UNKNOWN as 0 or OK as 0 but not failed.

message_definitions/v1.0/common.xml
```diff
@@ -3771,10 +3771,187 @@
       <entry value="3" name="CELLULAR_NETWORK_RADIO_TYPE_WCDMA"/>
       <entry value="4" name="CELLULAR_NETWORK_RADIO_TYPE_LTE"/>
     </enum>
-    <enum name="CELLULAR_NETWORK_STATUS_FLAG">
+    <enum name="CELLULAR_NETWORK_BAND">
+      <description>Cellular network radio type</description>
```
julianoes, line 3775: So is it the network band or radio type?

message_definitions/v1.0/common.xml
hamishwillee: Yes, CELLULAR_STATUS_FLAG_UNKNOWN better 0 value. I

message_definitions/v1.0/common.xml
```diff
@@ -3771,10 +3771,187 @@
...
+        <description>One or more packet data bearers is active and connected</description>
+      </entry>
+    </enum>
+    <enum name="CELLULAR_NETWORK_FAILED_REASON">
+      <description>These flags are used to diagnose the failure state of CELLULAR_STATUS</description>
+      <entry value="0" name="CELLULAR_NETWORK_FAILED_REASON_NONE">
+        <description>No error</description>
+      </entry>
+      <entry value="1" name="CELLULAR_NETWORK_FAILED_REASON_UNKNOWN">
+        <description>Error state is unknown</description>
```
hamishwillee, line 3948: Perhaps better as "Fail reason unknown". The error state is know - it has failed :-)

message_definitions/v1.0/common.xml
```diff
@@ -6296,13 +6473,14 @@
     <message id="334" name="CELLULAR_STATUS">
       <wip/>
       <description>Report current used cellular network status</description>
-      <field type="uint16_t" name="status" enum="CELLULAR_NETWORK_STATUS_FLAG" display="bitmask">Status bitmap</field>
+      <field type="uint8_t" name="status" enum="CELLULAR_STATUS_FLAG">Cellular modem status</field>
+      <field type="uint8_t" name="failure_status" enum="CELLULAR_NETWORK_FAILED_REASON">Failure status when status in in CELLUAR_STATUS_FAILED</field>
```
hamishwillee, line 6477: - rename to failure_reason
- change desc to "Failure reason when status in in CELLUAR_STATUS_FAILED"

message_definitions/v1.0/common.xml
hamishwillee: @Jaeyoung-Lim Don't forget to raise this in meeting.

message_definitions/v1.0/common.xml
```diff
@@ -6296,13 +6473,14 @@
...
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
-      <field type="uint8_t" name="quality">Cellular network RSSI/RSRP in dBm, absolute value</field>
+      <field type="uint8_t" name="quality">Signal Quality in percent</field>
+      <field type="uint16_t[5]" name="band" enum="CELLULAR_NETWORK_BAND">Currently used band</field>
```
hamishwillee, line 6480: This is a set of bands, max of 5.
1. Is 5 enought?
2. Change the description to "Set of supported bands". (plural)
3. Maybe name "supported_bands"

message_definitions/v1.0/common.xml
```diff
@@ -6296,13 +6473,14 @@
     <message id="334" name="CELLULAR_STATUS">
       <wip/>
       <description>Report current used cellular network status</description>
-      <field type="uint16_t" name="status" enum="CELLULAR_NETWORK_STATUS_FLAG" display="bitmask">Status bitmap</field>
+      <field type="uint8_t" name="status" enum="CELLULAR_STATUS_FLAG">Cellular modem status</field>
+      <field type="uint8_t" name="failure_status" enum="CELLULAR_NETWORK_FAILED_REASON">Failure status when status in in CELLUAR_STATUS_FAILED</field>
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
-      <field type="uint8_t" name="quality">Cellular network RSSI/RSRP in dBm, absolute value</field>
+      <field type="uint8_t" name="quality">Signal Quality in percent</field>
```
hamishwillee, line 6479: Will this always be sent? If not, then define value to indicate "not supported"

message_definitions/v1.0/common.xml
Jaeyoung-Lim: This field is not for showing what band the device supports, but what band the modem is currently using for communication.

message_definitions/v1.0/common.xml
Jaeyoung-Lim: I agree that the suggestion is more clear. However, this is a direct map to the failure reasons from `libmm` (lib-modem-manager): https://www.freedesktop.org/software/ModemManager/api/latest/ModemManager-Flags-and-Enumerations.html#MMModemStateFailedReason My intention was to keep it one to one so that the status of the modem is more transparent

message_definitions/v1.0/common.xml
hamishwillee: Then why is it an array of 5 things?

message_definitions/v1.0/common.xml
hamishwillee: Note, dev call agreed this not needed as is not supplied by known modems, and MAVLink should allow generation of this info outside of this context.

Jaeyoung-Lim: Discussed in the mavlink call, and we decided that the continuous quality metric should not be included in the mavlink message

message_definitions/v1.0/common.xml
Jaeyoung-Lim: Done!

message_definitions/v1.0/common.xml
Jaeyoung-Lim: Removed cellular band

message_definitions/v1.0/common.xml
Jaeyoung-Lim: Removed band field

message_definitions/v1.0/common.xml
Jaeyoung-Lim: Done!

hamishwillee:
Merging now that changes discussed in the Dev call have now been made. Thanks @Jaeyoung-Lim !

PR #1324, opened 2020-03-19
Fix description errors

ndepal:
I noticed a few errors in the message descriptions that this PR fixes:
- In `GPS_RAW_INT`: Uncertainties are described with "Positive for up". That makes no sense.
- Many messages have the timestamp field described with "by checking for the magnitude the number", should be "magnitude *of* the number".
- `HOME_POSITION`: A sentence was repeated, removed duplicate.


LorenzMeier:
Thanks, much appreciated!

PR #1325, opened 2020-03-19
common.xml: Clarify some OpenDroneID fields

friissoren:
The description for some fields were a bit misleading
and more precise information was missing in some.

Signed-off-by: Soren Friis <soren.friis@intel.com>

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6482,27 +6482,27 @@
       <wip/>
       <!-- This message is work-in-progress and it can therefore change, and should NOT be used in stable production environments -->
       <description>Data for filling the OpenDroneID Location message. The float data types are 32-bit IEEE 754. The Location message provides the location, altitude, direction and speed of the aircraft.</description>
-      <field type="uint8_t" name="status" enum="MAV_ODID_STATUS">Indicates whether the Unmanned Aircraft is on the ground or in the air.</field>
-      <field type="uint16_t" name="direction" units="cdeg">Direction over ground (not heading, but direction of movement) in degrees * 100: 0.0 - 359.99 degrees. If unknown: 361.00 degrees.</field>
-      <field type="uint16_t" name="speed_horizontal" units="cm/s">Ground speed. Positive only. If unknown: 255.00 m/s. If speed is larger than 254.25 m/s, use 254.25 m/s.</field>
-      <field type="int16_t" name="speed_vertical" units="cm/s">The vertical speed. Up is positive. If unknown: 63.00 m/s. If speed is larger than 62.00 m/s, use 62.00 m/s.</field>
-      <field type="int32_t" name="latitude" units="degE7">Current latitude of the UA (Unmanned Aircraft). If unknown: 0 deg (both Lat/Lon).</field>
-      <field type="int32_t" name="longitude" units="degE7">Current longitude of the UA (Unmanned Aircraft). If unknown: 0 deg (both Lat/Lon).</field>
+      <field type="uint8_t" name="status" enum="MAV_ODID_STATUS">Indicates whether the unmanned aircraft is on the ground or in the air.</field>
+      <field type="uint16_t" name="direction" units="cdeg">Direction over ground (not heading, but direction of movement) measured clockwise from true North: 0 - 35999 centi-degrees. If unknown: 36100 centi-degrees.</field>
+      <field type="uint16_t" name="speed_horizontal" units="cm/s">Ground speed. Positive only. If unknown: 25500 cm/s. If speed is larger than 25425 cm/s, use 25425 cm/s.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6483,11 +6483,11 @@
...
-      <field type="uint16_t" name="direction" units="cdeg">Direction over ground (not heading, but direction of movement) in degrees * 100: 0.0 - 359.99 degrees. If unknown: 361.00 degrees.</field>
-      <field type="uint16_t" name="speed_horizontal" units="cm/s">Ground speed. Positive only. If unknown: 255.00 m/s. If speed is larger than 254.25 m/s, use 254.25 m/s.</field>
-      <field type="int16_t" name="speed_vertical" units="cm/s">The vertical speed. Up is positive. If unknown: 63.00 m/s. If speed is larger than 62.00 m/s, use 62.00 m/s.</field>
-      <field type="int32_t" name="latitude" units="degE7">Current latitude of the UA (Unmanned Aircraft). If unknown: 0 deg (both Lat/Lon).</field>
-      <field type="int32_t" name="longitude" units="degE7">Current longitude of the UA (Unmanned Aircraft). If unknown: 0 deg (both Lat/Lon).</field>
+      <field type="uint16_t" name="direction" units="cdeg">Direction over ground (not heading, but direction of movement) in degrees * 100 measured clockwise from true North: 0 - 35999 centi-degrees. If unknown: 36100 centi-degrees.</field>
```
hamishwillee: How about remove "in degrees * 100". Units are in the units field. The rest is good.

message_definitions/v1.0/common.xml
```diff
@@ -6483,11 +6483,11 @@
...
+      <field type="uint16_t" name="speed_horizontal" units="cm/s">Ground speed. Positive only. If unknown: 25500 cm/s. If speed is larger than 25425 cm/s, use 25425 cm/s.</field>
```
hamishwillee: This is a 16 bit number, so max is 65K values. Why are we use numbers that probably were designed for an 8 bit value (ie around 25600)? Also we should juse use the last two biggest numbers right?

message_definitions/v1.0/common.xml
```diff
@@ -6483,11 +6483,11 @@
...
+      <field type="int16_t" name="speed_vertical" units="cm/s">The vertical speed. Up is positive. If unknown: 6300 cm/s. If speed is larger than 6200 cm/s, use 6200 cm/s. If lower than -6200 cm/s, use -6200 cm/s.</field>
```
hamishwillee: Why cropped to 6300? / or why this type to store the value?

message_definitions/v1.0/common.xml
```diff
@@ -6483,11 +6483,11 @@
...
+      <field type="int32_t" name="latitude" units="degE7">Current latitude of the UA (Unmanned Aircraft). If unknown: 0 degE7 (both Lat/Lon).</field>
```
hamishwillee: - Remove ` degE7 ` in the description.
- I'd be tempted to make "UA (Unmanned Aircraft)" to "unmanned aircraft". After all, you can't use the contraction UA elsewhere.

message_definitions/v1.0/common.xml
```diff
@@ -6483,11 +6483,11 @@
...
+      <field type="int32_t" name="longitude" units="degE7">Current longitude of the UA (Unmanned Aircraft). If unknown: 0 degE7 (both Lat/Lon).</field>
```
hamishwillee: Remove `degE7` in the description.

message_definitions/v1.0/common.xml
friissoren: The ASTM remote ID standard uses 9 bits to encode the direction and the horizontal speed.
You can see an earlier draft of the standard here: https://github.com/opendroneid/specs
Look at the Message specification.

message_definitions/v1.0/common.xml
friissoren: The standard uses only 8-bit but it allows 0.5 m/s values and I didn't want to push the quantization/conversion stuff to the user of the Mavlink API so I thought it was more clear if the same unit was used for both vertical and horizontal speed.

message_definitions/v1.0/common.xml
friissoren: Done

message_definitions/v1.0/common.xml
friissoren: Done

message_definitions/v1.0/common.xml
friissoren: Done

message_definitions/v1.0/common.xml
hamishwillee: Thanks. Makes sense.

message_definitions/v1.0/common.xml
hamishwillee: Fair enough. Though you

friissoren: Sorry, did your comment get cut off in half? It appears incomplete?
Do you have any other concerns or could this PR get merged?

hamishwillee: Sorry, no problem. I will merge now.

PR #1329, opened 2020-03-31
Add gimbal device id param to gimbal_manager_set_attitude

coderkalyan:
the GIMBAL_MANAGER_SET_ATTITUDE messsage is missing a parameter to specify gimbal device ID, this PR adds it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6074,6 +6074,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags to use.</field>
+      <field type="uint8_t" name="gimbal_device_id">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. (Send command multiple times for more than one but not all gimbals.)</field>
       <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation, the frame is depends on whether the flag GIMBAL_MANAGER_FLAGS_YAW_LOCK is set)</field>
       <field type="float" name="angular_velocity_x" units="rad/s">X component of angular velocity, positive is banking to the right, NaN to be ignored.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6077,6 +6077,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags to use.</field>
+      <field type="uint8_t" name="gimbal_device_id">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. (Send command multiple times for more than one but not all gimbals.)</field>
```
julianoes, line 6080: In `GIMBAL_INFORMATION` we call it `gimbal_component`. It would probably be better to have them consistent. Which one is better?

coderkalyan: @julianoes In my opinion, gimbal device id, or even just gimbal_id, because we cannot assume that the gimbal device is even a component - the manager provides a wrapper for PWM or other gimbals as well...

coderkalyan:
Update: @julianoes I renamed `gimbal_component` to `gimbal_device_id` in `GIMBAL_INFORMATION`. I also added this field to the `GIMBAL_MANAGER_STATUS` message, will probably be useful to know which gimbal manager the status belongs to :thinking: :smile: 

I also renamed `angular rate` to `angular velocity`.

julianoes:
@olliw42 this makes sense, right?

olliw42:
hey folks

sorry for being so totally unresponsive to this mavlink gimbal stuff lately. I just want to quickly state, but really want to state:

You should NOT think that this is because I would have lost interest, to the contrary!

It's just that I don't have the concentration to look at it with sufficient care to make useful comments. And the reason for why I lack the concentration is because I'm heavily working on getting to a point there I actually can implement AND use this stuff in real world! (you know that QGC turned out to not be an efficient path to progress forward, so I had to establish a substitute which would work for me, and it's where my time goes into currently). That is, even though I'm silent on the topic now, I - hopefully - will come back with experience from practical usage - which I believe might be more helpful than just theories :):):)

So, please continue to keep me in the line as you did here, it's very much appreciated, and please don't be upset about my silence :)

cheers, and all the best,
Olli

julianoes:
@olliw42 phew, good to know! Thanks for the heads-up and happy Easter!

PR #1331, opened 2020-04-01
Add failure injection command with enums

julianoes:
In order to test failsafe mechanisms it would be nice to be able to inject failures using a command.

Note that for safetey and security, autopilots should only accept such a failure injection command if this has been specifically enabled using e.g. a parameter or even a flag set at build time.

This commit is just a first suggestion how this could look like and in no means exhaustive.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2039,6 +2039,14 @@
         <param index="6">Reserved</param>
         <param index="7">Reserved</param>
       </entry>
+      <entry value="420" name="MAV_CMD_INJECT_FAILURE">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Inject artificial failure for testing purposes. Note that autopilots should implement an additional protection before accepting this command such as a specific param setting.</description>
+        <param index="1" name="Failure unit" enum="FAILURE_UNIT">The unit which is affected by the failure.</param>
+        <param index="2" name="Failure type" enum="FAILURE_TYPE">The type how the failure manifests itself.</param>
+        <param index="3" name="Instance">Instance affected by failure (0 to signal all).</param>
+      </entry>
       <entry value="500" name="MAV_CMD_START_RX_PAIR" hasLocation="false" isDestination="false">
         <description>Starts receiver pairing.</description>
         <param index="1" label="Spektrum">0:Spektrum.</param>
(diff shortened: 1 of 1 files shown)


hamishwillee:
As discussed in dev call.
- This looks good for me, modulo a way to make sure it cannot be invoked by accident. 
- @auturgy Suggests there are possibly ArduPilot messages that do same thing. He will provide info so we can evaluate/compare.

dagar:
Looks good to me, but can we wait for @auturgy to make sure we don't introduce unnecessary duplication?

auturgy:
Ours are parameters not messages, so no duplicates 

Regards,

James

hamishwillee:
OK, merging as WIP.

@auturgy Can you help us understand what, if anything, in the ArduPilot parameters is not covered by this message? It would be good if this could offer an alternative path for ArduPilot in future.

message_definitions/v1.0/common.xml
```diff
@@ -4285,6 +4293,52 @@
       <entry value="2048" name="AIS_FLAGS_VALID_CALLSIGN"/>
       <entry value="4096" name="AIS_FLAGS_VALID_NAME"/>
     </enum>
+    <enum name="FAILURE_UNIT" bitmask="true">
+      <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
```
amilcarlucas, line 4297: This comment is NOT enough. You need to add a <wip> xml tag

message_definitions/v1.0/common.xml
```diff
@@ -4285,6 +4293,52 @@
...
+      <entry value="6" name="FAILURE_UNIT_SENSOR_VIO"/>
+      <entry value="7" name="FAILURE_UNIT_SENSOR_DISTANCE_SENSOR"/>
+      <entry value="100" name="FAILURE_UNIT_SYSTEM_BATTERY"/>
+      <entry value="101" name="FAILURE_UNIT_SYSTEM_MOTOR"/>
+      <entry value="102" name="FAILURE_UNIT_SYSTEM_SERVO"/>
+      <entry value="103" name="FAILURE_UNIT_SYSTEM_AVOIDANCE"/>
+      <entry value="104" name="FAILURE_UNIT_SYSTEM_RC_SIGNAL"/>
+      <entry value="105" name="FAILURE_UNIT_SYSTEM_MAVLINK_SIGNAL"/>
+    </enum>
+    <enum name="FAILURE_TYPE" bitmask="true">
```
amilcarlucas, line 4314: If this is a bitmask, the values need to be a power of two.

message_definitions/v1.0/common.xml
```diff
@@ -4285,6 +4293,52 @@
       <entry value="2048" name="AIS_FLAGS_VALID_CALLSIGN"/>
       <entry value="4096" name="AIS_FLAGS_VALID_NAME"/>
     </enum>
+    <enum name="FAILURE_UNIT" bitmask="true">
```
amilcarlucas, line 4296: This is not a bitmask!!!

message_definitions/v1.0/common.xml
hamishwillee: @amilcarlucas You can't add WIP to an enum - only an enum entry. So this is "best that can be done right now".

That seems like a flaw in the XML schema - can you add that support so we can update these properly?

message_definitions/v1.0/common.xml
hamishwillee: Thanks. Fixed by https://github.com/mavlink/mavlink/pull/1338

message_definitions/v1.0/common.xml
hamishwillee: Thanks. Fixed by https://github.com/mavlink/mavlink/pull/1338

julianoes:
Thanks for the review @amilcarlucas and thanks for the fix @hamishwillee!

PR #1337, opened 2020-04-02
Add RAW_RPM message

roman-dvorak:
Hi, I would like to add new MAVLINK message for RPM data. This message is related to TFRPM01A sensor implemented in PX4 PX4/Firmware#14018. It can be used with other RPM sensors or for broadcasting RPM from some CAN ESC.

RAW_RPM
```
uint64 timestamp
float32 indicated_frequency_rpm       # indicated rotor Frequency in Revolution per minute
float32 estimated_accurancy_rpm       # estimated accurancy in Revolution per minute
```

Plese, I'm not sure about message ID.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6402,6 +6402,11 @@
       <field type="uint8_t" name="tx_session_pending">1: Transmission session pending, 0: No transmission session pending.</field>
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
+    <message id="339" name="RAW_RPM">
+      <description>RPM sensor data message.</description>
+      <field type="uint8_t" name="index">Index of this RPM sensor (0-indexed)</field>
+      <field type="float" name="frequency" units="rpm">Indicated rate</field>
+    </message>
     <message id="340" name="UTM_GLOBAL_POSITION">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->


message_definitions/v1.0/common.xml
```diff
@@ -6314,6 +6314,12 @@
       <field type="uint8_t" name="tx_session_pending">1: Transmission session pending, 0: No transmission session pending.</field>
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
+    <message id="336" name="RAW_RPM">
```
hamishwillee, line 6317: Perhaps id 339? That means the status messages below can expand at least a few more.

message_definitions/v1.0/common.xml
```diff
@@ -6314,6 +6314,12 @@
...
+      <description>RPM sensor data message.</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="float" name="indicated_frequency_rpm">Indicated rotor Frequency in Revolution per minute.</field>
```
hamishwillee, line 6320: For both of these fields ...

NOrmally we put the units in an attribute. However there is no unit for rpm. So first step would be to add one by creating a PR in the XSD: https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L96

Then we would update this like:
```
<field type="float" name="indicated_frequency" units="rpm">Indicated rotor frequency.</field>
<field type="float" name="estimated_frequency" units="rpm">Estimated rotor frequency.</field>
```

Is it possible that  an RPM sensor might not be able to provide one of these fields? E.g. it could only provide indicated, but not estimated? If so, you should specify explicitly how this would be represented. :
```
<field type="float" name="estimated_frequency" units="rpm">Estimated rotor frequency. NaN indicates this field not supported.</field>
```

A few more questions
- Is there anything else an rpm sensor might reasonably send? E.g. temperature, run time, etc etc.
- Will there only ever be one of these per system? If there can be multiples you'd normally add:
  ```
  <field type="uint_8" name="index">Index of this RPM sensor (0-indexed).</field>
  ```

@auturgy Anything else you might add?

auturgy:
Relevant - https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/ardupilotmega.xml#L1556
Re schema for units: its there already, https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L106
The timestamp is interesting: I'm not convinced it's necessary as payload within the message.  Are you expecting significant and relevant lag between when the message is filled, sent and parsed?

auturgy:
I'd add that we have RPM as a field in other messages, as well as stand-alone (EFI for instance), so depending on the use case this isn't the only way ArduPilot sends/receives RPM data.

message_definitions/v1.0/common.xml
roman-dvorak: I think that RPM sensor should provide indicated value every time. The idea of missing an estimated value is interesting. And I think that it is possible. 

Temperature is also interesting. Because it is sometimes good to know the temperature of bearings or similar thing. Run time can be calculated additionally from those messages and I think that is a redundant data. 

Yes, in the future it will be possible for the UAV to have more sensors. Added msg ID

roman-dvorak:
@auturgy 
I know this message. That's why I renamed it RAW_RPM. But this message is not well designed. It assumes 2 RPM sensors (data) and it is not indexable. The temperature parameter is quite interesting and important - it is also missing in ardupilot msg.

ESC message contains many unused fields for us. And then it's an unnecessary waste of mavlink network capacity.

The question is, is uint16 not enough? So far we have a float.

message_definitions/v1.0/common.xml
roman-dvorak: Ok, changed in new commit.

auturgy:
I think you need to articulate the use case a bit more: who is sending the message, who is receiving it, what are they doing with it.
If the autopilot is consuming it, the sensor needs to generate the time stamp. Not sure that’s likely or necessary.
If the gcs is consuming it, why is the extra fidelity of a time stamp on the reading needed? It’ll (likely) be generated anyway for the onboard logging, but mavlink messages get an 8 bit time stamp in the gcs, which is probably enough for this use case.
If an offboard/companion process is using it: what is the use case for a time stamp on the reading rather than the message?

Also, Temp isn’t really “raw_rpm”. It’s potentially useful, but needs context, and once given context probably belongs in a different message -> ie it seems to be scope creep.

I agree the ardupilot rpm message is limited, but the majority of use cases for rpm are covered by other messages (esc telemetry, efi, wheel odometry etc).

Basically: if you want “raw_rpm”, don’t make this more complicated just because you can. Articulate what you need the message to achieve, and just implement that.

If you want an index, you could just add that as an extension to the ArduPilot message - extensions are zero trimmed, so it won’t break anything if not used.

roman-dvorak:
Ok, our main usage is to get RPM data of rotor of our autogyro. It is important to know it before take-off. This value should be displayed to some ground station (mavros, pymavlink, maybe even QGC). This message can be also connected to some application related to the automatic launch release externally. 

Our start looks like that rotor must have the required speed before starting. It is ensured with prerotator motor. The captain of flight checks the rpm and gives the start command when the minimum speed is reached.
So, it is really important for us. :)

So, index, estimated and indicated values are enought for us. 

RPM message from ArduPilot messages. We would like tu use it mainly with PX4. And this message is not avialible there. 

Is possible to move RPM message from Ardupilot to Common msg deffinitios? And add index value? We can use RPM1 as indicated and RPM2 as estimated value. So, it is useful for us. 

Thanks.

roman-dvorak:
@auturgy, after thinking of all these options, I'd rather add this new message. After all, it is different from existing Ardupilot, if it will be possible.

hamishwillee:
@auturgy @roman-dvorak  To me this hinges a bit on what is the  "estimated" value for?
- You shouldn't use the ardupilot message's second sensor for this. The fields have a standardised meaning - if you overload something a GCS won't work with it in a standard way.
- If this is needed you can't just rely on the ESC rpm.

I don't think the timestamp is needed.

Just a thought, but would it be better/is it possible for this use case to have a flag for "ready to take off" rather than sending the raw RPM? Or is this only something that a pilot can make use of knowing the details of the frame?

roman-dvorak:
For ```estimated_accurancy``` I have no use at GCS right now. But it is value obtained from sensor.

Ok, timestamp can be omitted, estimated value also. I will fix it in next commit. 

 > Just a thought, but would it be better/is it possible for this use case to have a flag for "ready to take off" rather than sending the raw RPM? Or is this only something that a pilot can make use of knowing the details of the frame?

No, it is not better. Because another necessary use is knowledge of RPM in autorotation mode. There is needed to know RPM to correct rotor position adjustment.

In the future, we may need such things as well. But this would help us very much for development right now.

auturgy:
I’m not convinced anything other than the measured rpm value is used, generated, or needed, but an instance/index could be useful if for some reason one of the other messages that already provides that isn’t appropriate.
I’ll jump out of this conversation now - I can’t see this getting to a point I’ll sign off on it.

Regards,

James

roman-dvorak:
So, only the data source index and RPM value remained in the last commit.
#7304b22

message_definitions/v1.0/common.xml
```diff
@@ -6368,6 +6368,11 @@
       <field type="uint8_t" name="tx_session_pending">1: Transmission session pending, 0: No transmission session pending.</field>
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
+    <message id="339" name="RAW_RPM">
+      <description>RPM sensor data message.</description>
+      <field type="uint8_t" name="index">Index of this RPM sensor (0-indexed).</field>
+      <field type="float" name="indicated_frequency" units="rpm">Indicated rotor Frequency in Revolution per minute.</field>
```
hamishwillee, line 6374: description - please change to "Indicated rotor frequency".

roman-dvorak: Done in #7304b22714a6dc7fce0a0417a981656c33c7e109

hamishwillee:
With the current definition you could basically use the ArduPilot message. We could pull it across and modify it later if we needed to have more items. Or we could do it "properly" in the first place, at the cost of breaking compatibility. Which approach is best is a philosophical discussion.

I am going to think on this for a few days - feel free to ping me next week Tuesday if no direction provided.

roman-dvorak:
Hi, @hamishwillee, do you have any new ideas or solutions?

hamishwillee:
Thanks for the reminder. In the interests of widest compatibility I think you should use the ArduPilot RPM message [this works now you just have one sensor field]. Can you create a PR that removes it from ardupilotmega.xml and inserts it into common.xml at about the right position?

roman-dvorak:
Well, thank you. I created a new PR #1342

tridge:
> Can you create a PR that removes it from ardupilotmega.xml and inserts it into common.xml at about the right position?

I'd prefer the original is left alone, given he wants different semantics

hamishwillee:
Reopening pull request. As above, ^^^requested modifications to https://github.com/mavlink/mavlink/pull/1342 were not appropriate.

@roman-dvorak OK, this is open again. 

```
    <message id="339" name="RAW_RPM">
      <description>RPM sensor data message.</description>
      <field type="uint8_t" name="index">Index of this RPM sensor (0-indexed).</field>
      <field type="float" name="indicated_frequency" units="rpm">Indicated rotor frequency</field>
    </message>
```
Do you want to add back the estimated frequency field? 
- Can you update to clearly explain the difference between indicated and estimated?
- I think we can assume that every rpm system will output what you are calling "indicated" but not all of them will output esimated. We should make it clear what value means "this sensor does not output this value". So in this case you might have a sentence at the end "NaN indicates value is not supported/supplied by sensor."

hamishwillee:
@peterbarker @auturgy If there is anything else you might like in a new common message of this type, speak now, or forever hold your peace :-)

hamishwillee:
Also @roman-dvorak Note Peter's comment here: https://github.com/mavlink/mavlink/pull/1342#issuecomment-613206853
Regarding perhaps having additional info in an autogyro message. I'm not advocating - but worth considering now.

amilcarlucas:
Change the "Indicated rotor frequency" to "Indicated rotations per second" . Maybe it is not a rotor speed that we are measuring.

peterbarker:
Doesn't Roman need a quality measure?

Putting any fields which can be zero last (give or take mavlink wire packing...) may reduce mavlink2 packet sizes.  The generated code shows these sizes reasonably clearly IIRC.

I would note that we *stopped* using RPM in ArduPilot's Rover for relaying wheel speeds (using the wheel_encoder message instead).

We use RPM in ArduPilot for ICE work - but that should probably be a separate message.

So this message is probably in the stopgap-measure category, when you want to just spit data across....

An enumerated type field might not be a bad option.  But don't take that as anything but spitballing :-)

roman-dvorak:
Hi, Thank you all for your feedback and solving the problem.

> Can you update to clearly explain the difference between indicated and estimated?
Sure. The indicated value is RPM value measured from the sensor. And the estimated value is the accuracy of that value. 

> Doesn't Roman need a quality measure?
We do not need to know the accuracy of the value at QGC. Accuracy is logged. 

> Change the "Indicated rotor frequency" to "Indicated rotations per second" . Maybe it is not a rotor speed that we are measuring.
Yes it is true. I'll do it

> I think we can assume that every rpm system will output what you are calling "indicated" but not all of them will output esimated. We should make it clear what value means "this sensor does not output this value". So in this case you might have a sentence at the end "NaN indicates value is not supported/supplied by sensor."
Yes, it was originally meant. That this value can be none (unsupported/unprovided)

Now I do not know how to do it with the index of source. I'd like to add it there because there may be more RPM sources/sensors.

roman-dvorak:
Are there any other requirements?

message_definitions/v1.0/common.xml
```diff
@@ -6395,6 +6395,11 @@
       <field type="uint8_t" name="tx_session_pending">1: Transmission session pending, 0: No transmission session pending.</field>
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
+    <message id="339" name="RAW_RPM">
+      <description>RPM sensor data message.</description>
+      <field type="uint8_t" name="index">Index of this RPM sensor (0-indexed).</field>
+      <field type="float" name="frequency" units="rpm">Indicated rotations per second</field>
```
hamishwillee, line 6401: Well, it can't be a unit of rpm and in rotations per second as they are different. 
I assume this is actually repm, in which case you should just have description "Indicated rate".

hamishwillee:
I addd a comment regarding a bug, but otherwise it is good for me.

message_definitions/v1.0/common.xml
amilcarlucas: Did he meant "Indicated rotations per *minute*" ?

amilcarlucas:
This also needs to get git squashed berfore it gets merged

message_definitions/v1.0/common.xml
roman-dvorak: I meant RPM "rotation per minute".

roman-dvorak:
Ok, @amilcarlucas, @hamishwillee 

Commits were squashed into a single commit and description is changed to "Indicated rate". Units are RPM (rotations per minute)

PR #1341, opened 2020-04-07
Fix a typo in minimal.xml

bys1123:


hamishwillee:
Thanks. I am fairly certain that if this breaks anything it will obvious at compile time. So merging.

PR #1342, opened 2020-04-08
Move RPM message from ardupilotmega.xml to common.xml 

roman-dvorak:
based on PR #1337 I have created this PR. The content of this issue is to move RPM message from ardupilotmega definition to common for widest compatibility. 

Into RPM message was also added index field and data units. 

I'm not sure if I can change message ID (and to which id to change it) and with the placement of message definitions in common messages.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5708,11 +5708,13 @@
       <extensions/>
       <field type="uint8_t" name="breach_mitigation" enum="FENCE_MITIGATE">Active action to prevent fence breach</field>
     </message>
+    <!-- imported from ardupilotmega.xml (2020) -->
     <message id="226" name="RPM">
       <description>RPM sensor output.</description>
+      <field type="float" name="rpm1" units="rpm">RPM values 1</field>
+      <field type="float" name="rpm2" units="rpm">RPM values 2</field>
+      <extensions/>
       <field type="uint8_t" name="index">Sensor ID</field>
-      <field type="float" name="rpm1" units="rpm">RPM Sensor1.</field>
-      <field type="float" name="rpm2" units="rpm">RPM Sensor2.</field>
     </message>
     <!-- MESSAGE IDs 180 - 229: Space for custom messages in individual projectname_messages.xml files -->
     <message id="230" name="ESTIMATOR_STATUS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1943,6 +1943,11 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <message id="226" name="RPM">
+        <description>RPM sensor output.</description>
+        <field type="float" name="rpm1">RPM Sensor1.</field>
+        <field type="float" name="rpm2">RPM Sensor2.</field>
+      </message>
       <entry value="240" name="MAV_CMD_DO_LAST" hasLocation="false" isDestination="false">
         <description>NOP - This command is only used to mark the upper limit of the DO commands in the enumeration</description>
         <param index="1">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5710,8 +5710,9 @@
     </message>
     <message id="226" name="RPM">
       <description>RPM sensor output.</description>
-      <field type="float" name="rpm1">RPM Sensor1.</field>
-      <field type="float" name="rpm2">RPM Sensor2.</field>
+      <field type="uint8_t" name="index">Sensor ID</field>
+      <field type="float" name="rpm1" units="rpm">RPM Sensor1.</field>
+      <field type="float" name="rpm2" units="rpm">RPM Sensor2.</field>
     </message>
     <!-- MESSAGE IDs 180 - 229: Space for custom messages in individual projectname_messages.xml files -->
     <message id="230" name="ESTIMATOR_STATUS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1943,11 +1943,6 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <message id="226" name="RPM">
-        <description>RPM sensor output.</description>
-        <field type="float" name="rpm1">RPM Sensor1.</field>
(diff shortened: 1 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5708,6 +5708,12 @@
       <extensions/>
       <field type="uint8_t" name="breach_mitigation" enum="FENCE_MITIGATE">Active action to prevent fence breach</field>
     </message>
+    <message id="226" name="RPM">
+      <description>RPM sensor output.</description>
+      <field type="uint8_t" name="index">Sensor ID</field>
```
hamishwillee, line 5713: Please remove the index. We'll do further updates as MAVLink 2 extensions if needed.

hamishwillee:
@auturgy @peterbarker @julianoes FYI moving the RMP message from ardupilotmega.xml to common.xml. 

The id is in the mavlink 1 range in the section that is reserved for any autopilot. So in theory another autopilot could be using this ID for something else and we'll be creating a clash. However I consider this is a relatively low risk, as it is not used in any of the dialect files in our repo.

HOw big an impact would it be to move it into the MAVLink 2 range? I'm guessing too high?

message_definitions/v1.0/common.xml
roman-dvorak: Set as mavlink2 value

peterbarker:
Two RPM values and a sensor ID.  That's... interesting.

You probably wouldn't have two RPM values in hindsight, of course - I am curious about how the "sensor-id" thing works

roman-dvorak:
@peterbarker yes, the second value is not well useful for us. Therefore the original definition was like this:
```
    <message id="339" name="RAW_RPM">
      <description>RPM sensor data message.</description>
      <field type="uint_8" name="index">Index of this RPM sensor (0-indexed).</field>
      <field type="float" name="indicated_frequency" units="rpm">Indicated rotor Frequency in Revolution per minute.</field>
      <field type="float" name="estimated_accurancy" units="rpm">Estimated accurancy in Revolution per minute. NaN indicates this field not supported.</field>
    </message>
```

Because our RPM sensor produces RPM value and estimated accuracy. The index value will be useful for planned support of multiple RPM sensors.

peterbarker:
> @peterbarker yes, the second value is not well useful for us. Therefore the original definition was like this:
> 
> ```
>     <message id="339" name="RAW_RPM">
>       <description>RPM sensor data message.</description>
>       <field type="uint_8" name="index">Index of this RPM sensor (0-indexed).</field>
>       <field type="float" name="indicated_frequency" units="rpm">Indicated rotor Frequency in Revolution per minute.</field>
>       <field type="float" name="estimated_accurancy" units="rpm">Estimated accurancy in Revolution per minute. NaN indicates this field not supported.</field>
>     </message>
> ```
> 
> Because our RPM sensor produces RPM value and estimated accuracy. The index value will be useful for planned support of multiple RPM sensors.

So you're using rpm1 for the value and rpm2 for an uncertainty?  That's a little contrary to intent and usage of that message in ArduPilot, as you might imagine :-)

It might actually be better to put together a separate message with instance, value, covariance and whatever in it.

I'll make sure this comes up on the ArduPilot devcall to see what people thing.

roman-dvorak:
Thank you! 

Yes, I have already suggested this use. RPM1 as value, and RPM2 as covariance. And @hamishwillee wrote on tahat:
> - You shouldn't use the ardupilot message's second sensor for this. The fields have a standardised meaning - if you overload something a GCS won't work with it in a standard way.

_ in https://github.com/mavlink/mavlink/pull/1337#issuecomment-608136391_

I also think that the message should be separate

hamishwillee:
@roman-dvorak At the end of discussion https://github.com/mavlink/mavlink/pull/1337 you had essentially dropped everything from the proposed message except for a single sensor value (e.g. no estimated accuracy). Given that I change I proposed bringing the ArduPilot message across and evolving it *compatibly* in future, if needed. That made sense because you could do that with the ArduPilot message.

For example, a compatible extension might be to add a (new) extension field that would allow you to specify (e.g. by setting value to 1) that the second rpm value was to be interpretted as the covariance.

@peterbarker See the discussion in #1337 . I would prefer a new message because the current ArduPilot message isn't particularly well designed. @auturgy thought this could reasonably be done with the old message. Looking forward to hearing what the devcall think. If ArduPilot can afford the break, then I'd like to do it.

peterbarker:
On Mon, 13 Apr 2020, Hamish Willee wrote:

> For example, a compatible extension might be to add a (new) extension field that would allow you to specify (e.g. by setting value to 1) that the second
> rpm value was to be interpretted as the covariance.

Better to just add two quality values as extensions, surely?!

> @peterbarker See the discussion in #1337 . I would prefer a new message because the current ArduPilot message isn't particularly well designed. @auturgy
> thought this could reasonably be done with the old message. Looking forward to hearing what the devcall think. If ArduPilot can afford the break, then I'd
> like to do it.

Yeah, sadly, it we didn't end up discussing it.  I remembered and 
mentioned part-way through the call but it didn't make the agenda (despite 
me asking), and I forget at the end of the call for the "off-the-floor" 
discussion.

It might make sense to spitball that better message again.

     <message id="339" name="RAW_RPM">
       <description>RPM sensor data message.</description>
       <field type="uint8_t" name="index">Index of this RPM sensor 
(0-indexed).</field>
       <field type="float" name="indicated_frequency" units="rpm">Indicated 
rotor frequency</field>
     </message>

Not sure about having a "quality" field on that message.  If it was an 
external sensor passing data in via mavlink, that would make sense (like 
the vision messages).  But the typical onboard sensors don't have that 
sort of thing on them.

(Today's devcall had us getting our RPM source from a vibe filter we're 
running, I should mention.  That would actually be relevant for your 
autogyro stuff.)

We should probably also discuss James' observation that having it as a 
field in a separate, dedicated message may also be an option.  So a 
dedicated "autogyro_rotor" message that contains all those extra fields 
mentioned (assuming you have sensors for them!) - bearing temperature, 
estimated RPM, actual RPM, whether the rotor is being actively spun by a 
startup motor, that sort of thing.

tridge:
please leave the existing ardupilot message alone and just add the new message in common.xml if you want it

hamishwillee:
@tridge Thanks very much. We certainly wouldn't have added it in a breaking way like this.

I'll reopen the other one an progress that.

PR #1343, opened 2020-04-13
Add message COPTER_DRIVE_STATUS

hsteinhaus:
Most hovering vehicles, including multicopters, helicopters and misc VTOLs and hybrid vehicles rely on drive thrust for attitude control and flight stability. A failed drive or a drive operating close to it's maximum (or even minimum) thrust limit, might seriously affect flight stability. But there might be recovery avaible if the operator knows about the situation early and is able to react and in a correct way, e.g. the operator could decide to enter an autorotation maneuver in case of a helicopter with a failed tail rotor drive. Therefore, this condition should be brought to the attention of the operator immediately, with enough detail information come to the correct conclusion.

Some very basic information for each drive should be enough for a sound decision. These are:
1. is the drive expected to be in use currently?
2. is it known to be failed?
3. is it performing at it's max power limit?
4. is it performing at it's min power limit?

Almost all combinations of failure could happen simultaneously, e.g. a hex-rotor with a drive known to be failed will probably max out two non-failed drives and throttle down to minimum at least one other drive, which is why a bit mask with one bit per drive is preferred to some a burst of per-drive messages in case of a single drive failure event.

To my best knowledge, there is no message for drive status in a vehicle with multiple/many drives in MAVLink currently. In the past, I have mad a hack for my vehicles in the flight controller specific error fields to make drive status information available at the GCS. The availability of this information has saved two vehicles so far, however the way of implementation is IMHO not the right one (other flight controllers might have the same problem) and has proven to be hard to maintain even in my fork of ArduPilot only.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6522,6 +6522,16 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
+    <message id="410" name="COPTER_DRIVE_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Drive status information, applicable to multicopters, helicopters and VTOL/hybrid vehicles. Failed drives or drives operates close to min or max power limit might affect maneuverability or flight stability and immediate operator attention is advised.</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint16_t" name="in_use" display="bitmask">Bitmap for drives that are in use in the current flight configuration and flight phase</field>
+      <field type="uint16_t" name="lower_limit" display="bitmask">Bitmap for drives that are running close to or at their minimum power limit.</field>
+      <field type="uint16_t" name="upper_limit" display="bitmask">Bitmap frives that are running close to or at their maxmimum power limit.</field>
+      <field type="uint16_t" name="failed" display="bitmask">Bitmap for drives which are known to be failed.</field>
+    </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
       <description>Cumulative distance traveled for each reported wheel.</description>


hamishwillee:
@dagar @auturgy Any thoughts on this? 

Things that spring to mind:
1. Is it also relevant to FW and other types?
2. Why timestamp needed?
3. Ever more than 16 motors?
4. Are there other things you might want to convey?
5. Would this be better to just send warning message for drives that are actually at the upper and lower limits? (ie a message with an index rather than a message with a bitmask?

hsteinhaus:
1. I think, the message and it's encoding is not specific to a particular autopilot implementation.
2. The message is intended to be send periodically, therefore the timestamp is important to correlate to a certain point in time. Additionally, the timestamp will be very helpful in post-mortem crash analysis (which drive failed first?).
3. Hard to tell. I hope 16 is good enough for the foreseeable future of 99.9% of all use cases. I could even live with 8 bits, however the demand for 16 is easy to understand for me (e.g. consider a VTOL plane with two regular props and an X8 config for vertical takeoff).
4. No, the scope of the message is very specific for the use case of periodic status monitoring of all drives. Any per-drive detail information would preferably be handled by one or more specialised messages.
5. I don't think so, as, at least in multi-rotors, common failures are very likely. Often all drives if the vehicle get warnings of different kind at exactly the same time (e.g. in the hexacopter example mentioned above). Therefore IMHO it makes no sense to flood the telemetry channel with a cascade of messages in the event of trouble, that might block the channel for other important information. like e.g status texts.

auturgy:
Is anything happening re the Events interface/microservice? (https://github.com/mavlink/rfcs/pull/14).  If so, imho better to roll this into it rather than a new standalone message.
Either way, I really need @lthall to have a look: he's doing the failure detection / recovery stuff for Copter.

hamishwillee:
@bkueng Any further work/planning on events interface? Might be a good candidiate for this. 

 @lthall Your thoughts appreciated. Ideally today so we can move this along.

lthall:
Hi all,

I will try to summarise my thoughts in very short points:

- The only field I see in the list above that is useful is "known to be failed".
- It is normal for motors to regularly hit maximum and minimum PWM outputs so this is not a good measure of the motors health unless the aircraft is never doing any form of dynamic flight or mildly aggressive maneuvering or flying in any sort of challenging conditions.
- "Motor in use" has very little utility in reality. It is needed only if the interpretation of other data relies on it. For example (motor in use) AND (motor at minimum) then BAD. Given that the only filed that I think is actually useful is "known to be failed" does not rely on "Motor in use" it is wasted data.
- A regular heart beat health message like this one is waisted bandwidth as we only care when something is wrong. The vast majority of time the power system will be completly healthy. During this time this data has no value.

I would suggest that a better approach would be to have messages that inform the user when there is a problem and is flexible to relevant information that we currently do not understand or appreciate. I would suggest a message that:
- The message is on a event base maybe repeated when problems exist
- The message is flexible enough to provide information form the ESC like, sync loss, restart, hover heat, power limited, low input voltage limited, loss of position sensor, FOC fallback to trapezoidal commutation. ect.
- A failure needs to be diagnosed on the flight controller and communicated to the ground station, we should not be expecting the ground station to diagnose a failure and attempting to send enough information for it to do that.

I hope that makes sence.

bkueng:
> @bkueng Any further work/planning on events interface? Might be a good candidiate for this.

It is a good candidate, but unfortunately not.

hsteinhaus:
Hi Leonard,

> It is normal for motors to regularly hit maximum and minimum PWM outputs so this is not a good measure of the motors health 
Couldn't disagree more. If my vehicle in AUTO mode while taking orhtophotos ever hits min or max motor limits, something is *seriously* wrong. Either I just push the vehicle to hard for the given weight and weather/wind speed-wise, or something is about to fail shortly. If I continue to push that hard for the next our or so, there is a very good chance that I will crash my vehicle.

So, at least for my use case, motors at limits (low or high) are a clear indication for *immediate* pilot action. At the same time, the limit saturation is the only indication (except may be motor temperature) that can warn the pilot about the trouble to be expected shortly. The ESC will turn the damaged motor with a partially broken winding wire or the dying bearing till it is to late and can no longer do so. The first ESC-originated error message will therefore come *after* the quadrotor has switched to ballistic flight mode. It has a completely different use case - post mortem analysis vs. early warning.

I cannot see, why the definition of a message could waste bandwith. If the user doesn't appreciate it, he doesn't have to enable it and it will consume zero bandwith. For my use, it would the third most important message after battery usage and radio link quality, in daily flying.

hsteinhaus:
After discussing the pros and cons in detail, is there any reason not to add the message? I don't see that there is any cost other than reserving a single enum value. On the other hand, It would help to prevent another nasty and incompatible fork of MAVLink.

I can assure, that there is a huge utility in reality in predictive motor diagnosis, and hopefully I am not the only one interested in that topic.

amilcarlucas:
Do not call it COPTER_DRIVE_STATUS. It should be VEHICLE_POWERTRAIN_STATUS

hamishwillee:
Closing as stale. 

There are arguments both for and against here - I can see how a message like this would be useful in some cases. You wouldn't want to stream it for all platforms (as per leonard's point) but for those where it would be useful you could request it. More platforms would be interested in notifications of problem cases/events.

Anyway, MAVLink is about defining standard interfaces. If we can't find enough parties in agreement it isn't standard.
This can be reopened if/when there is interest from multiple flight stacks.

PR #1349, opened 2020-04-21
Update Open Drone ID messages

friissoren:


auturgy:
@magicrub can you have a quick look

message_definitions/v1.0/common.xml
```diff
@@ -171,6 +171,9 @@
       <entry value="33" name="MAV_TYPE_SERVO">
         <description>Servo</description>
       </entry>
+      <entry value="34" name="MAV_TYPE_ODID">
+        <description>Open Drone ID. See https://mavlink.io/en/services/opendroneid.htm.</description>
```
hamishwillee, line 175: Perhaps Open Droid ID > Open Droid Tx/Rx module?

hamishwillee:
FWIW Thanks @friissoren. I think this does the job - essentially you can route to up to at least three different endpoints now over any network configuration using normal mavlink ruls. 

@magicrub I am happy with this, but I want to know that you are basically OK with it before pressing the merge button.

message_definitions/v1.0/common.xml
friissoren: Also the MAV_COMP_ID_ODID_USER_INPUT can generate these messages. I think it would be better to keep the type generic.

hamishwillee: Yes, that would be better. I'm not sure we need MAV_COMP_ID_ODID_USER_INPUT  though. When would we ever use it?

friissoren: For inputting the Operator ID, at the very least. Also the fields related to Area Count/Radius/Ceiling/Floor are likely to be input by the operator.
And there is the Self-ID field, which is just a plain text string that the operator an use to give general information about the purpose of the flight.

hamishwillee: @friissoren Sorry, just got back to this.

I don't see why you need this? Component IDs are for routing and handling - "I want this message to go to this component" or "I want to know that this message came from this component so I can decide to handle it appropriately"

So imagine a user enters the data into GCS. The message is then sent to the MAV_COMP_ID_ODID_TXRX_1. The MAV_COMP_ID_ODID_TXRX_1. That transmitter doesn't care about the origin being a specific component ID I don't think. If it does, then it probably would check on the GCS component ID.
The transmitter never sends info specifically to a MAV_COMP_ID_ODID_USER_INPUT - it sends it to a system or GCS.

Does that make sense?

hamishwillee: Other than that I'm OK with this PR.

friissoren: There is no GCS component ID (that I have found at least). If there is, please point me to it.
It must be possible for the end-user somehow to input information to the drone ID system, hence the need for this user input component.

hamishwillee: Hi Soren, 
a GCS has its own system ID, and you can identify that you're talking to a GCS using [MAV_TYPE_GCS](https://mavlink.io/en/messages/common.html#MAV_TYPE_GCS). Typically the component ID for a mission planner is [MAV_COMP_ID_MISSIONPLANNER](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_MISSIONPLANNER) but you don't need it - because you'd address a GCS via its system id after identifying its type.

> It must be possible for the end-user somehow to input information to the drone ID system, hence the need for this user input component.

Yes the user needs to be able to input information to the drone ID system, but IMO this does not imply the need for this user input component id. 

Consider the case with a GCS and Vehicle (with two drone id transmitters) on a MAVLink network, and the GCS sends the message with user input. The transmitter gets this message - why would it need to check that this is from a user input component? It won't be sent by another transmitter, and if it is, you could rule that out by saying "filtering on not of type transmitter". It can know the message is from a GCS based on the MAV_TYPE of the GCS system.

The only reason it needs a user input component is if you had multiple user input components on a sending system and wanted to preferentially chose to obey one and not the other. 

That is not to say that there aren't problems here :-(. Unfortunately again, it is hard to sure we have covered all bases without a set of diagrams showing all the possible message flows to-from transmitters, vehicles and GCSs, and verifying that in all cases the intended components (only) proces the messages.

For example, consider the case above where instead of just one GCS and vehicle you now have two vehicles on the network. If the GCS sends the message it is broadcast - which vehicle should get the message? Right now, both will. Which is why messages that sometimes need to go to a particular vehicle must have a target system id.

So I think you need to look at the messages and decide whether any can cause problems if received by an inappropriate system or component. If so, then these will need target ids.

Hope that makes sense.

friissoren: Thanks for the explanation and sorry for me taking to long to grasp how this should work. Somehow I hadn't understood that the GCS and the UA are two different systems with different system IDs.

I modified both the documentation and the message specification. Please let me know if this now makes more sense.

There is one thing I still don't understand. The GCS has it's own system id which if fills to the sysid field. What does the GCS put in the compid field when sending a message?

friissoren:
@magicrub: If you have time, please take a look at this one also. Thanks. It is linked to https://github.com/mavlink/mavlink-devguide/pull/227

message_definitions/v1.0/common.xml
hamishwillee: Hi @friissoren 
You're welcome. No worries. Unfortunately we have docs that explain how a mavlink network works, but not framed as "how you then need to think about message design". I'll write something like that at some point; I haven't done so because this is all abstract to me - I have never had to implement a system as you are doing.

> I modified both the documentation and the message specification. Please let me know if this now makes more sense.

I won't have time this week sorry. Ping me again next Wednesday if I have not done so.

> The GCS has it's own system ID which it fills to the sysid field. What does the GCS put in the compid field when sending a message?

To my knowledge, the existing GCS use [MAV_COMP_ID_MISSIONPLANNER](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_MISSIONPLANNER). I don't think it particularly matters though. What does matter (and what you should filter on for identity) is that is uses [MAV_TYPE_GCS](https://mavlink.io/en/messages/common.html#MAV_TYPE_GCS)

friissoren: Just take your time with the review. We will get there.

I think the MAV_TYPE_GCS is the main source of my confusion. You can only use that in the HEARTBEAT message. It is not used in any other message and you cannot use it either for filling the systemid or compid fields.

This then means that the transceiver components need to first listen for a HEARTBEAT message with type set to MAV_TYPE_GCS. From that message, they pick up the systemid of the GCS (I guess the compid can actually be ignored here). After that, they can then start listening to ODID messages with that same systemid, since they implicitly know that it is the GCS that has sent them.

If this is correct, I will add something about that to the documentation.

friissoren:
I can move those to the beginning. However, what is this timestamp you are referring to?

message_definitions/v1.0/common.xml
```diff
@@ -4133,49 +4145,49 @@
...
       <entry value="14" name="MAV_ODID_TIME_ACC_1_4_SECOND">
-        <description>The timestamp accuracy is smaller than 1.4 second.</description>
+        <description>The timestamp accuracy is smaller than or equal to 1.4 second.</description>
       </entry>
       <entry value="15" name="MAV_ODID_TIME_ACC_1_5_SECOND">
-        <description>The timestamp accuracy is smaller than 1.5 second.</description>
+        <description>The timestamp accuracy is smaller than or equal to 1.5 second.</description>
       </entry>
     </enum>
     <enum name="MAV_ODID_AUTH_TYPE">
```
amilcarlucas, line 4193: This starts with a "<enum name="MAV_ODID_AUTH_TYPE">" and ends with a "</message>" something is wrong here

message_definitions/v1.0/common.xml
```diff
@@ -6698,6 +6713,9 @@
       <field type="uint8_t" name="speed_accuracy" enum="MAV_ODID_SPEED_ACC">The accuracy of the horizontal and vertical speed.</field>
       <field type="float" name="timestamp" units="s">Seconds after the full hour with reference to UTC time. Typically the GPS outputs a time-of-week value in milliseconds. First convert that to UTC and then convert for this field using ((float) (time_week_ms % (60*60*1000))) / 1000.</field>
```
amilcarlucas, line 6714: This is the timestamp that I am referring to

message_definitions/v1.0/common.xml
```diff
@@ -6698,6 +6713,9 @@
       <field type="uint8_t" name="speed_accuracy" enum="MAV_ODID_SPEED_ACC">The accuracy of the horizontal and vertical speed.</field>
```
amilcarlucas, line 6713: Please add a message description text.

message_definitions/v1.0/common.xml
```diff
@@ -6709,13 +6727,19 @@
       <field type="uint8_t" name="length" units="bytes">This field is only present for page 0. Total bytes of authentication_data from all data pages. Allowed range is 0 - 109 (17 + 23*4).</field>
```
amilcarlucas, line 6727: Please add a message description text

message_definitions/v1.0/common.xml
```diff
@@ -6709,13 +6727,19 @@
...
       <field type="uint32_t" name="timestamp" units="s">This field is only present for page 0. 32 bit Unix Timestamp in seconds since 00:00:00 01/01/2019.</field>
```
amilcarlucas, line 6728: This is the timestamp that I am referring to.

message_definitions/v1.0/common.xml
```diff
@@ -6728,21 +6752,29 @@
       <field type="uint16_t" name="area_radius" units="m">Radius of the cylindrical area of the group or formation (default 0).</field>
```
amilcarlucas, line 6752: Please add a description text

message_definitions/v1.0/common.xml
friissoren: Sorry, you lost me here. What are you trying to say?

message_definitions/v1.0/common.xml
friissoren: Ahh, ok. The timestamp is specific to the location message. There is no need to have that for the other messages.
I will update the PR to have the target_system and target_component at the beginning of the messages.

message_definitions/v1.0/common.xml
friissoren: There is already "The accuracy of the horizontal and vertical speed."
And the Location message has already a description field at the beginning. I guess these are enough?

message_definitions/v1.0/common.xml
friissoren: Sorry, yes. There was a second timestamp field. But that one is specific to the authentication message and should not be generalized to all ODID messages.

message_definitions/v1.0/common.xml
friissoren: It is already there?

message_definitions/v1.0/common.xml
friissoren: Again already there?

hamishwillee:
@amilcarlucas FWIW I am pretty sure that "target_system" and "target_component" do not have to be at the beginning of the message. That said, it is better/more consistent if they are. They just have to be named correctly though.

amilcarlucas:
Yes, I mentioned that for consistency reasons. If Mavlink wants to be a clean standard, it needs to obey all sorts of consistency rules, the messages need to have a consistent "look-and-feel" :)

amilcarlucas:
@friissoren It looks like you removed some of the original PR changes. Was that intentional?

friissoren:
I am all for consistency. I moved the target_xxx to the beginning of the messages.

There are still three commits in the PR. I only needed to force push two of them, so the first (well, third one) stayed untouched.

BTW: Maybe force pushing isn't a good way of operating. I don't seem to be able to open your earlier comments anymore. I should probably have made a new commit instead. Seems that Hamish anyway squash them all before committing to the main repository.

I still don't get your comments about descriptions. All of the messages have a description field filled. Maybe it doesn't show properly if you look at the commits in the diff view? See here and forward:
https://mavlink.io/en/messages/common.html#OPEN_DRONE_ID_BASIC_ID

friissoren:
@hamishwillee @magicrub @amilcarlucas : Do you think there will be any further changes needed on this PR and the related https://github.com/mavlink/mavlink-devguide/pull/227 PR?

Both parts are still under WIP markings so there is time to fix things if needed.

I would like to get these current changes merged, before I add a few new fields that have been added in the compatible ASD-STAN Direct Remote ID standard for Europe.

message_definitions/v1.0/common.xml
Jaeyoung-Lim, line 175: I guess `https://mavlink.io/en/services/opendroneid.htm.` is a typo. Should be `https://mavlink.io/en/services/opendroneid.html` ?

friissoren: Thank you. Good catch. I will fix this.

hamishwillee:
@amilcarlucas can you please respond to https://github.com/mavlink/mavlink/pull/1349#issuecomment-634547611

I have no problem with the changes, but would be great if @magicrub  could comment. 
Let's time-box that. If he has not commented by next week Tuesday, please ping me and mege as is. As you say, it is WIP and it is an improvement.

magicrub:
hi all, sorry for hiding! I'll take a good hard look at this whole thing tonight. I SWEAR!!!!!

hamishwillee:
> hi all, sorry for hiding! I'll take a good hard look at this whole thing tonight. I SWEAR!!!!!

... on my dog's life?

amilcarlucas:
@friissoren ignore my coment about descriptions. It seams to have been caused by github presenting me a bad diff view and leading me astray.

amilcarlucas:
This can be merged once @magicrub approves it.

magicrub:
APPROVED! This looks good. Sorry for the late review. I have a few comments on the documents PR but this one looks as-is

magicrub:
needs a rebase

hamishwillee:
Thanks everyone. Merging!

PR #1352, opened 2020-04-23
common: added generator status message

CarlEngel:
I want to add a message for doing telemetry messages used for on board electrical generators such as alternators, generators, fuel cells, solar cells.

I tagged it WIP for now, not sure if this is the right thing to do.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3328,33 +3328,33 @@
         <description>Generator is charging the batteries (generating enough power to charge and provide the load).</description>
       </entry>
       <entry value="16" name="GENERATOR_STATUS_FLAG_REDUCED_POWER">
-        <description>Generator is providing the maximum output.</description>
-      </entry>
-      <entry value="32" name="GENERATOR_STATUS_FLAG_MAXPOWER">
         <description>Generator is operating at a reduced maximum power.</description>
       </entry>
+      <entry value="32" name="GENERATOR_STATUS_FLAG_MAXPOWER">
+        <description>Generator is providing the maximum output.</description>
+      </entry>
       <entry value="64" name="GENERATOR_STATUS_FLAG_OVERTEMP_WARNING">
-        <description>Generator is near the maximum operating temperature, cooling is insufficient (may be running at reduced power).</description>
+        <description>Generator is near the maximum operating temperature, cooling is insufficient.</description>
       </entry>
       <entry value="128" name="GENERATOR_STATUS_FLAG_OVERTEMP_FAULT">
         <description>Generator hit the maximum operating temperature and shutdown.</description>
       </entry>
-      <entry value="256" name="GENERATOR_STATUS_FLAG_RECTIFIER_OVERTEMP_WARNING">
-        <description>Power electronics are near the maximum operating temperature, cooling is insufficient (may be running at reduced power).</description>
+      <entry value="256" name="GENERATOR_STATUS_FLAG_ELECTRONICS_OVERTEMP_WARNING">
+        <description>Power electronics are near the maximum operating temperature, cooling is insufficient.</description>
       </entry>
-      <entry value="512" name="GENERATOR_STATUS_FLAG_RECTIFIER_OVERTEMP_WARNING">
+      <entry value="512" name="GENERATOR_STATUS_FLAG_ELECTRONICS_OVERTEMP_FAULT">
         <description>Power electronics hit the maximum operating temperature and shutdown.</description>
       </entry>
-      <entry value="1024" name="GENERATOR_STATUS_FLAG_RECTIFIER_FAULT">
-        <description>Power electronics experienced an error.</description>
+      <entry value="1024" name="GENERATOR_STATUS_FLAG_ELECTRONICS_FAULT">
+        <description>Power electronics experienced an fault and shutdown.</description>
       </entry>
       <entry value="2048" name="GENERATOR_STATUS_FLAG_POWERSOURCE_FAULT">
-        <description>The power source providing the generator failed e.g. mechanical generator stopped, tether is no longer providing power, solar cell is in shade, hydrogen reaction no longer happening.</description>
+        <description>The power source supplying the generator failed e.g. mechanical generator stopped, tether is no longer providing power, solar cell is in shade, hydrogen reaction no longer happening.</description>
       </entry>
-      <entry value="4096" name="GENERATOR_STATUS_FLAG_COMMUNICATION_ERROR">
+      <entry value="4096" name="GENERATOR_STATUS_FLAG_COMMUNICATION_WARNING">
         <description>Generator controller having communication problems.</description>
       </entry>
-      <entry value="8192" name="GENERATOR_STATUS_FLAG_COOLING_ERROR">
+      <entry value="8192" name="GENERATOR_STATUS_FLAG_COOLING_WARNING">
         <description>Power electronic or generator cooling system error.</description>
       </entry>
       <entry value="16384" name="GENERATOR_STATUS_FLAG_POWER_RAIL_FAULT">
@@ -6573,15 +6573,15 @@
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Telemetry of power generation system. Alternator or mechanical generator.</description>
-      <field type="uint64_t" name="status" display="bitmask" enum="GENERATOR_STATUS_FLAG">Status message</field>
-      <field type="uint16_t" name="generator_speed" units="rpm">Speed of electrical generator or alternator. 0: field not provided.</field>
-      <field type="float" name="battery_current" units="A">Current into/out of battery. Positive for out. Negative for in. 0: field not provided.</field>
-      <field type="float" name="load_current" units="A">Current going to the UAV. If battery current not available this is the DC current from the generator. Positive for out. Negative for in.</field>
-      <field type="float" name="power_generated" units="W">The power being generated.</field>
-      <field type="float" name="bus_voltage" units="V">Voltage of the bus seen at the generator or battery bus if battery bus is controlled by generator and at a different voltage to main bus.</field>
-      <field type="int16_t" name="rectifier_temperature" units="degC">The temperature of the rectifier or power converter.</field>
-      <field type="float" name="bat_current_setpoint" units="A">The target battery current. Positive for out. Negative for in. -10000: field not provided</field>
-      <field type="int16_t" name="generator_temperature" units="degC">The temperature of the mechanical motor, fuel cell core or generator. UINT16_MAX: field not provided.</field>
+      <field type="uint64_t" name="status" display="bitmask" enum="GENERATOR_STATUS_FLAG">Status flags.</field>
+      <field type="uint16_t" name="generator_speed" units="rpm">Speed of electrical generator or alternator. UINT16_MAX: field not provided.</field>
+      <field type="float" name="battery_current" units="A">Current into/out of battery. Positive for out. Negative for in. NaN: field not provided.</field>
+      <field type="float" name="load_current" units="A">Current going to the UAV. If battery current not available this is the DC current from the generator. Positive for out. Negative for in. NaN: field not provided</field>
+      <field type="float" name="power_generated" units="W">The power being generated. NaN: field not provided</field>
+      <field type="float" name="bus_voltage" units="V">Voltage of the bus seen at the generator, or battery bus if battery bus is controlled by generator and at a different voltage to main bus.</field>
+      <field type="int16_t" name="rectifier_temperature" units="degC">The temperature of the rectifier or power converter. INT16_MAX: field not provided.</field>
+      <field type="float" name="bat_current_setpoint" units="A">The target battery current. Positive for out. Negative for in. NaN: field not provided</field>
+      <field type="int16_t" name="generator_temperature" units="degC">The temperature of the mechanical motor, fuel cell core or generator. INT16_MAX: field not provided.</field>
     </message>
     <message id="375" name="ACTUATOR_OUTPUT_STATUS">
       <description>The raw values of the actuator outputs (e.g. on Pixhawk, from MAIN, AUX ports). This message supersedes SERVO_OUTPUT_RAW.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6451,14 +6451,14 @@
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Telemetry of power generation system. Alternator or mechanical generator.</description>
       <field type="uint64_t" name="status">Status message</field>
-      <field type="uint16_t" name="generator_speed" units="rpm">Speed of electrical generator or alternator. Set to 0 if non rotating generator such as fuel cell.</field>
-      <field type="float" name="battery_current" units="A">Current into/out of battery. Positive for out. Negative for in.</field>
-      <field type="float" name="load_current" units="A">Current going to the UAV.</field>
+      <field type="uint16_t" name="generator_speed" units="rpm">Speed of electrical generator or alternator. 0: field not provided.</field>
+      <field type="float" name="battery_current" units="A">Current into/out of battery. Positive for out. Negative for in. 0: field not provided.</field>
+      <field type="float" name="load_current" units="A">Current going to the UAV. If battery current not available this is the DC current from the generator. Positive for out. Negative for in.</field>
       <field type="float" name="power_generated" units="W">The power being generated.</field>
       <field type="uint16_t" name="bus_voltage" units="mV">Voltage of the bus seen at the generator.</field>
       <field type="int16_t" name="rectifier_temperature" units="degC">The temperature of the rectifier or power converter.</field>
-      <field type="float" name="bat_current_setpoint" units="A">The target battery current. Positive for out. Negative for in.</field>
-      <field type="int16_t" name="generator_temperature" units="degC">The temperature of the mechanical motor, fuel cell core or generator.</field>
+      <field type="float" name="bat_current_setpoint" units="A">The target battery current. Positive for out. Negative for in. -10000: field not provided</field>
+      <field type="int16_t" name="generator_temperature" units="degC">The temperature of the mechanical motor, fuel cell core or generator. UINT16_MAX: field not provided.</field>
     </message>
     <message id="375" name="ACTUATOR_OUTPUT_STATUS">
       <description>The raw values of the actuator outputs (e.g. on Pixhawk, from MAIN, AUX ports). This message supersedes SERVO_OUTPUT_RAW.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3259,8 +3259,6 @@
     </enum>
     <enum name="GENERATOR_STATUS_FLAG" bitmask="true">
       <description>Flags to report status/failure case for a power generator (used in GENERATOR_STATUS).</description>
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6446,20 @@
       <field type="uint16_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
+    <message id="373" name="GENERATOR_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Telemetry of power generation system. Alternator or mechanical generator.</description>
+      <field type="uint64_t" name="status">Status message</field>
```
hamishwillee, line 6453: Is intention that this is a string or a bitmask or something? Normally status information is conveyed by STATUSTEXT for "common UI handling" but cases like this will be better handled in future [Events Interface](https://docs.google.com/document/d/18qdDgfML97lItom09MJhngYnFzAm1zFdmlCKG7TaBpg/edit).

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6446,20 @@
...
+      <field type="uint16_t" name="generator_speed" units="rpm">Speed of electrical generator or alternator. Set to 0 if non rotating generator such as fuel cell.</field>
+      <field type="float" name="battery_current" units="A">Current into/out of battery. Positive for out. Negative for in.</field>
```
hamishwillee, line 6455: Are any of these fields "optional" - ie is a component always _required_ to provide all fields? If not, you need to explain how "not supported" is indicated.

hamishwillee:
Hi @CarlEngel 

We tend to tag WIP when we're not sure that the message is ideal, and remove as soon as there is a working implementation. So this is good.

I don't know anything about power generators. Generally though first thing we try do is see:
- if there is any message in the other main dialects that do this already (does not appear to be)
- check whether the message is sufficiently generalised for the particular use case - e.g. are there generators that might reasonably export other fields so this can be used with all similar generators.
- Further check whether the message can be usefully extended for the broader use case of power supplies. 

Since I know little about this invite some people to comment: @auturgy @LorenzMeier @dagar - have you opinions on this/anyone else with expertise we can ask?

auturgy:
We haven’t really needed a bespoke message, as ArduPilot already has messages that provide key data (multiple battery monitors, RPM, etc).
I’d support this going in, but suggest it worth considering what needs to be sent to the ground station vs logged onboard: if there are fields not used in flight, they could be removed to shrink the message (but still logged onboard).

message_definitions/v1.0/common.xml
CarlEngel: Is it okay to flag it as a enum, but not provide actual enum values. More intended for it just to be logged. The idea was to keep it generic so that different vendors can provide different statuses depending on the type of their generator. But if there is a better way to do this then I would be open to it.

message_definitions/v1.0/common.xml
CarlEngel: I added values for fields that I would see as not being required.

CarlEngel:
@hamishwillee If there are any other fields that can be added in order to make the message more generalized I would be happy to add them. These are the ones I have come up that I see as being useful for most cases and ones that I generally want to see when operating a vehicle.

@auturgy I felt that it would be helpful to have a standard message for different types of generators to be grouped. I found it useful to have all the data in one place while operating a drone to monitor the health of the generator without too much work. It might also allow for future development with on board generators. Are there any fields that you feel are not necessary?

message_definitions/v1.0/common.xml
julianoes: I'm also a bit confused about this one.

CarlEngel: Maybe I am naming it poorly, but I want it to be a status of the generator controller. After looking more, the end goal would be to have it as a bitmask enum. But in order to keep it general for different types of generators it is hard to know what type of enum values to give it yet. Some examples of this would be a flag for `high_temperature_warning` (for if the generator is getting close to its max operating temperature and the operator can know there is insufficient cooling) or `overtemp_shutdown_error` (for if generation stopped for an overtemp). An example message I was using was `failure_flags` in the `GIMBAL_DEVICE_ATTITUDE_STATUS` message.

hamishwillee: @CarlEngel
That's cool. Then create a placeholder enum and add it into this. E.g. This would be much better if you could think of a real first status, but if not, name first flag as `GENERATOR_STATUS_FLAG_RESERVED`.
```
    <enum name="GENERATOR_STATUS_FLAG" bitmask="true">
      <description>Flags to report status/failure case for a power generator (used in GENERATOR_STATUS).</description>
      <entry value="1" name="GENERATOR_STATUS_FLAG_?">
        <description>?</description>
      </entry>
    </enum>
....
 <field type="uint64_t" name="status" enum=>Status message</field>
....
````
I don't really like that this can't be zero-byte truncated, but don't see easy solution.

**Comment by lthall** [comment]
Hi all, The biggest mistake I see made with these things is assumptions on range. In this case I am really happy to see voltages, currents using a float.

In the same way the error messages should be able to be extremely flexible to account for the fact we have very little appreciation of what messages will need to be communicated back to the user. The ability to generate a random text string to convey information relevant to the specific implementation is invaluable. This has been great for the autopilot for similar reasons. My understanding of the code above is limited but this functionality may be a given and I misunderstood the conversation above.

The other generally useful feature is a status bitmask that can be used to summarise the various system states. For a standard generator it may be ['Ready', 'Idling', 'Charging', 'Max Power', 'Reduced Power', 'Over Temp Engine', 'Over Temp Generator', 'Over Temp Rectifier', 'Engine Fault' ect....]. For more complex generators I could easily see 16 bits being appricated.

I am sorry if I have not specifically addressed the PR. My intention was to simply highlight some factors to consider and be translated to the message requirements where appropriate. As I am not familiar with the challenges and limitations of mavlink message designs I can't really say what should be done and therefore must trust others to include or discard my thoughts as appropriate.

Happy to talk or answer questions if anything I have said raises a flag!!

**Comment by CarlEngel** [comment]
> In the same way the error messages should be able to be extremely flexible to account for the fact we have very little appreciation of what messages will need to be communicated back to the user. The ability to generate a random text string to convey information relevant to the specific implementation is invaluable. This has been great for the autopilot for similar reasons. My understanding of the code above is limited but this functionality may be a given and I misunderstood the conversation above.

For this functionality, the `STATUSTEXT ` message would still be used.

> The other generally useful feature is a status bitmask that can be used to summarise the various system states. For a standard generator it may be ['Ready', 'Idling', 'Charging', 'Max Power', 'Reduced Power', 'Over Temp Engine', 'Over Temp Generator', 'Over Temp Rectifier', 'Engine Fault' ect....]. For more complex generators I could easily see 16 bits being appreciated.

You hit the nail on the head with my intention of the `status` message. These are the kind of statuses I was expecting to use. I will do what @hamishwillee suggested and place these into a enum as well as a few other I had in mind and update the pr tomorrow.

**Comment by hamishwillee** [comment]
Thanks @lthall - awesome. Some points (which we discussed offline). 
- It might be worth splitting out state and fault status
- Carl proposed free text for the reasons you outlined. I prefer status enums as they are precise, and their meeting can be used by a GCS. Also they are translatable. 
  - I appreciate that the system is not mature yet, so you will have messages that do not fit this model. My proposal.
    - Use an enum, define the set of things you need now.
    - Mark the enum and message WIP. We can then iterate this "for now". 
    - If there is a public notification that a user needs to know, for now use STATUSTEXT message.  In future we can capture some of these in a separate events interface that is under discussion. 
    - During development you can use messages like [DEBUG_VECT](https://mavlink.io/en/messages/common.html#DEBUG_VECT) that aren't displayed by the GCS to send your strings. QGC allows you to see these strings via the inspector. 


To clarify the status field discussion, this was original jus a buffer for free text and it was me that suggested making it an enum. The enum isn't as flexible as being able to send what you like, but it is pretty big.
The reason I hate text is that a GCS can do nothing with it except display it, and it is not amenable to translation. Whereas an enum has a precise meaning, that can be mapped to any text we like.

**Comment by hamishwillee** [comment]
@CarlEngel We crossed messages, but essentially "yes, I feel your pain, as does @lthall (Leonard). STATUSTEXT does what you need for now. FYI there is a proposal for a proper events interface that will do this much better (allow translation, and better management of what gets displayed etc).

@lthall Is happy to jump on a call with you/us to discuss the fields and statuses - if you want, can you share a skype id or other contact via PX4 slack with me? Or you guys can continue refining the fields here. FYI Leonard is in the same timezone as me - AEST.

**Comment by CarlEngel** [comment]
I added some preliminary statuses, I will be happy to have a call to discuss and change as needed.

I also added a wip to the enum, but this gives an error in the mavgen and also seems to cause the build to fail. So I am guessing that the wip tag is not supported for enums.

**Comment by hamishwillee** [comment]
>  So I am guessing that the wip tag is not supported for enums. 

No, but it should be. Yes, remove please when you can.

**Comment by CarlEngel** [comment]
@hamishwillee I have spoken lthall and he believes that we have covered most of the cases with the status message. We are going to leave it in wip for a while and start doing testing with it to see if there are any other situations that come up that are not included. He did have one comment that I added a change for.

The build is currently failing because the W unit is not included in the generator. I have updated it in the main repo, just the current version that is being used in this repo does not have it in yet.

**Comment by auturgy** [comment]
The Watts PR in pymavlink has been merged, so updating the submodule will get it.

File: message_definitions/v1.0/common.xml [hunk 3]
```diff
@@ -3257,6 +3257,63 @@
...
+        <description>Generator is ready to start generating power.</description>
+      </entry>
+      <entry value="4" name="GENERATOR_STATUS_FLAG_GENERATING">
+        <description>Generator is generating power.</description>
+      </entry>
+      <entry value="8" name="GENERATOR_STATUS_FLAG_CHARGING">
+        <description>Generator is charging the batteries (generating enough power to charge and provide the load).</description>
+      </entry>
+      <entry value="16" name="GENERATOR_STATUS_FLAG_REDUCED_POWER">
+        <description>Generator is providing the maximum output.</description>
```
hamishwillee, line 3298: This text does not appear to match the entry name?

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+      </entry>
+      <entry value="32" name="GENERATOR_STATUS_FLAG_MAXPOWER">
+        <description>Generator is operating at a reduced maximum power.</description>
```
hamishwillee, line 3301: Comment seems wrong.

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+      </entry>
+      <entry value="64" name="GENERATOR_STATUS_FLAG_OVERTEMP_WARNING">
+        <description>Generator is near the maximum operating temperature, cooling is insufficient (may be running at reduced power).</description>
```
hamishwillee, line 3304: Does "may be running at reduced power" act as a hint that GENERATOR_STATUS_FLAG_REDUCED_POWER may also be true? In that case you'd just set the other flag right?

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+      </entry>
+      <entry value="128" name="GENERATOR_STATUS_FLAG_OVERTEMP_FAULT">
+        <description>Generator hit the maximum operating temperature and shutdown.</description>
+      </entry>
+      <entry value="256" name="GENERATOR_STATUS_FLAG_RECTIFIER_OVERTEMP_WARNING">
+        <description>Power electronics are near the maximum operating temperature, cooling is insufficient (may be running at reduced power).</description>
```
hamishwillee, line 3310: As above. Probably show reduced power indication using the other flag. If comment here is useful though I have no problem. 
Should the flag be "RECTIFIER" specifically? I ask only because you say "Power electronics" which seems more general.

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+      </entry>
+      <entry value="512" name="GENERATOR_STATUS_FLAG_RECTIFIER_OVERTEMP_WARNING">
```
hamishwillee, line 3312: Same entry name as previous case.

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+        <description>Power electronics hit the maximum operating temperature and shutdown.</description>
+      </entry>
+      <entry value="1024" name="GENERATOR_STATUS_FLAG_RECTIFIER_FAULT">
+        <description>Power electronics experienced an error.</description>
```
hamishwillee, line 3316: Just FMI, does this always cause a shutdown too?

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+      </entry>
+      <entry value="2048" name="GENERATOR_STATUS_FLAG_POWERSOURCE_FAULT">
+        <description>The power source providing the generator failed e.g. mechanical generator stopped, tether is no longer providing power, solar cell is in shade, hydrogen reaction no longer happening.</description>
```
hamishwillee, line 3319: providing the generator > supplying the generator

message_definitions/v1.0/common.xml
```diff
@@ -3257,6 +3257,63 @@
...
+      </entry>
+      <entry value="4096" name="GENERATOR_STATUS_FLAG_COMMUNICATION_ERROR">
+        <description>Generator controller having communication problems.</description>
+      </entry>
+      <entry value="8192" name="GENERATOR_STATUS_FLAG_COOLING_ERROR">
+        <description>Power electronic or generator cooling system error.</description>
+      </entry>
+      <entry value="16384" name="GENERATOR_STATUS_FLAG_POWER_RAIL_FAULT">
+        <description>Generator controller power rail experienced a fault.</description>
```
hamishwillee, line 3328: FMI what is difference between a fault and an error? Are these information only, or do they require some sort of response?

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
       <field type="uint16_t" default="0" name="cell_offset">The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.</field>
       <field type="uint16_t[16]" name="voltages" units="mV">Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.</field>
     </message>
+    <message id="373" name="GENERATOR_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Telemetry of power generation system. Alternator or mechanical generator.</description>
+      <field type="uint64_t" name="status" display="bitmask" enum="GENERATOR_STATUS_FLAG">Status message</field>
```
hamishwillee, line 6540: Status message > Status flags/bitmap

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="uint16_t" name="generator_speed" units="rpm">Speed of electrical generator or alternator. 0: field not provided.</field>
```
hamishwillee, line 6541: Isn't zero a valid value? I'd change this to UINT16_MAX

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="float" name="battery_current" units="A">Current into/out of battery. Positive for out. Negative for in. 0: field not provided.</field>
```
hamishwillee, line 6542: This is a float, so change field not provided to NaN

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="float" name="load_current" units="A">Current going to the UAV. If battery current not available this is the DC current from the generator. Positive for out. Negative for in.</field>
```
hamishwillee, line 6543: If potentially optional/not supplied, specify NaN

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="float" name="power_generated" units="W">The power being generated.</field>
```
hamishwillee, line 6544: Always supplied? If can be optional specify NaN

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="float" name="bus_voltage" units="V">Voltage of the bus seen at the generator or battery bus if battery bus is controlled by generator and at a different voltage to main bus.</field>
```
hamishwillee, line 6545: 1. Add comma "generator, or battery"
2. Is this optional ? if so specify NaN for that.

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="int16_t" name="rectifier_temperature" units="degC">The temperature of the rectifier or power converter.</field>
```
hamishwillee, line 6546: Optional? Us a min value for that

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="float" name="bat_current_setpoint" units="A">The target battery current. Positive for out. Negative for in. -10000: field not provided</field>
```
hamishwillee, line 6547: Use NaN for "-10000: field not provided"

message_definitions/v1.0/common.xml
```diff
@@ -6446,6 +6503,20 @@
...
+      <field type="int16_t" name="generator_temperature" units="degC">The temperature of the mechanical motor, fuel cell core or generator. UINT16_MAX: field not provided.</field>
```
hamishwillee, line 6548: I THINK it needs to be INT16_MAX not UINT16_MAX

hamishwillee:
@CarlEngel Bunch of questions above for you :-)

message_definitions/v1.0/common.xml
CarlEngel: I removed the reduced power comment since the reduced power flag would be set.

message_definitions/v1.0/common.xml
CarlEngel: It means that the maximum power that it is able to produce has been decreased in software to allow it to run while preventing overheating. So for example with an engine the maximum throttle may be capped in order to reduce the current through the electronics. This should be obvious for anybody operating a generator. (i.e. a generator capable of outputting 2000W may now only be able to output 1500W maximum because it is approaching the thermal limit)

message_definitions/v1.0/common.xml
CarlEngel: I changed the wording to be more consistent. Only using fault and warning now. A fault is something that causes the generator to fail. An warning is something that needs attention before the next use or is causing the system to not operate at 100%

message_definitions/v1.0/common.xml
CarlEngel: This value is required. If a system is not monitoring the voltage it can lead to overcharging of the battery. This can be used to make sure that the connections are made correctly and that the calibration of the sensor is correct.

hamishwillee:
@CarlEngel @lthall This looks great to me. Merged as WIP with very minor change since last iteration (added Carl's comment about faults vs warnings to the status)

Hope this proves useful to both of you. Remember that it is marked WIP, so you can iterate while you test this does what you need. 

Please let me know once you've got this integrate in your generator code. If there is anything in particular you need of QGroundControl, best to follow up in the slack channel with Don.

auturgy:
@peterbarker

Issue #1275, opened 2019-12-02
MAVLinkFTP: Require support of compressed (text) files


olliw42:
The MAVLink micro services involve a number of different files, which may be stored in the vehicle so that they can be fetched with MAVLinkFTP.

However, they can be very verbose, e.g. when they are text based, and thus consume much memory on the vehicle. For instance, the Camera micro service involves a camera definition XML file, COMPONENT_INFORMATION involves a component definition XML file.

The proposal is to specify at least one compressed format, which must be supported by any MAVLinkFTP client, such as GCSes. I would propose this to be .zip.

For instance, if it is .zip, a camera definition file might be stored on the vehicle as cameraxyz_def.xml.zip and the MAVlinkFTP client would automatically unzip it into cameraxyz_def.xml before processing.

To give you some numbers from my own "experience":
* The STorM32 has currently 3 camera and 2 component definition xml files stored which together have a size of ca 250 kB, and which would compresses to slightly less than 20 kB with .zip. (I currently use my own compression scheme for storing them since the MCU's memory is only 256kB, but it would be much easier if it just could store the zips)
* I've tentatively created a component definition xml file for ArduPilot, and it has a size of 710 kB, but would compress to 64 kB.

Note that some of the files one typically would NOT want to store on a on-board SD card (if the component has any) since it wouldn't then be easily possible to upgrade these files with the "normal" firmware upgrade method of flashing the firmware to the MCU.

hamishwillee:
Sounds like a good idea to me. Generally the gimbal or whatever will have pre-compressed, and it doesn't seem particularly costly for a GCS to have a library to decompress. @julianoes @dogmaphobic Any thoughts on this?

julianoes:
> cameraxyz_def.xml.zip and the MAVlinkFTP client would automatically unzip it into cameraxyz_def.xml 

I think that makes sense.

hamishwillee:
I very much like the end goal, but the FTP protocol itself feels like it might not be the right place. Mostly because it is a generic file transfer protocol and adding specific behaviour for zip file handling and extraction is odd. 

The "right" place feels like the protocol that uses the service - e.g. camera and component information.

I'm not sure how to progress this.

olliw42:
I actually had the same consideration and initially attached it to the micro services in my mind. 

[...]

PR #1355, opened 2020-04-27
common.xml: Camera definition files should support xz compression

hamishwillee:
The component and camera definition files may be compressed using .xz, and GCS that implement the protocol must support this. Already covered in other docs, e.g.  https://mavlink.io/en/services/component_def.html#file-compression

Fixes https://github.com/mavlink/mavlink/issues/1275


hamishwillee:
@dogmaphobic There is a big discussion of this in https://github.com/mavlink/mavlink/issues/1275 - essentially this expects the GCS to note the .xml.gz extension and decrypt the text file. Saves quite a bit of flash space.

hamishwillee:
@dogmaphobic Can you confirm this is correct please - ie that QGC supports automatic extraction of a camera definition file with extension .xml.gz ?

hamishwillee:
@julianoes @LorenzMeier We will support **.xz** compressed component information files. However according to Gus it is NOT the case that QGC supports compressed *camera* information files (as this PR says ".xml.gz"). 

What's the best way forward?
- Close this (no requirement)
- Modify compression required to **.xz** and merge, even though QGC does not support it.
- As above, but someone gets allocated to making QGC support it
- Someone works to move the camera definition files and schema into component information format?
- Turn this into an issue - we'd like to do this at some point. I have no problem leaving desirable issues open, but would prefer that PRs were closed.

julianoes:
MAVSDK does not support `.xz` yet but it can easily be added once required.

hamishwillee:
> MAVSDK does not support .xz yet but it can easily be added once required.

@julianoes True, but when is something "required"? I'd say that this is "shortly after I submit this", which means that I shouldn't do this until you or someone else has volunteered to make that change, and ideally prototyped it. Volunteering? :-)

julianoes:
I'll add it as soon as someone complains that their .xz file doesn't work :smile:. It's called just-in-time-driven-programming :smile:

hamishwillee:
Hmmm. Well, I'll hit merge, but QGC is now non compliant with this boo hiss

PR #1360, opened 2020-04-28
fix MAV_CMD_INJECT_FAILURE attribute

BazookaJoe1900:
name is not valid attribute for parameter, label is
That fails the validator


hamishwillee:
Thanks. Missed that you had done this and fixed it in your other issue.

PR #1361, opened 2020-04-28
Fixe Common.xml to be valid to schema

BazookaJoe1900:
Fixed XML to apply schema rules
Fixed typo 'True' to 'true'
Removed 'wip' element from messgaes that are depricated. both element can't be together


hamishwillee:
Thanks @BazookaJoe1900 . Yes, those are bugs. I also fixed up and issue where a parameter was using name rather than label. 
Will merge

amilcarlucas:
It does not make sense.... but .... some messages ARE both wip (never got really finished) and deprecated (got replaced with something newer)

hamishwillee:
And I'm actually a bit nervous about this. Deprecation is the right way to go, but in fact the path for a GCS is not completely clear because some camera hardware (for example) will not yet be updated to support the new requestor message. 
This is why we need microservices versioning.

PR #1362, opened 2020-04-29
Set and Get WiFi configuration, adding new fields and enums in the pr…

auturgy:
As discussed in dev call

@dogmaphobic Dev call discussed this and created PR. 

Main points:
- Need to make sure that extensions default to 0 for zero truncation. Also that if this is sent by a system that does not know about the fields, this will set the modes to the values  you have specified - is that appropriate? Normally you'd have a 0 value which says - don't change the mode or whatever.
- The interaction is that you can send MAV_CMD_REQUEST_WIFI_CONFIG to get the WIFI_CONFIG_AP:
  - Should it send back the password? If you do this then anyone who sends the message gets whatever the person set, which might be their phone number or whatever. Perhaps if emitting this message it should leave password empty or a hash?
  - Use MAV_CMD_REQUEST_MESSAGE and id for WIFI_CONFIG_AP - ie delete MAV_CMD_REQUEST_WIFI_CONFIG 
- Similarly, Jonas seemed to think that WIFI_CONFIG_AP being received should result in it being emitted again (as an acknowledgement)
  - If this is the case, it should be documented
  - Again, the password should not really be emitted ?
  - This is a broadcast message so will be processed by everyone. SHould it be? I guess only a wifi can act on it - everything else will just ignore, and that you'll only have one Wifi set up in a MAVLink network?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1246,6 +1246,33 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
     </enum>
+    <enum name="WIFI_CONFIG_AP_RESPONSE">
+      <description>Possible responses from a WIFI_CONFIG_AP message.</description>
+      <entry value="0" name="WIFI_CONFIG_AP_RESPONSE_ACCEPTED">
+        <description>Changes accepted.</description>
+      </entry>
+      <entry value="1" name="WIFI_CONFIG_AP_RESPONSE_REJECTED">
+        <description>Changes rejected.</description>
+      </entry>
+      <entry value="2" name="WIFI_CONFIG_AP_RESPONSE_MODE_ERROR">
+        <description>Invalid Mode.</description>
+      </entry>
+      <entry value="3" name="WIFI_CONFIG_AP_RESPONSE_SSID_ERROR">
+        <description>Invalid SSID.</description>
+      </entry>
+      <entry value="4" name="WIFI_CONFIG_AP_RESPONSE_PASSWORD_ERROR">
+        <description>Invalid Password.</description>
+      </entry>
+    </enum>
+    <enum name="WIFI_CONFIG_AP_MODE">
+      <description>WiFi Mode.</description>
+      <entry value="0" name="WIFI_CONFIG_AP_MODE_AP">
+        <description>WiFi configured as an access point.</description>
+      </entry>
+      <entry value="1" name="WIFI_CONFIG_AP_MODE_STATION">
+        <description>WiFi configured as a station connected to an existing local WiFi network.</description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM and MISSION_ITEM_INT messages) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
@@ -2279,6 +2306,16 @@
       <entry value="2600" name="MAV_CMD_CONTROL_HIGH_LATENCY" hasLocation="false" isDestination="false">
         <description>Request to start/stop transmitting over the high latency telemetry</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">Control transmission over high latency telemetry (0: stop, 1: start)</param>
+        <param index="2">Empty</param>
+        <param index="3">Empty</param>
+        <param index="4">Empty</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
+      <entry value="2620" name="MAV_CMD_REQUEST_WIFI_CONFIG" hasLocation="false" isDestination="false">
+        <description>Request WiFi configuration (WIFI_CONFIG_AP)</description>
+        <param index="1">Empty</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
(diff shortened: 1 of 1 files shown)


dogmaphobic:
> * Perhaps if emitting this message it should leave password empty or a hash?

It makes sense to send a hash only. We just need to agree on what hash type to use.

>Also that if this is sent by a system that does not know about the fields,

Yes, I guess it would be safer to define something like:
```
 <entry value="0" name="WIFI_CONFIG_AP_MODE_UNDEFINED">
```
and
```
<entry value="0" name="WIFI_CONFIG_AP_RESPONSE_UNDEFINED">
```
That would indicate this message comes from an older version, which does not know about the new fields.

> * Use MAV_CMD_REQUEST_MESSAGE and id for WIFI_CONFIG_AP - ie delete MAV_CMD_REQUEST_WIFI_CONFIG

Yup, it makes sense, especially given that this is a new command.

With that said, I just need to know what's the final conclusion so it can be implemented.

hamishwillee:
@dogmaphobic Final conclusion on what? If you just mean the hash, then we'd be led by what you suggest. 

Otherwise, if you update the PR as discussed/you indicated I'd be good with it.

dogmaphobic:
I think it's all there. 

* Added  WIFI_CONFIG_AP_MODE_UNDEFINED and WIFI_CONFIG_AP_RESPONSE_UNDEFINED to handle old code (pre extension).
* Added documentation of using MD5 hash for password when message is sent back as a response.
* Removed the MAV_CMD_REQUEST_WIFI_CONFIG command in lie of MAV_CMD_REQUEST_MESSAGE.

message_definitions/v1.0/common.xml
```diff
@@ -6223,8 +6256,11 @@
     </message>
     <message id="299" name="WIFI_CONFIG_AP">
       <description>Configure AP SSID and Password.</description>
-      <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID). Leave it blank to leave it unchanged.</field>
-      <field type="char[64]" name="password">Password. Leave it blank for an open AP.</field>
+      <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID). Blank to leave it unchanged when setting. Current SSID when sent back as a response.</field>
```
hamishwillee, line 6259: We should still say how this is requested and confirm when this is emitted. Just in response to MAV_CMD_REQUEST_MESSAGE, or also when the WIFI_CONFIG_AP message is received?

Either way, there are potential problems if the there can ever be more than one AP on the network (?)

The reason is that this is broadcast. So if, for example, I send MAV_CMD_REQUEST_MESSAGE to emit WIFI_CONFIG_AP, then even if I send it direct to a particular AP, the other AP will get the response and copy that setting (with the hash).

Also if I set the WIFI_CONFIG_AP then all APs will take the same settings.

If there is only ever one API that might get this message then there is no problem.

If there can be multiple APs then you might consider having both a setter and informational message, with the setter having a target sys id.

dogmaphobic: > We should still say how this is requested and confirm when this is emitted. Just in response to MAV_CMD_REQUEST_MESSAGE, or also when the WIFI_CONFIG_AP message is received?

Yes, the idea is for the AP to send a `WIFI_CONFIG_AP` in response (with the results) each time it receives a `WIFI_CONFIG_AP` (with the settings to be made). That however, will make it the only MAVLink message that behaves this way. I can't see an alternative though. Messages do not "ACK". The ground station has no way to know if the `WIFI_CONFIG_AP` message was received or accepted.

Given that, yes. The AP will send an `WIFI_CONFIG_AP` message in response to both. A specific `MAV_CMD_REQUEST_MESSAGE` command (with the current settings) _and_ in reply to a `WIFI_CONFIG_AP` which is assigning new settings.

>Either way, there are potential problems if the there can ever be more than one AP on the network (?)

No. The way this works, you can only be connected to the one and only AP. Normally, it comes from factory with a default SSID and password. You associate the ground station with this AP and once done, you start receiving MAVLink traffic. At that point, you would set/change new SSIDs and password. Existing systems, such as the ESP8266, have their own compID. They emit heartbeats and even send RADIO_STATUS messages (with RSSI, etc.)

If the AP is set to "Station" (a client of an existing WiFi network), you would have multiple vehicles but you would be addressing them individually using different IP addresses.

hamishwillee: >Yes, the idea is for the AP to send a WIFI_CONFIG_AP in response (with the results) each time it receives a WIFI_CONFIG_AP (with the settings to be made). That however, will make it the only MAVLink message that behaves this way. I can't see an alternative though. Messages do not "ACK". The ground station has no way to know if the WIFI_CONFIG_AP message was received or accepted.

Just for completeness, a more clear interation that you could use is to define two messages so it looks like below. If you don't like that let me know and I'll just make sure the mention of MAV_CMD_REQUEST_MESSAGE is included in the current approach.

```
    <message id="298" name="WIFI_CONFIG_AP_INFO">
      <description>Report Wireless AP settings. Emitted whenever AP settings change (i.e. in response to SET_WIFI_CONFIG_AP) or in response to MAV_CMD_REQUEST_MESSAGE.</description>
      <field type="char[32]" name="ssid">Name of current Wi-Fi network (SSID).</field>
      <field type="char[64]" name="password">MD5 hash of password</field>
      <extensions/>
      <field type="int8_t" name="mode" enum="WIFI_CONFIG_AP_MODE">WiFi Mode.</field>
      <field type="int8_t" name="response" enum="WIFI_CONFIG_AP_RESPONSE">Message acceptance response (sent back to GS).</field>
    </message>
    <message id="298" name="WIFI_CONFIG_AP">
      <description>Configure AP SSID and Password. WIFI_CONFIG_AP_INFO must be emitted after change (even if not successful).</description>
      <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID). Blank to leave it unchanged when setting. </field>
      <field type="char[64]" name="password">Password. Blank for an open AP.</field>
      <extensions/>
      <field type="int8_t" name="mode" enum="WIFI_CONFIG_AP_MODE">WiFi Mode.</field>
      <field type="int8_t" name="response" enum="WIFI_CONFIG_AP_RESPONSE">Message acceptance response (sent back to GS).</field>
    </message>
```

hamishwillee:
@dogmaphobic OK, I've just added info about when this is emitted and I'll merge now. No need to complicate things if this works as you indicated.

PR #1364, opened 2020-05-02
Added yaw to CMD_DO_SET_HOME

dayjaby:
Adding a yaw angle to CMD_DO_SET_HOME. Implementation for PX4 already done: https://github.com/PX4/Firmware/pull/14810

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1586,7 +1586,7 @@
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
-        <param index="4">Empty</param>
+        <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
         <param index="7" label="Altitude" units="m">Altitude</param>


hamishwillee:
FWIW if you propose a change it may sometimes be rejected, so usually best to discuss first before playing with implementation.

@auturgy This makes sense to me. If you're OK with this can you share in ArduPilot community?

message_definitions/v1.0/common.xml
```diff
@@ -1586,7 +1586,7 @@
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
-        <param index="4">Empty</param>
+        <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading</param>
```
hamishwillee, line 1589: If old QGC sends PX4 this message then the value that will be sent is 0 (the default value it sends unless the reserved value is explicitly set to NaN in the original message). That means PX4 will be explicitly set as 0 yaw in most cases. Is that OK?

dayjaby: Px4 used a heading of 0 before as well if it got this command, so nothing would change there. It was just weird to be able to set the yaw on landing items, but if you trigger RTL, it always faces north.

hamishwillee: OK. Though FWIW this mostly should be sent with yaw as NaN to use the system default.

hamishwillee: @auturgy @julianoes Are you OK to merge this? The change is consistent - all other messages that take a Yaw use this parameter in this way.

The only problem I see is that if this command is sent by a GCS that has not been updated then it will send zero by default [this is what QGC does and what the spec says a GCS should do if no default is specified]. So the system will always be set with a home location heading of north in this case.

This should not be a problem for PX4 because according to @dayjaby PX4 always sets north for home location by default. It might be a problem for ArduPilot if it does something else - like set the home location to the current heading of the vehicle.

If no response I will merge tomorrow because I think this is the "right message". Mostly I'm highlighting for awareness.

amilcarlucas: The way we solve this in AP is to use 0.0 as "no change" and to use 360.00 as "0 deg".

hamishwillee:
Thanks - merged!

PR #1375, opened 2020-05-08
VFR_HUD: edit airspeed comment to also allow CAS if available instead of IAS

sfuhrer:
The "airspeed" field in VFR_HUD should be equal to the calibrated airspeed (CAS) if available, as that is corrected for sensor/pitot and placement errors. If no CAS is available IAS (indicated airspeed) can be used.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4983,7 +4983,7 @@
     </message>
     <message id="74" name="VFR_HUD">
       <description>Metrics typically displayed on a HUD for fixed wing aircraft.</description>
-      <field type="float" name="airspeed" units="m/s">Vehicle speed in form appropriate for vehicle type. For standard aircraft this is typically calibrated airspeed (CAS) or indicated airspeed (IAS) - either of which can be used to estimate stall speed.</field>
+      <field type="float" name="airspeed" units="m/s">Vehicle speed in form appropriate for vehicle type. For standard aircraft this is typically calibrated airspeed (CAS) or indicated airspeed (IAS) - either of which can be used by a pilot to estimate stall speed.</field>
       <field type="float" name="groundspeed" units="m/s">Current ground speed.</field>
       <field type="int16_t" name="heading" units="deg">Current heading in compass units (0-360, 0=north).</field>
       <field type="uint16_t" name="throttle" units="%">Current throttle setting (0 to 100).</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4983,7 +4983,7 @@
     </message>
     <message id="74" name="VFR_HUD">
       <description>Metrics typically displayed on a HUD for fixed wing aircraft.</description>
-      <field type="float" name="airspeed" units="m/s">Current calibrated airspeed (CAS), or indicated airspeed (IAS) if CAS is not available.</field>
+      <field type="float" name="airspeed" units="m/s">Vehicle speed in form appropriate for vehicle type. For aircraft this is typically calibrated airspeed (CAS) or indicated airspeed (IAS), which can be used to estimate stall speed.</field>
       <field type="float" name="groundspeed" units="m/s">Current ground speed.</field>
       <field type="int16_t" name="heading" units="deg">Current heading in compass units (0-360, 0=north).</field>
       <field type="uint16_t" name="throttle" units="%">Current throttle setting (0 to 100).</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4983,7 +4983,7 @@
     </message>
     <message id="74" name="VFR_HUD">
       <description>Metrics typically displayed on a HUD for fixed wing aircraft.</description>
-      <field type="float" name="airspeed" units="m/s">Current indicated airspeed (IAS).</field>
+      <field type="float" name="airspeed" units="m/s">Current calibrated airspeed (CAS), or indicated airspeed (IAS) if CAS is not available.</field>
       <field type="float" name="groundspeed" units="m/s">Current ground speed.</field>
       <field type="int16_t" name="heading" units="deg">Current heading in compass units (0-360, 0=north).</field>
       <field type="uint16_t" name="throttle" units="%">Current throttle setting (0 to 100).</field>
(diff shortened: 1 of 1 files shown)


dagar:
I'm fine with this change, but I don't know if others would consider it a change in behavior. 

Anyone feel strongly one way or the other? Maybe @WickedShell or @auturgy?

auturgy:
Im not hard over on it, but it would go against normal aviation convention.  VFR_HUD” is used for more than it’s initial design, but in the context of feeding data to an ASI, by definition that’s IAS, not CAS.

sfuhrer:
@auturgy I really do not know much about aviation standards, but have seen a couple of examples of HUDs displaying CAS and not IAS on google, e.g. [this](https://www.google.com/search?q=head+up+display+airspeed&tbm=isch&ved=2ahUKEwiSra6KvqvpAhUJIMUKHSMCCvIQ2-cCegQIABAA&oq=head+up+display+airspeed&gs_lcp=CgNpbWcQAzoCCAA6BggAEAUQHjoGCAAQCBAeOgQIABAYUIUKWPcYYP0ZaABwAHgBgAGDAYgB4AWSAQM4LjGYAQCgAQGqAQtnd3Mtd2l6LWltZw&sclient=img&ei=Oxm5XtKjPInAlAajhKiQDw&bih=922&biw=1853&client=ubuntu&hs=ELn#imgrc=wJ5NRvDM1DyoKM
). I don't really understand the reason why you should show IAS to the pilot, if you already know that there are some instrument and position errors on the airspeed reading, and you have gone through the process of calibrating it correctly and thus have a different CAS than IAS. It feels like I'm missing here something.

@hamishwillee I don't know much about mavlink processes, but I guess changing fields in the messages isn't really a possibility. But maybe adding fields is possible? 
In the end I'm just trying to find a way to send CAS to the groundstation, such that the user has the same airspeed displayed as is also used internally.

hamishwillee:
@auturgy @TSC21 @dagar @WickedShell If no one is *actively opposed* to this then I plan to merge it tomorrow morning my time. I don't see a case where someone using a HUD would prefer the uncalibrated version if calibration is possible.

@sfuhrer If there _are_ objections, then we can add an extension like this:
```
<extensions/>
<field type="float" name="calibrated_airspeed" units="m/s">Calibrated airspeed (CAS), or NaN if CAS is not available. </field>
```
The downsides of this is that the message no longer benefits from zero-byte truncation (even if CAS is not supplied) and that it only works on MAVLink 2. Also if sent by a system that does not support this field the value will always be zero, which might be a problem. 

This can be improved but it makes the field use a bit weird - e.g. the below allows zero truncation and make it clear that if 0 is received it is actually "not supported" rather than 0 airspeed. 
```
<extensions/>
<field type="float" name="calibrated_airspeed" units="m/s">Calibrated airspeed (CAS).  NaN is should be used in used in place of a 0 measurement. 0 should be used if CAS information is not available. </field>
```
Or you could just make your new field the same as the one your were proposing, but specify NaN for CAS not supported.

auturgy:
No objection, just a comment :)
As a note (and relevant to why aviation general uses IAS): make sure that what you send here matches whatever you are using for stall measurement and prevention. 
fwiw ArduPilot fills this field from a variety of sources depending on vehicle type and available sensors, so the specificity is somewhat misleading anyway (ie rover, boat, etc).

hamishwillee:
Thanks @auturgy - I'm sorry, I'm confused by "I would object to changing this actual message though - it would break a lot of things in a lot of disconnected places."

What I think you are saying is that you wouldn't want an incompatible change - you'd be "ok" with the proposed change to the text (kind of) or to a compatible extension.

WickedShell:
I'm pretty against the extension idea. It's a large waste of bandwidth for most stuff, if you really want it then sticking it in some different more airspeed oriented message would be more reasonable.

I'd support @auturgy's comment that general aviation at least seems to only fly on IAS, not CAS. Particularly since backup gauges aren't ever CAS. You only really start getting CAS as a possible option once you swap to a glass cockpit.

hamishwillee:
@auturgy @WickedShell @sfuhrer So I see that general aviation uses IAS but 
a) this HUD is used for all sorts ofvehicles, 
b) the fact that they do this sounds like "because we don't or can't support CAS" more than "we actually prefer a less reliable and predictable airspead measure". 

So I am still happy to merge this (though @sfuhrer if you really need to rely on the type being CAS then I tend to agree with @WickedShell that having it in another message might be better).

Two last options: 
- we make this _more_ general to reflect that this message is used as a HUD for all vehicle types and have a description like 
  > Vehicle speed. For aircraft this is typically indicated airspeed (IAS) or calibrated airspeed (CAS). For ground vehicles it is typically tachometer measurement."
- we allow the sender to be specific if they want by specifying the type in an optional extra field that can be zero-byte truncated by default (zero cost). This way the sender can choose to be specific and absorb the cost if they so wish. 

If the second option you would do something like:

```xml
    <enum name="VFR_HUD_SPEED_TYPE">
      <description>Type of airspeed data in VFR_HUD</description>
      <entry value="0" name="VFR_HUD_SPEED_TYPE_DEFAULT">
        <description>Vehicle speed is default assumed appropriate for vehicle type by autopilot. Typically IAS or CAS for aircraft.</description>
      </entry>
      <entry value="1" name="VFR_HUD_SPEED_TYPE_IAS">
        <description>Indicated airspeed (IAS).</description>
      </entry>
      <entry value="2" name="VFR_HUD_SPEED_TYPE_CAS">
        <description>Calibrated airspeed (CAS).</description>
      </entry>
    </enum>
```
New extension
```
<extensions/>
<field type="uint8_t" name="speed_type" enum="VFR_HUD_SPEED_TYPE">Type of speed indicated by field airspeed. Default depends on vehicle type and system: nominally IAS or CAS for aircraft.</field>
```
And the field itself

```
<field type="float" name="airspeed" units="m/s">Vehicle speed. The  speed type depends on airframe. For aircraft it is usually either calibrated airspeed (CAS), or indicated airspeed (IAS).</field>
```

Let's make a final decision in the mavcall tonight.

TSC21:
Nothing against this change.

sfuhrer:
@hamishwillee for me your summary makes very much sense. CAS=IAS, unless the user or the manufacturer of the drone deliberately calibrated the airspeed (and the flight control software they use even support CAS). In that case they would probably expect that the airspeed readings on the groundstation to reflect this calibration, as this is then also what the flight controller uses internally. 
For me the extra field would not be necessary. I would expect that the airspeed displayed is equal to the CAS if I've calibrated the sensor, and IAS otherwise.

hamishwillee:
As discussed, the requirement here is that the sender be _allowed_ to send CAS. The field has therefore  beeen generalised to state that the format is dependent on the vehicle type. It also notes that either CAS or IAS are typically used as either can be used by pilot used to estimate stall speed.

This can be merged once CI passes.

PR #1381, opened 2020-05-20
Remove deprecation/rename of MAV_FRAME_BODY_FRD

hamishwillee:
This undos the deprecation and rename of MAV_FRAME_BODY_FRD in #1112, for the reasons given in the comment: https://github.com/mavlink/mavlink/pull/1112#issuecomment-625050092. 

Essentially this is still needed and should not have been marked deprecated. I am self merging to get this in.

@FYI @kamilritz @julianoes

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -879,9 +879,8 @@
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
         <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
-      <entry value="12" name="MAV_FRAME_RESERVED_12">
-        <deprecated since="2019-04" replaced_by="MAV_FRAME_BODY_OFFSET_NED"/>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@TSC21 See this. MAV_FRAME_BODY_FRD recovered.

kamilritz:
@hamishwillee Thank you very much for reverting this. 
I noticed that at the moment `MAV_FRAME_BODY_NED` and `MAV_FRAME_BODY_OFFSET_NED` are at the moment marked as [deprecated](https://github.com/mavlink/mavlink/pull/1381/files#diff-ceacc1e2390186650146a45be8fcb691R869-R873). In my view those frames are really deprecated and they don't seem to be used  in PX4. Other frame that are deprecated, got renamed to `MAV_FRAME_RESERVED_*` in #1112. Would it be an option to do the same with `MAV_FRAME_BODY_NED` and `MAV_FRAME_BODY_OFFSET_NED`?

hamishwillee:
@kamilritz If you would like to create a PR renaming them to reserved values (also moving the original name into the description) we can discuss that.

I'm not at all sure about the use of `MAV_FRAME_RESERVED_` in this way. What I thought should happen is that a value is deprecated. Then at some future point the value gets changed to MAV_FRAME_RESERVED_ to tell everyone that they cannot use it any more and that it may be used for a new purpose soon. Changing it at the point of deprecation feels premature - forces a quick update when the user might not be ready.

So if you create the PR I will use it as an opportunity to clarify the process :-)

kamilritz:
So it was not the intent to rename all deprecated frames. I am fine with this. Thanks again

PR #1383, opened 2020-05-26
MAV_CMD_NAV_LOITER_TIME: add Heading Required field

sfuhrer:
Other then for the MAV_CMD_NAV_LOITER_ALT, for MAV_CMD_NAV_LOITER_TIME there was no possibility to add a "heading wait" requirement. In my opinion this field should be consistent for both of these loiter types.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1317,10 +1317,10 @@
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="19" name="MAV_CMD_NAV_LOITER_TIME" hasLocation="true" isDestination="true">
-        <description>Loiter at the specified latitude, longitude and altitude for a certain amount of time. Additionally (for fixed wing vehicles), if the Heading Required parameter (2) is non-zero the aircraft will only leave the loiter path when heading towards the next waypoint.</description>
+        <description>Loiter at the specified latitude, longitude and altitude for a certain amount of time. Multicopter vehicles stop at the point (within a vehicle-specific acceptance radius). Forward-moving vehicles (e.g. fixed-wing) circle the point with the specified radius/direction. If the Heading Required parameter (2) is non-zero forward moving aircraft will only leave the loiter circle once heading towards the next waypoint.</description>
         <param index="1" label="Time" units="s" minValue="0">Loiter time (only starts once Lat, Lon and Alt is reached).</param>
-        <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter only when heading towards the next waypoint (0 = False)</param>
-        <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise.</param>
+        <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter circle only once heading towards the next waypoint (0 = False)</param>
+        <param index="3" label="Radius" units="m">Loiter radius around waypoint for forward-moving vehicles. If positive loiter clockwise, else counter-clockwise.</param>
         <param index="4">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle.  NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1318,9 +1318,9 @@
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="19" name="MAV_CMD_NAV_LOITER_TIME" hasLocation="true" isDestination="true">
-        <description>Loiter around this waypoint for X seconds</description>
-        <param index="1" label="Time" units="s" minValue="0">Loiter time.</param>
-        <param index="2">Empty</param>
+        <description>Loiter at the specified Latitude, Longitude and Altitude for a certain amount of time.  Additionally, if the Heading Required parameter is non-zero the  aircraft will not leave the loiter until heading toward the next waypoint.</description>
+        <param index="1" label="Time" units="s" minValue="0">Loiter time (only starts once Lat, Lon and Alt is reached).</param>
+        <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Heading Required (0 = False)</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise.</param>
         <param index="4">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle.  NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
         <param index="5" label="Latitude">Latitude</param>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@sfuhrer Do you mean [MAV_CMD_NAV_LOITER_TO_ALT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT)? 

This and other loiter commands specify an Xtrack location, which I interpret to mean an "exit track" location. That sounds a bit like the heading. On LOITER_TO_ALT

> Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).

What I think that means is that the system will leave the loiter:
- 0: When vehicle reaches line between current waypoint (centre) and next waypoint. [Essentially this would be at 90 degrees to the path you want :-( )
- 1: For exit location [don't know what this means - do you?]
- NaN: yaw to system location - I think this means will exit when yaw matches system setting for yaw - ie normally when facing next waypoint (which is what you want) but could be when facing home.
- Other value is "specific value" 

So it sounds like this is setting the exit heading, and probably you want either 1 or NaN. The thing it is missing is just "head off whenever you have reached the altitude].

I'd like to clarify what Xtrack does, because if we want to be consistent, then we need to adopt the same terminology. Does this make sense?

message_definitions/v1.0/common.xml
```diff
@@ -1317,9 +1317,9 @@
...
-        <description>Loiter around this waypoint for X seconds</description>
-        <param index="1" label="Time" units="s" minValue="0">Loiter time.</param>
-        <param index="2">Empty</param>
+        <description>Loiter at the specified Latitude, Longitude and Altitude for a certain amount of time.  Additionally, if the Heading Required parameter is non-zero the  aircraft will not leave the loiter until heading toward the next waypoint.</description>
+        <param index="1" label="Time" units="s" minValue="0">Loiter time (only starts once Lat, Lon and Alt is reached).</param>
+        <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Heading Required (0 = False)</param>
         <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise.</param>
         <param index="4">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle.  NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
```
amilcarlucas, line 1324: Why not use the index 4 for this? It already has an effect for forward moving vehicles, just add the descrition for what it should do for "non-forward moving vehicles". Something like:
`<param index="4" label="xtrack location/heading">For forward moving vehicles, this sets exit xtrack location: 0 for center of loiter, 1 for exit location. Else, this is desired yaw angle.  NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.). For other vehicles 0 or NAN means do not change Yaw behaviour, 1 means Yaw to next WP before leaving this point, other values (2.. 361) are a direct Yaw command.</param>`

sfuhrer:
@hamishwillee here the promised screen shot of a mission containing MAV_CMD_NAV_LOITER_ALT with different configuration (at least that's how PX4 is interpreting it atm): 
with param1: heading reaquired, param4: xtrack location --> (param1, param4)
WP 2: (0,0), WP 4: (1,0), WP 7: (1,1), WP 9: (0,1)
image

hamishwillee:
@sfuhrer I got carried away and tried to tidy up the MC vs FW story on this. Would be good if you could sanity check this.

Decided NOT to change comment about existing on tanjent of loiter to the center of the next waypoint. That isn't helpful because a system might well want to exit on the tanjent of one waypoint to the tanjent of the loiter circle of the next waypoint - and we shouldn't prohibit it.

sfuhrer:
@hamishwillee your changes sound good yes, it's much better defined like this. Only one thing: For param4, isn't there a "For" missing, so "For forward moving aircraft..."?

amilcarlucas:
The current text also talks about wp instead of the lat/lon of this message

hamishwillee:
@amilcarlucas 

> The current text also talks about wp instead of the lat/lon of this message

This means the same thing. IMO it is an unnecessary change, but not incorrect.

The Dev call are OK with the message as is. I'm going to look again tomorrow. I THINK this is OK, and that the change allows some things which cannot be done purely with X-track.

Note, I will go through all these messages and update them consistently at some point.

message_definitions/v1.0/common.xml
hamishwillee: This doesn't support case of leaving when heading is reached but then go back to the center line.

hamishwillee:
@sfuhrer I'm merging this now as agreed with the dev call.

There have been general improvements to text which will roll out later.

hamishwillee:
@sfuhrer I did not make the Xtrack change yet. The current definition of Xtrack is:

> For forward-moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).

I interpret the exit xtrack to be the path that the vehicle will trend towards after exiting the loiter. So the values are.
- `0`: Track is from centre of loiter waypoint to centre of next waypoint, so we exit from tranjent (or wherever) but trend back to the centre line.
  image

- `1`: Track is from loiter circle exit location to centre of the next waypoint. 
  image
  This one is a bit weird because it looks like the exit point is where it leaves crosses the original circle, not where it leaves the original path. 
  image
- Other values indicate desired yaw angle. This isn't at all clear to me - can you clarify. My GUESS is that the yaw angle is the angle of entry into the next waypoint *in degrees* relative to the track between the centres. So for an angle of around 45 you'd get a path like below. If not, can you clarify?
   image
- NaN is system yaw behaviour. By default that is towards the next waypoint - so I'd expect this case to be the same as "0". But what about if yaw setting is towards home? Does that mean that the angle from the vehicle to home at the point the vehicle exits the radius is the angle it would approach the next waypoint (relative to the center-centre path as in previous point?)

sfuhrer:
@hamishwillee I first want to point out that I'm much more familiar with the fixed-wing use case than the rotary-wing one, and I actually think that the second and third sentence are actually referring to MC use cases

> Else, this is desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).

For MC this seems to make sense, as you there can fly with arbitrary yaw. For FW it doesn't, an thus if you set anything beside exactly 1 it should do the default ("0") behavior.

About the 3rd pic you posted, about the weird exit and then going to the tangent path of the opposite side of the circle: that seems to be a bug of PX4. I though want to point out that this setting (no heading wait but then xtrack=1) really makes no sense and should even be prevented (only allow xtrack=1 with also heading wait =1). Not sure if a re-write of this mavlink command though is necessary, or if it is rather for the ground station to disable setting anything other than heading wait =1 if xtrack=1.

hamishwillee:
@sfuhrer Thank you. 
1. We really don't have any mechanism other than documentation to specify that something behaves a particular way for a particular vehicle, or how parameter values affect each other. 
1. The docs on xtrack explicitly say "For forward-moving aircraft" - so any options that don't make sense for FW should be removed. Let's ask James ...

@auturgy Xtrack defines the path that a fixed-wing vehicle will converge towards upon leaving loiter. I've put some examples above - https://github.com/mavlink/mavlink/pull/1383#issuecomment-635053440

The first two options above make sense - a tanjential path from the loiter to the center of the next waypoint, or a path from the tanjent that convergest to the direct line between the two points.

The other two xpath settings are not clear, but I think they define "an angle of entry to the next waypoint relative to the center line between the waypoints". The angle being either explicit (degrees) or NaN ("system yaw"). In particular the system yaw option makes little sense since the behavour would be very unpredictable. 

Does ArduPilot use those, and if so, what for/when do you want to do this?

My proposal is that we simply remove xtrack options other than 0 and 1.

hamishwillee:
@auturgy When you get a chance, can you please look at my last post above

PR #1385, opened 2020-05-26
Add image count field to CAMERA_CAPTURE_STATUS message.

MatejFranceskin:
Problem:
- there is no way to determine the current image count that could be used if any [CAMERA_IMAGE_CAPTURED](https://mavlink.io/en/messages/common.html#CAMERA_IMAGE_CAPTURED) messages were lost.
- camera_index field in CAMERA_IMAGE_CAPTURED specifies that image count is reset at arming. This doesn't allow implementation of a persistent photo log that could be produced combining several flights.

This PR adds image_count field to CAMERA_CAPTURE_STATUS message. Implementation of camera manager can use this field to implement persistant photo log and remotes can request missing entries with MAV_CMD_REQUEST_MESSAGE.
Persistent photo log image count can be reset with added parameter for  MAV_CMD_STORAGE_FORMAT command.
Definition of image_index was changed so that arming is not mentioned anymore, leaving the actual interpretation to the camera manager.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6133,6 +6133,7 @@
       <field type="float" name="image_interval" units="s">Image capture interval</field>
       <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started</field>
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
+      <extensions/>
       <field type="int32_t" name="image_count">Total number of images.</field>
     </message>
     <message id="263" name="CAMERA_IMAGE_CAPTURED">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6132,6 +6133,7 @@
       <field type="float" name="image_interval" units="s">Image capture interval</field>
       <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started</field>
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
+      <field type="int32_t" name="image_count">Total number of images.</field>
```
julianoes, line 6136:
```suggestion
      <extensions/>
      <field type="int32_t" name="image_count">Total number of images.</field>
```

hamishwillee:
This looks great. Devcall agreed to merge. Action on @MatejFranceskin to update https://mavlink.io/en/services/camera.html with section on how to use this.

PR #1387, opened 2020-05-27
common: HIL_GPS: extend with an ID field

TSC21:
If one wants to simulate multiple GPS inputs from simulation, and using the same messaging structure, i.e., `HIL_GPS` messages, we need a field that identifies the sensor source so it can be properly handled and logged in the autopilot side.

An alternative could be use `GPS_INPUT` instead of `HIL_GPS`, but then we would be using messages which purpose is not really to be used as simulation in/outputs.


TSC21:
> add note that is zero indexed.

Interesting detail :D

hamishwillee:
@TSC21  This makes sense to me. This will be interpretted correctly if sent by a system that doesn't know about the extension. If received by a system that does not know about the extension then it will be handled incorrectly for indexes of >0. I assume we're OK with that?

@auturgy Any comment. I'm inclined to merge.

auturgy:
looks fine

PR #1390, opened 2020-06-02
MAV_CMD_STORAGE_FORMAT - clarify reset of image count

hamishwillee:
This clarifies what exactly is meant by "reset image count" in MAV_CMD_STORAGE_FORMAT to mean both the CAMERA_CAPTURE_STATUS.image_count and CAMERA_IMAGE_CAPTURED.image_index [you have to do both, or subsequent captures will result in a mismatch].

This is needed because otherwise the words "image count" mean very little in the context of this message.

FWIW it does feel a bit odd to use this message to reset a count value, but I don't have a better option to offer.

@MatejFranceskin Can you confirm this matches your intent?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6151,7 +6151,7 @@
       <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started</field>
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
       <extensions/>
-      <field type="int32_t" name="image_count">Total number of images.</field>
+      <field type="int32_t" name="image_count">Total number of images captured ('forever', or until reset using MAV_CMD_STORAGE_FORMAT).</field>
     </message>
     <message id="263" name="CAMERA_IMAGE_CAPTURED">
       <description>Information about a captured image. This is emitted every time a message is captured. It may be re-requested using MAV_CMD_REQUEST_MESSAGE, using param2 to indicate the sequence number for the missing image.</description>
@@ -6163,7 +6163,7 @@
       <field type="int32_t" name="alt" units="mm">Altitude (MSL) where image was taken</field>
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
       <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 0, 0, 0, 0)</field>
-      <field type="int32_t" name="image_index">Zero based index of this image (image count -1)</field>
+      <field type="int32_t" name="image_index">Zero based index of this image (i.e. a new image will have index CAMERA_CAPTURE_STATUS.image count -1)</field>
       <field type="int8_t" name="capture_result">Boolean indicating success (1) or failure (0) while capturing this image.</field>
       <field type="char[205]" name="file_url">URL of image taken. Either local storage or http://foo.jpg if camera provides an HTTP interface.</field>
     </message>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2155,8 +2155,8 @@
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT" hasLocation="false" isDestination="false">
         <description>Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.</description>
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
-        <param index="2" label="Format" minValue="0" maxValue="1" increment="1">0: No action 1: Format storage</param>
-        <param index="3" label="Reset Image Count (CAMERA_CAPTURE_STATUS.image_count and CAMERA_IMAGE_CAPTURED.image_index)" minValue="0" maxValue="1" increment="1">0: No action 1: Reset Image Count</param>
+        <param index="2" label="Format" minValue="0" maxValue="1" increment="1">Format storage (and reset image log). 0: No action 1: Format storage</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6151,7 +6151,7 @@
       <field type="uint32_t" name="recording_time_ms" units="ms">Time since recording started</field>
       <field type="float" name="available_capacity" units="MiB">Available storage capacity.</field>
       <extensions/>
-      <field type="int32_t" name="image_count">Total number of images.</field>
+      <field type="int32_t" name="image_count">Total number of images captured ('forever', or until reset using MAV_CMD_STORAGE_FORMAT).</field>
```
hamishwillee, line 6154: @MatejFranceskin It is unlikely this will exceed max int32 value, but if it does, should both the index and count rollover to 0 again?

message_definitions/v1.0/common.xml
```diff
@@ -6163,7 +6163,7 @@
       <field type="int32_t" name="alt" units="mm">Altitude (MSL) where image was taken</field>
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
       <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 0, 0, 0, 0)</field>
-      <field type="int32_t" name="image_index">Zero based index of this image (image count -1)</field>
+      <field type="int32_t" name="image_index">Zero based index of this image (i.e. a new image will have index CAMERA_CAPTURE_STATUS.image count -1)</field>
```
hamishwillee, line 6166: This clarification needed because the index does not have to be for a new image, where the total count is relevant. Also, in this message `image_count` doesn't mean anything - so full name of message needed.

message_definitions/v1.0/common.xml
```diff
@@ -2156,7 +2156,7 @@
         <description>Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.</description>
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
         <param index="2" label="Format" minValue="0" maxValue="1" increment="1">0: No action 1: Format storage</param>
-        <param index="3" label="Reset Image Count" minValue="0" maxValue="1" increment="1">0: No action 1: Reset Image Count</param>
+        <param index="3" label="Reset Image Count (CAMERA_CAPTURE_STATUS.image_count and CAMERA_IMAGE_CAPTURED.image_index)" minValue="0" maxValue="1" increment="1">0: No action 1: Reset Image Count</param>
```
hamishwillee, line 2159: Should this say "Reset Image Log" (not count?)

message_definitions/v1.0/common.xml
dogmaphobic: 2,147,483,647 * ~1MB = 2048TB or 2PB. Not taking into account that images tend to get bigger, not smaller as technology advances, which would bring the above to even more absurd storage requirements. My camera for instance, produces a 60MB image. That would mean 120PB of storage :P

hamishwillee: @dogmaphobic @MatejFranceskin I'm OK with the wording through, because there won't be new captures if the disc is full. 

This does make me think though:
- what if they format the disc. Shouldn't this automatically also format the image count? As defined, whether that happens or not would depend on where the log/counts are stored. We should be explicit?

MatejFranceskin: I agree with your changes.
Regarding exceeding max int32 I think that it would be worse if we rollover then if we stay at the maximum value and don't take more photos.
Regarding storage format - it actually doesn't make sense that you would erase the images but keep the log counts. Probably we should be explicit.

message_definitions/v1.0/common.xml
```diff
@@ -2155,8 +2155,8 @@
       <entry value="526" name="MAV_CMD_STORAGE_FORMAT" hasLocation="false" isDestination="false">
         <description>Format a storage medium. Once format is complete, a STORAGE_INFORMATION message is sent. Use the command's target_component to target a specific component's storage.</description>
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
-        <param index="2" label="Format" minValue="0" maxValue="1" increment="1">0: No action 1: Format storage</param>
-        <param index="3" label="Reset Image Count" minValue="0" maxValue="1" increment="1">0: No action 1: Reset Image Count</param>
+        <param index="2" label="Format" minValue="0" maxValue="1" increment="1">Format storage (and reset image log). 0: No action 1: Format storage</param>
+        <param index="3" label="Reset Image Log (without formatting storage medium). This will reset CAMERA_CAPTURE_STATUS.image_count and CAMERA_IMAGE_CAPTURED.image_index." minValue="0" maxValue="1" increment="1">0: No action 1: Reset Image Count</param>
```
hamishwillee, line 2159: @MatejFranceskin @dogmaphobic As [discussed](https://github.com/mavlink/mavlink/pull/1390/files#r434392975), now says that format also clears the log.

Do we know what will happen if an image that is requested doesn't exist? [and do we need to specify that]?

message_definitions/v1.0/common.xml
hamishwillee: OK, but since Gus thinks that this will require images to get a bit more out of control I'm proposing to leave specifying this for now.

message_definitions/v1.0/common.xml
hamishwillee: Merging now anyway, but we can look at ^^^ as a post process if needed.

amilcarlucas: The length of this label is huge. A GCS will have problems redering this!!! Most of the text belongs in the param description, not it the label attribute.

hamishwillee: Thanks very much @amilcarlucas ! This will be fixed by https://github.com/mavlink/mavlink/pull/1410

PR #1392, opened 2020-06-08
Added a MAVLink-ArduCopter SITL Example

dongsin-kim:
This is a simple MAVLink example for beginners who are interested in receiving and sending some messages from/to the ArduPilot SITL (Simulation In The Loop) virtual vehicle, specifically focus on ArduCopter, using the MAVLink C libraries.

This code contains three parts:
1. Configure the connection with the vehicle using the address "127.0.0.1:14550".
2. Send a SET_MODE command to the vehicle.
3. Receive a GLOBAL_POSITION_INT message from the vehicle.

### DIFF
--- a/examples/ArduCopter/mavlink_copter.c
+++ b/examples/ArduCopter/mavlink_copter.c
@@ -1,0 +1,213 @@
+/*   
+Dongsin Kim (ds.justin.kim @ gmail.com)
+*/
+
+
+/*
+MIT License
+
+Copyright (c) [2020] [Dongsin Kim]
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


examples/ArduCopter/mavlink_copter.c
```diff
@@ -0,0 +1,213 @@
...
+		{
+			for (i = 0; i < recsize; ++i)
+			{
+				if (mavlink_parse_char(MAVLINK_COMM_0, buf[i], &msg, &status)) // Parse packet
+				{
+					/* Packet received */
+					// printf("\nReceived packet: SYS: %d, COMP: %d, LEN: %d, MSG ID: %d\n", msg.sysid, msg.compid, msg.len, msg.msgid);
+					switch(msg.msgid)
+					{
+						// case MAVLINK_MSG_ID_HEARTBEAT: // #0
```
hamishwillee, line 150: Do we need commented out stuff in the example? If you expect people to need this, then perhaps making it easy to uncomment using /* and */ and a note explaining when/why they do this would be good.

hamishwillee:
@dongsin-kim What does this add over the other examples?

- Ideally we want to try and remove any flight-stack specific behaviour from the examples - or more generally a well designed example should work on any flight stack. 
- As far as I can tell, except for the mode that is set, this is actually generic. The mode stuff is flight stack specific. The address for the simulator is also "common".
- SET_MODE is deprecated, examples should use the latest and greatest commands - e.g. MAV_CMD_DO_SET_MODE

I'm not opposed to the example using ArduPilot or PX4 simulators but we should make it very clear what is actually ArduPilot or PX4 specific, and ideally make it general if at all possible.

dongsin-kim:
@hamishwillee , @LorenzMeier thank you for the consideration and helpful comments. I could have learned a lot from this request and discussion. I would also need to consider more about the concept and objective of this example. Thank you so much.

PR #1393, opened 2020-06-09
MAV_CMD_DO_UPGRADE: Align with thinking on long running ops

hamishwillee:
@TSC21 I've been having a discussion with @julianoes on long running operations here: https://github.com/mavlink/mavlink-devguide/pull/243#discussion_r435564953

This changes the behaviour on the upgrade message based on the result of that discussion. Essentially it says that:
-  if you cancel a message the system should respond with MAV_RESULT_CANCELLED
- restarting shouldn't be allowed - you'd cancel the operation and then run a new update. 

This provides a far more predictable behaviour for the state machine for the generic case.

Julian, does not need to merge immediately - I did this now so that Nuno is aware of change before implementation done.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2055,9 +2055,9 @@
       </entry>
       <entry value="247" name="MAV_CMD_DO_UPGRADE" hasLocation="false" isDestination="false">
         <wip/>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@TSC21 On hold. Lorenz suggested we use a separate message for that. See https://github.com/mavlink/mavlink/pull/1401

hamishwillee:
Hi @TSC21 

We have now clarified how [long running operations should work](http://mavlink.io/en/services/command.html#long_running_commands). In summary, they have a new message to cancel them, and to restart you cancel then send the message again. 
This updates the upgrade message appropriately, linking to the updated docs.

Does this make sense?

TSC21:
@hamishwillee yes this looks fine.

Issue #1370, opened 2020-05-05
Gimbal V2 Protocol Discussion II: Gimbal device information


olliw42:
I am not totally happy with the GIMBAL_DEVICE_INFORMATION message.

It's fields are:

- time_boot_ms
- vendor_name, model_name, firmware_version
- cap_flags
- tilt_max, tilt_min, tilt_rate_max, pan_max, pan_min, pan_rate_max

First, I wonder about the vendor_name, model_name, firmware_version fields. These are also contained in the COMPONENT_INFORMATION message, so there is a substantial duplication here. 
I'm not sure if there is an implicit decision about the design or even fate of COMPONENT_INFORMATION in here, but it somehow makes much more sense to me to have a message which is generic for all kinds of components to request such information instead of having very specific messages for specific types of components. Furthermore, I think that besides the firmware version also the hardware version can be an equally relevant piece of information, and I find it thus somewhat regretable that here the layout of CAMERA_INFORMATION and not of COMPONENT_INFORMATION seems to be followed.

**My suggestion:** 
Either remove these fields and shift to COMPONENT_INFORMATION. Or, if they should remain here, for whatever reasons, then please add hardware_version.

Next, it is actually totally unclear to me what the purpose of time_boot_ms could be. This is totally static data, and thus valid at every time, and not just at the time there the data was requested. So, this field appears to be totally superfluous and just eat away bandwidth.

I realize that this comment is also appllies to messages like CAMERA_INFORMATION and COMPONENT_INFORMATION, there CAMERA_INFORMATION was kind of the template for inspiration or the others. But this doesn't make the argument wrong. 

**My suggestion:** 
If the field time_boot_ms is really here just because it is present also in other similarly static messages or because of the better looks of symmetry, I would suggest to remove it, and also e.g. in the COMPONENT_INFORMATION message.

[...]

Issue #1346, opened 2020-04-18
Using COMPONENT_INFORMATION to obtain vehicle metadata


DonLakeFlyer:
This is the higher level proposal related to #1339 for how you get the various pieces of metadata which a GCS needs from a connected vehicle. I've also included my proposed spec for the parameter metadata json format.

# Vehicle metadata request protocol

* GCS sends ```MAV_CMD_REQUEST_MESSAGE.param1 = COMPONENT_INFORMATION ``` ```MAV_CMD_REQUEST_MESSAGE.param2 = COMP_METADATA_TYPE_VERSION ``` to the vehicle
* If ```comp_metadata_uid``` is not already cached GCS does ftp download from specified vehicle uri.
* GCS sends ```MAV_CMD_REQUEST_MESSAGE.param1 = ...``` for each optional metadata it wants. Including the cache dance semantics.

# ```MAV_CMD_REQUEST_MESSAGE```

Notes:

* I think param2-6 should all be "The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).". 
* I don't think it makes sense for param2 to be "Index id (if appropriate). The use of this parameter (if any), must be defined in the requested message.".

# ```COMPONENT_INFORMATION```

Notes:

* The removal of all fields other than what is needed to download metadata files. The fields that were previously include here are now part of the ```COMP_METADATA_TYPE_VERSION``` type metadata file spec.
* The type of requested metadata is specified in ```MAV_CMD_REQUEST_MESSAGE.param2```

| Type        | Field           | Description  |
| ------------- | -------------| -----|
| ```uint32_t``` | ```time_boot_ms``` | Timestamp (time since system boot) |
| ```uint32_t``` | ```comp_metadata_type``` | The type of metadata returned (```COMP_METADATA_TYPE```). |
| ```uint32_t``` | ```comp_metadata_uid``` | Unique uid for this metadata which a gcs can use for created cached metadata and understanding whether it's cache it up to date or whether it needs to download new data. |
| ```char[140]``` | ```component_metadata_uri``` | Component definition URI. If prefix ```mavlinkftp:\\``` file is downloaded from vehicle over mavlink ftp protocol. If prefix ```http[s]:\\``` file is downloaded over internet. Files are json format. They can end in ```.gz``` to indicate file is in gzip format. |
| ```uint32_t``` | ```comp_translation_uid``` | Unique uid for the translation file associated with the metadata. |
[...]

PR #1394, opened 2020-06-09
COMPONENT_INFORMATION: add actual_compid field

ndepal:
As discussed in https://px4.slack.com/archives/C52DXKYLV/p1591641605371800

This is useful to communicate component information for components that do not have their own mavlink interface. For example, a GPS device connected to a flight controller.

@julianoes I made it part of the message since it's still marked WIP, but I can make it an extension if you prefer.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6677,6 +6677,7 @@
       <field type="uint32_t" name="capability_flags" enum="COMPONENT_CAP_FLAGS">Bitmap of component capability flags.</field>
       <field type="uint16_t" name="component_definition_version">Component definition version (iteration)</field>
       <field type="char[140]" name="component_definition_uri">Component definition URI (if any, otherwise only basic functions will be available). The XML format is not yet specified and work in progress. </field>
+      <field type="uint8_t" name="actual_compid">ID of the component this message is about. Used if not the same as component ID of message sender. 0 if not used.</field>
     </message>
     <message id="400" name="PLAY_TUNE_V2">
       <wip/>


julianoes:
I think this is a good start to discuss in the MAVLink call tomorrow, 9am CEST. The chance to get it in is best if you can show up.

ndepal:
Ok, I'll try to attend.

hamishwillee:
Not sure about this. FWIW I think I'd rather that either:
1. We made it easy (in general) for an autopilot to emulate the behaviour of mavlink components with heartbeat etc.
2. We made it easy to fake component id for outbound messages, and told the autopilot (or other system) it has to accept inbound commands to those components.
3. We allow an autopilot (actually any component in system) to output the component metadata for any/all other components if it chooses (sorted by type/component id). That way the autopilot can output this information for the attached "not real mavlink" components. If there is conflict the one with the component source id is used.

I tend towards 3 because it is the minimal solution.

hamishwillee:
Hi @ndepal 
No, because you wouldn't request the component specifically. You'd request "all metadata from autopilot" and the autopilot (presumably) would send back not just its own information, but that for all components.

We talked about this in the dev call. The problem with option three is that it requires quite a lot of infrastructure to get the information - MAVFTP for a start, and the format of that metadata isn't defined. Essentially we think it is a good idea, but not when all the pieces of the story are still in flux.

The right way to do this is to be able to update the generated libraries to support setting an arbitrary component ID, but that may be contrary to the spec. We may well need this in future anyway, but we don't have it now.

So even though none of us liked this proposal, but it is the only one that delivers what you need in a reasonable timeframe. So we thought maybe merge as WIP but it might well change at some point before it stabilises.

Leaving to @julianoes to confirm that.

ndepal:
Ok thanks for the explanation.

julianoes:
@hamishwillee I found one more problem with this. I'm gonna try to explain it:

So basically @ndepal needs the version of a component which is not a MAVLink thing itself.

You suggested with 3. to enable the autopilot to share some sort of json file with versions of components. This would then be downloaded using MAVLink FTP. I already brought up in the call that while I see how all this meta data can be useful and makes sense, that it is also basically a way to share data that is less (or differently) standardized, so it is more flexible for specific use cases. Basically, the meta data is a workaround for the fact that the MAVLink messages don't really have versioning (apart from extensions) and arbitrary size, whereas the json schema will have a version which can be iterated and more data can be added more easily (and MAVLink FTP takes care of the retransmission of the file chunks).

Anyway, as we're discussing the COMPONENT_INFORMATION in https://github.com/mavlink/mavlink/issues/1346, part of the proposal is to move the version fields into metadata as well (search for `firmwareVersion`). This means that once #1346 goes in the version field that @ndepal requires is removed :sob:.

This means that we either need to leave the version field in the message and not move it to the json "metadata" file, or we need to go towards that new proposal already meaning that we need to have a way for any component to be represented in the json file and @ndepal needs to implement MAVLink FTP in whatever client he is using (if it's MAVSDK it's already included) to get the version.

DonLakeFlyer:
This is superceded by #1408

hamishwillee:
@DonLakeFlyer I'm not sure the new PR addresses the use case (though it does make this one not work anymore)

The need here is to be able to get the version information from a non-mavlink component in a system, allowing all the components in a system to be enumerated. This basically added a field to say that I really want the details for the specified component id.

 I did suggest above we could have the autopilot returning version info for all attached components. Maybe that would work.

Suggestions?

julianoes:
@DonLakeFlyer @hamishwillee  can we bring the version field back to the COMPONENT_INFORMATION. It's something needed for @ndepal and the v2 gimbal proposal as mentioned in #1370.

DonLakeFlyer:
> I'm not sure the new PR addresses the use case (though it does make this one not work anymore)

I was thinking for some reason the version json had component id which it doesn't. So my comment is wrong.

hamishwillee:
@julianoes As you know the component information message has pretty much turned into a system for supplying metadata file locations - ie the message itself doesn't carry much other than information about what metadata is supported and where you can get it. There isn't space in the payload for this any more. Also, this wouldn't be the "right" way to do this.

I think now what you would do is send MAV_CMD_REQUEST_MESSAGE to the system and component you want specifying the id for the COMPONENT_INFORMATION message. All components should respond with their version information in this format https://github.com/mavlink/mavlink/blob/master/component_information/version.schema.json

Components will have to support MAVFTP and include a version file. I suspect that an autopilot that wants to supply information for "not real" components will have to fake the source component ID.

Thoughts?

hamishwillee:
This is no longer appropriate, but might be considered for COMPONENT_INFORMATION_BASIC.

Issue #1376, opened 2020-05-08
Gimbal V2 Protocol Discussion II: Gimbal manager set attitude command vs set attitude message 


olliw42:
I find the two gimbal manager control messages DO_GIMBAL_MANAGER_ATTITUDE and GIMBAL_MANAGER_SET_ATTITUDE, and especially their interplay and inter-dependence, somewhat not so intuitive. I will comment on the interplay in an extra issue, but want to here suggest the following:

It is somewhat strange that both refer to ATTITUDE, but DO_GIMBAL_MANAGER_ATTITUDE in fact related to tilt and pan angles and not the attitude quaternion., as all the others do.

Now, I very well understand the motivation for that, since for setting a mission one really doesn't want to fool around with a quaternion but wants more human friendly variables. And this generally also bodes very well with the idea of the gimbal manager to abstract the gimbal operation, as it's also explicitly mentioned in the documentation.

However, IMHO, a very similar argument could be made also for GIMBAL_MANAGER_SET_ATTITUDE. In many if not most typical applications one would prefer to deal directly with tilt and pan angles and not with a quaternion. As example I mention a GCS doing stick control (which is also mentioned in the docs).

In addition, I feel that appropriately descriptive command/message names also can help a lot.

**My suggestion:**

* I would rename the existing command DO_GIMBAL_MANAGER_ATTITUDE into DO_GIMBAL_MANAGER_TILTPAN (or DO_GIMBAL_MANAGER_PITCHYAW)

* I would add a new message GIMBAL_MANAGER_SET_TILTPAN (or GIMBAL_MANAGER_SET_PITCHYAW), which in function is analogous to the existing GIMBAL_MANAGER_SET_ATTITUDE except for more human friendly tilt, pan fields (and no roll related fields)

* Possibly, largely for the sake of symmetry (the mathematics version of it I love so much), one also could add a new command DO_GIMBAL_MANAGER_ATTITUDE, which uses a quaternion (at the cost of no rates due to the 7 param limit). But I would agree that it might not be the most often used command, and it wouldn't be the biggest loss to world if it wouldn't be added.

:)

Cheers,
Olli

julianoes:
> However, IMHO, a very similar argument could be made also for GIMBAL_MANAGER_SET_ATTITUDE. In many if not most typical applications one would prefer to deal directly with tilt and pan angles and not with a quaternion. As example I mention a GCS doing stick control (which is also mentioned in the docs).

[...]

PR #1397, opened 2020-06-10
add GIMBAL_MANAGER_SET_TILTPAN, rename MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN

olliw42:
does https://github.com/mavlink/mavlink/issues/1376

for the units in GIMBAL_MANAGER_SET_TILTPAN I have chosen rad and rad/s. This is in contrast to the units in the MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN command, which uses deg and deg/s. 

Since GIMBAL_MANAGER_SET_TILTPAN is also a "high level" API like MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN one might argue that deg would be better than rad, but I went with rad here since rad is used also in the GIMBAL_MANAGER_INFORMATION message.

in MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN I also did some changes in the descriptions in the field,to make things more uniform across the series of GIMBAL_XXX messages/commands.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2226,14 +2226,14 @@
         <param index="1" label="Tag" minValue="0" increment="1">Target tag to jump to.</param>
         <param index="2" label="Repeat" minValue="0" increment="1">Repeat count.</param>
       </entry>
-      <entry value="1000" name="MAV_CMD_DO_GIMBAL_MANAGER_ATTITUDE" hasLocation="false" isDestination="false">
+      <entry value="1000" name="MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN" hasLocation="false" isDestination="false">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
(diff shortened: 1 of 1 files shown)


olliw42:
@julianoes @hamishwillee irrespective of gimbal manager details, renaming to MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN and adding GIMBAL_MANAGER_SET_TILTPAN makes IMHO sense in any case. It probably should be done rather sooner than later (at least the renaming).

hamishwillee:
@julianoes Seems fine to me, but you get to approve. Please update the docs to go with this. 

FWIW I have no strong feelings on deg vs radians; I know Lorenz treats anything in MAVLink as low level and should use the most convenient mathematical units. High level would be SDK.

olliw42:
commands are also used in missions
you know from scratch what 1.047 is?

hamishwillee:
@olliw42 "you know from scratch what 1.047 is?". No, but the GCS could easily do conversions. FWIW I have no personal conviction on this, other than a desire for following common patterns.

Anyway, Julian is happy, so I have merged.

olliw42:
great! THX.

PR #1398, opened 2020-06-10
common.xml: Add new Open Drone ID fields for classification

friissoren:
Signed-off-by: Soren Friis <soren.friis@intel.com>

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4066,6 +4066,9 @@
       <entry value="2" name="MAV_ODID_STATUS_AIRBORNE">
         <description>The UA is in the air.</description>
       </entry>
+      <entry value="3" name="MAV_ODID_STATUS_EMERGENCY">
+        <description>The UA is having an emergency.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4228,17 +4231,65 @@
         <description>Free-form text description of the purpose of the flight.</description>
       </entry>
     </enum>
-    <enum name="MAV_ODID_LOCATION_SRC">
-      <entry value="0" name="MAV_ODID_LOCATION_SRC_TAKEOFF">
+    <enum name="MAV_ODID_LOCATION_TYPE">
```
hamishwillee, line 4234: Is the location type always known?

friissoren: This enum is only used for the operator location, so yes, when the GCS/operator fills the location, it also knows the type and will fill that.
Maybe it would be more clear to rename this. I will do an update.

hamishwillee:
Thanks @friissoren ! Merged..

PR #1400, opened 2020-06-10
Correct description in ESTIMATOR_STATUS_FLAGS

olliw42:
It seems to me that there is no EKF_STATUS message anymore, but that it has been replaced with  ESTIMATOR_STATUS. So the description in ESTIMATOR_STATUS_FLAGS would be wrong.

Pl double check if I'm not just running wild. :)


hamishwillee:
Nope, you're on target. Thanks for this!

PR #1401, opened 2020-06-11
Add COMMAND_CANCEL - message to cancel long running commands

hamishwillee:
@julianoes @auturgy This adds message to cancel a long running command, as suggested by @LorenzMeier 

The message just takes a target sysid and compid, and the command being cancelled (not strictly needed if concurrency not supported, but a good check to make).

I don't have to implement this, so important you check the logic.

Essentially the sequence continues with progress updates until it completes - with one of acceptance, failure or cancellation. If the cancel message was sent, received and obeyed the flow would look like this.

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG4gICAgR0NTLT4-RHJvbmU6IENPTU1BTkRfTE9ORygpXG4gICAgR0NTLT4-R0NTOiBTdGFydCB0aW1lb3V0XG4gICAgRHJvbmUtPj5HQ1M6IENPTU1BTkRfQUNLKHJlc3VsdD1NQVZfUkVTVUxUX0lOX1BST0dSRVNTLHByb2dyZXNzPT8pXG4gICAgR0NTLT4-R0NTOiBTdGFydCAobG9uZ2VyKSB0aW1lb3V0XG4gICAgTm90ZSByaWdodCBvZiBHQ1M6IC4uLiBtb3JlIHVwZGF0ZXNcbiAgICBHQ1MtPj5Ecm9uZTogQ09NTUFORF9DQU5DRUwoY29tbWFuZClcbiAgICAgICAgR0NTLT4-R0NTOiBTdGFydCB0aW1lb3V0XG4gICAgRHJvbmUtPj5HQ1M6IENPTU1BTkRfQUNLKHJlc3VsdD1NQVZfUkVTVUxUX0NBTkNFTExFRCkiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ)

If no cancellation or other completion ACK is received, the cancel request can be retried
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG4gICAgR0NTLT4-RHJvbmU6IENPTU1BTkRfTE9ORygpXG4gICAgR0NTLT4-R0NTOiBTdGFydCB0aW1lb3V0XG4gICAgRHJvbmUtPj5HQ1M6IENPTU1BTkRfQUNLKHJlc3VsdD1NQVZfUkVTVUxUX0lOX1BST0dSRVNTLHByb2dyZXNzPT8pXG4gICAgR0NTLT4-R0NTOiBTdGFydCAobG9uZ2VyKSB0aW1lb3V0XG4gICAgTm90ZSByaWdodCBvZiBHQ1M6IC4uLiBtb3JlIHVwZGF0ZXNcbiAgICBHQ1MtPj5Ecm9uZTogQ09NTUFORF9DQU5DRUwoY29tbWFuZClcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXRcbiAgICBEcm9uZS0-PkdDUzogQ09NTUFORF9BQ0socmVzdWx0PU1BVl9SRVNVTFRfSU5fUFJPR1JFU1Mgb3IgTUFWX1JFU1VMVF9BQ0NFUFRFRClcbiAgICBHQ1MtPj5Ecm9uZTogQ09NTUFORF9DQU5DRUwoY29tbWFuZClcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXRcbiAgICBEcm9uZS0-PkdDUzogQ09NTUFORF9BQ0socmVzdWx0PU1BVl9SRVNVTFRfSU5fQ0FOQ0VMTEVEKSIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

If the operation has completed, the cancel message can be ignored. 

Does this make sense?

- [ ] Do we need a capability bit to indicate cancellation is supported? I guess "yes" because we can't assume handlers just check for "not in progress" to decide the sequence is done. Wish we had microservices.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5178,7 +5178,7 @@
       <field type="uint8_t" name="target_system">WIP: System which requested the command to be executed</field>
       <field type="uint8_t" name="target_component">WIP: Component which requested the command to be executed</field>
     </message>
-    <message id="78" name="COMMAND_CANCEL">
+    <message id="80" name="COMMAND_CANCEL">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to the original command with result=MAV_RESULT_CANCELLED if the long running process was cancelled. If it has already completed, the cancel action can be ignored. The cancel action can be retried until some sort of acknowledgement to the original command has been received. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5116,7 +5116,9 @@
       <field type="uint8_t" name="target_component">WIP: Component which requested the command to be executed</field>
     </message>
     <message id="78" name="COMMAND_CANCEL">
-      <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to original command with result=MAV_RESULT_CANCELLED if the long running process was cancelled. If it has already completed, the cancel action can be ignored. The cancel action can be retried until some sort of acknowledgement to the original command has been received. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to the original command with result=MAV_RESULT_CANCELLED if the long running process was cancelled. If it has already completed, the cancel action can be ignored. The cancel action can be retried until some sort of acknowledgement to the original command has been received. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
       <field type="uint8_t" name="target_system">System executing long running command. Should not be broadcast (0).</field>
       <field type="uint8_t" name="target_component">Component executing long running command.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID (of command to cancel).</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2850,7 +2850,7 @@
       <entry value="5" name="MAV_RESULT_IN_PROGRESS">
         <description>Command is valid and is being executed. This will be followed by further progress updates, i.e. the component may send further COMMAND_ACK messages with result MAV_RESULT_IN_PROGRESS (at a rate decided by the implementation), and must terminate by sending a COMMAND_ACK message with final result of the operation. The COMMAND_ACK.progress field can be used to indicate the progress of the operation.</description>
       </entry>
-      <entry value="5" name="MAV_RESULT_CANCELLED">
+      <entry value="6" name="MAV_RESULT_CANCELLED">
         <description>Command has been cancelled (as a result of receiving a COMMAND_CANCEL message).</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5116,7 +5116,7 @@
       <field type="uint8_t" name="target_component">WIP: Component which requested the command to be executed</field>
     </message>
     <message id="78" name="COMMAND_CANCEL">
-      <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to original command with result=MAV_RESULT_CANCELLED (whether or not an operation is running). Cancelling must always succeed. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
+      <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to original command with result=MAV_RESULT_CANCELLED (whether or not an operation is running). The command should retry if no MAV_RESULT_CANCELLED acknowledgement is recieved. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
       <field type="uint8_t" name="target_system">System executing long running command. Should not be broadcast (0).</field>
       <field type="uint8_t" name="target_component">Component executing long running command.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID (of command to cancel).</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5115,6 +5115,12 @@
       <field type="uint8_t" name="target_system">WIP: System which requested the command to be executed</field>
       <field type="uint8_t" name="target_component">WIP: Component which requested the command to be executed</field>
     </message>
+    <message id="78" name="COMMAND_CANCEL">
```
hamishwillee, line 5118: OK to use this ID?

julianoes: Doesn't look like it's taken in https://github.com/ArduPilot/mavlink/blob/master/message_definitions/v1.0/ardupilotmega.xml, so I assume it's ok.

message_definitions/v1.0/common.xml
```diff
@@ -5115,6 +5115,12 @@
...
+      <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to original command with result=MAV_RESULT_CANCELLED (whether or not an operation is running). The command should retry if no MAV_RESULT_CANCELLED acknowledgement is recieved. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
```
julianoes, line 5119: This would be my implementation suggestion:

```suggestion
      <description>Cancel a long running command. The target system should respond with a COMMAND_ACK to original command with result=MAV_RESULT_CANCELLED if the long running process was cancelled. If it has already completed, the cancel action can be ignored. The cancel action can be retried until some sort of acknowledgement to the original command has been received. The command microservice is documented at https://mavlink.io/en/services/command.html</description>
```

My thinking is to prevent potential races where the actor trying to cancel does not know if the cancel was successful or not (e.g. too late).

hamishwillee: Hi @julianoes 

I _think_ precisely the opposite - there can't be a race condition the way it is because even after the request is accepted, the actor trying to do the cancelled request will still get the cancelled ack.

The way this is now worded, if you send a cancel and the receiving system has just completed then it will ignore the cancel command. You could then resend that cancel multiple times and it would do nothing: neither ack or cancel. But perhaps I don't understand the sequence you are proposing. 

The only problem I see here is actually separate to cancellation. For a short running request if you don't get an ACK then you resend the request.  So you are guaranteed an ACK.
But for the long running case if you drop the ACK there is no way to check if you missed it or not because you don't resend the request. 

So it may be that we should store completion status of the last request and the cancellation request and return either cancelled or accepted state if we receive a resend of the cancellation request. Note though, the current request doesn't support resends - ie  it doesn't have fields to say I'm a resend of an earlier message.

What am I missing. I think I might need sequences showing what you plan.

hamishwillee: Or to summarise, we need to make sure that we always get a completion ack of some kind, and we need some way to make sure that if it didn't arrive we can check it.

julianoes: I see, it's tricky.

My view comes from a users perspective:

> I think precisely the opposite - there can't be a race condition the way it is because even after the request is accepted, the actor trying to do the cancelled request will still get the cancelled ack.

This is wrong and confusing for a user. Let's say I am starting an update and then I see the progress as it goes. Then I decide to cancel it. What I'm now interested is if the update succeeded or was actually cancelled. If I receive CANCELLED no matter what I don't know what actually happened. If I get ACCEPTED I know that my request to cancel did not have an effect and the command went through anyway. If I get CANCELLED, I can be sure that the update did not go through and I'm still on the old version.

message_definitions/v1.0/common.xml
hamishwillee: OK. I'll stick with that then.

hamishwillee:
@julianoes Thanks very much. Good point. I have accepted your suggestion as it makes sense - the sequence can only complete once, and with one result. 

1. How do we handle the case where no completion ACK received - ie say either a cancel, accepted or fail was sent but not received. Or to put it another way, what happens following our IN_PROGRESS timeout? We haven't specified this in [the protocol](http://mavlink.io/en/services/command.html)?
The easiest thing would be to perhaps just note "unknown problem" in UI and move back to idle.

1. Do we use a capability bit to indicate this new option? Microservices versioning or a metadata to indicate this message supported would be better. I kind of think that we could ignore this for now - if the command isn't supported by a system nothing should change for the recipient. The sender of course will just keep resending the cancel and it will be ignored.

hamishwillee:
@julianoes When you get a chance - new related question above: https://github.com/mavlink/mavlink/pull/1401#issuecomment-643839421

hamishwillee:
@julianoes I can merge this, but I'd quite like to resolve what we do if the final Cancel or Fail or Accepted ACK is missed?

If not, then can merge this and defer to the dev call.

julianoes:
@hamishwillee sorry I missed that:

1. I would say that's a timeout. In the same way that no answer after sending a command is a timeout.
2. Yes, I think it's fine if cancel is ignored. There are not many (if any) commands yet that are long running and would need this.

hamishwillee:
@julianoes

> I would say that's a timeout. In the same way that no answer after sending a command is a timeout.

The difference is, if there is a timeout after sending a command you can re-request it.

So to be very clear, you're saying that if a component misses the final ACK (for any reason) it will timeout and return to idle (presumably notifying the user that the operation timed out).

The only other way I can see doing this is another message to request a progress update - the system then returns either the current progress or the **last sent ACK**.  

Lets chat in the call.

hamishwillee:
Merging the cancel message as WIP. Associated docs in https://github.com/mavlink/mavlink-devguide/pull/243

PR #1402, opened 2020-06-11
MAV_CMD_NAV_LOITER_TIME: xtrack, improve definition

hamishwillee:
This follows from discussion of xtrack in https://github.com/mavlink/mavlink/pull/1383#issuecomment-637952305 (and the dev call). 

@auturgy Does this capture what you meant? I hope to have diagrams in the guide too.

@sfuhrer This clarifies what the angles and "system yaw" mean. The confusion before was because the value defines **either** the xtrack path or the xtrack exit location, depending on value.

The Xtrack is the path that the vehicle will take/converge towards between the current and next waypoints. The center xpath is the path between the centers of the waypoints. So if you set 0 you're not saying where you'll exit the path, just that you'll converge towards the center.

If you set 1 you're defining the xpath to be from where the vehicle leaves the loiter to the center of the next waypoint. 

Any other value is an exit location to the center xpath - it if 45, you're saying that the vehicle will leave the loiter when the angle from the loiter tangent to the center xpath is 45 degrees, and converge to the center xpath. This is actually useful :-). Make sense?


sfuhrer:
> Any other value is an exit location to the center xpath - it if 45, you're saying that the vehicle will leave the loiter when the angle from the loiter tangent to the center xpath is 45 degrees, and converge to the center xpath. This is actually useful :-). Make sense?

Aha, I had thought that the manually entered value only refers to MC (=desired yaw after leaving the loiter). So how does this then now work for MC (as then you don't have any loiter circle).

hamishwillee:
@sfuhrer The xtrack explicitly only applies to forward moving vehicles "For forward-moving aircraft this sets xtrack path or exit location". So "not applicable".

This otherwise looks OK?

message_definitions/v1.0/common.xml
```diff
@@ -1421,11 +1421,11 @@
...
-        <param index="1" label="Heading Required" minValue="0" maxValue="1" increment="1">Heading Required (0 = False)</param>
-        <param index="2" label="Radius" units="m">Radius. If positive loiter clockwise, negative counter-clockwise, 0 means no change to standard loiter.</param>
+        <description>Begin loiter at the specified Latitude and Longitude.  If Lat=Lon=0, then loiter at the current position.  Don't consider the navigation command complete (don't leave loiter) until the altitude has been reached. Additionally, if the Heading Required parameter is non-zero the aircraft will not leave the loiter until heading toward the next waypoint.</description>
+        <param index="1" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter circle only once heading towards the next waypoint (0 = False)</param>
```
hamishwillee, line 1425: MAV_CMD_NAV_LOITER_TO_ALT is inconsistent - all the other loiters have heading at 2, and radius at 1. Is it possible to change this for consistency? @auturgy @sfuhrer

message_definitions/v1.0/common.xml
```diff
@@ -1333,19 +1333,19 @@
       <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint for X turns</description>
         <param index="1" label="Turns" minValue="0">Number of turns.</param>
-        <param index="2">Empty</param>
-        <param index="3" label="Radius" units="m">Radius around waypoint. If positive loiter clockwise, else counter-clockwise</param>
-        <param index="4" label="Xtrack Location">Forward moving aircraft this sets exit xtrack location: 0 for center of loiter wp, 1 for exit location. Else, this is desired yaw angle. NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
+        <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter circle only once heading towards the next waypoint (0 = False)</param>
```
hamishwillee, line 1336: @sfuhrer Note, added heading here as was missing. OK?

message_definitions/v1.0/common.xml
hamishwillee: @magicrub @WickedShell @sfuhrer Would it be possible/manageable to change the MAV_CMD_NAV_LOITER_TO_ALT to move the param 1 and param 2 down one slot. 
This would make the user of params consistent across all the loiter messages.

By no means critical of course, so this is an "estimate of impact".

sfuhrer: I would also prefer to have it consistent, and feel like impact is acceptable.

message_definitions/v1.0/common.xml
sfuhrer: Guess that makes perfect sense, yes :+1:

message_definitions/v1.0/common.xml
hamishwillee: OK. I'll merge this and create a separate dedicated PR for discussion.

PR #1403, opened 2020-06-14
common: distance_sensor: add signal_quality field

TSC21:
Adds a relevant field to the `distance_sensor` message, representing the quality of the sensor signal. If the sensor is external to the flight controller, meaning, not a device connected to the flight controller board and having the reading processed in a driver, but instead connected to a companion computer, even more relevant it becomes to have a way of sending this data, considering we are to process the raw data in the flight controller side. The inverse direction stream, meaning from the flight controller to a ground station for example, is also relevant for sensor debugging. For the simulated sensors, which data is being streamed using MAVLink, this is also a relevant field.

Solves https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_receiver.cpp#L719.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5676,6 +5676,7 @@
       <field type="float" name="horizontal_fov" units="rad">Horizontal Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to 0.</field>
       <field type="float" name="vertical_fov" units="rad">Vertical Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to 0.</field>
       <field type="float[4]" name="quaternion">Quaternion of the sensor orientation in vehicle body frame (w, x, y, z order, zero-rotation is 1, 0, 0, 0). Zero-rotation is along the vehicle body x-axis. This field is required if the orientation is set to MAV_SENSOR_ROTATION_CUSTOM. Set it to 0 if invalid."</field>
+      <field type="uint8_t" name="signal_quality" units="%">Signal quality of the sensor. Specific to each sensor type, representing the relation of the signal strength with the target reflectivity, distance, size or aspect, but normalised as a percentage. 0 = unknown/unset signal quality, 1 = invalid signal, 100 = perfect signal.</field>
     </message>
     <message id="133" name="TERRAIN_REQUEST">
       <description>Request for terrain data and terrain status</description>


message_definitions/v1.0/common.xml
```diff
@@ -5676,6 +5676,7 @@
       <field type="float" name="horizontal_fov" units="rad">Horizontal Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to 0.</field>
       <field type="float" name="vertical_fov" units="rad">Vertical Field of View (angle) where the distance measurement is valid and the field of view is known. Otherwise this is set to 0.</field>
       <field type="float[4]" name="quaternion">Quaternion of the sensor orientation in vehicle body frame (w, x, y, z order, zero-rotation is 1, 0, 0, 0). Zero-rotation is along the vehicle body x-axis. This field is required if the orientation is set to MAV_SENSOR_ROTATION_CUSTOM. Set it to 0 if invalid."</field>
+      <field type="int8_t" name="signal_quality" units="%">Signal quality of the sensor. Specific to each sensor type, representing the relation of the signal strength with the target reflectivity, distance, size or aspect, but normalised as a percentage. 0 = invalid signal, 100 = perfect signal, and -1 = unknown signal quality.</field>
```
hamishwillee, line 5679: Generally the 0 value should mean "information not sent/you can ignore this value". That's because [extensions](http://mavlink.io/en/guide/define_xml_element.html#message_extensions) sent values default to zero if the sender does not know about the field - so a recipient needs to know that it should ignore a zero value. Can this be modified such that 0 is invalid signal OR not supported?

@auturgy @julianoes Does this make sense to you? I don't know enough about what a recipient would be able to do with this data to understand how useful it is.

julianoes: Isn't this potentially the same as the `covariance` field above? Or could `covariance` at least be used for it?

auturgy:
are there any distance sensors that provide this, or is this just a solution waiting for a problem?
I understand the intent here, just want to make sure that if it's implemented, it at least has a reference implementation.

message_definitions/v1.0/common.xml
TSC21: > Generally the 0 value should mean "information not sent/you can ignore this value". That's because extensions are received as zero if the sender does not know about the field - so a recipient needs to know 0 always means "ignore this". Can this be modified such that 0 is invalid signal OR not supported?

@hamishwillee the problem is that in this case we do need to split between a known quality invalid signal (=0) and marking the quality as unknown. Maybe use `uint8_t` and set it to 255, considering that this is supposed to be normalised 0.100?

TSC21: > Isn't this potentially the same as the `covariance` field above? Or could `covariance` at least be used for it?

Not in this case. The covariance should come of the statistical distribution of the measurements (distances), while the signal quality comes from normalised signal strength.

TSC21:
> are there any distance sensors that provide this, or is this just a solution waiting for a problem?
> I understand the intent here, just want to make sure that if it's implemented, it at least has a reference implementation.

Yes you have for example the signal strength in the Lidar Lites, and this basically picks that up and normalises that value.

message_definitions/v1.0/common.xml
auturgy: > > Generally the 0 value should mean "information not sent/you can ignore this value". That's because extensions are received as zero if the sender does not know about the field - so a recipient needs to know 0 always means "ignore this". Can this be modified such that 0 is invalid signal OR not supported?
> 
> @hamishwillee the problem is that in this case we do need to split between a known quality invalid signal (=0) and marking the quality as unknown. Maybe use `uint8_t` and set it to 255, considering that this is supposed to be normalised 0.100?

To reinforce Hamish’s gentle nudge: use 0 for not used, to allow trimming. This isn’t going to be a frequently used field, so let’s not force it to be sent unnecessarily.

auturgy:
> > are there any distance sensors that provide this, or is this just a solution waiting for a problem?
> > I understand the intent here, just want to make sure that if it's implemented, it at least has a reference implementation.
> 
> Yes you have for example the signal strength in the Lidar Lites, and this basically picks that up and normalises that value.

Sure. And the Benewake protocol has a “reliability” field you could fudge into this if so inclined. 
But why push it over mavlink? It’s useful to log on the vehicle, but when is an offboard system going to need this data in real-time? How is it used?

TSC21:
> > > are there any distance sensors that provide this, or is this just a solution waiting for a problem?
> > > I understand the intent here, just want to make sure that if it's implemented, it at least has a reference implementation.
> > 
> > 
> > Yes you have for example the signal strength in the Lidar Lites, and this basically picks that up and normalises that value.
> 
> Sure. And the Benewake protocol has a “reliability” field you could fudge into this if so inclined.
> But why push it over mavlink? It’s useful to log on the vehicle, but when is an offboard system going to need this data in real-time? How is it used?

As explained, this is can also be brought from the simulation side. I brought the logging as an option as well, and this doesn't actually need to be logged in the vehicle, can also be brought to QGC.

BazookaJoe1900:
@TSC21  great!
if we can make SIM_GPS_BLOCK as bit field parameter that can block either of the multi GPS signal, that will be nice

message_definitions/v1.0/common.xml
TSC21: > use 0 for not used, to allow trimming

@hamishwillee @auturgy done.

hamishwillee:
I have no problem with the message itself. 

@auturgy convinced by the use case? [FWIW I don't completely "get" the benefit of exporting from simulator if that is the selling point here - is it that you can emit the value from a simulator so log it on PX4?]

hamishwillee:
PS Build failed. I have restarted but if it fails again please do a check to see if we broke anything.

TSC21:
> PS Build failed. I have restarted but if it fails again please do a check to see if we broke anything.



> PS Build failed. I have restarted but if it fails again please do a check to see if we broke anything.

https://github.com/mavlink/mavlink/issues/1404

hamishwillee:
Thanks. I've asked PeterB if he knows about the issue.

auturgy:
He does. Just hit it with one of his own pr’s

TSC21:
> [FWIW I don't completely "get" the benefit of exporting from simulator if that is the selling point here - is it that you can emit the value from a simulator so log it on PX4?]

Consider that the simulator functions as a driver for each sensor - what the driver provides to you mostly is raw data and/or other information that your propagate in the flight controller internals, to arrive to the respective places where this data is to be processed and handled. In the case of PX4, EKF2 uses the signal quality of the range sensors to evaluate if the data is good to go or not. Since the simulator depends on the MAVLink stream to propagate its "driver" data, this is also an information that needs to be provided from the "driver" in the simulator, to the PX4 side. This is used for more than logging, and it's relevant for testing the flight controller behaviour to signal quality changes on the range sensors.

TSC21:
Thanks @hamishwillee!

Issue #1345, opened 2020-04-14
Reserve Component ID 68 for RADIO


olliw42:
The SiK radios, which are still in wide-spread use, for historical reasons send out RADIO_STATUS messages with source id 51 and component ID 68.

Component ID 68 is however marked in the spec as MAV_COMP_ID_USER44, i.e. as an id freely chosable by a user.

I suggest to remove MAV_COMP_ID_USER44, and replace it with MAV_COMP_ID_RADIO, in order to put things on a proper formal basis (and avoid pitfalls even if admittedly unlikely)

Thanks
Olli

hamishwillee:
I think this makes sense.

julianoes:
Make sense to me.

olliw42:
for your convenience: https://github.com/mavlink/mavlink/pull/1405

:)

olliw42:
done. THX!
hence closing

PR #1405, opened 2020-06-16
replace MAV_COMP_ID_USER44 with MAV_COMP_ID_RADIO

olliw42:
does https://github.com/mavlink/mavlink/issues/1345

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -514,7 +514,7 @@
         <description>Id for a component on privately managed MAVLink network. Can be used for any purpose but may not be published by components outside of the private network.</description>
       </entry>
       <entry value="68" name="MAV_COMP_ID_RADIO">
-        <description>Telemetry radio component that sends out RADIO_STATUS messages (e.g. SiK radios).</description>
+        <description>Telemetry radio (e.g. SiK radio, or other component that emits RADIO_STATUS messages).</description>
       </entry>
       <entry value="69" name="MAV_COMP_ID_USER45">
         <description>Id for a component on privately managed MAVLink network. Can be used for any purpose but may not be published by components outside of the private network.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -514,7 +514,7 @@
         <description>Id for a component on privately managed MAVLink network. Can be used for any purpose but may not be published by components outside of the private network.</description>
       </entry>
       <entry value="68" name="MAV_COMP_ID_RADIO">
-        <description>Radio which sends out RADIO_STATUS messages, such as the SiK radios which send with source ID 51 and component ID 68.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -513,8 +513,8 @@
       <entry value="67" name="MAV_COMP_ID_USER43">
         <description>Id for a component on privately managed MAVLink network. Can be used for any purpose but may not be published by components outside of the private network.</description>
       </entry>
-      <entry value="68" name="MAV_COMP_ID_USER44">
-        <description>Id for a component on privately managed MAVLink network. Can be used for any purpose but may not be published by components outside of the private network.</description>
+      <entry value="68" name="MAV_COMP_ID_RADIO">
+        <description>Telemetry radio component that sends out RADIO_STATUS messages (e.g. SiK radios).</description>
```
hamishwillee, line 517: @olliw42 Thanks very much. 

Is the fact that it sends out RADIO_STATUS messages key to the definition, or are we being too specific. I.e. we might reasonably just have "Telemetry radio (e.g. SiK radio)" and this could then be used  for other kinds of radios.

olliw42: that's a very good question. I think I fully agree in principle that it shouldn't be different to any other, and it would be great if it would be non specific. Like GIMBAL = gimbal, SERVO = servo, etc, which would suggest RADIO = telemetry radio. Sending RADIO_STATUS is probably what makes a component a RADIO component, but that's so for e.g. a camera too, and it's not said in the ID definition. 

However, there is some historical specific use of this ID, and if another thing would use that ID in different ways it most likely would break the existing, wide-spread GCS and/or autopilot implementations. That is, these other kinds of radios are not as free in what they can do as other standard IDs. And I think this knowledge should not be lost. "Radio" should also not be confused with other things colloquially named radio, e.g. a transmitter. 

Use of this ID should not break the existing implementations which rely on it to identify a very specific thing. Not sure how to best put it into language. I tried it by being somewhat specific. Might not have been the best attempt.

hamishwillee: OK, so how about we compromise a little - generic definition, more precise example

> Telemetry radio (e.g. SiK radio, or other system that emits RADIO_STATUS messages).

olliw42: sounds great to me!

replaced system with component

hamishwillee: Awesome. I also just changed the component name to include "telemetry". I think it helps to be that specific (unless you strongly disagree).

olliw42: :+1:

hamishwillee:
OK, IMO this is good to merge once we resolve the issues (unrelated) with CI. Thanks @olliw42

hamishwillee:
Thanks @olliw42 . Merging/

olliw42:
THX!

PR #1408, opened 2020-06-21
Updates for new COMPONENT_INFORMATION support

DonLakeFlyer:
Changes for #1346

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1295,6 +1295,15 @@
         <description>WiFi disabled.</description>
       </entry>
     </enum>
+    <enum name="COMP_METADATA_TYPE">
+      <description>Possible values for COMPONENT_INFORMATION.comp_metadata_type.</description>
+      <entry value="0" name="COMP_METADATA_TYPE_VERSION">
+        <description>Version information which also includes information on other optional supported COMP_METADATA_TYPE's. Must be supported. Only downloadable from vehicle.</description>
+      </entry>
+      <entry value="1" name="COMP_METADATA_TYPE_PARAMETER">
+        <description>Parameter meta data.</description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM and MISSION_ITEM_INT messages) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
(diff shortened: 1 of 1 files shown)


hamishwillee:
@olliw42 FYI changes to component_information

message_definitions/v1.0/common.xml
```diff
@@ -6672,13 +6681,11 @@
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Information about a component. For camera components instead use CAMERA_INFORMATION, and for autopilots use AUTOPILOT_VERSION. Components including GCSes should consider supporting requests of this message via MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint8_t[32]" name="vendor_name">Name of the component vendor</field>
```
hamishwillee, line 6675: @DonLakeFlyer @julianoes The theory is that there will be metadata to get the values of these fields. 
1. Removing these before said metadata exists may well break something more seriously than just "I need to update my messages".
2. There is a question about what should be in metadata and what should be in the message itself. The theory is that the message should have stuff that everyone will need - getting the metadata should be a choice. [So for camera, the common settings are in message, and the extended are in metadata]. Is this stuff something that we should not force users to implement the whole mavftp etc infrastructure to get. 

I'm to some extent playing devils advocate, but:
1. For next time perhaps add our new fields as extensions. Then remove the fields when we have a complete story.
2. Can we also add an enum and metadata for this info?

olliw42:
@hamishwillee thx for the FYI. IMHO this should win the gold medal for the greatest nonsense of the year. I can't believe it, what a nonsense. This could be achieved much easier by, e.g., simply reading the ftp dir without any need of all this cruft and overhead. It's sad to see a good idea turned into a terrible result. IMO.

message_definitions/v1.0/common.xml
```diff
@@ -6672,13 +6681,11 @@
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
```
hamishwillee, line 6681: @DonLakeFlyer If using MAV_CMD_REQUEST_MESSAGE, what param maps to the metadata type you are after?

hamishwillee:
@olliw42 The authors obviously thought the "cruft" had purpose. This provides:
- location of metadata
- ability to query what metadata is supported
- ability to query specific metadata
- Ability to query whether file has been updated and needs to be re-queried.
- Information about whether translations files exist and need to be re-queried. 

Yes you could have a folder for component metadata, perhaps hard coded and allow the vehicle to "see what is there". Is that what you were envisaging?

Or to put it another way, if you want to educate, perhaps you should expand a bit on how your solution would allow us to do the same kinds of things (or why the kinds of things we want to do are unnecessary).

PR #1412, opened 2020-06-24
BATTERY_STATUS.voltage: 10 cells to 12 cells

muramura:
A 12-cell battery is available for personal purchase.
12-cell batteries are being majored in by DJI.
The ArduPilot is a 12-cell battery with a voltage of cell 10 to cell 10
The system is designed to set the minimum voltage at 11 and 12.
I would prefer to be able to transmit all cell voltages.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5872,6 +5872,7 @@
       <extensions/>
       <field type="int32_t" name="time_remaining" units="s">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
We shouldn't do this because it breaks the message, forcing everyone to upgrade. Further, we're adding 2 cells, but there are already 14s batteries out there.

So proposal is to add a new `voltage_ext` field as an extension (or similar) at the end of the current extensions. This should have at least 4 slots (how many can be fit?) and the zero value will indicate "information not supplied). This allows the data to be truncated if not supplied. If the voltage is measured as zero then the system should send 1, and the recipient should assume that 1 and zero are synonymous.

muramura:
@hamishwillee san. I understand. At the end of this message, the voltage_ext is an array of four arrays of type uint16_t Add in.

muramura:
@hamishwillee san.  I've changed.

hamishwillee:
@muramura Thanks. I couldn't edit or merge this for some reason (weird, since I have some admin rights). Anyway, to allow the needed updates, I recreated in https://github.com/mavlink/mavlink/pull/1423 - can you please check that meets your requirements?

PR #1413, opened 2020-06-24
MAV_CMD_NAV_LOITER_TO_ALT - move heading/radius params for consistency

hamishwillee:
All the other MAV_CMD_NAV_LOITER_ messages have the same patterns for the loiter radius (and heading required) params, if present. This moves MAV_CMD_NAV_LOITER_TO_ALT to match.

I think consistency is desirable but of course this will have a code impact. Do you implement it? Is this something you see as worth the cost? 

@magicrub @WickedShell @sfuhrer @DonLakeFlyer @auturgy @meee1


WickedShell:
I don't think it's worth the cost, as it will be surprisingly annoying in terms of versioning, and it will definitely affect people, and be a challenge to make obvious enough in the changelogs/versioning everywhere. Given the GCS has to present the Ux for the item anyways I don't think that's a strong enough reason to rearrange the order, as the GCS is already responsible for sanely mapping input to params, so this should be mostly non user facing already, or at least assisted with labels.

DonLakeFlyer:
This can't be changed. It's used in missions which have already been persisted to disk.

WickedShell:
> This can't be changed. It's used in missions which have already been persisted to disk.

ArduPilot will be fine for that with anything already uploaded to the autopilot, but it will definitely break in any missions which people have preplanned and saved to their computers, which is even harder to manage.

hamishwillee:
Thanks guys. Closing. No action.

PR #1414, opened 2020-06-26
Add auxiliary actuator commands

coderkalyan:
This PR adds 2 commands for controlling auxiliary actuators (primarily designed for servos). MAV_CMD_DO_SET_ACTUATOR_AUX replaces MAV_CMD_DO_SET_SERVO, being less ambiguous and more flexible. MAV_CMD_DO_CYCLE_ACTUATOR_AUX replaces MAV_CMD_DO_REPEAT_SERVO but adds more functionality.

@bkueng @hamishwillee let me know if these messages make sense. I have gone a little extra with the CYCLE_ACTUATOR_AUX command, but let me know if you're fine with it or would prefer not to have this message. I think it could be useful in many cases, potentially even my own payload drop mechanism.

Note: For the PX4 implementation, if control group 3 is used, only the first 3 values will be mapped to actually do anything. I am simply adding 7 for future proofing/because I can :joy:

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1304,27 +1304,6 @@
         <description>Parameter meta data.</description>
       </entry>
     </enum>
-    <enum name="ACTUATOR_CYCLE_INTERPOLATION">
-      <description>Type of interpolation to use when cycling actuator values using MAV_CMD_DO_CYCLE_ACTUATOR</description>
-      <entry value="0" name="ACTUATOR_CYCLE_INTERPOLATION_CONSTANT">
-        <description>Toggles between the two cycle values, no interpolation done.</description>
-      </entry>
-      <entry value="1" name="ACTUATOR_CYCLE_INTERPOLATION_LINEAR">
-        <description>Linear interpolation between the two cycle values.</description>
-      </entry>
-      <entry value="2" name="ACTUATOR_CYCLE_INTERPOLATION_QUADRATIC">
-        <description>Interpolates between the two cycle value with a quadratic curve.</description>
-      </entry>
-      <entry value="3" name="ACTUATOR_CYCLE_INTERPOLATION_CUBIC">
-        <description>Interpolates between the two cycle value with a cubic curve.</description>
-      </entry>
-      <entry value="4" name="ACTUATOR_CYCLE_INTERPOLATION_QUARTIC">
-        <description>Interpolates between the two cycle value with a quartic curve.</description>
-      </entry>
-      <entry value="5" name="ACTUATOR_CYCLE_INTERPOLATION_SINUSOIDAL">
-        <description>Interpolates between the two cycle value with a sinusoidal curve.</description>
-      </entry>
-    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
@@ -1747,15 +1726,6 @@
         <param index="5" label="Actuator 5" minValue="-1" maxValue="1">Actuator 5 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
-      </entry>
-      <entry value="188" name="MAV_CMD_DO_CYCLE_ACTUATOR" hasLocation="false" isDestination="false">
-        <description>Cycle an actuator between two settings for a desired number of cycles with a desired period.</description>
-        <param index="1" label="Instance" minValue="1" increment="1">Actuator instance number (0 to control all actuators).</param>
-        <param index="2" label="Value 1" minValue="-1" maxValue="1">Actuator value 1, scaled from -1 to 1. NaN to use "nominal" value.</param>
-        <param index="3" label="Value 2" minValue="-1" maxValue="1">Actuator value 2, scaled from -1 to 1. NaN to use "nominal" value.</param>
-        <param index="4" label="Count" minValue="1" increment="1">Cycle counts.</param>
-        <param index="5" label="Period" minValue="0" units="s">Period of cycle.</param>
-        <param index="6" label="Interpolation" enum="ACTUATOR_CYCLE_INTERPOLATION">Interpolation to use for cycle values.</param>
       </entry>
       <entry value="189" name="MAV_CMD_DO_LAND_START" hasLocation="true" isDestination="false">
         <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts. It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used. The Latitude/Longitude is optional, and may be set to 0 if not needed. If specified then it will be used to help find the closest landing sequence.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1718,6 +1718,8 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="187" name="MAV_CMD_DO_SET_ACTUATOR" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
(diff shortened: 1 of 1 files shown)


coderkalyan:
Update: as for the cycle, I'm not sure exactly what would be the best way to "cancel" it, especially if it is set to loop forever. The existing REPEAT_SERVO message does not seem to indicate what to do either.

hamishwillee:
Looks cool to me, but I want @bkueng general opinion before progressing.

bkueng:
I do not see this as a replacement for `MAV_CMD_DO_SET_SERVO`, as they act on different levels (before or after mixing).
I realize that AUX is confusing here, as some understand it as the AUX ports. This is not what I meant though, as it does not refer to any physical pins, but to mixer input channels (for example [feed-through of RC AUX1 channel](https://dev.px4.io/v1.9.0/en/airframes/airframe_reference.html#quadrotor-x), and gives full flexibility on which physical outputs it's actually used). The AUX naming might be a bit PX4-specific and if we can generalize further I'm fine with that (it's on the same level as roll-command for example). But I also wanted it to be a bit specific so it's easy to use and implement.

hamishwillee:
@bkueng This feels very PX4-specific - Do the concepts of mixer, and RC AUX passthrough channels even appear in all the other flight stacks we know about? I'm not familiar with them from ArduPilot (but I could be wrong on that).

We can't have flight-stack specific commands in  common.xml. 

But perhaps I simply don't understand how this will work or am being naive- ie if I set actuator 1 and it is before the mixer then  it maps to an input group. That means the output could actually be output to anything. That might be OK, but we need that to be very clear in the documentation for the message.

coderkalyan:
@bkueng @hamishwillee What if we make do with DO_SET_SERVO on PX4, and just document the fact that it's usage is a bit non standard - then the folks already using that message won't be affected - then focus on designing a more future proof set of messages for actuator control? I feel like with the exception of gimbals, actuator control for payloads or anything else that's not part of the main flight control is lacking, maybe @autergy has some ideas about that?

hamishwillee:
@coderkalyan The discussion on DO_SET_SERVO should not be held here. That said, sure, why not. 
Nothing changes in MAVLink we just provide an explanation of how it is used in PX4. That doesn't change anything in MAVLink because the reason we have this problem is that the definition is so loose.

I was kind of hoping we would do that "as well".

hamishwillee:
@bkueng I _think_ there is confusion over the intent. I have been interpreting "actuator 3" as the 3rd output on the AUX (either UAVCAN or PWM), and that isn't sufficiently flexible. 

But what I think you actually saying is that each actuator is a "designated" servo output that is being mapped by some other mechanism (e.g. a parameter), and which could correspond to any output you want on UAVCAN, MAIN, AUX PWM or whatever. 

This isn't what I expected, but given the use case it is actually better:
- You don't worry in mavlink where the servo is mapped. 
- You can map it whereever you like, without changing the message.
- It scales well. You probably don't need 16 items because you only map the ports you actually know are going to be used for servos
- Because you know which servos are mapped the checking on what is sensible and safe is easier - you don't need to check that AUX3 is free and safe - you check that the parameter is assigned. 

Was that closer to your intent?

So the message might look like below - not sure if we really need the index now.

<entry value="187" name="MAV_CMD_DO_SET_ACTUATOR" hasLocation="false" isDestination="false">
        <description>Sets actuators (e.g. servos) to a desired value. The actuator numbers are mapped to specific outputs (e.g. on any MAIN or AUX PWM or UAVCAN) using a flight-stack specific mechanism (i.e. a parameter).</description>
        <param index="1" label="Actuator 1" minValue="-1" maxValue="1">Actuator 1, scaled from [-1 to 1]. NaN to ignore. If index param is non-zero, this is Actuator "6*index+1"</param>
        <param index="2" label="Actuator 2" minValue="-1" maxValue="1">Actuator index+1 value, scaled from [-1 to 1]. NaN to ignore.</param>
        <param index="3" label="Actuator 3" minValue="-1" maxValue="1">Actuator index+1 value, scaled from [-1 to 1]. NaN to ignore.</param>
        <param index="4" label="Actuator 4" minValue="-1" maxValue="1">Actuator index+2 scaled from [-1 to 1]. NaN to ignore.</param>
        <param index="5" label="Actuator 5" minValue="-1" maxValue="1">Actuator index +3 value, scaled from [-1 to 1]. NaN to ignore.</param>
        <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator index + 4 value, scaled from [-1 to 1]. NaN to ignore.</param>
        <param index="7" label="index" minValue="0" increment="1">Index of actuator set (i.e. if non-zero, Actuator 1 is actually actuator 7)</param>
      </entry>

bkueng:
> @bkueng @hamishwillee What if we make do with DO_SET_SERVO on PX4, and just document the fact that it's usage is a bit non standard

Fine with me too.

@hamishwillee yes you got it!
It's mostly about the naming and how you map it exactly that is autopilot-specific. For example ardupilot has RC passthrough controls/channels as well: https://ardupilot.org/copter/docs/common-rcoutput-mapping.html#rcpassthru.

The message looks good - I'm not sure we need the index though (and if I'd call it first_index or index_offset).

auturgy:
Hamish has almost described rc_override.

hamishwillee:
@auturgy Sure, except it isn't tied to the constraints of mapping an RC receiver. Or to put it another way, this message still looks sensible to me, and a significant step up from SET_SERVO. Or am I missing something?

auturgy:
We generally refrain from introducing new messages/commands where the outcome can sensibly be achieved with an existing one.
For inputs to the mixer, we currently have manual_control and rc_override. For direct output control we have do_set_servo (and a bunch of device/peripheral specific stuff for cameras/grippers/package delivery/gimbals/undercarriage/shortly a sprayer, etc).

I agree that this is an area that can be reworked and improved, as there’s a lot of legacy rc and pwm influenced stuff that should be factored out. I’m not convinced that the proposed design is enough of an improvement though: particularly as there is apparent confusion as to it’s scope, intent and implementation.
I suggest we discuss at devcall and/or someone drafts an RFC.

hamishwillee:
@auturgy I think the confusion about scope and intent and implementation no longer exists - the message proposal captures what is desired.

The message is very obviously an improvement over DO_SET_SERVO. The only question then is whether there is any value/synergy in trying to find overlap with RC_OVERRIDE. To which I say "you're to experienced for me to dismiss the idea, but my gut feeling is that there is a lot to be said for keeping the evolution of these two things separate".

I'll add to the devcall. @bkueng Can you attend - I think this is early morning (8am for you) next Weds?

coderkalyan:
Ok then I will get started implementing this with DO_SET_SERVO for now. I will also be on the PX4 dev call this week if you want to discuss it @bkueng @hamishwillee.

hamishwillee:
Thanks @coderkalyan , catch you tomorrow then!

bkueng:
Just to avoid confusion: we are talking about the [MAVLink dev call](https://mavlink.io/en/about/support.html#dev_call), not PX4, right?

coderkalyan:
I can be on both :) What time is the MAVLink call? (PDT here)

hamishwillee:
Yes, MAVLInk dev call. In about 6 hours. 
@coderkalyan It is marked on the calendar https://www.dronecode.org/calendar/ - that allows you to pick the timezone you need.

coderkalyan:
Ouch, looks like its at midnight my time - probably can make it. I'll try to be on the PX4 call, though.

coderkalyan:
@hamishwillee let me know if it looks good now. I removed the AUX terminology and switched param 7 to an index.

About the CYCLE_ACTUATOR, I added it to be a replacement for REPEAT_SERVO but added a little more functionality since REPEAT_SERVO seemed very imprecisely defined. If this message isn't wanted altogether, I can remove it.

hamishwillee:
@coderkalyan Thanks very much! I'm very happy with MAV_CMD_DO_SET_ACTUATOR.

Can you please remove ACTUATOR_CYCLE_INTERPOLATION and MAV_CMD_DO_CYCLE_ACTUATOR and then I will merge this. 

I am **not** saying that I am not happy with MAV_CMD_DO_CYCLE_ACTUATOR. But I think with all the discussion of the first command this has not been fully considered. If you think this is useful then please add that as a separate PR with a couple of use cases you see.

coderkalyan:
@hamishwillee Okay. I've backed up the cycle command onto an offline branch and removed it from this PR. You should be free to merge now.

hamishwillee:
@coderkalyan Sorry, one last thing. Can you also add the two lines used for wip markup ? That allows us to change this until we have an implementation.

coderkalyan:
@hamishwillee Should be good now?

hamishwillee:
@coderkalyan This is great, but github is not letting me update the branch (and hence I can't merge). Does it give you the option to update?

image

@julianoes This has happened to me a few times today, but never before. How about you? Ideas?

coderkalyan:
@hamishwillee Just updated :)

hamishwillee:
Awesome. Merged! Usually this makes its way into PX4 within a week, though it might be faster, and you could of course PR your own update.

Thanks for taking the lead on this!

coderkalyan:
No problem, glad to help! I'll definitely put the update in my PR for do_set_servo support, where I can also support this command.

hamishwillee:
Very cool. Presumably we'll just support 3 or 4 of those servo slots :-).

PR #1421, opened 2020-07-07
Define cellular config messages for configuring Modems

Jaeyoung-Lim:
Currently there is no mavlink message to enable configuring cellular modems.
This message is required to configure the cellular model information.

The information includes
- APN
- PIN numbers
- Puk codes
- enabling /disabling pins
- enabling /disabling roaming
- black listing bands

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6639,12 +6639,14 @@
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
     <message id="336" name="CELLULAR_CONFIG">
-      <description>Configure Cellular Modems. This message is re-emitted as an acknowledgement by the modem. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE</description>
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Configure cellular modems. This message is re-emitted as an acknowledgement by the modem. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint8_t" name="enable_pin">Enable / disable PIN on the SIM card. 0: Unchange setttings 1: PIN disabled, 2: PIN enabled.</field>
       <field type="char[32]" name="pin">PIN sent to the simcard. Blank when PIN is disabled. Empty when message is sent back as a response.</field>
       <field type="char[32]" name="apn">Name of the cellular APN. Blank to leave it unchanged when setting. Current APN when sent back as a response.</field>
-      <field type="char[32]" name="puk">Required Puk code in case the user failed to authenticate 3 times with the PIN.</field>
-      <field type="uint8_t" name="roaming">Configure whether roaming is allowed, 0: Unchange setttings 1: roaming disabled, 2: roaming enabled.</field>
+      <field type="char[32]" name="puk">Required PUK code in case the user failed to authenticate 3 times with the PIN.</field>
+      <field type="uint8_t" name="roaming">Configure whether roaming is allowed, 0: settings not changed, 1: roaming disabled, 2: roaming enabled.</field>
       <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="339" name="RAW_RPM">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1281,21 +1281,18 @@
       </entry>
     </enum>
     <enum name="CELLULAR_CONFIG_RESPONSE">
-      <description>Possible responses from a WIFI_CONFIG_AP message.</description>
-      <entry value="0" name="CELLULAR_CONFIG_RESPONSE_UNDEFINED">
-        <description>Undefined response. Likely an indicative of a system that doesn't support this request.</description>
-      </entry>
-      <entry value="1" name="CELLULAR_CONFIG_RESPONSE_ACCEPTED">
+      <description>Possible responses from a CELLULAR_CONFIG message.</description>
+      <entry value="0" name="CELLULAR_CONFIG_RESPONSE_ACCEPTED">
         <description>Changes accepted.</description>
       </entry>
-      <entry value="2" name="CELLULAR_CONFIG_RESPONSE_REJECTED">
-        <description>Changes rejected.</description>
-      </entry>
-      <entry value="3" name="CELLULAR_CONFIG_RESPONSE_APN_ERROR">
+      <entry value="1" name="CELLULAR_CONFIG_RESPONSE_APN_ERROR">
         <description>Invalid APN.</description>
       </entry>
-      <entry value="4" name="CELLULAR_CONFIG_RESPONSE_PIN_ERROR">
-        <description>Invalid Password.</description>
+      <entry value="2" name="CELLULAR_CONFIG_RESPONSE_PIN_ERROR">
+        <description>Invalid PIN.</description>
+      </entry>
+      <entry value="3" name="CELLULAR_CONFIG_RESPONSE_REJECTED">
+        <description>Changes rejected.</description>
       </entry>
     </enum>
     <enum name="WIFI_CONFIG_AP_MODE">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1280,6 +1280,24 @@
         <description>Invalid Password.</description>
       </entry>
     </enum>
+    <enum name="CELLULAR_CONFIG_RESPONSE">
+      <description>Possible responses from a WIFI_CONFIG_AP message.</description>
```
julianoes, line 1284:
```suggestion
      <description>Possible responses from a CELLULAR_CONFIG message.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -1280,6 +1280,24 @@
...
+      <entry value="0" name="CELLULAR_CONFIG_RESPONSE_UNDEFINED">
+        <description>Undefined response. Likely an indicative of a system that doesn't support this request.</description>
+      </entry>
```
julianoes, line 1287: I probably wouldn't spec "undefined" as it's very unclear. If you want to indicate not supported, then just add `UNSUPPORTED`.

message_definitions/v1.0/common.xml
```diff
@@ -1280,6 +1280,24 @@
...
+      <entry value="1" name="CELLULAR_CONFIG_RESPONSE_ACCEPTED">
```
julianoes, line 1288: Also, I would probalby have 0 as `ACCEPTED` like `COMMAND_ACK`s.

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
       <field type="uint8_t" name="tx_session_pending">1: Transmission session pending, 0: No transmission session pending.</field>
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
+    <message id="336" name="CELLULAR_CONFIG">
+      <description>Configure Cellular Modems. This message is re-emitted as an acknowledgement by the modem. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE</description>
+      <field type="uint8_t" name="enable_pin">Enable / disable PIN on the SIM card.</field>
+      <field type="char[32]" name="pin">PIN sent to the simcard. Black when PIN is disabled. MD5 hash when message is sent back as a response.</field>
```
julianoes, line 6647: Black?

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
...
+      <extensions/>
```
julianoes, line 6648: No need for extensions as this is a new message.

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
...
+      <field type="char[32]" name="apn">Name of the cellular APN. Blank to leave it unchanged when setting. Current SSID when sent back as a response.</field>
```
julianoes, line 6649: APN or SSID?

message_definitions/v1.0/common.xml
julianoes: MD5? MD5 is not secure, shouldn't this be something like SHA2?

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
...
+      <field type="char[32]" name="puk">Required Puk code in case the user failed to authenticate 3 times with the PIN.</field>
+      <field type="uint8_t" name="Roaming">Configure whether roaming is allowed.</field>
```
julianoes, line 6651:
```suggestion
      <field type="uint8_t" name="Roaming">Configure whether roaming is allowed, 0: roaming disabled, 1: roaming enabled.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
...
+      <field type="uint16_t" name="band">Black list bands.</field>
```
julianoes, line 6652: What about "block list" or "deny list"?

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
...
+      <field type="int8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
```
julianoes, line 6653:
```suggestion
      <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -1280,6 +1280,24 @@
...
+        <description>Changes accepted.</description>
+      </entry>
+      <entry value="2" name="CELLULAR_CONFIG_RESPONSE_REJECTED">
+        <description>Changes rejected.</description>
+      </entry>
+      <entry value="3" name="CELLULAR_CONFIG_RESPONSE_APN_ERROR">
+        <description>Invalid APN.</description>
+      </entry>
+      <entry value="4" name="CELLULAR_CONFIG_RESPONSE_PIN_ERROR">
+        <description>Invalid Password.</description>
```
julianoes, line 1298:
```suggestion
        <description>Invalid PIN.</description>
```

message_definitions/v1.0/common.xml
Jaeyoung-Lim: Ah, sorry I wanted to write "Black". The md5 is sent back also in the `WIFI_CONFIG_AP` message, so I wanted to be consistent

DanielePettenuzzo: "Blank"?

message_definitions/v1.0/common.xml
DanielePettenuzzo: APN is correct. I assume LTE modems don't have an SSID

message_definitions/v1.0/common.xml
JonasVautherin: I think this comes from [WIFI_CONFIG_AP_RESPONSE](https://github.com/mavlink/mavlink/blob/1a909cf855e8c8c27253d9d174e00a8a3702737b/message_definitions/v1.0/common.xml#L1264). I do agree that `UNSUPPORTED` would be clearer, but I wanted to mention it for the sake of consistency.

message_definitions/v1.0/common.xml
JonasVautherin, line 6647:
```suggestion
      <field type="char[32]" name="pin">PIN sent to the simcard. Blank when PIN is disabled. MD5 hash when message is sent back as a response.</field>
```

JonasVautherin:
```suggestion
      <field type="char[32]" name="pin">PIN sent to the simcard. Blank when PIN is disabled. MD5 hash when message is sent back as a response.</field>
```

message_definitions/v1.0/common.xml
julianoes: I wouldn't copy another message if we can do it better for the sake of consistency but it's up to you.

message_definitions/v1.0/common.xml
julianoes: Ah :smile:

message_definitions/v1.0/common.xml
hamishwillee: FWIW agree - if it really means unsupported then it should say that.

message_definitions/v1.0/common.xml
```diff
@@ -1280,6 +1280,24 @@
...
     </enum>
+    <enum name="CELLULAR_CONFIG_RESPONSE">
+      <description>Possible responses from a CELLULAR_CONFIG message.</description>
+      <entry value="0" name="CELLULAR_CONFIG_RESPONSE_UNDEFINED">
+        <description>Undefined response. Likely an indicative of a system that doesn't support this request.</description>
+      </entry>
+      <entry value="1" name="CELLULAR_CONFIG_RESPONSE_ACCEPTED">
+        <description>Changes accepted.</description>
+      </entry>
+      <entry value="2" name="CELLULAR_CONFIG_RESPONSE_REJECTED">
```
hamishwillee, line 1291: Are all the remainder errors, and this is the "generic error you get if you don't know why it failed"? If so, perhaps this should be last.

message_definitions/v1.0/common.xml
```diff
@@ -6623,6 +6641,17 @@
       <field type="uint8_t" name="tx_session_pending">1: Transmission session pending, 0: No transmission session pending.</field>
       <field type="uint8_t" name="rx_session_pending">1: Receiving session pending, 0: No receiving session pending.</field>
     </message>
+    <message id="336" name="CELLULAR_CONFIG">
+      <description>Configure Cellular Modems. This message is re-emitted as an acknowledgement by the modem. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE</description>
+      <field type="uint8_t" name="enable_pin">Enable / disable PIN on the SIM card.</field>
```
hamishwillee, line 6646: This is a pin number? Maximum size of 256 reasonable?

message_definitions/v1.0/common.xml
hamishwillee: What does this even mean - real example please. Specifically, this is a 16 bit number - what number would you put in here and why?

hamishwillee:
@auturgy We hope to discuss this message this evening. Also https://github.com/mavlink/mavlink/pull/1323

message_definitions/v1.0/common.xml
Jaeyoung-Lim: Some regions of the world (e.g. US) require legally to black list certain LTE bands. For example you are not allowed to operate commercially in band5 in the US. I defined it as `uint16_t` since the CELLULAR_BAND enum has the same type in https://github.com/mavlink/mavlink/pull/1323, but wasn't sure if the cellular status message will get approved

message_definitions/v1.0/common.xml
Jaeyoung-Lim: This is just a flag to enable / disable pin authentication on the SIM!

hamishwillee: We're normally very precise - this isn't C++ or some other programming language where you can assume that a particular value is true or false. So something like (to your taste):

> Enable setting the PIN on the SIM card. 0: Pin setting disabled, 1: Pin setting enabled.

message_definitions/v1.0/common.xml
hamishwillee: Yes, but the confusion is that here you have just one "slot", so you can only blacklist one band right? Intentional?

hamishwillee:
@Jaeyoung-Lim Thanks for making the changes. I did a minor update to mark this a wip. That allows us to change it in future. Once you have an implementation in the wild and we are happy with this message, let me know and we can remove it.

Jaeyoung-Lim:
Thanks @hamishwillee !

PR #1423, opened 2020-07-08
BATTERY_STATUS.voltage_ext: add cells 10 to 14

hamishwillee:
This replaces https://github.com/mavlink/mavlink/pull/1412, which for some reason Github will not allow me to update or merge.

@muramura Thanks for this. I have had to recreate the PR and commit to get the change to allow me to edit it. Can you please check that this is still correct. The main change is to clarify that since 0 now means "not supported" they have to send 1 to mean "0mV".

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5888,7 +5888,7 @@
       <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
       <field type="uint8_t" name="type" enum="MAV_BATTERY_TYPE">Type (chemistry) of the battery</field>
       <field type="int16_t" name="temperature" units="cdegC">Temperature of the battery. INT16_MAX for unknown temperature.</field>
-      <field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells. Cells above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).</field>
+      <field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells 1 to 10 (see voltages_ext for cells 11-14). Cells in this field above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).</field>
       <field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes @auturgy This look OK to you? As per dev call, it just adds 4 cells as an extension field, 0 means not supported, and 1 means "approximately 0".

auturgy:
Makes sense to me.
We’ll probably have to scale it again in the future but I think this provides an acceptable pattern for doing so.

Regards,

James

hamishwillee:
@auturgy Hmmm. What do you see as reasonable future proofing in terms of "number of cells"?

It might make sense to assign more now. Argument against is that if we add any other field at the end that can't be truncated we'll be creating a less efficient message than we have now.

rmackay9:
I think 12S is relatively common.  I think I've only heard of a user using 14S once.

muramura:
@hamishwillee san.  Thank you for the review & re PR. I have checked the changes. My thoughts included. Thank you.

auturgy:
14S is relatively common for trad heli, and there’s at least one user already asking about 24S (which is insane). But highlights that if we’re going to touch messages that are so fundamental, we should do it in a way that allows for growth and minimises the pain for when we have to do it again. “Good enough for most people now” isn’t good enough.

rmackay9:
at the same time, adding extra fields comes with a bandwidth cost for everyone I think.

hamishwillee:
@rmackay9 Due to empty byte truncation the cost of this addition is only paid if a system actually uses the extra cells. It is essentially free _now_ for us to add as many cells as we like.

My concern though is that if later someone says I want to add a "thingy" field at the end , truncation no longer works, so `voltage_ext` immediately gets paid by everyone.

If we think this message is "done forever" we must make the voltage_ext really big. If we think it is likely to extend, we make it smaller, and accept that we may have to invent `voltage_ext2` in future.

I lean towards leaving it as it is now (14s) and inventing `voltage_ext2` later if needed. At that point we might decide we're better off having a new message.

rmackay9:
@hamishwillee, ok, sounds good.  I actually agree with you that 14S is enough for now.  Nobody knows the future but.. seems OK to me.

JackWakleyMA:
Thanks to James and Randy for pointing me to this. For our aircraft, 12S is the minimum we fly with, however I'm not that interested in presenting individual cells to the end user. The thing I need to know about a battery before a customer flies it is the max cell deviation. That is the information I would want a pre-arm checks against; along with pack voltage. In the future we may also build current monitoring into the battery's BMS so that would be a third piece of information to send to the flight controller from the same source.

rmackay9:
Looks like this needs a rebase but I vote for this going in.

hamishwillee:
@JackWakleyMA So what fields would you actually want/can you infer them from the current message? Propose them here in XML form please. 
Because depending on the size and how they are defined it might be good to insert these before the `voltage_ext`.

rmackay9:
@hamishwillee, @JackWakleyMA,

We can handle the pre-arm check portion of Jack's request without adding to the xml message I think. It would only be if Jack wanted a continuous real-time view of the maximum cell voltage deviation that we would need an additional field.

hamishwillee:
@rmackay9 Fair enough - if we needed that then I'd propose we think about a redesign of the messages into high and low rate version. 

Anyway, I'll give Jack a chance to comment. I'm not working today, so as long as there are no objections, I will merge Monday. [ping again if this is urgent your side]

auturgy:
Whilst I don’t think it’s implemented anywhere, we did create the smart_battery messages a while back, which include cell voltages etc.
I think rolling this message in as-is, and look to implementing smart_battery for Jack, might be a sensible path.

Regards,

James

hamishwillee:
Great. Merged!

rmackay9:
@amilcarlucas, no pressure of course but I'll make use of this change very soon if we merge it to the ArduPilot fork.

muramura:
@hamishwillee san. Thank you.

hamishwillee:
@muramura You are very welcome! Thanks for your patience on the back-and-forth.

PR #1425, opened 2020-07-15
Update cellular config and message

mattstraehl:
This is a follow-up PR on PR: https://github.com/mavlink/mavlink/pull/1421

It contains the following changes:
* Add `CELLULAR_CONFIG_BLOCKED_PUK_REQUIRED` response code to `CELLULAR_CONFIG_RESPONSE`. This response code is needed to indicate that the PUK code is required to unblock SIM card.
* Add field `enable_lte` to CELLULAR_CONFIG message. This field is needed to enable/disable LTE on the vehicle.
* Add field `new_pin` to `CELLULAR_CONFIG` message. This field is needed when changing the PIN.
* Reduce length of the message fields for PIN and PUK to 16.
* Some minor message field description changes.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1344,6 +1344,9 @@
       <entry value="3" name="CELLULAR_CONFIG_RESPONSE_REJECTED">
         <description>Changes rejected.</description>
       </entry>
+      <entry value="4" name="CELLULAR_CONFIG_BLOCKED_PUK_REQUIRED">
+        <description>PUK is required to unblock SIM card.</description>
+      </entry>
     </enum>
     <enum name="WIFI_CONFIG_AP_MODE">
       <description>WiFi Mode.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6783,6 +6783,7 @@
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Configure cellular modems. This message is re-emitted as an acknowledgement by the modem. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE.</description>
+      <field type="uint8_t" name="enable_lte">Enable/disable LTE. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
       <field type="uint8_t" name="enable_pin">Enable/disable PIN on the SIM card. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
       <field type="char[16]" name="pin">PIN sent to the SIM card. Blank when PIN is disabled. Empty when message is sent back as a response.</field>
       <field type="char[16]" name="new_pin">New PIN when changing the PIN. Blank to leave it unchanged. Empty when message is sent back as a response.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6780,11 +6783,12 @@
...
-      <field type="char[32]" name="pin">PIN sent to the simcard. Blank when PIN is disabled. Empty when message is sent back as a response.</field>
-      <field type="char[32]" name="apn">Name of the cellular APN. Blank to leave it unchanged when setting. Current APN when sent back as a response.</field>
-      <field type="char[32]" name="puk">Required PUK code in case the user failed to authenticate 3 times with the PIN.</field>
-      <field type="uint8_t" name="roaming">Configure whether roaming is allowed, 0: settings not changed, 1: roaming disabled, 2: roaming enabled.</field>
+      <field type="uint8_t" name="enable_pin">Enable/disable PIN on the SIM card. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
+      <field type="char[32]" name="pin">PIN sent to the SIM card. Blank when PIN is disabled. Empty when message is sent back as a response.</field>
+      <field type="char[32]" name="new_pin">New PIN when changing the PIN. Blank to leave it unchanged. Empty when message is sent back as a response.</field>
```
hamishwillee, line 6788: @mattstraehl So how does this work? My guess is that to set a new pin you send both the old pin and new pin values? And you'd get CELLULAR_CONFIG_RESPONSE_PIN_ERROR if the pin was wrong?

message_definitions/v1.0/common.xml
```diff
@@ -6780,11 +6783,12 @@
...
-      <field type="uint8_t" name="enable_pin">Enable / disable PIN on the SIM card. 0: Unchange setttings 1: PIN disabled, 2: PIN enabled.</field>
-      <field type="char[32]" name="pin">PIN sent to the simcard. Blank when PIN is disabled. Empty when message is sent back as a response.</field>
-      <field type="char[32]" name="apn">Name of the cellular APN. Blank to leave it unchanged when setting. Current APN when sent back as a response.</field>
-      <field type="char[32]" name="puk">Required PUK code in case the user failed to authenticate 3 times with the PIN.</field>
-      <field type="uint8_t" name="roaming">Configure whether roaming is allowed, 0: settings not changed, 1: roaming disabled, 2: roaming enabled.</field>
+      <field type="uint8_t" name="enable_pin">Enable/disable PIN on the SIM card. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
+      <field type="char[32]" name="pin">PIN sent to the SIM card. Blank when PIN is disabled. Empty when message is sent back as a response.</field>
```
julianoes, line 6787: All the pins and puks I've seen were always 4 or 6 letters long? Are they actually up to 32 letters long?

message_definitions/v1.0/common.xml
mattstraehl: Yes exactly, that's also how I was thinking it would work.

message_definitions/v1.0/common.xml
mattstraehl: It looks like the PUK code can be up to 8 digits long: https://en.wikipedia.org/wiki/Personal_unblocking_key

So I think we would need at least 9 characters (8 plus 1 for the trailing zero). So how about we use `char[16]` for `pin` and `puk` which will give some extra space to be safe?

hamishwillee:
Looks good. Merged. Is WIP, so we can iterate further if needed later.

PR #1426, opened 2020-07-16
common: add note about lan and lon = 0

julianoes:
This just describes the state as it is implemented by ArduPilot and QGC.

Relevant discussion in https://github.com/PX4/Firmware/pull/15308.

MAVSDK change to match it: https://github.com/PX4/Firmware/pull/15308

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4912,8 +4912,8 @@
       <description>The filtered global position (e.g. fused GPS and accelerometers). The position is in GPS-frame (right-handed, Z-up). It
                is designed as scaled integer message since the resolution of float is not sufficient.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="int32_t" name="lat" units="degE7">Latitude, expressed</field>
-      <field type="int32_t" name="lon" units="degE7">Longitude, expressed</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude, expressed (if both lat and lon are 0, it means that they are not available yet)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude, expressed (if both lat and lon are 0, it means that they are not available yet)</field>
       <field type="int32_t" name="alt" units="mm">Altitude (MSL). Note that virtually all GPS modules provide both WGS84 and MSL.</field>
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
       <field type="int16_t" name="vx" units="cm/s">Ground X Speed (Latitude, positive north)</field>


message_definitions/v1.0/common.xml
```diff
@@ -4918,8 +4918,8 @@
       <description>The filtered global position (e.g. fused GPS and accelerometers). The position is in GPS-frame (right-handed, Z-up). It
                is designed as scaled integer message since the resolution of float is not sufficient.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="int32_t" name="lat" units="degE7">Latitude, expressed</field>
-      <field type="int32_t" name="lon" units="degE7">Longitude, expressed</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude, expressed (if both lat and lon are 0, it means that they are not available yet)</field>
```
hamishwillee, line 4921: is ", expressed" needed here? I suspect a hangover from when the units were removed.

auturgy:
Let’s not scope creep: I’d get this in (the edge case, aka Null Island, isn’t likely to be encountered), and perhaps raise an issue to review/add extras

LorenzMeier:
Entering magic values into a valid range is against best practices. It is unlikely that a lat / lon of 0/0 ever hits a real system - yes. But if I were to review the MAVLink spec and found stuff like that I would loose trust into the implementers because very apparently they are too lazy to implement things correctly and have all sorts of magic easter eggs hidden. 

Magic values are fine until they are not. And it is very simple to objectify this: If a unit test that does an input data sweep would fall over, then it is a bad change. In this case a sweep across all possible coordinates would generate an incorrect result here, so it is a bad change.

@DonLakeFlyer If you need altitude, please use the correct packet for it: http://mavlink.io/en/messages/common.html#ALTITUDE

hamishwillee:
@LorenzMeier Is it worth though tightening the spec to do this "correctly"? i.e. specify reasonable values like `INT32_MAX ` for the  values that potentially might not be sent, and which are therefore sent as zero.

DonLakeFlyer:
> specify reasonable values like `INT32_MAX ` for the values that potentially might not be sent

We should do that. There needs to be something, otherwise a GCS doesn't know how to get altitude information from what message.

LorenzMeier:
The INT32 max approach is perfectly adequate and fine by me. It might however break existing receivers which lack range checking.

@DonLakeFlyer You really should be getting altitude information from the ALTITUDE message. GLOBAL_POSITION_INT is designed as a global position estimate that ONLY gets sent if you actually have one. The MAVLink spec has an approach to do all of this correctly for a long time, so it should be used.

DonLakeFlyer:
> The MAVLink spec has an approach to do all of this correctly for a long time, so it should be used.

Doc'ed where?

LorenzMeier:
https://github.com/mavlink/MAVSDK/pull/1151/files

> @DonLakeFlyer If you need altitude, please use the correct packet for it: http://mavlink.io/en/messages/common.html#ALTITUDE

DonLakeFlyer:
Ages ago, this was changed: https://github.com/mavlink/qgroundcontrol/pull/4346. It made GLOBAL_POSITION_INT take precedence over ALTITUDE. No comments in there as to why unfortunately.

So the question for mavlink spec is which one of these statements is correct:
* Send of ALTITUDE is required and should be the only place GCS gets altitude telemetry from
* Send of ALTITUDE is optional and if not sent a GCS should fall back to GLOBAL_POSITION_INT for altitudes

PR #1427, opened 2020-07-17
Add a couple of fields and status flags to GENERATOR_STATUS

peterbarker:


rmackay9:
This looks good to me.  I think the key question is whether the WIP field means that we can add these fields (without using mavlink2 extensions) which will break applications that are using the old xml.  If the answer is "yes" then I think this PR can be merged from my point of view anyway.  If the answer is "no" then the WIP label should be removed.

hamishwillee:
@rmackay9 WIP means that the message can be modified at will without any great concern for potential breaks in users - essentially it means we do not need to add extensions.

We remove the WIP when there is a product-ready implementation and you guys are fairly sure it is stable. Because once removed you will be able to add extensions and new enums, but the name, field type, and size of the existing fields will be fixed.

This looks great to me. But I'd like @CarlEngel to comment first. If you really need it merged soonest I can do that, but be prepared for later removal if he has a problem with it.

peterbarker:
@hamishwillee Carl as OK'd this elsewhere.

PR #1428, opened 2020-07-21
extend param_ext protocol by adding messages to start and commit a parameter transaction

DanielePettenuzzo:
Lately we have created several mavlink messages to set some settings on the vehicle (eg. the wifi credentials message and the cellular config message. 

In mavlink we already have a parameter protocol that could be used although it currently has some limitations. The three main limitations are the following:
- We cannot set multiple parameters at the same time. Every parameter is set in a separate mavlink message and is applied as soon as it's received. For settings such as wifi SSID and password we would like to first set both parameters before applying them for obvious reasons.
- Currently we would need to use the PARAM_EXT to set such strings. The PARAM_EXT has a value field that is 128 bytes long and even if not all these bytes are used there is no trimming therefore we are always transmitting them for each parameter.
- There is no way to attach some description/metadata to a given parameter.

With this pr I try to solve the first issue. To do this I created two new messages to start and stop (commit) a parameter transaction. With the start command we tell the vehicle that we will set some parameters but they should not be applied right away. With the commit message we tell the vehicle that we are done setting parameters and they can now be applied. 

Both messages are sent out by the ground station and the vehicle will respond with the same message but adding its transaction id (server) and a response to the message. This response works as an ack for the messages.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3923,7 +3923,7 @@
         <description>Parameter failed to set</description>
       </entry>
       <entry value="3" name="PARAM_ACK_IN_PROGRESS">
-        <description>Parameter value received but not yet set/accepted. A subsequent PARAM_TRANSACT_ACK or PARAM_EXT_ACK with the final result will follow once operation is completed. This is returned immediately for parameters that take longer to set, indicating taht the the parameter was recieved and does not need to be resent.</description>
+        <description>Parameter value received but not yet set/accepted. A subsequent PARAM_ACK_TRANSACTION or PARAM_EXT_ACK with the final result will follow once operation is completed. This is returned immediately for parameters that take longer to set, indicating taht the the parameter was recieved and does not need to be resent.</description>
       </entry>
     </enum>
     <enum name="CAMERA_MODE">
@@ -4801,7 +4801,7 @@
       <field type="uint32_t" name="custom_mode">The new autopilot-specific mode. This field can be ignored by an autopilot.</field>
     </message>
     <!-- IDs 15-17 reserved for PARAM_VALUE_UNION and other param messages -->
-    <message id="19" name="PARAM_TRANSACT_ACK">
+    <message id="19" name="PARAM_ACK_TRANSACTION">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Response from a PARAM_SET message when it is used in a transaction.</description>
@@ -4833,7 +4833,7 @@
       <field type="uint16_t" name="param_index">Index of this onboard parameter</field>
     </message>
     <message id="23" name="PARAM_SET">
-      <description>Set a parameter value (write new value to permanent storage). Within a transaction the recieving componenent should respond with PARAM_TRANSACT_ACK to the setter component. IMPORTANT: If sent outside a transaction the receiving component should acknowledge the new parameter value by broadcasting a PARAM_VALUE message to all communication partners (broadcasting ensures that multiple GCS all have an up-to-date list of all parameters). If the sending GCS did not receive a PARAM_VALUE or PARAM_TRANSACT_ACK message within its timeout time, it should re-send the PARAM_SET message. The parameter microservice is documented at https://mavlink.io/en/services/parameter.html</description>
+      <description>Set a parameter value (write new value to permanent storage). Within a transaction the recieving componenent should respond with PARAM_ACK_TRANSACTION to the setter component. IMPORTANT: If sent outside a transaction the receiving component should acknowledge the new parameter value by broadcasting a PARAM_VALUE message to all communication partners (broadcasting ensures that multiple GCS all have an up-to-date list of all parameters). If the sending GCS did not receive a PARAM_VALUE or PARAM_ACK_TRANSACTION message within its timeout time, it should re-send the PARAM_SET message. The parameter microservice is documented at https://mavlink.io/en/services/parameter.html</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="char[16]" name="param_id">Onboard parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1372,19 +1372,31 @@
         <description>Parameter meta data.</description>
       </entry>
     </enum>
-    <enum name="PARAM_EXT_TRANSACTION_RESPONSE">
-      <description>Possible responses from a PARAM_EXT_START_TRANSACTION and PARAM_EXT_COMMIT_TRANSACTION messages.</description>
-      <entry value="0" name="PARAM_EXT_TRANSACTION_RESPONSE_ACCEPTED">
+    <enum name="PARAM_TRANSACTION_RESPONSE">
+      <description>Possible responses from a PARAM_START_TRANSACTION and PARAM_COMMIT_TRANSACTION messages.</description>
+      <entry value="0" name="PARAM_TRANSACTION_RESPONSE_ACCEPTED">
         <description>Transaction accepted.</description>
       </entry>
-      <entry value="1" name="PARAM_EXT_TRANSACTION_RESPONSE_FAILED">
+      <entry value="1" name="PARAM_TRANSACTION_RESPONSE_FAILED">
         <description>Transaction failed.</description>
       </entry>
-      <entry value="2" name="PARAM_EXT_TRANSACTION_RESPONSE_UNSUPPORTED">
+      <entry value="2" name="PARAM_TRANSACTION_RESPONSE_UNSUPPORTED">
         <description>Transaction unsupported.</description>
       </entry>
-      <entry value="3" name="PARAM_EXT_TRANSACTION_RESPONSE_INPROGRESS">
+      <entry value="3" name="PARAM_TRANSACTION_RESPONSE_INPROGRESS">
         <description>Transaction in progress.</description>
+      </entry>
+    </enum>
+    <enum name="PARAM_TRANSACTION_TRANSPORT">
+      <description>Possible transport layers to set and get parameters via mavlink during a parameter transaction.</description>
+      <entry value="0" name="PARAM_TRANSACTION_TRANSPORT_PARAM">
+        <description>Transaction over param transport.</description>
+      </entry>
+      <entry value="1" name="PARAM_TRANSACTION_TRANSPORT_PARAM_EXT">
+        <description>Transaction over param_ext transport.</description>
+      </entry>
+      <entry value="2" name="PARAM_TRANSACTION_TRANSPORT_MAVFTP">
+        <description>Transaction over mavftp transport.</description>
       </entry>
     </enum>
     <!-- The MAV_CMD enum entries describe either: -->
@@ -6717,21 +6729,18 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
     </message>
-    <message id="325" name="PARAM_EXT_START_TRANSACTION">
-      <description>Request to start a new parameter transaction. The response (ack) will contain the same message but with a response attached to it.</description>
+    <message id="325" name="PARAM_START_TRANSACTION">
+      <description>Request to start a new parameter transaction. Multiple kinds of transport layers can be used to exchange parameters in the transaction (param, param_ext and mavftp). The response (ack) will contain the same message but with a response attached to it.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint32_t" name="client_transaction_id">This field contains the transaction id for the sending side. </field>
-      <field type="uint32_t" name="server_transaction_id">This field contains the transaction id for the receiving side. It will be set to zero in the request and will be filled in only in the response. </field>
-      <field type="uint8_t" name="response" enum="PARAM_EXT_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
+      <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Message acceptance response (sent back to GS).</field>
+      <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
-    <message id="326" name="PARAM_EXT_COMMIT_TRANSACTION">
-      <description>Request to end the current parameter transaction. This message will have effect only if a transaction was previously opened using the PARAM_EXT_START_TRANSACTION message. The response will contain the same message but with a response attached to it. The response can either be a success/failure or and in progress in case the receiving side takes some time to apply the parameters.</description>
+    <message id="326" name="PARAM_COMMIT_TRANSACTION">
+      <description>Request to end the current parameter transaction. This message will have effect only if a transaction was previously opened using the PARAM_START_TRANSACTION message. The response will contain the same message but with a response attached to it. The response can either be a success/failure or and in progress in case the receiving side takes some time to apply the parameters.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint32_t" name="client_transaction_id">This field contains the transaction id for the sending side.  </field>
-      <field type="uint32_t" name="server_transaction_id">This field contains the transaction id for the receiving side. It will be set to zero in the request and will be filled in only in the response. </field>
-      <field type="uint8_t" name="response" enum="PARAM_EXT_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
+      <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="330" name="OBSTACLE_DISTANCE">
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6729,14 +6729,14 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
     </message>
-    <message id="325" name="PARAM_START_TRANSACTION">
+    <message id="328" name="PARAM_START_TRANSACTION">
       <description>Request to start a new parameter transaction. Multiple kinds of transport layers can be used to exchange parameters in the transaction (param, param_ext and mavftp). The response (ack) will contain the same message but with a response attached to it.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Message acceptance response (sent back to GS).</field>
       <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
-    <message id="326" name="PARAM_COMMIT_TRANSACTION">
+    <message id="329" name="PARAM_COMMIT_TRANSACTION">
       <description>Request to end the current parameter transaction. This message will have effect only if a transaction was previously opened using the PARAM_START_TRANSACTION message. The response will contain the same message but with a response attached to it. The response can either be a success/failure or and in progress in case the receiving side takes some time to apply the parameters.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1372,6 +1372,21 @@
         <description>Parameter meta data.</description>
       </entry>
     </enum>
+    <enum name="PARAM_EXT_TRANSACTION_RESPONSE">
+      <description>Possible responses from a PARAM_EXT_START_TRANSACTION and PARAM_EXT_COMMIT_TRANSACTION messages.</description>
+      <entry value="0" name="PARAM_EXT_TRANSACTION_RESPONSE_ACCEPTED">
+        <description>Transaction accepted.</description>
+      </entry>
+      <entry value="1" name="PARAM_EXT_TRANSACTION_RESPONSE_FAILED">
+        <description>Transaction failed.</description>
+      </entry>
+      <entry value="2" name="PARAM_EXT_TRANSACTION_RESPONSE_UNSUPPORTED">
+        <description>Transaction unsupported.</description>
+      </entry>
+      <entry value="3" name="PARAM_EXT_TRANSACTION_RESPONSE_INPROGRESS">
+        <description>Transaction in progress.</description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
@@ -6701,6 +6716,22 @@
       <field type="char[128]" name="param_value">Parameter value (new value if PARAM_ACK_ACCEPTED, current value otherwise)</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
+    </message>
+    <message id="325" name="PARAM_EXT_START_TRANSACTION">
+      <description>Request to start a new parameter transaction. The response (ack) will contain the same message but with a response attached to it.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="client_transaction_id">This field contains the transaction id for the sending side. </field>
+      <field type="uint32_t" name="server_transaction_id">This field contains the transaction id for the receiving side. It will be set to zero in the request and will be filled in only in the response. </field>
+      <field type="uint8_t" name="response" enum="PARAM_EXT_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
+    </message>
+    <message id="326" name="PARAM_EXT_COMMIT_TRANSACTION">
+      <description>Request to end the current parameter transaction. This message will have effect only if a transaction was previously opened using the PARAM_EXT_START_TRANSACTION message. The response will contain the same message but with a response attached to it. The response can either be a success/failure or and in progress in case the receiving side takes some time to apply the parameters.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="client_transaction_id">This field contains the transaction id for the sending side.  </field>
+      <field type="uint32_t" name="server_transaction_id">This field contains the transaction id for the receiving side. It will be set to zero in the request and will be filled in only in the response. </field>
+      <field type="uint8_t" name="response" enum="PARAM_EXT_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="330" name="OBSTACLE_DISTANCE">
       <description>Obstacle distances in front of the sensor, starting from the left in increment degrees to the right</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3912,7 +3912,7 @@
       </entry>
     </enum>
     <enum name="PARAM_ACK">
-      <description>Result from a PARAM_SET or PARAM_EXT_SET message.</description>
+      <description>Result from PARAM_EXT_SET message (or a PARAM_SET within a transaction).</description>
       <entry value="0" name="PARAM_ACK_ACCEPTED">
         <description>Parameter value ACCEPTED and SET</description>
       </entry>
@@ -3923,7 +3923,7 @@
         <description>Parameter failed to set</description>
       </entry>
       <entry value="3" name="PARAM_ACK_IN_PROGRESS">
-        <description>Parameter value received but not yet validated or set. A subsequent PARAM_ACK or PARAM_EXT_ACK will follow once operation is completed with the actual result. These are for parameters that may take longer to set. Instead of waiting for an ACK and potentially timing out, you will immediately receive this response to let you know it was received.</description>
+        <description>Parameter value received but not yet set/accepted. A subsequent PARAM_TRANSACT_ACK or PARAM_EXT_ACK with the final result will follow once operation is completed. This is returned immediately for parameters that take longer to set, indicating taht the the parameter was recieved and does not need to be resent.</description>
       </entry>
     </enum>
     <enum name="CAMERA_MODE">
@@ -4800,9 +4800,13 @@
       <field type="uint8_t" name="base_mode" enum="MAV_MODE">The new base mode.</field>
       <field type="uint32_t" name="custom_mode">The new autopilot-specific mode. This field can be ignored by an autopilot.</field>
     </message>
-    <!-- reserved for PARAM_VALUE_UNION -->
-    <message id="19" name="PARAM_ACK">
-      <description>Response from a PARAM_SET message.</description>
+    <!-- IDs 15-17 reserved for PARAM_VALUE_UNION and other param messages -->
+    <message id="19" name="PARAM_TRANSACT_ACK">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Response from a PARAM_SET message when it is used in a transaction.</description>
+      <field type="uint8_t" name="target_system">Id of system that sent PARAM_SET message.</field>
+      <field type="uint8_t" name="target_component">Id of system that sent PARAM_SET message.</field>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="float" name="param_value">Parameter value (new value if PARAM_ACCEPTED, current value otherwise)</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type.</field>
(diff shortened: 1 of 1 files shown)


DanielePettenuzzo:
DC4AA0CC-78F9-4A21-BFE4-8CE7DBDA0BBB

message_definitions/v1.0/common.xml
```diff
@@ -1372,6 +1372,21 @@
...
+      </entry>
+      <entry value="1" name="PARAM_EXT_TRANSACTION_RESPONSE_FAILED">
+        <description>Transaction failed.</description>
+      </entry>
+      <entry value="2" name="PARAM_EXT_TRANSACTION_RESPONSE_UNSUPPORTED">
+        <description>Transaction unsupported.</description>
+      </entry>
+      <entry value="3" name="PARAM_EXT_TRANSACTION_RESPONSE_INPROGRESS">
+        <description>Transaction in progress.</description>
+      </entry>
```
mattstraehl, line 1388: Will the "in progress" response be sent to the GS periodically until followed by either success/failure, or does the GS have to check back periodically until either success/failure is received?

message_definitions/v1.0/common.xml
```diff
@@ -6702,6 +6717,22 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
     </message>
+    <message id="325" name="PARAM_EXT_START_TRANSACTION">
+      <description>Request to start a new parameter transaction. The response (ack) will contain the same message but with a response attached to it.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
```
mattstraehl, line 6723: If I'm not mistaken, aren't **System ID** and **Component ID** included in the messages by default? At least for e.g. the WiFi Config message they seem to be included although not explicitly specified as part of the message definition:
https://github.com/mavlink/c_library_v2/blob/master/common/mavlink_msg_wifi_config_ap.h#L85-L86

hamishwillee:
> There is no way to attach some description/metadata to a given parameter.

parameter metadata - should be handled as part of https://github.com/mavlink/mavlink/issues/1346. 

The metadata are almost the same - though there is a new type "MAV_PARAM_EXT_TYPE_CUSTOM" and we probably need to identify that these are extended not ordinary params (?). Also, if these need to be set "as a group" then do we need to identify the associated parameters in metadata?

message_definitions/v1.0/common.xml
hamishwillee: @mattstraehl The sender's system/component ID are always included in the message as part of the packet definition. The **target** system/component id are included in the payload only if explicitly defined as above (i.e. the names are a convention). If these are not specified then this message would be a broadcast, effectively telling every component on the network to start a transaction.

hamishwillee:
@DanielePettenuzzo So in current version it sequence is:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG4gICAgR0NTLT4-RHJvbmU6IFBBUkFNX1NFVCAocGFyYW0gbmFtZSwgdmFsdWUsIC4uLilcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXQgKGZvciBQQVJBTV9WQUxVRSlcbiAgICBEcm9uZS0-PkRyb25lOiBXcml0ZSBwYXJhbWV0ZXIgdmFsdWVcbiAgICBEcm9uZS0-PkdDUzogUEFSQU1fVkFMVUUgKG5hbWUsIHZhbHVlIC4uLilcbiAgICBHQ1MtPj5HQ1M6IChvcHRpb25hbCkgVXBkYXRlIGNhY2hlIGZvciBQQVJBTV9WQUxVRVxuICAgIEdDUy0tPj5Ecm9uZTogT24gdGltZW91dCByZXN0YXJ0IHRoaXMgc2VxdWVuY2UiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ)

And the change is ()?:

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG5cbiAgICBcblxuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9FWFRfU1RBUlRfVFJBTlNBQ1RJT04gKHRhcmdldCwgY2xpZW50aWQpXG4gICAgR0NTLT4-R0NTOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fRVhUX1NUQVJUX1RSQU5TQUNUSU9OKVxuICAgIERyb25lLT4-R0NTOiBbQUNLXSBQQVJBTV9FWFRfU1RBUlRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLGlkKVxuICAgIERyb25lLT4-RHJvbmUgOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fU0VUIG9yIFBBUkFNX0VYVF9DT01NSVRfVFJBTlNBQ1RJT04pXG4gICAgTm90ZSBvdmVyIEdDUyxEcm9uZTogUmVwZWF0IGJlbG93IG9yIGFsbCBwYXJhbXNcbiAgICBHQ1MtPj5Ecm9uZTogUEFSQU1fU0VUIChwYXJhbSBuYW1lLCB2YWx1ZSwgLi4uKVxuICAgIEdDUy0-PkdDUzogU3RhcnQgdGltZW91dCAoZm9yIFBBUkFNX1ZBTFVFKVxuICAgIERyb25lLT4-RHJvbmU6IFdyaXRlIHBhcmFtZXRlciB2YWx1ZVxuICAgIERyb25lLT4-R0NTOiBQQVJBTV9WQUxVRSAobmFtZSwgdmFsdWUgLi4uKVxuICAgIEdDUy0-PkdDUzogKG9wdGlvbmFsKSBVcGRhdGUgY2FjaGUgZm9yIFBBUkFNX1ZBTFVFXG4gICAgR0NTLS0-PkRyb25lOiBPbiB0aW1lb3V0IHJlc3RhcnQgdGhpcyBzZXF1ZW5jZVxuICAgIE5vdGUgb3ZlciBHQ1MsRHJvbmU6IFJlcGVhdCBhYm92ZSBmb3IgYWxsIHBhcmFtc1xuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9FWFRfQ09NTUlUX1RSQU5TQUNUSU9OICh0YXJnZXQpXG4gICAgR0NTLT4-RHJvbmU6IFtBQ0xdIFBBUkFNX0VYVF9DT01NSVRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLi4uKSIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

Questions:
1. Why do we need a client id and why do we need a server transaction id? 
   - You know the source system and component id in all cases, so there is no question about the origin. 
   - The response field alone is enough to know if this is a response or not.
   - So unless you are allowing multiple transactions at a time (which you cannot support, since the system would not know which PARAM_SET belong to which commit) what is the need for multiple transaction ides?
2. If there is a reason, how are these allocated, and why do they need to be so big?
3. Is a long running transaction cancelled? IF so, recommend another message for that following pattern here: https://mavlink.io/en/services/command.html#long_running_commands
4. If you're going to send "in progress notifications" that's nice, but not much a GCS can do with it without knowing either what proportion is complete - and in this case we don't know how many your sending. Might be worth the start commit providing a count.

hamishwillee:
@auturgy FYI, hope to talk about this in tonight's MAVLink meeting. @DanielePettenuzzo @mattstraehl - can you attend?

DanielePettenuzzo:
I will be there but might be a few minutes late.

auturgy:
Perhaps have a look at https://github.com/ArduPilot/ardupilot/pull/13911

hamishwillee:
Note that PARAM_VALUE should probably not be sent as ack in transaction - only send after actually set as can fail.

message_definitions/v1.0/common.xml
DanielePettenuzzo: Usually with mavlink commands QGC checks periodically and the vehicle continues responding with in progress until it's done. What do you think?

message_definitions/v1.0/common.xml
DanielePettenuzzo: Exactly. So in our case that we want to be able to use this protocol with multiple mavlink components in the future it is better that we target a specific component with the transaction request instead of broadcasting it.

DanielePettenuzzo:
@hamishwillee I'm not sure how we should handle this. We probably still need acks for each param_set otherwise the retry mechanism will not work. Probably in case of a commit transaction success QGC can keep what it received in each ack and in case the vehicle fails to commit then qgc should request those params again. 
Otherwise if we want to avoid sending parameters twice we should probably store the previous values on QGC so that it can automatically revert to them in case the transaction fails. 
How do you think we should handle this? @julianoes @mattstraehl  any ideas?

DanielePettenuzzo:
I will remove the _EXT_ from PARAM_EXT_START_TRANSACTION and PARAM_EXT_COMMIT_TRANSACTION since we want this to be transport independent. I will also add the transport as an enum in the message. So far the options will be:
- PARAM
- PARAM_EXT
- Mavftp
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG5cbiAgICBcblxuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTiAodGFyZ2V0LCBjbGllbnRpZClcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXQgKGZvciBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTilcbiAgICBEcm9uZS0-PkdDUzogW0FDS10gUEFSQU1fU1RBUlRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLGlkKVxuICAgIERyb25lLT4-RHJvbmUgOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fU0VUIG9yIFBBUkFNX0NPTU1JVF9UUkFOU0FDVElPTilcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBSZXBlYXQgYmVsb3cgb3IgYWxsIHBhcmFtc1xuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9TRVQgKHBhcmFtIG5hbWUsIHZhbHVlLCAuLi4pXG4gICAgR0NTLT4-R0NTOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fVkFMVUUpXG4gICAgRHJvbmUtPj5Ecm9uZTogV3JpdGUgcGFyYW1ldGVyIHZhbHVlXG4gICAgRHJvbmUtPj5HQ1M6IFBBUkFNX1ZBTFVFIChuYW1lLCB2YWx1ZSAuLi4pXG4gICAgR0NTLT4-R0NTOiAob3B0aW9uYWwpIFVwZGF0ZSBjYWNoZSBmb3IgUEFSQU1fVkFMVUVcbiAgICBHQ1MtLT4-RHJvbmU6IE9uIHRpbWVvdXQgcmVzdGFydCB0aGlzIHNlcXVlbmNlXG4gICAgTm90ZSBvdmVyIEdDUyxEcm9uZTogUmVwZWF0IGFib3ZlIGZvciBhbGwgcGFyYW1zXG4gICAgR0NTLT4-RHJvbmU6IFBBUkFNX0NPTU1JVF9UUkFOU0FDVElPTiAodGFyZ2V0KVxuICAgIERyb25lLT4-R0NTOiBbQUNMXSBQQVJBTV9DT01NSVRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLi4uKSIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

DanielePettenuzzo:
@hamishwillee I pushed a commit with the following changes:
- remove client_transaction_id and server_transaction_id fields
- remove _EXT_ from the transaction message names since these messages are independent from the transport
- add enum to choose transaction transport (only to the start message). Do we need it also on the commit message?

DanielePettenuzzo:
I also updated the message numbers so that we can also get this one in: https://github.com/mavlink/mavlink/pull/1430

message_definitions/v1.0/common.xml
mattstraehl: That makes sense, I was actually wondering if it's the MAVLink convention that for every message exactly one response is sent back. If that's the case it should be clear to anyone familiar with MAVLink.

message_definitions/v1.0/common.xml
mattstraehl: Got it, thank you both for the clarification.

message_definitions/v1.0/common.xml
```diff
@@ -1372,6 +1372,33 @@
...
+    </enum>
+    <enum name="PARAM_TRANSACTION_TRANSPORT">
+      <description>Possible transport layers to set and get parameters via mavlink during a parameter transaction.</description>
+      <entry value="0" name="PARAM_TRANSACTION_TRANSPORT_PARAM">
+        <description>Transaction over param transport.</description>
+      </entry>
+      <entry value="1" name="PARAM_TRANSACTION_TRANSPORT_PARAM_EXT">
+        <description>Transaction over param_ext transport.</description>
+      </entry>
+      <entry value="2" name="PARAM_TRANSACTION_TRANSPORT_MAVFTP">
```
hamishwillee, line 1398: FWIW I don't understand how this case is used. Perhaps it could be removed until that is clarified/implementable.

@LorenzMeier you mentioned this. What are you envisaging here? I.e how does this fit with https://mavlink.io/en/services/ftp.html#uploading-a-file

message_definitions/v1.0/common.xml
```diff
@@ -6702,6 +6729,19 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
     </message>
+    <message id="328" name="PARAM_START_TRANSACTION">
+      <description>Request to start a new parameter transaction. Multiple kinds of transport layers can be used to exchange parameters in the transaction (param, param_ext and mavftp). The response (ack) will contain the same message but with a response attached to it.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Message acceptance response (sent back to GS).</field>
```
hamishwillee, line 6736: Description wrong.

hamishwillee:
@DanielePettenuzzo 

> add enum to choose transaction transport (only to the start message). Do we need it also on the commit message?

I don't think so. 

> @hamishwillee I'm not sure how we should handle this. We probably still need acks for each param_set otherwise the retry mechanism will not work. Probably in case of a commit transaction success QGC can keep what it received in each ack and in case the vehicle fails to commit then qgc should request those params again.
> Otherwise if we want to avoid sending parameters twice we should probably store the previous values on QGC so that it can automatically revert to them in case the transaction fails.

In normal use the PARAM_VALUES are sent as broadcast to ensure that other systems that also have parameters get them too - it is a sort of "poor mans" way of allowing multiple GCS or other systems on network to keep track of param changes (it is unreliable of course, because there is no way for a system that missed the message to know it was sent, other than the original setter).

The point being is that if they are sent they will be considered as "SET" by **other systems** irrespective of what your first GCS does. 

Can we use an alternative/new ACK for the transaction case (only) ? The drone could send all the PARAM_VALUES out after committing the transaction. The GCS would ignore these.

message_definitions/v1.0/common.xml
DanielePettenuzzo: My understanding was that we start/commit a transaction in the same way but setting the parameter won't happen via the mavlink param/param_ext protocols but storing parameters in files and uploading/downloading them via mavlink ftp. If this needs a further discussion I can remove it for now

DanielePettenuzzo:
@hamishwillee so the transaction would basically look the same as before but with one additional step after the commit transaction.

Moreover I think there is also some naming confusion. 

For the param protocol we have:
- PARAM_SET: sets parameter
- PARAM_VALUE: acks by sending value back with a response and is also used for a "param get"

For the param_ext protocol we have:
- PARAM_EXT_SET: sets parameter
- PARAM_EXT_VALUE: sends value back during a "param get"
- PARAM_EXT_ACK: acks by sending value back with a response

Maybe we could create also the PARAM_ACK and then have the two protocols consistent and do exactly what you proposed. Respond the the PARAM_SET with a PARAM_ACK so that only the sender will update it's parameter and then set a PARAM_VALUE only when the commit is done and we know if it was successful or not. With this last message then also the other components can update their params and the sender can also be sure that what he got from the ACK makes sense or not.

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG5cbiAgICBcblxuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTiAodGFyZ2V0LCBjbGllbnRpZClcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXQgKGZvciBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTilcbiAgICBEcm9uZS0-PkdDUzogW0FDS10gUEFSQU1fU1RBUlRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLGlkKVxuICAgIERyb25lLT4-RHJvbmUgOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fU0VUIG9yIFBBUkFNX0NPTU1JVF9UUkFOU0FDVElPTilcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBSZXBlYXQgYmVsb3cgZm9yIGFsbCBwYXJhbXNcbiAgICBHQ1MtPj5Ecm9uZTogUEFSQU1fU0VUIChwYXJhbSBuYW1lLCB2YWx1ZSwgLi4uKVxuICAgIEdDUy0-PkdDUzogU3RhcnQgdGltZW91dCAoZm9yIFBBUkFNX0FDSylcbiAgICBEcm9uZS0-PkRyb25lOiBXcml0ZSBwYXJhbWV0ZXIgdmFsdWVcbiAgICBEcm9uZS0-PkdDUzogUEFSQU1fQUNLIChuYW1lLCB2YWx1ZSAuLi4pXG4gICAgR0NTLT4-R0NTOiAob3B0aW9uYWwpIFVwZGF0ZSBjYWNoZSB1c2luZyBQQVJBTV9BQ0tcbiAgICBHQ1MtLT4-RHJvbmU6IE9uIHRpbWVvdXQgcmVzdGFydCB0aGlzIHNlcXVlbmNlXG4gICAgTm90ZSBvdmVyIEdDUyxEcm9uZTogUmVwZWF0IGFib3ZlIGZvciBhbGwgcGFyYW1zXG4gICAgR0NTLT4-RHJvbmU6IFBBUkFNX0NPTU1JVF9UUkFOU0FDVElPTiAodGFyZ2V0KVxuICAgIERyb25lLT4-R0NTOiBbQUNMXSBQQVJBTV9DT01NSVRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLi4uKVxuICAgIE5vdGUgb3ZlciBHQ1MsRHJvbmU6IFJlcGVhdCBiZWxvdyBmb3IgYWxsIHBhcmFtc1xuICAgIERyb25lLT4-R0NTOiBQQVJBTV9WQUxVRSAobmFtZSwgdmFsdWUgLi4uKVxuICAgIEdDUy0-PkdDUzogKG9wdGlvbmFsKSBVcGRhdGUgY2FjaGUgZm9yIFBBUkFNX1ZBTFVFXG4gICAgTm90ZSBvdmVyIEdDUyxEcm9uZTogUmVwZWF0IGFib3ZlIGZvciBhbGwgcGFyYW1zIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

DanielePettenuzzo:
As soon as we clarify this one I will also address the remaining review points and push them

message_definitions/v1.0/common.xml
hamishwillee: @DanielePettenuzzo The confusion might just be with me, but this feels too theoretical since we don't have a mechanism for uploading params via mavlinkftp yet. We are talking about using MAVFTP to get the set of parameters from vehicle, but we haven't talked about how we might use it for provisioning parameters yet. 

My view is that there is no cost to adding an additional enum value if we need it later. Perhaps @julianoes has more clarity though.

hamishwillee:
@DanielePettenuzzo You learn something new every day; I had assumed that the protocols mirrored each other with respect to sending out a value after setting (as well as on request). Thanks for that. 

I like your solution of adding a PARAM_ACK. Nothing broken provided it is only used for the transaction case.

Arguably there is no need to even send out all the PARAM_VALUEs at the end because 
- using transactions is completely new sequence and nothing is monitoring for their absence (only sending out incorrect would be "wrong")
- the ext versions doesnt broadcast this, so I guess the expectation is that it isn't needed; which works for me because it is broken.

@julianoes make sense to you? Essentially if using "normal" params with transactions instead of sending back the value we send a new ack message.

DanielePettenuzzo:
@hamishwillee 
Ok then to summarize:
- When using transactions parameters cannot be set unless a transaction has started and parameters won't be applied until a transaction has been commited
- To a PARAM_SET we reply with a PARAM_ACK (the latter is the only missing message at the moment)
- No need to send PARAM_VALUE after transaction has finished (because of the two reasons mentioned by Hamish)
- PARAM_VALUE will now be used only for reading parameters (usually during connection phase)
- For the PARAM_EXT I will mirror 1 to 1 what we do with regular PARAM. For PARAM_EXT no new messages are needed since we already have both the PARAM_EXT_VALUE and PARAM_EXT_ACK messages.

The up to date protocol is the following:
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG5cbiAgICBcblxuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTiAodGFyZ2V0LCBjbGllbnRpZClcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXQgKGZvciBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTilcbiAgICBEcm9uZS0-PkdDUzogW0FDS10gUEFSQU1fU1RBUlRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLGlkKVxuICAgIERyb25lLT4-RHJvbmUgOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fU0VUIG9yIFBBUkFNX0NPTU1JVF9UUkFOU0FDVElPTilcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBSZXBlYXQgYmVsb3cgZm9yIGFsbCBwYXJhbXNcbiAgICBHQ1MtPj5Ecm9uZTogUEFSQU1fU0VUIChwYXJhbSBuYW1lLCB2YWx1ZSwgLi4uKVxuICAgIEdDUy0-PkdDUzogU3RhcnQgdGltZW91dCAoZm9yIFBBUkFNX0FDSylcbiAgICBEcm9uZS0-PkRyb25lOiBXcml0ZSBwYXJhbWV0ZXIgdmFsdWUgKGRvIG5vdCBjb21taXQgeWV0KVxuICAgIERyb25lLT4-R0NTOiBQQVJBTV9BQ0sgKG5hbWUsIHZhbHVlIC4uLilcbiAgICBHQ1MtPj5HQ1M6IChvcHRpb25hbCkgVXBkYXRlIGNhY2hlIHVzaW5nIFBBUkFNX0FDS1xuICAgIEdDUy0tPj5Ecm9uZTogT24gdGltZW91dCByZXN0YXJ0IHRoaXMgc2VxdWVuY2VcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBSZXBlYXQgYWJvdmUgZm9yIGFsbCBwYXJhbXNcbiAgICBHQ1MtPj5Ecm9uZTogUEFSQU1fQ09NTUlUX1RSQU5TQUNUSU9OICh0YXJnZXQpXG4gICAgRHJvbmUtPj5Ecm9uZTogQ29tbWl0IHBhcmFtZXRlciB2YWx1ZShzKVxuICAgIERyb25lLT4-R0NTOiBbQUNMXSBQQVJBTV9DT01NSVRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLi4uKSIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In0sInVwZGF0ZUVkaXRvciI6ZmFsc2V9)

message_definitions/v1.0/common.xml
DanielePettenuzzo: I am totally ok with this. For now I will remove the mavftp value from the enum and we should follow up on that before adding it back

DanielePettenuzzo:
@hamishwillee I addressed your last review and also added the PARAM_ACK message. I have two questions about this:
- PARAM_ACK is already also a name of an enum. Will this create conflicts?
- I added PARAM_ACK (message id = 19) before all the other PARAM_* to group them all together but I noticed there is this comment between message 11 and 20 that I don't know if it's outdated or not: `<!-- reserved for PARAM_VALUE_UNION -->`. And if it's not outdated does this reserve all messages between 11 and 20?

message_definitions/v1.0/common.xml
```diff
@@ -4800,9 +4800,13 @@
       <field type="uint8_t" name="base_mode" enum="MAV_MODE">The new base mode.</field>
       <field type="uint32_t" name="custom_mode">The new autopilot-specific mode. This field can be ignored by an autopilot.</field>
     </message>
-    <!-- reserved for PARAM_VALUE_UNION -->
-    <message id="19" name="PARAM_ACK">
-      <description>Response from a PARAM_SET message.</description>
+    <!-- IDs 15-17 reserved for PARAM_VALUE_UNION and other param messages -->
+    <message id="19" name="PARAM_TRANSACT_ACK">
+      <wip/>
```
hamishwillee, line 4805: @DanielePettenuzzo FYI, added wip and comment to all new lines. These get removed when you are happy open questions are all answered and there is an implementation.

message_definitions/v1.0/common.xml
```diff
@@ -4776,7 +4800,18 @@
       <field type="uint8_t" name="base_mode" enum="MAV_MODE">The new base mode.</field>
       <field type="uint32_t" name="custom_mode">The new autopilot-specific mode. This field can be ignored by an autopilot.</field>
     </message>
-    <!-- reserved for PARAM_VALUE_UNION -->
+    <!-- IDs 15-17 reserved for PARAM_VALUE_UNION and other param messages -->
```
hamishwillee, line 4803: @DanielePettenuzzo 

> I added PARAM_ACK (message id = 19) before all the other PARAM_* to group them all together but I noticed there is this comment between message 11 and 20 that I don't know if it's outdated or not: . And if it's not outdated does this reserve all messages between 11 and 20?

Good place to put this. Highly likely this was just "keep some space free for this single message in this region".  I have cleaned up the intent.

message_definitions/v1.0/common.xml
```diff
@@ -4776,7 +4800,18 @@
...
+    <message id="19" name="PARAM_ACK_TRANSACTION">
```
hamishwillee, line 4804: > PARAM_ACK is already also a name of an enum. Will this create conflicts?

Yes. I have renamed this to make it clear this ACK is ONLY sent in transactions. You can't undo expected behaviour which is sending PARAM_VALUE when param is set. 

Name OK with you?

message_definitions/v1.0/common.xml
```diff
@@ -4776,7 +4800,18 @@
...
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Response from a PARAM_SET message when it is used in a transaction.</description>
+      <field type="uint8_t" name="target_system">Id of system that sent PARAM_SET message.</field>
```
hamishwillee, line 4808: Note addition of system and component id. This is not broadcast - it is intended for the sender only. Note, descrption above makes clear this is only sent for transaction sets.

message_definitions/v1.0/common.xml
```diff
@@ -4798,7 +4833,7 @@
...
+      <description>Set a parameter value (write new value to permanent storage). Within a transaction the recieving componenent should respond with PARAM_ACK_TRANSACTION to the setter component. IMPORTANT: If sent outside a transaction the receiving component should acknowledge the new parameter value by broadcasting a PARAM_VALUE message to all communication partners (broadcasting ensures that multiple GCS all have an up-to-date list of all parameters). If the sending GCS did not receive a PARAM_VALUE or PARAM_ACK_TRANSACTION message within its timeout time, it should re-send the PARAM_SET message. The parameter microservice is documented at https://mavlink.io/en/services/parameter.html</description>
```
hamishwillee, line 4836: @DanielePettenuzzo And note here, I modified this back so PARAM_VALUE is sent for non-transactions and PARAM_ACK_TRANSACTION is sent within transactions.

It does not say that PARAM_VALUE is cached and emitted at the end of the transaction. @julianoes My feeling is that even though we emit PARAM _VALUE outside transactions to update all other GCS, we are not required to do so at the end of a transaction. Reasoning being that approach is not used in the more recent param protocol - I think it is flawed and we we don't have to for consistency, we should not. OK?

hamishwillee:
@DanielePettenuzzo I've fixed up to address your comments and added notes FYI. If you're happy and there is no strong objection then either I or @julianoes can merge and we can iterate (don't wait on me Julian if Daniele is happy, it is Friday!)

message_definitions/v1.0/common.xml
DanielePettenuzzo: Sound good. So this is the equivalent of PARAM_EXT_ACK but for PARAM_* when using transactions, right?

DanielePettenuzzo:
@hamishwillee @julianoes this looks good for me now. If there are no other issues or things to be discussed then we can merge. I will come back to this next week after I start implementing the protocol on the vehicle side. 

Would be cool to also get this one in if there are no further issues: https://github.com/mavlink/mavlink/pull/1430

hamishwillee:
Merging now, as discussed.

DonLakeFlyer:
FYI: This is a massive amount of work to support in QGC.

I don't understand the problem this is trying to solve. Seems like possibly hitting a simpler problem with a sledgehammer solution. Can someone explain in more detail why this is needed?

hamishwillee:
@DonLakeFlyer Yes. The transaction stuff is provided to allow multiple params to be committed in one go, rather than being set and acknowledged individually. The use case is setting something like wifi or LTE credentials where the multiple parameter settings like password and new password must be set together and fail or succeed as a group - having a transaction allows the values to be set together and only rejected at the end when we have an end result (which make come some point after setting).

Originally we had discussed a configuration message for this, but with so many different types of settings there thinking was that it is better to reuse parameters for this and have all the common interactions for checking resending etc in one place.

Note as part of this we wanted to take the opportunity to reimplement PARAM_EXT variant with the fields properly ordered so that the data can be truncated. 

@julianoes shares your concerns on this because it has to be implemented by mavsdk too. Can you and Julian and @DanielePettenuzzo get in a meeting and discuss?

LorenzMeier:
@DonLakeFlyer You could call the current usage (not the parameter protocol itself) in the GCS of the autopilot parameters incomplete and in some cases dangerous (it works right now because people are aware, but nothing in the implementation ensures safety). E.g. setting PID gains independently during flight can crash a vehicle, despite of the individual values being correct if you look at them one-by-one. There is no method from switching from one PID set to another that would be atomic and thus safe. There are plenty of other cases, like camera triggers, radio link / MAVLink settings (e.g. if you have to set baud rate and parity, and commit them at the same time to switch from one configuration to another).

Where I'm getting at: The fact that transactions are not supported yet is something we have worked around by forcing system reboots and/or being disarmed, which is a crude approach of having a transaction. But it is needed. And so even though this might introduce churn in the GCS, it is very much necessary as we are starting to have more industrial systems out there that can't be power-cycled all the time entirely as the hobby-systems could be.

DonLakeFlyer:
Thanks @LorenzMeier. Good examples, I get it now. Will take me a little time to look through this. I'm in the middle of a big pile of other high prio stuff.

DonLakeFlyer:
It also would have been nice if someone pinged me about this before it was merged!

DonLakeFlyer:
> Arguably there is no need to even send out all the PARAM_VALUEs at the end because

They changes should be broadcast at the end to update other gcs which may also be connected to the vehicle. Even given all the caveats of possible lost packets with this it's still better to make a best effort to keep all connected gcs up to date on what parameter values are.

DonLakeFlyer:
My comments from a quick look:
* I would change the message set to: `PARAM_TRANSACTION_START, PARAM_TRANSACTION_COMMIT, PARAM_TRANSACTION_ACK`.
* `PARAM_TRANSACTION_ACK` is used as the ack for everything. It's odd to have some commands be multi-purpose in both directions. Just include what is being acked and a result code. I think this makes thing clearer if you are looking at mavlink message traffic. Since you need to add an `ACK` command anyway for `PARAM_SET`, make it generic.

DonLakeFlyer:
What happens if the ack from `PARAM_TRANSACTION_COMMIT` is lost from vehicle->gcs. I assume the gcs should resend the commit but that won't work because the transaction will be closed on the vehicle side. There might be a need to add some sort of transaction id which is used throughout. Then on the commit message to the vehicle if you get a commit with the same id more than once it's ok you just ack it again.

DanielePettenuzzo:
@DonLakeFlyer Yes this is a good point. In case we use the id to figure out which transaction a commit is for, does this mean that I would need to keep track of some sort of transaction history on the vehicle side where I can go back and check on the status of an older transaction?

DonLakeFlyer:
I think you'll need to keep track of transactions by id and relate it to the gcs system id which sent it. From a state standpoint you could have multiple GCS requesting parameter transactions at the same time. Then the question becomes whether you support multiple transactions at the same time from the same gcs. I would say no. So if you get a new transaction from the same gcs while and old transaction is pending you throw the old one out.

DanielePettenuzzo:
Ok sounds good. So from my understanding from the current protocol we are missing the part where we broadcast again all parameters after a transaction so that all ground stations can be up to date. 

Instead, @DonLakeFlyer @hamishwillee  regarding the param_v3 vs param_ext messages, @dogmaphobic raised the concern that we will break compatibility between qgc and all mavlink components by doing this change. Is there any chance we can be compatible with both for some time so we can use the new param_v3 that has its value trimmed but still we don't break compatibility with all mavlink components?

FYI this is the param_v3 pr: https://github.com/mavlink/mavlink/pull/1430

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtO1xuICAgIHBhcnRpY2lwYW50IEdDU1xuICAgIHBhcnRpY2lwYW50IERyb25lXG5cbiAgICBcblxuICAgIEdDUy0-PkRyb25lOiBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTiAodGFyZ2V0LCBjbGllbnRpZClcbiAgICBHQ1MtPj5HQ1M6IFN0YXJ0IHRpbWVvdXQgKGZvciBQQVJBTV9TVEFSVF9UUkFOU0FDVElPTilcbiAgICBEcm9uZS0-PkdDUzogW0FDS10gUEFSQU1fU1RBUlRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLGlkKVxuICAgIERyb25lLT4-RHJvbmUgOiBTdGFydCB0aW1lb3V0IChmb3IgUEFSQU1fU0VUIG9yIFBBUkFNX0NPTU1JVF9UUkFOU0FDVElPTilcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBSZXBlYXQgYmVsb3cgZm9yIGFsbCBwYXJhbXNcbiAgICBHQ1MtPj5Ecm9uZTogUEFSQU1fU0VUIChwYXJhbSBuYW1lLCB2YWx1ZSwgLi4uKVxuICAgIEdDUy0-PkdDUzogU3RhcnQgdGltZW91dCAoZm9yIFBBUkFNX0FDSylcbiAgICBEcm9uZS0-PkRyb25lOiBXcml0ZSBwYXJhbWV0ZXIgdmFsdWUgKGRvIG5vdCBjb21taXQgeWV0KVxuICAgIERyb25lLT4-R0NTOiBQQVJBTV9BQ0sgKG5hbWUsIHZhbHVlIC4uLilcbiAgICBHQ1MtPj5HQ1M6IChvcHRpb25hbCkgVXBkYXRlIGNhY2hlIHVzaW5nIFBBUkFNX0FDS1xuICAgIEdDUy0tPj5Ecm9uZTogT24gdGltZW91dCByZXN0YXJ0IHRoaXMgc2VxdWVuY2VcbiAgICBOb3RlIG92ZXIgR0NTLERyb25lOiBSZXBlYXQgYWJvdmUgZm9yIGFsbCBwYXJhbXNcbiAgICBHQ1MtPj5Ecm9uZTogUEFSQU1fQ09NTUlUX1RSQU5TQUNUSU9OICh0YXJnZXQpXG4gICAgRHJvbmUtPj5Ecm9uZTogQ29tbWl0IHBhcmFtZXRlciB2YWx1ZShzKVxuICAgIERyb25lLT4-R0NTOiBbQUNMXSBQQVJBTV9DT01NSVRfVFJBTlNBQ1RJT04gKHJlc3BvbnNlLi4uKVxuICAgIE5vdGUgb3ZlciBHQ1MsRHJvbmU6IFJlcGVhdCBiZWxvdyBmb3IgYWxsIHBhcmFtc1xuICAgIERyb25lLT4-R0NTOiBQQVJBTV9WQUxVRSAobmFtZSwgdmFsdWUgLi4uKVxuICAgIEdDUy0-PkdDUzogKG9wdGlvbmFsKSBVcGRhdGUgY2FjaGUgZm9yIFBBUkFNX1ZBTFVFXG4gICAgTm90ZSBvdmVyIEdDUyxEcm9uZTogUmVwZWF0IGFib3ZlIGZvciBhbGwgcGFyYW1zIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

PR #1429, opened 2020-07-22
common: SCALLED_PRESURE: add press diff temperature

IamPete1:
Adds a second temperature field to SCALLED_PRESSURE to allow sending back of the air temp from the airspeed sensor. This is often quite different to the baro temp due to location. 

https://github.com/ArduPilot/mavlink/pull/142

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4879,7 +4879,9 @@
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="float" name="press_abs" units="hPa">Absolute pressure</field>
       <field type="float" name="press_diff" units="hPa">Differential pressure 1</field>
-      <field type="int16_t" name="temperature" units="cdegC">Temperature</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
If this will not always be sent, you should define a value to represent "not sent" usually the max value for an int.

Given that we have three of these messages is it worth designing a single new one with an index? I'm asking, not advocating. @julianoes @auturgy

hamishwillee:
OK then. In it goes. Thanks!

amilcarlucas:
Using UINT16_MAX on a int16 type is wrong !!!!

IamPete1:
@amilcarlucas Opps, your right, I think we should be using 0 for 'no value' in any case since its a extension.

hamishwillee:
@amilcarlucas Good catch :-0. 

@IamPete1 can you create PR to fix?
I don't think it is too late as you're likely the only implementer at this point. Yes the default value should be 0 for not supported on an extension. Usually we'd use 0 indicates not supported and perhaps say 1 is used for 0 or 1

PR #1430, opened 2020-07-22
common: fix PARAM_EXT trimming

### LINKED ISSUE #1168: `PARAM_EXT_SET` and `PARAM_EXT_VALUE` not correctly truncated
It was raised at the PX4 DevSummit by one of the participants that the messages [PARAM_EXT_SET/VALUE/ACK](https://github.com/mavlink/mavlink/blob/95031e31a5eec47dc951fc09e2939071b8cf3c2b/message_definitions/v1.0/common.xml#L5008-L5030) are not correctly truncated (trimmed because of trailing zeros).

Looking at the C generated code of the packed struct this seems to be true:

https://github.com/mavlink/c_library_v2/blob/beb789085dde84cfafc3d227fca1be68199ac402/common/mavlink_msg_param_ext_value.h#L6-L13

@LorenzMeier are we missing something here? Or is the generated code not correctly ordering the fields for the arrays? Could it be an issue of the parser of uint8_t vs. char?

julianoes:
It turns out that PARAM_EXT messages were not trimmed as we thought they would be. The fields of a message are ordered by the size of the type, however it does not matter if the type is an array or not.
This means that char[128] is not at the end of the message and therefore not automatically trimmed.

This patch introduces new messages called _TRIMMED which has the param_value as part of the extension meaning that it will be at the end and trimmed correctly.

I'm open to other suggestions on how to do this or better naming suggestions.

FYI @DanielePettenuzzo.
Closes #1168

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6711,7 +6711,6 @@
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint16_t" name="param_count">Total number of parameters</field>
       <field type="uint16_t" name="param_index">Index of this parameter</field>
-      <extensions/>
       <field type="char[128]" name="param_value">Parameter value (zeros get trimmed)</field>
     </message>
     <message id="326" name="PARAM_EXT_SET_TRIMMED">
@@ -6720,7 +6719,6 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
-      <extensions/>
       <field type="char[128]" name="param_value">Parameter value (zeros get trimmed)</field>
     </message>
     <message id="327" name="PARAM_EXT_ACK_TRIMMED">
@@ -6728,7 +6726,6 @@
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
-      <extensions/>
       <field type="char[128]" name="param_value">Parameter value (new value if PARAM_ACK_ACCEPTED, current value otherwise, zeros get trimmed)</field>
     </message>
     <message id="330" name="OBSTACLE_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6715,7 +6715,7 @@
       <field type="uint8_t" name="target_component">Component ID</field>
     </message>
     <message id="322" name="PARAM_EXT_VALUE">
-      <deprecated since="2020-07" replaced_by="PARAM_EXT_VALUE_TRIMMED">The new version of this message correctly trims zeros.</deprecated>
+      <deprecated since="2020-07" replaced_by="PARAM_V3_VALUE">The new version of this message correctly trims zeros.</deprecated>
       <description>Emit the value of a parameter. The inclusion of param_count and param_index in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout.</description>
       <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="char[128]" name="param_value">Parameter value</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6696,12 +6698,39 @@
...
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
     </message>
+    <message id="325" name="PARAM_EXT_VALUE_TRIMMED">
+      <description>Emit the value of a parameter. The inclusion of param_count and param_index in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout. Replacement for PARAM_EXT_VALUE.</description>
+      <field type="char[16]" name="param_id">Parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
+      <field type="uint8_t" name="param_type" enum="MAV_PARAM_EXT_TYPE">Parameter type.</field>
+      <field type="uint16_t" name="param_count">Total number of parameters</field>
+      <field type="uint16_t" name="param_index">Index of this parameter</field>
+      <extensions/>
```
hamishwillee, line 6714: Why an extension? This should be reordered to the end by the parser anyway because a char is same size as uint8

hamishwillee:
> The inclusion of `param_count` and `param_index` in the message allows the recipient to keep track of received parameters and allows them to re-request missing parameters after a loss or timeout. 

How this should work is not clear to me. Are these numbers for the whole set or for the particular transaction? What is the sequence to set all parameters and read all parameters? - e.g. https://mavlink.io/en/services/parameter.html#read_all

hamishwillee:
The param name is still restricted to 16 chars. Is this something we need to reconsider?

message_definitions/v1.0/common.xml
julianoes: Oh, right. It just goes from big types to smaller types, and the ordering is stable, so we can have it last without `<extensions/>`.

julianoes:
> how do these know if they should return the new type or old type?

They should return the new type, however, if they happen to return the old type, they don't support the new type.

> The param name is still restricted to 16 chars. Is this something we need to reconsider?
> Now this is at the end, should we allow a larger param value?
> Not sure about the name either. Maybe PARAM_V3

It's simple they said, just fix the ordering :laughing:. I don't know.

DanielePettenuzzo:
So my understanding to wrap up this PR is we need to:
- remove the extension
- maybe rename the message name (Hamish's proposal: PARAM_V3)

Then we could merge this but keep these new messages as WIP since I might have some useful feedback because @mattstraehl and I will implement this in the next few days. In the meantime we can continue the discussion about:
- param_name size: should we increase this and by how much
- param_value size: by how much can we increase this to reach the limit of the mavlink message? @julianoes do you know?

For these last two points we can maybe also get feedback from @LorenzMeier

hamishwillee:
@DanielePettenuzzo That's pretty good summary.

- I would rename the message but that is to my taste; we're hoping to get rid of the the EXT imminently, so calling it EXT_TRIMMED then becomes redundant.
- Definitely mark all of new messages as WIP so we can modify them until we have a stable implementation.
- I would get this in wip without the changes for max payload size or parameter size. 

@bkueng we're creating a new message for parameters to fix inefficiency in the current PARAM_EXT_ messages. Opportunity to increase the param name length and the size of data that can be sent. This is obviously easy in MAVLink, but I'm guessing that it might have significant cost in terms of implementation and consumed resources to do PX4 side. Is this worth thinking about?

julianoes:
Ok, I've removed the `<extensions/>` tag as it wasn't required.

Now, I see two ways going forward with this:
1. Leave everything as is and thus be compatible with previous PARAM_EXT. This makes it easier for implementers (e.g. me for MAVSDK) to support it because it's basically a drop in replacement. And we also re-use some of the messages and only use the TRIMMED variant for the ones that were flawed.
2. We design PARAM_v3, and fix all the things that we think are worth fixing. However, note that this will take time to spec as well as to implement. As an example, changing the param length has implications for the MAVSDK implementation as well as the camera xml definition. For instance we now need to know the param type when parsing the xml in order to know if a param name is too long or ok.

I would be all for speccing PARAM_V3 if we had seen a problem with either param names being too short, or the 128 param bytes not being enough, however, I have never heard that. The only thing I have heard was "param_value is not truncated and therefore inefficient"!

bkueng:
>   @bkueng we're creating a new message for parameters to fix inefficiency in the current PARAM_EXT_ messages. Opportunity to increase the param name length and the size of data that can be sent. This is obviously easy in MAVLink, but I'm guessing that it might have significant cost in terms of implementation and consumed resources to do PX4 side. Is this worth thinking about?


Should be fine for now. Having longer names would in some cases be nice, but is not required.

hamishwillee:
Thanks @bkueng - OK so we won't change anything else on this for now (no increased parameter name size or value size). I figure not increasing the size is OK - having a further field later that extends the size actually can be done efficiently.

@julianoes @DanielePettenuzzo I have renamed from `_TRIMMED` suffix, to using `PARAM_V3_` prefix - e.g. `PARAM_V3_ACK`. This is a preference because it provides an obvious upgrade path if we ever have to modify again. I also did `PARAM_V3_ACK` rather than `PARAM_ACK_V3` to indicate third version of the param protocol rather than message. Marked the messages as WIP so we can still modify if anyone hates this.

merging now so @DanielePettenuzzo can get to work on an implementation.

julianoes:
>  Marked the messages as WIP so we can still modify if anyone hates this.

I hate the fact that we now have V3 and EXT mixed up. We either have to add a full V3 set or call it EXT_TRIMMED.

hamishwillee:
Agreed. I prefer full V3 set.

julianoes:
A full set makes the implementation harder for MAVSDK which is why I preferred the incremental _TRIMMED approach. Instead of having a small switch to try trimmed and non trimmed for param ext I now need to make a third param implementation.

hamishwillee:
@julianoes I prefer the v3 approach but then I don't have to support this change as you do. My recollection is that @LorenzMeier originally suggested we should try to get the _EXT version "forgotten" but perhaps I misinterpreted. Could you could chat to him in person about this?

julianoes:
It's always easy to say "foo should be removed/forgotton" if you are not the one having to maintain it. This was the same with the old gimbal messages, it's the same with switching the camera definition xml to json and it's the same here.

And that's why I kept emphasizing that we either fix this in the smallest way possible (my suggestion) or come up with the next big protocol solving all of the problems which then makes the maintenance burden worth it because it we actually improved it. And I had warned that this would likely take a bit longer, and that's exactly where we are now.

dogmaphobic:
Note that by deprecating `PARAM_EXT_xxx` you are breaking every camera implementation out in the wild. They will all have to switch to `PARAM_V3_xxx`. As to the GCS, it will need to implement both as it needs to support existing cameras (using `PARAM_EXT`) and newer ones using `PARAM_V3`.

julianoes:
@dogmaphobic if it is deprecated but still implemented, that's fine, right? It won't actually disappear.

dogmaphobic:
To me it means: _don't use it_. In this case, it will be up to QGC. In order to support the new `PARAM_V3`, it will need to either duplicate everything so it supports both or drop `PARAM_EXT`. That also means every custom implementation of QCC (and there are various) will need to handle this as well. 

Yes, it can be done but it is rather drastic for a bug in the trimming logic.

julianoes:
Yes, which is why I initially only suggested to add two _TRIMMED variants for the messages that were flawed and leave the rest as is.

PR #1432, opened 2020-07-27
Common: add WINCH_STATUS message

rmackay9:
This PR adds a new WINCH_STATUS message for use with the Daiwa winch. https://ardupilot.org/copter/docs/common-daiwa-winch.html

I have a PR which will improve AP's driver to make use of the telemetry connection from the winch and send it to the ground station using this message.

The equivalent PR to add this message to ArduPilot's fork of common.xml is here: https://github.com/ArduPilot/mavlink/pull/144

All feedback welcome!

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4729,6 +4729,21 @@
         <description>Unit is sometimes working, sometimes not.</description>
       </entry>
     </enum>
+    <enum name="MAV_WINCH_STATUS_FLAG" bitmask="true">
+      <description>Winch status flags used in WINCH_STATUS</description>
+      <entry value="1" name="MAV_WINCH_STATUS_HEALTHY">
+        <description>Winch is healthy</description>
+      </entry>
+      <entry value="2" name="MAV_WINCH_STATUS_FULLY_RETRACTED">
+        <description>Winch thread is fully retracted</description>
+      </entry>
+      <entry value="4" name="MAV_WINCH_STATUS_MOVING">
+        <description>Winch motor is moving</description>
+      </entry>
+      <entry value="8" name="MAV_WINCH_STATUS_CLUTCH_ENGAGED">
+        <description>Winch clutch is engaged allowing motor to move freely</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="0" name="HEARTBEAT">
@@ -7093,6 +7108,19 @@
       <field type="uint64_t" name="time_usec" units="us">Timestamp (synced to UNIX time or since system boot).</field>
       <field type="uint8_t" name="count">Number of wheels reported.</field>
       <field type="double[16]" name="distance" units="m">Distance reported by individual wheel encoders. Forward rotations increase values, reverse rotations decrease them. Not all wheels will necessarily have wheel encoders; the mapping of encoders to wheel positions must be agreed/understood by the endpoints.</field>
+    </message>
+    <message id="9005" name="WINCH_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change, and should NOT be used in stable production environments -->
+      <description>Winch status.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (synced to UNIX time or since system boot).</field>
+      <field type="float" name="line_length" units="m">Length of line released. NaN if unknown</field>
+      <field type="float" name="speed" units="m/s">Speed line is being released or retracted. Positive values if being released, negative values if being retracted, NaN if unknown</field>
+      <field type="float" name="tension" units="kg">Tension on the line. NaN if unknown</field>
+      <field type="float" name="voltage" units="V">Voltage of the battery supplying the winch. NaN if unknown</field>
+      <field type="float" name="current" units="A">Current draw from the winch. NaN if unknown</field>
+      <field type="int16_t" name="temperature" units="degC">Temperature of the motor. INT16_MAX if unknown</field>
+      <field type="uint32_t" name="status" display="bitmask" enum="MAV_WINCH_STATUS_FLAG">Status flags</field>
     </message>
     <message id="12900" name="OPEN_DRONE_ID_BASIC_ID">
       <wip/>


message_definitions/v1.0/common.xml
```diff
@@ -4720,6 +4720,21 @@
         <description>Unit is sometimes working, sometimes not.</description>
       </entry>
     </enum>
+    <enum name="MAV_WINCH_STATUS_FLAG" bitmask="true">
+      <description>Winch status flags used in WINCH_STATUS</description>
+      <entry value="1" name="MAV_WINCH_STATUS_HEALTHY">
+        <description>Winch is healthy</description>
+      </entry>
+      <entry value="2" name="MAV_WINCH_STATUS_THREAD_END">
```
hamishwillee, line 4728: 1. Would MAV_WINCH_STATUS_RETRACTED be better (ie. are all winches mechanically the same such that thread end makes sense for all winches). More generally, retracted means something to me, while THREAD end requires explanation.
2. Is there need for a complementary flag indicating fully extended?

message_definitions/v1.0/common.xml
```diff
@@ -7062,6 +7077,17 @@
       <field type="uint8_t" name="count">Number of wheels reported.</field>
       <field type="double[16]" name="distance" units="m">Distance reported by individual wheel encoders. Forward rotations increase values, reverse rotations decrease them. Not all wheels will necessarily have wheel encoders; the mapping of encoders to wheel positions must be agreed/understood by the endpoints.</field>
     </message>
+    <message id="9005" name="WINCH_STATUS">
+      <description>Winch status.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (synced to UNIX time or since system boot).</field>
+      <field type="float" name="line_length" units="m">Length of line released. NaN if unknown</field>
+      <field type="int16_t" name="speed" units="d%">Speed line is being released as a percentage. Negative values if retracting. INT16_MAX if unknown</field>
```
hamishwillee, line 7084: Probably "Speed line is being released/retracted" (units state that this is decipercentage).

message_definitions/v1.0/common.xml
```diff
@@ -7062,6 +7077,17 @@
...
+      <field type="uint16_t" name="tension" units="g">Tension on the line in grams. UINT16_MAX if unknown</field>
+      <field type="float" name="voltage" units="V">Voltage of the battery supplying the winch. NaN if unknown</field>
```
hamishwillee, line 7086: Why do we need the current and voltage? I'm not arguing it, just interested.

hamishwillee:
Just a few questions indicating my ignorance. Shouldn't we also have a MAV_CMD to control retract/extend?

message_definitions/v1.0/common.xml
rmackay9: Sure, I can rename it to MAV_WINCH_STATUS_FULLY_RETRACTED.  On the Daiwa winch this status is directly connected to a lever/switch that is depressed by a ball bearing at the end of the line.  There's no way that the winch can detect if it's reached the other end (i..e fully extended) so no point in adding this complementary flag for now at least.

message_definitions/v1.0/common.xml
rmackay9: Sure, I'll change this to be in m/s.  The winch actually doesn't know how fast the line is being released in m/s but we can multiply the percentage but a fixed constant and it will be close.

message_definitions/v1.0/common.xml
rmackay9: The winch provides the battery voltage and the amount of current it is drawing to retract the payload.  I suppose it might be useful for the pilot to know the current so they can see if the winch is pulling too hard unnecessarily.  I.e. you could imagine the payload is fully retracted but the winch motor is still pulling very hard to no avail.

rmackay9:
@hamishwillee, there is actually already a MAV_CMD_DO_WINCH in ardupilotmega.xml that I added many years ago.  We could move it to common.xml but I'm not sure anybody is interested at this point.

rmackay9:
@hamishwillee, thanks very much for the review.  I've updated the message as discussed.

hamishwillee:
Sounds good. 

My only concern is future proofing. I'm interested not just in how much I have spooled out, but how much I have in total to play with. Worth adding a field for that now? 

That case you were talking about where it has stalled and the pilot has to guess based on voltage/current might be better done as a status bit too - that the controller can supply if it can. I guess the worry being that if you don't get that field you think everything is fine?

@auturgy @julianoes Anything to add to this?

message_definitions/v1.0/common.xml
```diff
@@ -7062,6 +7077,17 @@
       <field type="uint8_t" name="count">Number of wheels reported.</field>
       <field type="double[16]" name="distance" units="m">Distance reported by individual wheel encoders. Forward rotations increase values, reverse rotations decrease them. Not all wheels will necessarily have wheel encoders; the mapping of encoders to wheel positions must be agreed/understood by the endpoints.</field>
     </message>
+    <message id="9005" name="WINCH_STATUS">
+      <description>Winch status.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (synced to UNIX time or since system boot).</field>
+      <field type="float" name="line_length" units="m">Length of line released. NaN if unknown</field>
+      <field type="float" name="speed" units="m/s">Speed line is being released or retracted. Positive values if being released, negative values if being retracted, NaN if unknown</field>
+      <field type="uint16_t" name="tension" units="g">Tension on the line in grams. UINT16_MAX if unknown</field>
```
julianoes, line 7085: What about?

```suggestion
      <field type="float" name="tension" units="kg">Tension on the line in grams. NAN if unknown</field>
```

Just because it would be consistent with the other flags.

rmackay9: @julianoes, txs for the review.  I generally think your suggestion makes sense, I didn't do it this way because from what I could see, "kg" is not in the list of units ([see here](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd)) while "g" (for grams) is.  I could add kg but that would also require updating pymavlink.

rmackay9:
@hamishwillee, I'm not personally a fan of adding stuff that might be used in the future because that very often leads to cruft that nobody uses.  I think it's best to keep it simple and then add extra flags later if the day comes that we can support them.

message_definitions/v1.0/common.xml
julianoes: Oh, I see. That's funny since `kg` is the SI one. Anyway, your call. I'm ok with this otherwise.

rmackay9:
@hamishwillee, I'm happy to rebase this before we merge.  I might do the change that Julien has suggested as well.

As a side note, I think moving to a model where all PRs are rebased before merging would be best.  That avoids the ugly "Merge branch 'master'"..." commits.

hamishwillee:
@rmackay9 We squash and merge anyway - doesn't that avoid the ugly commit? If not, yes please, do rebase.

> I'm not personally a fan of adding stuff that might be used in the future because that very often leads to cruft that nobody uses. I think it's best to keep it simple and then add extra flags later if the day comes that we can support them.

As an ideology I agree 100% with your statement, but this doesn't feel like cruft. To me it feels like the winch is flawed by not providing the total length. Still, Julian thumbs-ed up your comment, so I guess he is OK with the design.

FWIW I guess mostly we work in vehicles where grams make more sense than kg, so I am ambivalent as the data size is the same.

hamishwillee:
@rmackay9 FYI I didn't merge this because you said you might edit it. But one change I would make is to mark it `wip` - that way you can change it without much discussion until you have an implementation.

rmackay9:
@hamishwillee, txs for not merging.  I have an driver implementation that is undergoing review on the AP side.  i suspect that will be complete in a week or so.

.. re the hardware flaw.. who knows.. that's out of my control.  We work with what we have.

rmackay9:
I've made both the suggested changes:

- Added WIP label.  I agree this is a good idea at least until the driver gets merged into AP master
- Switched the tension to be in kg.   This will mean that travis will probably fail because this pymavlink PR https://github.com/ArduPilot/pymavlink/pull/437 needs to be merged first.  It'll take a little longer but still seems like the right choices.

Thanks again for the feedback.

hamishwillee:
Awesome. I'll merge after the XSD change goes in. Can you ping me on that in case I miss it - as I will have to update the submodule here first too.

rmackay9:
@hamishwillee, great thanks.  The pymavlink change has gone in now so I guess if mavlink's pymavlink module is updated and this PR is debased it will pass Travis.  Txs!

rmackay9:
@hamishwillee, I've rebased this on master and I think it's ready to merge once it re-passes travis.  If you're happy can you merge this for me?

hamishwillee:
@rmackay9 Done. I do think we will need to modify this in future for other winches, but hopefully in a non-breaking way.

rmackay9:
Great, thanks!

PR #1433, opened 2020-07-29
Add NAV_VTOL_LAND options for the pattern

WickedShell:
Add options to NAV_VTOL_LAND to allow more control over how the approach is managed. Relates to ArduPilot/ardupilot#14874


hamishwillee:
@sfuhrer @RomanBapst This adds specificity to the VTOL land command on intended landing behaviour. Can you please comment - ie are the options reasonable, should there be more? etc.

WickedShell:
@sfuhrer you're proposed options were what I was attempting to encompass here actually :) The only tweak I have is that on `NAV_VTOL_LAND_OPTIONS_HOVER_DESCENT` the altitude field is describing this approach altitude that the vehicle should be descending/climbing to enroute to the transition point. I can reuse your labels if you prefer, I'm not to fussed as long as we can start flagging options into the landing approach.

hamishwillee:
@WickedShell @sfuhrer Do you still have an interest in this PR? If so did you reach agreement on what the changes are?  Please rebase and we can discuss getting this in?

If not, then suggest we close this.

sfuhrer:
> @WickedShell @sfuhrer Do you still have an interest in this PR? If so did you reach agreement on what the changes are? Please rebase and we can discuss getting this in?
> 
> If not, then suggest we close this.

Okay to close it in my eyes. We're currently working on a slightly different approach for VTOL landings anyway.

WickedShell:
Yes I still want it and need it. (Anything else I've thought of will just
be a proliferation of more commands which won't help)

DonLakeFlyer:
I don't see how you can set up a spiral approach with just this one command? Don't you need to spiral and then head into the wind to transition and then when transition completes multi-rotor down. That requires a landing pattern of multiple commands to setup against wind correctly doesn't it?

WickedShell:
> I don't see how you can set up a spiral approach with just this one command? Don't you need to spiral down and then head into the wind to transition which then takes a decent amount of distance and then when transition completes multi-rotor down. That requires a landing pattern of multiple commands to setup against wind correctly doesn't it?

I spiral down over the landing point, to the approach altitude, then select the into the wind direction and detransition before descending. This uses a single waypoint to run the approach at the moment, but I've had to do it as a vehicle level switch, which inhibits straight in landings.
land5

DonLakeFlyer:
> then select the into the wind direction

How do you do that? There is no value for that in the command.

WickedShell:
The vehicle is using it's own estimate of the wind heading, and selecting the path itself.

DonLakeFlyer:
> The vehicle is using it's own estimate of the wind heading, and selecting the path itself.

Ah, interesting. Understood. Wonder if the option should have more text explaining that somehow to be clearer?

hamishwillee:
@sfuhrer "We're currently working on a slightly different approach for VTOL landings anyway.". Can you explain what you're proposing and whether this should be a variation or extension as well?

@WickedShell 
1. Thanks for the spiral approach diagram. So the approach altitude is param3. Where are the spiral radius and transition altitude defined? Is the yaw param respected after transition? I see that we specify ground altitude - just FMI why is that needed?
2. The difference between this and the other option is defined by @sfuhrer [here](https://github.com/mavlink/mavlink/pull/1433#pullrequestreview-459812745). Essentially both fly to landing point at approach altitude - in one case you spiral down to transition altitude in FW and then transition, in the other you just transition and land in MC. Is that correct?

image

I can probably help reword this once you guys can confirm the above.

hamishwillee:
@WickedShell @sfuhrer I have proposed a rewording based on @sfuhrer text. I've tried to make it pretty simple either "Land in mc mode all the way" or "fixed wing until close to ground, then land in MC mode". 

Questions
1. I would still like to know what PX4 approach is/how that fits in
2. What if the FW landing point is sent (e.g. as a command) when you're in MC mode? [if in MC and got MC mode I'd assume you'd just land]
3. If this is sent without a position will it land where it is? This is theoretical, because according to MAVLink it must have a position. But that could just be a "glitch".
4. Does this do what people want?

sfuhrer:
> @sfuhrer "We're currently working on a slightly different approach for VTOL landings anyway.". Can you explain what you're proposing and whether this should be a variation or extension as well?

What we are currently experimenting with is a VTOL land command that has as additional arguments (2 min needed) a safe area (specified by at least two values, e.g. radius and cone angle). The vehicle would than automatically chose the optimal approach direction in this cone, based on current onboard wind estimate. Ideally, the command would also expose an transition (aka approach) altitude. 
It's gonna be very hard to combine all these (I guess all very valid) cases of 
- descend completely in hover
- descend in a spiral around home
- descend in defined safe area and then transition towards home

If we want to combine all of them into one command then I guess we would need like a "mode" param in the beginning (either 1,2 or 3 from above), and then have adaptive other params (e.g. for the option 3 we could sacrifice the Heading, as it should be defined by the transition direction and/or wind). Maybe that's also a no-go, and in that case I don't see any real other option than to switch to a new command for it, and your proposal @hamishwillee would be okay.
@RomanBapst maybe you could also comment here.

message_definitions/v1.0/common.xml
```diff
@@ -4720,6 +4720,19 @@
...
+    <enum name="NAV_VTOL_LAND_OPTIONS">
+      <entry value="0" name="NAV_VTOL_LAND_OPTIONS_DEFAULT">
+        <description>Default autopilot landing behaviour.</description>
+      </entry>
+      <entry value="1" name="NAV_VTOL_LAND_OPTIONS_FW_DESCENT">
+        <description>Descend in fixed wing mode, transitioning to multicopter mode for vertical landing when close to the ground.
+          The fixed wing descent pattern is at the discretion of the vehicle (e.g. transition altitude, loiter direction, radius, and speed, etc.).        
+        </description>
+      </entry>
+      <entry value="2" name="NAV_VTOL_LAND_OPTIONS_HOVER_DESCENT">
```
WickedShell, line 4732: I can't tell if this is different then what I described or not. The name makes me think it's an immediate transition to MC flight modes then land, but the description implies you fly as a FW then go MC just at the end (which is the option I was trying to cover here)

WickedShell (commented):
> 2. What if the FW landing point is sent (e.g. as a command) when you're in MC mode? [if in MC and got MC mode I'd assume you'd just land]

ArduPilot doesn't really map modes out in that manner. I'd sorta expect you to go back to FW mode and fly the approach, but that's just my gut reaction for how to handle that.

> 3. If this is sent without a position will it land where it is? This is theoretical, because according to MAVLink it must have a position. But that could just be a "glitch".

I think it just requires a position, and you nack it otherwise.... Arguably something like spiral could use the current position when it starts the command, but I think that's just a can of worms, and probably bad planning, so I'd prefer to just nack it.

> What we are currently experimenting with is a VTOL land command that has as additional arguments (2 min needed) a safe area (specified by at least two values, e.g. radius and cone angle).

I'd be happy for more options. I actually messed with adding this as well in ArduPilot (somewhere or other I have a working implementation, but it quickly became "I want to stack multiple safe zones together" and users were bad at giving me a reasonably into the wind option, so I ended up just swapping to the spiral stuff, as at the time it was imperative to do transitions into the wind. I think it's worth noting that a common use case is your safe zone is actually mirrored on the opposite side (think about a runway for example), and you'd like the plane to be able to approach either side.

message_definitions/v1.0/common.xml
hamishwillee: Your assumption from the name is correct - the vehicle transitions to MC and does a hover descent all the way to the ground as soon as it reaches the landing position/target. I don't see the description as ambiguous - it says "reaching the landing position" not "altititude". Let me change that to co-ordinates. 

The other option is the one you've proposed. Essentially you fly to the landing co-ordinates and fly in fixed wing to some other altitude before descending. I haven't specified spiral or the landing altitude or whatever - that's up to your vehicle.

hamishwillee:
@sfuhrer Thanks for the explanation. We can't really mess with the meaning of existing fields in the command. While it is desirable to support this in the same message as the others, I think that will be quite hard. When you have a prototype implementation we can revisit a new command.

@WickedShell I have updated the hover descent description - is this clear/are you happy with text now.

>>  If this is sent without a position will it land where it is? This is theoretical, because according to MAVLink it must have a position. But that could just be a "glitch".

> I think it just requires a position, and you nack it otherwise.... Arguably something like spiral could use the current position when it starts the command, but I think that's just a can of worms, and probably bad planning, so I'd prefer to just nack it.

An MC that gets a land command without a position just lands at the current position. There isn't any can of worms - you're just saving something having to first fetch the current position and then set it in the message (by which time it might be a little "off").  It isn't really bad planning either - a user can decide to land outside a mission with this message right?

Anyway, just a thought.

message_definitions/v1.0/common.xml
WickedShell: My concern is we maybe just want an additional option right now of "immediately go into MC mode and do the entire approach + landing in MC mode" because otherwise I think there is some ambiguity possible. (It's also just the other logical landing type I can think of off the top of my head that people will probably want)

WickedShell:
> An MC that gets a land command without a position just lands at the current position. There isn't any can of worms - you're just saving something having to first fetch the current position and then set it in the message (by which time it might be a little "off"). It isn't really bad planning either - a user can decide to land outside a mission with this message right?

My only concern is if it's a FW approach then switch to MC what will happen is you detransition immediately potentially, then have to fly as a MC backwards to your spot where you started the transition, which could be far away. We can accept 0,0 to mean current position I guess, but it looks like a messy situation to me, but I might be being missing the case where it works great, and it's not a big deal to me either way.

sfuhrer:
>When you have a prototype implementation we can revisit a new command.

Alright, then I have nothing against getting this in like it is now (though I'm not convinced that we need the "Default autopilot landing behaviour" option).

message_definitions/v1.0/common.xml
hamishwillee: @WickedShell If I wanted to do that I would just have a MAV_CMD_DO_VTOL_TRANSITION at the preceding point and fly my approach in MC mode. I think that complicates this for no benefit.

hamishwillee:
> I'm not convinced that we need the "Default autopilot landing behaviour" option).

@sfuhrer The default landing behaviour value is OK with me. It allows the system to have other interesting options we don't know about.

hamishwillee:
OK, @sfuhrer is good with this. @WickedShell if you're OK too ping here and I will merge.

message_definitions/v1.0/common.xml
WickedShell: @hamishwillee I still think there's some benefit, but it's also something I think we can revisit it and add more options in the future if needed, and this definetly gets me what I need at the moment, so let's go with it.

hamishwillee:
Merging - this won't break anyone, can be extended with new pattern definitions as needed, and has agreement from two autopilots. Also, anyone who wanted to comment has had lots of opportunities.

message_definitions/v1.0/common.xml
hamishwillee: I've got to say I'd be in the "opposed" camp on such an approach. Mostly because I think of the mavlink messages as ideally "primitives". Others might disagree.

Anyway, it is in. Hopefully does what you need. Thanks for your patience (also thanks @sfuhrer for your feedback).

PR #1434, opened 2020-07-29
add missing things for param v3 protocol

DanielePettenuzzo:
This pr fixes the following in the new param v3 protocol:

- MAV_PARAM_V3_TYPE (same as the MAV_PARAM_EXT_TYPE)
- add PARAM_TRANSACTION_TRANSPORT_PARAM_V3 transport
- extend PARAM_REQUEST_LIST and PARAM_REQUEST_READ to be used also for the param v3 protocol instead of creating new messages

FYI @hamishwillee

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1395,7 +1395,7 @@
       <entry value="1" name="PARAM_TRANSACTION_TRANSPORT_PARAM_EXT">
         <description>Transaction over param_ext transport.</description>
       </entry>
-      <entry value="1" name="PARAM_TRANSACTION_TRANSPORT_PARAM_V3">
+      <entry value="2" name="PARAM_TRANSACTION_TRANSPORT_PARAM_V3">
         <description>Transaction over param_v3 transport.</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1395,6 +1395,9 @@
       <entry value="1" name="PARAM_TRANSACTION_TRANSPORT_PARAM_EXT">
         <description>Transaction over param_ext transport.</description>
       </entry>
+      <entry value="1" name="PARAM_TRANSACTION_TRANSPORT_PARAM_V3">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4818,11 +4858,15 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="char[16]" name="param_id">Onboard parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
       <field type="int16_t" name="param_index">Parameter index. Send -1 to use the param ID field as identifier (else the param id will be ignored)</field>
+      <extensions/>
+      <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Possible transport layers to set and get parameters via mavlink during a parameter transaction.</field>
```
hamishwillee, line 4862: @DanielePettenuzzo @julianoes The `PARAM_REQUEST_LIST.param_transport` and `PARAM_REQUEST_READ.param_transport` need to be discussed.

Firstly you don't need to think about transactions for reading, only for setting.

I assume that this is a way around having to declare a new PARAM_V3 read and list message? You're probably better off doing that - after all we already have PARAM_EXT_REQUEST_READ.

Julian might take a different view, but I think the symmetry is better.

DanielePettenuzzo:
@hamishwillee @julianoes Ok makes sense. I will make the change. In this case that we don't have any available message ids close to the other PARAM_V3 should I just take the first available at the bottom of the list?

julianoes:
@DanielePettenuzzo if you create a complete new set as V3, then please move it together so the message IDs make sense. You can just move the new V3 ones.

DanielePettenuzzo:
@julianoes ok got it. Thank you

julianoes:
@DanielePettenuzzo if you don't want to do all that work, then just rename V3 to _TRIMMED and call it a day :wink:.

PR #1435, opened 2020-07-29
common: message to present camera field of view

julianoes:
This message enables a ground station to present the field of view of the camera in the map view.

The alternative to this new message would be to add the required fields to CAMERA_CAPTURE_STATUS.

FYI @jkflying @mrivi @batinkov

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6437,7 +6437,7 @@
       <field type="int32_t" name="lon" units="degE7">Longitude where capture was taken</field>
       <field type="int32_t" name="alt" units="mm">Altitude (MSL) where image was taken</field>
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
-      <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 0, 0, 0, 0)</field>
+      <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="int32_t" name="image_index">Zero based index of this image (i.e. a new image will have index CAMERA_CAPTURE_STATUS.image count -1)</field>
       <field type="int8_t" name="capture_result">Boolean indicating success (1) or failure (0) while capturing this image.</field>
       <field type="char[205]" name="file_url">URL of image taken. Either local storage or http://foo.jpg if camera provides an HTTP interface.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6437,7 +6437,7 @@
       <field type="int32_t" name="lon" units="degE7">Longitude where capture was taken</field>
       <field type="int32_t" name="alt" units="mm">Altitude (MSL) where image was taken</field>
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
-      <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 0, 0, 0, 0)</field>
+      <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
```
hamishwillee, line 6440: Would anyone be depending on the previous version of this value?

hamishwillee:
@julianoes 

> The alternative to this new message would be to add the required fields to CAMERA_CAPTURE_STATUS.

So when does this get sent? Same time as CAMERA_CAPTURE_STATUS or on some different cycle and trigger? I.e if this is something you'll always want at the same time they might as well be in the same message. However if only some status messages might need these fields then doing as you have is the right thing. As you know :-)

My only other question is who supplies this information and where does it come from? For example, the lat_camera and friends are not likely to be known by the camera - which is OK if this is from the autopilot and it can assume its own position and use a param say to work out the camera offset. How about the image height?

message_definitions/v1.0/common.xml
julianoes: I don't think so. That was just wrong.

julianoes:
@hamishwillee I think the rate should not be tied to to CAMERA_CAPTURE_STATUS.

> My only other question is who supplies this information and where does it come from?

This is for cameras that do know all this information.

hamishwillee:
How many cameras are going to know their own altitude, latitude and longitude? I put it to you - none.

julianoes:
> How many cameras are going to know their own altitude, latitude and longitude? I put it to you - none.

hamishwillee:
Wow. I guess phones have cameras and GPS chips, so maybe not so unrealistic. Though maybe not to mm precision :-). Maybe the intent is just approximate - ie on a companion you could just take the vehicle position as camera position.

julianoes:
> ie  on a companion you could just take the vehicle position as camera position.

Exactly!

hamishwillee:
Is it then worth making a note of this, since the units are mm and a user might assume they need to mess around with offsets to vehicle GPS etc to get this kind of accuracy? Something like this:

> Information about the field of view of a camera. **Camera location is typically from onboard GPS (typically no need to compensate for offset between GPS and camera).**  Can be requested with a MAV_CMD_REQUEST_MESSAGE command.

Anyway, if you add `wip` I'm fine with this being merged (ie. just do it)

julianoes:
Added wip.

hamishwillee:
Cheers. Merged as WIP.

PR #1436, opened 2020-07-29
common: remove camera ID

julianoes:
As far as I know the concept of camera ID was abandonded and this is just a leftover that should be removed.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6432,7 +6432,7 @@
       <description>Information about a captured image. This is emitted every time a message is captured. It may be re-requested using MAV_CMD_REQUEST_MESSAGE, using param2 to indicate the sequence number for the missing image.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint64_t" name="time_utc" units="us">Timestamp (time since UNIX epoch) in UTC. 0 for unknown.</field>
-      <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="camera_id">Deprecated/unused. Please use additional component IDs for multiple cameras.</field>
       <field type="int32_t" name="lat" units="degE7">Latitude where image was taken</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6432,7 +6432,7 @@
       <description>Information about a captured image. This is emitted every time a message is captured. It may be re-requested using MAV_CMD_REQUEST_MESSAGE, using param2 to indicate the sequence number for the missing image.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint64_t" name="time_utc" units="us">Timestamp (time since UNIX epoch) in UTC. 0 for unknown.</field>
-      <field type="uint8_t" name="camera_id">Camera ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="camera_id">Deprecated/unused. Please use additional component IDs for multiple cameras.</field>
```
hamishwillee, line 6435: @julianoes Is the theory here that the originating camera will be determine by the sender component id? What if the message is emitted from a camera manager component that has its own component ID.

In any case, I don't know whether or not this is true. @dogmaphobic is the owner of the API - Gus can you please comment.

julianoes: > What if the message is emitted from a camera manager component that has its own component ID.

I don't know. It could send MAV_TYPE to camera. That's a general problem with components, isn't it?

@dogmaphobic switched from camera ID to having multiple camera component IDs but stream IDs.
https://github.com/mavlink/mavlink/pull/747#discussion_r127045273

So the question is if this should be a stream ID or nothing. :thinking:

dogmaphobic:
You use the compID to identify the camera. Early on, the protocol used to define `camera_id` and it was removed in favor of compID. This `camera_id` field is a left over and it is not used anywhere.

>What if the message is emitted from a camera manager component that has its own component ID.

That would be wrong and it would confuse the ground station. Different processes, using their own instance of MAVLink should never send messages on behalf of others as this will mess up the counters within the message.

As a camera may have multiple video streams, `stream_id` is used to identify them. This is not relevant to this message.

julianoes:
Thanks @dogmaphobic.

PR #1439, opened 2020-07-31
common: add param_action to param transaction commit

DanielePettenuzzo:
This is a continuation for the parameter transaction protocol design. I added a field to the commit message to decide whether we want to commit or cancel the transaction. The reason for this is for bigger sets of parameters. In QGC we could open a transaction when the user opens a new view in the UI and set the parameters as the user is inserting them and then just send the commit when the user presses on the submit button. In this way we avoid sending all parameters at the end making the submit action slower. If we want to do this we should also be able to cancel a transaction in case a user changes some parameters but then goes to another view. 

FYI @dogmaphobic

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1396,6 +1396,15 @@
         <description>Transaction over param_ext transport.</description>
       </entry>
     </enum>
+    <enum name="PARAM_TRANSACTION_ACTION">
+      <description>Possible parameter transaction action during a commit.</description>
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
Why is this in a separate pull? This should be part of the other pull for transactions.

DonLakeFlyer:
Now I realize that other transaction change has been merged!

DanielePettenuzzo:
Yes it was merged as WIP so that we can iterate on it

PR #1441, opened 2020-08-03
Image tracking

dogmaphobic:
Moved the image tracking work that started within the gimbal api to the camera api while adding missing messages and commands.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6609,15 +6609,15 @@
       <field type="int32_t" name="lat" units="degE7">Latitude of tracked object</field>
       <field type="int32_t" name="lon" units="degE7">Longitude of tracked object</field>
       <field type="float" name="alt" units="m">Altitude of tracked object(AMSL, WGS84)</field>
-      <field type="float" name="h_acc" units="m">Horizontal accuracy, NAN if unknown</field>
-      <field type="float" name="v_acc" units="m">Vertical accuracy, NAN if unknown</field>
-      <field type="float" name="vel_n" units="m/s">North velocity of tracked object, NAN if unknown</field>
-      <field type="float" name="vel_e" units="m/s">East velocity of tracked object, NAN if unknown</field>
-      <field type="float" name="vel_d" units="m/s">Down velocity of tracked object, NAN if unknown</field>
-      <field type="float" name="vel_acc" units="m/s">Velocity accuracy, NAN if unknown</field>
-      <field type="float" name="dist" units="m">Distance between camera and tracked object, NAN if unknown</field>
-      <field type="float" name="hdg" units="rad">Heading in radians, NAN if unknown, in NED</field>
-      <field type="float" name="hdg_acc" units="rad">Accuracy of heading, NAN if unknown, in NED</field>
+      <field type="float" name="h_acc" units="m">Horizontal accuracy. NAN if unknown</field>
+      <field type="float" name="v_acc" units="m">Vertical accuracy. NAN if unknown</field>
+      <field type="float" name="vel_n" units="m/s">North velocity of tracked object. NAN if unknown</field>
+      <field type="float" name="vel_e" units="m/s">East velocity of tracked object. NAN if unknown</field>
+      <field type="float" name="vel_d" units="m/s">Down velocity of tracked object. NAN if unknown</field>
+      <field type="float" name="vel_acc" units="m/s">Velocity accuracy. NAN if unknown</field>
+      <field type="float" name="dist" units="m">Distance between camera and tracked object. NAN if unknown</field>
+      <field type="float" name="hdg" units="rad">Heading in radians, in NED. NAN if unknown</field>
+      <field type="float" name="hdg_acc" units="rad">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
     <message id="280" name="GIMBAL_MANAGER_INFORMATION">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1100,12 +1100,6 @@
       <entry value="131072" name="GIMBAL_MANAGER_CAP_FLAGS_CAN_POINT_LOCATION_GLOBAL">
         <description>Gimbal manager supports to point to a global latitude, longitude, altitude position.</description>
       </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2409,6 +2385,28 @@
         <param index="2" label="Reset" minValue="-1" maxValue="1" increment="1">1 to reset the trigger sequence, -1 or 0 to ignore</param>
         <param index="3" label="Pause" minValue="-1" maxValue="1" increment="2">1 to pause triggering, but without switching the camera off or retracting it. -1 to ignore</param>
       </entry>
+      <entry value="2004" name="MAV_CMD_CAMERA_TRACK_POINT" hasLocation="false" isDestination="false">
```
hamishwillee, line 2388: Will it track forever if no stop tracking is received?

message_definitions/v1.0/common.xml
```diff
@@ -3890,6 +3897,45 @@
         <description>Stream is h.264 on MPEG TS (URI gives the port number)</description>
       </entry>
     </enum>
+    <enum name="CAMERA_TRACKING_STATUS_FLAGS">
+      <description>Camera tracking status flags</description>
+      <entry value="0" name="CAMERA_TRACKING_STATUS_FLAGS_IDLE">
```
hamishwillee, line 3902: Is it ever possible for (all) these enums that the status will not be known?

message_definitions/v1.0/common.xml
dogmaphobic: Yes, it will stay in tracking mode until told otherwise. It may lose track of the target (which would be indicated in the status), but it will keep tracking (or attempting).

message_definitions/v1.0/common.xml
dogmaphobic: Not really. It is either not tracking (IDLE) or tracking (ACTIVE or ERROR).

olliw42:
tracking is not a gimbal feature, but likewise it is not a camera feature
on a chat with Julian last Fr I called it a "super mode". Maybe connect with @julianoes for latest insight. 
:)

julianoes:
@olliw42 I can see how an overall tracking mode is a "super mode", however, I think it makes sense that an actual tracking command goes to a camera which then tries to track. This can be with or without gimbal or autopilot, whatever the use case requires.

dogmaphobic:
> tracking is not a gimbal feature, but likewise it is not a camera feature

I would agree with that. My worry is that by attaching it to either one of these APIs, the compID may become an issue if the code handling the tracking is in a different process. It won't work. That is, if it's part of the camera API, it would inherit the camera compID. If it running in a different process, it would have its own instance of mavlink and that would mess up the messages (counters, etc.) If it runs on a different process, it needs its own compID.

With that said, every tracking device I've seen is a "camera" (camera/gimbal combination), exposing one single SDK.

Jaeyoung-Lim:
@hamishwillee Any chance that we can get this in?

LorenzMeier:
There is an unrelated CI failure, so I'm still merging this.

message_definitions/v1.0/common.xml
```diff
@@ -6555,6 +6601,39 @@
...
       <field type="float" name="vfov" units="deg">Vertical field of view (NaN if unknown).</field>
     </message>
+    <message id="275" name="CAMERA_TRACKING_IMAGE_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Camera tracking status, sent while in active tracking. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval.</description>
+      <field type="uint8_t" name="tracking_status" enum="CAMERA_TRACKING_STATUS_FLAGS">Current tracking status</field>
+      <field type="uint8_t" name="tracking_mode" enum="CAMERA_TRACKING_MODE">Current tracking mode</field>
+      <field type="uint8_t" name="target_data" enum="CAMERA_TRACKING_TARGET_DATA">Defines location of target data</field>
+      <field type="float" name="point_x">Current tracked point x value if CAMERA_TRACKING_POINT (normalized 0..1, 0 is left, 1 is right), NAN if unknown</field>
```
amilcarlucas, line 6611: Why no minValue="0" maxValue="1" attributes here ?

julianoes: @amilcarlucas thanks. I guess that was forgotten. Would be great if you could follow up @dogmaphobic.

dogmaphobic: This is a read only message (status). There won't ever be an UI to edit these fields.

amilcarlucas: The point is not to create a GUI the point is to document the data for multiple purposes. GUI is only one of the possible purposes

message_definitions/v1.0/common.xml
```diff
@@ -2409,6 +2385,28 @@
...
       </entry>
+      <entry value="2004" name="MAV_CMD_CAMERA_TRACK_POINT" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>If the camera supports point visual tracking (CAMERA_CAP_FLAGS_HAS_TRACKING_POINT is set), this command allows to initiate the tracking.</description>
+        <param index="1" label="Point x" minValue="0" maxValue="1">Point to track x value (normalized 0..1, 0 is left, 1 is right).</param>
+        <param index="2" label="Point y" minValue="0" maxValue="1">Point to track y value (normalized 0..1, 0 is top, 1 is bottom).</param>
+        <param index="3" label="Radius" minValue="0" maxValue="1">Point radius (normalized 0..1, 0 is image left, 1 is image right).</param>
+      </entry>
+      <entry value="2005" name="MAV_CMD_CAMERA_TRACK_RECTANGLE" hasLocation="false" isDestination="false">
```
Jaeyoung-Lim, line 2396: Is there a reason why `MAV_CMD_CAMERA_TRACK_RECTANGLE` and `MAV_CMD_CAMERA_TRACK_POINT` needs to be separate messsages?

dogmaphobic: > Is there a reason why `MAV_CMD_CAMERA_TRACK_RECTANGLE` and `MAV_CMD_CAMERA_TRACK_POINT` needs to be separate messsages? For me it looks like it is sending quite similar information

Come to think of it, no. There could be a flag indicating what type.

Issue #1344, opened 2020-04-14
Component Id for a companion computer


olliw42:
I'd like to make my companion computer (a RPi) to be conform with MAVLink, and hence want to give it a component ID. Unfortunately, it's not clear to me what it should be, and it in fact seems to me a suitable id is missing.

The issue I face is that a companion CAN be of some of the available ids, but that it also may not be, and more over might fit MORE THAN ONE of the available ids.

For instance:

MAV_COMP_ID_LOG: it may log or may not log
MAV_COMP_ID_MISSIONPLANNER: it may do a mission plan or may not
MAV_COMP_ID_PATHPLANNER: it may do path planning or may not
MAV_COMP_ID_OBSTACLE_AVOIDANCE: it may do avoidance or may not
MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY: it may do odometry or may not

so the closest seems to be
MAV_COMP_ID_PERIPHERAL
but here to: it may provide parameters or it may not.

I thus suggest to add a new component id

MAV_COMP_ID_ONBOARD_COMPANION

It's id might be e.g. 161, 176, 199, 194, ...

Thanks
Olli

hamishwillee:
Hi @olliw42 

Not sure, even after some discussion.

@baumanta You worked on obstacle avoidance. Can you confirm what MAV_TYPE and MAV_COMPONENT you use for a companion computer when it has multiple ROS nodes - ie it also supports VIO?

@auturgy Says that his understanding is that the component id belongs to the process. So if you have a companion that does VIO and Obstacle avoidance, those nodes should emit their own heartbeats with those ids. But would be great if you can confirm.

Not sure about what you should use if there are choices. A MAV_COMP_ID_OBSTACLE_AVOIDANCE **is a** path planner. It probably won't be a MAV_COMP_ID_MISSIONPLANNER though - that is intended for something that sends out complete plans like a GCS. 

A further inconsistency is that we don't expect an autopilot to necessarily emit a separate HEARTBEAT if it implements the behaviour of a component. Another complication is when to use MAV_TYPES vs components. 

Now I've added to the confusion, perhaps we can converge on an answer based on "what is done now".

baumanta:
[...]

PR #1443, opened 2020-08-05
Improve param transactions

DanielePettenuzzo:
This pr solves 2 issues with parameter transactions and also introduces a new generic component ID for a mission computer (companion computer). I saw that there was already some discussion about this component id here: https://github.com/mavlink/mavlink/issues/1344. The component IDs we currently have for the mission computer are very specific. In my case I have a process on the mission computer that handles various types of mavlink parameters and interacts with a few other processes that are not exposed via mavlink. This is the reason why I would like to have a generic comp id and also reserve 3 more ids for other mission computer related components in the future.

Instead, regarding the parameter transaction protocol I extended with two fields:
- transaction_id: this was a proposal from @DonLakeFlyer since on the vehicle side we will need to keep some kind of transaction history. This is because if the ack to a transaction_commit is dropped then the next time qgc will retry the transaction_commit the transaction will already be closed (on the vehicle side) and we would reply with a fail. With this transaction id we can look up also through the closed transactions history and respond with the right state.
- param_group: in our process on the vehicle we have multiple groups of parameters (grouped by functionalities). We made a clear interface to be able to easily add new sets of parameters that could be used by other processes in the future. The architecture is already in place that multiple transactions could happen at once as long as they are from different groups (to avoid conflicts). To keep the protocol simple and easy to understand I think it would make sense to have this new field so that we already know which group a transaction will be setting (even before the first param set). Otherwise when we receive the transaction_start we need a lot of additional logic for this that would be removed with just one new field in the message. For now I made it 32 bits but it could also be 16. We could also make this field optional (eg. unused if set to 0).

@dogmaphobic can you please review this just to be sure that this matches what we discussed earlier today?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -709,6 +709,10 @@
       <entry value="190" name="MAV_COMP_ID_MISSIONPLANNER">
         <description>Component that can generate/supply a mission flight plan (e.g. GCS or developer API).</description>
       </entry>
+      <entry value="191" name="MAV_COMP_ID_MISSION_COMPUTER">
+        <description>Component that lives on the mission computer (companion computer) and has some generic functionalities, such as settings system parameters and monitoring the status of some processes that don't directly speak mavlink and so on.</description>
+      </entry>
+      <!-- Component ids from 192-194 are reserved for any future mavlink components related to the mission computer. -->
       <entry value="195" name="MAV_COMP_ID_PATHPLANNER">
         <description>Component that finds an optimal path between points based on a certain constraint (e.g. minimum snap, shortest path, cost, etc.).</description>
       </entry>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6830,7 +6830,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Possible transport layers to set and get parameters via mavlink during a parameter transaction.</field>
-      <field type="uint32_t" name="transaction_id">Identifier for a specific transaction.</field>
+      <field type="uint16_t" name="transaction_id">Identifier for a specific transaction.</field>
       <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="329" name="PARAM_COMMIT_TRANSACTION">
@@ -6840,7 +6840,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="param_action" enum="PARAM_TRANSACTION_ACTION">Commit or cancel an ongoing transaction.</field>
-      <field type="uint32_t" name="transaction_id">Identifier for a specific transaction.</field>
+      <field type="uint16_t" name="transaction_id">Identifier for a specific transaction.</field>
       <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="330" name="OBSTACLE_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -709,8 +709,8 @@
       <entry value="190" name="MAV_COMP_ID_MISSIONPLANNER">
         <description>Component that can generate/supply a mission flight plan (e.g. GCS or developer API).</description>
       </entry>
-      <entry value="191" name="MAV_COMP_ID_MISSION_COMPUTER">
-        <description>Component that lives on the mission computer (companion computer) and has some generic functionalities, such as settings system parameters and monitoring the status of some processes that don't directly speak mavlink and so on.</description>
+      <entry value="191" name="MAV_COMP_ID_ONBOARD_COMPUTER">
+        <description>Component that lives on the onboard computer (companion computer) and has some generic functionalities, such as settings system parameters and monitoring the status of some processes that don't directly speak mavlink and so on.</description>
       </entry>
       <!-- Component ids from 192-194 are reserved for any future mavlink components related to the mission computer. -->
       <entry value="195" name="MAV_COMP_ID_PATHPLANNER">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6826,6 +6826,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Possible transport layers to set and get parameters via mavlink during a parameter transaction.</field>
+      <field type="uint32_t" name="transaction_id">Identifier for a specific transaction.</field>
       <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="329" name="PARAM_COMMIT_TRANSACTION">
@@ -6835,6 +6836,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="param_action" enum="PARAM_TRANSACTION_ACTION">Commit or cancel an ongoing transaction.</field>
+      <field type="uint32_t" name="transaction_id">Identifier for a specific transaction.</field>
       <field type="uint8_t" name="response" enum="PARAM_TRANSACTION_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
     <message id="330" name="OBSTACLE_DISTANCE">


DonLakeFlyer:
"param_group" I would disagree with this. The gcs should be as stupid as possible to firmware internals. This also limits transactions to these "internal groups". Transactions can be useful for more than that. For example let's say you are loading parameters from a saved file. I think it would be nice if that were treated as a transaction. Either they all get changed to what is in the file or none of them get changed. I'm sure there are other examples of where transactions could be handy.

DonLakeFlyer:
Also why is `transaction_id` not some sort of int?

DanielePettenuzzo:
@DonLakeFlyer @dogmaphobic For the transaction_id I can change it to int. I was just wondering if it will just be a number or an alphanumeric string. If a number is good enough then I can change it to uint32. Regarding the param_group I understand your point. My proposal was to have this as an optional field just for specific applications where qgc has predefined transactions such as setting wifi credentials then it will know exactly the parameter names before hand since they are tightly coupled to the UI. With this one simple number we could already match the transaction to a set of parameters and avoid a lot of code in trying to understand what is the action has to be taken on the vehicle side. We could make it default equal to zero that means it's disabled and then just for the "predefined" transactions we could use it. Do you think this would be acceptable?

DonLakeFlyer:
If a gcs specifies a magic group id for a transaction then the firmware is going to need to verify that only parameters which are part of that magic group id are specified in the commit. Adding a parameter to that transaction which is outside the group is an error. In order to do that you will need to be able to quickly go from parameter name to magic group id. If you can do that then there is no need specify the group id in the protocol. If you can go from parameter name to group then you  know what action(s) the firmware needs to take when the commit happens. 

Also the protocol is either generic of it isn't. It seems like you are creating something which is not generic and specific to a use case. You can't assume a gcs is going to "do the right thing".

For example this should be completely supported:
* There is a magic group in the firmware called "foo" which contains parameters "foo1" and "foo2"
* There is also a magic group in the firmware called "bar" which contains parameters "bar1" and "bar2"
* There is also a random parameter call "baz" which is not part of any group
* I should be able to send this group of params in a transaction and it should work: foo1, foo2, bar2, baz
* The firmware should be smart enough to understand the the "foo" special commit thingy needs to run after the commit
* The firmware should be smart enough to understand that the "bar" special commit thingy code needs to run after the commit
* Things should still work fine even though every param in the "bar" group is not part of the transaction and just use old values for the missing params
* It's perfectly fine that "baz" was included in the transaction 
* If the user comes in through the parameter editor and changes "foo1" the firmware should still know to run the special commit thingy code on the "foo" group based on new foo1 and old foo2

DanielePettenuzzo:
Ok I understand your point. Thank you for the detailed explanation. I will push two changes so we can wrap this up:
- change transaction_id to uint32
- remove param_group

message_definitions/v1.0/common.xml
```diff
@@ -709,6 +709,10 @@
       <entry value="190" name="MAV_COMP_ID_MISSIONPLANNER">
         <description>Component that can generate/supply a mission flight plan (e.g. GCS or developer API).</description>
       </entry>
+      <entry value="191" name="MAV_COMP_ID_MISSION_COMPUTER">
```
hamishwillee, line 712: Is this so that QGC can send the transaction messages to the mission computer, or so that the autopilot (for example) knows that the messages are from the mission computer rather than say QGC? Or some other reason?

I know Lorenz likes the term mission computer but no one outside Auterion uses it - so if it were me I'd probably call it `MAV_COMP_ID_ONBOARD_CONTROLLER` to match with the `MAV_TYPE_ONBOARD_CONTROLLER`.


--- aside ---

Mission Computer id has come up a few times. This is a bit of a can of worms that should have been resolved, but I don't feel as - see https://github.com/mavlink/mavlink/issues/1344

For example, it isn't very clear if a companion should emit both this and MAV_COMP_ID_PATHPLANNER, and if it does, what the MAV_TYPE in the heartbeat should be - since that would be interpreted as a system with two companion computers.

auturgy:
Not critical but “onboard computer” would be better than “mission computer”. I know you’re shooting for aviation terminology, but mission computers, payload computers and flight computers are all different things, and in the drone space not clearly differentiated. I think it’s better to just use a generic descriptor than mis-use a specific one.

hamishwillee:
@DanielePettenuzzo I have no problem with the new transaction id. I would change the name of the mission computer id. I'm happy to merge modulo that as this is WIP. @julianoes as I might be asleep by the time there are responses, feel free to merge if you think it is fine.

@DonLakeFlyer Re groups, FWIW only
- the use case that Daniele is first addressing is using the parameters to configure the LTE or WIFI params. It would presumably be much easier for a flight stack to know this intent when it gets a transaction rather than wait until it gets all the parameters and then work out the intent - particularly if we start to use this for lots of different types of settings. So even though this is a magic number and might be ignored, and it doesn't help the GCS with any checking, it might still be ease flight stack-side processing to include an "intent field". I'm not a developer though.
- There are good reasons to go with params for this and I appreciate that this is generic GCS code. But it sux that every single GCS will have to reinvent the wheel here to work with PX4, and again for other flight stacks. 
- With params they will always be device specific, but we can perhaps later use the approach of camera definition XML to create a UI so that a GCS does not need to know anything about the underlying system - e.g. the GCS would need to know the common id below, but nothing else to know that it is setting LTE:
```
<setting id="common id used by systems for setting LTE" name="LTE Setting">
<transaction>
  <param name="whatever"></param>
  <param name="whatever2"></param>
</transaction>
</setting>
```


```

File: message_definitions/v1.0/common.xml (hunk 1 above)
Comment (DanielePettenuzzo) [reply message_definitions/v1.0/common.xml:712]: I am ok with MAV_COMP_ID_ONBOARD_CONTROLLER although maybe MAV_COMP_ID_ONBOARD_COMPUTER is still more generic. Regarding the MAV_COMP_ID_PATHPLANNER comp id my understanding that this is a more specific case. This is the component ID that something like mavros uses since it's actually performing some sort of path planning. In our system we run both mavros and this other process for parameters so I couldn't use this comp id twice. Moreover this new process that deals with parameters could be extended to many more things to get status and more control on the mission computer so it might do more that just one specific thing.

Instead with the MAV_TYPE I am very confused. My understanding what that this was used to identify what type of vehicle/hardware you are running on. For example quadrotor, hexarotor, vtol and ground station. For me it's a bit strange that then mission computer, gimbal, etc. are also mav types as they sound to me more like mav components. Again I'm not an expert with mavlink but my understanding is that the mav_type is more related to the vehicle and then inside we can have many mav components such as autopilot, gimbal, camera, pathplanner, etc. but all with the same mav_type. Does this make sense?

Maybe @julianoes has a better understanding here

**Comment by DanielePettenuzzo** [comment]
@hamishwillee for now I will rename the comp id and then we can merge this. I would say we could continue with the magic number discussion separately from this pr.

File: message_definitions/v1.0/common.xml (hunk 1 above)
Comment (DanielePettenuzzo) [reply message_definitions/v1.0/common.xml:712]: And regarding the mav type, this is something that is defined by the autopilot since it's there that you choose whether you are a vtol or a quadrotor. This would then mean that every mavlink process that is not running on the autopilot should wait for the autopilot's heartbeat (as we do for system and component id) and take also the mav_type from it since this will then contain the "airframe" type.

Comment (julianoes) [reply message_definitions/v1.0/common.xml:712]: I suggest MAV_COMP_ID_ONBOARD_CONTROLLER or MAV_COMP_ID_ONBOARD_COMPUTER. MISSION_COMPUTER is almost the same as MISSIONPLANNER, so I don't see the point in that.

File: message_definitions/v1.0/common.xml [hunk 2]
```diff
@@ -6822,6 +6826,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="param_transport" enum="PARAM_TRANSACTION_TRANSPORT">Possible transport layers to set and get parameters via mavlink during a parameter transaction.</field>
+      <field type="uint32_t" name="transaction_id">Identifier for a specific transaction.</field>
```
julianoes, line 6829: Do we need 2^32 transactions? Is this a random ID or counted from 0?

hamishwillee:
@DanielePettenuzzo 

> for now I will rename the comp id and then we can merge this. I would say we could continue with the magic number discussion separately from this pr.

Works for me. Push for Julian to merge. I'm nearly done for day. I will comment on the type/id in that thread. We shouldn't let it side track us.

message_definitions/v1.0/common.xml
DanielePettenuzzo: We have not defined this but I would prefer that during a session (from boot to shutdown) we cannot have two transactions with the same id and since on the vehicle we will keep some sort of transaction history it should probably be either 2^16 or 2^32. Should I reduce it to 16?

message_definitions/v1.0/common.xml
julianoes: > And regarding the mav type, this is something that is defined by the autopilot since it's there that you choose whether you are a vtol or a quadrotor.

That has changed, see: https://mavlink.io/en/messages/common.html#MAV_TYPE

DanielePettenuzzo: I will go with MAV_COMP_ID_ONBOARD_COMPUTER since I saw we already have a message called ONBOARD_COMPUTER_STATUS. This message was created by Tanja to report the companion computer's status to the ground station or to the autopilot.

julianoes: I think ONBOARD_COMPUTER is best, it's also what @auturgy suggests.

hamishwillee: @DanielePettenuzzo The thinking as it was on this is here: https://github.com/mavlink/mavlink/issues/1344#issuecomment-669639853

In summary, the type was what you though originally, but then got repurposed for what seemed like good reasons to Julian and Ollie, but which I never really groked. 

@julianoes and I are chatting about this, our thinking is leaning towards reverting the MAV_TYPE to mean what you think. This would be emitted by the autopilot and have the value of the type of vehicle (this is true now too). It would also be emitted by stand alone gimbals etc. However it would not "none" for any component that is a subcomponent.

Then we would use component as we have been historically too. 

A companion computer might emit both path planner and this id or neither, indicating that there are multiple processes that might want to use HEARTBEAT for proof of life and that we need to address. The implication though is probably that we need a clear description that this means "a component representing a set of services on the companion", not "the companion has died, along with all components on it. If you see what I mean?

Anyway, I'll think about it over the weekend. Nothing you need to worry about.

message_definitions/v1.0/common.xml
julianoes: I would think that's enough, yes.

DanielePettenuzzo: ok I will change it to 16

message_definitions/v1.0/common.xml
DanielePettenuzzo: Got it. I can also join the next call if this comes up again. 

So I guess MAV_COMP_ID_ONBOARD_COMPUTER should be fine for now

DanielePettenuzzo:
@hamishwillee @julianoes thank you for being so responsive. I pushed the new comp id name and change the transaction id from uint32 to uint16.

DanielePettenuzzo:
also rebased since github was complaining

LorenzMeier:
@DanielePettenuzzo There is an actual CI failure here. Could you please run the generator locally to resolve it?

julianoes:
CI looks unrelated. It's 404 of npm install as far as I can see.

LorenzMeier:
Ok, so the same as on the other PR, I missed that originally.

julianoes:
I assume so. Not sure if it's an intermittent thing or if some libraries will need updating (in another PR).

PR #1444, opened 2020-08-06
common: add "error" flag for calibration state

julianoes:
This is not an error per se but something that could be happening and a flag would explain why the gimbal is not working properly as it is calibrating.

FYI @olliw42


olliw42:
shouldn't it be xxx_ERROR_FLAGS_yyy and not xxx_ERROR_yyy 

maybe ERROR_FLAGS_IS_CALIBRATING or ERROR_FLAGS_CALIBRATION_RUNNING or such, to maybe make it somewhat better clear that this isn't an error but a current condition/state

julianoes:
Absolutely, thanks @olliw42.

julianoes:
This is all still WIP, and this was just a small addition, so I went ahead and merged it.

PR #1446, opened 2020-08-07
common: minor cleanup regarding scaled param5/param6

dayjaby:
Whether param5/param6 are scaled or not depends on if it is sent as COMMAND_LONG/COMMAND_INT/MISSION_INT. It is not a property of the command itself, so we can cleanup any remarks regarding the scaling in the commands.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1897,8 +1897,8 @@
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5" label="Latitude" units="degE7">Latitude of ROI location</param>
-        <param index="6" label="Longitude" units="degE7">Longitude of ROI location</param>
+        <param index="5" label="Latitude" units="deg">Latitude of ROI location</param>
+        <param index="6" label="Longitude" units="deg">Longitude of ROI location</param>
         <param index="7" label="Altitude" units="m">Altitude of ROI location</param>
       </entry>
       <entry value="196" name="MAV_CMD_DO_SET_ROI_WPNEXT_OFFSET" hasLocation="false" isDestination="false">
@@ -2627,8 +2627,8 @@
         <param index="2" label="Approach Vector" units="deg" minValue="-1" maxValue="360">Desired approach vector in compass heading. A negative value indicates the system can define the approach vector at will.</param>
         <param index="3" label="Ground Speed" minValue="-1">Desired ground speed at release time. This can be overridden by the airframe in case it needs to meet minimum airspeed. A negative value indicates the system can define the ground speed at will.</param>
         <param index="4" label="Altitude Clearance" units="m" minValue="-1">Minimum altitude clearance to the release position. A negative value indicates the system can define the clearance at will.</param>
-        <param index="5" label="Latitude" units="degE7">Latitude. Note, if used in MISSION_ITEM (deprecated) the units are degrees (unscaled)</param>
-        <param index="6" label="Longitude" units="degE7">Longitude. Note, if used in MISSION_ITEM (deprecated) the units are degrees (unscaled)</param>
+        <param index="5" label="Latitude" units="deg">Latitude</param>
+        <param index="6" label="Longitude" units="deg">Longitude</param>
         <param index="7" label="Altitude" units="m">Altitude (MSL)</param>
       </entry>
       <entry value="30002" name="MAV_CMD_PAYLOAD_CONTROL_DEPLOY" hasLocation="false" isDestination="false">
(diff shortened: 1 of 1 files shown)


amilcarlucas:
For consistency, please add "deg" units to all instances

hamishwillee:
@amilcarlucas @dayjaby My understanding is that we do mean degE7 - you must use the variant of mission or command message that can send this format. Generally LONG variants should not be used except where "really" needed - ie where param5/6 are really floats.

So my take is that we don't change the units on those that have units. Note sure about the "unscaled" comments - this just means you use deg or degE7 depending on your type?

@julianoes @auturgy Am I misinterpreting?

dayjaby:
Can't we create a new annotation which allows to describe it more explicitely, like <send-as>COMMAND_INT</send-as>?

hamishwillee:
> Can't we create a new annotation which allows to describe it more explicitly, like COMMAND_INT?

YOu mean a tag to indicate that a particular MAV_CMD can only be sent in a particular message type?

dayjaby:
Yes, I think "degE7" tried to provide this information, but the user needs some prior knowledge so it's good to have a clear unambigiuous distinction instead.

hamishwillee:
Thanks for the clarification. Then IMO the units should revert to degE7 as that is the expected unit (and would be sent as that unit with loss of precision in a float). 

If we wish to specify that a command can only be sent in a message with INT for params 5/6 then that should be done as new markup and will require updates to the XSD - @amilcarlucas is the expert on that - any thoughts/objections?

julianoes:
> My understanding is that we do mean degE7 - you must use the variant of mission or command message that can send this format. Generally LONG variants should not be used except where "really" needed - ie where param5/6 are really floats.

I expect deg for float/COMMAND_LONG and degE7 for int32/COMMAND_INT. As far as I'm aware both are allowed, at least I don't know of any place where we have specified this on purpose.

I don't know how that would be documented in xml. Both, deg and degE7 can be right or wrong, right?

amilcarlucas:
Due to precision issues, most users will want to use the COMMAND_INT/degE7 variant of this command. So I vote to document that in the units .xml attribute. The other ( float/COMMAND_LONG ) use case is used less often and the users will need to read the COMMAND_LONG documentation to figure out that they must not multiply by 1E7 before sending the message. That is a good compromise in my eyes

hamishwillee:
Yes, either unit can be used, depending on the message it is sent in. However no one _should_ use COMMAND_LONG for sending positional information. So to me these should **all** be set as degE7. I'm certainly not going to merge a change that makes those into deg.

The right thing to do here is 
- anything that was original degE7 be reverted. 
- if something had no units but description said unscaled then that is deg - so `<param index="5" label="Latitude">Latitude unscaled</param>` goes to `<param index="5" label="Latitude" unit="deg">Latitude</param>`

@dayjaby Are you happy to make that change?

dayjaby:
That's just the old approach which feels wrong to me.

The scaling is a property of COMMAND_INT vs COMMAND_LONG, not of the command itself. Imagine there will be a new message like COMMAND_INT64 for even better precision. All the degE7 comments will be wrong then.

So what I expect to see is that a command like MAV_CMD_DO_SET_ROI_LOCATION has a warning like "must be sent as a COMMMAND_INT or MISSION_INT" and that COMMAND_INT/MISSION_INT documents the E7 scaling, like it already does nicely!

"Must be sent as COMMAND_INT" is also less ambigious for the firmware or ground control developers, which might ask themselves whether to use COMMAND_INT or COMMAND_LONG and might end up implementing both because there is no strict guideline.

In the command documentation it says `Any command may be packaged in this message, but there may be some loss of precision for positional co-ordinates (latitude, longitude).`, which is wrong for firmware versions which strictly deny positional commands on COMMAND_LONG.

hamishwillee:
@dayjaby This stuff can get pretty pedantic. 

For example **Any command may be packaged in this message, but there may be some loss of precision for positional co-ordinates (latitude, longitude)** is correct - the fact that some flight stacks might choose to deny positional commands on COMMAND_LONG does not stop them being packaged.

I agree with you that this feels wrong and that the scaling is a property of the message.  But I don't particularly like this new solution better than the old one. For a start, as soon as I add "Must be sent as COMMAND_INT or MISSION_INT" someone will come along and tell me that they send it already using the COMMAND_LONG and I can't retrospectively change the protocol - or they will come along and say "yes, let's invent another message that takes double precision floats, so all messages can be sent in anything". 

I'm slightly partial to auto-adding text to at least make the _recommendation_ of using COMMAND_INT/MISSION_INT for any field that has degE7 in param5, 6. Would still keep degE7.

As you can probably tell, this is not the first time this has come around. I've been trying to schedule a discussion in the dev call, but it keeps on being deferred.

amilcarlucas:
let's just add the degE7 then.

hamishwillee:
I am closing this - it reduces the usefulness for people who know how this works, while at the same time does not really help newbies.

PS
- I have added some more information in [mission protocol docs](https://mavlink.io/en/services/mission.html#frames--positional-information) about how this works.
- there is also an interesting proposal here: https://mavlink.io/en/services/mission.html#frames--positional-information which might remove the need for encoding at all. 



https://github.com/mavlink/mavlink/issues/1523

PR #1451, opened 2020-08-20
Add yaw to HIL_GPS messsage

Jaeyoung-Lim:
This PR adds a yaw field to the `HIL_GPS` message.

This is needed in order to simulate GPS sensors that provide yaw information in HITL / SITL simulations. 

This is added as an `extention` and is a identical field with the yaw field https://mavlink.io/en/messages/common.html#GPS_INPUT
- yaw is in cdeg which indicate the yaw from north.
- North is marked as 36000 and yaw=0 invalidates the field.


bresch (approved):
Thanks!

LorenzMeier (approved):

PR #1452, opened 2020-08-20
Move transaction from messages to a command.

dogmaphobic:
Making this as a command instead of messages allows the GCS to make use of the existing state machine already in place.

To start a transaction, send the `MAV_CMD_PARAM_TRANSACTION` command using `PARAM_TRANSACTION_ACTION_START` for `param1`, `PARAM_TRANSACTION_ACTION_COMMIT` to commit and so on.

Command responses are handled as usual, using the `MAV_RESULT` responses.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1373,21 +1373,6 @@
         <description>Parameter meta data.</description>
       </entry>
     </enum>
-    <enum name="PARAM_TRANSACTION_RESPONSE">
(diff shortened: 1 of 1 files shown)


DanielePettenuzzo:
@hamishwillee This is a continuation of the parameter transactions that we discussed in the past weeks.

hamishwillee:
@dogmaphobic This makes sense, given that we get to re-use all the resending logic. Is the intent to implement the commit message as a [long running command protocol operation](https://mavlink.io/en/services/command.html#long_running_commands)?

I should write a checklist on new protocol design that covers this sort of design best practise.

PR #1453, opened 2020-08-21
message_definitions: add an all.xml to try to avoid conflicts in files which don't already conflict

peterbarker:


auturgy:
Any comments on this before merge?
It will greatly aid standardization attempts, and as it isn't intended to be used outside CI/test suits, doesn't break anything.

hamishwillee:
@auturgy I think it makes sense. Technically it isn't "All", but it is everyone who seems to want to be compatible.  Is it worth fixing up the clashes - i.e. in the test?

@LorenzMeier @julianoes Any objections to this?

julianoes (approved):
Great idea!

auturgy:
@hamishwillee we need to engage with whoever manages the xml's with clashes.  It may be some of the xml's can be retired?

hamishwillee:
@auturgy I guess so. Let's merge and iterate. 
@LorenzMeier Any objections?

hamishwillee:
Merged, as discussed in dev call.

Jaeyoung-Lim:
@peterbarker The fact that ASLUAV dialect was the only dialect not included, and got merged without any notification or discussion with the ASL dialect is very unfortunate.

This also very much contradicts the discussion that happened in https://github.com/mavlink/mavlink/pull/1596

auturgy:
@Jaeyoung-Lim excluding the ASL dialect is completely consistent with what @acfloria requested here: https://github.com/mavlink/mavlink/pull/1596#issuecomment-797320569

peterbarker:
> @peterbarker This PR essentially makes it extremely hard to keep our work up to date with upstream, preventing us to contribute back

I'm sorry to hear that.  In what ways does the presence of `all.xml` make it difficult to keep up-to-date or contribute back?

> The fact that ASLUAV dialect was the only dialect not included, and got merged without any notification or discussion with the ASL dialect is very unfortunate.

ASLUAV is most certainly not the only dialect not included in `all.xml`.  There are 5 dialects mentioned in `all.xml` which did not get in - I noted the first reason in XML comments in there, and I'm hopeful that over time the conflicts can be resolved somehow - perhaps through common messages being adopted by the relevant project, or by messages moving ID.  `SENSOR_OFFSETS` in the ArduPilot is long-standing - but I'm really not sure we can't move it - in all the log analysis I've ever done it hasn't actually been useful to my recollection.  That being said, if `AQ_TELEMETRY_F` was no longer in use then removing that message definition would be preferable.  I've raised https://github.com/ArduPilot/ardupilot/pull/18368 for discussion to see what people think of the concept.

IIRC (and this was 8 months ago!) I did attempt to search the projects out to work out if the message was still in use.
 
> IMHO, the prerequisite for this PR to be merged is to at least come up with an action item of how the dialects should be

I've been kind of hoping that GCSs might start compiling against `all.xml` rather than against `ardupilotmega.xml`.  Hasn't happened so far.  That will make swiping good ideas from other dialects down into `common.xml` rather more easy.

> handled(or a discrete decision that dialects are no longer the way to handle these message definition), since ardupilotmega dialect is also one of the dialects that seems to get preferential priorities over other dialects.

Presence in `all.xml` does not make a dialect valid or invalid.  Yes, `ardupilotmega.xml` got preferential treatment here because I did the work, and I didn't want the people who'd put so much effort in removing duplication issues to have their work wasted.

Perhaps `all.xml` was a bit presumptuous/inconsiderate... but I initially was really hoping that we'd be able to wedge all of them in.

hamishwillee:
@Jaeyoung-Lim I'm also confused by this necrobump. I don't understand the issue you have raised, but I know we won't be able to address it without an explanation of how this makes it harder for you to keep work up to date with upstream. 

Can you please create a new issue to discuss this. Perhaps come along to the mavlink call today?

PR #1456, opened 2020-08-24
BATTERY_STATUS add faults, remove SMART_BATTERY_STATUS

hamishwillee:
smart_battery* messages do not include battery function ( e.g. avionics, payload, whatever) or chemistry. It occurs to me the function is necessary if multiple smart batteries might be present 

@DonLakeFlyer Looking at https://github.com/mavlink/qgroundcontrol/pull/9019 you support battery_info now. I presume it would be useful to be able to treat the smart battery messages as a superset. Is there anything else this would need in order for a smart battery to be used in the same way as a battery_status? Note that there are two smart battery messages.

Note also, I added these inline rather than extensions as messages are WIP.

@AlexKlimaj @LorenzMeier @julianoes @auturgy Thoughts?


DonLakeFlyer:
Yes I saw `SMART_BATTERY_STATUS` and to me something smells wrong with respect to having two different messages for this. The fact is that 'SMART_BATTERY_STATUS` is going to need every field that is already in `BATTERY_STATUS`. Just add `fault_bitmask` to `BATTERY_STATUS` and call it good.

hamishwillee:
@DonLakeFlyer Have you seen both [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO) and [SMART_BATTERY_STATUS](https://mavlink.io/en/messages/common.html#SMART_BATTERY_STATUS)? Essentially we've separated out the static or slow update info from the dynamic info. There is a lot more in there than just a fault too.
There is a bit of design by committee here, but this is what got settled on after a lot of discussion.

What I'd rather do is make sure that the smart_battery messages could be used instead of BATTERY_STATUS - ie provide an upgrade path.

amilcarlucas (approved):

DonLakeFlyer:
> What I'd rather do is make sure that the smart_battery messages could be used instead of BATTERY_STATUS

And you will in turn make every GCS have to write new code supporting both `BATTERY_STATUS` and `SMART_BATTERY_STATUS` which doesn't make sense. Basic battery status is basic battery status no matter what kind of battery is it and that is covered well already by `BATTERY_STATUS`.

DonLakeFlyer:
Also I assume `SMART_BATTERY_INFO` is static metadata. If so, that information should come through the `COMPONENT_INFORMATION` system which is the new standard for getting metadata about a component. No need to make any more new messages.

hamishwillee:
@DonLakeFlyer I wish you would occasionally attend the discussions on messages. The insights are valuable, but applying them more than year after the messages went in is a bit unhelpful.

Yes, SMART_BATTERY_INFO is basically static info. It would be convenient for a GCS to do have this as metadata but it would force every smart battery to implement COMPONENT_INFORMATION and MAVLink FTP in order to send this information. Do you think that is reasonable or it just mean no one bothers?
In any case this could be considered a completely separate "question" from the status update.

W.r.t the battery vs smart battery info perhaps we could just use BATTERY_STATUS. Differences below. The main one being that SMART_BATTERY is more efficient, and already able to be extended to larger cell counts  via offsets. 

BATTERY_STATUS 

Field Name | Type | Units | Values | Description
-- | -- | -- | -- | --
id | uint8_t |   |   | Battery ID
battery_function | uint8_t |   | MAV_BATTERY_FUNCTION | Function of the battery
type | uint8_t |   | MAV_BATTERY_TYPE | Type (chemistry) of the battery
temperature | int16_t | cdegC |   | Temperature of the battery. INT16_MAX for unknown temperature.
voltages | uint16_t[10] | mV |   | Battery voltage of cells 1 to 10 (see voltages_ext for cells 11-14). Cells in this field above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).
current_battery | int16_t | cA |   | Battery current, -1: autopilot does not measure the current
current_consumed | int32_t | mAh |   | Consumed charge, -1: autopilot does not provide consumption estimate
energy_consumed | int32_t | hJ |   | Consumed energy, -1: autopilot does not provide energy consumption estimate
battery_remaining | int8_t | % |   | Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.
time_remaining ** | int32_t | s |   | Remaining battery time, 0: autopilot does not provide remaining battery time estimate
charge_state ** | uint8_t |   | MAV_BATTERY_CHARGE_STATE | State for extent of discharge, provided by autopilot for warning or external reactions
voltages_ext ** | uint16_t[4] | mV |   | Battery voltages for cells 11 to 14. Cells above the valid cell count for this battery should have a value of 0, where zero indicates not supported (note, this is different than for the voltages field and allows empty byte truncation). If the measured value is 0 then 1 should be sent instead.


SMART_BATTERY_STATUS

Field Name | Type | Units | Values | Description
-- | -- | -- | -- | --
id | uint16_t |   |   | Battery ID
capacity_remaining | int16_t | % |   | Remaining battery energy. Values: [0-100], -1: field not provided.
current | int16_t | cA |   | Battery current (through all cells/loads). Positive if discharging, negative if charging. UINT16_MAX: field not provided.
temperature | int16_t | cdegC |   | Battery temperature. -1: field not provided.
fault_bitmask | int32_t |   | MAV_SMART_BATTERY_FAULT | Fault/health indications.
time_remaining | int32_t | s |   | Estimated remaining battery time. -1: field not provided.
cell_offset | uint16_t |   |   | The cell number of the first index in the 'voltages' array field. Using this field allows you to specify cell voltages for batteries with more than 16 cells.
voltages | uint16_t[16] | mV |   | Individual cell voltages. Batteries with more 16 cells can use the cell_offset field to specify the cell offset for the array specified in the current message . Index values above the valid cell count for this battery should have the UINT16_MAX value.

....

DonLakeFlyer:
> but it would force every smart battery to implement COMPONENT_INFORMATION and MAVLink FTP in order to send this information. Do you think that is reasonable or it just mean no one bothers?

No I don't think it's reasonable and is why I deleted the comment a few minutes after I realized it was stupid.

hamishwillee:
@DonLakeFlyer Yes, but the theory is nice. I do like the idea of adding SMART_BATTERY_INFO  to the API for those systems that want it and using BATTERY_STATUS for the SMART_BATTERY status. Will add to devcall.  @AlexKlimaj - any opinion/thoughts?

DonLakeFlyer:
I don't know how safe it is to do this but if you add `MAV_SMART_BATTERY_FAULT` faults to `MAV_BATTERY_CHARGE` state then you can get rid of `SMART_BATTERY_STATUS` and you are done.

hamishwillee:
Sounds like a plan to discuss in dev call. @auturgy @MaEtUgR Anyone we should involve in discussion of changes to WIP smart batteries as per https://github.com/mavlink/mavlink/pull/1456#issuecomment-679443398 ?

auturgy:
I don’t think the smart_battery messages are implemented anywhere?
If not, basically agree with Don, although the component_information mechanism won’t work - it’s a mavlink service whereas most/all smart batteries are smbus or can.

AlexKlimaj:
I agree that extending charge_state in BATTERY_STATUS could handle almost everything SMART_BATTERY_STATUS does. SMART_BATTERY_INFO is probably the more useful of the messages for someone implementing a smart battery on their system.

message_definitions/v1.0/common.xml
```diff
@@ -2833,24 +2833,24 @@
         <description>Battery is charging.</description>
       </entry>
     </enum>
-    <enum name="MAV_SMART_BATTERY_FAULT" bitmask="true">
+    <enum name="MAV_BATTERY_FAULT" bitmask="true">
```
hamishwillee, line 2836: FYI, renamed. @DonLakeFlyer @julianoes How would we/do we need to indicate that faults are not supported by the battery?

hamishwillee:
@DonLakeFlyer We discussed in dev call and everyone liked your proposal re smart batteries. The PR now modified to 
- remove SMART_BATTERY_STATUS (was WIP)
- Update BATTERY_STATUS to mention SMART_BATTERY_INFO for smart batteries, and visa versa
- Add BATTERY_STATUS fault reporting as an extension (with rename of the enum to MAV_BATTERY_FAULT). 
  - **Question** - do we need to indicate the case where fault reporting not supported, and if so, how? - as is, all current batteries will show as having no faults, even if faults might exist.

Is there anything we can do about this field? Specifically the proposed smart battery was gong to use negative values for charging. Given this is an int, that would be useful. Normally we'd use a max value for "not supported".
```
<field type="int16_t" name="current_battery" units="cA">Battery current, -1: autopilot does not measure the current</field>
```

@julianoes Any comments?

DonLakeFlyer:
How does 'fault_bitmask' relate to 'charge_state' values which also have failure indicators.

message_definitions/v1.0/common.xml
julianoes: We can't have 0 to signal unsupported because these are power of 2 flags and there is no way to set 0. So it would have to be 1, I think.

hamishwillee: @DonLakeFlyer Does a GCS need to know that fault reporting is supported or not? If the answer is no, then let's leave as is.

You shouldn't really add a bit to indicate "faults are not supported" - for a start it will mean that we can't trim the message if they aren't supported and there are no faults.

hamishwillee:
> How does 'fault_bitmask' relate to 'charge_state' values which also have failure indicators.

@DonLakeFlyer My assumption is that faults indicate that the battery has failed (can't be used) or is unhealthy (can be used). So any fault will result in either of `MAV_BATTERY_CHARGE_STATE_FAILED` or `MAV_BATTERY_CHARGE_STATE_UNHEALTHY` also being set. 

@AlexKlimaj @MaEtUgR The faults we came up with are listed: `MAV_SMART_BATTERY_FAULT_SPIKES`, `MAV_SMART_BATTERY_FAULT_SINGLE_CELL_FAIL`, `MAV_SMART_BATTERY_FAULT_OVER_CURRENT`, `MAV_SMART_BATTERY_FAULT_OVER_TEMPERATURE`, `MAV_SMART_BATTERY_FAULT_UNDER_TEMPERATURE`

Which of these indicate failed vs unwell ?

cmic0:
I think would also make sense to introduce a `mode` field in `BATTERY_STATUS` to handle the various states of a battery (e.g. storage / Disarmed (powers the vehicle but with limited current) / Armed (disabled over current protection for flight)). Thoughts?

hamishwillee:
@AlexKlimaj Can you comment on @cmic0 post above? How would the information be used? Is there anyone else we know who has this information?

cmic0:
For the use case I am targeting I want to use this information to define a preflight check:
You have multiple batteries that you want to connect on the vehicle. When you plug in a battery, this one will check if the bus voltage of the vehicle is compatible with its own voltage and only transition to "ready to fly" if the delta voltage is OK. This will basically prevent flying if you plug an empty battery and a fully charged battery

DonLakeFlyer:
> So any fault will result in either of `MAV_BATTERY_CHARGE_STATE_FAILED` or `MAV_BATTERY_CHARGE_STATE_UNHEALTHY` also being set.

Ok, that make sense then.

message_definitions/v1.0/common.xml
DonLakeFlyer: You need to specifically connect charge_state reporting a fault to these bits and then you know if unsupported. For example:
* 'charge_state' equal to 'FAILED' or 'UNHEALTHY' signals that 'fault_bitmask' should also be set
* then if 'fault_bitmask' = 0 for that case it indicates that 'fault_bitmask' is not supported

hamishwillee:
@cmic0 Can you please post the additions you like to the PR here (as a comment). 

Once you've done that will ensure gets discussed in the dev call

message_definitions/v1.0/common.xml
hamishwillee: Thanks. I've done that in the `fault_bitmask`

hamishwillee:
@DonLakeFlyer @AlexKlimaj Do Either of you have an opinion on @cmic0 suggestion to add a `BATTERY_STATUS.mode` field to handle the various states of a battery (e.g. storage / Disarmed (powers the vehicle but with limited current) / Armed (disabled over current protection for flight)). 

From above he says 

> For the use case I am targeting I want to use this information to define a preflight check:
You have multiple batteries that you want to connect on the vehicle. When you plug in a battery, this one will check if the bus voltage of the vehicle is compatible with its own voltage and only transition to "ready to fly" if the delta voltage is OK. This will basically prevent flying if you plug an empty battery and a fully charged battery

DonLakeFlyer:
If you plug in a discharged battery it should report `charge_state=MAV_BATTERY_CHARGE_STATE_LOW' or something along those lines. Two things I don't understand:
* Why that isn't enough information to determine a pre-flight check
* What the whole "bus voltage of the vehicle" stuff is talking about

AlexKlimaj:
@hamishwillee I agree with @DonLakeFlyer . I don't think the mode of the battery is needed. The FC can look at the voltage or % remaining and determine not to arm if there is a large delta.

hamishwillee:
Thanks @AlexKlimaj ! Do you have a view on whether the faults/which faults are fail vs unhealthy?

AlexKlimaj:
Off the top of my head: Over voltage, under voltage, over temperature, under temperature, voltage imbalance, state of health (capacity degradation), short circuit, and over discharge.
I'll try to think of some other potential faults.

hamishwillee:
@AlexKlimaj Thanks, but that is not quite what I meant. We already have those faults listed (MAV_SMART_BATTERY_FAULT_SPIKES, MAV_SMART_BATTERY_FAULT_SINGLE_CELL_FAIL, MAV_SMART_BATTERY_FAULT_OVER_CURRENT, MAV_SMART_BATTERY_FAULT_OVER_TEMPERATURE, MAV_SMART_BATTERY_FAULT_UNDER_TEMPERATURE)



The question is, what does each of these mean in terms of battery health? For example, If I get MAV_SMART_BATTERY_FAULT_OVER_TEMPERATURE fault does that mean the whole battery has failed and should be thrown away, or does it just mean that the battery is unhealthy and we might limp home on it :-).

Or to put it another way, are any of these just indicators of potential problems that we can ignore in the short term, or are they all critical immediately?

AlexKlimaj:
Hm, I think people will have different ideas of what is a failed battery for their own setup. But here is my interpretation.

MAV_SMART_BATTERY_FAULT_SPIKES - Not sure what this one means. 
MAV_SMART_BATTERY_FAULT_SINGLE_CELL_FAIL - One cell has failed therefore the pack has failed and should not be used. 
MAV_SMART_BATTERY_FAULT_OVER_CURRENT - Load is sinking more current than battery can handle. Not necessarily a failure. 
MAV_SMART_BATTERY_FAULT_OVER_TEMPERATURE - Not necessarily a failure in flight but might be a pre flight check error.
MAV_SMART_BATTERY_FAULT_UNDER_TEMPERATURE - Same as above.

hamishwillee:
Thanks very much @AlexKlimaj ! 
I have noted that if any faults are set either the unhealthy or failed charge state must also be set. For the specific case of cell fail I've noted that MAV_BATTERY_CHARGE_STATE_FAILED should also be set. 

I haven't done anything for the others - because IMO it is up to the battery vendor to decide whether the fault is critical or just an indication of a potential problem.

@DonLakeFlyer I think this is good to merge - any objections? I think @cmic0 proposals are orthogonal so we could merge this and discuss those separately.

cmic0:
Thanks for the feedback @AlexKlimaj / @DonLakeFlyer .
By looking at the FAULTS definitions I think that adding something like `MAV_SMART_BATTERY_FAULT_VOLTAGE_INCONSISTENCY` to indicate that the battery has a different voltage compared to the vehicle's one would address the use case i am targeting.

> The FC can look at the voltage or % remaining and determine not to arm if there is a large delta.

I think this should be handled by the battery itself as different smart batteries might have a different threshold for the inconsistency, the autopilot should only receive the info.

DonLakeFlyer:
> The question is, what does each of these mean in terms of battery health? For example, If I get MAV_SMART_BATTERY_FAULT_OVER_TEMPERATURE fault does that mean the whole battery has failed and should be thrown away, or does it just mean that the battery is unhealthy and we might limp home on it

I would say that is not up to mavlink spec but up to firmware/battery implementation.  The correct status of that should be reported in 'charge_state'.

DonLakeFlyer:
> I think this is good to merge - any objections?

Looks good to me

AlexKlimaj:
> 
> 
> Thanks for the feedback @AlexKlimaj / @DonLakeFlyer .
> By looking at the FAULTS definitions I think that adding something like `MAV_SMART_BATTERY_FAULT_VOLTAGE_INCONSISTENCY` to indicate that the battery has a different voltage compared to the vehicle's one would address the use case i am targeting.
> 
> > The FC can look at the voltage or % remaining and determine not to arm if there is a large delta.
> 
> I think this should be handled by the battery itself as different smart batteries might have a different threshold for the inconsistency, the autopilot should only receive the info.

Batteries aren't publishing MAVLINK directly. They are using UAVCAN or SMBUS to the flight controller. So I don't know if it makes sense.

hamishwillee:
> Batteries aren't publishing MAVLINK directly. They are using UAVCAN or SMBUS to the flight controller. So I don't know if it makes sense.

The wording in the PR just allows the sender to decide whether the fail is critical or warning for the types where it can be either. The flight controller might be what decides. Either way, by not saying "you must send this" we allow the flexibility to be defined in the implementation.

message_definitions/v1.0/common.xml
```diff
@@ -2834,35 +2834,44 @@
...
-      <entry value="16" name="MAV_SMART_BATTERY_FAULT_OVER_TEMPERATURE">
+      <entry value="16" name="MAV_BATTERY_FAULT_OVER_TEMPERATURE">
         <description>Over-temperature fault.</description>
       </entry>
-      <entry value="32" name="MAV_SMART_BATTERY_FAULT_UNDER_TEMPERATURE">
+      <entry value="32" name="MAV_BATTERY_FAULT_UNDER_TEMPERATURE">
         <description>Under-temperature fault.</description>
       </entry>
+      <entry value="64" name="MAV_BATTERY_FAULT_CELL_IMBALANCE">
+        <description>Cells have significantly different voltages fault.</description>
```
hamishwillee, line 2867: @cmic0 Can you check this is a reasonable description of MAV_BATTERY_FAULT_CELL_IMBALANCE, and provide alternative if not.

message_definitions/v1.0/common.xml
```diff
@@ -2834,35 +2834,44 @@
...
+      </entry>
+      <entry value="128" name="MAV_BATTERY_FAULT_INCOMPATIBLE_BUS_VOLTAGE">
+        <description>Bus voltage cannot be matched by battery fault.</description>
```
hamishwillee, line 2870: @cmic0 Can you check this is a reasonable description of MAV_BATTERY_FAULT_INCOMPATIBLE_BUS_VOLTAGE, and provide alternative if not.

message_definitions/v1.0/common.xml
```diff
@@ -2834,35 +2834,50 @@
...
+        <description>Battery is diagnosed to be defective or an error occurred, usage is discouraged / prohibited. Possible causes (faults) are listed in MAV_BATTERY_FAULT.</description>
       </entry>
       <entry value="7" name="MAV_BATTERY_CHARGE_STATE_CHARGING">
         <description>Battery is charging.</description>
       </entry>
     </enum>
-    <enum name="MAV_SMART_BATTERY_FAULT" bitmask="true">
-      <description>Smart battery supply status/fault flags (bitmask) for health indication.</description>
-      <entry value="1" name="MAV_SMART_BATTERY_FAULT_DEEP_DISCHARGE">
+    <enum name="MAV_BATTERY_MODE">
```
hamishwillee, line 2846: @cmic0 Can you please give me a list of other battery modes that we might want to report on?

hamishwillee:
@DonLakeFlyer @cmic0 @auturgy @LorenzMeier I have added a "battery mode" field before the fault reporting. An example that Lorenz provided on why this might be useful is to be able to know that the reason for a crash is that someone switched to storage mode in flight.

This is not the implementation suggested in the dev call, which was to change the fault bitmask to be more general "state" information that includes modes as a flag - e.g. MAV_BATTERY_STATE_STORAGE_MODE. Once I tried it I realised it can work, but it is hard to understand.

An alternative would have been to simply word the problem cases as a fault - e.g. MAV_BATTERY_FAULT_STORAGE_MODE_WHEN_ARMED. However I have assumed that it is generally useful to know the mode.

Any concerns with this?

amilcarlucas:
Please do not forget to squash before merging

DonLakeFlyer:
So working under the assumption that "storage mode" means that the battery is discharging itself for long term storage. I would think this is more appropriate to be included in 'charge_state' which already has a `MAV_BATTERY_CHARGE_STATE_CHARGING`. To me storage mode is pretty similar to charging mode.

All of that said I know little about how these batteries work.

hamishwillee:
> So working under the assumption that "storage mode" means that the battery is discharging itself for long term storage. I would think this is more appropriate to be included in 'charge_state' which already has a MAV_BATTERY_CHARGE_STATE_CHARGING. To me storage mode is pretty similar to charging mode.

1. Still waiting on @cmic0 to post the set of modes so we can test the assumption - it isn't just storage mode - there are modes like "battery can be hot-swapped" apparently. 
2. If a battery is both in storage mode and at a critical battery level, what would you expect it to set for the charging state? 
3. The reason a distinct state was suggested was so that we can catch cases like "battery switched to storage mode during flight" - I worry about the potential for losing this information.

cmic0:
I thought about it and i believe that the reasonable modes are:

- `MAV_BATTERY_MODE_AUTO_DISCHARGING`: battery is auto discharging, this might be triggered due to idle time, i.e if the battery has not been used for more than a certain time auto discharges until cells have reached the storage voltage. If the battery has already reached the storage voltage we can flag it by introducing `MAV_BATTERY_CHARGE_STATE_STORAGE`

- `MAV_BATTERY_MODE_HOT_SWAP`: battery limits its current to prevent short bursts of high currents that might lead to sparks and damage sensitive electrical circuits.

- `MAV_BATTERY_MODE_FLIGHT`: battery current limits are disabled. If a failure occurs the battery will only notify it (with `MAV_BATTERY_FAULT`) without switching out from the flight mode.

So thinking on a real use case (for context):
If i pick a battery of the shelf and i plug it on my vehicle this one will be in `MAV_BATTERY_MODE_HOT_SWAP` (I am only doing bench tests, so it's good that the battery is power-limiting the output) and most likely will also report `MAV_BATTERY_CHARGE_STATE_STORAGE`. When i'll be out flying the battery will transition to `MAV_BATTERY_MODE_FLIGHT`

hamishwillee:
@DonLakeFlyer Can I please have your advice.  @cmic0 has proposed a new storage charge state and a mode [above ^^^^](https://github.com/mavlink/mavlink/pull/1456#issuecomment-695767885)

This can't work because `charge_state` is full (8 bit). If it was 16 bit I'd probably go along with your suggestion of making all of these into charge states. 

Putting all these in a mode field works except that:
- If the field is used, all trimming is lost most of the time. Could we choose to not have the "normal operation" and assume that? That way this could be zero most of the time except when landed?
- It isn't clear what charging state you should use for the storage mode etc. Thoughts?

Alternatively we could put them in with the faults field - and change meaning to "flags". But then how do I indicate the charge state? I currently say that you must set a charge state of failed or unhealthy for the faults as a way to indicate whether or not fault reporting is supported.
The only workaround I see is to change the meaning of unhealthy to be more "not ready" - either a fault or simply not in the right state yet. 

Suggestions welcome. I'm a bit stumped on best way forward.
Worth a quick chat tomorrow morning?

DonLakeFlyer:
My main concern is understanding what a GCS would do with this information. I'll explain with how QGC works:
* QGC shows battery indicators to the in the toolbar, one for each autopilot battery
* It shows an icon and either battery percentage or voltage depending on what is available
* It colors the information for each battery based on the `charge_state` field:
```
            function getBatteryColor() {
                switch (battery.chargeState.rawValue) {
                case MAVLink.MAV_BATTERY_CHARGE_STATE_OK:
                    return qgcPal.text
                case MAVLink.MAV_BATTERY_CHARGE_STATE_LOW:
                    return qgcPal.colorOrange
                case MAVLink.MAV_BATTERY_CHARGE_STATE_CRITICAL:
                case MAVLink.MAV_BATTERY_CHARGE_STATE_EMERGENCY:
                case MAVLink.MAV_BATTERY_CHARGE_STATE_FAILED:
                case MAVLink.MAV_BATTERY_CHARGE_STATE_UNHEALTHY:
                    return qgcPal.colorRed
                default:
                    return qgcPal.text
                }
            }
```

If this new battery mode thing is just additional information user then fine. I would continue to assume that 'charge_state" will reflect a state which is appropriate for whatever "mode" the battery is in. Or simply put if the battery is in some "mode" which you shouldn't fly with then 'charge_state' is updated to reflect it somehow such that the indicator would say go red.

Unfortunately in looking at the possible values of `MAV_BATTERY_CHARGE_STATE' you'll find two problems:
* It already has somewhat conflated what I would consider "status" and "mode" since it has a `MAV_BATTERY_CHARGE_STATE_CHARGING` value. (And in hindsight QGC is likely doing to wrong thing with coloring for this value)
* Also it is missing some sort of general "not ready" status which could be used to indicate a smart battery which is in some mode you shouldn't fly with.

Because of this it means that if a GCS wants to show the user some high level indication of battery readiness with color it would need to look at both 'charge_state' and 'mode'. To me that seems really crap and fraught with failure or flakiness. The GCS has to make too many assumptions about what these modes actually mean. A GCS should not have to do that. It should have a single simple way of knowing "don't fly with this battery now".

A possible solution would be to include something as simple as a `NOT_READY` state to charge state or something along those lines. And then the firmware is in charge of determining which 'mode' you shouldn't fly with by updating 'charge_state' appropriately based on mode. And then a GCS has a simple way of not needed to understand these modes, but still show the user useful status.

hamishwillee:
@DonLakeFlyer Yes, for now at least `mode` field is information only, primarily for logging. Reasoning for this at end FYI only.

> **On that basis I assume you are happy for me to merge and will do so probably tomorrow.**

With respect to MAV_BATTERY_CHARGE_STATE being a bit messy I see the problem and I would like it clarified. However I don't think we should block this PR to do so. OK? If not, see my suggestion below and I can try do that as a separate PR.

> Can you create a PR that would work for you? 

With respect to that my leaning is towards:
-  adding MAV_BATTERY_CHARGE_STATE_NOT_READY which is any state where you can't take off. GCS would set this based on mode or whatever.
- Not sure about how we handle the charging state. My leaning is to deprecate it and move this to being a mode. The UI could choose to have additional icon to show charging when in the not ready state and in charging mode. Right now, if GCS got this it would treat as NOT_READY.

FYI Reasoning as to why mode is information only is mostly around "right now we do not know enough about possible impact of modes to define the integration with charging states. But with a mode field we have flexibility to work on this later and we get what we need for now. Specifically: 
- I suggested a not_ready charging state in which these modes should be active, but even that isn't quite suitable. Reason - as soon as the battery is connected to a system it should go out of these modes if the battery is smart enough.
- I suggested that these just be treated as faults and pushed out only when they are a problem (i.e. when armed). That sorts out this use case  - but then we would still likely end up being requested to add more modes in future so we end up having the discussion again. [If we do add as faults in future this will not be a problem for GCS because I would expect the Firmware to report the condition]

message_definitions/v1.0/common.xml
```diff
@@ -2834,35 +2834,50 @@
...
+      <description>Battery mode. Note, the normal operation mode (i.e. when flying) should be reported as MAV_BATTERY_MODE_UNKNOWN to allow message trimming in normal flight.</description>
+      <entry value="0" name="MAV_BATTERY_MODE_UNKNOWN">
```
hamishwillee, line 2848: FYI @DonLakeFlyer We also use MAV_BATTERY_MODE_UNKNOWN for "normal use". This is a compromise 
- if we had a separate mode for normal use it would be enabled when flying, removing trimming on this message.
- This means that there is no way to know just by looking at the message whether battery mode reporting is supported or not or if just the current mode is unknown, or if we're in normal mode. 

Upshot, this should just be treated like an informational flag for "special" modes. 

I don't see any workaround other than a completely new message.

DonLakeFlyer:
> Yes, for now at least `mode` field is information only, primarily for logging. Reasoning for this at end FYI only.
> 
> > **On that basis I assume you are happy for me to merge and will do so probably tomorrow.**

Yup, seems fine then.

DonLakeFlyer:
> With respect to that my leaning is towards:
> 
> * adding MAV_BATTERY_CHARGE_STATE_NOT_READY which is any state where you can't take off. GCS would set this based on mode or whatever.
> * Not sure about how we handle the charging state. My leaning is to deprecate it and move this to being a mode. The UI could choose to have additional icon to show charging when in the not ready state and in charging mode. Right now, if GCS got this it would treat as NOT_READY.

That all makes sense to me. How you handle that from a spec deprecation standpoint hard to say.

message_definitions/v1.0/common.xml
```diff
@@ -6492,8 +6509,10 @@
     <message id="370" name="SMART_BATTERY_INFO">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Smart Battery information (static/infrequent update). Use for updates from: smart battery to flight stack, flight stack to GCS. Use instead of BATTERY_STATUS for smart batteries.</description>
+      <description>Smart Battery information (static/infrequent update). Use for updates from: smart battery to flight stack, flight stack to GCS. Use BATTERY_STATUS for smart battery frequent updates.</description>
       <field type="uint8_t" name="id">Battery ID</field>
+      <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
```
hamishwillee, line 6514: @DonLakeFlyer FYI these are duplicates of what is in BATTERY_STATUS. 

They went in before you convinced us to use BATTERY_STATUS. I am however keeping them because I really hope one day we recreate a new BATTERY_STATUS and pull them out.

message_definitions/v1.0/common.xml
```diff
@@ -6492,8 +6509,10 @@
     <message id="370" name="SMART_BATTERY_INFO">
```
hamishwillee, line 6509: @DonLakeFlyer @auturgy Doesn't it make sense to rename `SMART_BATTERY_INFO` to `BATTERY_INFO`? while this is all WIP?

hamishwillee:
As Don as agreed this and everyone else has had a chance to comment I am merging this as-is. 

It should not impact anyone who does not want to implement it/has not implemented it because:
- Changes to BATTERY_STATUS default to 0 - if not implemented trimming still happens etc.
- Changes to SMART_BATTERY_INFO are breaking, but that is WIP and not implemented

Follow up changes required:
- Perhaps rename SMART_BATTERY_INFO to BATTERY_INFO
- Refactor battery charge status as per [here](https://github.com/mavlink/mavlink/pull/1456#issuecomment-698066628)

mRwu-Xian:
Is time_remaining ** in BATTERY_STATUS calculated based on the rate of remaining power and current power consumption?

Issue #1455, opened 2020-08-21
Docs: MAV_CMD_PREFLIGHT_UAVCAN.param1 not quite correct


DonLakeFlyer:
[MAV_CMD_PREFLIGHT_UAVCAN](https://mavlink.io/en/messages/common.html#MAV_CMD_PREFLIGHT_UAVCAN). 
It should also include 0 to stop assignment. Also I think "*Start* actuator ID assignment and direction mapping." instead of *Trigger* seems more understandable to me. When I first looked at the description for param1 = 1 I though it was wrong as well.

hamishwillee:
@DonLakeFlyer Better to (separate). create PRs for these things. Easier to review. 

So who came up with the protocols and what messages are involved? I think you're saying it goes like below - you send the same message to start and end update with different params, with settings being in the middle?

[...]

PR #1457, opened 2020-08-24
MAV_CMD_PREFLIGHT_UAVCAN - improve wording

hamishwillee:
@DonLakeFlyer This what you had in mind in https://github.com/mavlink/mavlink/issues/1455#issuecomment-678865832 ?


DonLakeFlyer:
Yup. That said I am not even remotely close to a UAVCAN expert. I know next to nothing about it. I only know what QGC does and what it tells the user to do. I don't know if there is missing detail here I don't know about.

hamishwillee:
@dagar @bkueng What is the message flow for uavcan preconfiguration?

Don things maybe that uavcan preconfiguration works like this:
- GCS sends message to tell autopilot to do "UAVCAN preconfig activities". That might be turning motors etc.
- User makes whatever changes are needed.
- User presses "Done"
- GCS sends second message to tell drone that preconfig is done.

Don noted that the message only gives an option to tell the autopilot that uavcan preconfig has started - and no way to say stopped. This change suggests how you might send stop.

My suspicion though is that there is just one command to trigger. So the current text is correct. I.e. 
- GCS sends message to tell autopilot to do "UAVCAN preconfig activities". 
- Autopilot does these activities - ie runs motors slowly for 5 seconds. At this point it stops. GCS has reset itself to steady state
- User fixes anything they need to.
- User presses "Done", GCS just closes the window.

So no need for a second message.

DonLakeFlyer:
> Don things maybe that uavcan preconfiguration works like this:

I don't really think that. That is exactly what QGC does.

hamishwillee:
@pavel-kirienko Do you have any insight on the intention for the (MAVLink) message that starts UAVCAN pre-flight checking (e.g. of motor direction)? The main thoughts on the topic here: https://github.com/mavlink/mavlink/pull/1457#issuecomment-678871488

If not, I will just update the behaviour to match what Don has implemented.

pavel-kirienko:
Hey guys, the original description is correct but the command is named incorrectly. To avoid confusion (case in point here) we should rename it into something like `UAVCAN_ENUMERATE`. It is intended to be executed not as a preflight procedure but only once in the process of the initial vehicle configuration (by a vendor or a maintenance technician; never in the field). This process can only be triggered; there is no need for a stop command because it is stopped automatically upon completion. I suggest closing this PR.

This function is only available for the legacy UAVCAN v0; there is currently no alternative in v1 and it may not be needed there at all because it is mostly valuable for a limited number of hobby users. At any rate, this question will be decided in a later stage of DS-015 development.

Links:

- https://github.com/PX4/Firmware/pull/2748
- https://github.com/UAVCAN/public_regulated_data_types/tree/legacy-v0/uavcan/protocol/enumeration
- https://dev.px4.io/v1.9.0/en/uavcan/node_enumeration.html
- it should be part of https://mavlink.io/en/guide/uavcan_interaction.html
- Feature in action (in this case it is used to enumerate motors, but it is equally applicable to any kind of onboard equipment): https://www.youtube.com/watch?v=4nSa8tvpbgQ

DonLakeFlyer:
Does sending MAV_CMD_PREFLIGHT_UAVCAN.param1=0 cancel an in progress configuration that is incomplete or does it do nothing?

pavel-kirienko:
I don't know for sure since I didn't write the code, but I suspect that zero does nothing. I would recommend looking at the UAVCAN driver sources.

DonLakeFlyer:
This: https://github.com/PX4/Firmware/blob/master/src/drivers/uavcan/uavcan_servers.cpp#L481 points to the fact that param1 does support both 0 and 1 as values. Not only that there is something connected to param2 as well. What it all means don't know.

hamishwillee:
@DonLakeFlyer @pavel-kirienko I read that code as a fall though - yes it accepts 0 or 1, but it does the same thing in either case. Not clear what would happen if you sent it twice - either it would do the action twice or return a failed. I expect your code might be just ignoring that fail.

Upshot, the message is probably correct. If we were to change it I'd propose we just made the clarification that it applies only to UAVCAN v0 - ie description to something like

> Trigger UAVCAN v0 enumeration (actuator ID assignment and direction mapping). This command will be only accepted if in pre-flight mode. Does not affect UAVCAN v1 and later.

OK?

pavel-kirienko:
OK. Is it an option to rename the command to `UAVCAN_ENUMERATE` though? The current name is misleading.

hamishwillee:
Not really/not worth the effort. We can add the information in the description.

hamishwillee:
@DonLakeFlyer @pavel-kirienko How about this? 
- It reverts the changes - this is trigger, not start/stop
- Even though it will also do the same thing for param1= 0, this only mentions 1 - the "0" is kind of "out of spec".
- Added note that this is misnamed, and that it should only be used once during vehicle setup.

OK?

pavel-kirienko (approved):

DonLakeFlyer:
It seems oddly specific:
* Talking about only available during pre-flight for example could be added to tons of other commands as well
* Talking about it only being used by a vendor or maintenance tech is also strange. Wouldn't a DIY'er building their own drone use this as well?

Also it still seems that param1=0 cancels the enumeration if you look at the code. Param 1 is stuffed into a variable which seems to indicate running or not: https://github.com/PX4/Firmware/blob/master/src/drivers/uavcan/uavcan_servers.cpp#L491

hamishwillee:
@DonLakeFlyer YOu make good points, but that does not stop me feeling like I am wasting my life :-(. 

Would the following satisfy you and @pavel-kirienko? 
```
      <entry value="243" name="MAV_CMD_PREFLIGHT_UAVCAN" hasLocation="false" isDestination="false">
        <description>Trigger UAVCAN configuration (actuator ID assignment and direction mapping).  Note: This maps to the legacy UAVCAN v0 function UAVCAN_ENUMERATE, which is intended to be executed just once during initial vehicle configuration.</description>
        <param index="1" label="Actuator ID">1: Trigger actuator ID assignment and direction mapping. 0: Cancel command.</param>
        <param index="2">Reserved</param>
        <param index="3">Reserved</param>
        <param index="4">Reserved</param>
        <param index="5">Reserved</param>
      </entry>
```

Note, 
- yes, the invention of cancellation via sending zero is "new" but is is not a "break" since a system that does not implement this will just continue the operation through to completion
- I actually like making it clear when a message can be called (i.e. in preflight) and I'd see those other commands as a bit defective. Making it clear that the intent is that this called is in vehicle bring-up does the same thing for me.

pavel-kirienko:
I am fine with the revised description if we return the part about poor naming:

> (i.e. it is not a normal pre-flight command, and has been poorly named)

Whether this issue applies to other commands is irrelevant, it is an issue nevertheless and it should be fixed.

hamishwillee:
@pavel-kirienko The poor naming thing is correct, but it isn't generally necessary to state that once you know this is based on UAVCAN_ENUMERATE. I.e. what would a user do knowing that it is poorly named. Then we have to answer - why can't you just change it etc etc. Upshot, I would not bother adding that back. 

And I agree that if it is useful to know that something can only be used in a certain way, then whatever other commands document would not affect doing the right thing here. We're after continuous improvement.

pavel-kirienko:
> what would a user do knowing that it is poorly named.

I think the remark about the poor name is helpful because it eliminates the confusion caused by the conflict between the name and the description: the name says that the command triggers a preflight activity (because it has `PREFLIGHT` in the name); the description says the opposite ("intended to be executed *just once during initial vehicle configuration*").

hamishwillee:
OK, updated as per back-and-forth. @DonLakeFlyer @pavel-kirienko If there are no final comments, I plan to merge this shortly.

hamishwillee:
Merging. Propose new PR if you're not happy.

PR #1460, opened 2020-08-25
COMPONENT_INFORMATION: Parameter meta data - Support indexed naming

DonLakeFlyer:
* Parameter name can be in format `FOO<#>_BAR` which will match `FOO1_BAR`, `FOO_6_BAR` and so forth
* Reduces duplication and file size

### DIFF
--- a/component_information/parameter.schema.json
+++ b/component_information/parameter.schema.json
@@ -26,7 +26,8 @@
                 "name": {
                     "description":  "Parameter Name.",
                     "type":         "string",
-                    "pattern":      "^[a-zA-Z0-9_]{1,16}$"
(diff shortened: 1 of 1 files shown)


hamishwillee:
So what does the XML look like for FOO_6_BAR? i.e. is it just that you use `FOO_<#>_BAR` in name field and when GCS sees anything in that format it associates the metadata?

What about instance info in the description - ie I might need to have "This is foo <#1> that does X, Y, Z"

DonLakeFlyer:
> is it just that you use `FOO_<#>_BAR` in name field and when GCS sees anything in that format it associates the metadata?

Correct.

> What about instance info in the description - ie I might need to have "This is foo <#1> that does X, Y, Z"

That is correct as well.

PR #1463, opened 2020-08-26
Update common.xml to import minimal

hamishwillee:
This includes minimal.xml into common.xml. I removed the xml entities from common that were already in minimal.xml and merged all changes back to common.xml.

Open question - should the MAV_CMDs be in minimal? The argument "for" is that these are required to check the version of mavlink you are really talking to. The argument against is that you can't use them without the command protocol. My feeling is that these should only be in common.

@LorenzMeier @julianoes @peterbarker @auturgy FYI


peterbarker:
@hamishwillee in the interests of improving structure, could we leave get-message and protocol-version in common.xml for the time being, please?

hamishwillee:
> in the interests of improving structure, could we leave get-message and protocol-version in common.xml for the time being, please?

Not sure how this would "improve structure". Let's see how the discussion falls out - but they are in there currently so if no one has strong feelings I will just leave them there.

peterbarker:
On Tue, 25 Aug 2020, Hamish Willee wrote:

>       in the interests of improving structure, could we leave get-message and protocol-version in common.xml for the time being, please?
> 
> Not sure how this would "improve structure". Let's see how the discussion falls out - but they are in there currently so if no one has strong feelings I
> will just leave them there.

Who implements PROTOCOL_VERSION?

To be clear - I'd like minimal.xml to be inheritted, and that's the 
structural improvement I was refering to.

AFAICS, HEARTBEAT and its associated enums are absolutely uncontroversial 
as minimal.xml entries.

hamishwillee:
Don't know who implements PROTOCOL_VERSION? It should be automatic.

Minimal.xml is inherited by this PR. My concern is that as-is it includes commands, so to fully implement a system you would have to also implement the command protocol - making it really needed for minimal. I guess the first thing to decide is how essential these really are.

peterbarker:
On Tue, 25 Aug 2020, Hamish Willee wrote:

> Don't know who implements PROTOCOL_VERSION? It should be automatic.
> 
> Minimal.xml is inherited by this PR. My concern is that as-is it includes commands, so to fully implement a system you would have to also implement the
> command protocol - making it really needed for minimal. I guess the first thing to decide is how essential these really are.

Well, given that ArduPilot doesn't implement PROTOCOL_VERSION, it's 
definitely not essential :-)

Technically get-message wouldn't need to be minimal either if there was a 
requirement for PROTOCOL_VERSION to be emitted in the same way that 
heartbeat is.  Not that I would advocate for that.

hamishwillee:
OK. I'll merge as-is. Might get moved. Will ask in dev call this week for opinions.

peterbarker:
@hamishwillee perhaps I wasn't clear?  `PROTOCOL_VERSION` has no business being in `minimal.xml` at this point.

auturgy:
Please revert.  This needs to be all, but only, those things necessary to participate in a mavlink system.  PROTOCOL_VERSION isn't used for version negotiation, so one really wonders what it is used for....

hamishwillee:
@auturgy We can discuss on Wednesday. This reflects what was in minimal.xml at for the last whatever months, and while I personally agree it should be out I only want to change it once at most more.

hamishwillee:
And assuming nesting is all working, changing it later or now should have no impact on end users :-)

PR #1419, opened 2020-07-02
common: add intent flags

julianoes:
@olliw42 discovered that the current gimbal v2 proposal still does not properly address deconfliction of various sources. The rule was to favor commands over the message streamed at a certain rate. This worked for the case where manual control is overridden by a command as part of a mission, however, it does not support e.g. a smart shot being sent from a companion computer using the message (and not the command).

Therefore, the idea here is to add flags for the intent of the source.
This then lets the gimbal manager decide which intent it wants to allow at any given time. The default order of the intents would be:
PILOT
GIMBAL_OPERATOR
SMARTSHOT
MISSION
TEST
CUSTOM1
CUSTOM2
with PILOT being the lowest and CUSTOM2 the highest "priority". However, a gimbal manager is allowed to be configured differently depending on the use case.

Clients sending gimbal control or nudging messages or commands are supposed to set NUDGE or CONTROL if they want to start nudging or controlling and follow up with at least one message with both NUDGE and CONTROL unset in order to give up their control.

This proposal is by no means perfect but I argue it's better than the current state and should enable many use cases, as well as make the deconfliction rules easier to grasp.

Somewhat resolves #1380

FYI @olliw42

PR #1466, opened 2020-09-01
Gimbal v2 changes

julianoes:
This is a collection of changes/improvements on the initial gimbal v2 spec.
The changes mostly come out of the discussion with @olliw42 in #1419 and through experience after some implementation of the v2 protocol.

Basically, we found that the deconfliction/prioritization rules were not very clear and not covering more than a few specific cases. After some more iterations on how this could be done, @olliw42 and I gave up on such a design partly because we couldn't agree on the specifics and partly because it seemed to get very complex for not much benefit. We therefore went back to a similar approach as it was before with a configure message, although somewhat different. Instead of setting up the mode of the gimbal, we now set the sysid/compid of the component in control. The component can then use whichever messages and mode it needs. Additionally, there can be secondary input from a component for things like nudging, reposition, etc..

What we also determined was that for the most part all that's required is really just manual control without a unit, so basically just pan and tilt from -1 to 1, either in angle mode or angular rate mode. The gimbal can use whatever smoothness/speed makes sense, or is configured.

Note that the concept of gimbal manager and device stays mostly the same as before.

For more detail, check the individual commits.

Further todos:
- [x] Clean up flags
- [x] Check status and information messages
- [x] Update docs
- [ ] Figure out remaining questions in #1419.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -460,15 +460,6 @@
       <entry value="131072" name="GIMBAL_MANAGER_CAP_FLAGS_CAN_POINT_LOCATION_GLOBAL">
         <description>Gimbal manager supports to point to a global latitude, longitude, altitude position.</description>
       </entry>
-      <entry value="1048576" name="GIMBAL_MANAGER_CAP_FLAGS_SUPPORTS_FOCAL_LENGTH_SCALE">
-        <description>Gimbal manager supports pitching and yawing at an angular velocity scaled by focal length (the more zoomed in, the slower the movement).</description>
-      </entry>
-      <entry value="2097152" name="GIMBAL_MANAGER_CAP_FLAGS_SUPPORTS_NUDGING">
-        <description>Gimbal manager supports nudging when pointing to a location or tracking.</description>
-      </entry>
-      <entry value="4194304" name="GIMBAL_MANAGER_CAP_FLAGS_SUPPORTS_OVERRIDE">
-        <description>Gimbal manager supports overriding when pointing to a location or tracking.</description>
-      </entry>
     </enum>
     <enum name="GIMBAL_DEVICE_FLAGS" bitmask="true">
       <description>Flags for gimbal device (lower level) operation.</description>
@@ -505,16 +496,7 @@
       <entry value="16" name="GIMBAL_MANAGER_FLAGS_YAW_LOCK">
         <description>Based on GIMBAL_DEVICE_FLAGS_YAW_LOCK</description>
       </entry>
-      <entry value="1048576" name="GIMBAL_MANAGER_FLAGS_ANGULAR_VELOCITY_RELATIVE_TO_FOCAL_LENGTH">
-        <description>Scale angular velocity relative to focal length. This means the gimbal moves slower if it is zoomed in.</description>
-      </entry>
-      <entry value="2097152" name="GIMBAL_MANAGER_FLAGS_NUDGE">
-        <description>Interpret attitude control on top of pointing to a location or tracking. If this flag is set, the quaternion is relative to the existing tracking angle.</description>
-      </entry>
-      <entry value="4194304" name="GIMBAL_MANAGER_FLAGS_OVERRIDE">
-        <description>Completely override pointing to a location or tracking. If this flag is set, the quaternion is (as usual) according to GIMBAL_MANAGER_FLAGS_YAW_LOCK.</description>
-      </entry>
-      <entry value="8388608" name="GIMBAL_MANAGER_FLAGS_NONE">
+      <entry value="1048576" name="GIMBAL_MANAGER_FLAGS_NONE">
         <description>This flag can be set to give up control previously set using MAV_CMD_DO_GIMBAL_MANAGER_ATTITUDE. This flag must not be combined with other flags.</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1718,17 +1718,17 @@
         <param index="3" label="Tilt angle" units="deg" minValue="-180" maxValue="180">Tilt/pitch angle (positive to tilt up, relative to vehicle for PAN mode, relative to world horizon for HOLD mode).</param>
         <param index="4" label="Pan angle" units="deg" minValue="-180" maxValue="180">Pan/yaw angle (positive to pan to the right, relative to vehicle for PAN mode, absolute to North for HOLD mode).</param>
         <param index="5" label="Gimbal manager flags" enum="GIMBAL_MANAGER_FLAGS">Gimbal manager flags to use.</param>
-        <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. (Send command multiple times for more than one but not all gimbals.)</param>
+        <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one but not all gimbals.</param>
       </entry>
       <entry value="1001" name="MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE" hasLocation="false" isDestination="false">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Gimbal configuration to set which sysid/compid is in primary and secondary control.</description>
-        <param index="1" label="sysid primary control">Sysid currently in primary control (0: no one in control).</param>
-        <param index="2" label="compid primary control">Compid currently in primary control (0: no one in control).</param>
-        <param index="3" label="sysid secondary control">Sysid currently in secondary control (0: no one in control).</param>
-        <param index="4" label="compid secondary control">Compid currently in secondary control (0: no one in control).</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6187,6 +6183,31 @@
       <field type="float" name="tilt_rate" units="rad/s">Tilt/pitch angular rate (positive: up, negative: down, NaN to be ignored).</field>
       <field type="float" name="pan_rate" units="rad/s">Pan/yaw angular rate (positive: to the right, negative: to the left, NaN to be ignored).</field>
     </message>
+    <message id="288" name="GIMBAL_MANAGER_SET_MANUAL_CONTROL">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>High level message to control a gimbal manually, so without units. The actual angles or angular rates will be produced by the gimbal manager based on settings. This message is to be sent to the gimbal manager (e.g. from a ground station). Angles and rates can be set to NaN according to use case.</description>
```
hamishwillee, line 6189: "based on settings". By settings, do you mean "settings that are outside of the scope of MAVLink"? I.e. I think you're just saying that "the rates depend on gimbal hardware and configuration - so you can't set actual rates, just as a proportion of the total.

I guess also that the implementation might choose to not do this proportionally - but follow some rate curve based on the value.

message_definitions/v1.0/common.xml
```diff
@@ -6187,6 +6183,31 @@
...
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags to use.</field>
```
hamishwillee, line 6192: " to use" is redundant

message_definitions/v1.0/common.xml
hamishwillee, line 6189: ", so without units" does not follow the "High level message to control a gimbal manually" logically. ie just because it is a high level message that does not mean that it has to be unitless. 

Perhaps "High level message to control a gimbal manually. The angles or angular rates are unitless; the actual rates will depend on internal gimbal manager settings/configuration. ...

message_definitions/v1.0/common.xml
```diff
@@ -6187,6 +6183,31 @@
...
+      <field type="float" name="pan">Pan/yaw angle unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
+      <field type="float" name="tilt_rate">Tilt/pitch angular rate unitless (-1..1, positive: up, negative: down, NaN to be ignored).</field>
+      <field type="float" name="pan_rate">Pan/yaw angular rate unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
+    </message>
+    <message id="289" name="GIMBAL_DEVICE_SET_MANUAL_CONTROL">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Low level message to control a gimbal manually, so without units. The actual angles or angular rates will be produced by the gimbal manager based on settings and by the gimbal device if available. This message is sent by the gimbal manager to the gimbal device. Angles and rates can be set to NaN according to use case. The message should not be sent at the same time as GIMBAL_DEVICE_SET_ATTITUDE, otherwise the two setpoints would conflict.</description>
```
hamishwillee, line 6202: Same comment about "so without units" - that bit belongs in the next sentence.

message_definitions/v1.0/common.xml
```diff
@@ -1732,6 +1714,16 @@
...
       </entry>
+      <entry value="1001" name="MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Gimbal configuration to set which sysid/compid is in primary and secondary control.</description>
+        <param index="1" label="sysid primary control">Sysid currently in primary control (0: no one in control).</param>
+        <param index="2" label="compid primary control">Compid currently in primary control (0: no one in control).</param>
+        <param index="3" label="sysid secondary control">Sysid currently in secondary control (0: no one in control).</param>
+        <param index="4" label="compid secondary control">Compid currently in secondary control (0: no one in control).</param>
+        <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. (Send command multiple times for more than one but not all gimbals.)</param>
```
hamishwillee, line 1725: "components. (Send command multiple times for more than one but not all gimbals.)" to either
- "components (send command multiple" or 
- "components. Send command multiple times"

message_definitions/v1.0/common.xml
```diff
@@ -1732,6 +1714,16 @@
         <param index="5" label="Gimbal manager flags" enum="GIMBAL_MANAGER_FLAGS">Gimbal manager flags to use.</param>
         <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. (Send command multiple times for more than one but not all gimbals.)</param>
       </entry>
+      <entry value="1001" name="MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Gimbal configuration to set which sysid/compid is in primary and secondary control.</description>
+        <param index="1" label="sysid primary control">Sysid currently in primary control (0: no one in control).</param>
```
hamishwillee, line 1721: For all of these do you mean "currently in" or "Sysid to take primary control"

julianoes: It means that this component is in control (not to take). It means that messages arriving from that sysid/compid will be used/applied.

julianoes:
Thanks for the review @hamishwillee. I'll do a pass later.

message_definitions/v1.0/common.xml
hamishwillee: @julianoes English doesn't make sense (to me). As I understand it the  ownership of the gimbal and who is the primary and secondary is managed by the gimbal manager. So the manager broadcasts the current controller(s) using GIMBAL_MANAGER_STATUS. Another system can set itself as the "current controller" using *this* message MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE.

The point is that this is a setter so you probably shouldn't use "currently" because it makes it sound like the message is a getter - getting the current controller id. Also it is redundant.

Suggested changes:
- "Sysid currently in primary control (0: no one in control)." TO "Sysid for primary control (0: no one in control)"
- Same for the other fields.
- In the description, it is worth saying here something like "Check current controllers in GIMBAL_MANAGER_STATUS before arbitrarily seizing control. For more info see ....)

julianoes: Oh, sorry, you're right, thx!

message_definitions/v1.0/common.xml
julianoes: Yes, it would bee based on settings outside of MAVLink, e.g. params.

julianoes: Thx, yes!

julianoes:
Thanks @hamishwillee. I've added one commit trying to address your review comments.

hamishwillee:
@julianoes Much better thanks. I did a minor subedit for consistency. 

The only concern I have may well be resolved by the other docs - still don't understand the model for primary and secondary control. But aside from that this is good to go (feel free to self-merge)

julianoes:
> still don't understand the model for primary and secondary control.

That's up to the implementation and not really specified by the protocol. For some use cases there will only be a primary, and for others there might be a primary e.g. for the general angle and a secondary for slight nudging, and others might just add the two sources together.

hamishwillee:
> That's up to the implementation and not really specified by the protocol. For some use cases there will only be a primary, and for others there might be a primary e.g. for the general angle and a secondary for slight nudging, and others might just add the two sources together.

Not just that - see my comments in the docs PR. Specifically what happens if multiple things try to set these and the interactions. Can I set myself as the secondary if something else has set itself as primary etc. Who owns what.

julianoes:
> Can I set myself as the secondary if something else has set itself as primary etc. Who owns what.

Yes, anyone can change it as they want. I think that's something that we should explain in the FAQ part of the docs.

julianoes:
I have come to the conclusion - at least for now - that we should only have the manual control message from ground station/companion/other clients to the gimbal manager, however, not from the gimbal manager to the gimbal device.

Here is my reasoning:
- The gimbal device does not know how to mix an angle setpoint and a manual control setpoint (e.g. nudging). This is rather something that can/should be done by the gimbal manager.
- The gimbal device would get confused if there are attitude and manual control messages arriving at the same time. It's not specified what should happen then. If possible, I would like to prevent that case by protocol design.
- One of advantages of manual control is that the pan/tilt speed can get adjusted to a camera zoom level so that panning is nice and slow when zoomed in. This would happen in the gimbal manager because a gimbal with integrated camera is typically the gimbal manager (and not device).

What's not so clear?
- The gimbal device knows its limits/settings like angular rate and smoothness best. It does communicate max angles and max angular velocities using GIMBAL_DEVICE_INFORMATION which allows the gimbal manager to translate manual control setpoints to angular velocities, or angles, but it doesn't seem necessarily seem optimal.

I'm going to merge this PR now and work on the implementation. Once I find how it does or does not work, I'll come back to the protocol which is still marked WIP.

PR #1468, opened 2020-09-01
Add ID field to HIL_SENSOR for multisensor  support

Jaeyoung-Lim:
This PR adds a ID field as an extension to the `HIL_SENSOR` message. This enables streaming multiple instances of `HIL_SENSOR` data so that we can simulate in HIL/SITL multi IMU/Airspeed/Barometer sensor scenarios.

I believe simulating multiple sensors have become of interest for various reasons and PX4 already has support for multi airspeed sensors. For me it doesn't make sense that this cannot be simulated because the message definition is unable to represent multiple instances, hence the extension.

The same approach is already being utilized in the `HIL_GPS` message for simulating multiple GPS sensors.

**Alternative approach**
An alternative approach would be to actually redefine the `HIL_SENSOR` message so that we have separate messages for each sensors. However, I think this might bring huge regression problems and a synchronous updates of each sensors are already being nicely handled with the `fields_updated` field. 

@TSC21 @RomanBapst FYI


TSC21 (approved):
Looks good. My only concern will be regarding the lockstep feature and the corner cases around it. But I guess that's a concern for the autopilot implementation,

LorenzMeier (approved):
This looks good and is logical. I waited for the dev call initially, but this should go in now.

PR #1469, opened 2020-09-02
Add some "missing" fields to AUTOPILOT_VERSION, CAMERA_INFORMATION

olliw42:
The current COMPONENT_INFORMATION has become pretty much useless for typical components or embedded gcses. However, info such as vendor, model, firmware version, and hardware version are quite useful to the user even for typical components. The AUTOPILOT_VERSION message is the typical component's best bet to provide it, and CAMERA_INFORMATION for cameras.

Unfortunately, some of these info is missing in each of them, and this PR is attempting to make them more complete and consistent as regards these info. That is, the two messages are extended such that they each provide these four items
* vendor name
* model name
* firmware version (aka flight_sw_version in AUTOPILOT_VERSION)
* hardware version (aka board_version in AUTOPILOT_VERSION)

For AUTOPILOT_VERSION this is a quite natural extension with essentially no consequences, and thus should be agreeable.

For CAMERA_INFORMATION it would be the first extension, and thus has to follow the large uri field (of type char[140]). It thus breaks the effectiveness of zero-byte truncation. However, CAMERA_INFORMATION is exchanged only very rarely, and I thus argue that zero-byte truncation is not really of importance here. I thus think that the extension should also be agreeable.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5544,6 +5544,8 @@
       <field type="uint64_t" name="uid">UID if provided by hardware (see uid2)</field>
       <extensions/>
       <field type="uint8_t[18]" name="uid2">UID if provided by hardware (supersedes the uid field. If this is non-zero, use this field, otherwise use uid)</field>
+      <field type="uint8_t[32]" name="vendor_name">Name of vendor</field>
+      <field type="uint8_t[32]" name="model_name">Name of model</field>
     </message>
     <message id="149" name="LANDING_TARGET">
       <description>The location of a landing target. See: https://mavlink.io/en/services/landing_target.html</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5851,6 +5851,8 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration)</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol).</field>
+      <extensions/>
+      <field type="uint32_t" name="hardware_version">Version of the camera hardware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff)</field>
     </message>
     <message id="260" name="CAMERA_SETTINGS">
       <description>Settings of a camera. Can be requested with a MAV_CMD_REQUEST_MESSAGE command.</description>


message_definitions/v1.0/common.xml
```diff
@@ -5849,6 +5851,8 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration)</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol).</field>
+      <extensions/>
+      <field type="uint32_t" name="hardware_version">Version of the camera hardware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff)</field>
```
hamishwillee, line 5855: @dogmaphobic Can you please review this?

olliw42:
ping :)

message_definitions/v1.0/common.xml
hamishwillee: @olliw42 What would you do with the information, and expect the recipient to do with the information. Note, I'm just trying to get arguments together so this can be discussed.

olliw42: That's a bit of a tricky question, I guess. I think one could with some justification generally ask if metadata such as firmware and/or hardware version is needed, and so on. I'm not sure I would want to go into such a discussion. I think anyone of us easily can produce plenty of use cases there such info would be helpful, but I don't see value in this. So, the part concerning what I actually would do or expect anyone to do with it I deliberately leave unanswered.

My line of reasoning is much simpler: 

1. It obviously was anticipated by the original creator(s) of these messages that these kind of metadata can have some value. This I use to bypass the said discussion and just accept the result as a given. 

2. I cannot see any convincing argument which would say that the one type of metadata is generally more or less important for an autopilot than for a general component. From this I conclude that it should be the superset of metadata types for both autopilot and components. This makes me conclude that for the camera the said field is missing.

For the moment I'm actually not doing anything spectacular with these information: I'm displaying it. But this hardly can be relevant for the discussion.

hamishwillee: Hi @olliw42,
Thanks very much. 

My take on this:
- for the camera the things that the software will care about are all defined in the camera_uri - ie the hardware version specific handling is all abstracted for the software into the definition fie.. 
- for AUTOPILOT_VERSION the fields should be inferrable from the board_version, vendor and product id. 

I could be wrong, but my concern is that it is very difficult to get changes in where there is no stakeholder pull or a firm use case.  So even if I'm wrong in my opinion, I'd rather close this until someone says that they really want it. Happy to raise in the dev call.

olliw42: > my concern is that it is very difficult to get changes in where there is no stakeholder pull or a firm use case. So even if I'm wrong in my opinion, I'd rather close this until someone says that they really want it. 

I have no argument to counter this argument. It's a policy decision.

As regards the camera I find it highly illogical to argue that it needs a software version field but not a hardware version field since it can be read of the xml file. The same argument applies to the software version field too. Doesn't make sense to me, frankly. But I don't have a very strong opinion on the camera thing, so ok.

As regards the autopilot message I think you forgot about the first paragraph of the first post, which is really the key point of the PR. One implication is that there most likely never ever would be any vendor and product ids for typical components. 

As side note, I btw think that it would be worthwhile to think a bit deeper about what this first paragraph means on a more general scale, but that's not the topic of this PR.

As another side note, I'd like to say that IMHO you are really moving in a unfortunate direction for some pieces (not you personally, but you as the mavlink group). You are repeating making the same mistake which has created quite some not-so-nice aspects in mavlink (attention: British understatement), namely to not really realize that there are not only autopilots but also components. You want to move in that directions, ok, your decision, but I think typical components nevertheless should get at least a bit of a fair share in the game. 

So, as regards the suggested autopilot version extensions I actually have kind of clear opinion :)

olliw42:
have seen the dev call's conclusion, so close it :)
but many thanks for considering it
I guess typical components continue to have a hard time ;)

hamishwillee:
> I guess typical components continue to have a hard time ;)

@olliw42 You are right - we are focused on the autopilot, and this has on occasion been to the detriment of both components and GCS.

We do our best, but if something is blocking you come along to the devcall and defend it. You will do a much better job of presenting the affect of particular decisions that I will.

olliw42:
thx for your kind response. I definitely didn't wanted to put in doubt that you guys wouldn't try your best.

but I frankly don't feel like I want to defend anything ...

I guess my ultimate hope is that the new governance policy will allow message id spaces for projects like the STorM32 ... I guess it won't, but I still can hope, right LOL

thx again for your efforts

Issue #1464, opened 2020-08-31
MAVLink docs: incomplete for common


olliw42:
minimal.xml is now included in common.xml, but the respective messages, enums and so on are NOT also listed on the documents page
e.g, an important message like HEARTBEAT is now not listed anymore :)

hamishwillee:
https://mavlink.io/en/messages/minimal.html#HEARTBEAT

Common includes minimal.xml and this is listed. Are you suggesting that the dialect files should be auto-expanded out? I'm somewhat tempted by the idea.

olliw42:
> Are you suggesting that the dialect files should be auto-expanded out?

yes

I use the docs very often, and find it very unfortunate and less practical as it is right now.

hamishwillee:
Fair enough. I'll have a look at the docs generator in the next week or so. I think I can do a relatively easy hack.

olliw42:
great!

I realize I might have been a bit unclear. I guess I'm not actually suggesting to generally expand out the included xmls, i.e., also for dialects (dialects in the sense of the one google doc ). I suggest so only for standard and common (and other similar, if there ever are other similar). For dialects I guess I would handle it as you currently do it, i.e., just show the additions by that dialect.

hamishwillee:
@olliw42 Thanks. I misinterpreted you, but it doesn't matter because the dev call came to the same conclusion :-). I created a fix but waiting on review of the "approach".

olliw42:
this seems to have addressed to some reasonable extend, so I close.

PR #1470, opened 2020-09-03
Docs generator include minimal in common

hamishwillee:
Devcall discussed whether it makes sense to show included definitions in parent dialects and files (e.g. show minimal.xml definitions in the docs for common.xml) as requested in #1464. General feeling was that dialects didn't want this, but we do want it for common.xml.

This is a very simple implementation that simply appends the minimal.xml definitions to the end of common docs:
image

This is mentioned/forward linked from the introduction.
image

This also updates the text for the minimal.xml doc file, which was just referred to as a dialect. This now looks like:
image

Note, a nicer implementation would be to merge the two sets and add output to indicate which set the entities come from. Decided that is way too much effort for now. Might make sense to do that eventually as a new output format from the docs generator.


hamishwillee:
Self merging. I'm pretty sure this breaks nothing, and it can be iterated if needed.

PR #1472, opened 2020-09-05
move MAG_CAL_REPORT and MAG_CAL_STATUS ardupilotmega -> common

dagar:


dagar:
Looks like we can't get away with this one. Suggestions?

``` Console
ERROR: Duplicate message id 192 for MAG_CAL_REPORT (message_definitions/v1.0/common.xml:5600) also used by PTZ_STATUS (message_definitions/v1.0/slugs.xml:266)
```

auturgy:
The generator should overwrite common with the dialect, and as slugs aren’t using the ardupilotmega message I don’t think this conflict actually matters. But we need to test that. Is slugs a live dialect?

dagar:
>  Is slugs a live dialect?

Seems unlikely, the last real commit was 2014. https://github.com/mavlink/mavlink/pull/175

LorenzMeier:
It has to go: https://github.com/mavlink/mavlink/pull/1473

hamishwillee:
Slugs deleted which fixes above issue. There is no change to the message so no reason to block this. Merged.

PR #1473, opened 2020-09-06
Delete slugs.xml

LorenzMeier:
This is an unused dialect, with the last relevant commit six years ago. It is cluttering the standards repo and should be maintained downstream if desired.


dagar (approved):

hamishwillee:
Yes. I might argue with this approach except worse case the "owner" will get back to us and we'll at least have someone we can engage with on this. Merging.

hamishwillee:
PS/FWIW @LorenzMeier @auturgy We should add simple rules on addition/removal of dialects in the governance draft so this is just a part of the maintenance process. Though of course with all.xml the chance of clashes in the first place will be much lower.

PR #1478, opened 2020-09-15
MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE: clarify param4 values

hamishwillee:
QGC is using NaN rather than 0 to indicate "not valid" in [MAV_CMD_IMAGE_START_CAPTURE.param4](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE) (sequence number) when param3 != 1. This works with PX4 and as it is clearly an invalid value I would expect a system getting this to handle it. 

What this change does is explicitly allow the NaN case. 

I also took opportunity to mark up the other parameters as reserved with an NaN value using the expected syntax.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1736,8 +1736,8 @@
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
         <param index="1">Reserved (Set to 0)</param>
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
-        <param index="3" label="Capture Count" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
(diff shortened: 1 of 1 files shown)


hamishwillee:
From Don

> I'm guessing when param3 == 1 px4 firmware will puke on param4 = NaN to ignore instead of 0. Hmm, which in turn I think means QGC should really be changed. Otherwise the spec just gets too polluted. The spec should probably update to say somehow 0 is used to not only specify no camera sequence number specified but also used when param4 is not used at all.

hamishwillee:
@DonLakeFlyer 

>  I'm guessing when param3 == 1 px4 firmware will puke on param4 = NaN to ignore instead of 0. Hmm, which in turn I think means QGC should really be changed. Otherwise the spec just gets too polluted. The spec should probably update to say somehow 0 is used to not only specify no camera sequence number specified but also used when param4 is not used at all.

And so it should - **either** NaN or 0 are invalid if param3 ==1. 

Even so, I've modified to say "always set to zero unless `param3 == 1`. If you're not happy with this, propose a change.

```
<param index="4" label="Sequence Number" minValue="1" increment="1">
Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1), otherwise set to 0. 
Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted.</param>
```

DonLakeFlyer:
I think that's good. Thanks @hamishwillee

hamishwillee:
Self merging. This does not change the meaning, just clarifies the value.

PR #1479, opened 2020-09-16
common: Add a DECAROTOR type

muramura:
I would add a 10-rotor helicopter with ArduPilot.
I have added a type for this.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1,8 +1,10 @@
 <?xml version="1.0"?>
 <mavlink>
+  <include>minimal.xml</include>
   <version>3</version>
   <dialect>0</dialect>
   <enums>
+<!--
     <enum name="MAV_AUTOPILOT">
       <description>Micro air vehicle / autopilot classes. This identifies the individual model.</description>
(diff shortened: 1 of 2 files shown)


hamishwillee:
@muramura This has conflicts because MAV_TYPE was removed from common.xml. We now just have it in minimal.xml and include minimal.xml into common.xml.

Can you please rebase to fix the conflict.

Otherwise sounds good to me - @auturgy @julianoes any comment (this is just addition of a deca-rotor type (we already have a **do**decorator, but this one is new).

muramura:
@hamishwillee san.   I put the common.xml back in.
I had assigned 34 in common.xml MAV_TYPE in common.xml.
I assigned 35 to DECAROTOR.

hamishwillee:
Hi @muramura 

Thanks. Unfortunately there still appears to be some conflict with minimal and it won't let me rebase. Can you rebase. Please use whatever the next free id is for the MAV_TYPE.

muramura:
I've included minimal.xml in common.xml.
I was able to verify that mavgenerate.py was successful.

Screenshot from 2020-09-19 15-40-51

muramura:
I PR'd again in #1483.

PR #1480, opened 2020-09-17
MAV_CMD_NAV_FENCE_RETURN_POINT - deprecate

hamishwillee:
As discussed in the dev call, `MAV_CMD_NAV_FENCE_RETURN_POINT` predates the use of rally points and is superseded by them. This change deprecates the command, and further clarifies that if a rally point is defined as well then that rally point would be used instead. 

FYI @DonLakeFlyer ,@meee1 
- We wont be removing this arbitrarily because we never remove things arbitrarily :-). But in particular, because this is still being used by some ArduPilot systems that won't update. 
- PX4 appears to support setting this through QGC, but it also appears to ignore the setting. Worth removing as an option in the UI IMO.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1915,6 +1915,7 @@
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="5000" name="MAV_CMD_NAV_FENCE_RETURN_POINT" hasLocation="true" isDestination="true">
+        <deprecated since="2020-09" replaced_by="MAV_CMD_NAV_RALLY_POINT">This mechanism was superseded by rally points. If rally points are defined then they take precedence.</description>
         <description>Fence return point. There can only be one fence return point.
         </description>
         <param index="1">Reserved</param>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Self merging if this passes the tests, as this was agreed in dev call by main stakeholders.

DonLakeFlyer:
> As discussed in the dev call, `MAV_CMD_NAV_FENCE_RETURN_POINT` predates the use of rally points and is superseded by them.

That is not quite the case. ArduCopter supports rally points whereas ArduPlane supported a single breach return point. This was here to replicate/support that difference. I don't know the exact state of ArduPilot support for the new fence protocol. If they have moved fully to rally points only this is fine. If not, this needs to stay.

hamishwillee:
@DonLakeFlyer There is no intent to remove this, just deprecate it. We're using deprecate to mean "superseded by". 

It shouldn't be enabled for PX4 though.

DonLakeFlyer:
You can't deprecate this if it is a needed existing ArduPilot feature. I don't myself understand whether rally points fully cover what breach return point was doing or not. ArduPilot folks would have to answer that. If that was already resolved/answered in the dev call then it's fine to deprecate.

hamishwillee:
@auturgy Can you please comment. In the dev call we id not discuss ArduPlane, which  may not support rally points.

Either way, we are taking deprecated to mean "there is something else that people should use instead of this now" - both GCS and flight stacks.

auturgy:
there's a PR in active development to add support for the new protocol.

https://github.com/ArduPilot/ardupilot/pull/15288

PR #1482, opened 2020-09-17
Command to allow geofence action to be set per-fence

hamishwillee:
MAVLink has a [FENCE_ACTION](https://mavlink.io/en/messages/common.html#FENCE_ACTION) that defines a number of possible actions on geofence breach. This is currently used only by ArduPilot to define the values for the system action on fence breach (it is used directly in a parameter).

This PR generalises the action, and defines a command that can be included in a geofence definition (as many times as you like) to allow the geofences in a plan to have separate actions assigned.

The message can also be used via the command protocol to set the global/system-default fence action.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2037,18 +2037,7 @@
         <param index="6" label="Longitude">Longitude</param>
         <param index="7">Reserved</param>
       </entry>
-      <entry value="5005" name="MAV_CMD_SET_FENCE_BREACH_ACTION" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Sets the action on geofence breach.
-          If sent using the command protocol this sets the system-default geofence action.
-          As part of a mission protocol plan it sets the fence action for the next complete geofence definition *after* the command.
-          Note: A fence action defined in a plan will override the default system setting (even if the system-default is `FENCE_ACTION_NONE`).
-          Note: Every geofence in a plan can have its own action; if no fence action is defined for a particular fence the system-default will be used.
-          Note: The flight stack should reject a plan or command that uses a geofence action that it does not support and send a STATUSTEXT with the reason.
-        </description>
-        <param index="1" label="Action" enum="FENCE_ACTION">Fence action on breach.</param>
-      </entry>
+      <!-- value 4005 reserved for MAV_CMD_SET_FENCE_BREACH_ACTION (see development.xml). -->
       <entry value="5100" name="MAV_CMD_NAV_RALLY_POINT" hasLocation="true" isDestination="false">
         <description>Rally point. You can have multiple rally points defined.
         </description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -314,20 +314,30 @@
     </enum>
     <!-- fenced mode enums -->
     <enum name="FENCE_ACTION">
+      <description>Actions following geofence breach.</description>
       <entry value="0" name="FENCE_ACTION_NONE">
-        <description>Disable fenced mode</description>
+        <description>Disable fenced mode. If used in a plan this would mean the next fence is disabled.</description>
       </entry>
       <entry value="1" name="FENCE_ACTION_GUIDED">
-        <description>Switched to guided mode to return point (fence point 0)</description>
+        <description>Fly to geofence MAV_CMD_NAV_FENCE_RETURN_POINT in GUIDED mode. Note: This action is only supported by ArduPlane, and may not be supported in all versions.</description>
       </entry>
       <entry value="2" name="FENCE_ACTION_REPORT">
         <description>Report fence breach, but don't take action</description>
       </entry>
       <entry value="3" name="FENCE_ACTION_GUIDED_THR_PASS">
-        <description>Switched to guided mode to return point (fence point 0) with manual throttle control</description>
+        <description>Fly to geofence MAV_CMD_NAV_FENCE_RETURN_POINT with manual throttle control in GUIDED mode. Note: This action is only supported by ArduPlane, and may not be supported in all versions.</description>
       </entry>
       <entry value="4" name="FENCE_ACTION_RTL">
-        <description>Switch to RTL (return to launch) mode and head for the return point.</description>
+        <description>Return/RTL mode.</description>
+      </entry>
+      <entry value="5" name="FENCE_ACTION_HOLD">
+        <description>Hold at current location.</description>
+      </entry>
+      <entry value="6" name="FENCE_ACTION_TERMINATE">
+        <description>Termination failsafe. Motors are shut down (some flight stacks may trigger other failsafe actions).</description>
+      </entry>
+      <entry value="7" name="FENCE_ACTION_LAND">
+        <description>Land at current location.</description>
       </entry>
     </enum>
     <enum name="FENCE_BREACH">
@@ -2026,6 +2036,18 @@
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
         <param index="7">Reserved</param>
+      </entry>
+      <entry value="5005" name="MAV_CMD_SET_FENCE_BREACH_ACTION" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Sets the action on geofence breach.
+          If sent using the command protocol this sets the system-default geofence action.
+          As part of a mission protocol plan it sets the fence action for the next complete geofence definition *after* the command.
+          Note: A fence action defined in a plan will override the default system setting (even if the system-default is `FENCE_ACTION_NONE`).
+          Note: Every geofence in a plan can have its own action; if no fence action is defined for a particular fence the system-default will be used.
+          Note: The flight stack should reject a plan or command that uses a geofence action that it does not support and send a STATUSTEXT with the reason.
+        </description>
+        <param index="1" label="Action" enum="FENCE_ACTION">Fence action on breach.</param>
       </entry>
       <entry value="5100" name="MAV_CMD_NAV_RALLY_POINT" hasLocation="true" isDestination="false">
         <description>Rally point. You can have multiple rally points defined.
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -44,7 +44,6 @@
         <description>Synthetic/calculated airspeed.</description>
       </entry>
     </enum>
-    
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -26,6 +26,25 @@
         <description>WPA3.</description>
       </entry>
     </enum>
+    <!-- The MAV_CMD enum entries describe either: -->
+    <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
+    <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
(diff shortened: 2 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1970,6 +1979,13 @@
         <param index="6" label="Longitude">Longitude</param>
         <param index="7">Reserved</param>
       </entry>
+      <entry value="5005" name="MAV_CMD_NAV_FENCE_BREACH_ACTION" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
```
hamishwillee, line 1984: @DonLakeFlyer @meee1 

So the design says that geofence applies to next fence defined in the plan and all the ones after that. So you'd apply it before start of each geofence. If you applied it within a geofence it would only work for the next new geofence, and might be replaced by another one inside the same geofence (last one wins). 

Questions:
1. Would it be better from a GCS point of view if the command applies just to the single next new geofence rather than all subsequent geofences. In other words the alternative model is that if you don't specifically assign a geofence for a each fence then the default will be used?
2. + @julianoes This says that if the vehicle does not understand the fence action it will reject the plan. Is this reasonable? Or should it fall back to the system default. My thinking is that as a user I would want to only see valid options, but if that was not possible I'd want to know that what I asked for was done. But I am not sure that this is as feasible as just using system default.

message_definitions/v1.0/common.xml
```diff
@@ -311,6 +311,7 @@
     </enum>
     <!-- fenced mode enums -->
     <enum name="FENCE_ACTION">
+       <description>Actions following geofence breach.</description>
       <entry value="0" name="FENCE_ACTION_NONE">
         <description>Disable fenced mode</description>
       </entry>
```
hamishwillee, line 317: @auturgy I'm trying to rationalise the modes.  What does "guided mode" mean in FENCE_ACTION_GUIDED and FENCE_ACTION_GUIDED_THR_PASS?

I know that QGC has the concept of guided meaning "a mode where I can set goto and orbit instructions" but in PX4 this actually maps to Hold mode. 

What I "think" this is saying is that the first one sets to the ArduPilot equivalent - a mode that holds and moves under GCS direction and the return point is set to the GCS fence point 0 (the center of the fence?).  The second one is the same except the RC throttle can be changed still.

DonLakeFlyer: The whole "guided" thing is mostly ardupilot terminology/support bleeding through. I don't think it should say or be named anything with "guided" since the concept of "guided" is not a mavlink thing but a firmware specific thing. I'd consider renaming.

Trying to go back in time and remember, these breach action I believe are basically a union of what arducopter/plane supported such that the new protocol could support what ardupilot already supported. Although I know parts or ardupilot support the new fence protocol I'm not sure that ArduPlane does yet. I'm also not sure what for example copter does with these breach actions. For example I though a fence breach actions would go to the closes rally point, not to fence point 0.

message_definitions/v1.0/common.xml
DonLakeFlyer: > 1. Would it be better from a GCS point of view if the command applies just to the single next new geofence rather than all subsequent geofences.

The fence breach spec is basically how ardupilot fence breach control worked from the previous ardupilot specific fence protocol. The support was global fence breach settings which is reflected in the new spec. I'm not convinced per-fence breach control adds value and in fact will make things worse from a standpoint of further complicating ui to create fences. Also you can't have per-fence breach control without also having per-fence rally point definition. I don't think it's rat hole you want to go down.

If someone thinks per-fence control is valuable then the discussion should start with why that is.

hamishwillee: > If someone thinks per-fence control is valuable then the discussion should start with why that is.

@DonLakeFlyer @LorenzMeier suggested this with what _seemed_ to me pretty broad support in the dev call across flight stacks and other GCS - @auturgy @meee1 etc.

A few ideas off the top of my head on the value - first geofence with a warning only, followed by a second fence with a hold. Then it is also likely you might have different actions for when you see a vehicle approaching (hold/land) vs entering restricted airspace. 

> Also you can't have per-fence breach control without also having per-fence rally point definition. I don't think it's rat hole you want to go down.

Can you explain that? Why do fences need different destinations just because they have different actions? By definition if the action is a return for any violated fence then the vehicle will go to the closest rally point or home or whatever else it's internal settings say. To me that is independent of the action setting.

I'm far more concerned about what a system should do if it doesn't support a particular GF action.

message_definitions/v1.0/common.xml
hamishwillee: @auturgy Do you think you can clarify what these do? I'd like to rename and describe them in a flight-stack agnostic way (if that makes sense when we know what they do).

message_definitions/v1.0/common.xml
DonLakeFlyer: It the mavlink folks have thought this through including things like a single set of rally points and so forth then I assume it's fine and useful.

Given that, what is the need for 'FENCE_ACTION_NONE'? Why would you add a `MAV_CMD_NAV_FENCE_BREACH_ACTION` at all just to set it to no action?

hamishwillee: @DonLakeFlyer The idea is clearly desirable. The rally point concern is not IMO an issue - the two things are mutually exclusive.

> Given that, what is the need for 'FENCE_ACTION_NONE'? Why would you add a MAV_CMD_NAV_FENCE_BREACH_ACTION at all just to set it to no action?

You wouldn't - I don't see a case where a mission would want to both declare a geofence and turn off any action for it. 

But we are starting from the ArduPilot geofence enum that predates the discussion and which was intended for setting the global fence action via a parameter. I haven't seen this before, but, as with the guided options, am trying to work out an approach that works without breaking anything legacy.

Upshot, annoying and unnecessary option in this context, but not one that breaks the GUI to have present right?

hamishwillee: So assuming this all makes sense, back to the original question - for a GCS UI is it better to assume that any setting applies for all future geofence definitions after declaration until replaced, or is it better to assume that each instance of this command applies just to the next fence?

DonLakeFlyer: > for a GCS UI is it better to assume that any setting applies for all future geofence definitions after declaration until replaced, or is it better to assume that each instance of this command applies just to the next fence?

I believe there are global fence breach settings which can be tweaked via parameters correct? Given that, a specific setting for fence breach pushed up with the plan can only apply to the next fence. Otherwise there is no way to create a set of fences, some which use global parameter based settings and some which don't.

DonLakeFlyer: > But we are starting from the ArduPilot geofence enum that predates the discussion and which was intended for setting the global fence action via a parameter.

Ah, ok. Understood.

hamishwillee: >> for a GCS UI is it better to assume that any setting applies for all future geofence definitions after declaration until replaced, or is it better to assume that each instance of this command applies just to the next fence?

> I believe there are global fence breach settings which can be tweaked via parameters correct? Given that, a specific setting for fence breach pushed up with the plan can only apply to the next fence. Otherwise there is no way to create a set of fences, some which use global parameter based settings and some which don't.

I will change it to say that then. Should we perhaps have and option "use system default" which makes it explicit?

Note though, that while yes they can be tweaked using parameters, this does not imply that you can't create a fence that mixes global and non-global params. For example, I might define an action "use system default" which would apply from insertion point on until it was replaced.

message_definitions/v1.0/common.xml
auturgy: https://ardupilot.org/plane/docs/parameters.html#fence-action-action-on-geofence-breach provides the descriptions, but basically fence_action_none disables the fence.

auturgy: Also worth linking this in: WIP PR for the newer fence protocol into ArduPilot Plane.
https://github.com/ArduPilot/ardupilot/pull/15288

hamishwillee:
@julianoes @auturgy @DonLakeFlyer 

I think this is nearly ready to go - can you give it a review?

The message sets the default system action if used in command protocol, and the action for the next fence if used in a plan. I have [stated that](https://github.com/mavlink/mavlink/pull/1482/files#diff-ceacc1e2390186650146a45be8fcb691R1989) the value in the mission plan overrides the system default **even if the default action is to turn off geofences**. Thoughts?

Otherwise, the FENCE_ACTION just extends the options from ArduPilot with those from PX4. A system is expected to reject a command/plan that sets a value it does not support.

hamishwillee:
@DonLakeFlyer @julianoes Reminder - needs a sanity check.

message_definitions/v1.0/common.xml
```diff
@@ -311,20 +311,30 @@
...
       <entry value="1" name="FENCE_ACTION_GUIDED">
-        <description>Switched to guided mode to return point (fence point 0)</description>
+        <description>Fly to specified MAV_CMD_NAV_FENCE_RETURN_POINT in GUIDED mode.</description>
       </entry>
       <entry value="2" name="FENCE_ACTION_REPORT">
         <description>Report fence breach, but don't take action</description>
       </entry>
       <entry value="3" name="FENCE_ACTION_GUIDED_THR_PASS">
-        <description>Switched to guided mode to return point (fence point 0) with manual throttle control</description>
+        <description>Fly to MAV_CMD_NAV_FENCE_RETURN_POINT with manual throttle control in GUIDED mode.</description>
```
DonLakeFlyer, line 325: Is ArduPlane keeping the concept of fence return points in the new implementation or is it switching to rally points? I'm concerned about reusing these actions for the new fence protocol and carrying over things which may not make sense any more.

hamishwillee: @DonLakeFlyer I _understand_ that ArduPlane will not use these fence return points any more. However they will also have legacy vehicles that still support the setting. 

@auturgy are you OK with us marking the cases like FENCE_ACTION_GUIDED_THR_PASS as deprecated and having adding text to description "do not use". If we can change the name to reserved as well that would be good - but I presume might break "someone".

auturgy:
I’d prefer to wait until the new fence protocol is merged for ArduPilot-Plane, to be sure we make an accurate assessment.

Regards,

James

hamishwillee:
@auturgy Fair enough. This is pause until https://github.com/ArduPilot/ardupilot/pull/15288 merged.

FWIW commitment from ArduPilot and PX4 to implement this would be highly desirable before merging.

DonLakeFlyer:
> FWIW commitment from ArduPilot and PX4 to implement this would be highly desirable before merging.

I would say that it should be required for at least one of them to commit. Otherwise it's just cruft in the spec.

hamishwillee:
>> FWIW commitment from ArduPilot and PX4 to implement this would be highly desirable before merging.

> I would say that it should be required for at least one of them to commit. Otherwise it's just cruft in the spec.

And you'd be right. This evolved from a chat in a dev call where everyone said "what a good idea" but that's not the same as saying "we'll implement it".

joshanne:
> Fair enough. Blocked/paused until [ArduPilot/ardupilot#15288](https://github.com/ArduPilot/ardupilot/pull/15288) merged.

hamishwillee:
Now no longer blocked by ArduPlane, as geofence implementation went in. 

@joshanne Can you confirm which fence actions from the set in this PR are/are not supported by the new implementation? (e.g. I'm guessing you might have dropped FENCE_ACTION_GUIDED_THR_PASS )
Also if there are any new ones?

Lastly, did you drop support for the [MAV_CMD_NAV_FENCE_RETURN_POINT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_RETURN_POINT) (which makes no sense now)

hamishwillee:
We agreed in the dev call this should be "standard". I have updated to clarify that  the `FENCE_ACTION_GUIDED_` variants are only supported by ArduPlane. The enum remains in common.xml as it is an update. The new message command is in  development.xml as per new process.

@DonLakeFlyer There is an action on @auturgy to confirm whether the new ArduPlane updates dropped support for `FENCE_ACTION_GUIDED_` action variants, and for the `MAV_CMD_NAV_FENCE_RETURN_POINT`. The PR says that an autopilot that does not support the time of action should reject it. QGC might reasonably decide not to expose the `FENCE_ACTION_GUIDED_` options directly.

@LorenzMeier @auturgy Any comments before I merge?

hamishwillee:
Self merged.

PR #1483, opened 2020-09-19
common: Add a DECAROTOR type

muramura:
I've updated the source and added a 10-rotor helicopter type.

Before:
#1479


hamishwillee:
Thanks! Merging.

muramura:
@hamishwillee san. thanks.

PR #1484, opened 2020-09-21
CMake: Removing slugs.xml

thopiekar:
Looks like this XML has been removed over time. I'm just wondering why nobody noticed that it is missing.

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com> (github: thopiekar)


hamishwillee:
It was removed recently, and this was missed. Thank you for the catch!

PR #1485, opened 2020-09-21
CMake: Using find_program to find mavgen(.py)

thopiekar:
If neither mavgen nor mavgen.py have been found as executables in $PATH, then we will fallback to the hardcoded location (submodule).

Signed-off-by: Thomas Karl Pietrowski <thopiekar@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,7 @@ set(LIBRARY_SOVERSION "0.0.0")
 # none required
 
 # options
+option(USE_SYSTEM_PYMAVLINK "Generate mavlink files with system-wide installed mavgen" OFF)
 option(WITH_TESTS "Build test programs." OFF)
 option(WITH_BUILD_DEPS "Build dependencies." OFF) # no deps currently to build
 option(WITH_BUILD_STATIC "Build preferring static linking." ON)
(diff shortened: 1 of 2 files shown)


hamishwillee:
@julianoes This changes cmake file from always taking the version in the current submodule to preferentially taking a version it finds on the system path. 
I don't know if that is desirable or not, or if the cmake changes are correct - can you please review?

thopiekar:
Well, the pro is that you don't need to mess with submodules. Looking around the repositories here (other projects included), I noticed that you love to use submodules.
That makes sense if you are using Makefiles, but as you are using CMake, I would recommend to make use of it's "ExternalProject" functionalities.

I'm perfectly fine, if we add an if() around this section. I just managed to build and package pymavlink separately and it is installing scripts into /usr/bin/, so I want to make use of these.

hamishwillee:
@thopiekar Thanks for the explanation. All sounds reasonable to me, but I'm deferring to @julianoes who will have a much more educated view :-)

julianoes:
@thopiekar I think that an option to enable this would be good, so `-DUSE_SYSTEM_PYMAVLINK=ON`.

hamishwillee:
This went stale - @thopiekar Are you planning on making the [change suggested](https://github.com/mavlink/mavlink/pull/1485#issuecomment-698718752) or should I close this?

thopiekar:
Hello @hamishwillee !
Thank you for coming back on this! 

I lost this PR on my desk. I'll continue on this and it should be ready at the end of the week at the latest. 😉

thopiekar:
Ok, correcting myself. Just found time and finished it before it forget about it again.

hamishwillee:
:-) Thanks very much!
@julianoes You happy with this?

julianoes (approved):

PR #1486, opened 2020-09-22
Gimbal Device: Add rc flags

olliw42:
The gimbal protocol v2 messages may not be in a good shape and it's outcoming a bit unpredictable, but this pr is based on the assumption that the set of messages dealing with what is called the gimbal device is pretty solid and will remain irrespective of what happens to the other parts. Indeed, the latest pr #1466 is not affecting those (except of improvements of descriptions).

The pr adds RC flags to the gimbal device messages. The rational is the observation that many gimbals have the ability to digest an input rc signal, be it a PWM signal or other protocols like sbus and friends, and it even could be a RC_CHANNELS mavlink message. This rc input is processed by gimbal specific procedures possibly using gimbal specific parameters with the effect to move the gimbal. The added rc flags allow us to tell the gimbal what it is supposed to do with that rc input signal (possibly overriding settings made by the user via a gimbal specific gui, or gimbal default settings). It is functionality on the same level like RETRACT or NEUTRAL.

On this occasion I also took the liberty to expand the size of the gimbal device flags to uint32. I first note that the gimbal device failure_flags is also of size uint32, which appears to be massively oversized and very future proof. It doesn't make much sense to me to expect up to 32 error flags but only 16 state flags at most. I rather think the opposite seems more likely.

Currently the gimbal protocol v2's idea is that the 16 bits of the gimbal device flags are 1-to-1 mirrored into the first 16 bits of the 32-bit wide gimbal manager flags. If that is to prevail, the proposal here would be to continue to do so. That is, the lower 16 bit of the gimbal device flag would be mirrored into the lower 16 bits of the gimbal manager flags, but the higher 16 bits of the gimbal device flag would not be carried over to the gimbal manager flags (and are so to say hidden).

Indeed, the added rc gimbal device flags need not to show up in the gimbal manager flags at all, and in fact should not show up up there, since rc control would be handled by other means by the gimbal manager. Accordingly, I have placed the rc flags in the upper 16 bits.

In the case the current approach to mirror the gimbal device flags into the gimbal manager should not prevail, no harm has been done either. So, it makes much sense to me.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -415,6 +415,9 @@
       <entry value="2048" name="GIMBAL_DEVICE_CAP_FLAGS_SUPPORTS_INFINITE_YAW">
         <description>Gimbal device supports yawing/panning infinetely (e.g. using slip disk).</description>
       </entry>
+      <entry value="4096" name="GIMBAL_DEVICE_CAP_FLAGS_HAS_RC">
+        <description>Gimbal device supports control or nudge via an RC input signal.</description>
+      </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_CAP_FLAGS" bitmask="true">
       <description>Gimbal manager high level capability flags (bitmap). The first 16 bits are identical to the GIMBAL_DEVICE_CAP_FLAGS which are identical with GIMBAL_DEVICE_FLAGS. However, the gimbal manager does not need to copy the flags from the gimbal but can also enhance the capabilities and thus add flags.</description>
@@ -486,6 +489,12 @@
       </entry>
       <entry value="16" name="GIMBAL_DEVICE_FLAGS_YAW_LOCK">
         <description>Lock yaw angle to absolute angle relative to North (not relative to drone). If this flag is set, the quaternion is in the Earth frame with the x-axis pointing North (yaw absolute). If this flag is not set, the quaternion frame is in the Earth frame rotated so that the x-axis is pointing forward (yaw relative to vehicle).</description>
+      </entry>
+      <entry value="65536" name="GIMBAL_DEVICE_FLAGS_RC_CONTROL">
+        <description>Set to RC control. The RC input signal fed to the gimbal device controls the gimbal's orientation.</description>
+      </entry>
+      <entry value="131072" name="GIMBAL_DEVICE_FLAGS_RC_NUDGE">
+        <description>Set to RC nudge. The RC input signal fed to the gimbal device nudges the gimbal's orientation.</description>
       </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_FLAGS" bitmask="true">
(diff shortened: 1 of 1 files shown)


olliw42:
@hamishwillee @julianoes 
May I get a _preliminary_ response on the PRs #1486 #1487 #1490, like a tentative assessment if they may be accepted or not? I would like to know the direction I should move on. :)

julianoes (changes requested):
@olliw42 I'm sorry I disagree with your suggestions. As laid out earlier I don't think RC should be routed directly through to the gimbal. It would be a backdoor based on an old protocol. Instead, I've suggested a separate messages GIMBAL_MANAGER_SET_MANUAL_CONTROL and GIMBAL_DEVICE_SET_MANUAL_CONTROL in #1466.

Also, I got rid of the control and nudge flags and all that remains is primary and secondary control, and it's up to the implementation on how to mix it.

Regarding the size of the flag field. I'm confused why you would want "hidden" flags. To me it makes perfect sense to have less flags for the device which means that they can always be mapped to the manager. Or let me ask the question: what does it solve if these flags are also 32 bit? To me the only effect will be developer confusion.

olliw42:
MANY thx for the quick response

> I've suggested a separate messages 

I've seen that, and it actually triggered asking :)

I continue to argue that they are totally unneeded and superfluous

>  I don't think RC should be routed directly through to the gimbal

this is how many people's setup works and it's how it's going to be for a long while ... you effectively kick out many users for a long time from benefiting from the new messages and force them into the old protocol

>  I'm confused why you would want "hidden" flags.

I don't want hidden flags, I just said that this would be so IFF one wants to maintain to mirror the flags

> To me the only effect will be developer confusion.

frankly, I'm tired of this argument. 1. The "developer"s seem to really be a very vulnerable low capability bread of humans. 2. In many MAVLink places many things are NOT made simple but complicated even though simpler solution would be possible, this argument is mostly used when it is convenient :)

>  I'm sorry I disagree with your suggestions.

ok. I fully accept this. Thanks for being clear, this is helpful, it clarifies things on my side. :+1: 

I guess it means these PRs can be closed. So I do :)

julianoes:
> 1. The "developer"s seem to really be a very vulnerable low capability bread of humans.

That's not what I'm saying but when you are new to something from the outside and there is a lot of context and legacy things like RC pass-through it can get complicated. And you can't always infer from clear it is to you to others. I'm speaking from experience when I bring up this point because I'm often the one trying to explain how it works to implementers.

> this is how many people's setup works and it's how it's going to be for a long while ... you effectively kick out many users for a long time from benefiting from the new messages and force them into the old protocol

This is where I'm confused. What's so hard in parsing the MANUAL_CONTROL message and use it, just like you would use the RC_CHANNELS_OVERRIDE (or whatever)? I'm really not trying to kick out users. It's just that I believe in abstractions and not weird spaghetti design.

auturgy:
this is somewhat of a philosophical argument.  I'm not sure there's ever a clear answer when a pragmatist and idealogue engage in a debate :)
In this case I agree with both perspective: from a purist perspective, Julian is right.  Using a normalised representation makes for a cleaner API.  If we were designing from scratch, and did not have to consider backwards compatibility, this is the approach I think we should take.
But I agree with Olli.  There are too many existing systems - not just STorM32, but other very high end, professional equipment - that use the legacy RC inputs, so we should allow them to be explicitly handled.

olliw42:
> This is where I'm confused. What's so hard in parsing the MANUAL_CONTROL message

it's not hard. It just needs to be implemented system wide. Most things are not hard, and other systems have it for years, yet we wait for them since years to become available or even possible in MAVLink systems ... it's called facing reality :)

Example? Camera micro service not useable with ArduPilot, even though out for years. How long will it still take for users of it to become possible? I know for sure that a - I believe well recognized - AP user made a significant effort last year ... we are still waiting ...

olliw42:
> from a purist perspective ... Using a normalised representation makes for a cleaner API.

even then it is totally superfluous since it's only about passing a stick position from a gcs all throught the gimbal manger to the gimbal device ... and needs TWO new messages just for that.

In other places MAVLink is always very eager to avoid introducing new messages if the very same can be achieved with existing messages ... and it makes some sense. Here it is argued exactly the opposite way.

so, passing through the stick positions with +-1 units is super super purist

olliw42:
this thought crossed my mind, kind of as compromise between the worlds:

if custom cap flags could be acceptable for the gimbal device information message one could think that a uint16 custom flag could also be acceptable for the gimbal device control & status messages. This would allow putting the suggested RC flags and absolute yaw flags into the custom flag, and would help resolving the limitations of the current gimbal device messages, and also allow so to say "inofficial" paths for backwards compatibility.

PR #1487, opened 2020-09-22
Gimbal Device Information: Add hardware version field

olliw42:
This PR adds a field hardware_version to the GIMBAL_DEVICE_INFORMATION message.

I can't see any rational for why the message should carry info such as vendor name, model name, and software version but not also reveal info about the hardware. (I had brought it up also in #1370, and at this time it seems this point wasn't disputed but not acted on because of the general faith of the message)

Given the apparently contentious nature of the topic, as it appears to be from recent prs and issues, I've put it into this separate pr.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6123,20 +6123,20 @@
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Information about a low level gimbal. This message should be requested by the gimbal manager or a ground station using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint8_t[32]" name="vendor_name">Name of the gimbal vendor</field>
-      <field type="uint8_t[32]" name="model_name">Name of the gimbal model</field>
-      <field type="uint8_t[32]" name="user_name">Name of the gimbal given to it by the user</field>
-      <field type="uint32_t" name="firmware_version">Version of the gimbal firmware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff)</field>
-      <field type="uint32_t" name="hardware_version">Version of the gimbal hardware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff)</field>
+      <field type="uint8_t[32]" name="vendor_name">Name of the gimbal vendor.</field>
+      <field type="uint8_t[32]" name="model_name">Name of the gimbal model.</field>
+      <field type="uint8_t[32]" name="user_name">Name of the gimbal given to it by the user.</field>
+      <field type="uint32_t" name="firmware_version">Version of the gimbal firmware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff).</field>
+      <field type="uint32_t" name="hardware_version">Version of the gimbal hardware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff).</field>
       <field type="uint64_t" name="uid">UID of gimbal hardware (0 if unknown).</field>
       <field type="uint16_t" name="cap_flags" enum="GIMBAL_DEVICE_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
       <field type="uint16_t" name="custom_cap_flags" display="bitmask">Bitmap for use for gimbal-specific capability flags.</field>
-      <field type="float" name="tilt_max" units="rad">Maximum tilt/pitch angle (positive: up, negative: down)</field>
-      <field type="float" name="tilt_min" units="rad">Minimum tilt/pitch angle (positive: up, negative: down)</field>
-      <field type="float" name="tilt_rate_max" units="rad/s">Maximum tilt/pitch angular rate (positive: up, negative: down)</field>
-      <field type="float" name="pan_max" units="rad">Maximum pan/yaw angle (positive: to the right, negative: to the left)</field>
-      <field type="float" name="pan_min" units="rad">Minimum pan/yaw angle (positive: to the right, negative: to the left)</field>
-      <field type="float" name="pan_rate_max" units="rad/s">Minimum pan/yaw angular rate (positive: to the right, negative: to the left)</field>
+      <field type="float" name="tilt_max" units="rad">Maximum tilt/pitch angle (positive: up, negative: down).</field>
+      <field type="float" name="tilt_min" units="rad">Minimum tilt/pitch angle (positive: up, negative: down).</field>
+      <field type="float" name="tilt_rate_max" units="rad/s">Maximum tilt/pitch angular rate (positive: up, negative: down).</field>
+      <field type="float" name="pan_max" units="rad">Maximum pan/yaw angle (positive: to the right, negative: to the left).</field>
+      <field type="float" name="pan_min" units="rad">Minimum pan/yaw angle (positive: to the right, negative: to the left).</field>
+      <field type="float" name="pan_rate_max" units="rad/s">Minimum pan/yaw angular rate (positive: to the right, negative: to the left).</field>
     </message>
     <message id="284" name="GIMBAL_DEVICE_SET_ATTITUDE">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6127,6 +6127,7 @@
       <field type="uint8_t[32]" name="model_name">Name of the gimbal model</field>
       <field type="uint32_t" name="firmware_version">Version of the gimbal firmware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff)</field>
       <field type="uint32_t" name="hardware_version">Version of the gimbal hardware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff)</field>
+      <field type="uint64_t" name="uid">UID of gimbal hardware (0 if unknown).</field>
       <field type="uint16_t" name="cap_flags" enum="GIMBAL_DEVICE_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
       <field type="uint16_t" name="custom_cap_flags" display="bitmask">Bitmap for use for gimbal-specific capability flags.</field>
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Makes sense, thanks.

olliw42:
question, pl: 

I also would like to suggest adding a uint64 UID field. 

The purpose would be that e.g. a GCS can check if the gimbal was swapped out. That is, a user may have two or more gimbals which may be installed, but some configurations might have to be somewhat different. With a UID one could do things like offering profiles which need to be set up only once, and could then be selected by a GCS, or even an autopilot (e.g. the autopilot may need to know if the gimbal has pan or not, like in ArduPilot, but it's out of my mind that there ever will be a MAVLink way to reveal such deep details) 

Would you say yes and I should add it to this PR or would you say no?

olliw42:
> I would push back mostly because we also have the component information capabilities for anything custom

the component information capabilities is only a theoretical possibility for typical components, and really not a practical one ... in other places you find sort of attmitance of this

olliw42:
I've added a custom cap field and a UID field in two commits, so you can decide and cherry pick :)

olliw42:
I'm sorry to further bother. I recalled that we actually once had yet another field in discussion, and I believe to recall that the response wasn't outright negative either.  So

3. I would also suggest adding a uint8_t[32] user_name field.

Some implementations allow a user to set a name for their component. This allows them to easily distinguish between two or more hardware items which may come from the same vendor and be of the same model.

Would you say yes and I should add it to this PR or would you say no?

hamishwillee:
As I understand it the gimbal device is the low level information about what the gimbal supports. The  gimbal manager uses this to decide how to reformulate higher level commands to something the gimbal understands. Is that right?
So to be useful in a mavlink sense it sounds like your gimbal manager and the device info need to have an understanding of each other?

So if you add these fields, what would a GCS have to do to use them? What would you have to change in the gimbal manager on PX4 or ArduPilot have to do? 

I look at the UID in the same way. Are you expecting the gimbal manager to process the meaning of the UID change? If this is just a way of telling the manager it needs to re-query the gimbal capabilities, are there better way. 

Perhaps I have completely missed the point but it feels to me that these fields might be interesting for display, but they are a step away from providing a common interface for things that want to use the gimbal.

olliw42:
hey @hamishwillee 
your points are well thought out, and from the perspective you take I would not disagree with your points. I however do have a different point of view, and this let's me come to different conclusions. IMHO the gimbal manager is a bit of a more variable beast. First, I feel it is still a way to go to have it finally defined, but maybe more importantly, the gimbal manager in my view can be much more implementation dependent than the gimbal device. I mean, you can have many types of software on your PC using a harddisk, but the low-level driver for the harddisk would be a pretty fixed thing. So I think that a particular implementor/integrator might swap out a gimbal manager for its own (and IMHO that's actually one of the key benefits of the introduction of a gimbal manager), but she/he would still deal with the same gimbal device. As a result, the gimbal device set of messages may have different and possibly more states/flags/fields than what a particular gimbal manager would use or expose as its API. The gimbal device messages should IMHO expose the relevant gimbal capabilities.

I thus am currently trying to make the gimbal device messages by themselves as useful as possible, without thinking very much about a gimbal manager would actually use this information. This is what the recent 3 or so PRs on the gimbal device were about, including this one. 

In some way my logic is a simple as that: Any info the gimbal device doesn't expose can't be known by a gimbal manager (or anyone else), while a gimbal manager may not necessarily be interested in all info a gimbal device exposes.

As regards the gimbal device information message it is e.g., IMHO, not finally clear if that is not exposed also to the outside. I think currently the idea is that a gimbal manager, and only a gimbal manager, reads this message, and that the gimbal manager provides a gimbal manager information message which exposes all the info. And I think that this is largely the basis of your line of thoughts. 

However, this implies that all fields must be mirrored, and that the gimbal manager information message potentially becomes large. Also, pl note, that even though this appears to be the current intention, there do not seem to be efforts to actually do this, i.e., several fields of the current gimbal device information message are not exposed in the gimbal manager information message. 

This logic appears to be also concluded for the gimbal device status message, a GCS should read it, it's not mirrored by a respective gimbal manager message. Also just try to expand the thinking for a moment and let's consider COMPONENT_INFORMATION. This clearly would be located on the gimbal device, and it may reveal gimbal device related specific info. So, do you really expect that the gimbal manager also should mirror this gimbal device specific info, wouldn't it be just so much easier to say that e.g. a GCS should read it directly from the gimbal if it wants to do so?

So, I "predict" that the prevailing approach will be that also e.g. a GCS may read the gimbal device information, if it is interested in the info it provides.

Therefore I'm trying to make gimbal device information as useful as possible.

The suggested info may seem at first to be primarily about only displaying. But I don't think so, it is largely also about identification. 

There might be reasons why someone else, be it a gimbal manager implementation, a companion, a GCS, or whoever, wants to know what type of gimbal this is, e.g., because that gimbal provides non-standardized features an implementation wants to use. You clearly note that this is exactly the situation with e.g. PX4 or ArduPilot. You need to know which one it is in in many places. And that's not a malfunction! It's a necessary implication of the fact that MAVLink allows dialects. If MAVLink allows dialects, then there must also be reasonable ways to allow identification of whether a component speaks that dialect or not. You can't say A and not B. As much as I can see, the gimbal device information is the "best" current place to do that for a gimbal device. 

MAVLink might also eventually come to a point there it enables firmware upgrade of a component via MAVLink. Form a user perspective this is clearly a much wanted feature. This also will require proper identification.

And so on.

I know that you guys often argue to not add fields/messages for which there is no clear actual need or implementation, and while I agree that sometimes this is a good position, I think that more often so one should better think ahead and anticipate what can be reasonably anticipated already now.

I hope this long piece gives a better impression on how I look at it.

I don't expect that this text makes you guys to now agree. For some, probably mostly idealistic reasons I simply suggest what I think it should look like, you guys can then agree or not, or ignore, but I then can go ahead with my stuff. :):)

hamishwillee:
@olliw42 

Thanks for explaining the logic. You're way more expert on gimbals than I am  - and while it would be better if we had a much broader engagement with gimbal manufacturers, you're the guy in the room. For which I am grateful.

Hmmmm, so a more capable device driver might allow more/better options for the gimbal manager to exploit if they are present. 
- This allows room for innovation and differentiation without breaking the standard for anyone else (provided the core primitives exist). 
- I am not sure how likely such innovation is - a manufacturer with a gimbal that exposes a gimbal manager API can access the low level things internally. Most system integrators will just take the easiest route - use the PX4 gimbal manager.

Yes, there is interest in being able to do a mavlink firmware upgrades. Currently the integration is that a system that can manage an upgrade gets sent [MAV_CMD_DO_UPGRADE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_UPGRADE). The upgrader system is set up by an integrator and is assumed to know what the thing being upgraded is. The current use case being a companion computer that knows how to upgrade the firmware of a flight stack.

That works, but I find the idea of being able to upgrade arbitrary components attractive. I envisage that the component information incudes a flag that the component is upgradable and a server to check for the latest version. The GCS checks for this and has the detailed dependency info in the component metadata. It then offers option to update. Upshot, the information is needed for the use case but not needed in the message. 

Attempting to channel Lorenz/play devils advocate - we want to keep the APIs as simple appearing as possible so they are not intimidating. That leans me towards asking whether these can be added later if needed. Or exposed in different ways - i.e. through metadata - for those that want them?

If it were my decision I would lean towards yes for the gimbal device capabilities. I'd try encourage a generic approach based on component information for the rest of the version data. That is what it is for. @julianoes call though.

olliw42:
> while it would be better if we had a much broader engagement with gimbal manufacturers, you're the guy in the room.

I think this is then a bit of a problem, and might worry you. Greemsy should be involved.

As regards me being the man in the room I guess I have to unfortunately tell you then that I'm not really in that room anymore. I've decided to go with my own messages, and the only unclear point is if I can use this gimbal device information message or not. 

As regards the other points you address, I guess I just can repeat myself. I don't disagree with them, I just take another position. I basically just say, if you say so in one place you also should say so in another place and not differently.

> I am not sure how likely such innovation is - a manufacturer with a gimbal that exposes a gimbal manager API can access the low level things internally. Most system integrators will just take the easiest route - use the PX4 gimbal manager.

that's the division point. You may be correct on the innovations and that most will just use the PX4 gimbal manager. However, this also very much depends on how the messages had been drafted in the first place. If they are basically just made to suit PX4, they likely won't be picked up by others since they don't suit their needs. So, there is a bit of a self-fulfilling prophecy aspect here.

E.g. I was strongly advocating for making the gimbal device robotics like, which included the angular rate vector. However, it's probably safe to say that there is no implementation yet supporting or needing this. So, according to "your" logic these fields shouldn't be there and go away. Similar so with some other fields and flags. So, the gimbal device messages are clearly made with a grander picture in mind. Irrespective of current or likely use, but because it makes it a concept. One of course could argue, let's add them then they are needed. However, experience shows that by going that path too often mistakes had been ingrained which are then hard to root out later. You can see that throughout MAVlink. That's why there is still not a clear scheme for component discovery, is PARAM and PARAM_EXT and PARAM_EXT_TRIMMED and still perception of insufficiency and need for a v3, is MAVLink 1 and MAVLink 2 and still major flaws and limitation needing improvements, and some more. Not always, but often looking for the grander picture produces results which are more future proof. That's at least how I generally approach the things (or try to do things). 

>  That leans me towards asking whether these can be added later if needed.

yes, this is a possible way. But it's not my way for reasons given here and elsewhere :). And as said I'm not actually trying to convince you guys or argue. I just like to know.

My feeling would be however that if you decide against adding fields like hardware, uid, etc, you probably should also be removing the two name and firmware fields. They came for camera information. This would make the message at least concentrate on the gimbal essentials.

julianoes:
@olliw42 I challenge two things, the rest, I would like to merge.

`user_name`:
- First of all, I find the name confusing and I would rename it to `user_given_name`. That's because `username` reminds me of a login box which is not what thi is.
- Secondly, I think the point of this message is that you can distinguish one gimbal from another one. If you really happen to have two gimbals of the same type (so same Vendor and Product name) then the uid is what you use to distinguish it. If you want, you can keep an index of which uid has which name on the ground station or wherever you need it. My argument here is mostly based on the fact that I don't think it will get much use and taste trying to save 32 bytes, so not very good.

`custom_cap_flags`:
- I can't think someone would need this. And, if there is something missing from the "official" capability flags, why not suggest it there? Or to rephrase: give me an example of a custom capability flag that you want and can't be in the standard and I reconsider this!

olliw42:
1. changing name is perfectly fine to me
2. well, this is true, code can easily differentiate by uid, I'm not convinced users can equally easily. I agree though that this is not the most relevant field, it would however make it rounded.
3. #1486, #1490. A user cap flag would be the most crucial addition to me ;)

julianoes:
I'm still not convinced on adding the 32 bytes for a name but I can live with it :smile:.

About the `custom_cap_flags`, I still don't think it's required, but we might as well have it for whenever it comes up. Now I wonder if it should be called `custom_flags` instead, so that it's not necessarily related to only capabilities?

olliw42:
> About the custom_cap_flags, ... but we might as well have it for whenever it comes up.

I'm not sure if that counts as "whenever it comes up": Let me be explicit. I have a very clear plan for STorM32 now, and the only undecided point is if I use this gimbal information message or use my own. In that sense one could argue that it came up since I would need that custom_cap_flags since otherwise I couldn't expose my flags. Given that I'm prepared to anyway do what I want one however also legitimately could argue that one doesn't have to care about what I need. So, please follow what YOU think is the best for MAVLink. I just made the suggestion and gave my arguments.

julianoes:
Yes, I plan to add it to unblock you and potentially others in the future. Could it be called `custom_flags` or do you think `custom_cap_flags` is better?

olliw42:
that's great to hear :)
I guess I would keep custom_cap_flags. It's an information message with static flags, so I wouldn't know what flags could be there which not in one way or the other could be understood as capability of the gimbal. But it's really semantics. If you prefer custom_flags, so it be :)

olliw42:
it seems you've an opinion now. Could you tell me what changes I need to make, and then possible merge?

I'm sorry if I come a cross a bit pushy, I'd like to get forward :)

julianoes (approved):

olliw42:
very cool, THANK you so much, sir !!

julianoes:
Welcome and thanks for the ongoing effort to improve this.

PR #1488, opened 2020-09-22
AUTOPILOT_STATE_FOR_GIMBAL_DEVICE: it's Hamilton

olliw42:
Hamiltonian convention is nonsense, it's Hamilton convention, or Hamilton's convention, or whatever

(Hamiltonian is the Hamilton function or Hamilton operator, or related mathematical constructs, nothing to do at all with quaternions)

I also grabbed the opportunity to move the two target fields to the front. A simple search through common.xml shows that this is so for all other messages (except one which has a network thing).


julianoes (approved):
Makes sense, thanks @olliw42!

amilcarlucas (approved):

hamishwillee:
Thanks @olliw42

PR #1489, opened 2020-09-22
Ardupilot downstream updates up-to 2020-09-22

amilcarlucas:


amilcarlucas:
I think we should increment the mavlink version after this gets merged

hamishwillee:
@amilcarlucas Thanks

> I think we should increment the mavlink version after this gets merged

Why? We've merged plenty of stuff before and not updated things :-)

Where did `instance="true"` come from? What is it for, when is it applied? I'm not very keen on ardupilot making changes in common and pushing up - always slows things down.

auturgy:
Schema changes occur in pymavlink. The instance tag was added here:
https://github.com/ArduPilot/pymavlink/pull/459

amilcarlucas:
@hamishwillee the [latest version (1.0.12)](https://github.com/mavlink/mavlink/releases)  is from 30 May 2019
We now have:
* minimal.xml,
* Improved markup for deprecation, wip and other stuff
* deprecated slugs
* added instance=true attributes :)

So yes, I think it's time for a 1.0.13 release

amilcarlucas:
@hamishwillee any progress on this one? I added the dev cal flag in an attempt to speed things up :)

amilcarlucas:
Rebased on master and added pymavlink updates

hamishwillee:
@amilcarlucas Thanks for the prompt. I'm part time right now which makes it hard to find time for this. Good-ish news though is that while my time is reduced, MAVLink is the first priority for what time I do have starting from a week or 2 from now.

All of this makes sense to me except for `instance=true`. I looked at the PR "This allows for array indexing of mavlink messages, so that array syntax can be used for messages like BATTERY_STATUS which have an instance field" - and I still can't work out why you would bother. Ie what benefit does this give you and how do you use it differently. 
I can see that this syntax can be applied to command and messages, but that at time of writing it only affects message generation, and only for the generated Python code.

Can you explain?

Further, most of the instances are applied to a variable that takes an id. So even though I don't see the value (yet) it makes sense how they are applied. BUT, see `NAMED_VALUE_FLOAT.name` is an instance, and that is a `char[10]`.

amilcarlucas:
The instance=true is a truly great change!!!!
Allows GCS to for example understand the BATTERY_STATUS message. Let's say you have two batteries, both with different voltages, both sending BATTERY_STATUS messages. In the GCS if you are graphing the voltages, the current graph will be complete garbage, both voltages will be interleaved....
This change allows the GCS in a true dynamic way to understand that there are two instances and that it should group the messages from one instance in one graph and from the second instance on the second graph.
It also works if the id's are strings (char[10]) .

Less hardcoded stuff in the GCS, more explicit markup in the protocol.!!

Like I said a truly great improvement.

hamishwillee:
@amilcarlucas Thank you for the explanation! 

@DonLakeFlyer You know about the changes to support "is instance" markup ? Amlicar explains it above ^^^. The example is that one message is used to send the information for different batteries, with the battery instances being identified by a particular field.  
- right now if you want to plot the behaviour for different batteries you'd have to hard code that the message is an "instance" and which field is the instance 
- with this change you can work that out from the xml
- and if you were using Python library you could get the info direct from the generated library. Might be worth updating the library you use similarly.

hamishwillee:
@amilcarlucas Merged. Thanks for your work and explanations. (FYI @auturgy )

I would **prefer** not to do a 1.0.13 release. Why? 
- No one is using the release number for filtering/compatibility because we don't do it often enough for anyone to bother.
- to make it worth people doing so we would have to do this regularly, and do change list info to help them understand the changes. And then we run into the problem that we will only be able to do this 256 times.
- This is not a break of any kind. Even the switch to minimal.xml is compatible. 

I would like to do the next break when we move to using standard.xml. That is going to be a significant break, worthy of the work.

If you strongly disagree and create a PR then I am happy for this to be discussed in the dev call. I would not be strongly opposed though I would raise the above concerns.

amilcarlucas:
@hamishwillee please do not squash my commits. I do that myself, and I make sure I do proper commit messages than can be tracked in git history.

@hamishwillee Regarding the version, here are my counter arguments:

1. I use the release number on our commercial projects
2. I plan to increment this number in the future when significative changes happen (like the ones we have now). It does not need to be done all the time, but is a good practice to do it at least from time to time.
3. the minimal.xml change did break mavros. They had to fix the compilation on mavros 1.40. So your third argument does not apply.

I will do a PR to increase the version number. Discussion can continue on that new PR.

hamishwillee:
> @hamishwillee please do not squash my commits. I do that myself, and I make sure I do proper commit messages than can be tracked in git history.

Sorry, I assumed that the squishing with the default commit message did that.

Re the rest - fair enough :-)

amilcarlucas:
@hamishwillee can you do a release? Looks like it is a github tagging stuff that I can not do with a PR.

hamishwillee:
@amilcarlucas I'm not working on a reduced cycle, so sorry for the delay. Since we generally do not release ever, I actually don't know what you are asking for. All I would do is update https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L4 (version) to 4.

amilcarlucas:
@hamishwillee No, I did not mean that. I meant tag a release like this: https://github.com/mavlink/mavlink/releases

hamishwillee:
@amilcarlucas Thanks for clarification. I see no harm in this, and if it helps you more than happy to do so. See https://github.com/mavlink/mavlink/releases/tag/1.0.13

hamishwillee:
@DonLakeFlyer Pretty much nothing, and yes they are. But @amilcarlucas wanted this.

Issue #1369, opened 2020-05-04
Gimbal V2 Protocol Discussion II: Absolute/relative yaw


olliw42:
In the long gimbal thread the topic of absolute and relative yaw, and the related issues such as with smart gimbals which need autopilot assistance, had been discussed at quite some length, since it's not an easy topic. I had the impression that some conclusion was reached in that thread. It is even addressed in the google document.

However, I do not see any appearance of this in the current mavlink gimbal v2 messages. It is in fact totally unclear to me how that should be dealt with.

The issue becomes more involved when in addition autopilot assistance through the AUTOPILOT_STATE_FOR_GIMBAL_DEVICE message is at play. This introduces some additional time lines in the whole flow, and I can't see how one could resolve that with what is existing.

So, in short, no idea how that all is supposed to work, and I would appreciate some clarifying comments.

:)

julianoes:
You're right that this topic is not addressed enough in the documentation. Let's improve it, and by the way, feel free to make pull requests to the docs and suggest changes and additions as you would like to see them. The current iteration is merely a first try of mine.

Now, here are the notes coming from the message definitions that do already exist:

1.

Quaternion in [GIMBAL_MANAGER_SET_ATTITUDE](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_SET_ATTITUDE)
> the frame is depends on whether the flag GIMBAL_MANAGER_FLAGS_YAW_LOCK is set

Quaternion in [GIMBAL_DEVICE_SET_ATTITUDE](https://mavlink.io/en/messages/common.html#GIMBAL_DEVICE_SET_ATTITUDE))
> the frame is depends on whether the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set

This is not very clear, maybe something like this would be better:

The frame is the world frame if ..._FLAGS_YAW_LOCK is set (e.g. yaw 0 points North).
The frame is the world frame, however, aligned with yaw of the drone (e.g. yaw 0 points forward).

2. 

The meaning of FOLLOW and LOCK is explained in [GIMBAL_MANAGER_CAP_FLAGS](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_CAP_FLAGS)/[GIMBAL_DEVICE_CAP_FLAGS](https://mavlink.io/en/messages/common.html#GIMBAL_DEVICE_CAP_FLAGS), however, it would make sense to add a point to the docs to explain the meaning.

3.

[...]

PR #1490, opened 2020-09-23
Gimbal Device: Handling of absolute yaw

olliw42:
This is my take at the issue of handling absolute yaw for gimbal devices. (related also to #1369).

Currently, the frame is linked to wether the yaw axis is in locked (= absolute yaw) or in follow (= relative yaw) mode. 

IMHO, this is not appropriate and a quite serious restriction to practical applications, making it essentially impossible for many gimbal users to put their gimbal into lock mode. Note that this can be a much desired operation, even if the gimbal doesn't have absolute yaw. If you want to dispute that watch the videos posted here done in a commercial context:  https://www.rcgroups.com/forums/showpost.php?p=45371373&postcount=12922

The idea of the flags is that a gimbal may not be able to support absolute yaw for every point in time during operation. For instance, it may take a while before it gets the supporting data from the autopilot. There might be also longer or shorter outages during operation.

The added capability flag indicates if the gimbal may support absolute yaw during runtime, or not. 

The added CAN_ACCEPT_YAW_ABSOLUTE is the flag which reports to the outside world if the gimbal device is currently able to accept absolute yaw angle inputs.

Finally, the added YAW_ABSOLUTE flag has the usual dual role of these flags, i.e. it can be set in a SET_ATTITUDE message which when indicates to the gimbal device that the data is in absolute frame, or else in relative frame. It is also raised in the STATUS message if the reported data is in absolute frame or not.

One point is left open, namely when should the gimbal device send STATUS with absolute yaw, and when not. A simple rule would be that gimbal device always sends in absolute frame whenever it can, else not. However, one also equally well could it just leave to the gimbal device implementation. Outside code needs to adapt to both cases anyhow.

The proposal is only for the gimbal device, at the level of the gimbal manager one may want to do whatever else.

EDIT:
If the extension of the gimbal device flag to uint32 as proposed in #1486 would be accepted, one also could consider to move the two newly added flags to the higher 16-bit area, if one thinks they should be hidden to a gimbal manager user.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -412,8 +412,11 @@
       <entry value="1024" name="GIMBAL_DEVICE_CAP_FLAGS_HAS_YAW_LOCK">
         <description>Gimbal device supports locking to an absolute heading (often this is an option available)</description>
       </entry>
-      <entry value="2048" name="GIMBAL_DEVICE_CAP_FLAGS_SUPPORTS_INFINITE_YAW">
-        <description>Gimbal device supports yawing/panning infinetely (e.g. using slip disk).</description>
+      <entry value="2048" name="GIMBAL_DEVICE_CAP_FLAGS_HAS_INFINITE_YAW">
+        <description>Gimbal device supports yawing/panning infinitely (e.g. using slip ring).</description>
+      </entry>
+      <entry value="4096" name="GIMBAL_DEVICE_CAP_FLAGS_HAS_ABSOLUTE_YAW">
+        <description>Gimbal device supports absolute yaw angles (this usually requires support by an autopilot, and can be dynamic, i.e., go on and off during runtime).</description>
       </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_CAP_FLAGS" bitmask="true">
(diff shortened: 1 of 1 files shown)


julianoes (commented):
I'm confused. To me the question whether a gimbal supports absolute yaw is already provided by the flag `GIMBAL_DEVICE_CAP_FLAGS_HAS_YAW_LOCK`. So if it can't lock to an absolute direction, it also can't accept an absolute yaw angle.

olliw42:
this are totally two different things

the STorM32 gimbal (and any other gimbal controller I know of) could do yaw lock since day one, and it still can, without knowing anything about absolute yaw. And it was useful since day one and is still useful, the video is just a latest example.

Anyway, my feeling is that it isn't in your concept of gimbal device & manager and a lengthy discussion would probably rather be a distraction than helpful, so I'd rather close this. Many thx.
(I found it important to give it a try)

julianoes:
Right ok. But hypothetically if a gimbal controller can implement yaw lock, it might as well implement knowing absolute yaw by using the data in AUTOPILOT_STATE_FOR_GIMBAL_DEVICE?

olliw42:
I repeat, this are totally two different things

of course, IFF a gimbal controller implements AUTOPILOT_STATE_FOR_GIMBAL_DEVICE it knows about absolute yaw, if not then not. But in either case it can implement yaw lock. Just look at all the hand held gimbals out there. They all can do yaw lock, but I don't know any which has an AHRS or a connection to an autopilot ...

julianoes:
Ok, now I understand what you mean.

I would argue that if a gimbal implements the gimbal v2 device protocol it also needs to implement/understand AUTOPILOT_STATE_FOR_GIMBAL_DEVICE, and thus the additional flag is not required. Does that make sense?

olliw42:
> I would argue that if a gimbal implements the gimbal v2 device protocol it also needs to implement/understand AUTOPILOT_STATE_FOR_GIMBAL_DEVICE, and thus the additional flag is not required. Does that make sense?

not for me. See 3rd paragraph in intro post. But as I said in https://github.com/mavlink/mavlink/pull/1490#issuecomment-698723522: "Anyway, my feeling is that it isn't in your concept of gimbal device & manager and a lengthy discussion would probably rather be a distraction than helpful, so I'd rather close this. Many thx."

PR #1491, opened 2020-09-23
common: correct scaled pressure extension not available value

IamPete1:
This fixes a error I introduced in https://github.com/mavlink/mavlink/pull/1429

The change is to send zero for not available as its a extension.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4411,7 +4411,7 @@
       <field type="float" name="press_diff" units="hPa">Differential pressure 1</field>
       <field type="int16_t" name="temperature" units="cdegC">Absolute pressure temperature</field>
       <extensions/>
-      <field type="int16_t" name="temperature_press_diff" units="cdegC">Differential pressure temperature (UINT16_MAX, if not available)</field>
+      <field type="int16_t" name="temperature_press_diff" units="cdegC">Differential pressure temperature (0, if not available). Report values of 0 (or 1) as 1 cdegC.</field>
     </message>
     <message id="30" name="ATTITUDE">
       <description>The attitude in the aeronautical frame (right-handed, Z-down, X-front, Y-right).</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4390,7 +4390,7 @@
       <field type="float" name="press_diff" units="hPa">Differential pressure 1</field>
       <field type="int16_t" name="temperature" units="cdegC">Absolute pressure temperature</field>
       <extensions/>
-      <field type="int16_t" name="temperature_press_diff" units="cdegC">Differential pressure temperature (UINT16_MAX, if not available)</field>
+      <field type="int16_t" name="temperature_press_diff" units="cdegC">Differential pressure temperature (0, if not available)</field>
```
hamishwillee, line 4393: @IamPete1 Sorry for delay - but why not INT16_MAX? Isn't 0 a valid value for this field? In which case you now have the question "how do I represent". Of course if 0 is not a valid value, then this is good :-) ditto for other fields.

IamPete1: @hamishwillee My understanding is that because this is a extensions we have to use 0 so that 'did not get a value' is the same as 'got a value but it not valid'. In the AP implementation I just fudge the temp to 1cdeg if we get a valid value of 0.

hamishwillee: @IamPete1 That is correct :-). But then you still need to clarify that "fudging" in the message description.

So perhaps 

> Differential pressure temperature (0, if not available). Report values of 0 (or 1) as 1 cdegC.

IamPete1: Done, thanks

hamishwillee:
@IamPete1 Looks good. Unfortunately travis check has failed, for what looks like things outside our control. Can you re-trigger the build on travis - it isn't letting me because it is not my PR. If not, I can "touch" this to force  a rebuild.

IamPete1:
@hamishwillee I weren't allowed to re-start so I have changed the commit name slightly and force pushed, hopefully it passes this time

hamishwillee:
@IamPete1 What a pain - thanks for that. Merged :-)

PR #1494, opened 2020-09-26
minimal: shift deprecated MAV_CMD_REQUEST_PROTOCOL_VERSION to common.xml

auturgy:
MAV_CMD_REQUEST_PROTOCOL_VERSION should never have been in minimal, as it is not part of the minimal subset of mavlink required to establish mavlink communication.  As it is deprecated in any case, shifting it back to common where it can quietly see out it's days.


amilcarlucas (approved):

hamishwillee:
I agree. Merged.

PR #1500, opened 2020-10-02
Introduce image groups to allow for correlating images with surveys

remster:
This PR re-approaches the image->survey correlation problem - one of two that [this](https://github.com/mavlink/mavlink/pull/1319) (now stale and closed) PR aimed to solve.
I choose to start a new PR because:
- the other problem, one of lost images, has been since solved at a higher lever ([with `MAV_CMD_REQUEST_MESSAGE`](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE))
- to shake off the comment clutter the previous PR amassed.

### The problem:
In a general case a mission features multiple surveys and each survey captures multiple images. Surveys are processed to produce orthomosaics, digital elevation models or 360 panoramas. The processing software needs to pick all images pertaining to a specific survey and no image beyond it (no image from other surveys) it thus needs the images grouped per survey. It could, in theory, attempt it by careful analysis of geotags and timestamps, but this is a blunt tool and one difficult to use. 

### Proposal
This PR proposes an "image group" that it asks to be communicated both in the;
- image exifs and 
- `CAMERA_IMAGE_CAPTURED` broadcast - to relieve the processors from needing to parse exif. A most simplistic processor - e.g.: [a photo gallery](https://github.com/mavlink/qgroundcontrol/blob/srr-qgc/src/PhotoGallery/PhotoGalleryVehicleGlue.cc#L102) arguably should group the photos by survey. 

[As suggested earlier](https://github.com/mavlink/mavlink/pull/1319#issuecomment-612441126) this PR reluctantly introduces a dedicated message (`MAV_CMD_IMAGE_START_GROUP`) that allows the GCS to set a group for all followup images. It does it reluctantly because I think the image->survey correlation problem is more fundamental and deserves to be the first class citizen. Many surveys per mission is not a corner case. Survey is a a thing within a plan and natural grouping. Images are captured for a purpose and very often that purpose is automated processing. This is why I suggest to scrap `MAV_CMD_IMAGE_START_GROUP` and allow the image group within [`MAV_CMD_IMAGE_START_CAPTURE`](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE) - there are still vacant parameter slots.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1785,6 +1785,14 @@
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Stops ongoing tracking.</description>
       </entry>
+      <entry value="2020" name="MAV_CMD_IMAGE_START_GROUP" hasLocation="false" isDestination="false">
+        <description>Instruct the camera to tag all subsequently captured images with the specified group ID. This is to later correlate images with mission items (surveys). Each camera should track independent groupings (as cameras may be occupied capturing different targets at different schedules). Vehicles are encouraged to bake the group ID into the image's exif and it must report the group ID within the CAMERA_IMAGE_CAPTURED.</description>
+        <param index="1" label="Number" minValue="0">Image Group ID.</param>
+        <param index="2" reserved="true" default="NaN"/>
+        <param index="3" reserved="true" default="NaN"/>
+        <param index="4" reserved="true" default="NaN"/>
+        <param index="7" reserved="true" default="NaN"/>
+      </entry>
       <entry value="2500" name="MAV_CMD_VIDEO_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Starts video capture (recording).</description>
         <param index="1" label="Stream ID" minValue="0" increment="1">Video Stream ID (0 for all streams)</param>
@@ -5975,6 +5983,7 @@
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
       <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="int32_t" name="image_index">Zero based index of this image (i.e. a new image will have index CAMERA_CAPTURE_STATUS.image count -1)</field>
+      <field type="int32_t" name="image_group">Image Group ID (if set by MAV_CMD_IMAGE_START_GROUP preceeding this image capture) or -1 if no image group set. Image downloaders should consider grouping images into folders named after 'image_group'.</field>
       <field type="int8_t" name="capture_result">Boolean indicating success (1) or failure (0) while capturing this image.</field>
       <field type="char[205]" name="file_url">URL of image taken. Either local storage or http://foo.jpg if camera provides an HTTP interface.</field>
     </message>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1786,7 +1786,7 @@
         <description>Stops ongoing tracking.</description>
       </entry>
       <entry value="2020" name="MAV_CMD_IMAGE_START_GROUP" hasLocation="false" isDestination="false">
-        <description>Instruct the camera to tag all subsequently captured images with the specified group ID. This is to later correlate images with mission items (surveys). Each camera should track independent groupings (as cameras may be occupied capturing different targets at different schedules). Vehicles are encouraged to bake the group ID into the image's exif and it must report the group ID within the CAMERA_IMAGE_CAPTURED.</description>
+        <description>Instruct the camera to tag all subsequently captured images with the specified group ID. This is to later correlate images with mission items (surveys). Each camera should track independent groupings (as cameras may be occupied capturing different targets at different schedules). Vehicles are encouraged to bake the group ID into the image's exif and it must report the group ID within the CAMERA_IMAGE_CAPTURED. A group ID once set can only be unset by setting another group ID (with another MAV_CMD_IMAGE_START_GROUP).</description>
         <param index="1" label="Image Group" minValue="0">Image group ID (an integer)</param>
         <param index="2" reserved="true" default="NaN"/>
         <param index="3" reserved="true" default="NaN"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5983,7 +5983,6 @@
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
       <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="int32_t" name="image_index">Zero based index of this image (i.e. a new image will have index CAMERA_CAPTURE_STATUS.image count -1)</field>
-      <field type="int32_t" name="image_group">Image Group ID (if set by MAV_CMD_IMAGE_START_GROUP preceeding this image capture) or -1 if no image group set. Image downloaders should consider grouping images into folders named after 'image_group'.</field>
       <field type="int8_t" name="capture_result">Boolean indicating success (1) or failure (0) while capturing this image.</field>
       <field type="char[205]" name="file_url">URL of image taken. Either local storage or http://foo.jpg if camera provides an HTTP interface.</field>
     </message>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1787,7 +1787,7 @@
       </entry>
       <entry value="2020" name="MAV_CMD_IMAGE_START_GROUP" hasLocation="false" isDestination="false">
         <description>Instruct the camera to tag all subsequently captured images with the specified group ID. This is to later correlate images with mission items (surveys). Each camera should track independent groupings (as cameras may be occupied capturing different targets at different schedules). Vehicles are encouraged to bake the group ID into the image's exif and it must report the group ID within the CAMERA_IMAGE_CAPTURED.</description>
-        <param index="1" label="Number" minValue="0">Image Group ID.</param>
+        <param index="1" label="Image Group" minValue="0">Image group ID (an integer)</param>
         <param index="2" reserved="true" default="NaN"/>
         <param index="3" reserved="true" default="NaN"/>
         <param index="4" reserved="true" default="NaN"/>


message_definitions/v1.0/common.xml
```diff
@@ -1793,6 +1793,14 @@
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Stops ongoing tracking.</description>
       </entry>
+      <entry value="2020" name="MAV_CMD_IMAGE_START_GROUP" hasLocation="false" isDestination="false">
+        <description>Instruct the camera to tag all subsequently captured images with the specified group ID. This is to later correlate images with mission items (surveys). Each camera should track independent groupings (as cameras may be occupied capturing different targets at different schedules). Vehicles are encouraged to bake the group ID into the image's exif and it must report the group ID within the CAMERA_IMAGE_CAPTURED.</description>
+        <param index="1" label="Number" minValue="0">Image Group ID.</param>
```
hamishwillee, line 1798: I'd probably make this match the equivalent field name - label  "Group" or "Image Group" and descrition "Image group ID (an integer).

hamishwillee:
@remster The idea of making it easier for geotagging software to work out what is relevant seems a good one to me. How would the `MAV_CMD_IMAGE_START_GROUP` know when it should end the group?

Pending your answer to the above, my leaning would be to agree that the  `MAV_CMD_IMAGE_START_CAPTURE` is a better place to start the group. Then the group would apply to all images in the current capture initiated by the call and restart next time. The default would be 0 and we'd fall back to that.
Other questions 

@MatejFranceskin @dogmaphobic @DanielePettenuzzo What do you guys think of this idea - having an image group associated with a set of images for easier sorting of what belongs in a survey etc.). Is there a better way?

message_definitions/v1.0/common.xml
```diff
@@ -5982,6 +5990,7 @@
       <field type="int32_t" name="relative_alt" units="mm">Altitude above ground</field>
       <field type="float[4]" name="q">Quaternion of camera orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="int32_t" name="image_index">Zero based index of this image (i.e. a new image will have index CAMERA_CAPTURE_STATUS.image count -1)</field>
+      <field type="int32_t" name="image_group">Image Group ID (if set by MAV_CMD_IMAGE_START_GROUP preceeding this image capture) or -1 if no image group set. Image downloaders should consider grouping images into folders named after 'image_group'.</field>
```
hamishwillee, line 5993: Fields cannot be added arbitrarily in messages without breaking compatibility. For this kind of thing we add [extensions](https://mavlink.io/en/guide/define_xml_element.html#message_extensions). One implication is that we need to use 0 for the default value - ie "not supported" 

Why an int32? How many groups were you thinking might be needed? Should these iterate forever? If we're going to do that, should the group be implicitly set by the camera based on the image capture call and become a read-only value?

remster: I feared I couldn't just bold it on. Sorry, int32 was a "clipboard inheritance". I have removed it as i know it can't stay that way, but I don't want to remove the problem of the group ID being notionally present in the `CAMERA_IMAGE_CAPTURED`

remster:
> How would the `MAV_CMD_IMAGE_START_GROUP` know when it should end the group?

It would end the group only with explicit setting of a new group. I don't see terrible consequences of this, i.e.: every photo implicitly belongs to a group and that group is '0' unless `MAV_CMD_IMAGE_START_GROUP` changes it.

DonLakeFlyer:
For something like QGC it would be fairly simple to support this from a survey standpoint. Each survey just adds the additional `MAV_CMD_IMAGE_START_GROUP` prior to each survey area. And then bumps it by one for each area. A little tricky to reorder the group ids for all surveys in a plan when the user inserts/deletes surveys while editing but not super hard.

I think it might also be useful to have the group id in the vehicle log somewhere. That way post-processing geotag could have an option to stuff the group id into the exif after the fact for the case where the vehicle is not capable of futzing the exif itself.

hamishwillee:
> It would end the group only with explicit setting of a new group. I don't see terrible consequences of this, i.e.: every photo implicitly belongs to a group and that group is '0' unless MAV_CMD_IMAGE_START_GROUP changes it.

That's fine. It just needs to be stated/clear or you'll have someone else having a different implementation.

dogmaphobic:
It makes a whole lot more sense to have the "group" number as part of the MAV_CMD_IMAGE_START_CAPTURE command. As it's a command, the value (param) is 32-bit in size so the question of 32-bit versus anything else is irrelevant.

Having it as part of a MAV_CMD_IMAGE_START_CAPTURE command allows for out of band pictures to be taken and not mixed in. Meaning, you can manually shoot a picture for whatever reason and that manual capture would have its own, default group number, distinct from an ongoing survey. Why would you do such a thing is beyond me but it can be done (using a start/stop group would make something like this quite a pain in the neck).

Currently MAV_CMD_IMAGE_START_CAPTURE param1 is unused ("reserved"). This is because it used to be a "execute/not execute" command that didn't make sense and it was dropped early on, before this command was ever used. I would think it would be safe to use it for the group number. Otherwise, params 5 through 7 are free for the taking as well. Older systems will simply ignore all those params any way (0, 5, 6 and 7).

We do need a few reserved group numbers though:

0 or NaN -> not used
1 -> default group (manual capture)
2 - 100 -> Reserved for some future, unforeseen reason.

The trick though, is that this group number should be added to the CAMERA_IMAGE_CAPTURED message and that will be tricky. It's an extension past the `file_url` field. In addition, what will the `image_index` show? The total count (current value) or a count for the group?

message_definitions/v1.0/common.xml
remster: Is this what you mean: https://github.com/mavlink/mavlink/pull/1500/commits/f635bc7fe58a1630d9ec0036967b500e866d84af ?

remster:
> > It would end the group only with explicit setting of a new group. I don't see terrible consequences of this, i.e.: every photo implicitly belongs to a group and that group is '0' unless MAV_CMD_IMAGE_START_GROUP changes it.
> 
> That's fine. It just needs to be stated/clear or you'll have someone else having a different implementation.

You think this is explicit enough: https://github.com/mavlink/mavlink/pull/1500/commits/2e919aa7ebc3686782c942fa3e7f3f81267cef96

remster:
> It makes a whole lot more sense to have the "group" number as part of the MAV_CMD_IMAGE_START_CAPTURE command. 
> We do need a few reserved group numbers though:

I am very glad you agree @dogmaphobic , but we would either:
- need to extend this reasoning also to MAV_CMD_DO_SET_CAM_TRIGG_DIST (because this is the command QGC uses to capture grid survey - the number one candidate for image groups)
- migrate QGC grid surveys to use MAV_CMD_IMAGE_START_CAPTURE

> We do need a few reserved group numbers though:
> 0 or NaN -> not used
> 1 -> default group (manual capture)
> 2 - 100 -> Reserved for some future, unforeseen reason.

I can see some healthy paranoia in the last item, but i fail to see a difference between "not used" and "default group". In other words; i would propose the default is 0, which puts all photos associated with it into the default group, which is equivalent to being in no group at all, which is equivalent to be "not used". This a minor point for me and you have way more experience negotiating with firmwares.

> The trick though, is that this group number should be added to the CAMERA_IMAGE_CAPTURED message and that will be tricky. It's an extension past the `file_url` field.

Again, I am glad you agree. Do you examples of similar extensions? Presumably it can be in the extension of MAVLINK 2.0 and in MAVLINK main for 3.0

> In addition, what will the `image_index` show? The total count (current value) or a count for the group?

I think we should separate concerns. Image index has clear value in book-keeping. There can be a component that tracks received `CAMERA_IMAGE_CAPTURED` and requests missing ones and its implementation is way simpler if the `image_index` is a flat monotonic number.

LorenzMeier:
Remek, we're making it a hard requirement in the MAVLink governance process to have at least one open source implementation and a 2nd one on a roadmap of any proposal going forward. Could you point us to the repo where this is being implemented?

remster:
> Remek, we're making it a hard requirement in the MAVLink governance process to have at least one open source implementation and a 2nd one on a roadmap of any proposal going forward. Could you point us to the repo where this is being implemented?

Hi @LorenzMeier - i take it that you want to see a firmware with this proposal implemented. If that's so, apologies for not reading up, I do not have a repo with this, I can see why you might ask for one and I do have at least one in mind with which we could showcase the implementation. However, this PR has one big dilemma: `MAV_CMD_IMAGE_START_GROUP` vs `MAV_CMD_IMAGE_START_CAPTURE` and I would like to ask the guardians of MAVLINK to take a stance / express their preference so I will bring the favoured alternative.

DonLakeFlyer:
> It makes a whole lot more sense to have the "group" number as part of the MAV_CMD_IMAGE_START_CAPTURE command. As it's a command, the value (param) is 32-bit in size so the question of 32-bit versus anything else is irrelevant.

That requires changing every camera command to add a new field. Changes like this to existing commands creates havoc for QGC.

> * migrate QGC grid surveys to use MAV_CMD_IMAGE_START_CAPTURE

That is not possible. This command only supports time based capture. QGC supports time based, distance based as well as  direct cameras commands when using hover and capture.

I'm not very interesting in having to dive into rewriting complex item code in a big way to support this. It's hard enough already keeping up with required changes.

hamishwillee:
@DonLakeFlyer @dogmaphobic 
So to summarise, I think you the result is that IFF we want to do this then we would
-  be back to using a `MAV_CMD_IMAGE_START_GROUP` to set the current group name. GCS could set this before a survey to some unique value, or alongside a trigger request, and otherwise reset it to zero.
- The problem then is what [CAMERA_IMAGE_CAPTURED](https://mavlink.io/en/messages/common.html#CAMERA_IMAGE_CAPTURED) should report. I appreciate that it would be after `file_url` so we lose truncation - I don't know if we can live with that. I would say that `image_index` has to remain the total. If we wanted to record per-folder counts we'd need to provide another index.

Is that about right, and could we live with the camera images not being truncated? Generally we're sending these over fast links right?

hamishwillee:
@DonLakeFlyer @dogmaphobic Is my summary above ^^ correct?

DonLakeFlyer:
> * be back to using a `MAV_CMD_IMAGE_START_GROUP` to set the current group name. GCS could set this before a survey to some unique value, or alongside a trigger request, and otherwise reset it to zero.

I believe that is correct.

> * The problem then is what [CAMERA_IMAGE_CAPTURED](https://mavlink.io/en/messages/common.html#CAMERA_IMAGE_CAPTURED) should report. I appreciate that it would be after `file_url` so we lose truncation - I don't know if we can live with that. I would say that `image_index` has to remain the total. If we wanted to record per-folder counts we'd need to provide another index.

No comment on that. I don't use anything other than lat/lon/alt and capture_result from that message.

hamishwillee:
OK. @remster Hope that is helpful. I will ask about the second point in dev call next week.

hamishwillee:
This appears to be very stale. Last discussion in dev call was that this is on border of what MAVLink should be covering. 

@remster Do you wish to continue this. If so we can't progress without prototype that goes through to some cameras?

If not, then suggest we close it.

hamishwillee:
Closing. Feel free to reopen with end to end prototype.

PR #1501, opened 2020-10-06
it's Hamilton, move target to front again

olliw42:
#1466 has reverted #1488, which I suspect (hope :)) was unintentional. This PR corrects this.

Just as comment: The last sentence in the description could probably need better wording, I couldn't come up with a few words change, so I left it as is :).


julianoes (approved):
I'm sorry, my bad fixing the conflicts!

PR #1502, opened 2020-10-06
gimbal: use char array for strings

julianoes:
Otherwise the compiler complains.


julianoes:
Self merging as it's trivial.

hamishwillee:
FWIW good :-)

PR #1503, opened 2020-10-07
Add instance=true to some messages

olliw42:
I think this is what the instance attribute is for, adds it to some more messages, especially the gimbal v2 messages.

There might be few more places there this should have been added; I've done it just where I was kind of sure it's what it should be :)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5114,7 +5114,7 @@
     <message id="106" name="OPTICAL_FLOW_RAD">
       <description>Optical flow from an angular rate flow sensor (e.g. PX4FLOW or mouse sensor)</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
-      <field type="uint8_t" name="sensor_id">Sensor ID</field>
+      <field type="uint8_t" name="sensor_id" instance="true">Sensor ID</field>
       <field type="uint32_t" name="integration_time_us" units="us">Integration time. Divide integrated_x and integrated_y by the integration time to obtain average flow. The integration time also indicates the.</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@olliw42 Thanks very much. 

The `instance="true"` indicates that the message can contain from a number of sources (e.g. a sensor), and identifies the field that can be used to differentiate those sources. This allows the information from two sensors to be carried in the same message in separate calls, and the recipient to be able to group the messages for each sensor together.

I think that this works provided each message carries just the data for one sensor. This therefore will work for batteries, optical flow, gimbal etc. However I don't think that it will work for ESC_INFO etc, where the index is not the identify of a single ESC, but rather the index into an array containing data for multiple ESC.
Thoughts?

@amilcarlucas I would appreciate your review.

olliw42:
yeah, I was thinking about the ESC case. I concluded however that, for a given setup, each index into the array is tautologous to the id of the esc, i.e. the data of one esc will always end up in exactly the same array index ... so, if one wanted to plot the data of one esc one would need exactly this value ... 

> The instance="true" indicates that the message can contain from a number of sources (e.g. a sensor), and identifies the field that can be used to differentiate those sources. This allows the information from two sensors to be carried in the same message in separate calls, and the recipient to be able to group the messages for each sensor together.

I am not sure this is what the the attribute is for, my understanding is slightly different. At the code level of digesting a message the recipient can differentiate the different sources by the mere fact that there is an id field, without any need of any attribute. In my understanding the attribute is pure metadata which e.g. allows plotting tools to automatically produce instance specific plots.

hamishwillee:
> recipient can differentiate the different sources by the mere fact that there is an id field

I'm not sure the distinction you're trying to draw, but the id field doesn't have to be called id.

@amilcarlucas Can you please check the ESC_INFO fields that @olliw42 has marked up and convince yourself that these would work properly? I *think* it probably works for the reasons Ollie says.

amilcarlucas (approved):
yes, you nailed it.

hamishwillee:
Thanks @olliw42 @amilcarlucas!

PR #1504, opened 2020-10-07
COMMAND_ACK - remove WIP on extensions. Clarify use

hamishwillee:
These can't be WIP after so much time. The explanations also a little ambiguous.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4844,10 +4844,10 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID (of acknowledged command).</field>
       <field type="uint8_t" name="result" enum="MAV_RESULT">Result of command.</field>
       <extensions/>
-      <field type="uint8_t" name="progress">WIP: Also used as result_param1, it can be set with a enum containing the errors reasons of why the command was denied or the progress percentage or 255 if unknown the progress when result is MAV_RESULT_IN_PROGRESS.</field>
-      <field type="int32_t" name="result_param2">WIP: Additional parameter of the result, example: which parameter of MAV_CMD_NAV_WAYPOINT caused it to be denied.</field>
-      <field type="uint8_t" name="target_system">WIP: System which requested the command to be executed</field>
-      <field type="uint8_t" name="target_component">WIP: Component which requested the command to be executed</field>
+      <field type="uint8_t" name="progress">Also used as result_param1, it can be set with an enum containing the errors reasons of why the command was denied, or the progress percentage when result is MAV_RESULT_IN_PROGRESS (255 if the progress is unknown).</field>
+      <field type="int32_t" name="result_param2">Additional parameter of the result, example: which parameter of MAV_CMD_NAV_WAYPOINT caused it to be denied.</field>
+      <field type="uint8_t" name="target_system">System ID of the target recipient. This is the ID of the system that sent the command for which this COMMAND_ACK is an acknowledgement.</field>
+      <field type="uint8_t" name="target_component">Component ID of the target recipient. This is the ID of the system that sent the command for which this COMMAND_ACK is an acknowledgement.</field>
     </message>
     <message id="80" name="COMMAND_CANCEL">
       <wip/>


auturgy:
Please revert. They’re bad extensions and won’t be merged downstream. Sys ID’s will be trimmed if 0 but 0 is a valid value.

Regards,

James

hamishwillee:
Hi James/@auturgy 

I _think_ this is completely safe and reasonable - can you please have another look at the actual change I made?

- The fields went in some years ago, so almost the entire world that uses the command protocol is already updated to this file. 
- The change is ONLY to the description (ie removal of a WIP prefix). So there is no excluding code that was previously operational based on a tag that is now doing something different. 
- 0 is actually an invalid sysid for the target of a command response in 99% of time. There are a couple of broadcast commands, but not many. 
- If this did have a non-zero value set and the receiving end did not know the value and trimmed it, then there is no impact - ie this is "recipient does not know how to handle the extension case"
- If the sender does not know about the field the recipient gets 0. That is a legitimate cause for concern, except that by now every system should be setting this. 

Given that I could add back the WIP text, but would it make any difference? Fixing this will break all of use equally.

auturgy:
the target ID extensions are one of the deltas between this repo and ArduPilot's common.xml.  They will not be merged downstream.  They were introduced for arm auth, the design of which was rejected by ArduPilot.  They need to remain WIP or be removed, as they are not agreed as "common".  Anyone who has implemented WIP messages/changes has done so in the knowledge that they are not stable messages/changes.

hamishwillee:
@auturgy I have added to dev call and reverted by https://github.com/mavlink/mavlink/pull/1506 . Ie the WIP is back but the improved description remains.

I think the fields make for a better command protocol. I hope that we can discuss it in that context.

julianoes:
I would say the `target_system` and `target_component` make sense. However, I was never fond of the way arm auth was implemented. It's hard to use and implement from my experience.

olliw42:
If I may: I would say target_system and target_component are very important! By general reasons! CMDs are required by the standard to be always ACKed, and there are commands which are potentially send at relatively high rates between components on the vehicle (such as e.g. the old but still prevailing DO_MOUNT_CONTROL) ... without target this spills the complete MAVLink network, including the potentially weak telemetry link to the GCS ... Besides the target fields in ACK there would be ideally even an option to suprress ACKs at all, but I understand that's unlikely to come LOL. The issue could only be avoided iff (if and only if) there always would be targeted message counter parts, which however might be a rather larger effort and hence somewhat unrealistic.

I would wish ArduPilot would be less destructive but more cooperative when it comes to MAVLink.

auturgy:
@olliw42 my goal here isn't destructive - quite the opposite.  It may not be obvious, but my over arching goal (and that of @amilcarlucas ) is to reduce the delta between this repo and ArduPilot/mavlink.  That process is requiring me to be somewhat robust both here and downstream, but I think it is in the interests of the entire mavlink ecosystem to work towards that goal.

olliw42:
many thx for your comments and clarification. First pl note that I haven't mentioned you or anyone else in my statement but ArduPilot as the larger entity. I certainly can't know who pulls which strings, makes which decisions, or has whatever intentions, so it's certainly not meant as personal offense, and  especially not that it would be your goal to be destructive. I however can know what the situation is as regards using mavlink, and that the results of ArduPilot's actions have dire consequences. The situation is - IMHO - clearly NOT to the benefit of the mavlink ecosystem.  

E.g. this ACK wip thing is there since quite some time now, with the said consequences, as well as other wip issues and routing issues which are around since quite a long while, which e.g. totally prevent us from using the camera microservice with native ArduPilot. I believe to have had mentioned somewhere before the example that an - in my opinion - respected ArduPilot user who had made a serious effort in that matter, which was: more than a year ago. No positive response, no nothing, it's still as it is. I could mention more points but don't want to, it's sufficiently clear even though it's well covered behind the politeness of the mavlink folks.   

I understand what you are saying, and as said it's not a personal offense, but that's not the point at all. The point is that as regards the factual situation ArduPilot's attitude is not constructive and cooperative. Not only as regards wips but also other mavlink related things. I believe I can put this conclusion on objective points, and I thus believe that it is not that easily argued away. Sorry for my opinion, but that's my opinion.

:)

PR #1508, opened 2020-10-08
add command 260 for Oblique Survey

igorsgcampos:
**Problem**
A solution commonly used to increase the coverage and efficiency in mapping surveys is to have multiple cameras at oblique angles to capture a wider areas per flight line, this also has potential benefits of capturing more data from facades and tall objects. However, adding cameras increase cost, weight and complexity of the equipment, on the other hand, mapping cameras are generally capable of shooting multiple frames per second and are usually triggered about once every two seconds or less frequently in mapping missions. We could better use the hardware capabilities if we moved the camera mount between shots while decreasing the shooting distance to keep the overlap settings.

**Solution**
I decided to implement a solution that uses the already existing MAV_CMD_DO_SET_CAM_TRIGG_DIST, plus some unused parameters, 4~6, to issue MAV_CMD_DO_MOUNT_CONTROL between camera shots to move the camera from left to right in stages, right after it passes half the shooting distance, allowing time for mount actuation and camera exposure. To keep the frontlap the solution expects the GCS to adjust the trigger distance accordingly. The parameters usage are as follows:

4: number of oblique positions (this will generally be 2 or 3, but could be more, depending on desired frontlap, flying speed and hardware capabilities (how fast are the camera and mount))
5: the maximum angle to roll the camera left and right from the center, the solution calculates the interval angle between positions from this information and the previous parameter
6: the pitch angle for mounts that support pitching

**Implementation**
I implemented the proposed solution here: https://github.com/PX4/Firmware/pull/15882 and added to QGC here: https://github.com/mavlink/qgroundcontrol/pull/9103

**Alternative**
Other way would be to add a new command, but since it has many aspects that are essentially the same as the CAM_TRIGG_DIST, I thought it would be better to add a few parameters to the existing command, as it would also allow it to work under Mavlink 1 with the limited number of 255 commands.

**Test data / coverage**
I tested it with the gazebo_typhoon_h480 model and checked the outcome from the video stream in QGC.

**Additional context**
Later I plan to write a paper comparing the gains in area coverage and any accuracy change from using this solution with relation to the traditional survey mode. I am feeling confident about a positive outcome.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1589,7 +1589,7 @@
       </entry>
       <entry value="260" name="MAV_CMD_OBLIQUE_SURVEY" hasLocation="false" isDestination="false">
         <wip/>
-        <description>Mission command to set a Camera Auto Mount Pivoting Oblique Survey. The camera is triggered each time this distance is exceeded, then the mount moves to the next position. Params 4~6 set-up the angle limits and number of positions for oblique survey, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV). This command can also be used to set the shutter integration time for the camera.</description>
+        <description>Mission command to set a Camera Auto Mount Pivoting Oblique Survey (Replaces CAM_TRIGG_DIST for this purpose). The camera is triggered each time this distance is exceeded, then the mount moves to the next position. Params 4~6 set-up the angle limits and number of positions for oblique survey, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV). This command can also be used to set the shutter integration time for the camera.</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="0" increment="1" default="0">Camera shutter integration time. 0 to ignore</param>
         <param index="3" label="Min Interval" units="ms" minValue="0" maxValue="10000" increment="1" default="0">The minimum interval in which the camera is capable of taking subsequent pictures repeatedly. 0 to ignore.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1345,13 +1345,13 @@
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">
-        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. Params 4~6 enable a pseudo-oblique mode, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup.</description>
+        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. WIP: Params 4~6 enable a pseudo-oblique mode, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup.</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Positions" minValue="0" increment="1" default="0">Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (Greater or equal to 2 enables this feature)</param>
-        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is greater or equal to 2) </param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is greater or equal to 2)</param>
+        <param index="4" label="Positions" minValue="0" increment="1" default="0">WIP: Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (Greater or equal to 2 enables this feature)</param>
+        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">WIP: Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is greater or equal to 2) </param>
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">WIP: Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is greater or equal to 2)</param>
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1372,13 +1372,13 @@
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">
-        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. WIP: Params 4~6 enable a pseudo-oblique mode, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV).</description>
+        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera.</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Positions" minValue="0" increment="1" default="0">WIP: Total number of roll positions at which the camera will capture photos (images captures spread evenly across the limits defined by param5). Feature enabled if value is greater or equal to 2.</param>
-        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">WIP: Angle limits that the camera can be rolled to left and right of center. Enabled if param 4 is greater or equal to 2.</param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">WIP: Fixed pitch angle that the camera will hold in pseudo oblique mode if the mount supports it. Enabled if param 4 is greater or equal to 2.</param>
+        <param index="4">Empty</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
@@ -1586,6 +1586,17 @@
         <param index="5" label="Latitude/X">Latitude/X position.</param>
         <param index="6" label="Longitude/Y">Longitude/Y position.</param>
         <param index="7" label="Altitude/Z">Altitude/Z position.</param>
+      </entry>
+      <entry value="260" name="MAV_CMD_OBLIQUE_SURVEY" hasLocation="false" isDestination="false">
+        <wip/>
+        <description>Mission command to set a Camera Auto Mount Pivoting Oblique Survey. The camera is triggered each time this distance is exceeded, then the mount moves to the next position. This command can also be used to set the shutter integration time for the camera. Params 4~6 set-up the angle limits and number of positions for oblique survey, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV).</description>
+        <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
+        <param index="2" label="Shutter" units="ms" minValue="0" increment="1" default="0">Camera shutter integration time. 0 to ignore</param>
+        <param index="3" label="Min Interval" units="ms" minValue="0" maxValue="10000" increment="1" default="0">The minimum interval in which the camera is capable of taking subsequent pictures repeatedly. 0 to ignore.</param>
+        <param index="4" label="Positions" minValue="2" increment="1">Total number of roll positions at which the camera will capture photos (images captures spread evenly across the limits defined by param5).</param>
+        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">Angle limits that the camera can be rolled to left and right of center.</param>
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">Fixed pitch angle that the camera will hold in pseudo oblique mode if the mount supports it.</param>
+        <param index="7">Empty</param>
       </entry>
       <entry value="300" name="MAV_CMD_MISSION_START" hasLocation="false" isDestination="false">
         <description>start running a mission</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1589,13 +1589,13 @@
       </entry>
       <entry value="260" name="MAV_CMD_OBLIQUE_SURVEY" hasLocation="false" isDestination="false">
         <wip/>
-        <description>Mission command to set a Camera Auto Mount Pivoting Oblique Survey. The camera is triggered each time this distance is exceeded, then the mount moves to the next position. This command can also be used to set the shutter integration time for the camera. Params 4~6 set-up the angle limits and number of positions for oblique survey, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV).</description>
+        <description>Mission command to set a Camera Auto Mount Pivoting Oblique Survey. The camera is triggered each time this distance is exceeded, then the mount moves to the next position. Params 4~6 set-up the angle limits and number of positions for oblique survey, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV). This command can also be used to set the shutter integration time for the camera.</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="0" increment="1" default="0">Camera shutter integration time. 0 to ignore</param>
         <param index="3" label="Min Interval" units="ms" minValue="0" maxValue="10000" increment="1" default="0">The minimum interval in which the camera is capable of taking subsequent pictures repeatedly. 0 to ignore.</param>
         <param index="4" label="Positions" minValue="2" increment="1">Total number of roll positions at which the camera will capture photos (images captures spread evenly across the limits defined by param5).</param>
         <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">Angle limits that the camera can be rolled to left and right of center.</param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">Fixed pitch angle that the camera will hold in pseudo oblique mode if the mount supports it.</param>
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">Fixed pitch angle that the camera will hold in oblique mode if the mount is actuated in the pitch axis.</param>
         <param index="7">Empty</param>
       </entry>
       <entry value="300" name="MAV_CMD_MISSION_START" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1349,7 +1349,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Positions" minValue="0" increment="1" default="0">Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (greater or equal to 2 enables this feature) (more info at: https://github.com/mavlink/mavlink/pull/1508)</param>
+        <param index="4" label="Positions" minValue="0" increment="1" default="0">Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (Greater or equal to 2 enables this feature)</param>
         <param index="5" label="Roll Angle" units="deg" minValue="0" default="20">Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is greater or equal to 2) </param>
         <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="-90">Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is greater or equal to 2)</param>
         <param index="7">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1350,8 +1350,8 @@
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
         <param index="4" label="Positions" minValue="0" increment="1" default="0">Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (Greater or equal to 2 enables this feature)</param>
-        <param index="5" label="Roll Angle" units="deg" minValue="0" default="20">Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is greater or equal to 2) </param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="-90">Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is greater or equal to 2)</param>
+        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is greater or equal to 2) </param>
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is greater or equal to 2)</param>
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1345,13 +1345,13 @@
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">
-        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. WIP: Params 4~6 enable a pseudo-oblique mode, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup.</description>
+        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. WIP: Params 4~6 enable a pseudo-oblique mode, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV).</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Positions" minValue="0" increment="1" default="0">WIP: Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (Greater or equal to 2 enables this feature)</param>
-        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">WIP: Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is greater or equal to 2) </param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">WIP: Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is greater or equal to 2)</param>
+        <param index="4" label="Positions" minValue="0" increment="1" default="0">WIP: Total number of roll positions at which the camera will capture photos (images captures spread evenly across the limits defined by param5). Feature enabled if value is greater or equal to 2.</param>
+        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">WIP: Angle limits (in degrees) that the camera can be rolled to left and right of center. Enabled if param 4 is greater or equal to 2.</param>
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">WIP: Fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. Enabled if param 4 is greater or equal to 2.</param>
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1345,13 +1345,13 @@
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">
-        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. Params 4~6 enable additional image captures at different gimbal orientations at each trigger point (this allows, for example, automatically roll of the camera between shots to emulate an oblique camera setup).</description>
+        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. Params 4~6 enable a special mode (CAMPOS) where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup.</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Positions" minValue="0" increment="1" default="0">Number of additional gimbal positions at which to capture photos at each trigger point. The gimbal orientations for each capture are defined by params 5 (roll) and param 6 (pitch). Default is 0 (gimbal orientation is fixed so only default position is captured)</param>
-        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">The angle limits (in degrees) that the camera will be rolled to left and right of the centerline.</param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">The fixed pitch angle that the camera will hold while capturing images at different roll angles (if the mount supports it).</param>
+        <param index="4" label="Positions" minValue="0" increment="1" default="0">Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (>= 2 enables this feature) (more info at: https://github.com/mavlink/mavlink/pull/1508)</param>
+        <param index="5" label="Roll Angle" units="deg" minValue="0" default="20">Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is >= 2) </param>
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="-90">Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is >= 2)</param>
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1345,13 +1345,13 @@
         <param index="7" label="Mode" enum="MAV_MOUNT_MODE">Mount mode.</param>
       </entry>
       <entry value="206" name="MAV_CMD_DO_SET_CAM_TRIGG_DIST" hasLocation="false" isDestination="false">
-        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. Params 4~6 enable a special mode (CAMPOS) where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup.</description>
+        <description>Mission command to set camera trigger distance for this flight. The camera is triggered each time this distance is exceeded. This command can also be used to set the shutter integration time for the camera. Params 4~6 enable a pseudo-oblique mode, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup.</description>
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4" label="Positions" minValue="0" increment="1" default="0">Extended parameter to specify the total number of positions the camera will capture photos by rolling repeatedly to emulate an oblique camera, providing an increased HFOV. (>= 2 enables this feature) (more info at: https://github.com/mavlink/mavlink/pull/1508)</param>
-        <param index="5" label="Roll Angle" units="deg" minValue="0" default="20">Extended parameter to specify the angle limits (in degrees) that the camera will be rolled to left and right of center in pseudo oblique mode. (Enabled if param 4: Positions is >= 2) </param>
-        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="-90">Extended parameter to specify the fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. (Enabled if param 4: Positions is >= 2)</param>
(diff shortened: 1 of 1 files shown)


igorsgcampos:
I would appreciate if someone could mark this with the dev-call label for discussion in the next meeting.

hamishwillee:
@igorsgcampos I have added to the dev call agenda, but it is unlikely that we will get to this issue as we plan to have an extended discussion around another matter.

My initial reactions are:
- Solution is pretty clever :-)
- Does this need to be supported by MAVLink 1? I would say not.
- Feels like we're adding very use-case specific functionality to what is supposed to be a "standard" API. I worry that we're potentially complicating the API, and we'll create edge cases for implementers.

But I'm just a facilitator :-). @julianoes @auturgy FYI - any thoughts on thsi?

igorsgcampos:
> @igorsgcampos I have added to the dev call agenda, but it is unlikely that we will get to this issue as we plan to have an extended discussion around another matter.
> 
> My initial reactions are:
> 
> * Solution is pretty clever :-)
> * Does this need to be supported by MAVLink 1? I would say not.
> * Feels like we're adding very use-case specific functionality to what is supposed to be a "standard" API. I worry that we're potentially complicating the API, and we'll create edge cases for implementers.
> 
> But I'm just a facilitator :-). @julianoes @auturgy FYI - any thoughts on this?

Would you think this would better fit a new command? What are your thoughts on that?

hamishwillee:
> Would you think this would better fit a new command? What are your thoughts on that?

I think it would feel less like hijacking the existing behaviour and more obvious when/how this should be used. I'll see if I can get some discussion in the meeting.

@DonLakeFlyer is also a good person to comment.

igorsgcampos:
> > Would you think this would better fit a new command? What are your thoughts on that?
> 
> I think it would feel less like hijacking the existing behaviour and more obvious when/how this should be used. I'll see if I can get some discussion in the meeting.
> 
> @DonLakeFlyer is also a good person to comment.

Thanks, I made a small video of the concept with the SITL. It is available at the following link:
https://photos.app.goo.gl/PDEye5Jx66f4f9E46

Thanks for your attention and support!

julianoes:
Firstly, @igorsgcampos nice pull requests with context, video, that's nice to see! And thanks a lot for the contribution!

Now if I had to do this, I would have taken the approach of assembling many mission items together to do the same.
E.g.:
1. NAV_WAYPOINT
2. DO_MOUNT_CONTROL
3. IMAGE_START_CAPTURE
4. DO_MOUNT_CONTROL
5. IMAGE_START_CAPTURE
6. DO_MOUNT_CONTROL
7. IMAGE_START_CAPTURE
Back to 1.

This gives you full control, however, has the disadvantage that you have to pre-generate it all and deal with huge mission item lists.

So considering that I can see how your solution is simple and enabling this use case. And given you already added support in PX4 and QGC, I'm ok with the proposal.

auturgy:
This consumes 3 params for a very specific use case.  Is it possible to maintain the intent, but be less specific re "CAMPOS".  Really we're just linking a pose to the trigger command, which could be generalised (unless I'm not understanding properly)

hamishwillee:
@igorsgcampos Thanks for coming to the meeting last night - really helps to have people come and explain their use cases.

https://github.com/mavlink/mavlink/pull/1508#issuecomment-708206963 above summarises the response - we'd like this to be modified to allow more "generic" gimbal control at the trigger point rather than this specific case - if at all possible (it might not be). I have done the FIRST part of this - removing the "CAMPOS" terminology. 

My understanding of this is that at the moment(please confirm):
- you set the number of positions in param 4. I THINK this means that you capture the normal center position, then this many additional images to either side of the camera centerline - ie 4 to left and 4 to right if the value was 4
- The roll angle is set as "limits", so I think that you're saying this is the max angle and you'd take images at this angle divided y the number of positions - correct?
- The pitch angle is fixed - I think you're saying that the camera will go to this position for all the additional pictures and then do the roll. Is that correct?

So what do you think could be more general? If I'm correct above we only allow roll control and limited pitch control. Note sure how you (or if you would want to) specify flexible pitch control and a "scan" of roll at the same time.

igorsgcampos:
> @igorsgcampos Thanks for coming to the meeting last night - really helps to have people come and explain their use cases.
> 
> [#1508 (comment)](https://github.com/mavlink/mavlink/pull/1508#issuecomment-708206963) above summarises the response - we'd like this to be modified to allow more "generic" gimbal control at the trigger point rather than this specific case - if at all possible (it might not be). I have done the FIRST part of this - removing the "CAMPOS" terminology.
> 
> My understanding of this is that at the moment(please confirm):
> 
> * you set the number of positions in param 4. I THINK this means that you capture the normal center position, then this many additional images to either side of the camera centerline - ie 4 to left and 4 to right if the value was 4

The way I actually thought would be more intuitive is if this parameters specify the total number of positions the camera will assume to capture images. I updated the parameter description to better reflect it. I also added a reference to this PR (don't know if it is OK, I can drop it, but the idea is that if somebody is confused they can come up here and see the demo video)

> * The roll angle is set as "limits", so I think that you're saying this is the max angle and you'd take images at this angle divided y the number of positions - correct?

You got the concept right, but this limit is for left and right, so I will actually divide twice this angle by the number of positions. (i.e. if you use 3 positions and 20 deg roll limit the camera will take a picture at -20, another at 0 and the last at 20 and repeat the cycle until the end of the command)

> * The pitch angle is fixed - I think you're saying that the camera will go to this position for all the additional pictures and then do the roll. Is that correct?

Yes, the camera will remain at this pitch angle, without swiveling up and down.

> 
> So what do you think could be more general? If I'm correct above we only allow roll control and limited pitch control. Note sure how you (or if you would want to) specify flexible pitch control and a "scan" of roll at the same time.

The challenge of having this enabled for both roll and pitch is that we would have roll positions multiplied by pitch positions, which - if possible - would be the natural way of implementing this, but then the camera trigger distance would be divided by the total number of positions and this would likely be too much for the hardware capabilities of either gimbal or camera, unless the vehicle flew in a considerable slower speed, which would defeat the purpose of adding this feature.

If the intent to control pitch is to achieve better 3D models then the best approach would probably be to set a fixed pitch angle at -45~-70 and have the sidelap increased so that each flight line would capture both sides of the objects at enough overlap. Or even employing a double grid, where the rolling motion would then suffice.

DonLakeFlyer:
> This consumes 3 params for a very specific use case. Is it possible to maintain the intent, but be less specific re "CAMPOS". Really we're just linking a pose to the trigger command, which could be generalised (unless I'm not understanding properly)

I agree.

igorsgcampos:
> > This consumes 3 params for a very specific use case. Is it possible to maintain the intent, but be less specific re "CAMPOS". Really we're just linking a pose to the trigger command, which could be generalised (unless I'm not understanding properly)
> 
> I agree.

What are your thoughts on the latest version?

DonLakeFlyer:
> What are your thoughts on the latest version?

It just seems like a really really (really?) specific use case jammed into CAM_TRIGGER_DISTANCE which is a very generic command.

igorsgcampos:
> > What are your thoughts on the latest version?
> 
> It just seems like a really really (really?) specific use case jammed into CAM_TRIGGER_DISTANCE which is a very generic command.

Oh, I thought you were talking about being less specific in the parameters description, as was the case in the dev-call.

I understand your point of view, the idea is to increase the coverage and efficiency of the vehicles, so it might have a pretty big adoption. We also discussed the possibility of creating missions with individual commands instead of this solution, but agreed that in missions with thousands of pictures it would have multiple thousands of waypoints, thus being unpractical.

DonLakeFlyer:
To me this concept should be encapsulated in sort of of "camera pose" system/messages. Which then affect all camera commands. So you specify a pose and then do your normal trigger distance stuff. Or have camera commands associated with a pose if or something. That what you aren't adding really detailed stuff to all camera commands.

igorsgcampos:
> To me this concept should be encapsulated in sort of of "camera pose" system/messages. Which then affect all camera commands. So you specify a pose and then do your normal trigger distance stuff. Or have camera commands associated with a pose if or something. That what you aren't adding really detailed stuff to all camera commands.

How would that work? In this concept the camera is not kept static in a pose, so we would have several camera pose messages, one for each picture?

igorsgcampos:
> > > Would you think this would better fit a new command? What are your thoughts on that?
> > 
> > 
> > I think it would feel less like hijacking the existing behaviour and more obvious when/how this should be used. I'll see if I can get some discussion in the meeting.
> > @DonLakeFlyer is also a good person to comment.
> 
> Thanks, I made a small video of the concept with the SITL. It is available at the following link:
> https://photos.app.goo.gl/PDEye5Jx66f4f9E46
> 
> Thanks for your attention and support!

Here is a video illustrating how it works.

hamishwillee:
@DonLakeFlyer What you're suggesting does not meet the use case. 

The use case is to take a number of images on every camera trigger at a spread of yaws (and with a specific fixed pitch). Emulating that with a separate camera pose message would be "hard". You'd have to manage every single camera trigger as its own waypoint - e.g. fly to point, set yaw, take photo (repeat at different yaw as needed), go to next point. etc. 

So if we want this we either need to have it embedded in trigger message as shown and handled by flight stack, or have a separate message that somehow daisy-chains gimbal behaviour to a trigger event, or have the GCS magically generate the sequence. 

A GCS might well be the right place for this - how about it? :-)

DonLakeFlyer:
> To me this concept should be encapsulated in sort of of "camera pose" system/messages. _Which then affect all camera commands._

Thinking about this a bit more I think the italicized part is wrong. But I still think the concept of defining a pose externally somehow make more sense. Imagine something like this:

* A "pose" puts the gimbal though some set of movements possibly taking an image at each movement stop. The definition of how you define a pose is TBD. This pull has an example of one possible pose definition. Many more could be created.
* A "pose" has an id associated with it which allows it to be referenced
* The TRIGGER_DISTANCE command specifies the pose id it wants to use (if any)

This way you aren't locking down TRIGGER_DISTANCE to do only this one possible set of gimbal stops for the rest of it's life. You could also add pose ids to thing like the video commands. Not sure if they make sense there or not, but it would be easier this way.

Also the concept of a "pose" with multiple steps is way more complicated to do correctly than what is currently defined here. Examples:
* Gimbals take some amount of time to finish moving to a new position. You can't take the picture until you know the gimbal is done moving. How do you do that?
* Cameras can only take images one after another above a certain rate. If you try to shoot too fast between two images, the second capture will just fail. This means a "pose" must understand how fast a camera can shoot. You see some of this in QGC survey setup where the time between images is a known then and it will warn if you are trying to fly a survey for which camera capture is faster than the camera can physically do.

All of that said, conceptually I get the interest, but it all seems really complicated to pull off well.

DonLakeFlyer:
If there is interest to do something fancy like this I would include way more folks who do surveys for a living (that is not me) to comment on what would be needed from a feature standpoint.

igorsgcampos:
> * A "pose" puts the gimbal though some set of movements possibly taking an image at each movement stop. The definition of how you define a pose is TBD. This pull has an example of one possible pose definition. Many more could be created.
> * A "pose" has an id associated with it which allows it to be referenced
> * The TRIGGER_DISTANCE command specifies the pose id it wants to use (if any)

From what I understand from this concept there would be a set of camera pose definitions, to be given by other commands. Afterwards, the TRIGGER_DISTANCE would then reference the pose id it wanted to use. The problem is that the intent is to trigger the camera in different poses along the same flight line, following this design it would only be feasible to do so by having TRIGGER_DISTANCE commands running in parallel and sync to properly coordinate the camera movements and triggers, which would be highly complex and unpractical, if not unfeasible.

> 
> This way you aren't locking down TRIGGER_DISTANCE to do only this one possible set of gimbal stops for the rest of it's life. You could also add pose ids to thing like the video commands. Not sure if they make sense there or not, but it would be easier this way.
> 

Here I presented merely an example of a possible configuration, but the number of positions, roll and pitch angles are what determine the behavior of the camera/mount and are freely configurable by the user. The purpose being clear: acquiring images in oblique angles, for the benefits of having both a wider horizontal field of view and better information of facades of high objects. So in essence these three parameters define a set of poses to be subsequently used while flying along a single line in a survey mission.

> Also the concept of a "pose" with multiple steps is way more complicated to do correctly than what is currently defined here. Examples:
> 
> * Gimbals take some amount of time to finish moving to a new position. You can't take the picture until you know the gimbal is done moving. How do you do that?

That has to come from the experience of the user (or feedback from the mount), but given that servos usually move at approximately 60 deg / 0.1 sec the actuation speed will likely exceed the camera's shot-to-shot capabilities. The exposure time is also quite fast in mapping missions, in the order of a few thousands of milliseconds, otherwise you cause too much motion blur and this degrades the final output of the project.

Given this, in my [PX4 PR](https://github.com/PX4/Firmware/pull/15882) I chose to keep the design simple and send the mount control commands right after the the vehicle passes half the trigger distance. A more elaborate approach is possible, by using camera trigger feedback and a delay before moving, depending in which point of the curtain the feedback is given by the camera, but this would require more set-up from the user.

> * Cameras can only take images one after another above a certain rate. If you try to shoot too fast between two images, the second capture will just fail. This means a "pose" must understand how fast a camera can shoot. You see some of this in QGC survey setup where the time between images is a known then and it will warn if you are trying to fly a survey for which camera capture is faster than the camera can physically do.

That is why, following the Occam's razor principle, I chose to not mess up with what is already in place, thus, in my [QGC PR](https://github.com/mavlink/qgroundcontrol/pull/9103) I compute new trigger distances and adjusted side footprints for the generation of the mission commands, therefore allowing the warning already in place to remain working even with the addition of this functionality.

> 
> All of that said, conceptually I get the interest, but it all seems really complicated to pull off well.

I understand this could cause confusion, which is why I chose to send these three PRs in parallel, so things could be clearer on how it would work in the end.

igorsgcampos:
> If there is interest to do something fancy like this I would include way more folks who do surveys for a living (that is not me) to comment on what would be needed from a feature standpoint. Start with the feature definition not the mavlink spec for it.

Actually I got this idea as I work with surveys myself. This is something that I would love to have (and decided to share with others) for both the increased efficiency/coverage of the vehicle in regular surveys and much more in 3D mapping projects, as it can take two (or more) oblique pictures plus a NADIR in a single pass, giving it potential to reduce 3D surveys by a factor of 2 or more: normally for good results I use a double grid with the camera at an angle (pitch) and high overlap settings (in practice if you want 70% in the final result you fly at 85% so that you end up with 70% given the UAV captures only one side of the buildings per line) plus a NADIR flight. This PR would solve it with a double grid at regular overlap and dismiss the need of a NADIR flight.

igorsgcampos:
Basically this is what I am trying to emulate(for the roll axis), in an elegant and lightweight alternative to:
image
image
More info on the benefits of an oblique camera is avalable at the following papers (accompanied by their abstracts):

>Nesbit, P.R.; Hugenholtz, C.H. Enhancing UAV–SfM 3D Model Accuracy in High-Relief Landscapes by Incorporating Oblique Images. Remote Sens. 2019, 11, 239.

>Complex landscapes with high topographic relief and intricate geometry present challenges for complete and accurate mapping of both lateral (x, y) and vertical (z) detail without deformation. Although small uninhabited/unmanned aerial vehicles (UAVs) paired with structure-from-motion (SfM) image processing has recently emerged as a popular solution for a range of mapping applications, common image acquisition and processing strategies can result in surface deformation along steep slopes within complex terrain. Incorporation of oblique (off-nadir) images into the UAV–SfM workflow has been shown to reduce systematic errors within resulting models, but there has been no consensus or documentation substantiating use of particular imaging angles. To address these limitations, we examined UAV–SfM models produced from image sets collected with various imaging angles (0–35°) within a high-relief ‘badland’ landscape and compared resulting surfaces with a reference dataset from a terrestrial laser scanner (TLS). More than 150 UAV–SfM scenarios were quantitatively evaluated to assess the effects of camera tilt angle, overlap, and imaging configuration on the precision and accuracy of the reconstructed terrain. Results indicate that imaging angle has a profound impact on accuracy and precision for data acquisition with a single camera angle in topographically complex scenes. Results also confirm previous findings that supplementing nadir image blocks with oblique images in the UAV–SfM workflow consistently improves spatial accuracy and precision and reduces data gaps and systematic errors in the final point cloud. Subtle differences among various oblique camera angles and imaging patterns suggest that higher overlap and higher oblique camera angles (20–35°) increased precision and accuracy by nearly 50% relative to nadir-only image blocks. We conclude by presenting four recommendations for incorporating oblique images and adapting flight parameters to enhance 3D mapping applications with UAV–SfM in high-relief terrain.

>Paolo Rossi, Francesco Mancini, Marco Dubbini, Francesco Mazzone & Alessandro Capra (2017) Combining nadir and oblique UAV imagery to reconstruct quarry topography: methodology and feasibility analysis, European Journal of Remote Sensing, 50:1, 211-221, DOI: 10.1080/22797254.2017.1313097

>The feasibility of unmanned-aerial-vehicle-based photogrammetry was assessed for the reconstruction of high-resolution topography and geomorphic features of quarries by nadir and off-nadir imagery. The test site was a quarry located in the rural area of Turi (Bari, southern Italy). Two processing scenarios were created. Nadir images were initially used, and images acquired with off-nadir angles were added. An accurate set of ground control points (GCPs) were surveyed for both georeferencing purposes and validation processes. In the reconstruction of the surfaces, an accuracy of a few centimeters was achieved in the final positioning of point clouds representing the main geometries of quarry environment. However, greatest differences were found along the edges or the lines characterized by sudden slope changes. To better understand such results, some characteristic quarry shapes depicted by both the scenarios were compared to those surveyed by a total station used as an independent benchmark technique. It allowed to define the benefits introduced by the joint use of nadir and oblique images in the delineation of quarry shapes, surface discontinuities and better descriptions of sub-vertical walls. Beside the evaluation of benefits introduced by use of oblique cameras, the effectiveness of the proposed methodology was also discussed with alternative technologies. Unmanned aerial platforms represent an effective solution, with the need for few accurate GCPs.

igorsgcampos:
Hopefully there is enough evidence showcasing the benefits of the proposed PR and we will agree on the merit of merging it.

@hamishwillee, please let me know if the changes in the parameters description are sufficient to satisfy the generalization suggested in the dev-call and if there is anything else needed for merging.

DonLakeFlyer:
> From what I understand from this concept there would be a set of camera pose definitions, to be given by other commands. Afterwards, the TRIGGER_DISTANCE would then reference the pose id it wanted to use. The problem is that the intent is to trigger the camera in different poses along the same flight line, following this design it would only be feasible to do so by having TRIGGER_DISTANCE commands running in parallel and synchrony to properly coordinate the camera movements and triggers, which would be highly complex and unpractical, if not unfeasible.

Sorry but I don't understand any of that. I don't see how an indirect reference through an id to what could be the exact same set of information you propose to put directly in the message could cause any problems.

Regarding gimbal/camera trigger timing issues: Have you tried it by manually setting up a single transect with waypoints, mount controls and camera triggers, flying it for and looking at the results? You should be able to create in a mission exactly what your automatic thing does. I assume it looks something like this:
* WP, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTRO, CAPTURE_IMAGE
* WP, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTRO, CAPTURE_IMAGE
* WP, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTRO, CAPTURE_IMAGE
* ... till you get to the end of the transect
You would only need to test a couple for real to see if right off the bat you run into images taken while the gimbal is moving and/or the camera puking because the gimbal is really fast which means the image captures are right on top of each other.

igorsgcampos:
So I guess we misunderstood each other about the first part.

> Regarding gimbal/camera trigger timing issues: Have you tried it by manually setting up a single transect with waypoints, mount controls and camera triggers, flying it for and looking at the results? You should be able to create in a mission exactly what your automatic thing does. I assume it looks something like this:
> * WP, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTRO, CAPTURE_IMAGE
> * WP, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTRO, CAPTURE_IMAGE
> * WP, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTROL, CAPTURE_IMAGE, MOUNT_CONTRO, CAPTURE_IMAGE
> * ... till you get to the end of the transect
> 
> You would only need to test a couple repetitions for real to see if right off the bat you run into images taken while the gimbal is moving and/or the camera puking because the gimbal is really fast which means the image captures are right on top of each other.

This is the problem we spoke about having thousands of waypoints, cluttering the mission, and maybe not working for some hardware/firmwares. More importantly, this would involve stopping at the waypoints, which, again, would defeat the purpose of this: increasing coverage and efficiency. Holding at waypoints is not possible for planes, which would also limit the vehicles this could be used with. Alternatively, if one set a waypoint at the beginning and another at the end of the transect it would not allow for control of the trigger distance.

DonLakeFlyer:
> This is the problem we spoke about having thousands of waypoints, cluttering the mission, and maybe not working for some hardware/firmwares.

I understand that. This has nothing to do with that. This has to with whether you tried the command sequence on a real vehicle to see if you run into problems like this:
* The image being taken while the gimbal is still moving to the position you asked. As far as I understand the gimbal change command will return immediately, it doesn't wait for the gimbal to finish moving. The firmware will then execute the next command immediately which is image capture. How is that image not taken while the gimbal is still moving?
* If you can get past the first problem of not waiting for the gimbal then does this really work as well: You do the first gimbal move, trigger image, second gimbal move, trigger image, third gimbal move, trigger camera. Can the camera trigger taking images that fast? From my experience talking with survey users and the fact that the need to warning users about triggering cameras to quickly was a requested feature in QGC the answer as far as I understand it is no. That is not possible.

As far as I can tell from how things work in reality your images are either going to be taken at the wrong orientations or possibly just fail based on gimbal speed (which is never instantaneous) and/or how the camera deals with fast camera triggering. SITL proves nothing about whether the concept will work in reality.

DonLakeFlyer:
Also to implement this capability in a completely generic manner I think something like this should be done instead:
* BEGIN_SUBROUTINE, END_SUBROUTINE commands - This allows you to define a command sequence which can be reused as needed. The subroutines would be identified by an integer id.
* SUBROUTINE_TRIGGER_DISTANCE - This works just like CAMERA_TRIGGER_DISTANCE with same params but also include a subroutine id command 

So you use BEGIN_SUBROUTINE/END_SUBROUTINE to build your gimbal move, camera triggers sequence. And you run the survey using SUBROUTINE_TRIGGER_DISTANCE instead of CAMERA_TRIGGER_DISTANCE. This has the benefit of being completely generic. Probably makes sense to add SUBROUTINE_TRIGGER_TIME as well. You could add a SUBROUTINE_TRIGGER command to trigger it anytime you want as well. You can even use this to do things like set up some fancy gimbal/camera capture sequence in a subroutine and the trigger it from an RC switch and GCS ui button. This would allow you to fly to some plane manually looking for something and then when you found it you can trigger a known pattern to capture information.

igorsgcampos:
> * The image being taken while the gimbal is still moving to the position you asked. As far as I understand the gimbal change command will return immediately, it doesn't wait for the gimbal to finish moving. The firmware will then execute the next command immediately which is image capture. How is that image not taken while the gimbal is still moving?

This is why I send the command to the gimbal at halfway the trigger distance, to allow time for the movement to occur.

> * If you can get past the first problem of not waiting for the gimbal then does this really work as well: You do the first gimbal move, trigger image, second gimbal move, trigger image, third gimbal move, trigger camera. Can the camera trigger taking images that fast? From my experience talking with survey users and the fact that the need to warning users about triggering cameras to quickly was a requested feature in QGC the answer as far as I understand it is no. That is not possible.

I took references of the continuous shooting rates of some of the most used cameras for aerial surveys, such as the a6000 (11fps), RX0 (5.5fps), RXR2 (5fps), a7 (10fps).

> 
> As far as I can tell from how things work in reality your images are either going to be taken at the wrong orientations or possibly just fail based on gimbal speed (which is never instantaneous) and/or how the camera deals with fast camera triggering. SITL proves nothing about whether the concept will work in reality.

Based on these references, if one uses a vehicle that flies at approximately 15 m/s and wants a GSD of 3 cm, using the a6000 with a 20 mm lens the vehicle would have to fly at 150 m and the vertical image footprint would be approximately 118 m, which for a configured 80% frontlap would result in 23.6 m trigger distance, if one sets the configuration for 3 positions, the trigger distance would be reduced to 7.86 m, meaning the camera would be triggered approximately twice a second, which is well below its specifications. This is not the best camera of the mentioned ones, yet it easily has enough power to perform such a mission.

As I already mentioned, the servos generally move at approximately 60 deg / 0.1 sec, which is more than enough for such a mission, where it would have approximately 250 ms to move into position before the camera is triggered. This figure is for my present, naive implementation, which I chose for its simplicity, but having a properly configured camera feedback, the servo would be able to move immediately after the delay for the end of the exposure, depending on which point of the curtain the signal is sent by the camera, and considering the exposure is in the order of few milliseconds, this time would be extended to nearly half a second.

Sometimes the limitation is not in the camera or servo, but in the setup of the user, who did not properly adjust the firmware to match the camera shooting capabilities, when the standard values are pretty conservative. So unless both camera and servo manufacturers are presenting highly misleading specs, this is perfectly capable of working with big tolerance margins.

igorsgcampos:
> Also to implement this capability in a completely generic manner I think something like this should be done instead:
> 
> * BEGIN_SUBROUTINE, END_SUBROUTINE commands - This allows you to define a command sequence which can be reused as needed. The subroutines would be identified by an integer id.
> * SUBROUTINE_TRIGGER_DISTANCE - This works just like CAMERA_TRIGGER_DISTANCE with same params but also include a subroutine id command
> 
> So you use BEGIN_SUBROUTINE/END_SUBROUTINE to build your gimbal move, camera triggers sequence. And you run the survey using SUBROUTINE_TRIGGER_DISTANCE instead of CAMERA_TRIGGER_DISTANCE. This has the benefit of being completely generic. Probably makes sense to add SUBROUTINE_TRIGGER_TIME as well. You could add a SUBROUTINE_TRIGGER command to trigger it anytime you want as well. You can even use this to do things like set up some fancy gimbal/camera capture sequence in a subroutine and the trigger it from an RC switch and GCS ui button. This would allow you to fly to some plane manually looking for something and then when you found it you can trigger a known pattern to capture information.

This seems overly complicated, especially its implementation on the GCS side. The idea is not to have some fancy camera control for specialized missions, but instead to enhance the capabilities of vehicles performing survey missions.

The command that is used for surveying missions is the CAM_TRIGG_DIST, which is why this enhancement naturally falls within it. This new feature is to be used only for surveying missions, with the intent of increasing their efficiency by reducing the number of transects necessary to perform a mission and also allowing for generation of better 3D models in areas with high obstacles. This is all supported by the evidence I provided with the articles.

This PR merely hopes to use three otherwise unused parameters from a command that is unlikely to be extended in other manner, for great potential benefits that far outweigh the reasons to keep it as is.

igorsgcampos:
Don, I understand your concerns, and greatly respect and admire the contributions that you brought. I will likely not reach the amount of work you have dedicated to the community, and for this I am truly thankful. I apologize if in my fierce advocation for this feature feels disturbing in any manner, I would not be doing it unless I genuinely believed in its potential.

I fell that we are coming back to topics that were already discussed and brought up either in the dev-call, the proposal or in the comments of this PR. I hope that the PRs that I raised for a full implementation along with the figures that I mentioned are sufficient to demonstrate the viability of its use.

DonLakeFlyer:
So the thing that I had wrong was that I thought you were internally triggering multiple gimbal moves/shots for each distance trigger. When in fact I believe you are only moving the gimbal once for each distance trigger and then take a single shot. Is that correct?

igorsgcampos:
> So the thing that I had wrong was that I thought you were internally triggering multiple gimbal moves/shots for each distance trigger. When in fact I believe you are only moving the gimbal once for each distance trigger and then take a single shot. Is that correct?

Exactly, and I adjust the trigger distance according to the number of positions set in param4.

DonLakeFlyer:
Ok, my total screw up then. I see how this should work fine then. I'm still a bit concerned about bolting in specific gimbal sequence support into TRIGGER_DISTANCE. It could be further abstracted into a pose definition which would work just like my SUBROUTINE example above but the definition of a pose would only include gimbal commands. Then QGC could include survey support which defines a pose for this specific oblique setup. If I'm the only one whose concerned about making this a bit more generic with indirect reference to a pose sequence then this is fine without it.

kikislater:
> Thanks, I made a small video of the concept with the SITL. It is available at the following link:
> https://photos.app.goo.gl/PDEye5Jx66f4f9E46

Valuable implementation. If some want more validation : they made it in ebee with SODA 3D sensor which include a gimbal : 
https://www.sensefly.com/camera/sensefly-soda-3d-mapping-camera/

igorsgcampos:
> Valuable implementation. If some want more validation : they made it in ebee with SODA 3D sensor which include a gimbal :
> https://www.sensefly.com/camera/sensefly-soda-3d-mapping-camera/
> 
> Edit : They have a GUI to setup and recommend 45° tilt with 85/60% front/sidelap up to 85/10% front/sidelap for large mapping areas. If their claim are ok, definitively worth it in mapping application.
> 
> image
> 
> https://www.youtube.com/watch?v=EliGPYi8La4

This is really cool, I was unaware of that.They were pretty secretive about it when it launched, I thought it was a camera with multiple sensors or internal moving parts.

If no one else has any concerns about it I would appreciate if someone with write access could merge this PR.

hamishwillee:
>  If I'm the only one whose concerned about making this a bit more generic with indirect reference to a pose sequence then this is fine without it.

@DonLakeFlyer You're not, but I am having trouble seeing what the pose definition would look like in terms of messages - do you mean literally like your subroutine but named "TRIGGER_POSE_SEQUENCE" (or similar) with assumed camera capture after every gimbal movement?

I'm also not sure how this helps with making sure that gimbals can move as expected in time available etc.

@igorsgcampos Why I like this is that it allows that it is generic, which is what we were asked for by the dev call. 

This will need `<wip>` boilerplate added. The expected process is that you would create an implementation in a stack. At that point we merge with the wip flag. After a few months of shakedown/testing the wip tag gets removed.

DonLakeFlyer:
> do you mean literally like your subroutine but named "TRIGGER_POSE_SEQUENCE" (or similar) with assumed camera capture after every gimbal movement?

Sort of. Don't have time right now. I'll detail it out tomorrow to folks can consider.

igorsgcampos:
> @igorsgcampos Why I like this is that it allows that it is generic, which is what we were asked for by the dev call.
> 
> This will need `<wip>` boilerplate added. **The expected process is that you would create an implementation in a stack.** At that point we merge with the wip flag. After a few months of shakedown/testing the wip tag gets removed.

@hamishwillee I did an implementation in [PX4](https://github.com/PX4/Firmware/pull/15882) stack and [QGC](https://github.com/mavlink/qgroundcontrol/pull/9103), After speaking to the folks at PX4 in the dev-call last week, they are waiting the merging of this PR to approve its merging. I also just added the WIP tags to the parameters, in the same fashion as CMD 246: MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN.

Also, I do not see the need of being more generic than what is already proposed, as the objective of this PR is to improve coverage/efficiency in survey missions, which the case in which CAM_TRIGG_DIST is most widely used, so the addition of this extended functionality would not deviate the command from its intended purpose, and it would just work the same where it is already employed.

I feel that having a set of poses is also an interesting idea, but something that would be more suitable for complex missions, like a 3D model of a particular building or natural structure, as these missions generally involve more than CAM_TRIGG_DIST. Then there is the (difficult) challenge of generating these poses in the GCS software to achieve better accuracy, as it would be to cumbersome to be done by the user.

I believe that Sensefly's choice for a simple mechanism to capture images at different roll angles, exactly as in this PR, shows that it is a valuable addition and a reasonable way to go forward.

Actually, the way I felt about implementing this when considering which would be the proper way was: either have it inserted in the CAM_TRIGG_DIST command or add a new command like CAM_TRIGG_DIST_SWIVEL. I opted for the former as I believe adding such a similar command would break the principle of orthogonality and using it in the CAM_TRIGG_DIST would also have the benefit of complying with the mavlink 1 limitations.

hamishwillee:
@igorsgcampos 

1. FWIW IMO No stack relying on MAVLink 1 will implement this proposed change, so not a consideration.
2. If you have a branch of this working then cool, under current rules WIP state is not a procedural blocker.
3. Don is the "user" from a mavlink perspective. It will be up to him to create an interface that does what end users need. So lets see what he says. 
4. I'm a facilitator, not a decision maker (except in the cases where there has been clear consensus).

When we get to the end of the discussion I will ask James and Julian from the dev call to do final review and merge. Sorry this is slow, but I am not convinced of the "right thing".

igorsgcampos:
@hamishwillee Absolutely reasonable approach!

@DonLakeFlyer Thanks for taking the time and attention to evaluate this proposal and providing us an extensive discussion to have the best possible solution applied.

DonLakeFlyer:
So the concept of specifying a set of poses is like this set of commands:

* BEGIN_CAMERA_TRIGGER_POSE - This has an id in one of the params
* MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
* MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
* MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
* ...
* END_CAMERA_TRIGGER_POSE

And then commands like CAMERA_TRIGGER_DISTANCE and/or MAV_CMD_IMAGE_START_CAPTURE would reference this pose id. And then it would work just like @igorsgcampos proposes. Executing the next gimbal command in the list at each trigger and then starting over at the top when it reaches the end.

If you wanted you could get fancy and include built-in poses using negative numbers for ids. So for example the specific pose for oblique that @igorsgcampos has specified could be built into the firmware and then just be reference with say a -1 pose id.

In QGC the transect style complex items would then include direct support for this oblique triggering. The user wouldn't need to specify all the individual pose crap. QGC would deal with it for them. So the end user would not see any different of how it was implemented behind the scenes. But if it has indirect pose support they could also set up a manual transect survey with any sort of custom pose they want. Creating the pose in QGC would be fairly straightforward since the pose definitions stuff is just regular mission commands. So you just use the regular QGC mission item editing ui to do it. Nothing special there.

I don't know if it's worth it or not. That's up to real users thinking about the transects surveys they might want to do.

igorsgcampos:
> So the concept of specifying a set of poses is like this set of commands:
> 
> * BEGIN_CAMERA_TRIGGER_POSE - This has an id in one of the params
> * MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
> * MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
> * MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
> * ...
> * END_CAMERA_TRIGGER_POSE
> 
> And then commands like CAMERA_TRIGGER_DISTANCE and/or MAV_CMD_IMAGE_START_CAPTURE would reference this pose id. And then it would work just like @igorsgcampos proposes. Executing the next gimbal command in the list at each trigger and then starting over at the top when it reaches the end.
> 
> If you wanted you could get fancy and include built-in poses using negative numbers for ids. So for example the specific pose for oblique that @igorsgcampos has specified could be built into the firmware and then just be reference with say a -1 pose id.
> 
> In QGC the transect style complex items would then include direct support for this oblique triggering. The user wouldn't need to specify all the individual pose crap. QGC would deal with it for them. So the end user would not see any different of how it was implemented behind the scenes. But if it has indirect pose support they could also set up a manual transect survey with any sort of custom pose they want. Creating the pose in QGC would be fairly straightforward since the pose definitions stuff is just regular mission commands. So you just use the regular QGC mission item editing ui to do it. Nothing special there.
> 
> I don't know if it's worth it or not. That's up to real users thinking about the transects surveys they might want to do.

That is an interesting idea, but computing the adjusted footprint for such a custom arrangement of poses would be a real challenge, furthermore, having poses varying in both pitch and roll would create a ground footprint that would not be representable by a rectangular shape, which is the way the image footprint is computed today.

The way I implemented in [QGC](https://github.com/mavlink/qgroundcontrol/pull/9103), the extended parameters of CAM_TRIGG_DIST already provide the adjustment of the image footprint and therefore all other features that are in place work as expected (number of photos estimation, spacing between transects, etc...). I don't see a reasonable way to automatically adjust these settings and compute their values with a custom set of poses.

DonLakeFlyer:
> That is an interesting idea, but computing the adjusted footprint for such a custom arrangement of poses would be a real challenge, furthermore, having poses varying in both pitch and roll would create a ground footprint that would not be representable by a rectangular shape, which is the way the image footprint is computed today.

The generic support would not include automatically calculation of trigger distances in QGC. You would just set up a manual survey and specify your own trigger distance. Or just create your own mission with CAMERA_TRIGGER_DISTANCE and a pose. Automatic setup would only be for your specific oblique setup which would be a built in setup in QGC.

igorsgcampos:
I made a video of how it is currently working, including QGC UI settings. It is available at: https://youtu.be/p3yD-LFPS9E

Currently I don't have much time, I will take a look at it later and see whether I can make these changes, as it was already challenging for me to do this in QGC because I haven't worked with Qt before.

But if it is to troublesome to change the CAM_TRIGG_DIST, I would rather have the proposed behavior added as another command such as CAM_TRIGG_DIST_SWIVEL as previously mentioned. I guess there would be no objections with that.

DonLakeFlyer:
Don't worry about fighting with QGC. The survey stuff is some of the most complex part of QGC. I can add support once this gets into firmware side.

hamishwillee:
@auturgy @julianoes I take the dev call discussion to mean that this feature is desired and as there is a PX4 implementation ready to merge, this is in-line with our current interpretation of WIP. Don has also stated this can be supported in QGC and is not unreasonable (he did suggest a more generic approach that looks great, but would be IMO unlikely to be done in a timely manner as is much more work for all concerned - https://github.com/mavlink/mavlink/pull/1508#issuecomment-711347241). 

Are you OK with this to merge? 

@amilcarlucas Could you please give this a quick review for style?

message_definitions/v1.0/common.xml
```diff
@@ -1345,13 +1345,13 @@
...
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4">Empty</param>
-        <param index="5">Empty</param>
-        <param index="6">Empty</param>
+        <param index="4" label="Positions" minValue="0" increment="1" default="0">WIP: Total number of roll positions at which the camera will capture photos (images captures spread evenly across the limits defined by param5). Feature enabled if value is greater or equal to 2.</param>
+        <param index="5" label="Roll Angle" units="deg" minValue="0" default="0">WIP: Angle limits (in degrees) that the camera can be rolled to left and right of center. Enabled if param 4 is greater or equal to 2.</param>
```
amilcarlucas, line 1353: remove the "(in degrees)" it's duplicated information. And we want to motivate GCS developers to implement and support more mavlink attributes.

message_definitions/v1.0/common.xml
```diff
@@ -1345,13 +1345,13 @@
...
+        <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">WIP: Fixed pitch angle (in degrees) that the camera will hold in pseudo oblique mode if the mount supports it. Enabled if param 4 is greater or equal to 2.</param>
```
amilcarlucas, line 1354: remove the "(in degrees)" it's duplicated information. And we want to motivate GCS developers to implement and support more mavlink attributes.

message_definitions/v1.0/common.xml
amilcarlucas: So this can not be negative?

amilcarlucas (changes requested):

message_definitions/v1.0/common.xml
igorsgcampos: Yes, only positive values, as it is the amount it will roll to both sides, left and right of the center.

message_definitions/v1.0/common.xml
igorsgcampos: done

message_definitions/v1.0/common.xml
igorsgcampos: image

igorsgcampos:
For reference of parameters:

image

amilcarlucas (changes requested):
I think the nice pictures you have drawn belong in the mavlink documentation

igorsgcampos:
> I think the nice pictures you have drawn belong in the mavlink documentation

Where specifically? I am not sure how to add them there either.

amilcarlucas:
@hamishwillee can you point @igorsgcampos to the right location?
I think in the [camera microservice page](https://mavlink.io/en/services/camera.html) but you know better where to place it.

igorsgcampos:
Since I do not have write permission perhaps it is simpler for someone else to add the image there, referring to the update in CAM_TRIGG_DIST. Otherwise I can fork the repo and make the changes.

DonLakeFlyer:
> Where specifically? I am not sure how to add them there either.

As it currently stands there really isn't going to be a great place in mavlink docs that I know of. Since there isn't a place which provides additional details for an individual command. The QGC docs could certainly hold it in the docs for the new pattern(s) feature.

igorsgcampos:
> > Where specifically? I am not sure how to add them there either.
> 
> As it currently stands there really isn't going to be a great place in mavlink docs that I know of. Since there isn't a place which provides additional details for an individual command. The QGC docs could certainly hold it in the docs for the new pattern(s) feature.

I agree with Don, I could not see a fitting place in the docs. QGC is probably the best place to hold it.

DonLakeFlyer:
I would also add that I think this is going to need to be a new command. Without that there is not way for a GCS to understand whether this firmware supports the command or not. If you just "extend" an existing command and run against a firmware which doesn't support the extension it's going to report all is fine since it doesn't even look at those param values. And then you go to fly the thing and it mysteriously doesn't work at all. With a new command a GCS will be able to understand whether the vehicle supports it through the upcoming COMPONENT_INFORMATION work. And even without that the firmware will reject the command as not supported if no support.

igorsgcampos:
You are probably right Don. Hence in the beginning I was unsure about the best way to do it. I did it this way as I thought it would be possible to make a check if the vehicle has a mount and the firmware is above a certain version, enabling this option only if both were true, which made me chose to extend CAM_TRIGG_DIST.

DonLakeFlyer:
> I did it this way as I thought it would be possible to make a check if the vehicle has a mount and the firmware is above a certain version, enabling this option only if both were true

I'm putting myself out of the business of all the insanity around version checks on the GCS side. That just means the mavlink spec sucks. It's being fixed with the `COMPONENT_INFORMATION` work. I think it might be solvable there but I'm not 100% certain. The next metadata type I'm working on is `COMP_METADATA_TYPE_COMMANDS`.

So if folks are ok with leaving it on `CAM_TRIGG_DIST` with possibility of it needing to move to a new command once I work out the next step of `COMP_METADATA_TYPE_COMMANDS` then I"m fine with it as well. Otherwise I'm maybe a week away from getting there.

igorsgcampos:
I see no problem in waiting another week. Thanks Don, please let me know when you get it figured out what adjustments need to be made if necessary.

hamishwillee:
@DonLakeFlyer This is exactly what COMP_METADATA_TYPE_COMMANDS is supposed to do - identify what commands are supported and what parameters in those commands. 

I'd **prefer** a new command for this either way, 
- to allow the camera trigger the option to be extended like a subroutine at some point in the future if we wanted.
- to make a very clear separation of this use case.
If @igorsgcampos and @DonLakeFlyer are happy with that then go for it.


The place for images and additional information about MAV_CMDs is the protocol docs - either command or mission protocol would contain a section like this  [mission protocol](https://mavlink.io/en/services/mission.html#mission-command-detail). If this applies to both protocols you'd have full docs in one and just have a link in the other to those docs. 
The common.xml itself would also be updated with link to additional info - but that can happen later. 

Does that make sense?

igorsgcampos:
Which command ID should it use?

hamishwillee:
> Which command ID should it use?

Doesn't matter as long as it is unique. Perhaps 260?

igorsgcampos:
> Doesn't matter as long as it is unique. Perhaps 260?

Sounds good, I will do the changes.

igorsgcampos:
I also took some time to make some simulations to compare with regular surveys with only NADIR images. The results indicate a better accuracy and efficiency. The vehicle flew for 6 km in oblique mode with 80%/75% front/side lap at 20° roll and 9.2 km in NADIR only mode 80%/70% front/side lap. The resulting number of pictures where 338 and 213 respectively, both failed to align 3 pictures each.

Oblique:
image
[CAMPOS-80-75-3-20-cross-v3.pdf](https://github.com/mavlink/mavlink/files/5485276/CAMPOS-80-75-3-20-cross-v3.pdf)

NADIR:
image
[NADIR-80-70-cross.pdf](https://github.com/mavlink/mavlink/files/5485279/NADIR-80-70-cross.pdf)

The oblique flight is named v3 as I tried a few different markers to see which would be detected better by the metashape. Only after determining the cross pattern worked better than the circle I performed the NADIR experiment, which is why it has no version after it. I also tried manually marking the marker corners and midpoints to align photos that lacked features and presented only a single marker, but I determined that the outcome was worse than if I left these photos unaligned, thus there are points 1-9 in the former, but they are disabled. I plan to disclose the dataset later for others to experiment with it.

There is a large area with a hole in the point cloud due to capturing only homogeneous texture of the grass, which is a repetitive pattern that confuses the photogrammetry software, most of the photographs comprising only grass and markers only aligned due to the later. Despite this, the software did a pretty good job in filling this void in both the DEM and the 3D model. It is actually a good thing that the dataset is challenging, as it provides a better insight into the effective differences between both methods.

I also deliberately set the vehicle to fly at 0° to demonstrate with a larger sample size the reduction in the number of flight lines (transects).

To get a more realistic scenario I also used Sony RX0 parameters for the simulated camera and inputted these values in Metashape. The camera basically uses a very similar sensor to P4A/P4P, a well established industry standard, but Sony crops it to 15.3Mp (I just wish they gave us the full 21Mp).

kikislater:
Interesting comparison ... Better interest with less sidelap, any try ?

igorsgcampos:
@hamishwillee I moved the command to ID 260.

@DonLakeFlyer did you figure out where to go with COMP_METADATA_TYPE_COMMANDS?

@kikislater I will adjust the way sidelap is calculated in QGC to reflect the lower GSD at more inclined angles, but as I have not done it yet, I used a higher sidelap to compensate for it.

@amilcarlucas Since this command moved to a new ID with proper WIP tag, I prefer to document it later, when I deem it stable. I would appreciate your review on the current state of this command.

message_definitions/v1.0/common.xml
```diff
@@ -1587,6 +1587,17 @@
         <param index="6" label="Longitude/Y">Longitude/Y position.</param>
         <param index="7" label="Altitude/Z">Altitude/Z position.</param>
       </entry>
+      <entry value="260" name="MAV_CMD_OBLIQUE_SURVEY" hasLocation="false" isDestination="false">
+        <wip/>
+        <description>Mission command to set a Camera Auto Mount Pivoting Oblique Survey. The camera is triggered each time this distance is exceeded, then the mount moves to the next position. Params 4~6 set-up the angle limits and number of positions for oblique survey, where mount-enabled vehicles automatically roll the camera between shots to emulate an oblique camera setup (providing an increased HFOV). This command can also be used to set the shutter integration time for the camera.</description>
```
hamishwillee, line 1592: When should this be set? Ie. can it be sent in both mission and command protocol? Should it trigger only on distance, or should it trigger also on time like the message it derives from?

I guess what I'm saying is that this probably should have a note that this is expected to be used as a direct replacment for Camera_trig ...

hamishwillee:
Other than comment above looks good to me. @DonLakeFlyer Any comments/you OK with merge?

message_definitions/v1.0/common.xml
igorsgcampos: Hey Hamish! This command is supposed to be used only on mission. I will add a note citing it is a replacement for CAM_TRIGG_DIST.

hamishwillee:
Thanks @igorsgcampos. Looks great but I would like @DonLakeFlyer to comment before merging.

Dev call happy with this, but we're trying to only merge things that have an implementation or approved implementation in a flight stack - ie it doesn't go in until we know that someone is going to really use it. I have asked for confirmation here: https://github.com/PX4/PX4-Autopilot/pull/15882#issuecomment-725678773

igorsgcampos:
@hamishwillee great! I made the necessary adjustments in https://github.com/PX4/PX4-Autopilot/pull/15882 and https://github.com/mavlink/qgroundcontrol/pull/9103 to reflect the change to a new command, but buildings will break until the mavlink submodule is updated in each project.
I still plan an improvement in QGC, thus I left it as a draft for now.

igorsgcampos:
already pushed improvement on QGC side, but still awaiting the merging of this PR.

I joined the PX4-Autopilot dev-call to get their acknowledgement that this feature will be merged, pending the merging of this PR as it is a submodule for PX4.

hamishwillee:
@igorsgcampos Thanks very much. With that confirmation I have merged.

Would you be able to add a section on this trigger command with your images above to this section of docs? https://mavlink.io/en/services/mission.html#mission-command-detail

igorsgcampos:
@hamishwillee Yes, no problem in adding the image documenting the command. I will do it soon.

DonLakeFlyer:
> did you figure out where to go with COMP_METADATA_TYPE_COMMANDS?

Not yet. But since this is a new command it that doesn't matter with respect to the mavlink spec for this command or for the firmware side implementation. It only matters to implementation in QGC.

PR #1509, opened 2020-10-09
gimbal: include roll limits, roll/pitch/yaw instead of bank/tilt/pan

julianoes:
This includes a few follow up changes for gimbal:

- include roll limits
- use roll/pitch/yaw instead of bank/tilt/pan
- remove hardware roll/pitch/yaw rate limits as they are likely/realistically not going to be used
- consistent ordering of min/max

@olliw42 I would appreciate you review.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6221,8 +6221,8 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags.</field>
       <field type="uint8_t" name="gimbal_device_id">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</field>
-      <field type="float" name="pitch">Tilt/pitch angle unitless (-1..1, positive: up, negative: down, NaN to be ignored).</field>
-      <field type="float" name="yaw">Pan/yaw angle unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
+      <field type="float" name="pitch">Pitch angle unitless (-1..1, positive: up, negative: down, NaN to be ignored).</field>
+      <field type="float" name="yaw">Yaw angle unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
       <field type="float" name="pitch_rate">Pitch angular rate unitless (-1..1, positive: up, negative: down, NaN to be ignored).</field>
       <field type="float" name="yaw_rate">Yaw angular rate unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
     </message>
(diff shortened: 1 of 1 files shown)


olliw42:
I think this makes lots of sense, especially in the device information. 
(I have no opinion on the gimbal manager info, as I'm not using it)

I guess I would not call it bank though ... I generally like to use consistent notation for same things if possible, you would set the precedence here and I thus think it's relevant to spend a second on it. I've did a quick google and only found pan-tilt-roll but not pan-tilt-bank, there is though nick and bank, but I couldn't find pan-nick-bank and such, so ... pl double check what is most appropriate

I generally wonder if the xxx_rate_max fields are useful. I think originally these fields came out a bit as what one would use for joystick control, and here the max rate would be needed to scale the stick to something. But this isn't so anymore and these are the physical hardware limits. I'm not sure if there ever would be a gimbal which would set that fields to not NAN and/or some implementation which would use these fields for anything like constraining the vehicle rates accordingly ... just a thought in case one doesn't like the blow up of these messages 

btw, it would be great if it could be settled rather sooner than later since I now redid all my parts (storm32, betacopter, opentx for mavlink) and am close to release ... and I'm using gimbal_device_information LOL

julianoes:
Ok thanks for the feedback. Let me get that in but without rates as you suggest, and with better naming.

I wonder if roll, pitch, yaw would be better at that point?

julianoes:
Note to myself: pending docs...

olliw42:
> I wonder if roll, pitch, yaw would be better at that point?

I don't really mind

 I myself are using roll, pitch, yaw, but pan-tilt-roll seems to pretty common too ... and in the other messages you kind of already voted for pan/tilt ... so, you would have to also change these other messages ...

roll/pitch/yaw might be a tiny bit better as it might not attach so much to the aeronautic Euler angles, which for gimbals are not good, while pan-tilt-roll I would suspect would be clearly understood as aeronautic Euler angles by most ... but this is really such a technical distinction that I doubt anyone except of folks like me would even have such a thought LOL

> Regarding your releases: I hear you and I'll try but also changes have to be expected as long as it's marked WIP I'm afraid.

yes sure, it's WIP, and I have really no problem with that, and later changes ... it's just that it would be kind of sad if a change would happen just 1 minute after release ... :)

olliw42:
ah, yet another reasonable reasoning ... I'm not saying I would be against bank, just that it should be used if it's really what one wants to use :+1: 

btw, in the flags it's roll, pitch, yaw

olliw42:
GIMBAL_MANAGER_SET_TILTPAN -> GIMBAL_MANAGER_SET_PITCHYAW
MAV_CMD_DO_GIMBAL_MANAGER_TILTPAN -> MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
including descriptions

?

julianoes:
Yes, I was considering that too. I'll go ahead with it then, thanks!

olliw42:
fantastic
few little comments

* in MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW the two rate fields come before the two angle fields, in e.g. GIMBAL_MANAGER_SET_PITCHYAW it's the other way around. If it's not by purpose one maybe could swap it in the command
* in the information messages _max comes before _min fields, my brain usually thinks from min to max 
* in GIMBAL_MANAGER_SET_MANUAL_CONTROL descriptions it's Tilt/pitch and Pan/yaw

julianoes:
Thanks, let me fix that!

olliw42:
I tried hard, but can't find anything more = looks perfect to me :D:D:D:D LOL

olliw42:
if you're happy with it, could you consider merging? I could then take it over too.
THX :)

julianoes:
Merging since this has been reviewed and is still marked WIP.

olliw42:
fantastic!! thx a lot. :+1:

PR #1510, opened 2020-10-11
common: Add 'Inclusion Group' to fence for 'OR' fences

IamPete1:
This adds a `Inclusion Group` number to fence polygons and circular inclusion zones. The vehicle must stay within ALL zones of a single group, this is the current behavior in AP. The vehicle can be within one group OR another, this would be new behavior. 

The goal being to enable one to configure a few fences for typically flying locations and be OK to fly so long as you were inside one of them.

https://youtu.be/_IkjcgyQbRI

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1925,7 +1925,7 @@
         <description>Fence vertex for an inclusion polygon (the polygon must not be self-intersecting). The vehicle must stay within this area. Minimum of 3 vertices required.
         </description>
         <param index="1" label="Vertex Count" minValue="3" increment="1">Polygon vertex count</param>
-        <param index="2">Reserved</param>
+        <param index="2" label="Inclusion Group" minValue="0" increment="1">Vehicle must be inside ALL inclusion zones in a single group, vehicle must be inside at least one group, must be the same for all points in each polygon</param>
         <param index="3">Reserved</param>
         <param index="4">Reserved</param>
         <param index="5" label="Latitude">Latitude</param>
@@ -1947,7 +1947,7 @@
         <description>Circular fence area. The vehicle must stay inside this area.
         </description>
         <param index="1" label="Radius" units="m">Radius.</param>
-        <param index="2">Reserved</param>
+        <param index="2" label="Inclusion Group" minValue="0" increment="1">Vehicle must be inside ALL inclusion zones in a single group, vehicle must be inside at least one group</param>
         <param index="3">Reserved</param>
         <param index="4">Reserved</param>
         <param index="5" label="Latitude">Latitude</param>


hamishwillee:
@IamPete1 Thanks for the video - makes it a lot easier to understand. 

PX4 and ArduPilot behave the same way for the default which is good. Normally if you want to fly somewhere else you'd just upload a new fence or use the default fence if the autopilot supports it. 

Questions:
- At what point does the current fence get selected/become valid.
- Assume I have 2 groups and I am placed outside of both of them. Does that mean I am in violation (of both) or neither.
- Assume I have 2 groups that have an intersection:
  - I am placed in the first group and fly into the intersection. Does the first group apply, or both, or does it switch? (or to put it another way, if I fly out of the first group into the second is there a violation or am I now valid for the second group). 
  - What about if I take off in the intersection?
- What about exclusion fences?
- Is this that much easier than just saving a new set of geofence definitions and loading them when needed?

IamPete1:
> At what point does the current fence get selected/become valid.

I think its better to check all fences and then check that your inside enough to fulfill the new threshold. For avoidance in AP you just have to be inside the current fence.

>Assume I have 2 groups and I am placed outside of both of them. Does that mean I am in violation (of both) or neither.

You would have a fence breach, currently there is no difference in action from beaching one fence or another. We could add group numbers to fence rally points I guess.

>I am placed in the first group and fly into the intersection. Does the first group apply, or both, or does it switch? (or to put it another way, if I fly out of the first group into the second is there a violation or am I now valid for the second group).

Same as above because there is no difference of action it doesn't matter which fence your actually breached, just that you breached one of them. If you fly from one group to another at a overlap then your still valid because your inside one group or another.

I think considering overlaps makes this much more complex than it need be. Any number of overlapping 'AND' or 'OR' fences can be simplified to a single polygon.

>What about exclusion fences?

You still have to be outside all exclusion fences all the time.

> Is this that much easier than just saving a new set of geofence definitions and loading them when needed?

I think not having to connect a GCS and load in a fence is clearly easier than having to do that. Its also less prone to error, you only have to get it right one time.

IamPete1:
I wonder if we should have a magic group number, say -1. For 'I don't care about the other groups, you must be inside this fence all the time'. I guess that could even be 0 so that all current fences would be of this magic type. Any one would have to upload fences with groups numbers of 1 or higher to get this new behavior.

hamishwillee:
@IamPete1 

The reason I ask all these questions is that it isn't clear what the current fence is, particularly in the case where you're inside multiples. As result this response seems to be saying two mutually exclusive things are true:


I think you're saying that 


At what point does the current fence get selected/become valid.

I think its better to check all fences and then check that your inside enough to fulfill the new threshold. For avoidance in AP you just have to be inside the current fence.

Assume I have 2 groups and I am placed outside of both of them. Does that mean I am in violation (of both) or neither.

You would have a fence breach, currently there is no difference in action from beaching one fence or another. We could add group numbers to fence rally points I guess.

I am placed in the first group and fly into the intersection. Does the first group apply, or both, or does it switch? (or to put it another way, if I fly out of the first group into the second is there a violation or am I now valid for the second group).

Same as above because there is no difference of action it doesn't matter which fence your actually breached, just that you breached one of them. If you fly from one group to another at a overlap then your still valid because your inside one group or another.

I think considering overlaps makes this much more complex than it need be. Any number of overlapping 'AND' or 'OR' fences can be simplified to a single polygon.

What about exclusion fences?

You still have to be outside all exclusion fences all the time.

Is this that much easier than just saving a new set of geofence definitions and loading them when needed?

I think not having to connect a GCS and load in a fence is clearly easier than having to do that. Its also less prone to error, you only have to get it right one time.

IamPete1:
> Say I am inside one group and I fly into an intersection. I think you are saying that I remain in the first group and if I fly out of it then I am in breach. i.e. there is no re-evaluation of what group I am in. Or you could be saying that there is a re-evaluation, so if I fly into the next group the bigger polygon becomes my barrier.

The proposed logic can be stated as follows:
* You must be inside all fences of a single group to be considered inside that group
* You must be inside at least one group

Maybe I have not understood the question clearly, I think I'm saying that there is always revaluation of the current group/groups.  Maybe you could draw a diagram?

hamishwillee:
No need, that makes it clear. Thanks. 

@DonLakeFlyer @auturgy @julianoes Can you please comment on this proposed Geofence feature? In summary this change allows someone to fly in different locations, each with its own geofence definitions, without having to upload a separate plan that covers the geofence in each place. Normally this would not work because if both locations had their own geofence, you would always be violating the other location's fence.

Pete describes it well https://github.com/mavlink/mavlink/pull/1510#issue-501131525 - and provides a video.

DonLakeFlyer:
It's interesting. But to me fences are already complicated enough for an end user to setup. This just adds more stuff for them to understand for what to me seems like an edge use case. Simpler is better unless it's important to add complexity.

IamPete1:
@DonLakeFlyer If uses don't want the functionality there is no additional complexity. Also requires no GCS change to carry on working as it does now.

My argument is that this lets you set it up one time for several flying locations. Yes this is slightly more complex, but do don't have to do it before every flight in a new location as you do now.

DonLakeFlyer:
> My argument is that this lets you set it up one time for several flying locations.

In QGC just save a separate Plan which only has a fence in it for each flying location. Open it, add your mission save to a different file/upload. I don't see how that is so complicated. The steps are almost exactly the same if this new feature is added. The only difference is that you start from a single file. I don't see how that really adds much value.

> If uses don't want the functionality there is no additional complexity.

You need a UI to set it up. That UI needs to be more complicated to be able to set up a feature which is already fairly complex. If you make things so complicated to use you just drive users away to something else because they can't figure it out.

IamPete1:
> In QGC just save a separate Plan which only has a fence in it for each flying location. Open it, add your mission save to a different file/upload. I don't see how that is so complicated. The steps are almost exactly the same if this new feature is added. The only difference is that you start from a single file. I don't see how that really adds much value.

Many users do not upload a mission or even connect a GCS for every flight.

>You need a UI to set it up. That UI needs to be more complicated to be able to set up a feature which is already fairly complex. If you make things so complicated to use you just drive users away to something else because they can't figure it out.

There is no change required to the UI if you don't support the feature, fences as QGC currently writes them will continue to work exactly as the do now. I would be great to have a nice UI for it of course. But you need not do that right away, see if there are any requests for support once this gets into the wild.

Mission Planner will already support this in a basic way as it exposes all the params of each waypoint anyway. Or of-course one could edit the waypoint file manually.

DonLakeFlyer:
> There is no change required to the UI if you don't support the feature

A ground station can't decide to arbitrarily pick and choose what part of the mavlink spec for fence protocol it supports. What happens when you download from a vehicle which has a fence on it that uses this new feature? The gcs will show incorrect information in its ui. That's not quality software.

Anyway I think you've got my opinion. Simpler is better. There are other ways to accomplish this.

hamishwillee:
@auturgy @julianoes Do you have an opinion on this? 

What Pete is proposing will certainly save a bit of time for anyone who flies a few locations where they need a complex geofence and doesn't want to have to upload a new plan every time. 
However I kind of agree with Don on this one; it will be a pain to do nicely in a UI, there are other options that I don't consider onerous, and I kind of think there are  maybe better alternatives outside MAVLInk. Eg. a user agent running in GCS that checks your location and offers you relevant plans you have already created, a default simple geofence that you get for free on arming etc.

DonLakeFlyer:
I wonder why inclusion fences work this way in the first place. If this is how they work by default what is the value to ever add more than one. Is there a use case for that? To me, making them by default work to solve this described problem which is basically to make them be more like "OR" logic make more sense.

auturgy:
I think that this may allow easier integration of wide area maps / UTM / safe to fly and prohibited areas

hamishwillee:
@DonLakeFlyer @IamPete1 

We discussed this in the dev call. In overview we accept that Pete's approach complicates the protocol and is far less intuitive than making inclusion fences into "OR" as Don proposed: https://github.com/mavlink/mavlink/pull/1510#issuecomment-713296095.

I.e. changing safe fly area as shown
image


We're all pretty sure this use case was not thought about - ie it was not a decision or "better". Unfortunately making the the inclusion geofence into OR will potentially break existing geofence definitions - and this is a potentially serious safety issue. [We like that Pete's solution is transparent].

@DonLakeFlyer Any suggestions about how we might make the change to inclusion fences being OR safely? Presumably this is a change in all flight stacks and firmware?

IamPete1:
Thanks for discussing this on the call.

It was not my intention to suggest changing all fences to 'OR'. Both 'OR' and 'AND' fences have there advantages and disadvantages, I don't think its a good idea to just change all to 'OR'.

This PR is a way to let users pick for there specific situation. 

To be honest I was not really considering that fences might be touching, my use case was for fences in different locations that did not touch. No matter if 'OR' or 'AND' touching polygon fences can all be simplified to a single polygon.

DonLakeFlyer:
> I don't think its a good idea to just change all to 'OR'.

Can you explain why?

DonLakeFlyer:
> Any suggestions about how we might make the change to inclusion fences being OR safely? Presumably this is a change in all flight stacks and firmware?

Not really. I don't see how you work around the problem. The only possibility is for firmware support to define the inclusion behavior through a parameter. The default for it is current behavior so nothing changes.

auturgy:
Thinking about it, the current behaviour covers the case where a smaller fence is completely within a larger one: the conservative/safe default is that breaching the inner fence is treated as a breach. That makes sense to me. This change also makes sense to me, as it would allow regulators/service providers to create fence plans for large areas in a way that the autopilots and ground stations can handle.

IamPete1:
> Can you explain why?

Its a huge job to make the change, and it would be a breaking change. Its easier to support both, there is no breaking change and no need to simultaneously add support to all firmware and ground stations.

IamPete1:
The augment for 'AND' fences that @peterbarker put to me is that a vehicle might get fences from multiple sources (although its possible that this is currently broken in AP). For example you upload a fence from the GCS, a companion computer might upload a second fence and an on board script might give a third.

DonLakeFlyer:
I'm interested in the end user perspective not the dev perspective as to why OR is bad for inclusion. I understand the dev side is complicated.

davidbuzz:
Personally, I like the idea of a param that let's u choose to OR your fences, but i think it should be backward compatible with existing behaviour so as not to break existing setups, ... also... I can see a couple of different scenarios where one fence is totally inside the other.. such as XOR vs OR

hamishwillee:
Thanks Pete, et al. So I guess the summary is that arbitrarily changing fences to OR is not viable/AND is the safer option.

I'm not convinced of the parameter solution for allowing OR fences. That's just saying "mavlink gives up on this", and likely has unforseen safety implications.

The problem here is that it does complicate the protocol, in particular for the UI. I can see you might need to add support to show/hide the geofences that are "actively working" when flying, also display the current groups and allow them to be edited. I think that it will be much less obvious to users what is going on and what they need to set - all for a relatively niche use case that can be resolved just by uploading new plans for each location (albeit that is a pain in the butt).

How will Mission Planner do the UI?

IamPete1:
@hamishwillee This already works in MP, you can just edit the WP's directly. Admittedly it would be nice if they were automatic and displayed with different colors, but it does work.

image

This is a demo of AP doing overlapping 'OR' fences:

https://youtu.be/o4U2K7W_AFE 

In the example its using a new param to set all fences to 'OR' rather than the method proposed here, just as a way to quickly demonstrate the implementation.

hamishwillee:
@IamPete1 I do understand that it is functional, but that is a truly abysmal UI experience. 

@DonLakeFlyer I will add this to discussion in next dev call. Can you summarise your position on this - ie Impossible to create an acceptable UI, possible but likely to result in horrible edge cases, possible but just not worth the effort.

DonLakeFlyer:
Before getting into UI I think we should back up to what is the use case for this new feature specifically and not get into technical details. I will use two fields to generate the usage case, but it can extend to 2+.
* The user flies primarily or exclusively at two different flying areas
* The user want to have an inclusive geofence at each of these flying areas
* The geofences at each area differ enough or are polygonal such that the global generic support for home relative circular inclusive fences does not work
* The geofences for each area are completely discontiguous. If they intersect at all this will not work.
* The user want to set up two fences one time, upload and then leave them one the vehicle and move from one area to the other without needing to load a new fence each time they move.
* If the user ever moves away from flying at these two areas they will need to clear the fences. And then if they go back to these two areas they will need to reload the fences.

My opinion is that given the above you are already off the charts as to why you would consider adding a feature to support such an edge case. Using existing capabilities of either the global circular fence or loading the fence when you go to fly that field is a pretty simply workaround.

Can you make a user interface for this, absolutely. Can you make a really crappy user interface for it that nobody will understand, absolutely. Can you create additional confusion to already complex geofence creation for the 99.99% of users that will never use this by adding this, absolutely. 

The call on this should be made from a standpoint of whether the feature is interesting, will anyone actually use it and why the current capabilities are already not good enough.

IamPete1:
> * The geofences for each area are completely discontinuous. If they intersect at all this will not work.

This works fine for intersecting GeoFences see: https://youtu.be/o4U2K7W_AFE

> * If the user ever moves away from flying at these two areas they will need to clear the fences. And then if they go back to these two areas they will need to reload the fences.

You could just turn off polygon fences in the flight controller, no need to clear. Or you can add the new fence to your list. The same is true for the current fence support. With this change you will have to do this less often.

>Can you make a user interface for this, absolutely. Can you make a really crappy user interface for it that nobody will understand, absolutely. Can you create additional confusion to already complex geofence creation for the 99.99% of users that will never use this by adding this, absolutely.

There is no requirement to add support, the current interface will continue to work fine. If you feel that it is not going to be worth it for your users you don't have to add support. MAVLink is a communications protocol not a user interface, we should not be deciding what to add to the protocol solely based on how easy or hard it might be to integrate into one or other GCS.

IamPete1:
@hamishwillee 

> I do understand that it is functional, but that is a truly abysmal UI experience.

This is exactly the same way that users interact with all the waypoint parameters. Delay waypoints or loiter radius, its not unprecedented to have to change values there.

MAVLink is a protocol, there is no requirement that the we instantly get up and running with the perfect user interface. if this proves to be a feature that is widely used then the interface can be improved, I'm sure any pull requests in that direction would be gratefully received. The point is that it can be currently used so people can make use of feature now if they want to, we would not be adding something that users have no way to access.

IamPete1:
I think every one is too focused on the effort required to make a UI. MAVLink is not a UI, its a protocol. This is not a breaking change, all current UI's will continue to work exactly as they do now.

Would this benefit from a great UI? of course! But that is not a requirement that must be met before considering new features.

hamishwillee:
@IamPete1 I'm not worried about not having a "great UI", but I can't see anything from this other than a "poor one" (yet). It isn't the main consideration, but enough to tip me more to the "nay" camp.

@julianoes @auturgy @LorenzMeier If you could form an opinion on this for devcall it would be useful.

auturgy:
I'm thoroughly confused as to why this is even contentious.

hamishwillee:
> I'm thoroughly confused as to why this is even contentious. 

@auturgy Because it complicates the standard, and it is a lot of work for a GCS to implement (that they have no choice but to handle), for little _perceived_ benefit.

That said, I don't think it is "contentious" per-se. It is just that I will only merge PRs where I see strong technical merit and consensus. We don't have that consensus because no one "technical" on the PX4 side has properly engaged (Don counts as "neutral" for this).

Action from Dev call is for me to summarise the perceived benefit and get Lorenz to either comment or assign technical person to comment. I will do that next.

hamishwillee:
Dev call discussion this:
- Hamish: Don is opposed because it complicates the standard, and it is a lot of work for a GCS to implement (that they have no choice but to handle), for little perceived benefit. There are other ways to do this kind of thing
- James: Benefit is that in future this will make it easier for us to integrate maps of safe/prohibited areas, as discussed here: https://github.com/mavlink/mavlink/pull/1510#issuecomment-717754278. So for example if an authority published a list of safe and prohibited flight areas, this feature would allow us to directly map that to a base geofence plan, which a user could then build upon for their own geofence and still be compliant.
  - HW @auturgy not convinced on further thought:
    - The proposal is that if in any geofence group there is no violation. So if I add my geofence to the global one then as long as I am either in the official inclusions OR my own then I won't get warned. That's a problem.
    - Do authorities really care about inclusion areas? I think they only care about exclusions right?
- AO Hamish to get technical expertise from PX4 to also evaluate.

Sorry for the delay Pete!

LorenzMeier:
@DonLakeFlyer I'll pick up the discussion with you. A simplified UI (not doing everything for everyone) should be feasible.

DonLakeFlyer:
> I'll pick up the discussion with you. A simplified UI (not doing everything for everyone) should be feasible.

The UI shouldn't be the driver of the decision. That can be worked out. To me the decision should be based on adding complexity to a spec that would rarely be used and can already be done in other ways. Cluttering up an API with unneeded complexity is just as bad as cluttering up a UI with unneeded complexity.

hamishwillee:
Dev call discussed, result was that this seems like a good idea; Lorenz to discuss concerns with Don before merging.

@DonLakeFlyer The reason this seems like a good idea is that we found a use case which would mean that make this incorrect "that would rarely be used and can already be done in other ways".

Specifically we are aware of aviation authorities starting to specify allowed fly zones. Making this change would enable publishing of geofence plans that map these areas. That seems worthwhile to @LorenzMeier and @auturgy.

@LorenzMeier Please push this following your discussion with Don.

DonLakeFlyer:
> Please push this following your discussion with Don.

No need to wait for me to merge.

> Specifically we are aware of aviation authorities starting to specify allowed fly zones. Making this change would enable publishing of geofence plans that map these areas.

Interesting and this would certainly be some part of that. But keep in mind this alone won't make that feature work. That would be a global fence only. Users are still going to want to be able to also specify mission specific fences in addition to that.

IamPete1:
Thanks all, should be implemented in AP relatively soon.

message_definitions/v1.0/common.xml
```diff
@@ -1925,7 +1925,7 @@
         <description>Fence vertex for an inclusion polygon (the polygon must not be self-intersecting). The vehicle must stay within this area. Minimum of 3 vertices required.
         </description>
         <param index="1" label="Vertex Count" minValue="3" increment="1">Polygon vertex count</param>
-        <param index="2">Reserved</param>
+        <param index="2" label="Inclusion Group" minValue="0" increment="1">Vehicle must be inside ALL inclusion zones in a single group, vehicle must be inside at least one group, must be the same for all points in each polygon</param>
```
hamishwillee, line 1928: @IamPete1 @auturgy I reread this and I think this might not be what we ended up agreeing  - it's a little confusing.

My understanding is that the need for this was to allow maps of allowed areas to be published by government etc, so you would want to declare a bunch of inclusion fences in a group and allow a vehicle to fly within any of of those fences.  This is what @IamPete1 originally requested.

However if you have multiple groups, I would assume that the vehicle can be in any fence within a group, but must also be in the intersection of those groups. I.e. if you have two groups defined you'd need an overlap of those groups. I'm not sure where you would need this case.

The first sentence _Vehicle must be inside ALL inclusion zones in a single group_ sounds like the intersection. On the other hand, if you have multiple groups, I would expect that the vehicle would have to be in an interesection of those groups.

Sorry if I'm wrong - just making sure! This is what I think you mean:

```suggestion
        <param index="2" label="Inclusion Group" minValue="0" increment="1">A number identifying a group of inclusion fences. A vehicle can be within any fence (or fences) in the group but must be inside at least one. If there are multiple groups the vehicle must be inside the intersection of the fences within the two groups. The group must be specified for all points in a particular polygon and have the same value (or the fence should be rejected).</param>
```

IamPete1: Yeah, that looks good.

PR #1514, opened 2020-10-14
Mark COMMAND_INT.current as not used.

hamishwillee:
[COMMAND_INT.current](https://mavlink.io/en/messages/common.html#COMMAND_INT) doc just says it is a value of true or false (boolean). I searched PX4 and ArduPilot and I'm pretty sure it is not used. This makes sense because I think it is a hold over from whoever copied MISSION_INT (and in that context it meant "current mission item).

@auturgy @julianoes OK?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4815,7 +4815,7 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the COMMAND.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the mission item.</field>
-      <field type="uint8_t" name="current">false:0, true:1</field>
+      <field type="uint8_t" name="current">Not used.</field>
       <field type="uint8_t" name="autocontinue">autocontinue to next wp</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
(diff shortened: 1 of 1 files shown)


peterbarker:
```
    const uint8_t current = packet.current;
    const MAV_MISSION_TYPE type = (MAV_MISSION_TYPE)packet.mission_type;

    if (type == MAV_MISSION_TYPE_MISSION && (current == 2 || current == 3)) {
```

https://github.com/ardupilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L693

hamishwillee:
@peterbarker That is a MISSION_INT.continue. This is talking about COMMAND_INT.continue. 

Continue does not make sense in a command, because there is no implicit "next command".

I am 99.9% sure that COMMAND_INT was created by unthinkingly copying MISSION_INT and leaving this field in. 

The source doesn't seem to use it anywhere, and there is no "correct" way to do so, since the current docs just say it is a boolean.

hamishwillee:
@auturgy OK to merge? I'm pretty OK with self merging, but I want to make sure everyone understands I'm doing this.

julianoes (approved):
@hamishwillee thanks for trying to clarify this. In MAVSDK it's always set to 0, so not used. And on the PX4 side it doesn't seem to be read either.

hamishwillee:
People have had  a chance to comment on this. I am confident of this. Therefore will merge.

julianoes (approved):
Makes sense.

PR #1515, opened 2020-10-19
remove double quotes from strings

BazookaJoe1900:
I have mavlink generator that due to the double quite fails. because it break the string to:
"Loiter circle exit location and/or path to next waypoint ("xtrack") for forward-only moving vehicles ....."

I think its illegal to use double quotes inside string in xml, not without esc char


hamishwillee:
@BazookaJoe1900 While it is not valid to have unescaped quotes inside an _attribute_ value, it is perfectly fine to have double quotes inside a string (tag).  You should fix your parser. 

I'm closing this. It isn't that I have a problem with this particular change, its that I know in future this will keep on cropping up because the syntax is valid, and there is not way or reason for us to try prevent it with the validator.

BazookaJoe1900:
@hamishwillee Thanks for the notes.
it does seems that the tool I am using takes the attribute into the code as metadata, while pymavlink doesn't. so this is why I encounter the issue.
I will just fix that In my tool, replace  double to single quote

hamishwillee:
Thanks very much!

PR #1517, opened 2020-10-22
gimbal: change manual control input

julianoes:
I think it makes more sense to have a flag to signal angle or angular rate instead of having both fields and then setting two out of four to NAN.
I don't have a strong opinion on this but it just occurred to me.

@olliw42 what do you think?


olliw42:
I guess I don't have an opinion since I'm not really convinced that the manual control messages are of any use.

The suggestion however makes lots of sense, one wouldn't use angle and rate input at the same time, right, they are rather mutually exclusive.

I wonder if cases exist in which one would want to use different input modes for pitch and yaw, in which case the field would rather be  a bit mask.

julianoes:
> I wonder if cases exist in which one would want to use different input modes for pitch and yaw, in which case the field would rather be a bit mask.

Oh, that is a good point. I didn't think of this but it - of course - makes sense, and means I'll leave it as is! Thanks.

PR #546, opened 2016-04-25
update NAV_CONTROLLER_OUTPUT comment

dagar:
PX4 NAV_CONTROLLER_OUTPUT implemented here - https://github.com/PX4/Firmware/pull/4363

I'd personally rather have altitude/altitude_sp and airspeed/airspeed_sp instead of altitude/airspeed error. Any objection to adding the actual setpoints to this message as well?

PR #1518, opened 2020-10-23
NAV_CONTROLLER_OUTPUT: add airspeed setpoint field

sfuhrer:
It would be useful to have a means of conveying the current airspeed setpoint from the vehicle to a groundstation. I don't know what exactly this message was designed to do, but as it already has attitude setpoint fields it would make sense to me to also have an airspeed setpoint field. 

Further we should indicate the kind of airspeed in this message, whether it is meant to be indicated/calibrated or true airspeed. In my opinion it should be IAS/CAS, as that is normally the setpoint that the user can interface with and that is constrained by physical limits (stall/max airspeed). 

@RomanBapst FYI.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4678,6 +4678,7 @@
       <field type="float" name="alt_error" units="m">Current altitude error</field>
       <field type="float" name="aspd_error" units="m/s">Current airspeed error</field>
       <field type="float" name="xtrack_error" units="m">Current crosstrack error on x-y plane</field>
+      <field type="float" name="aspd_sp" units="m/s">Current airspeed setpoint</field>
     </message>
     <message id="63" name="GLOBAL_POSITION_INT_COV">
       <description>The filtered global position (e.g. fused GPS and accelerometers). The position is in GPS-frame (right-handed, Z-up). It  is designed as scaled integer message since the resolution of float is not sufficient. NOTE: This message is intended for onboard networks / companion computers and higher-bandwidth links and optimized for accuracy and completeness. Please use the GLOBAL_POSITION_INT message for a minimal subset.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4678,7 +4678,8 @@
       <field type="float" name="alt_error" units="m">Current altitude error</field>
       <field type="float" name="aspd_error" units="m/s">Current airspeed error</field>
       <field type="float" name="xtrack_error" units="m">Current crosstrack error on x-y plane</field>
-      <field type="float" name="aspd_sp" units="m/s">Current airspeed setpoint</field>
+      <extensions/>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@sfuhrer You can't just add fields to the ends of messages because the CRC check will fail - ie this breaks the message for all existing systems until they update. Given this is a MAVLink 1 message, that would also break lots of things that will never update.

To fix this we add [extensions](https://mavlink.io/en/guide/define_xml_element.html#message_extensions). There have some caveats (read link). One of those is that a receiving system that knows about this field may get 0 because zero was sent OR because the sending system had not been updated and does not sent this value. So 0 has to be something you can ignore.

Normally what we'd do in this case is say something like: `"<field type="float" name="aspd_sp" units="m/s">Current airspeed setpoint. 0 indicates airspeed information is not supplied, 0.001 can be used to indicate 0.</field>"` (the value used to represent zero being some near equivalent anyway.

Further, this suggestion was [previously rejected](https://github.com/mavlink/mavlink/pull/546) on the basis that it could not be done without breaking message. I suspect that predates Mavlink 2 being in widespread use. Also [suggested here](https://github.com/mavlink/mavlink/issues/929#issuecomment-558911121) but closed as "not sufficient interest".

@dagar any reason it can't be done now?

hamishwillee:
@sfuhrer 

[Devcall](https://github.com/mavlink/mavlink/wiki/20201111-Dev-Meeting) did not have strong opinions on this one. General questions were:
- If we add this setpoint, don't we also need altitude setpoint
- How do you plan on using it.
- Would we be better with a separate message, and if so, what would be in it?

If you have thoughts on this might be worth jotting them down here and also coming to a meeting and acting as stakeholder (?).

> Further we should indicate the kind of airspeed in this message, whether it is meant to be indicated/calibrated or true airspeed. In my opinion it should be IAS/CAS, as that is normally the setpoint that the user can interface with and that is constrained by physical limits (stall/max airspeed).

@auturgy Would ArduPilot assume the airspeed/error are w.r.t. true airspeed or IAS/CAS? If it is IAS then I will update doc on message to make that clear.

sfuhrer:
@hamishwillee thanks for coming back on this.
My goal is to get a way to rely setpoints that the drone takes in semi-autonomous modes to the ground station. In my example it is the airspeed, which we intend to let the drone adapt autonomously depending on wind speed in our case. I see that the same could be useful for altitude and heading, and maybe also for position (though that one we most likely already have somewhere). It would could then be used by the operator to compare the current setpoint with the current state.

hamishwillee:
Hi @sfuhrer ,
FYI We didn't discuss this at the devcall because you weren't there :-). Next meeting in 2 weeks.

sfuhrer:
@hamishwillee It actually has lower priority for me right now. We've found another way to handle the airspeeds, and feedback from the drone to the groundstation isn't as important anymore. And I don't wanna push for things in mavlink before I have a clear implementation in mind, so I'm closing this, to potentially reopen once we have the implementation more clearly planned.

PR #1519, opened 2020-10-26
component_information: fix parameter.schema.json array type

bkueng:
The syntax for an `array` was not correct.
Not yet reformatted, so the diff is clear.

In addition:
- adds `bitmask`
- allows '.' and '-' as part of the param names. PX4 has some of those that come from uavcan. Probably not strictly required.

Questions:
- how should the [uid](https://github.com/mavlink/mavlink/blob/master/component_information/parameter.schema.json#L13) be filled in? If it's a CRC of the file it's impossible to add it.
- how does the [scope](https://github.com/mavlink/mavlink/blob/master/component_information/parameter.schema.json#L17) work? I would expect the scope can be applied to a (sub-)set of parameters. Is that incorrect?

### DIFF
--- a/component_information/parameter.schema.json
+++ b/component_information/parameter.schema.json
@@ -10,10 +10,6 @@
             "type":         "integer",
             "minimum":      1
         },
-        "uid": {
-            "description":  "Unique id for this metadata. Same as ```COMPONENT_INFORMATION. comp_metadata_uid```.",
-            "type":         "integer"
-        },
         "scope": {
             "description":  "Scope to which this metadata applies. Firmware: Any vehicle running this same vehicles firmware type. VehicleGroup: Any vehicle running this same firmware and this vehicles group type (Fixed Wing, Multi-Rotor, VTOL, Rover). VehicleType: Any vehicle match this vehicles firmware type and specific vehicle type. Vehicle: Only applies to this specific vehicle.",
             "type":         "string",
@@ -134,6 +130,6 @@
         }
     }
     },
-    "required":             [ "version", "uid", "scope", "parameters" ],
+    "required":             [ "version", "scope", "parameters" ],
     "additionalProperties": false
 }
--- a/component_information/parameter.schema.json
+++ b/component_information/parameter.schema.json
@@ -3,7 +3,7 @@
     "$schema":      "http://json-schema.org/draft-07/schema",
     "description":  "Schema for COMP_METADATA_TYPE_PARAMETER",
     "type":         "object",
-    
+
     "properties": {
         "version": {
             "description":  "Version number for the format of this file.",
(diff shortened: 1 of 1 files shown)


DonLakeFlyer:
> * how should the [uid](https://github.com/mavlink/mavlink/blob/master/component_information/parameter.schema.json#L13) be filled in? If it's a CRC of the file it's impossible to add it.

Yup. That should not be in there.

> * how does the [scope](https://github.com/mavlink/mavlink/blob/master/component_information/parameter.schema.json#L17) work? I would expect the scope can be applied to a (sub-)set of parameters. Is that incorrect?

No. Scope tells you the scope for which this metadata applies. Using parameter meta data as example:
* For PX4 the scope would normally be "Firmware" specific since parameter metadata is the same for all vehicle types.
* For ArduPilot the scope is "VehicleGroup" specific since for example ArduCopter has different parameter set than ArduPlane.
* Scope is needed to build a fully qualified caching system for metadata on the GCS side.
* All of that said, the current spec for it is woefully missing much useful information on how it needs to work. I'll put some more work into it. For now for going forward with PX4 implementation don't worry about it. I need to put some more though into it and will do a write up.

DonLakeFlyer:
The Scope stuff crosses path with the metadata uid discussion. If the metadata uid is unique enough without collisions at a level of firmware type scope (PX4/ArduPilot/...). Then the lower level scopes are not needed and it's possible to build a two level only cache: `comp_metadata_type:firmware type`.

@bkueng It's seem like from are discussion that a two level cache would work for PX4. 

@auturgy I think the same would be true for ArduPilot. But I'm not sure if you've been following what is going on here that closely?

DonLakeFlyer:
@bkueng The rest of the changes look good. uid should be removed.

bkueng:
> No. Scope tells you the scope for which this metadata applies. Using parameter meta data as example:

That clears things, thanks.

> uid should be removed.

Removed.

LorenzMeier (approved):

LorenzMeier:
@auturgy Could you please do a quick review? Simple fix of the schema.

hamishwillee:
Looks good to me. How did you validate the schema? 
I get an error when I paste this into https://www.jsonschemavalidator.net/ "
Bad JSON escape sequence: \.." 

FWIW removing the UID means that a component can now only return its own metadata. A use case that was discussed was a system providing metadata about a non-MAVLink component - say a GPS - that was connected to it. We can decide we don't need that, or that it might be added back in a new revision

bkueng:
> How did you validate the schema?

I added a small script here: https://github.com/PX4/Firmware/pull/16039/files#diff-630e3de34d9e565ea2a0bcdecdc01fda2b3767d8db5d20f04a69700c2595cd07.

> I get an error when I paste this into https://www.jsonschemavalidator.net/ "
Bad JSON escape sequence: .."

I don't see that.

DonLakeFlyer:
> FWIW removing the UID means that a component can now only return its own metadata.

I don't quite understand why that is? 

That said though: @bkueng Does this mean uid uniqueness scoping should be down to the component level instead of just firmware level. What is a component generates its own metadata and uid. It could use a different mechanism for uid generations than the main firmware.

hamishwillee:
> I don't quite understand why that is?

@DonLakeFlyer Ignore. I was confused about what this UID was.

> Does this mean uid uniqueness scoping should be down to the component level instead of just firmware level. What is a component generates its own metadata and uid. It could use a different mechanism for uid generations than the main firmware.

Again, may misunderstand, but any component that supports the parameter protocol can potentially supply metadata. You don't need to go lower if only the autopilot supports the protocol.

bkueng:
> That said though: @bkueng Does this mean uid uniqueness scoping should be down to the component level instead of just firmware level. What is a component generates its own metadata and uid. It could use a different mechanism for uid generations than the main firmware.

Yes. For caching separation I'd use this:
- Firmware type (or vehicle group type for APM)
- metadata url (and for mavftp: if it does not specify the component ID, add the component you request from to the url to get a canonical url). The mavftp url could look like this: `mftp://[;comp=<id>]<path>`
- uid

DonLakeFlyer:
> * metadata url (and for mavftp: if it does not specify the component ID, add the component you request from to the url to get a canonical url). The mavftp url could look like this: `mftp://[;comp=<id>]<path>`

I don't quite get this. Let's talk through it on next Monday's call.

DonLakeFlyer:
FYI: My questions to Beat are not directly related to this pull. This pull is fine to merge from my standpoint.

hamishwillee:
Thanks Don. I'll merge then.

PR #1520, opened 2020-10-26
Adding storage name to storage info message.

dogmaphobic:
When displaying storage status (within the UI), this allows to identify which storage the data refers to.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -653,6 +653,36 @@
         <description>Camera does not supply storage status information. Capacity information in STORAGE_INFORMATION fields will be ignored.</description>
       </entry>
     </enum>
+    <enum name="STORAGE_TYPE">
+      <description>Flags to indicate the type of storage.</description>
+      <entry value="0" name="STORAGE_TYPE_UNKNOWN">
+        <description>Storage type is not known.</description>
+      </entry>
+      <entry value="1" name="STORAGE_TYPE_USB_STICK">
+        <description>Storage type is USB device.</description>
+      </entry>
+      <entry value="2" name="STORAGE_TYPE_SD">
+        <description>Storage type is SD card.</description>
+      </entry>
+      <entry value="3" name="STORAGE_TYPE_MICROSD">
+        <description>Storage type is microSD card.</description>
+      </entry>
+      <entry value="4" name="STORAGE_TYPE_CF">
+        <description>Storage type is CFast.</description>
+      </entry>
+      <entry value="5" name="STORAGE_TYPE_CFE">
+        <description>Storage type is CFexpress.</description>
+      </entry>
+      <entry value="6" name="STORAGE_TYPE_XQD">
+        <description>Storage type is XQD.</description>
+      </entry>
+      <entry value="7" name="STORAGE_TYPE_HD">
+        <description>Storage type is HD mass storage type.</description>
+      </entry>
+      <entry value="254" name="STORAGE_TYPE_OTHER">
+        <description>Storage type is other, not listed type.</description>
+      </entry>
+    </enum>
     <enum name="ORBIT_YAW_BEHAVIOUR">
       <description>Yaw behaviour during orbit flight.</description>
       <entry value="0" name="ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER">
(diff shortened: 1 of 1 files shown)


ArnaudThiercelin (approved):

LorenzMeier:
@auturgy Could you please have a quick look? This should be a simple change just requiring glancing over it. Thanks!

message_definitions/v1.0/common.xml
```diff
@@ -5952,6 +5982,9 @@
       <field type="float" name="available_capacity" units="MiB">Available storage capacity. If storage is not ready (STORAGE_STATUS_READY) value will be ignored.</field>
       <field type="float" name="read_speed" units="MiB/s">Read speed.</field>
       <field type="float" name="write_speed" units="MiB/s">Write speed.</field>
+      <extensions/>
+      <field type="uint8_t" name="type" enum="STORAGE_TYPE">Type of storage</field>
+      <field type="char[32]" name="name">Textual storage name to be used in UI (microSD 1, USB Stick, Internal Memory, etc.)</field>
```
hamishwillee, line 5987: Why can't the UI infer this? As it is, this then becomes untranslatable. 

If kept, is this like STATUSTEXT - "text without null termination character" or some other format?

message_definitions/v1.0/common.xml
```diff
@@ -5952,6 +5982,9 @@
       <field type="float" name="available_capacity" units="MiB">Available storage capacity. If storage is not ready (STORAGE_STATUS_READY) value will be ignored.</field>
       <field type="float" name="read_speed" units="MiB/s">Read speed.</field>
       <field type="float" name="write_speed" units="MiB/s">Write speed.</field>
+      <extensions/>
+      <field type="uint8_t" name="type" enum="STORAGE_TYPE">Type of storage</field>
```
hamishwillee, line 5986: 1. Do all systems supply this information? 
2. If a system does not know about this extension, does it matter that the recipient will assume that the drive is internal (when it might be another type)?

Essentially the "0" value should normally indicate something that is "unknown", unless you're worried about losing truncation of other fields and you're therefore prepared to report incorrect values for systems that do not understand the message.

message_definitions/v1.0/common.xml
```diff
@@ -653,6 +653,36 @@
         <description>Camera does not supply storage status information. Capacity information in STORAGE_INFORMATION fields will be ignored.</description>
       </entry>
     </enum>
+    <enum name="STORAGE_TYPE">
+      <description>Flags to indicate the type of storage.</description>
+      <entry value="0" name="STORAGE_TYPE_INTERNAL">
```
hamishwillee, line 658: Technically this is not a type. Ditto USB. Know what you mean though.

message_definitions/v1.0/common.xml
dogmaphobic: Yup, it became obvious once pointed out. Made the 0 value set to "Unknown". I also removed the "Internal" type. Some cameras have a small amount of "Internal Memory" for cases where you forgot to insert a memory card _and have to take this picture_. This is so unlikely in this context to the point of being useless.

message_definitions/v1.0/common.xml
dogmaphobic: I added a bit explaining it to be a regular, null terminated string. As to the rest, no the UI can't infer if you want to be very specific. "SD Card in Slot 1" and "SD Card in Slot 2" can't be inferred. The untranslatable problem is global and must be resolved elsewhere. Internally, if I see a string, that's what I show to the user. If no string, then I use something glanced from the type.

LorenzMeier (approved):

LorenzMeier:
This has been discussed on the dev call and is good to be merged.

message_definitions/v1.0/common.xml
hamishwillee: > he untranslatable problem is global and must be resolved elsewhere. Internally, if I see a string, that's what I show to the user. If no string, then I use something glanced from the type.

@dogmaphobic That explanation does not justify the existence of this field. Enums so they have a fixed meaning. When would it be desirable for a flight stack to have a different string? What if you're in French language and the flight stack sends an English string. What if it sends up gibberish?

Also you're forcing storage of a string on autopilot, which isn't going to use it.

IMO GCS should maintain a table of what the integers means for each value and a translation, just as is done for UI strings.

hamishwillee:
Well, actually I'd still like that cursed description string removed - https://github.com/mavlink/mavlink/pull/1520#discussion_r513854173 - "not good design"

PR #1521, opened 2020-10-28
Gimbal: CONFIGURE improvements

julianoes:
This removes the NONE flag and adds two special values to configure control. For more details check the commit messages.

@olliw42 I would appreciate your feedback.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -496,9 +496,6 @@
       <entry value="16" name="GIMBAL_MANAGER_FLAGS_YAW_LOCK">
         <description>Based on GIMBAL_DEVICE_FLAGS_YAW_LOCK</description>
       </entry>
-      <entry value="1048576" name="GIMBAL_MANAGER_FLAGS_NONE">
-        <description>This flag can be set to give up control previously set using MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW. This flag must not be combined with other flags.</description>
-      </entry>
     </enum>
     <enum name="GIMBAL_DEVICE_ERROR_FLAGS" bitmask="true">
       <description>Gimbal device (low level) error flags (bitmap, 0 means no error)</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1751,10 +1751,10 @@
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Gimbal configuration to set which sysid/compid is in primary and secondary control.</description>
-        <param index="1" label="sysid primary control">Sysid for primary control (0: no one in control).</param>
-        <param index="2" label="compid primary control">Compid for primary control (0: no one in control).</param>
-        <param index="3" label="sysid secondary control">Sysid for secondary control (0: no one in control).</param>
-        <param index="4" label="compid secondary control">Compid for secondary control (0: no one in control).</param>
+        <param index="1" label="sysid primary control">Sysid for primary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
+        <param index="2" label="compid primary control">Compid for primary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
+        <param index="3" label="sysid secondary control">Sysid for secondary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
+        <param index="4" label="compid secondary control">Compid for secondary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
         <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">


olliw42:
as I'm not using this gimbal manager but my own version I do not have a strong opinion nor competence in the matter. my 2 cents:

NONE flag:
If I understand it correctly this indeed should be a left over from earlier attempts and release is done now differently. So, this makes totally sense to me.

additional values to control control: 
I don't fully comprehend the consequences but they might make sense and might be needed to keep things going, but my clear feeling is that it doesn't look anything like simple anymore now and that rather something is just fundamentally wrong. 

-2: ??? I have no idea how that could happen and how any communication could work if sysid is not known. I mean, if sent this message is coming with a sysid in the header, so whoever is sending this with -2 can't have put -2 in the header's sysid field. Since it's a command it might not actually be send, and I think that's what you are concerned with. However, this ONLY can be the case if the mission manager and the gimbal manager are on the same device, e.g. the autopilot or the companion. It is a mistery however to me how an autopilot could not know it's sysid. A companion might auto detect the sysid and hence might not know it initially, however it then also couldn't do any mission, i.e., it would have to wait in any case until it knows the sysid...    

-3: wouldn't this be exactly the same as setting to 0?

-1: wasn't it said that it should be cooperative, so shouldn't there be no situation where two clients would do conflicting things at the same time? If so, every client at every time would know which clients are in control and how it wants to change it, i.e., it always should be able to set both field and not want to set only one field.  

as said,my feeling is that the mere fact that such values are considered necessary rather point to the fact that something is wrong in the concept, but at the same time I may not understand it well enough to make a comment.

:)

julianoes:
> as said,my feeling is that the mere fact that such values are considered necessary rather point to the fact that something is wrong in the concept,

You're probably not wrong :). Let me give you more context on the implementation I'm working on:

In the ground station we need a way to assemble the mission for a survey, it goes something like this:
1. Takeoff,
2. Go to survey begin, 
3. Configure autopilot to be in control of gimbal
4. Pitch gimbal down
5. Start taking pictures
6. Start flying survey waypoints
7. Stop taking pictures
8. Pitch gimbal back up
9. Release control

Such a mission can be configured offline, so e.g. at home before you fly. This mans that the sysid/compid of the autopilot is not yet known. Therefore, using `-2` in 3. allows to set the sysid/compid in control which is executing the mission.

Then `-3` is useful in 9. because it will release if the autopilot is still in control. If however, the ground station has interfered during the survey mission and is now in control (for whatever reason) the mission end is not conflicting with that. If we were using `0` we would remove the control of the ground station which then needs to acquire it again.

And lastly, `-1` is handy when a command wants to change e.g. primary control but not secondary control, so `-1` for secondary control just leaves it as is. Again, this is not required for e.g. a ground station reconfiguring it because a ground station knows the current status, however, a mission item does not.

olliw42:
many thx, this was very useful

-2:
and what happens if in the mission one wants a different component to get control, e.g 
3. Configure camera to be in control of gimbal?

-3:
I would argue that a client which is not in control can't anymore release itself. That is, in 9. the gimbal manager should realize that this message is coming from the autopilot and that the GCS is in control and thus should ignore the autopilot's release. 
I'm not sure if that rule could in your particular cooperative approach produces issues in other edge cases, but such a rule makes sense to me.

And again: What happens if in the mission one wants a different component to get control, e.g. in 3. configure camera to be in control of gimbal?

What happens if in your mission you want to allow the GCS to interfere at a certain point, and not just then the user wants to do it?
What happens if in your mission you want to set a GCS to secondary control, but don't yet know the sys/compid of that GCS (because you do the mission in some other tool, or the mission is no used with a differnt GCS)?

Your -2 and -3 are somewhat stuck to an autopilot. The considerations however show - as much as I can see - that what you actually want is a way to explicitly say "release autopilot" or "release camera", and so on. Which isn't easily modeled through negative numbers....

-1:
I understand the convenience but not fully the need.

BTW:
In "my" gimbal manager I kept the idea of earlier attempts to use "pseudo ids" or "aliases" or however one wants to call that in order to identify the clients (although I made it more systematic now). That is, I do have an enum  (in shorthand)

    <enum name="MAV_STORM32_GIMBAL_MANAGER_CLIENT">
      <entry value="0" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_NONE">
      <entry value="1" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_ONBOARD">
      <entry value="2" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_AUTOPILOT">
      <entry value="3" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_GCS">
      <entry value="4" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_CAMERA">
      <entry value="5" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_GCS2">
      <entry value="6" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_CUSTOM1">
      <entry value="7" name="MAV_STORM32_GIMBAL_MANAGER_CLIENT_CUSTOM2">

This avoids having to use two uints, but also avoids having to know sys/compids in advance. Through gimbal manager flags specific clients can be set or released, which would allow to exactly do what one needs in 9.

julianoes:
> -2:
> and what happens if in the mission one wants a different component to get control, e.g
> 3. Configure camera to be in control of gimbal?

I would say for a special case like that you have to have that information beforehand. Alternatively, we could say that sysid can be -2 and then compid can be for the camera, so it will get only the sysid at runtime.

> I would argue that a client which is not in control can't anymore release itself.

Ok, that's also a possibility but would have to be clearly documented like that. This means that no one can set 0 anymore without first being in control though. So if the participant that is in control suddenly disconnected, then no one else can reset everything to 0.

> What happens if in your mission you want to allow the GCS to interfere at a certain point, and not just then the user wants to do it?

If you really want to allow the GCS to interfere you would remove your own control with -3 (or just 0 if you want to do it no matter what). Then the GCS knows it can take control.

If a user on purpose wants to interfere, then nothing is stopping the GCS to send CONFIGURE and set itself back in control.

>What happens if in your mission you want to set a GCS to secondary control, but don't yet know the sys/compid of that GCS (because you do the mission in some other tool, or the mission is no used with a differnt GCS)?

I don't know but I would just leave the control open (0) and have the GCS claim it.

> This avoids having to use two uints, but also avoids having to know sys/compids in advance. Through gimbal manager flags specific clients can be set or released, which would allow to exactly do what one needs in 9.

That is indeed nice. On the flipside the participants need to now know this mapping and adhere to it but it's a viable approach.

olliw42:
> I would say for a special case like that you have to have that information beforehand. Alternatively, we could say that sysid can be -2 and then compid can be for the camera, so it will get only the sysid at runtime.

is this a special case? I believe to remember that you were moving some scenarios explicitly to the camera, such as some tracking things.

The alternative implies that sysid must be that of the autopilot. What happens if in the mission you want a secondary GCS to get control for dual-operator scenarios?

It seems to me that the general problem rather is that the mission does become not only a mission but more like a script. That is, that things like who is in control of the gimbal is also put into the mission, and not only basic controls like setting pitch. This to me seems to be a quite general problem with the missions, which is related to the unfortunate dual role which the MAV_CMDs do play. I wonder if missions shouldn't just die out and be replaced by some more generic scripting.

Anyway, I just can repeat that to me the -X thing doesn't feel right. I predict that over time one will end up with many -X, not just these three... 

> This means that no one can set 0 anymore without first being in control though. 

yes

> So if the participant that is in control suddenly disconnected, then no one else can reset everything to 0.

not sure, depends on the exact rules of "cooperative". I think somewhere earlier you were saying that "cooperative" for you means that everyone needs to release before someone else can claim control (I call this type of cooperativeness "exclusive"). This would have then been a general problem, and would require some mechanism in the gimbalmanager to handle sudden unexpected events (e.g. a timeout).

In your mission you now change that to a mode there everyone can also terminate anyone else. This isn't exactly "co-operative", right, but rather rude.

Cooperativeness always means having to spend more work in managing things, and I think this is what you hit here. And in order to get along and not to have to spend that effort one introduces many special rules ...
(that's why I like priority-based schemes)

> If a user on purpose wants to interfere, then nothing is stopping the GCS to send CONFIGURE and set itself back in control.

That's demonstrating explicitly what I was just saying, right, that it's now rude ;)

> If you really want to allow the GCS to interfere you would remove your own control with -3 (or just 0 if you want to do it no matter what). Then the GCS knows it can take control.

what I meant was that in the mission you want to give the GCS the control, not just to allow it to get it, i.e., something like

:
3. Configure autopilot to be in control of gimbal
:
xx. Configure GCS to be in control of gimbal
:
yy. Release control

> I don't know but I would just leave the control open (0) and have the GCS claim it.

what I meant was that in the mission you want to give the secondary GCS the control, not just to allow it to get it.

>  On the flipside the participants need to now know this mapping and adhere to it but it's a viable approach.

Each participant knows what type it is. At least every participant should know ;).

I think what I might have missed to say is that every message of course needs to be extended by a client field which tells which client it is. So no mapping between sysid/compid pairs to client types.  We determined earlier that this would be too phenomenological and thus error prone and ruled it out, Note that it was exactly what earlier had lead to the "pseudo ids", and also the conclusion that some field is needed.

Nevertheless, it is true that also this approach does have flipsides, as it seems there is no 100.00% perfect solution. I nevertheless decided to go with it since to me it seems to have acceptable flipsides.

Lastly, I should clearly say that your suggestions of course DO work as you want them to work, and in this sense they are perfectly ok. I'm not saying they wouldn't. All I'm saying is that to me it looks like it's not the right path to go.

julianoes:
> is this a special case? I believe to remember that you were moving some scenarios explicitly to the camera, such as some tracking things.

Yes, but I would not start it automatically from a mission but from the GCS (during a mission or manual flight). So the GCS would tell the camera to initiate tracking and the camera would tell the gimbal that it now wants to be in control to do the tracking.

> The alternative implies that sysid must be that of the autopilot. What happens if in the mission you want a secondary GCS to get control for dual-operator scenarios?

Again, I wouldn't try to configure all of that offline as part of a mission.

> I wonder if missions shouldn't just die out and be replaced by some more generic scripting.

Yes, I very much have that feeling too.

> not sure, depends on the exact rules of "cooperative". I think somewhere earlier you were saying that "cooperative" for you means that everyone needs to release before someone else can claim control

Yes it does. At this point I would describe cooperative as: you can intefere when someone else is in control but you should only do that for a reason (e.g. the pilot presses a popup confirming "take over manual gimbal control").

> what I meant was that in the mission you want to give the secondary GCS the control, not just to allow it to get it.

I know what you mean but that I don't think that works because when you create the mission you probably know very little. You don't know if there is a pilot or a videographer/gimbal operator that should be in control, and you don't know the ids of these clients, etc. All you know is that the survey is done and you don't care anymore about what the gimbal does.

> Nevertheless, it is true that also this approach does have flipsides, as it seems there is no 100.00% perfect solution. I nevertheless decided to go with it since to me it seems to have acceptable flipsides.

> Lastly, I should clearly say that your suggestions of course DO work as you want them to work, and in this sense they are perfectly ok. I'm not saying they wouldn't. All I'm saying is that to me it looks like it's not the right path to go.

Thanks, I appreciate you thinking through this. I guess I somewhat prefer the other flipsides :smile:. There certainly is no simple solution.

olliw42:
several of your comments suggest to me that what you say for other components like camera and GCSes also could be argued to apply to the autopilot.

That is, if a mission is started either

- the autopilot automatically claims gimbal control. This would avoid the need for the explicit configure and explicit release, which both cause the issue. One could refine and do this only if there is a gimbal cmd in the mission. One could further refine and do this only if not someone else is already in control. Since anyone can override one doesn't really have to explicitly release. One could further refine an automatically add a release at the end of a mission if (and only if) the autopilot is still in control.

- the whoever it is which starts the mission would also in addition tell that the autopilot is in gimbal control. This also would avoid the need for the explicit configure and explicit release, which both cause the issue. E.g. if the GCS initiates the mission it also would configure the autopilot to control. This was in fact my very first thought of how things should be, but I assumed that you wouldn't find that ok but wanted to have that explicitly as a cmd in the mission. However, for the camera and secondary GCS this is exactly what you suggest how it should work, so I would think the very same should be acceptable also for the autopilot.

julianoes:
> the autopilot automatically claims gimbal control.

That's actually what I implemented first but I ran into another issue: I needed to keep track during the mission if the autopilot is still in control because whenever a mission is finished, or we switch out of the mission mode, we need to release control, and set it to 0. However, if the ground station would have taken over the gimbal in the meantime (during the mission) we should not steal the control from it again but leave it untouched. This means that at the place where we would release control we need that information. This got a bit cumbersome implementation wise. Basically, this information needed to be wired all the way to the module which takes care of this. At that point is when I thought "wouldn't it be easier if ...".

These were the advantages of doing the magic values for this case (for me):
1. I avoid the implementation overhead.
2. There can't be a race between a GCS taking control and PX4 releasing control. (What could happen is that a GCS acquires control right the second before PX4 releases control. So PX4 does not know a GCS has the control and sets it to 0. The workaround would be that you cannot release control if you don't have it as you suggested.)
3. I needed to have magic values anyway in order not to touch e.g. secondary control, or run into races there.
4. The automatic configure "injection" on the autopilot side is not straight forward and needs to be documented and explained, also not trivial (I know you're going to object).

> the whoever it is which starts the mission would also in addition tell that the autopilot is in gimbal control. This also would avoid the need for the explicit configure and explicit release, which both cause the issue.

This is an interesting thought. So basically the GCS would say the following to start the mission: arm, takeoff, get gimbal control, start mission. This would work when a mission is always started by the GCS. However, it's of course possible to preload a mission and then start it later using the mission mode switch on RC. (It's so hard to do anything when everything is supposed to be generic.)

> this is the scenario you are looking at, and you are making it to work for that scenario.

This is the way I see it: I'm trying to make it work for an actual use case I have which has also been a use case for the past 3 years. And I'm not excluding any other use cases, I'm just saying that more involved use cases need to have more configuration or things known in advance to make it work, e.g. you have to know the camera sysid/compid in order to tell it tacking at location X.

olliw42:
>  I needed to keep track during the mission if the autopilot is still in control 

my first impulse is to argue that this is what the gimbal_manager_status message is for ...

> This means that at the place where we would release control we need that information. 

my first impulse is to argue that this is what the gimbal_manager_status message is for, i.e., the autopilot would track it and thus would know at the time it needs to know ...

> This got a bit cumbersome implementation wise. Basically, this information needed to be wired all the way to the module which takes care of this. At that point is when I thought "wouldn't it be easier if ...".

I totally understand this, and I totally am not surprised about it, its exactly what I meant with that cooperative approaches need lots of additional effort ...

with these efforts also the old gimbal protocol can be made to work ... 

> However, it's of course possible to preload a mission and then start it later using the mission mode switch on RC. 

yes, that's the problem. But that problem would exactly also exist for camera and secondary control ... what's why your answer to my  "What happens if..." isn't satisfying

> This is the way I see it: I'm trying to make it work for an actual use case I have which has also been a use case for the past 3 years. 

I fully understand this. And that's great, and what you proposse will do it.

However, the result is that "your" gimbal manger only works for you.

> And I'm not excluding any other use cases

I don't agree with that. You e.g. excluded RC targeting. You now exclude setting the camera into control during mission, even though that's your proposed need for doing things like tracking. 

> e.g. you have to know the camera sysid/compid in order to tell it tacking at location X.

no you don't. You only have to in your concept.

as said, my impression is that your gimbal manger only works for you. It shouldn't be in common or standard but in px4.xml.

that's perfectly fine if that is what you guys want and are happy with. But I think that this part of the gimbal protocol v2 should not be presented in the general docs. (unless you want to stir feelings that PX4 = mavlink),

> Another thing that came up during the implementation: the autopilot did not know when to release. So basically, the user can set the mission items in any way they want but what is the queue for the autopilot to release control? Is it if the flag NONE is set, is it if the flag Neutral is set? What if the user just sends a new pitch 0 command. The autopilot basically has to make some assumptions if it should release control or not.

I'm not sure I understand this. To me it looks that you want to write a mission which works in all sorts of circumstances without having to rewrite the mission. Not sure if that can be successful.

julianoes:
> I totally understand this, and I totally am not surprised about it, its exactly what I meant with that cooperative approaches need lots of additional effort ...
> with these efforts also the old gimbal protocol can be made to work ...

Correct. And I was against gimbal v2 in the first place but everyone on the mavlink dev call found "no it's a mess, we should come up with v2, Julian please make a proposal". (I should have never ever accepted to do that :smile:)

> You e.g. excluded RC targeting.

No I don't. I just require RC to be proxied by whoever receives RC.

> as said, my impression is that your gimbal manger only works for you.

Possible. Maybe because I'm the only one implementing it apart from you. So all I find and try to make work is what I come across.

> It shouldn't be in common or standard but in px4.xml.

That's fair, and that's why it's still flagged WIP (because we don't have px4.xml yet).

> But I think that this part of the gimbal protocol v2 should not be presented in the general docs.

Well, sorry for writing docs :confused:.

Ok, to flip this around. You are saying I don't cover the "camera tracking initiated by an offline produced mission" case. I would like to support it but I don't know how. If you do, please go ahead and make a suggestion.
Sidenote: I really don't think you can initiate tracking as part of a mission. If you do tracking you select the target on the image and directly send that tracking to the camera which will then track. You don't encode tracking into a lit of mission items, or do you?

olliw42:
> Correct. And I was against gimbal v2 in the first place but everyone on the mavlink dev call found "no it's a mess, we should come up with v2, Julian please make a proposal". (I should have never ever accepted to do that smile)

:) :) :)  

well, that's how life sometimes goes, one suddenly is trapped into something LOL

v1 had to be replaced however, so no way out here

and whatever pain points are still there v2 is conceptually much better. For instance, it's now not anymore required to work out who is in control from just who is sending (this old time based approach). Now there is something to explicitly set it by a dedicated message, which really removes many issues with v1. So clear progress. It doesn't (yet) remove all issues however. 

> Well, sorry for writing docs

oh. No, docs are certainly fantastic, and it's fantastic you're spending time on it. What I was meaning was that it shouldn't show up as standard.

> Ok, to flip this around. You are saying I don't cover the "camera tracking initiated by an offline produced mission" case. I would like to support it but I don't know how. If you do, please go ahead and make a suggestion.

not exactly, I'm saying that there likely will be cases coming up which are not covered by the current -X approach and will lead to more -X flags, which will make it more and more unpleasant.

as regards going ahead, I'm going with my "own" gimbal manger, which is based on suggestions I made earlier. So, I do not have a strong opinion here nor strong incentive to make it better. Sorry for that. 

As said, your extension certainly will do what you want to achieve, no doubt about it.

julianoes:
> oh. No, docs are certainly fantastic, and it's fantastic you're spending time on it. What I was meaning was that it shouldn't show up as standard.

Right, I get it.

> not exactly, I'm saying that there likely will be cases coming up which are not covered by the current -X approach and will lead to more -X flags, which will make it more and more unpleasant.

That's likely yes, not sure what to do about it.

> As said, your extension certainly will do what you want to achieve, no doubt about it.

Alright thanks.

julianoes:
Alright, I'm going to self-merge this since no one seemed to object, at least not strongly, and the messages are still marked WIP at the moment, so nothing should be broken. Thanks @olliw42 for thinking and talking this one through with me.

olliw42:
my pleasure :)

message_definitions/v1.0/common.xml
```diff
@@ -1754,10 +1751,10 @@
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Gimbal configuration to set which sysid/compid is in primary and secondary control.</description>
-        <param index="1" label="sysid primary control">Sysid for primary control (0: no one in control).</param>
-        <param index="2" label="compid primary control">Compid for primary control (0: no one in control).</param>
-        <param index="3" label="sysid secondary control">Sysid for secondary control (0: no one in control).</param>
-        <param index="4" label="compid secondary control">Compid for secondary control (0: no one in control).</param>
+        <param index="1" label="sysid primary control">Sysid for primary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
```
hamishwillee, line 1754: Do you mean -1, -2, -3 in all of these or 1, 2. 3?

hamishwillee: Ignore. Obvious.

hamishwillee:
Sorry, just got to this. Looks OK to me :-)

PR #1524, opened 2020-11-03
Remove WIP for messages implemented in MAVSDK

julianoes:
I went through the MAVSDK source and removed all WIP notes/flags in common for messages that are already implemented and released.

Regarding the gimbal v2 messages I would personally leave them WIP a few weeks longer, for the rest it should be fairly clear.

This is in response to the request by @LorenzMeier in the [last dev call](https://github.com/mavlink/mavlink/wiki/20201028-Dev-Meeting#notes). The same would have to be done for PX4.


hamishwillee:
Thanks @julianoes 
Makes sense.

PR #1525, opened 2020-11-09
Move MAV_CMD_DO_AUTOTUNE_ENABLE to common

jkflying:
We'll be bringing autotune capabilities to PX4 as well, so to avoid fragmentation I'd like to bring this message to common.

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -27,7 +27,7 @@
     </enum>
     <!-- ardupilot specific MAV_CMD_* commands -->
     <enum name="MAV_CMD">
-      <!-- 214 used by common.xml -->
+      <!-- 200 to 214 used by common.xml -->
       <entry value="215" name="MAV_CMD_DO_SET_RESUME_REPEAT_DIST">
         <description>Set the distance to be repeated on mission resume</description>
         <param index="1" label="Distance" units="m">Distance.</param>
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -27,17 +27,7 @@
     </enum>
     <!-- ardupilot specific MAV_CMD_* commands -->
     <enum name="MAV_CMD">
-      <!-- 211 used by common.xml -->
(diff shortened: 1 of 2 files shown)


hamishwillee:
Excellent. 

@auturgy Any objection before I merge (it is unchanged from ArduPilot version).
FMI is Autotuning a mode? i.e. how does GCS know when the task is finished?

jkflying:
Still working out the exact message flow we want, but it runs in a few seconds so the UX considerations are pretty different from previous implementations.

message_definitions/v1.0/ardupilotmega.xml
```diff
@@ -27,17 +27,7 @@
...
-        <description>Enable/disable autotune.</description>
-        <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">Enable (1: enable, 0:disable).</param>
-        <param index="2">Empty.</param>
-        <param index="3">Empty.</param>
-        <param index="4">Empty.</param>
-        <param index="5">Empty.</param>
-        <param index="6">Empty.</param>
-        <param index="7">Empty.</param>
-      </entry>
+      <!-- 214 used by common.xml -->
```
auturgy, line 30: 212?

auturgy:
Looks like a little typo, but beyond that I'm all for standardising

message_definitions/v1.0/ardupilotmega.xml
jkflying: Maybe I missed the convention here, the intent is more like "0 - 214 used by common.xml". Want me to change it?

hamishwillee: @jkflying the convention is/has been that when an id moves out of a dialect we list it here. You moved 212 out but added note for 214.
But in this case your suggestion is much better, and I have changed it.

hamishwillee:
As discussed in the dev call, I will merge. 

Given there is work going on here it would be good to know how ArduPilot knows when the tuning is finished so I can document this and avoid spec ambiguity. It could be that you monitor for a mode change, or maybe this is a blocking command, so you use long running command with progress notifications.  @meee1 @auturgy If you could confirm would be great.

PR #1527, opened 2020-11-24
common: add new_mission extension to MISSION_COUNT message

IamPete1:
This adds a extension to allow the flight controller to notify that the mission has been changed, this allows the GSC to re-load. The mission could have been changed by a second GCS, companion computer or onboard scripting.

The use case here is to deal with fences automatically set and loaded by scripting https://github.com/ArduPilot/ardupilot/pull/15598

This is the demo that just streams the fence back, this is not a feasible solution IRL due to flooding the link and/or packet loss, but it shows the functionality that I would like to duplicate with this extension. 
https://youtu.be/WakSwFTCeV8

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4618,6 +4618,7 @@
       <field type="uint16_t" name="count">Number of mission items in the sequence</field>
       <extensions/>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
+      <field type="uint8_t" name="new_mission">Boolean to indicate a new mission has be loaded, when true the MISSION_COUNT message is emitted without having received MISSION_REQUEST_LIST, this notifies the GCS the mission has changed and it should re-load</field>
     </message>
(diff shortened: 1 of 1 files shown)


hamishwillee:
I'd rather have https://github.com/mavlink/mavlink/pull/1172

IamPete1:
@hamishwillee I was not aware of that PR, as it currently stands that is for missions only so would not work for my fence use case. 

No reason we cant have both, I guess that PR is not popular since it has sat for over a year?

hamishwillee:
> No reason we cant have both

I think we should try at least for just one solution, and I think the other one seems more flexible to me.

> I guess that PR is not popular since it has sat for over a year?

It was essentially approved but waiting on microservices going in. Let's push it forward.

meee1:
how does this with packet loss?

IamPete1:
@meee1 Its better than my previous method of just spamming the points to GCS, At least it would now know how many points it should ask for and can re-request a point if its missing.

hamishwillee:
As discussed in the dev call we're closing this and iterating the other item - let's make sure that does what you need Pete!

PR #1529, opened 2020-11-27
smart_battery_info extend serial number

mrivi:
Some smart batteries have serial numbers with a alpha numeric serial number. I would propose to extend the serial number from an integer to a array of ASCII characters

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6569,7 +6569,7 @@
       <field type="int32_t" name="capacity_full_specification" units="mAh">Capacity when full according to manufacturer, -1: field not provided.</field>
       <field type="int32_t" name="capacity_full" units="mAh">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
       <field type="uint16_t" name="cycle_count">Charge/discharge cycle count. -1: field not provided.</field>
-      <field type="int32_t" name="serial_number">Serial number. -1: field not provided.</field>
+      <field type="char[16]" name="serial_number">Serial number in ASCII characters, 0 terminated. All 0: field not provided.</field>
       <field type="char[50]" name="device_name">Static device name. Encode as manufacturer and product names separated using an underscore.</field>
       <field type="uint16_t" name="weight" units="g">Battery weight. 0: field not provided.</field>
       <field type="uint16_t" name="discharge_minimum_voltage" units="mV">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>


message_definitions/v1.0/common.xml
```diff
@@ -6569,7 +6569,7 @@
       <field type="int32_t" name="capacity_full_specification" units="mAh">Capacity when full according to manufacturer, -1: field not provided.</field>
       <field type="int32_t" name="capacity_full" units="mAh">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
       <field type="uint16_t" name="cycle_count">Charge/discharge cycle count. -1: field not provided.</field>
-      <field type="int32_t" name="serial_number">Serial number. -1: field not provided.</field>
+      <field type="uint8_t[16]" name="serial_number">Serial number in ASCII characters. All 0: field not provided.</field>
```
julianoes, line 6572: You need to specify if 0 terminated or not. And I would probably use `char[16` instead of `uint_t[16]` to signal an ASCII string.

mrivi: ok, I made the changes

julianoes: What about 0 termination?

mrivi: @julianoes for my application 0 termination would be fine since the batteries s/n has 14 characters. I have added it to the description

julianoes (approved):
Looks good to me.

hamishwillee:
@julianoes This is technically much better. 
- Should we rename to "serial code"?
- Can we assume that serial numbers will only contain the ASCII set or do we need to think about Unicode here, while we're making breaking changes.
-  Changing the type of a field is a breaking compatibility change. Semantically, so is changing the value to 0000 for "not supported".

I am pretty sure that no one does "yet". Maybe @auturgy @DonLakeFlyer @AlexKlimaj can comment?

julianoes:
> Should we rename to "serial code"?

While it is technically a serial code I usually know what is meant with a serial number.

> Can we assume that serial numbers will only contain the ASCII set or do we need to think about Unicode here, while we're making breaking changes.

Yes, I'd say they are at most ASCI, often just hex or base64, so I don't think unicode is required. Plus there is no unicode support anywhere else in mavlink. (Now I'm thinking about param names made up of emojis :grin:.)

> Changing the type of a field is a breaking compatibility change. Semantically, so is changing the value to 0000 for "not supported".

Right.

hamishwillee:
Works for me. So as long as that set of people above tell me they are OK with the break, so am I :-)

DonLakeFlyer:
QGC doesn't use this message yet so it doesn't care.

hamishwillee:
Thanks Don. @auturgy @AlexKlimaj Any comments. I plan to merge this in a couple of days if there is no response - because I am "pretty sure" this is not used (in particular because still marked as WIP).

@mrivi Presumably you're changing this because of a real use case. Does this mean there is/will now be a real implementation and we should also remove the wip?

AlexKlimaj:
I think ASCII is okay. And I would go with "Serial Number", "Serial ID", or just "Serial".

auturgy:
It’s not implemented anywhere afaik and is WIP, so no blocking from me. The implementation needs to be referenced to remove WIP though.

Regards,

James

hamishwillee:
CI is taking FOREVER! Will merge when I'm allowed.

PR #1530, opened 2020-11-27
add message to turn on/off indicator lights

mrivi:
Add message to turn on and off the vehicle lights:
- visible light
- IR lights
- position lights (green/red for orientation information)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -785,6 +785,18 @@
         <description>Cancel the current parameter transaction.</description>
       </entry>
     </enum>
+    <enum name="MAV_LIGHT_SOURCE">
+      <description>Possible light sources to indicate the presence of a vehicle.</description>
+      <entry value="0" name="VISIBLE_LIGHT">
+        <description>Visible light source.</description>
+      </entry>
+      <entry value="1" name="INFRARED">
+        <description>Infrared source.</description>
+      </entry>
+      <entry value="2" name="POSITION">
+        <description>Position lights (red/green), also indicates heading.</description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
@@ -1623,6 +1635,14 @@
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Turns illuminators ON/OFF. An illuminator is a light source that is used for lighting up dark areas external to the sytstem: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
+      </entry>
+      <entry value="406" name="MAV_CMD_LIGHT_INDICATOR_ON_OFF" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Turns indicator light source ON/OFF. This light source is meant to show the presence of a vehicle.</description>
+        <field type="uint8_t" name="type" enum="MAV_LIGHT_SOURCE">Type of light source</field>
+        <param index="1" label="BlinkDuration" units="seconds">Portion of the period where the light is ON. </param>
+        <param index="2" label="Period" units="seconds">Period of the blink. </param>
       </entry>
       <entry value="410" name="MAV_CMD_GET_HOME_POSITION" hasLocation="false" isDestination="false">
         <description>Request the home position from the vehicle.</description>


message_definitions/v1.0/common.xml
```diff
@@ -1624,6 +1624,14 @@
         <description>Turns illuminators ON/OFF. An illuminator is a light source that is used for lighting up dark areas external to the sytstem: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
       </entry>
+      <entry value="406" name="MAV_CMD_LIGHT_INDICATOR_ON_OFF" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Turns light source ON/OFF.</description>
+        <param index="1" label="EnableEO" minValue="0" maxValue="1" increment="1">0: EO Indicator OFF, 1: EO Indicator ON</param>
```
dagar, line 1631: You can expand the labels if you want, they're just strings, not fields we can use programmatically.

```suggestion
        <param index="1" label="Electro-optical (EO)" minValue="0" maxValue="1" increment="1">0: EO Indicator OFF, 1: EO Indicator ON</param>
```

auturgy: I'm not sure what the label is indicating.  Is it an illuminator for an EO camera?  Is it a visible spectrum navigation light / beacon?  Is it a status light on an EO camera?  It's just not clear.

message_definitions/v1.0/common.xml
```diff
@@ -1624,6 +1624,14 @@
...
+        <param index="2" label="EnableIR" minValue="0" maxValue="1" increment="1">0: IR Indicator OFF, 1: IR Indicator ON</param>
```
auturgy, line 1632: As per EO indicator.  Needs clarification

auturgy:
Not sure if this is fine grained enough.  You'll likely want flash rates for these.

mrivi:
@auturgy please have a look at the changes I made

message_definitions/v1.0/common.xml
```diff
@@ -1624,6 +1624,16 @@
         <description>Turns illuminators ON/OFF. An illuminator is a light source that is used for lighting up dark areas external to the sytstem: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
       </entry>
+      <entry value="406" name="MAV_CMD_LIGHT_INDICATOR_ON_OFF" hasLocation="false" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Turns indicator light source ON/OFF. This light source is meant to show the presence of a vehicle.</description>
+        <param index="1" label="EnableVisibleLight" minValue="0" maxValue="1" increment="1">Visibile light 0: EO Indicator OFF, 1: EO Indicator ON</param>
```
hamishwillee, line 1631: typo Visibile

message_definitions/v1.0/common.xml
```diff
@@ -1624,6 +1624,16 @@
...
+        <param index="2" label="EnableInfraRed" minValue="0" maxValue="1" increment="1">0: IR Indicator OFF, 1: IR Indicator ON</param>
+        <param index="3" label="EnablePositionLight" minValue="0" maxValue="1" increment="1">Position lights (red/green) to indicate the vehicle heading, 0: Position lights OFF, 1: Positions Lights ON</param>
+        <param index="4" label="BlinkDuration" units="seconds">Portion of the period where the light is ON. </param>
```
hamishwillee, line 1634: Do all the lights blink, and do they all blink at same rate and period? The way this message is set up I would assume that any light that is on will also blink at the same specified rate.

hamishwillee:
Any synergy with [MAV_CMD_ILLUMINATOR_ON_OFF](https://mavlink.io/en/messages/common.html#MAV_CMD_ILLUMINATOR_ON_OFF)? I.e. could we merge one/supercede one?

hamishwillee:
@auturgy This command useful for ArduPilot too? I.e. under new governance approach it can go into common.xml once their is an implementation?

message_definitions/v1.0/common.xml
mrivi: My intent was that you have only one enabled and the rate applied to the enabled one.

mrivi:
@hamishwillee I have separated it from the illuminator because of the  different concept
- illuminator: gives light to an area such that the vehicle can see the environment with cameras etc
- indicator: the light source is meant to show the presence of the vehicle

message_definitions/v1.0/common.xml
hamishwillee: The thing is that there is nothing in the message that captures this intent or prevents it being mis-used.  To do what you want change param 1 to an enum with the different options you have specified (IR, visible, etc) and remove the specific params. Then it is only possible to enable one light at a time, and it is obvious how you are applying it. Also, if we come up with more light sources it is easy to extend. 

And now you can see why I thought that illuminator might be a reasonable option - this is really turn something on an off, possibly with a cycle.

mrivi:
@hamishwillee I made the changes

hamishwillee:
@mrivi 
- Thanks, this looks good to me. 
- FYI Mavlink moving towards a model where WIP means we have a working implementation a flight stack, so this is here for a shakedown couple of months while we do testing. Is this the case?

@auturgy Can you please re-review/are you OK with this going in (i.e. ArduPilot interested in implementing this).

PR #1531, opened 2020-11-27
add messages for events interface

bkueng:
From https://github.com/mavlink/rfcs/pull/14 (yes it's been a while).

Implemetation for PX4: https://github.com/PX4/PX4-Autopilot/pull/16293

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4299,6 +4299,18 @@
       <entry value="6" name="MAG_CAL_BAD_ORIENTATION"/>
       <entry value="7" name="MAG_CAL_BAD_RADIUS"/>
     </enum>
+    <enum name="MAV_EVENT_ERROR_REASON">
+      <description>Reason for an event error response.</description>
+      <entry value="0" name="MAV_EVENT_ERROR_REASON_UNAVAILABLE">
+        <description>The requested event is not available (anymore).</description>
+      </entry>
+    </enum>
+    <enum name="MAV_EVENT_CURRENT_SEQUENCE_FLAGS">
+      <description>Flags for CURRENT_EVENT_SEQUENCE.</description>
+      <entry value="1" name="MAV_EVENT_CURRENT_SEQUENCE_FLAGS_RESET">
+        <description>A sequence reset has happened (e.g. vehicle reboot).</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
@@ -6675,6 +6687,45 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
+    </message>
+    <!-- Events Protocol -->
+    <message id="410" name="EVENT">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Event message. Each new event from a particular component gets a new sequence number. The same message might be sent multiple times if (re-)requested. Most events are broadcast, some can be specific to a target component (as receivers keep track of the sequence for missed events, all events need to be broadcast. Thus we use destination_component instead of target_component).</description>
+      <field type="uint8_t" name="destination_component">Component ID</field>
+      <field type="uint8_t" name="destination_system">System ID</field>
+      <field type="uint32_t" name="id">Event ID (as defined in the component metadata)</field>
+      <field type="uint32_t" name="event_time_boot_ms" units="ms">Timestamp (time since system boot when the event happened).</field>
+      <field type="uint16_t" name="sequence">Sequence number.</field>
+      <field type="uint8_t" name="log_levels">Log levels: 4 bits MSB: internal (for logging purposes), 4 bits LSB: external. Levels: Emergency = 0, Alert = 1, Critical = 2, Error = 3, Warning = 4, Notice = 5, Info = 6, Debug = 7, Protocol = 8, Disabled = 9</field>
+      <field type="uint8_t[40]" name="arguments">Arguments (depend on event ID).</field>
+    </message>
+    <message id="411" name="CURRENT_EVENT_SEQUENCE">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Regular broadcast for the current latest event sequence number for a component. This is used to check for dropped events.</description>
+      <field type="uint16_t" name="sequence">Sequence number.</field>
+      <field type="uint8_t" name="flags" enum="MAV_EVENT_CURRENT_SEQUENCE_FLAGS" display="bitmask">Flag bitset.</field>
+    </message>
+    <message id="412" name="REQUEST_EVENT">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Request one or more events to be (re-)sent. If first_sequence==last_sequence, only a single event is requested. Note that first_sequence can be larger than last_sequence (because the sequence number can wrap). Each sequence will trigger an EVENT or EVENT_ERROR response.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint16_t" name="first_sequence">First sequence number of the requested event.</field>
+      <field type="uint16_t" name="last_sequence">Last sequence number of the requested event.</field>
+    </message>
+    <message id="413" name="RESPONSE_EVENT_ERROR">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Response to a REQUEST_EVENT in case of an error (e.g. the event is not available anymore).</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint16_t" name="sequence">Sequence number.</field>
+      <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
+      <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -772,6 +772,9 @@
       <entry value="3" name="COMP_METADATA_TYPE_PERIPHERALS">
         <description>Meta data which specifies potential external peripherals that do not talk MAVLink</description>
       </entry>
+      <entry value="4" name="COMP_METADATA_TYPE_EVENTS">
+        <description>Meta data for events interface</description>
+      </entry>
     </enum>
     <enum name="PARAM_TRANSACTION_TRANSPORT">
       <description>Possible transport layers to set and get parameters via mavlink during a parameter transaction.</description>


message_definitions/v1.0/common.xml
```diff
@@ -6667,6 +6682,37 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
+    <!-- Events Protocol -->
+    <message id="410" name="EVENT">
+      <description>Event message. The same message might be sent multiple times if (re-)requested. Most events are broadcast, some can be specific to a target component (as receivers keep track of the sequence for missed events, all events need to be broadcast. Thus we use destination_component instead of target_component).</description>
+      <field type="uint8_t" name="destination_component">Component ID</field>
```
hamishwillee, line 6688: This and following should be named `target_system` and `target_component`, and probably reordered system first. The C toolchain has helpers to ease extraction from the payload if these are named correctly.

hamishwillee: Sorry, ignore that, I missed the text " Thus we use destination_component instead of target_component". 

I'm not sure I understand the reasoning. I think you're saying that this is broadcast, and every component must look for every event using the sequence number. So if an event is intended for a particular component you have to use some other name than target_ so that it isn't treated as a targetted message for the purposes of routing/handling etc.

Is that right?

message_definitions/v1.0/common.xml
```diff
@@ -6667,6 +6682,37 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="format" enum="TUNE_FORMAT" display="bitmask">Bitfield of supported tune formats.</field>
     </message>
+    <!-- Events Protocol -->
+    <message id="410" name="EVENT">
+      <description>Event message. The same message might be sent multiple times if (re-)requested. Most events are broadcast, some can be specific to a target component (as receivers keep track of the sequence for missed events, all events need to be broadcast. Thus we use destination_component instead of target_component).</description>
```
hamishwillee, line 6687: FWIW I think it worth nothing here that each new event from a particular component gets a new sequence number

message_definitions/v1.0/common.xml
```diff
@@ -6667,6 +6682,37 @@
...
+      <field type="uint8_t" name="destination_component">Component ID</field>
+      <field type="uint8_t" name="destination_system">System ID</field>
+      <field type="uint32_t" name="id">Event ID</field>
```
hamishwillee, line 6690: FWIW I think it worth nothing here that the supported IDs are defined in component metadata XML files.

message_definitions/v1.0/common.xml
```diff
@@ -6667,6 +6682,37 @@
...
+      <field type="uint8_t" name="log_levels">Log levels: 4 bits MSB: internal, 4 bits LSB: external.</field>
+      <field type="uint8_t[40]" name="arguments">Arguments (depend on event ID).</field>
+    </message>
+    <message id="411" name="CURRENT_EVENT_SEQUENCE">
+      <description>Regular broadcast for the current latest event sequence number for a component. This is used to check for dropped events.</description>
+      <field type="uint16_t" name="sequence">Sequence number.</field>
+      <field type="uint8_t" name="flags" enum="MAV_EVENT_CURRENT_SEQUENCE_FLAGS" display="bitmask">Flag bitset.</field>
+    </message>
+    <message id="412" name="REQUEST_EVENT">
+      <description>Request one or more events to be (re-)sent. If first_sequence==last_sequence, only a single event is requested. Note that first_sequence can be larger than last_sequence. Each sequence will trigger an EVENT or EVENT_ERROR response.</description>
```
hamishwillee, line 6702: Re "Note that first_sequence can be larger than last_sequence." - presumably because the sequence number can wrap? Worth saying that ("Note that first_sequence can be larger than last_sequence (because the sequence number can wrap)"

hamishwillee:
@auturgy @LorenzMeier @julianoes Can you please review this PR. Note, something that surprised me was that the spec uses XML definition for the events and adds an id to allow this to be retrieved using a component information file. I'd quite like to see some example event XML appear in the repo along with this for review.

James can you please confirm that ArduPilot intends to also implement support for events in a reasonable timeframe so that this might be accepted into common.xml under "new governance approach".

bkueng:
> Note, something that surprised me was that the spec uses XML definition for the events and adds an id to allow this to be retrieved using a component information file. I'd quite like to see some example event XML appear in the repo along with this for review.

See https://github.com/bkueng/libevents/blob/master/events/common.json.
- This is a separate repo that contains validation scripts, parsers, protocol implementation and the common event definitions, and is meant to be used by different projects. I think it could be under https://github.com/mavlink/.
- It's separate from MAVLink, since events can be used outside of MAVLink, e.g. in logs.
- It's using JSON mainly just because the COMPONENT_INFORMATION API uses JSON.

message_definitions/v1.0/common.xml
bkueng: Ok, done

message_definitions/v1.0/common.xml
bkueng: Done

message_definitions/v1.0/common.xml
bkueng: Ok, done

message_definitions/v1.0/common.xml
bkueng: Yes

auturgy:
Let’s not rush this: the component.json that this leverages was merged without discussion, isn’t documented, and hence itself can’t be considered stable.
The way it’s implemented here broadens the original intent, from mavlink enabled peripherals to the autopilot itself. This may well be sound, but is significant and requires discussion.
We also need to consider how this is going to interact with tooling. Having disparate, uncoordinated json files means that every log would need to include every component definition file for the system, and every log analysis tool would need to be able to parse them to make sense of the log - and as there’s no deconfliction of event ID’s between components, this is a two step process.  That’s a considerable disruption to the mechanics of a significant number of tools (both open source and proprietary).

@peterbarker @tridge

hamishwillee:
@auturgy If you're saying that we should formally approve the component information stuff into MAVLink before basing other stuff on it, then yes. 

In detail though, we were always and openly intending that the component metadata be used for the autopilot (not just peripherals), yes some proposed .json is in the tree for parameters and versioning - there is nothing in for events, and IMO something should be proposed in this PR so we can evaluate it. 

The json files will need to be co-ordinated.

@bkueng Can we have you along to the MAVLink call to move this forward - there is one today; not sure when the next one is.

hamishwillee:
Discussed in [Devcall](https://github.com/mavlink/mavlink/wiki/20201209-Dev-Meeting).  Action on James to get some review from ArduPilot.

hamishwillee:
@bkueng Is this the PR you are using for Events (still) or did it get superseded. If it is "live" should we be pushing into development.xml until prototyped?

bkueng:
> @bkueng Is this the PR you are using for Events

Yes. It's already prototyped and running, we could mark it as wip?

hamishwillee:
> Yes. It's already prototyped and running, we could mark it as wip?

Yes.

The "right" way is that it goes in development.xml as wip, and lives there until we're all happy for it to go into common.xml - ie a very serious commitment from at least one other flight stack to integrate it.

But since we're only just adopting that process and ArduPilot have also indicated strong support for this, I think we can merge it if you update to WIP and rebase.

What we need though is a commitment/clear timeline for removal of that WIP (that's kind of what the migration from development.xml into common.xml gives you - a known point at which people can and should adopt this).

FYI @auturgy

auturgy:
Let’s get it in

Regards,

James

bkueng:
Rebased & added WIP

hamishwillee:
@julianoes This has failing tests related to "Node". That's a new one on me - do you know where it came from?

I'm guessing some mismatch with Pymavlink.

peterbarker:
On Wed, 31 Mar 2021, Hamish Willee wrote:

> @julianoes This has failing tests related to "Node". That's a new one on me - do you know where it came from?
> 
> I'm guessing some mismatch with Pymavlink.

FWIW I saw some traffic to do with updating security vulnerabilities to do 
with the javascript bindings.

julianoes:
> @julianoes This has failing tests related to "Node". That's a new one on me - do you know where it came from?

I just ported this test from travis-ci, I have no idea what they do 🤷‍♂️ .

hamishwillee:
I have asked what they do on ArduPilot discord - [this link - FMI only](https://discordapp.com/channels/674039678562861068/728017546313466047/829588195213049897).

davidbuzz:
ok, so a initial review of the logs, its pretty clear that the Node parser/test suite doesn't like this new 'event' packet type... you can see its got "ticks" against al the autotests except for these new ones...  I think i'm going to speculate that something in the parser node/javascript parser/generator/autotest code uses 'event'  for some other purpose....  i will pull this PR and try to replicate locally....
...
  ✓ id1353 encode and decode supported_tunes from C using ardupilotmega/2.0 
    ✓ id1354 encode and decode supported_tunes from C using ardupilotmega/2.0 signed
    3) id1355 encode and decode event from C using ardupilotmega/2.0 
    4) id1356 encode and decode event from C using ardupilotmega/2.0 signed
    ✓ id1357 encode and decode current_event_sequence from C using ardupilotmega/2.0 
    ✓ id1358 encode and decode current_event_sequence from C using ardupilotmega/2.0 signed
...

davidbuzz:
ok, ive identified that the Node/js is having a hissy-fit with the xml called 'arguments'   ....
name="arguments"

call it just about anything else and it'll pass (eg 'zzzarguments').    ... or if you prefer, i can do a PR to the node parser to fix it.

bkueng:
Thanks @davidbuzz. I changed it to `args`, CI is passing now. If it's easy to fix I prefer that, as I'm also using `arguments` in json files and code. Otherwise I'll have to update those as well.

message_definitions/v1.0/common.xml
```diff
@@ -6676,6 +6691,45 @@
...
+    <message id="410" name="EVENT">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Event message. Each new event from a particular component gets a new sequence number. The same message might be sent multiple times if (re-)requested. Most events are broadcast, some can be specific to a target component (as receivers keep track of the sequence for missed events, all events need to be broadcast. Thus we use destination_component instead of target_component).</description>
+      <field type="uint8_t" name="destination_component">Component ID</field>
+      <field type="uint8_t" name="destination_system">System ID</field>
+      <field type="uint32_t" name="id">Event ID (as defined in the component metadata)</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot when the event happened).</field>
+      <field type="uint16_t" name="sequence">Sequence number.</field>
+      <field type="uint8_t" name="log_levels">Log levels: 4 bits MSB: internal, 4 bits LSB: external.</field>
```
peterbarker, line 6704: This is somewhat opaque.

Any documentation on this (which should possibly be included right here....)?

message_definitions/v1.0/common.xml
```diff
@@ -4299,6 +4302,18 @@
...
+    <enum name="MAV_EVENT_ERROR_REASON">
+      <description>Reason for an event error response.</description>
+      <entry value="0" name="MAV_EVENT_ERROR_REASON_UNAVAILABLE">
+        <description>The requested event is not available (anymore).</description>
+      </entry>
+    </enum>
+    <enum name="MAV_EVENT_CURRENT_SEQUENCE_FLAGS">
+      <description>Flags for CURRENT_EVENT_SEQUENCE.</description>
+      <entry value="1" name="MAV_EVENT_CURRENT_SEQUENCE_FLAGS_RESET">
+        <description>A sequence reset has happened (e.g. vehicle reboot).</description>
```
peterbarker, line 4314: Really not sure how this helps anything?

A reset event is in the eye of the receiver, after all - if someone pops up after you've rebooted they're uninterested in this.

message_definitions/v1.0/common.xml
```diff
@@ -6676,6 +6691,45 @@
...
+      <field type="uint16_t" name="last_sequence">Last sequence number of the requested event.</field>
+    </message>
+    <message id="413" name="EVENT_ERROR">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Response to a REQUEST_EVENT in case of an error (e.g. the event is not available anymore).</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint16_t" name="sequence">Sequence number.</field>
+      <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
```
peterbarker, line 6730: Should this be `REQUEST_EVENT_ERROR`, then?

message_definitions/v1.0/common.xml
```diff
@@ -6676,6 +6691,45 @@
...
     </message>
+    <!-- Events Protocol -->
+    <message id="410" name="EVENT">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Event message. Each new event from a particular component gets a new sequence number. The same message might be sent multiple times if (re-)requested. Most events are broadcast, some can be specific to a target component (as receivers keep track of the sequence for missed events, all events need to be broadcast. Thus we use destination_component instead of target_component).</description>
+      <field type="uint8_t" name="destination_component">Component ID</field>
+      <field type="uint8_t" name="destination_system">System ID</field>
+      <field type="uint32_t" name="id">Event ID (as defined in the component metadata)</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot when the event happened).</field>
```
peterbarker, line 6702: Using `time_boot_ms` break's MAVProxy's reboot-detection code - it looks for `time_boot_ms` in any packet and decided whether a reboot has occured based on that.  Since these are always in the past for events (give or take integer wrapping), this is going to cause confusion.

message_definitions/v1.0/common.xml
bkueng: I added it here, the intention was to have this part of the protocol docs.

message_definitions/v1.0/common.xml
bkueng: A receiver needs to know as the sequence number jumps and it would misinterpret as dropped events otherwise.

message_definitions/v1.0/common.xml
bkueng: Not sure I understand? A separate request for error?

message_definitions/v1.0/common.xml
bkueng: I can rename the field if you prefer. `timestamp_ms`?

hamishwillee: @peterbarker Reminder, a couple of responses/questions for you on this PR. I'm hoping to get this merged soon, when we're all happy.

auturgy:
FYI https://github.com/ArduPilot/ardupilot/pull/17133

message_definitions/v1.0/common.xml
peterbarker: I still don't get how this works if people are dropping messages left-right-and-centre.  This flag is only set while no events have ever been sent?  So if the GCS never gets any of those they'll never see the "reset"?

message_definitions/v1.0/common.xml
peterbarker: Why not just re-broadcast rather than this pseudo-targetting?  I doubt the system who is re-requesting could be bothered to see if these two fields match their own IDs - just the sequence numbers.

message_definitions/v1.0/common.xml
peterbarker: `event_time_boot_ms`, perhaps?

message_definitions/v1.0/common.xml
peterbarker: I'm saying that you've added a field that appears to be specific to someone requesting a message which is out-of-range with `REQUEST_EVENT`.  That doesn't seem correct in a packet called `EVENT_ERROR`, thus my suggestion that you pair this message with `REQUEST_EVENT` by naming it `REQUEST_EVENT_ERROR`.

peterbarker (commented):
This PR doesn't seem to specify how the URL for the relevant metadata is retrieved.  How's that done?

My implementation of this protocol exploded (as well as my brain) when I started to look at how the component-metdata protocol is described in this standard.

message_definitions/v1.0/common.xml
bkueng: Because an event could be targeted at a certain component in the first place. It's a special use-case, the default is to broadcast (also for a re-request).

message_definitions/v1.0/common.xml
bkueng: Yes only initially, I implemented an additional detection based on timestamp. Both is needed to handle message reodering and e.g. quick sitl restarts.

message_definitions/v1.0/common.xml
bkueng: Got it. Just from the name I'd interpret `REQUEST_EVENT_ERROR` to request an event error. What about `RESPONSE_EVENT_ERROR`?

message_definitions/v1.0/common.xml
bkueng: Updated.

bkueng:
> This PR doesn't seem to specify how the URL for the relevant metadata is retrieved. How's that done?

There's a lot more to it than just mavlink message specs. Did you read 
https://docs.google.com/document/d/18qdDgfML97lItom09MJhngYnFzAm1zFdmlCKG7TaBpg/edit?
Feel free to contact me directly on PX4 slack if you have more questions.

bkueng:
@hamishwillee @peterbarker can we merge this? I'm ok with making additional changes later on should that be needed.

hamishwillee:
I have no comment, but it would be good if @peterbarker could confirm he is OK with your responses. I'll ping.

message_definitions/v1.0/common.xml
peterbarker: I still don't understand why you'd want to include the destination sysid/component id - *even* if the message was re-requested.  Where's the advantage?  I mean, some of the disadvantages are obvious - complexity and wire-size.

message_definitions/v1.0/common.xml
peterbarker: It really sounds like you want a generation rather than a simple flag here.  So, for example, ArduPilot keeps a value stored in a `STAT_BOOTCOUNT` which is incremented each time the autopilot boots.

bkueng: That would be an option, but I really don't want to add that additional requirement of keeping state between boots/restarts.

hamishwillee:
According to @bkueng this has been reviewed by @peterbarker and we're happy for this to be merged

peterbarker:
On Wed, 28 Apr 2021, Hamish Willee wrote:

> According to @bkueng this has been reviewed by @peterbarker and we're happy for this to be merged

To be clear, I didn't approve this.

I did discover several things which I believe *should* be addressed but I 
can't formalise those things without spending more time than I can really 
justify on this.

I just couldn't justify standing in the way of it being merged given the 
implementation isn't straight forward.  However, I doubt this would be 
implemented by ArduPilot in its current form.

hamishwillee:
@peterbarker Thanks for the clarification - I misinterpreted Beats comments on this.

The process as I understand it is that this will go through a period of prototyping, following which some variation of this will be "proved" as viable and become part of the standard. At that point if ArduPilot wanted to do something different you would need to propose an enhancement. Much better to get us all on the same page now. Appreciate this is time consuming.

Is there a list of the things that you think need to be addressed at "high level" (or is generating that list what you mean by not having time to formalise the problem)?

peterbarker:
On Wed, 28 Apr 2021, Hamish Willee wrote:

> Is there a list of the things that you think need to be addressed at "high level" (or is generating that list what you mean by not having time to formalise the problem)?

That's right.

Beat and others have obviously put a lot of time and thought into this, 
and I'm having trouble finding the headspace to make an equivalent effort 
(already letting clients down....).

hamishwillee:
@peterbarker I completely understand. Do you think you might find headspace in a known period - e.g. say 2-3 months? It may be that the number of deployed systems would be very low at that point and we could still make relatively constrained changes.

PR #1533, opened 2020-12-03
common: abolish PARAM_EXT_*_TRIMMED again

julianoes:
It turns out that supporting the _TRIMMED message set for PARAM_EXT is not as straightforward as it should be:
- Having to implement both variants of the messages ends up being a bit of code/binary bloat.
- When doing a param set without previously having done a get or request list is tricky because it is not clear whether TRIMMED can be used. This means that retransmission is required in case the wrong one is picked which makes it more brittle and can add some latency. I realized all this once I move to the implementation for MAVSDK.

Furthermore, the changes have not actually been adopted anywhere that I'm aware of, and they are not documented yet either.

I therefore came around and changed my opinion and vote to get rid of these changes again.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6345,23 +6345,18 @@
       <field type="uint32_t" name="sw_vcs_commit">Version control system (VCS) revision identifier (e.g. git short commit hash). Zero if unknown.</field>
     </message>
     <message id="320" name="PARAM_EXT_REQUEST_READ">
-      <description>Request to read the value of a parameter with either the param_id string id or param_index. PARAM_EXT_VALUE or PARAM_EXT_VALUE_TRIMMED should be emitted in response (see field: trimmed).</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Makes sense to me.

@auturgy @LorenzMeier I plan to merge this removal of the unused PARAM_EXT_ TRIMMED messages if there is no objection within a day or so (and I don't imagine there will be).

olliw42:
I did implement them in STorM32 ;) LOL
I think it would be high time to get to grips with an improved param protocol, and I looked at the trimmed messages as a sort of step towards this.
I didn't found it such a big code bloat to add them in, not really, but it's true things become unclear if one hasn't done a request list before. But my understanding was that doing a request list first is or was kind of a requirement anyhow, since if you just send a set it's not clear anyway if that parameter is supported in the first place, right. Handling that case is not less nasty than the trimmed thing. So, this didn't looked like a real issue to me, or at least I was thinking it isn't considered an issue for MAVLink.
my 2 cents.

julianoes:
> but it's true things become unclear if one hasn't done a request list before.

Correct. And for MAVSDK I actually don't do that step because I didn't see a reason why. I guess now I do :).

PR #1538, opened 2020-12-10
Sync to Ardupilot downstream 2020-12-10

amilcarlucas:


message_definitions/v1.0/common.xml
```diff
@@ -1276,8 +1276,8 @@
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5" label="Latitude" units="degE7">Latitude of ROI location</param>
-        <param index="6" label="Longitude" units="degE7">Longitude of ROI location</param>
+        <param index="5" label="Latitude">Latitude of ROI location</param>
```
hamishwillee, line 1279: Why remove the units?

message_definitions/v1.0/ardupilotmega.xml
```diff
@@ -1652,5 +1653,19 @@
       <field type="float" name="max_value">OSD parameter maximum value.</field>
       <field type="float" name="increment">OSD parameter increment.</field>
     </message>
+    <message id="11037" name="OBSTACLE_DISTANCE_3D">
```
hamishwillee, line 1656: @mrivi FYI, this is a new ArduPilot dialect message that gives obstacle distance in 3D. Is this of interest to PX4? Note, if you think PX4 might take this one day, then good to comment on anything that springs to mind now.

message_definitions/v1.0/common.xml
amilcarlucas: Beacuse of the hole COMMAND_INT vs COMMAND_LONG issues

hamishwillee: I'm pretty sure the result of our discussions was leave all these alone - you shouldn't be using COMMAND_LONG with a latitude or longitude.

amilcarlucas: No the the result was:
- me apealing for adding degE7 everywhere
- everybody else deciding to remove it everywhere

I'm jus complying with everybody else's whiches :)

hamishwillee: I don't know who "everyone" here is - me and Julian agreed with you and I haven't removed them.

amilcarlucas: I missunderstood it then. So should we add degE7 everywhere?

hamishwillee: If by everywhere you mean on all lat/lon values in the global frames then that is what I would prefer - unless the description explicitly specifies that the values might be in some other units. I think it is better for consistency - yes the lats/lon can be sent in a float variant of a message, but they shouldn't be.

amilcarlucas: How about doing that in another PR, and just merging this one?

hamishwillee: How about removing this change and the next line :-)

amilcarlucas: done

hamishwillee:
@amilcarlucas Thanks - looks great.

Unfortunately it is failing due to a units error. Since I _think_ we have the latest pymavlink, is this pending some PR?

amilcarlucas:
No the error is not in this PR. The error is already in master.
The cm^3 units are in master, and are not introduced by this PR.

amilcarlucas:
@hamishwillee ping

hamishwillee:
@amilcarlucas Thanks for the clarification and I've responded to your message https://github.com/mavlink/mavlink/commit/73afa7c60c3fa43d230b66a777eca2ec6b1fbcff#commitcomment-46997036

I appreciate that the problem is in master. Even though I am fairly confident that this has no problems, I feel more confident if it passes tests before merging. So unless you have strong need to get this in, lets sort out the units issue first.

amilcarlucas:
Yes. Please add cm^3 to the xsd and re-trigger the CI tests?

amilcarlucas:
All tests pass now (after I re-triggered CI with a force push)

hamishwillee:
Thanks @amilcarlucas - merged!

PR #1549, opened 2020-12-25
common.xml: UINT16_MAX-1 on RC_CHANNELS_OVERRIDE means return-to-rc-radio

peterbarker:
Equivalent PR into mavlink repo is here: https://github.com/ArduPilot/mavlink/pull/180

PR to ArduPilot to use these changes is here: https://github.com/ArduPilot/ardupilot/pull/15065

Allows for a mechanism for a GCS to release channels back to the RC; at the moment the lower channels allow for instant-release to RC but the higher don't.

Also clarify the descriptions by putting all of the user's options in one place.


LorenzMeier (approved):
Makes sense, thanks!

PR #1551, opened 2020-12-29
common.xml: smart_battery_info fix cycles not available to UINT16_MAX

hendjoshsr71:
Changes cycle_count to an int32 to allow for -1 if not available


LorenzMeier:
The issue with this change is that due to the hardening of the protocol, this message would be (intentionally) incompatible with its previous versions. How about you document instead that a value of UINT32_MAX indicates that it is not available?

hendjoshsr71:
Makes sense! I changed it so that cycles not available is set to UINT16_MAX.

LorenzMeier (approved):

LorenzMeier:
@auturgy I'm merging this right away since it is an obvious fix (-1 == UINT16_MAX for an unsigned int)

auturgy:
@LorenzMeier no worries: we knew about this one

PR #1552, opened 2020-12-30
MAV_STATE enum, specify values explicitely

olliw42:
this specifies the values for all entries of the MAV_STATE enum

they are currently auto-generated by the code generator(s). I filled them in with what mavgen is auto-generating.

It may be seen as just a matter of convenience, but I actually think that this is not good practice, i.e., a specification should not have to rely on whatever code generator is doing, but rather be explicit and specify what the specification is to be. So, I actually also think that all the code generators should rather throw an error instead of auto-generating enum values...

COMMENT: This appears to also apply to the enum definitions in the ualberta.xml dialect ... but this is not a standard dialect, so I haven't touched that.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -235,28 +235,28 @@
       <entry value="0" name="MAV_STATE_UNINIT">
         <description>Uninitialized system, state is unknown.</description>
       </entry>
-      <entry name="MAV_STATE_BOOT">
+      <entry value="1" name="MAV_STATE_BOOT">
         <description>System is booting up.</description>
       </entry>
-      <entry name="MAV_STATE_CALIBRATING">
+      <entry value="2" name="MAV_STATE_CALIBRATING">
         <description>System is calibrating and not flight-ready.</description>
       </entry>
-      <entry name="MAV_STATE_STANDBY">
+      <entry value="3" name="MAV_STATE_STANDBY">
         <description>System is grounded and on standby. It can be launched any time.</description>
       </entry>
-      <entry name="MAV_STATE_ACTIVE">
+      <entry value="4" name="MAV_STATE_ACTIVE">
         <description>System is active and might be already airborne. Motors are engaged.</description>
       </entry>
-      <entry name="MAV_STATE_CRITICAL">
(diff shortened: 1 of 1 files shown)


hamishwillee:
@olliw42 If you feel strongly that the code generate should throw an error on missing enum, perhaps post an issue on ArduPilot/pymavlink to discuss and cross link back here. 

Personally I agreed; it is too easy for the addition of an enum value to break ordering unintentionally and "invisibly". Either way this is a no brainer; merging.

olliw42:
THX

I do feel strongly that the code generators should throw an error on missing enums, but not enough so to make me becoming active on ArduPilot/pymavlink (I had my share of it). Wanted to mention it though :)

in some sense it should actually not be a code generator thing but a mavlink thing, i.e., such enums should simply never make it into any xml (and the code generator should just be the last safety net). No idea if a corresponding (auto-) test could be run. 

I quickly looked into pymavklink/mavschema.xsd. I can't say I really understand this, but note that for enum there is not even an attribute "value". Otherwise I would have thought that maybe <xs:attribute ref="value" use="required"/> might have done it.

Maybe it's just easier to "manually" ensure that enums are properly populated.

Having said this, if you want to, pl feel free to direct me to also do it for the enum definitions in the ualberta.xml dialect ...

hamishwillee:
@amilcarlucas Many enums do not have values, or just have the first value and the rest are automatically populated with incremented values by the generator. OllieW takes the position that a missing enum value is a bug, and this should therefore result in an error being thrown for the XML definition.

I tend to agree. The reason being that it is very easy to inject an enum value inside an existing sequence and break anyone who is using a later value.

What do you think? If you agree we could go through and add values for all dialects, then add an appropriate check in the XSD.

olliw42:
> Many enums do not have values

> we could go through and add values for all dialects

When running the C generator the affected enums are displayed (a message "An enum value was auto-generated: xxx" is printed). 

When running it for all.xml, it are only the three enum definitions in the ualberta.xml dialect which are incomplete.

hamishwillee:
@olliw42 If it is just ualberta.xml and you're happy to make the updates then that would be great. We'd then ask amlicar to add a test to the xsd.

PR #1554, opened 2021-01-05
changed uint8_t[x] to char[x] where used for strings

zzaurak:
I've noticed that when using messages that contain uint8_t[] arrays for stringvalues the generated code for python (pymavlink) throws index out of range errors in case the string is not exactly the required length or longer. This can be prevented by padding the Strings with the appropriate number of zero bytes, but there is no easy and clean(as in simple) way to do this in python. At first I thought I needed to change the generator, but then I noticed that the python struct.pack method padds data automatically when it is a char array 

> For the 's' format character, the count is interpreted as the length of the bytes, not a repeat count like for the other format characters; for example, '10s' means a single 10-byte string, while '10c' means 10 characters. If a count is not given, it defaults to 1. For packing, the string is truncated or padded with null bytes as appropriate to make it fit. 
[https://docs.python.org/3/library/struct.html](https://docs.python.org/3/library/struct.html)

As I can see no negative impact to the change of this datatype at the few places where it is used for a string, I made this pull-request. I am looking forward to the discussion!

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5574,9 +5574,9 @@
       <description>The autopilot is requesting a resource (file, binary, other type of data)</description>
       <field type="uint8_t" name="request_id">Request ID. This ID should be re-used when sending back URI contents</field>
       <field type="uint8_t" name="uri_type">The type of requested URI. 0 = a file via URL. 1 = a UAVCAN binary</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@zzaurak The problem with this is that it is an incompatible change. Every message is sent with a CRC that is generated from the field types and names, so any library that has not been updated will not be able to handle this changed message. 

You would be better off changing the generator

PR #1555, opened 2021-01-06
improve format_xml.sh script, tab to spaces

olliw42:
this attempts to improve the script format_xml.sh, also making it to work as it is saying itself:

* It cannot be run with a filename argument, i.e., it cannot be applied to one specific file. That's particularly weird, since when run with option -c it is saying "run ./scripts/format_xml.sh $f" when a file needs formatting! This has been corrected by removing all options from the arguments list and filling xml_files with the first argument if there is an argument. (lines 27 - 37).

* It could not be easily applied to a specific directory, such as (the potentially upcoming) mavlink/external/dialects. Currently it would process ALL xml files it finds in the tree of subfolders from which the script is run. This has been overcome by a new option -d, which sets the $(find ...) accordingly.

* It cannot be easily figured out what exactly the formatting has changed. This has been overcome by a new option -o, which copies the old file to $f.old before it overwrites with the formatted file. This is for sure only a matter of convenience.

* I did change slightly the output. (line 37, line 58)

* In some places indenting was achieved by tabs in others by spaces; tabs have been replaced by spaces now.

Comment: show_help doesn't work for me. Since I don't know if that is just my installation, I haven't fooled around with this (I'm using git powershell on win7). I guess it would be more robust to simply do some echo.

### DIFF
--- a/scripts/format_xml.sh
+++ b/scripts/format_xml.sh
@@ -6,8 +6,10 @@ OPTIND=1         # Reset in case getopts has been used previously in the shell.
 
 # Initialize variables
 mode="format"
+xml_dir="."
+keep_old=0
 
-while getopts "h?c" opt; do
+while getopts "h?cd:o" opt; do
     case "$opt" in
     h|\?)
         show_help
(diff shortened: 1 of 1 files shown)


hamishwillee:
Looks very sane.  @amilcarlucasWould you like to test/review? If not let me know and I'll look at it next week.

olliw42:
blank missing between amilcarlucas and Would :)

hamishwillee:
@olliw42 There doesn't appear to be a facepalm or muppet icon on github, or I'd use it. 

@amilcarlucas Please see https://github.com/mavlink/mavlink/pull/1555#issuecomment-759849626

amilcarlucas:
Looks good to me, but I have no time to test it.

hamishwillee:
@julianoes Can you review this in the next couple of weeks?

julianoes (approved):
I'm not good at bash but if it works then I'd say it's good to go in.

olliw42:
I think what @hamishwillee would love to hear is that you or someone not me has tested if it works :)

julianoes:
I see :). I just tested it and the canonical `scripts/format_xml.sh` which I would use seems to work fine.

hamishwillee:
Good enough for me. Thanks guys.

olliw42:
GREAT! MANY THX

hamishwillee:
Thanks all.

PR #1559, opened 2021-01-20
common: add MAV_CMD_CAMERA_TRACK_SCENE

JonasVautherin:
This adds a message for a "TRACK_SCENE" mode, next to the existing "TRACK_POINT" and "TRACK_RECTANGLE" messages.


JonasVautherin:
Note: should be renamed to something like "OPTICAL_STABILIZE" and come with evidence of a working implementation.

PR #1562, opened 2021-01-21
Enums: Specify values explicitely in all enums

olliw42:
does https://github.com/mavlink/mavlink/pull/1552#issuecomment-763305515

Adds values to all enum entries which didn't yet had a value.

Procedure:
I ran mavgen for C for each and every xml file, and checked for the "An enum value was auto-generated: xxx = yy" print outs. I thus did not only check and corrected the xmls covered by all.xml, but all.

Comment:
For autoquad.xml this resulted in MAV_DATA_STREAM_PROPULSION = 13, which IMHO just shows how 'dangerous' the autogeneration is, since this value would have changed if the MAV_DATA_STREAM enum in common.xml would have changed, breaking implementations...

### DIFF
--- a/message_definitions/v1.0/ASLUAV.xml
+++ b/message_definitions/v1.0/ASLUAV.xml
@@ -48,7 +48,7 @@
       <entry value="0" name="GSM_MODEM_TYPE_UNKNOWN">
         <description>not specified</description>
       </entry>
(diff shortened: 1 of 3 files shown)


hamishwillee:
Thanks @olliw42 Looks great. Merging.

PR #1566, opened 2021-01-29
ack waypoints may violate geofence

chobitsfan:
It allows FC informs GCS or companion computer that a mission item or waypoint may violates no-fly zone

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2470,6 +2470,9 @@
       <entry value="15" name="MAV_MISSION_OPERATION_CANCELLED">
         <description>Current mission operation cancelled (e.g. mission upload, mission download).</description>
       </entry>
+      <entry value="16" name="MAV_MISSION_GEOFENCE_VIOLATED">
+        <description>Mission item violates geofence.</description>
+      </entry>
     </enum>
     <enum name="MAV_SEVERITY">
       <description>Indicates the severity level, generally used for status messages to indicate their relative urgency. Based on RFC-5424 using expanded definitions at: http://www.kiwisyslog.com/kb/info:-syslog-message-levels/.</description>


LorenzMeier (approved):

LorenzMeier:
@auturgy This makes total sense and is easy to implement in PX4 with fast turnaround - how do you see this?

message_definitions/v1.0/common.xml
```diff
@@ -2470,6 +2470,9 @@
       <entry value="15" name="MAV_MISSION_OPERATION_CANCELLED">
         <description>Current mission operation cancelled (e.g. mission upload, mission download).</description>
       </entry>
+      <entry value="16" name="MAV_MISSION_NFZ_VIOLATION">
+        <description>Mission item violates no-fly zone.</description>
```
hamishwillee, line 2474: What is a no fly zone?

My concern is that this means "geofence". Imagine GCS uploads a plan with a mission, geofence and rally points. If the new mission is uploaded first and violates the pre-existing on-board geofence the upload will fail.

chobitsfan: Hi @hamishwillee  Thank you

> What is a no fly zone? My concern is that this means "geofence".

Yes. May I should rename it to geofence?

> Imagine GCS uploads a plan with a mission, geofence and rally points. If the new mission is uploaded first and violates the pre-existing on-board geofence the upload will fail.

Yes. I think this is the purpose of this enum. To inform GCS that new mission is rejected due to geofence violation.  Users should change missions or geofence setting.

hamishwillee: > Yes. I think this is the purpose of this enum. To inform GCS that new mission is rejected due to geofence violation. Users should change missions or geofence setting.

I don't think you understand my concern. The GCS is uploading a mission and _compatible_ geofence. But if it uploads the mission first and that mission is incompatible with the existing geofence, it will never get to upload the geofence that would work with the mission. Plan uploading will therefore fail.
For this to work, the geofence would have to be uploaded first, then the mission. Which might be OK, but is not something we specify.

chobitsfan: Thank you @hamishwillee 

> But if it uploads the mission first and that mission is incompatible with the existing geofence, it will never get to upload the geofence that would work with the mission

Yes. I agree with you. But I think it is up to how autopilot or GCS implement it. This enum simply provide a feedback entry

hamishwillee: @chobitsfan So you want to add a new feature that potentially breaks every GCS and you think that it is up to them to "live with it"? That's not how we roll :-)

hamishwillee:
@chobitsfan Had a discussion on this in the Devcall last night. Generally we see why it is useful for a companion to get this feedback. Call agrees that this might well break a GCS. What testing have you done? For example, what happens if you upload a plan where the mission conflicts with the existing geofence? Have you tried on Mission Planner and QGroundControl? 

@DonLakeFlyer @meee1  FYI, this proposes a new error during mission upload if the new mission violates the geofence (main use case is update from companion computer, where mission gets varied independent of geofence).
I am concerned this will break GCS uploading a plan if the mission is uploaded first and violates the geofence already on the vehicle (before the matching geofence plan is uploaded). If I'm right, can you suggest a path forward for supporting this use case? (e.g. can  we recommend that if a whole plan is updated a GCS should always do the geofence and rally points first. I guess the problem is that the parts of the plan are conceptually linked, at least in QGC, but not linked by MAVLink.)

message_definitions/v1.0/common.xml
chobitsfan: Thank you @hamishwillee I understand your concern now. You are right.

chobitsfan:
Thank you @hamishwillee 

> or example, what happens if you upload a plan where the mission conflicts with the existing geofence? Have you tried on Mission Planner and QGroundControl?

I tried mission planner. Mission planner did not uploading mission and geofence at the same time.
1. geofence existed -> upload "safe" mission (ok)
2. geofence existed -> upload "not safe" mission (failed)
3. mission existed -> upload new geofence which makes existed mission "not safe" (ok)

hamishwillee:
Thanks. I have asked @DonLakeFlyer to comment.

DonLakeFlyer:
Given the dynamic ordered nature of this I would expect this to be a feasibility failure (which fails when you try to actually run the mission) as opposed to an upload failure. Similar to how landing patterns for fixed wing work. You can upload the mission fine but the feasibility checker can fail on it due to a bad glide slope. Also for geofence there is a global fence which is controlled through parameters. That could be changed after the mission is already on the vehicle.  So the upload would succeed but then trying to run the mission would fail.

Same problem with a no fly zone. Those can update dynamically and may differ from the time the mission was uploaded to the vehicle from the time the mission is actually flown.

chobitsfan:
Hi @DonLakeFlyer Thank you for feedback

> You can upload the mission fine but the feasibility checker can fail on it due to a bad glide slope. 

Yes. I agree with you. The purpose of this new item is to allow FC to inform GCS or companion computer that new mission obviously violates geofence.

hamishwillee:
@DonLakeFlyer makes sense. So rather than getting a failure of mission upload, the companion will get some other notification that there is a feasibility failure and the new mission needs to be modified.

What is the form of that notification typically? Or is this all code that would need to be added to the autopilot?

DonLakeFlyer:
There is no mavlink standard for failures at feasibility check time. The firmware just refuses to fly the mission and spits out some status text as to why.

hamishwillee:
Not sure how to proceed this so have deferred back to the dev call next week. 

Unfortunately  [MAV_MISSION_RESULT](https://mavlink.io/en/messages/common.html#MAV_MISSION_RESULT) indicates that the upload has completed - ie with success or an error. If it was just a notification then the recipient could decide to discontinue the discussion.

chobitsfan:
Thank you @hamishwillee 

I agree with you and @DonLakeFlyer. This PR may cause problems to GCS and confusion for users. I will close it and try to find better solution. Thank you very much

hamishwillee:
Thanks @chobitsfan - good luck finding a solution to the problem!

PR #1569, opened 2021-02-03
Add MAV_VTOL_STATE_QC field

ThomasRigi:
Linked to https://github.com/PX4/PX4-Autopilot/pull/16691.

I would like to add the ability to trigger a QuadChute from an external command over MAVLINK (in our case using MAVROS). Our main motivation for it is to be able to react and stop the drone as fast as possible e.g. if it is about to breach a Geofence in FW mode.

The new field would also allow to send additional info on the drone's state in https://mavlink.io/en/messages/common.html#EXTENDED_SYS_STATE so that any offboard control is aware that there was a quadchute.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1933,7 +1933,8 @@
       </entry>
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
-        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used. To force an immediate transition to MC (QuadChute), use MAV_VTOL_STATE_QC.</param>
+        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
+        <param index="2" label="mc_immediate">A transition to MAV_VTOL_STATE_MC will be immediate. Can be used, for example, to engage rotors as an emergency 'parachute'.</param>
       </entry>
       <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST" hasLocation="false" isDestination="false">
         <description>Request authorization to arm the vehicle to a external entity, the arm authorizer is responsible to request all data that is needs from the vehicle before authorize or deny the request. If approved the progress of command_ack message should be set with period of time that this authorization is valid in seconds or in case it was denied it should be set with one of the reasons in ARM_AUTH_DENIED_REASON.
@@ -3019,9 +3020,6 @@
       </entry>
       <entry value="4" name="MAV_VTOL_STATE_FW">
         <description>VTOL is in fixed-wing state</description>
-      </entry>
-      <entry value="5" name="MAV_VTOL_STATE_QC">
-        <description>VTOL is in QuadChute state</description>
       </entry>
     </enum>
     <enum name="MAV_LANDED_STATE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1934,7 +1934,7 @@
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
         <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
-        <param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopilot's implementation of this command.</param>
+        <param index="2" label="Immediate">Force immediate transition to the specified MAV_VTOL_STATE. 1: Force immediate, 0: normal transition. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous/damage vehicle, depending on autopilot implementation of this command.</param>
       </entry>
       <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST" hasLocation="false" isDestination="false">
         <description>Request authorization to arm the vehicle to a external entity, the arm authorizer is responsible to request all data that is needs from the vehicle before authorize or deny the request. If approved the progress of command_ack message should be set with period of time that this authorization is valid in seconds or in case it was denied it should be set with one of the reasons in ARM_AUTH_DENIED_REASON.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1934,7 +1934,7 @@
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
         <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
-        <param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopiliot's implementation of this command.</param>
+        <param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopilot's implementation of this command.</param>
       </entry>
       <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST" hasLocation="false" isDestination="false">
         <description>Request authorization to arm the vehicle to a external entity, the arm authorizer is responsible to request all data that is needs from the vehicle before authorize or deny the request. If approved the progress of command_ack message should be set with period of time that this authorization is valid in seconds or in case it was denied it should be set with one of the reasons in ARM_AUTH_DENIED_REASON.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3020,7 +3020,7 @@
       <entry value="4" name="MAV_VTOL_STATE_FW">
         <description>VTOL is in fixed-wing state</description>
       </entry>
-      <entry value="4" name="MAV_VTOL_STATE_QC">
+      <entry value="5" name="MAV_VTOL_STATE_QC">
         <description>VTOL is in QuadChute state</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1934,7 +1934,7 @@
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
         <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
-        <param index="2" label="mc_immediate">A transition to MAV_VTOL_STATE_MC will be immediate. Can be used, for example, to engage rotors as an emergency 'parachute'.</param>
+        <param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopiliot's implementation of this command.</param>
       </entry>
       <entry value="3001" name="MAV_CMD_ARM_AUTHORIZATION_REQUEST" hasLocation="false" isDestination="false">
         <description>Request authorization to arm the vehicle to a external entity, the arm authorizer is responsible to request all data that is needs from the vehicle before authorize or deny the request. If approved the progress of command_ack message should be set with period of time that this authorization is valid in seconds or in case it was denied it should be set with one of the reasons in ARM_AUTH_DENIED_REASON.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1933,7 +1933,7 @@
       </entry>
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes @auturgy I don't know enough to evaluate this. Can you guys help?

Note, on PX4 at least you can deploy a [parachute during flight termination](http://docs.px4.io/master/en/peripherals/parachute.html#parachute) - and that can be triggered using [MAV_CMD_DO_FLIGHTTERMINATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_FLIGHTTERMINATION) right now.

julianoes:
No idea, @sfuhrer?

sfuhrer (approved):
makes sense to me, looks clean.

ThomasRigi:
@hamishwillee My hopes were to not have to use a parachute (which is a big incident) if we can quadchute before (a minor incident with hopefully no damages). We do have a parachute on board, but it's really the last resort if all else goes bad.

hamishwillee:
Thanks. OK, so it makes sense why you want this, and it can't break anything "existing" out there. 

Is it really valid to think of as a "state"? i.e. the other states are MC or FW and you can switch between those more or les at will. Quadchute means nothing to me but my first thought would have been that this is more like turning on a peripheral. If that is not the case I'm OK with this.

@auturgy Any objections/thoughts on this new VTOL state?

ThomasRigi:
@hamishwillee you're right that currently I only use it as a trigger, so it's no full state yet. A Quadchute basically means an immediate hard switch from FW -> MC that you use as a failsafe action if you encounter problems in FW flight. In PX4 currently when the drone quadchutes it raises a flag in `vtol_vehicle_status` (called `vtol_transition_failsafe`), which then also gets reported to commander (in `vehicle_status_flags`, called `vtol_transition_failure`). This failsafe state could with the corresponding code changes be reflected in this new MAV_VTOL_STATE_QC state, at which point I think we could count it as a real state. IMO it would be helpful to expose this failsafe state over MAVLINK to peripherals such as the offboard computer. For now we listen for STATUSTEXT messages containing "Abort: QuadChute ..." warnings, but having it reflected in ExtendedSysState would make the procedure smoother.

If we don't want to treat Quadchute as a state, an alternative would be to use a second parameter for https://mavlink.io/en/messages/common.html#MAV_CMD_DO_VTOL_TRANSITION, something like "force" or "Quadchute" and if set to 1 and used in combination with Param1 set to MAV_VTOL_STATE_MC it will trigger a Quadchute in PX4. 

(I'm always only writing about PX4 because I have never used ArduPilot)

hamishwillee:
Hi @ThomasRigi,
I'm only a facilitator, but to me this feels a little like a misuse of the message. Others may see otherwise. The easiest way to get clear direction on this is to attend the MAVLink dev call next week. Do you think you might be able to come? https://mavlink.io/en/about/support.html#schedule ?

Otherwise I can ask on your behalf.

ThomasRigi:
I have added the call to my calendar, I hope I'll be able to attend :+1:

hamishwillee:
Discussed in MAV call. 

Quadchute is essentially using the MC rotors as a parachute when needed. Our understanding is that it doesn't require any different handling than MC mode - so you're not actually planning on defining a new state with different behaviour than MC. 
Upshot, you don't need a new state unless you have new state-like behaviour. 

We think you can use the current message as it is defined to change to MC when you need quadchute behaviour. However we "guess" there might be conditions under which an autopilot might reject the call the switch states.  So one option is to add a force bit in a spare/unused parameter. Ie a value of 0 means "as it is now", a value of 1 means "obey this command immediately". 

@sfuhrer Make sense? Are there conditions in PX4 where this force bit might be needed, or does transition always happen?

ThomasRigi:
Thanks for bringing it up and sorry for not being able to attend the call. 

The current message definition doesn't allow to distinguish a regular back transition (which for our drones can take up to 100-150m and up to 8s) from a quadchute (an immediate switch to MC as an emergency reaction, where we decelerate with much greater force). It's not that the autopilot would reject a switch to MC (I don't know if this could happen), it's that it normally takes too much time if it's an emergency. You don't wait 8s before triggering a parachute either.

> So one option is to add a force bit in a spare/unused parameter. Ie a value of 0 means "as it is now", a value of 1 means "obey this command immediately".

Yes, this force bit is an alternative that works just as well as adding a new "state". I hope that in the next days I'll find some time to create a new PR for this and adapt and test the PX4 PR accordingly. 

There is one question though, should this force bit also work for transitions to FW? An immediate state switch is certainly no good idea as the plane would stall immediately. But you could enforce that the normal front transition is triggered immediately no matter what.

sfuhrer:
Given that `MAV_VTOL_STATE` has fields for `MAV_VTOL_STATE_TRANSITION_TO_FW` and `MAV_VTOL_STATE_TRANSITION_TO_MC`, isn't the current way of doing it (giving either `MAV_VTOL_STATE_MC` or `MAV_VTOL_STATE_FW` as arguments in `MAV_CMD_DO_VTOL_TRANSITION`) actually the wrong way to do it? The clean way would be to have `MAV_VTOL_STATE_TRANSITION_TO_FW` or `MAV_VTOL_STATE_TRANSITION_TO_FW` as arguments. And then add `MAV_VTOL_STATE_FW` and `MAV_VTOL_STATE_MC` as arguments to completely bypass the transition logic and directly switch into those modes. 
On the MAVLINK side, only the comment "Only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used" would need to be updated.
Thoughts?

ThomasRigi:
@sfuhrer I like your thinking! This really looks much cleaner to me. However, this change would lead to dangerous behaviour if suddenly the old version of using this command becomes the immediate switch. We'd need some way to properly sync the external mavlink devices (GCS, Offboard computers etc) with the the autopilot's firmware version.

sfuhrer:
> However, this change would lead to dangerous behaviour if suddenly the old version of using this command becomes the immediate switch

Agree. On the PX4 side we could introduce a new param, like VT_ALLOW_QC_CMD, that by default is to false, thus will always trigger a transition if any VTOL_TRANSITION command is received. At some point in the future we could then drop this again.

sfuhrer:
> On the PX4 side we could introduce a new param, like VT_ALLOW_QC_CMD, that by default is to false, thus will always trigger a transition if any VTOL_TRANSITION command is received

That will though of course only cover the "old QGC, new flight controller" case, but not the opposite, where you then would send VTOL_Transition commands with `MAV_VTOL_STATE_TRANSITION_TO_FW`/`MAV_VTOL_STATE_TRANSITION_TO_MC, and the old flight controller then doesn't know what to do with that, expecting `MAV_VTOL_STATE_MC`/`MAV_VTOL_STATE_FW`. 
So in the end, just adding a "do transition immediate" field in the `MAV_CMD_DO_VTOL_TRANSITION` may be easier still in the end. 
@hamishwillee how are these incapability issues between groundstation and flight controller handled in other cases? If there is a fairly easy way to solve that problem then going for the 4 argument approach in MAV_CMD_DO_VTOL_TRANSITION looks neater, if not, let's add the additional "do transition immediate"  param.

hamishwillee:
@sfuhrer @ThomasRigi 

I wish we didn't use MAV_VTOL_STATE  for both state and "command" - it is awkward, and changing the MAV_CMD_DO_VTOL_TRANSITION to use different states for different things doesn't really fix that. Also the break would be significant.

> how are these incapability issues between groundstation and flight controller handled in other cases? If there is a fairly easy way to solve that problem then going for the 4 argument approach in MAV_CMD_DO_VTOL_TRANSITION looks neater, if not, let's add the additional "do transition immediate" param.

It is very hard to break anything in MAVLink. Often we live with crap situations because they aren't worth the cost of a fix. If we can we extend existing commands/messages in a non-breaking way. If we have to we'd typically create a new message. 

As already discussed the three options are:
- a new command MAV_CMD_VTOL_ENABLE_ROTORCHUTE
- use a "mc_immediate" bit
    ```
      <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
        <description>Request VTOL transition</description>
        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
        <param index="1" label="mc_immediate">A transition to MAV_VTOL_STATE_MC will be immediate. Can be used, for example, to engage rotors as an emergency 'parachute'.</param>
      </entry>
  ```
- a new state for quadchute

I lean towards the mc_immediate bit as suggested. It is the minimum change needed to achieve the desired behaviour and it won't break anything.

ThomasRigi:
> Often we live with crap situations because they aren't worth the cost of a fix

I think this is one of these cases where a clean fix which breaks things is too much a hassle if we can get the desired behaviour by just adding one line to use a previously unused parameter. @hamishwillee I pushed your "mc_immediate" bit solution. I agree it's the easiest way to get the desired behaviour.

message_definitions/v1.0/common.xml
```diff
@@ -1933,7 +1933,8 @@
       </entry>
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
-        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. Only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
+        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
+        <param index="2" label="mc_immediate">A transition to MAV_VTOL_STATE_MC will be immediate. Can be used, for example, to engage rotors as an emergency 'parachute'.</param>
```
sfuhrer, line 1937: Could we maybe add the term "quadchute" as well? At least on the PX4 side that's a common keyword that people may be looking for.

sfuhrer: Also, are we sure that we want to have it asymmetric? I think rather that the mavlink command should be independent of whether it's to MC or FW - if the flight controller doesn't support immediate MC to FW state changes then it should be prevented there - and/or on the groundstation ofc. 
So my proposal would be to have the param named "skip transition logic" or something along that. "immediate transition" could also work.

AndreasAntener:
Adding an additional opinion here: both sofar favoured solutions, 1) a new command, or 2) using a command param on MAV_CMD_DO_VTOL_TRANSITION sound reasonable to me.

As for @sfuhrer symmetry suggestion, yes I have a case where the opposite direction is desireable too. A failing vertical system at high altitude can be saved by forcing it into fixed-wing mode where the transition logic potentially fails.

In addition I was wondering if the naming and description should define if this forces an immediate switch or if it allows for a near immediate transition (the back transition still ramps up motors to not cause a huge power spike for example).

hamishwillee:
FWIW

> As for @sfuhrer symmetry suggestion, yes I have a case where the opposite direction is desirable too. A failing vertical system at high altitude can be saved by forcing it into fixed-wing mode where the transition logic potentially fails.

Silvan, is this something you think we need to cater to? If so, we could change `mc_immediate` into `mc_immediate` and document the cases they are intended for.

> should define if this forces an immediate switch or if it allows for a near immediate transition (the back transition still ramps up motors to not cause a huge power spike for example).

I'd assume that any command will be tuned to not break the vehicle, unless otherwise specified as a likely result (we do this for the flight termination commands, which make the vehicle crash).  So even setting immediate would mean be "as immediate as can be without breaking the vehicle".
That might mean that for some vehicles setting this bit would not be fast enough for quadchute. 

Reasonable @sfuhrer @ThomasRigi ?

ThomasRigi:
I'm good with having it symmetric, I just didn't see the use case for MC->FW immediate switch. Now that there is one, yes certainly let's have it symmetric. I'd vote for naming the field "immediate_transition" or "force_immediate_transition".

To address @hamishwillee's concerns about breaking the vehicle I'm in favor of calling it "force_immediate_transition". If you need to force something, it's an action done on purpose because you're in a situation where the normal behaviour doesn't work anymore. That forcing actions you usually wouldn't do can put your vehicle in danger is IMO clear enough. We can add a warning message to the description, and leave it up to the autopilot's implementation on when it allows you to do the immediate switch and refuse the command in a too dangerous situation:
`<param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopiliot's implementation of this command.</param>`

For example, @AndreasAntener 's use case can work in manual flight modes, but I have troubles imagining how PX4 would handle an immediate switch to FW if the drone is in Hold mode (or any automatic mode for that matter)

AndreasAntener:
@ThomasRigi naming/description sounds good

As for the force fixed-wing transition, I understand you work with PX4, but we shouldn't look too hard at what one implementation can do currently for defining this. Having said that, with a stable aircraft, one that maintains a proper glide position without propultion, this is feasible, also with PX4.

sfuhrer (changes requested):
Alright, sounds like we're aligned. 

@ThomasRigi can you push your proposition?

><param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopiliot's implementation of this command.</param>

ThomasRigi:
Ok done. I'm on holidays next week. Edits by maintainers are allowed in case of need :)

message_definitions/v1.0/common.xml
```diff
@@ -1933,7 +1933,8 @@
       </entry>
       <entry value="3000" name="MAV_CMD_DO_VTOL_TRANSITION" hasLocation="false" isDestination="false">
         <description>Request VTOL transition</description>
-        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. Only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
+        <param index="1" label="State" enum="MAV_VTOL_STATE">The target VTOL state. For normal transitions, only MAV_VTOL_STATE_MC and MAV_VTOL_STATE_FW can be used.</param>
+        <param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopiliot's implementation of this command.</param>
```
sfuhrer, line 1937:
```suggestion
        <param index="2" label="force_immediate_transition">Force an immediate transition to the desired MAV_VTOL_STATE. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous. Check your autopilot's implementation of this command.</param>
```

sfuhrer: @hamishwillee beside the typo this looks good for me.

hamishwillee:
Thanks guys. @sfuhrer I have modified this slightly to 
- specify what value counts as "forced transition" (i.e. "1"). 
- Clarified the main risk is vehicle damage
- The field "label" is an actual label, so force_immediate_transition casing is unhelpful. I've just changed to immediate, and the comment in QGC can help people understand values needed.

> `<param index="2" label="Immediate">Force immediate transition to the specified MAV_VTOL_STATE. 1: Force immediate, 0: normal transition. Can be used, for example, to trigger an emergency "Quadchute". Caution: Can be dangerous/damage vehicle, depending on autopilot implementation of this command.</param>`

I think this is good to go!

PR #1571, opened 2021-02-03
Fix badge and CI without changes

julianoes:
This adds a link to the badge, and let's the script pass if there is no diff in messages.


LorenzMeier (approved):
Thanks!

PR #1573, opened 2021-02-03
Clarify MAV_CMD_NAV_FENCE_RETURN_POINT vs rally points

hamishwillee:
@auturgy As discussed the deprecation of MAV_CMD_NAV_FENCE_RETURN_POINT was removed. I've proposed new wording in the description that captures when this should be used vs rally points - essentially IF rally points are supported they should be trusted instead. 

I think this works for systems that support the point (ArduPlane), systems that don't (ArduCopter, PX4) and things that support both (none that I know of, though possibly ArduPlane will do so in when updated to support rally points).

What this means though is that if a system supports both, then it must use the rally points, if enabled.


hamishwillee:
Self merging. The comment makes sense - newer systems are all using geofence instead of this.

PR #1577, opened 2021-02-08
Add battery faults flags for wrong firmware and battery models

cmic0:
This PR extends `MAV_BATTERY_FAULT` with two new fault types:
- Battery firmware incompatible: when the smart battery has a too old/new firmware compared to the autopilot.
- Battery model incompatible: when the battery model (e.g 2s1p, 2s2p,...) is not compatible with the vehicle

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2929,6 +2929,12 @@
       <entry value="64" name="MAV_BATTERY_FAULT_INCOMPATIBLE_VOLTAGE">
         <description>Vehicle voltage is not compatible with this battery (batteries on same power rail should have similar voltage).</description>
       </entry>
+      <entry value="128" name="MAV_BATTERY_FAULT_INCOMPATIBLE_FIRMWARE">
+        <description>Battery firmware is not compatible with current autopilot firmware.</description>
+      </entry>
+      <entry value="256" name="MAV_BATTERY_FAULT_INCOMPATIBLE_MODEL">
+        <description>Battery model is not compatible.</description>
+      </entry>
     </enum>
     <enum name="MAV_GENERATOR_STATUS_FLAG" bitmask="true">
       <description>Flags to report status/failure cases for a power generator (used in GENERATOR_STATUS). Note that FAULTS are conditions that cause the generator to fail. Warnings are conditions that require attention before the next use (they indicate the system is not operating properly).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2932,8 +2932,8 @@
       <entry value="128" name="MAV_BATTERY_FAULT_INCOMPATIBLE_FIRMWARE">
         <description>Battery firmware is not compatible with current autopilot firmware.</description>
       </entry>
-      <entry value="256" name="MAV_BATTERY_FAULT_INCOMPATIBLE_MODEL">
-        <description>Battery model is not compatible.</description>
+      <entry value="256" name="BATTERY_FAULT_INCOMPATIBLE_CELLS_CONFIGURATION">
+        <description>Battery is not compatible due to cell configuration (e.g. 5s1p when vehicle requires 6s).</description>
       </entry>
     </enum>
     <enum name="MAV_GENERATOR_STATUS_FLAG" bitmask="true">


message_definitions/v1.0/common.xml
```diff
@@ -2929,6 +2929,12 @@
       <entry value="64" name="MAV_BATTERY_FAULT_INCOMPATIBLE_VOLTAGE">
         <description>Vehicle voltage is not compatible with this battery (batteries on same power rail should have similar voltage).</description>
       </entry>
+      <entry value="128" name="MAV_BATTERY_FAULT_INCOMPATIBLE_FIRMWARE">
```
hamishwillee, line 2932: What does this really mean/how can this happen? The battery is _outputting_ mavlink battery messages  which are by nature "compatible" (or at least the Autopilot knows what is incompatible). This therefore implies that the battery is getting something from the autopilot firmware that makes it incompatible - but nothing is coming from the autopilot to the battery.

message_definitions/v1.0/common.xml
```diff
@@ -2929,6 +2929,12 @@
...
+        <description>Battery firmware is not compatible with current autopilot firmware.</description>
+      </entry>
+      <entry value="256" name="MAV_BATTERY_FAULT_INCOMPATIBLE_MODEL">
+        <description>Battery model is not compatible.</description>
```
hamishwillee, line 2936: I guess it would be convenient to push this to the battery, but what would make a battery incompatible, and how would the battery know?

hamishwillee:
@auturgy Any comment? I have no particular problem with how these might be used - I am assuming that the GCS would warn "you're using an incompatible battery". I just don't know how the battery could a) be incompatible on a software level, b) know that it is incompatible on a hardware level.

message_definitions/v1.0/common.xml
cmic0: Most of the smart batteries I've seen so far are not directly talking MAVLink but they use SMBus or CAN (or in general  BUS oriented protocols) for the communication towards the autopilot. 
On the autopilot part there is then a driver which handles the abstraction towards MAVLink.

In this view it might happen that the battery has an incompatible firmware compared to the autopilot driver.

message_definitions/v1.0/common.xml
cmic0: Moving towards smart battery standards (e.g Pixhawk DS-013) there will be more batteries with the same connector; in this situation it might well be that the user plugs a non supported battery model into the vehicle (eg a 5s1p instead of a 6s). 

Following the view of the comment above, the driver would recognize the bad model and flag it to the user with a MAVLink message

message_definitions/v1.0/common.xml
hamishwillee: OK, makes sense. I'll confirm at MAV meeting no one has objections.

message_definitions/v1.0/common.xml
hamishwillee: @cmic0 is "model" the right term (I don't know) - we could do `MAV_BATTERY_FAULT_INCOMPATIBLE` and this could be used for any non-firmware incompatibility - chemistry etc. Up to you.

But, I you should give an example in the description - something like

>Battery model is not compatible **(e.g. 5s1p when vehicle requires at least 6s)**.

hamishwillee:
@cmic0 Discussed in dev call. Those present had no objections. 

I added a comment/question in https://github.com/mavlink/mavlink/pull/1577#discussion_r578075803

We can merge once that is addressed. If I'm not around ask Julian.

hamishwillee:
@cmic0 No comment on ^^^ ?

message_definitions/v1.0/common.xml
cmic0: Thanks @hamishwillee for following up!
Probably a more appropriate term would be `BATTERY_FAULT_INCOMPATIBLE_CELL_CONFIGURATION` , as the 5s1p and so on defines it's configuration. Thoughts? 

> we could do MAV_BATTERY_FAULT_INCOMPATIBLE and this could be used for any non-firmware incompatibility - chemistry etc. Up to you

I'd not "merge" the faults into a single flag as in general the user would be required to do different operations (e.g update firmware in one case and just swap the battery in the other case) thus having such level of granularity is beneficial

hamishwillee:
@cmic0 I updated to BATTERY_FAULT_INCOMPATIBLE_CELLS_CONFIGURATION as you suggested. I'm OK with this and I suspect you're the first to add this. 

@auturgy any objection. I plan to merge tomorrow if no one comments. This has been sitting here for comments for a good long while, this change makes sense, and it can't break anyone else.

julianoes (approved):
Seems fine to me.

cmic0:
Thanks @hamishwillee! Looks really good now IMO 👌🏻

hamishwillee:
Awesome. It's in :-)

PR #1578, opened 2021-02-09
common: add gimbal manager attitude status message

julianoes:
This adds an additional message which basically mirrors the GIMBAL_DEVICE_ATTITUDE_STATUS message.

The initial idea was that the gimbal device broadcasts its attitude status to everyone in the mavlink network and thus also a ground station. However, during implementation, we realized that this has some downsides:

1. In many network/messaging setups the attitude status message needs to be forwarded by mavlink components to finally reach the ground station. For instance, if the autopilot is the gimbal manager, it needs to bridge the mavlink connection between the gimbal device and the ground station. This is not always trivial, e.g. can require more RAM usage, more configuration required, or lead to inefficiencies because messages are forwarded from or to the gimbal which are not required.
2. This concept somewhat falls apart if there is no gimbal device because e.g. we are talking to an integrated camera/gimbal which does not actually contain or talk to a gimbal device.
3. This concept is a bit tricky to explain/document because it does not follow the rest of the messages which are symmetric between device and manager.

Basically, this change would move from the diagram on the left to the right:

gimbal attitude feedback

FYI @olliw42

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6299,12 +6299,12 @@
     <message id="289" name="GIMBAL_MANAGER_ATTITUDE_STATUS">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Message reporting the status of a gimbal manager. This message should be broadcasted by a gimbal manager component and will basically mirror GIMBAL_DEVICE_ATTITUDE_STATUS. The angles encoded in the quaternion are in the global frame (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right). This message should be broadcast at a low regular rate (e.g. 10Hz).</description>
+      <description>Message reporting the attitude status by a gimbal manager. This message should be broadcasted by a gimbal manager component and will basically mirror GIMBAL_DEVICE_ATTITUDE_STATUS. The angles encoded in the quaternion are in the global frame (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right).</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">Current gimbal flags set.</field>
-      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation, the frame is depends on whether the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set)</field>
+      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation, the frame depends on whether the flag GIMBAL_MANAGER_FLAGS_YAW_LOCK is set)</field>
       <field type="float" name="angular_velocity_x" units="rad/s">X component of angular velocity (NaN if unknown)</field>
       <field type="float" name="angular_velocity_y" units="rad/s">Y component of angular velocity (NaN if unknown)</field>
       <field type="float" name="angular_velocity_z" units="rad/s">Z component of angular velocity (NaN if unknown)</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6296,6 +6296,20 @@
       <field type="float" name="pitch_rate">Pitch angular rate unitless (-1..1, positive: up, negative: down, NaN to be ignored).</field>
       <field type="float" name="yaw_rate">Yaw angular rate unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
     </message>
+    <message id="289" name="GIMBAL_MANAGER_ATTITUDE_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Message reporting the status of a gimbal manager. This message should be broadcasted by a gimbal manager component and will basically mirror GIMBAL_DEVICE_ATTITUDE_STATUS. The angles encoded in the quaternion are in the global frame (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right). This message should be broadcast at a low regular rate (e.g. 10Hz).</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">Current gimbal flags set.</field>
+      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation, the frame is depends on whether the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set)</field>
+      <field type="float" name="angular_velocity_x" units="rad/s">X component of angular velocity (NaN if unknown)</field>
+      <field type="float" name="angular_velocity_y" units="rad/s">Y component of angular velocity (NaN if unknown)</field>
+      <field type="float" name="angular_velocity_z" units="rad/s">Z component of angular velocity (NaN if unknown)</field>
+      <field type="uint32_t" name="device_failure_flags" display="bitmask" enum="GIMBAL_DEVICE_ERROR_FLAGS">Gimbal device failure flags (0 for no failure)</field>
+    </message>
     <message id="290" name="ESC_INFO">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->


julianoes:
Any thoughts @olliw42? I would, as always, appreciate your opinion :).

hamishwillee:
From a "MAVLink point of view" a broadcast message should _always_ be forwarded. I've never had to consider the potential "additional" cost of internally routing such a thing.

My two bits - I like the symmetry of the API, and I think it makes a lot more sense for a gimbal manager to emit a gimbal attitude status than a device attitude status if it is wrapping the device behaviour. 

@julianoes 
Before I look at the specifics of the docs, can you confirm that .... instead of *broadcasting* `GIMBAL_DEVICE_ATTITUDE_STATUS` it would be addressed specifically to the gimbal manager? The Gimbal manager would get this and then resend as a GIMBAL_MANAGER_ATTITUDE_STATUS, this time broadcast (or would it be addressed?)

julianoes:
> Before I look at the specifics of the docs, can you confirm that .... instead of broadcasting GIMBAL_DEVICE_ATTITUDE_STATUS it would be addressed specifically to the gimbal manager? The Gimbal manager would get this and then resend as a GIMBAL_MANAGER_ATTITUDE_STATUS, this time broadcast (or would it be addressed?)

Yes, that's the idea but I should change the description then. However, it is still allowed to be broadcast, it's just not required or recommended I'd say.

hamishwillee:
> Yes, that's the idea but I should change the description then. However, it is still allowed to be broadcast, it's just not required or recommended I'd say.

Well you can always broadcast, but sounds like routing a broadcast message is a problem. At least if you're explicit it shouldn't be then you will be respecting MAVLink's authori-tay :-)

message_definitions/v1.0/common.xml
```diff
@@ -6296,6 +6296,20 @@
       <field type="float" name="pitch_rate">Pitch angular rate unitless (-1..1, positive: up, negative: down, NaN to be ignored).</field>
       <field type="float" name="yaw_rate">Yaw angular rate unitless (-1..1, positive: to the right, negative: to the left, NaN to be ignored).</field>
     </message>
+    <message id="289" name="GIMBAL_MANAGER_ATTITUDE_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Message reporting the status of a gimbal manager. This message should be broadcasted by a gimbal manager component and will basically mirror GIMBAL_DEVICE_ATTITUDE_STATUS. The angles encoded in the quaternion are in the global frame (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right). This message should be broadcast at a low regular rate (e.g. 10Hz).</description>
```
hamishwillee, line 6302: "will basically mirror GIMBAL_DEVICE_ATTITUDE_STATUS" and "This message should be broadcast at a low regular rate (e.g. 10Hz)" seem to be a little opposed. 

I assume you mean "we get a GIMBAL_DEVICE_ATTITUDE_STATUS and immediately emit a GIMBAL_MANAGER_ATTITUDE_STATUS; this is expected to be at a low regular  rate (e.g. 10Hz). 

But you could mean that the gimbal manager caches the device status messages and emits the last one received a its own rate, independent of whatever the gimbal device sends.

message_definitions/v1.0/common.xml
```diff
@@ -6296,6 +6296,20 @@
...
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">Current gimbal flags set.</field>
+      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation, the frame is depends on whether the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set)</field>
```
hamishwillee, line 6307: the frame is depends > the frame depends OR the frame is dependent on.

message_definitions/v1.0/common.xml
julianoes: It could be either, probably mostly in sync but it could be at a slower rate if set like that by a GCS.

message_definitions/v1.0/common.xml
julianoes: Ops, thx.

julianoes:
I pushed another try.

olliw42:
thx for keeping me in the loop :)

as regards my opinion, I think I understand the idea to have a more clean api, on the other hand it seems to me that this message does not carry any information which is not already available through other messages, and thus is of purely cosmetic type. I think if such a message would be proposed in another context it would be turned down directly with that argument. :)

E.g. one could argue that the camera protocol could be made cleaner and nicer too by not reusing otherwise introduced messages (which if I recall correctly was once explicitly stated is not what mavlink wants to do).

I'm also not comfortable with that it adds latency, since it needs to be parsed and resend by the gimbal manger, which would compromise tighter closed loop control in e.g higher-performance gimbal tracking. (we want to get better with time, right, not stick with 2010)

That's generally a bit of an issue (it e.g. applies to routing (which could be made much faster in terms of latency than it currently is)), and also applies to the gimbal manager control messages. It could be avoided there, but with the given messages (of both your an my set) it is so. However, for the status it is easily avoided, by doing as it currently is been done.

Hence, the cosmetic benefit of a cleaner api is IMHO of less benefit than the disadvantages.

> makes a lot more sense for a gimbal manager to emit a gimbal attitude status than a device attitude status

the gimbal manager is never to emit a device attitude status

julianoes:
> I think if such a message would be proposed in another context it would be turned down directly with that argument. :)

Probably true!

> I'm also not comfortable with that it adds latency, since it needs to be parsed and resend by the gimbal manger

That's also my fear.

> the gimbal manager is never to emit a device attitude status

Right but how would you emit the attitude status if there is no gimbal device, e.g. the gimbal manager is an all-in-one camera/tracking/gimbal/thing?

olliw42:
> Right but how would you emit the attitude status if there is no gimbal device, e.g. the gimbal manager is an all-in-one camera/tracking/gimbal/thing?

I don't interpret it that way. I interpret it to mean that there are both a gimbal manager and a gimbal device in one housing.

That's IMHO the only reasonable interpretation. E.g. parameters. Should a gimbal manger have parameters itself? If so, wouldn't it have to emit a heartbeat with a comp id? What would the comp id be?

In my interpretation I simply have a gimbal device in the housing, i.e. a component with a gimbal comp id, which emits heartbeats, has parameters, maybe mavftp, and whatever a component may do. Since it's a gimbal component, it emits the gimbal device status message. In addition to that I have the gimbal manager in the housing. The gimbal manager however is not a component, by design and construction.

If there is addition a camera in the housing (which the STorM32 gimbal actually does, see NT Camera), when there is also a camera component with a camera comp id, which emits heartbeats, has parameters, maybe mavftp, and whatever a component may do.

Anything else doesn't IMHO make sense with the fabric and design of mavlink. Where to should all the camera communication go to if there would be not a camera component? And it actually makes sense to me and it works also fine, proof: STorM32.

julianoes:
Right, sorry I wasn't clear. In that case this integrated thing would be one mavlink component, probably the choice would be COMPID_CAMERA, it would also send out gimbal manager information, so it would signal to the GCS that it is a gimbal manager.

Now I'm wondering if it could actually send GIMBAL_DEVICE_ATTITUDE_STATUS. I suppose it could but it doesn't seem too intuitive, right?

olliw42:
as I say, I think that's the miss interpretation, that one physical device must and can be also only one component ...

this necessarily must run into all sorts of problems with all the micro services

> Now I'm wondering if it could actually send GIMBAL_DEVICE_ATTITUDE_STATUS. I suppose it could but it doesn't seem too intuitive, right?

I find your point of view counter intuitive (and just to complicate things)

A physical gimbal is a gimbal device. Period and end of story.

so if there is a physical gimbal device around there ought to be a gimbal component with a gimbal comp id.  

Similarly so with a camera. A physical camera is a camera. Period and end of story.

so if there is a physical camera around there ought to be a camera component with a camera comp id.

A gimbal manager is a gimbal manager. Period and end of story.

A gimbal manger is NOT a mavlink component, and has thus no comp id itself.

With every other approach you neccessarily run into huge issues E.g. just the simple matter of discovering if there is a gimbal or a camera in the system will immediately break down. That's so because one heartbeat can carry only one type, thus if you need to announce the presence of two types you need to emit two heartbeats. And so on and on and on. The hole eco system will break down with your thinking.

> probably the choice would be COMPID_CAMERA, 

no way, it can be COMPID_MISSIONPLANNER, or whatever else but not CAMERA and NOT GIMBAL.

> Now I'm wondering if it could actually send GIMBAL_DEVICE_ATTITUDE_STATUS.

it not only could, it must

and it must send a heartbeat with a gimbal comp_id

julianoes:
> A physical gimbal is a gimbal device. Period and end of story.
> so if there is a physical gimbal device around there ought to be a gimbal component with a gimbal comp id.
> Similarly so with a camera. A physical camera is a camera. Period and end of story.
> so if there is a physical camera around there ought to be a camera component with a camera comp id.
> A gimbal manager is a gimbal manager. Period and end of story.
> A gimbal manger is NOT a mavlink component, and has thus no comp id itself.

I agree with everything except that there might not be a mavlink gimbal device. Basically, my suggestion is that if there is some sort of gimbal (could be SBUS/PWM/<whatever protocol) it can be attached to some component (e.g. the autopilot or a camera).
In that case I would (as an implementor) not want to bother of adding the mavlink gimbal device component because I can already do everything with the gimbal manager API that I expose. The only exception to that is currently the gimbal attitude. That's the only part that needs to come from a gimbal device (so the actual mavlink component).

So, I'm arguing that it is allowed to abstract a gimbal device away while you seem to be suggesting that no matter the setup, that mavlink gimbal device should exits as a mavlink component, right?

I think that's possible but a bit of a problem to implement. For instance for PX4 I'd have to support (at least) these two cases:

1. PX4 is the gimbal manager and has a gimbal device connected via mavlink (the nice case). In this case, the gimbal device message is forwarded by the mavlink network to the ground station. This means that forwarding needs to be enabled for this case which is a headache in itself (and part of the reason of this PR).
2. PX4  is the gimbal manager and is driving an old PWM gimbal. In this case PX4 has to create a second mavlink instance with the component ID of a gimbal in order to be a gimbal device. This means having to create a second mavlink channel and allocate the buffer for it, etc. It's do-able but it's again a bit of a pain. With this PR, it would only implement the gimbal manager attitude message and already be done.

> With every other approach you neccessarily run into huge issues E.g. just the simple matter of discovering if there is a gimbal or a camera in the system will immediately break down. That's so because one heartbeat can carry only one type, thus if you need to announce the presence of two types you need to emit two heartbeats. And so on and on and on. The hole eco system will break down with your thinking.

Discovery is a good point. So the way we have it implemented is for a gimbal manager to discover a gimbal device by the heartbeat. The ground station can then detect the gimbal manager by listening for GIMBAL_MANAGER_STATUS. If that arrives, it can ask for GIMBAL_MANAGER_INFORMATION if it hasn't received it already earlier. This works because the status is broadcast at some rate.

olliw42:
I agree that the case of an old pwm gimbal is less clean as it could be otherwise. HOWEVER, besides this one single point, there are LOTs of HOWEVERs (I'm not sure I list all)

- you can't fully abstract a gimbal device/component into a gimbal manager anyhow: I repeat (and I hope you listen a bit more): parameters. A mavlink gimbal (the nice case) may and most likely will have parameters, and users might actually be interested and request to be able to access them. So, unless you also abstract these into your gimbal manager, whch really should be a coding fest, there is no way around doing the mavlink message forwarding and routing. Similar so if the gimbal component supports mavFtp, or as a matter of fact any other mavlink thing. Or if there is also the camera. So, my clear advice, get your routing right.
- you are flip flopping. Quite some comments and suggestions were turned down with the argument that the new protocol should not be for old fashioned things. I think I could provide many cites along this line (some with sad results for me, such as a storm32.xml, but that's ok LOL). Here now you however are saying that the whole eco system should be substantially broken just for a somewhat easier support of old crappy pwm gimbals .... 
- all of the other points I have mentioned before
- I agree that a pwm gimbal or multiple components needs a bit more coding effort, but it is much less than you seem to claim. Much code can be reused.

> In this case PX4 has to create a second mavlink instance with the component ID of a gimbal in order to be a gimbal device. This means having to create a second mavlink channel and allocate the buffer for it, etc. It's do-able but it's again a bit of a pain. With this PR, it would only implement the gimbal manager attitude message and already be done.

I don't know PX4, but I know that I didn't had to do that and that I do not understand why it should be so more difficult to send gimbal_device_status + heartbeat instead of gimbal_manager_status ... especially I don't see why it needs a new channel since transported all on one and the same channel ... 

frankly, given this and given that routing/forwarding seems to be also a problem it seems to me that PX4 rather needs a redesign and not mavlink

> Discovery is a good point. So the way we have it implemented is for a gimbal manager to discover a gimbal device by the heartbeat. The ground station can then detect the gimbal manager by listening for GIMBAL_MANAGER_STATUS. If that arrives, it can ask for GIMBAL_MANAGER_INFORMATION if it hasn't received it already earlier. This works because the status is broadcast at some rate.

you seemed to have missed the crucial point with camera. How would the camera be discovered if there wouldn't be also a camera component with it's own id and heartbeat?

julianoes:
> you can't fully abstract a gimbal device/component into a gimbal manager anyhow: I repeat (and I hope you listen a bit more): parameters. A mavlink gimbal (the nice case) may and most likely will have parameters, and users might actually be interested and request to be able to access them. So, unless you also abstract these into your gimbal manager, whch really should be a coding fest, there is no way around doing the mavlink message forwarding and routing. Similar so if the gimbal component supports mavFtp, or as a matter of fact any other mavlink thing. Or if there is also the camera. So, my clear advice, get your routing right.

Of course you can. The device which implements the gimbal manager can have as many parameters and mavlink things as it wants. E.g. if the autopilot is the gimbal manager it already has params, and for a camera it can have the camera definition with params (or the component definition with params), and these param can of course also cover gimbal functions.

But, and I think that's what you mean. If you have a new mavlink v2 gimbal device it should actually be properly connected to the network because otherwise we really can't make use of all these possibilities! You have a point here, and I clearly need to rethink that.

- you are flip flopping. Quite some comments and suggestions were turned down with the argument that the new protocol should not be for old fashioned things. I think I could provide many cites along this line (some with sad results for me, such as a storm32.xml, but that's ok LOL). Here now you however are saying that the whole eco system should be substantially broken just for a somewhat easier support of old crappy pwm gimbals ....

There is a distinction there. I said that the protocol should not re-use/depend on details of a previous/old design. E.g. a PWM value should not be used in a new API, instead it should be abstracted with something "better", so -1..1 instead of 1000..2000. That's what you do in software architecture, you create abstractions over details that you don't want to carry.
And, yes I definitely don't want to break the whole eco system, and you are right that for nice new things forwarding is indeed required!

> I don't know PX4, but I know that I didn't had to do that and that I do not understand why it should be so more difficult to send gimbal_device_status + heartbeat instead of gimbal_manager_status ... especially I don't see why it needs a new channel since transported all on one and the same channel ...

It needs a new MAVLINK_CHANNEL_x, otherwise the sequence number is incorrect and it looks like message were dropped. That's because there is one sequence per MAVLINK_CHANNEL. Maybe we can get away with one receive buffer, I have not thought that one through.

> frankly, given this and given that routing/forwarding seems to be also a problem it seems to me that PX4 rather needs a redesign and not mavlink

Yes, that might indeed be required.

> you seemed to have missed the crucial point with camera. How would the camera be discovered if there wouldn't be also a camera component with it's own id and heartbeat? Why would your points 1. and 2. not give you headaches with PX4?

Ok, I'm assuming here that a camera is directly connected to the mavlink network while a gimbal is wired through the autopilot. That's because a camera usually comes with a live stream which requires an IP (or something similar) anyway, and some sort of Linux onboard computer (in or next to the camera). For that case we can use mavlink-router and connect it all, no problem.

Summary: I realize what you're saying. The forwarding of this one message solves the problem for one case but just creates confusion for the case where the gimbal device actually wants to do more than report its attitude! Given that it might be a good idea to stick with forwarding.

olliw42:
> Of course you can. Of course you can. The device which implements the gimbal manager can have as many parameters and mavlink things as it wants. E.g. if the autopilot is the gimbal manager it already has params, and for a camera it can have the camera definition with params (or the component definition with params), and these param can of course also cover gimbal functions.

First off, I was clearly discussing "A mavlink gimbal (the nice case)", so that's a bit of distraction. However, even when talking about a legacy gimbal, it is logically not sound. Of course you can add whatever parameters to the autopilot, but these are then autopilot parameters accessed with the autopilot comp id. If there is no gimbal heartbeat and no means to access these parameters via param messages with a gimbal comp id, no one would be able (withing the mavlink fabric) to realize that these are gimbal parameters! That's not abstraction, that's introducing explicit implementation specific dependencies.

Please note also that there had been quite a shift of topic, since initially it actually started out with "Right but how would you emit the attitude status if ... e.g. the gimbal manager is an all-in-one camera/tracking/gimbal/thing?" while when it moved to case 1. and 2. My position of how to do it fortunately addresses properly each of these cases. The shifts however make it a bit difficult to put statements into the correct context.

> There is a distinction there. 

There is a reason why I felt forced to go with a storm32.xml, I didn't do that out of pleasure but because relevant use cases and gimbal limitations were considered to not be adressed in a new API.

> Summary: I realize what you're saying. The forwarding of this one message solves the problem for one case but just creates confusion for the case where the gimbal device actually wants to do more than report its attitude!

I guess that's my concern concerning the gimbal manager status message.

I'm in addition saying that a gimbal device should be a component with a gimbal id and a camera should be a component with a camera id, independent of whether it are separate physical devices connected to the mavlink network or integrated as code into some other physical device. In the case of a camera/gimbal/thing the thing should behave like two components on one physical link.

julianoes:
> I'm in addition saying that a gimbal device should be a component with a gimbal id and a camera should be a component with a camera id, independent of whether it are separate physical devices connected to the mavlink network or integrated as code into some other physical device. In the case of a camera/gimbal/thing the thing should behave like two components on one physical link.

Right, but that's not required for a gimbal manager, agreed?

olliw42:
my statement (which spurred the discussion :D) was even stronger:

"the gimbal manager is never to emit a device attitude status"

"A gimbal manger is NOT a mavlink component, and has thus no comp id itself."

If your "not required" is a polite rephrasing for "must not", then, yes, agreed :)

julianoes:
@olliw42 right!

PR #1579, opened 2021-02-10
common: make tracking mode enum consistent

julianoes:
The entries of CAMERA_TRACKING_MODE were missing the _MODE suffix which is not consistent and might to mix up with CAMERA_TRACKING_STATUS_FLAGS and CAMERA_TRACKING_TARGET_DATA.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3377,13 +3377,13 @@
     </enum>
     <enum name="CAMERA_TRACKING_MODE">
       <description>Camera tracking modes</description>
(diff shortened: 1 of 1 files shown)


LorenzMeier (approved):

hamishwillee (approved):
This makes perfect sense to me - anything that breaks will do so at build time with a clear warning so is OK by me. 

@auturgy you guys don't use this right?

@julianoes I am pretty sure only PX4 is impacted. So if there is no reply from James by tomorrow, feel free to self merge.

PR #1581, opened 2021-02-11
MAV_CMD_PREFLIGHT_STORAGE: add Reset sensors to factory calibration

bkueng:


message_definitions/v1.0/common.xml
```diff
@@ -1559,7 +1559,7 @@
       </entry>
       <entry value="245" name="MAV_CMD_PREFLIGHT_STORAGE" hasLocation="false" isDestination="false">
         <description>Request storage of different parameter values and logs. This command will be only accepted if in pre-flight mode.</description>
-        <param index="1" label="Parameter Storage" minValue="0" maxValue="2" increment="1">Parameter storage: 0: READ FROM FLASH/EEPROM, 1: WRITE CURRENT TO FLASH/EEPROM, 2: Reset to defaults</param>
+        <param index="1" label="Parameter Storage" minValue="0" maxValue="3" increment="1">Parameter storage: 0: READ FROM FLASH/EEPROM, 1: WRITE CURRENT TO FLASH/EEPROM, 2: Reset to defaults, 3: Reset sensors to factory calibration (or default if not available)</param>
```
hamishwillee, line 1562: I have no problem with this, but are we perhaps now at the point where it makes sense to specify an enum for this?

bkueng: Done

message_definitions/v1.0/common.xml
```diff
@@ -4284,6 +4284,24 @@
       <entry value="6" name="MAG_CAL_BAD_ORIENTATION"/>
       <entry value="7" name="MAG_CAL_BAD_RADIUS"/>
     </enum>
+    <enum name="MAV_STORAGE_ACTION">
+      <description>Storage action.</description>
+      <entry value="-1" name="MAV_STORAGE_ACTION_NOTHING">
```
hamishwillee, line 4289: That -1 in an enum feels very odd :-). All good though.

message_definitions/v1.0/common.xml
```diff
@@ -4284,6 +4284,24 @@
...
+        <description>Read from FLASH/EEPROM.</description>
+      </entry>
+      <entry value="1" name="MAV_STORAGE_ACTION_WRITE">
+        <description>Write current values to FLASH/EEPROM</description>
+      </entry>
+      <entry value="2" name="MAV_STORAGE_ACTION_RESET">
+        <description>Reset to defaults</description>
+      </entry>
+      <entry value="3" name="MAV_STORAGE_ACTION_RESET_SENSOR_FACTORY">
+        <description>Reset sensors to factory calibration (or default if not available)</description>
```
hamishwillee, line 4302: What is the last option for? I.e. I understand that `MAV_STORAGE_ACTION_RESET` resets your values of parameters or missions - presumably to defaults defined in Firmware?

But what are the sensors in `MAV_STORAGE_ACTION_RESET_SENSOR_FACTORY` that you are resetting?

bkueng: The calibration data for accel, gyro, mag. Maybe needs a better description?

message_definitions/v1.0/common.xml
bkueng: Yes to me as well, and apparently CI does not like it either:
```
ERROR: /home/runner/work/mavlink/mavlink/message_definitions/v1.0/common.xml:4289:0:ERROR:SCHEMASV:SCHEMAV_CVC_PATTERN_VALID: Element 'entry', attribute 'value': [facet 'pattern'] The value '-1' is not accepted by the pattern '0[bB][0-1]{1,32}'.
```

bkueng: @hamishwillee how do you want to proceed here?

message_definitions/v1.0/common.xml
hamishwillee: Yes.

hamishwillee: Or to put it another way, this says "reset sensors" which is not obviously anything to do with storage.

message_definitions/v1.0/common.xml
hamishwillee: @bkueng Apologies, I somehow missed this notification. 

Can you push this to  a positive number? That will be much faster probably than enabling mavlink to support negative enum values.

message_definitions/v1.0/common.xml
bkueng: `Reset sensor calibration to factory default (or firmware default if not available)`?

message_definitions/v1.0/common.xml
bkueng: Not really, this is already in use and would break existing setups.

message_definitions/v1.0/common.xml
hamishwillee: How about (?)
- MAV_STORAGE_ACTION_RESET_SENSOR_FACTORY > MAV_STORAGE_ACTION_RESET_SENSOR_CALIBRATION (or maybe MAV_STORAGE_ACTION_RESET_SENSOR_CALIB_DATA
- Description to `Reset sensor calibration parameter data to factory default (or firmware default if not available)`

I'm assuming the calibration data is in the form of parameters, so you would only use this for MAV_CMD_PREFLIGHT_STORAGE.param1. Would you expect MAV_STORAGE_ACTION_RESET
 on parameters to also clear this data?

message_definitions/v1.0/common.xml
hamishwillee: Hmmmm well "someone" would need to make the changes needed in ArduPilot/Pymavlink and "get them through". This would first involve changing the XSD to allow the value to be negative: https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L20-L27

Can you do that?

Then you'd have to do some testing on the generators handling negative values.

@peterbarker This is a case where we have a parameter that takes a few numbers, which we want to turn into an enum. Unfortunately one of the existing numbers is negative, and the XSD prevents negative values on the enum. IMO it should not. Do you know of anything else that might be impacted?

bkueng: I can do it, but I can imagine some languages don't support negative values for enum's.

message_definitions/v1.0/common.xml
bkueng: Works for me. No, `MAV_STORAGE_ACTION_RESET` would not reset that. I can imagine that factory calibration might be read-only on some hardware in the future.

message_definitions/v1.0/common.xml
LorenzMeier: No, we should not go negative in an enum. All hell could break loose years from now.

hamishwillee: Then I don't see any other option than to break existing setups (do you?). E.g. move `MAV_STORAGE_ACTION_NOTHING` to an unused value.

message_definitions/v1.0/common.xml
hamishwillee: We could make this change, but it feels like this  storage really a _peer_ of parameter and mission storage. So we might have
```
<param index="4" label="Sensor Calibration" enum="MAV_STORAGE_ACTION">Sensor calibration data. Set as  action MAV_STORAGE_ACTION_RESET to reset sensor calibration parameter data to factory default (or firmware default if not available)</param>
```

message_definitions/v1.0/common.xml
bkueng: If you really want an enum, then yes, but at this point I'd rather just keep the not-so-nice current state.

message_definitions/v1.0/common.xml
bkueng: Hmm not sure if that's better, at least at the moment they are a subset of the parameter storage (and thus operations to that affect the sensors as well).

hamishwillee:
@bkueng I guess it is better to revert this to just numbers then. Sorry for the back and forth. If we ever revisit this for a new or equivalent message though, then we'll fixt it.

bkueng:
> @bkueng I guess it is better to revert this to just numbers then. Sorry for the back and forth. If we ever revisit this for a new or equivalent message though, then we'll fixt it.

Done

hamishwillee:
Thanks @bkueng . Merged.

PR #1583, opened 2021-02-13
message_definitions: Add auterion dialect

TSC21:
1. Adds the Auterion dialect through the `auterion.xml` file;
2. Updates references for appropriate automatic deployments for the header repos.

I require admin permissions in the following repos:

1. This same repo, so to configure the `GH_TOKEN` with the `auterionci` one;
2. `auterion/c_library_v2` so to add writing permissions to the `auterionci` bot.

Follow up PRs will include some commands and enums required for some specific internal projects and also to validate the implementations so these messages can be upstreamed to the `common` dialect instead.


TSC21:
Sorry, wrong base branch.

PR #1587, opened 2021-02-16
Added Android example

mareksuma1985:
Added Android (API 28) example based on [Linux example](https://github.com/mavlink/mavlink/tree/master/examples/linux) by [Bryan Godbolt](https://github.com/godbolt). It sends heartbeat, attitude, heading, and GPS coordinates. Receives and displays manual control commands.


hamishwillee:
@mareksuma1985 Great idea - thanks very much!

@julianoes @JonasVautherin Suggestions on how we test/review this?

examples/android/mavlink-udp/src/main/cpp/receive.c
```diff
@@ -0,0 +1,94 @@
+        //FIXME: przenieść odbieranie wiadomości do osobnego wątku, bez oczekiwania 1 sekundy
```
amilcarlucas, line 1: Comments in non-english Language!!

amilcarlucas (changes requested):

hamishwillee:
@amilcarlucas @julianoes @JonasVautherin I think having an Android example that can easily be used is a good think, though I worry it will break over time.

But we can't shouldn't merge this untested. Will one of you have time to try it out?

julianoes:
@hamishwillee are there instructions on how to try it? And to make sure it doesn't break later it needs a test in CI.

mareksuma1985:
The [project won't build](https://github.com/firebase/FirebaseUI-Android/issues/1349) after downloading it from GitHub, Android projects rarely do. I've uploaded a compiled apk file under: [mavlink/examples/android/mavlink-udp/build/outputs/apk/debug/mavlink-udp-debug.apk](https://github.com/mareksuma1985/mavlink/blob/master/examples/android/mavlink-udp/build/outputs/apk/debug/mavlink-udp-debug.apk).

JonasVautherin:
> The project won't build after downloading it from GitHub, Android projects rarely do.

@mareksuma1985: Would you mind elaborating on that? I have been maintaining something like 20 Android projects, and they all just built after a clone with `./gradlew build`. Never had an issue with that.

mareksuma1985:
After downloading zip or cloning the repository, menu items: _Build / Make Project_ and _Build / Make Modules_ are inactive and all run configurations are gone. I've tried some tricks from the thread above but they don't work.

JonasVautherin:
> After downloading zip or cloning the repository, menu items: Build / Make Project and Build / Make Modules are inactive and all run configurations are gone.

I'm assuming that you're talking about Android-Studio here. But I don't think that this prevents a CI from building (and testing) the app using gradle.

Not saying that there must be a CI for it (though I think it's always preferable), but just saying that I am pretty sure it is possible.

examples/android/README.md
```diff
@@ -0,0 +1,33 @@
...
+LOCAL_PATH := $(call my-dir)
+
+include $(CLEAR_VARS)
+LOCAL_CFLAGS += -I ~/mavlink/generated/include
+LOCAL_MODULE    := mavlink_udp
+LOCAL_SRC_FILES := mavlink_udp.c
+include $(BUILD_SHARED_LIBRARY)
+```
+
+According to [MAVLink Step by Step](https://discuss.ardupilot.org/t/mavlink-step-by-step/9629) by [Pedro Albuquerque](https://discuss.ardupilot.org/u/Pedro_Albuquerque) `System ID` should be a number between 2 and 254 and `Component ID` should be 1. Only then you can establish connection with QGroundControl and other ground control applications.
```
hamishwillee, line 27: It sounds like you expect people to set `System ID` and `Component ID` **of the SDK** - is that correct?? If so, you should tell them **where/how** to do this. 

Generally System ids are allocated to new vehicles on the network incrementally up from 1, while ground stations are allocated decrementally down from 255. So on a single vehicle/single ground station network any id for the SDK between 2 and 254 works for the SDK - but the key here is "using a unique id for the SDK". 

Again assuming you are talking about the SDK, the component ID should NOT be 1 - that would imply an autopilot: https://mavlink.io/en/messages/common.html#MAV_COMPONENT
You would generally use 191 - for a companion computer.

mareksuma1985:
I've changed `THIS_COMPONENT` back to 1 (`MAV_COMP_ID_AUTOPILOT1`) because QGroundControl would not connect when it was set to 191. UDP server now starts immediately if device is online. Latest build is here: [mavlink-udp-debug.apk](https://github.com/mareksuma1985/mavlink/blob/master/examples/android/mavlink-udp/build/outputs/apk/debug/mavlink-udp-debug.apk).

hamishwillee:
> I've changed THIS_COMPONENT back to 1 (MAV_COMP_ID_AUTOPILOT1) because QGroundControl would not connect when it was set to 191. UDP server now starts immediately if device is online. Latest build is here: mavlink-udp-debug.apk.

That's a QGC issue - it clearly only wants to talk to a component that it thinks of as an autopilot "by default". Perhaps we can cover this by changing the docs FROM:

> According to [MAVLink Step by Step](https://discuss.ardupilot.org/t/mavlink-step-by-step/9629) by [Pedro Albuquerque](https://discuss.ardupilot.org/u/Pedro_Albuquerque) `System ID` should be a number between 2 and 254 and `Component ID` should be 1. Only then you can establish connection with QGroundControl and other ground control applications.

TO

> The `System ID` should be a number between 1 and 255 and `Component ID` should be one of the values in [MAV_COMPONENT](https://mavlink.io/en/messages/common.html#MAV_COMPONENT). In order to be able to connect to QGroundControl we have set the System ID as X and the component id as 1.

hamishwillee:
@julianoes I think the readme "is" the instructions. IF this is not enough, then it needs to be modified to allow someone to test this. Is it enough in your opinion?

julianoes:
1. I still think it should be built in CI to prevent it from getting broken in the future.
2. I wonder if it should have its own repo, and just be linked from here. In the same way that https://github.com/mavlink/c_uart_interface_example has its own repo.

hamishwillee:
@julianoes 

> I wonder if it should have its own repo, and just be linked from here. In the same way that https://github.com/mavlink/c_uart_interface_example has its own repo.

I kind of feel that all examples should be here or none. My leaning is to agree with you - better to have own repo. But we should confirm we're happy to accept it and under what circumstances before asking for that.

What needs to happen to get this in CI? I.e. is this something you would do?

julianoes:
> I kind of feel that all examples should be here or none.

Yes. If it is here then it's nicely in sync. However, that means that whenever we break the example in a PR we need to fix it as well.

> What needs to happen to get this in CI? I.e. is this something you would do?

No, sorry, once I touch it then it becomes my responsibility, and I'm responsible for way too much CI things already.

hamishwillee:
@julianoes Fair enough. I've added to the dev call. We need to make a clear decision on what happens with this (which might be merging, or holding off until we have someone to do the necessary checks)

mareksuma1985:
I've tried [committing](https://github.com/mavlink/mavlink/pull/1587/commits/5a4a6c0fbb0d30289e7dba1c4941b4842461e6fe) the Android Studio project without gitignore files but it still won't build after downloading or cloning. I can sent the compressed project to someone more experienced in Git as I'm out of ideas how to make it work.

hamishwillee:
Thanks @mareksuma1985 , in particular for attending the meeting.

PR #1592, opened 2021-02-17
Specify enum values explicitely III

olliw42:
somewhat embarrassing but my previous method of using mavgen to find enums which do not have a value field didn't tracked these ones here

hence a 3rd PR along this line :(


amilcarlucas (approved):

hamishwillee:
Thank you!

PR #1593, opened 2021-02-17
Add MAVLINK_INCOMPAT_FLAGS to minimal.xml

olliw42:
the mavgen C language specifies e.g. a define MAVLINK_IFLAG_SIGNED in order to flag in the header's incompat_flags field that a packet is signed

however, that's clearly a protocol thing and not a thing of a particular implementation, and I thus think that it should (must) be specified in an xml accordingly. Since it is so basic I put it into minimal.xml.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -2,6 +2,12 @@
 <mavlink>
   <version>3</version>
   <enums>
+    <enum name="MAVLINK_INCOMPAT_FLAGS" bitmask="true">
+      <description>These flags encode the incompat_flags field in the header of MAVLink v2 messages.</description>
+      <entry value="1" name="MAVLINK_INCOMPAT_FLAGS_SIGNED">
+        <description>The packet is signed (a signature has been appended to the packet).</description>
+      </entry>
+    </enum>
     <enum name="MAV_AUTOPILOT">
       <description>Micro air vehicle / autopilot classes. This identifies the individual model.</description>
       <entry value="0" name="MAV_AUTOPILOT_GENERIC">


hamishwillee:
I don't understand the value of this/who/how it helps. But, things should only go in minimal.xml that are required to set up a minimal system. Signing does not fit into that category.

olliw42:
> I don't understand the value of this/who/how it helps. 

The value is that the variable parts of the protocol is defined via the xml, no confusion can happen and everyone refers to the same names. Isn't this what the xml is all about? I mean, anyone could also define the msg IDs themselves and given them whatever name they like ... the xml is not needed except that it is massively convenient. So, I actually don't understand that question ...

As a matter of fact, I don't understand why one has to define that one value by hand :)

It btw also would define which values are valid.

> But, things should only go in minimal.xml that are required to set up a minimal system. Signing does not fit into that category.

true ...

hamishwillee:
I always though the XML was a programming language agnostic way of representing messages and commands for auto-generation :-). I don't see what anyone will do differently with this information.

But perhaps others will see it differently. @julianoes

olliw42:
> I always though the XML was a programming language agnostic way

exactly !

now tell me, in how far is the incompat flags programming language dependent? 

I find it is not, it is as programming language agnostic as any other enum in these xmls are.

> representing messages and commands

that's one possible interpretation ... another is to have it representing the variable parts in the protocol

the variable part includes the incompat flags as it determines e.g. the length of a message as much as a msg ID does ... 

>  I don't see what anyone will do differently with this information

anyone using v2 messages will explicitly or implicitly have to use that flag !! and be it only through a code generator which provides convenient functions for you

let's hear @julianoes.  I however get that what looked so logical to me doesn't raise frenetic enthusiasm on your side. Again, it is not the most important thing in the world and anyone can live without, so, if you feel it's more of a time waister pl feel free to close :)

hamishwillee:
@julianoes Any opinion on this? 

What I see happening is that we add this, then nothing else happens - nothing changes in code or what people do. At some point more things get added to the code (if we're lucky) and they get out of sync. Upshot, it might make some kind of sense, but there isn't any real drive to do it.

olliw42:
> What I see happening is that we add this, then nothing else happens - nothing changes in code or what people do. 

incorrect. I would use it instead of having to define such things myself (this is what triggered the PR in fact, I found it wired that I have to define that flag myself or use an implementation specific definition of what is a protocol standard).

Second, a standard can't force people to maintain their code, but I find the logic strange to say that this is a reason for not advancing the standard ...

> At some point more things get added to the code (if we're lucky) and they get out of sync. 

how could this be possible? Nothing would ever be added to the standard which would get out of sync with itself, and if some code gets out of sync with the standard then this is a natural process for when code is not being maintained, and if the standard gets out of sync with some code then said code is development and not the definition of the standard and by definition ahead of the standard ... 

:)

julianoes:
> anyone using v2 messages will explicitly or implicitly have to use that flag !! and be it only through a code generator which provides convenient functions for you

If I understand this thread correctly, you want to add some enum flags to the xml so that they don't need to be hard-coded in the code generator, right?

While I'm all against hard-coded values and like anything that we can enforce through type safety (e.g. an enum class) I'm not sure if defining these flags in xml would help the code generator implementation due to something like a circular dependency. Basically, the parser and message generation code is written in C without the xml available. So, I think you would have to write the C parser assuming that the MAVLINK_INCOMPAT_FLAGS will later be available through some header file that does not exist just yet.

olliw42:
sharp and interesting argument, but I think it does not apply

in the code generator you would NOT have to define the value of that flag... very much in contrast to now. 

in the code generator one would of course have to use it's name in some code fragments. That's however the nature of the thing. For instance, you also need to know in advance the msgid of the heartbeat, in order to e.g. handle the mavlink_version field, even though at time of writing of the code generator the xml have never been processed. Another even more serious example would be the handling of the target fields. You need to know at time of writing of teh code generator the names of all the target fields in the messages even though the code generator doesn't know the messages yet. That is, the concern you raise is already deeply in there!

the proposal makes it easier however since you don't have to specify the value explicitly (and then hope that some tests pick up an error) but only have to specify the flag's name, which means that already any compiler will pick it up if in the code generator you would have used a wrong name

julianoes:
> any compiler will pick it up if in the code generator you would have used a wrong name

Right, that's true and probably a good argument in favor. I looked at the C helper files (such as https://github.com/ArduPilot/pymavlink/blob/master/generator/C/include_v2.0/mavlink_helpers.h) and none of them have includes depending on the xml, so that would have to be implemented. I hate paths and dependencies, so I'm not sure if it's worth the effort.

olliw42:
if you look at the mavlink.h, there is a large blocks of enum defines ... these are totally independent on everything ... which means they can be safely moved to the very top ... which means that anything which is included lateron has that incap flag enum available ...

so, technically (at least for C) it is perfectly working and no issue

> I hate paths and dependencies

my argument is you reduce dependencies LOL

> I guess I'm scared to try.

well, that's the argument - and also the above argument if it's worth the effort - which I have nothing to counter against :)

and as said, earth won't stop rotating if this would not be included, and I do have some level of sympathy for "never change a running system" which is kind of what stands against

but, you'll understand, I would cut another notch on my gun of not well designed features in paymavlink-mavgen LOL :D

julianoes:
> if you look at the mavlink.h, there is a large blocks of enum defines ... these are totally independent on everything ... which means they can be safely moved to the very top ... which means that anything which is included lateron has that incap flag enum available

julianoes:
Right, so include-wise it could work but it still requires the file `mavlink_helpers.h` to now include the file defining that enum, so it would require `mavlink_helpers.h` to include `minimal/mavlink`, right?

I guess I would be in favor of documenting enums but I would try to not mix MAVLink internal APIs (e.g. the header) with MAVLink external/user APIs (e.g. a payload field). These are enums at different levels of abstraction and should probably not be defined in the same place, in the same way that the magic start byte of MAVLink is defined in the specs and implementation but not in the minimal message set.

julianoes:
> The incompat flag is a part of the message like all the other variable fields like sysid, compid, and in fact also msgid etc. pp ... so, it is at the same level like msgid.

Right, I guess so.

> what I actually would find the logically cleanest would be a protocol.xml

That sounds better.

hamishwillee:
Hi @olliw42
We discussed this in dev call. There were some concerns that order of import might need to be looked at to make sure that the parser picked up the definitions needed everything else "first". 

The main consensus was that to proceed this we would need this an implementation that proves this for at least the the main languages - C++, Java and Python. If you would like to provide that then great!

If you don't think you will have time for that, can you please close this issue?

olliw42:
hahaha ... very smart, very smart indeed :) I like that.

I totally understand your position.

I however totally moved away from pymavgen and don't use it anymore in any of my projects (but use what I called fastMavlink) ... I assume that it wouldn't count as demonstration if I would implement it in my own lib, so I'm sorry having to say that I won't make an attempt ... (1/2 year earlier I would have done it)

so, closing :)

PR #1596, opened 2021-02-22
Make maklink a better standard, use CI to prevent message conflicts

amilcarlucas:
Needs review and approval from ASLUAV developers

### DIFF
--- a/message_definitions/v1.0/ASLUAV.xml
+++ b/message_definitions/v1.0/ASLUAV.xml
@@ -88,13 +88,6 @@
       <field type="float" name="param6">Parameter 6, as defined by MAV_CMD enum.</field>
       <field type="float" name="param7">Parameter 7, as defined by MAV_CMD enum.</field>
     </message>
-    <message id="201" name="SENS_POWER">
-      <description>Voltage and current sensor data</description>
-      <field type="float" name="adc121_vspb_volt" units="V"> Power board voltage sensor reading</field>
-      <field type="float" name="adc121_cspb_amp" units="A"> Power board current sensor reading</field>
-      <field type="float" name="adc121_cs1_amp" units="A"> Board current sensor 1 reading</field>
-      <field type="float" name="adc121_cs2_amp" units="A"> Board current sensor 2 reading</field>
-    </message>
     <message id="202" name="SENS_MPPT">
       <description>Maximum Power Point Tracker (MPPT) sensor data for solar module power performance tracking</description>
       <field type="uint64_t" name="mppt_timestamp" units="us"> MPPT last timestamp </field>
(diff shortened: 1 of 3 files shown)


julianoes (approved):
Fine from my side. Who would be a ASLUAV dev?

amilcarlucas:
@acfloria are you the ASLUAV maintainer?

acfloria:
We do not have a designated maintainer for it but I can for sure review the PR.

I am not that familiar with the latest decisions made with regards to MAVLink. However, I remember and there is still a comment in the code indicating that the message ID's between 180 to 229 are reserved for the individual dialects:

`<!-- MESSAGE IDs 180 - 229: Space for custom messages in individual projectname_messages.xml files -->`

Since the ardupilotmega and ASLUAV are two seperate dialects I think an ID collision in this range should be fine. I would argue the CI should only check the if there is a collision in each dialect and not between the different dialects.

While for now it would be fine to change the ID and add this check it might be an issue in the future if you keep adding messages in that range in the ardupilot dialect and we would eventually run out of free ID's.

Is there any specific reason to add this check? Do you want use some of the message definitions?
If that is the case we would be open to porting the message definitions to common.xml since this anyway our workflow for definitions that are stable and might be useful to the community.

amilcarlucas:
Yes, ardupilotmega and ASLUAV are two seperate dialects and can therefor conflict (without violating the sandard) and have been conflicting.

My proposal here is to reduce those conflicts. This makes the standard more consistent. And allows GCS to implement the full standard without needing a lot of switch-case statements (some are probably needed, but this reduces that number).

It makes the standard a lot simpler IMHO. And the CI test make sure it stays conflict free.

@LorenzMeier what do you think about this?

hamishwillee:
Hi @acfloria

As @amilcarlucas says, this is designed to help dialect owners (and the standard) by reducing id conflicts. With this change the dialect can be added to all.xml, which means that going forward you'll be able to test any conflicts in your allocations. That will further mean that vehicles running auslav can safely co-exist on a mavlink 2 network with other compliant dialects.

The cost is that all systems that use these messages will need to be updated.

This is not mandatory; if you don't can't  or won't do it then we simply won't include auslav.xml in all.xml. But we'd like to (any new dialects will need to avoid conflicts).

Please let us know whether we can proceed merging this. 

FYI (you probably know this!)
- MAVLink 2 moved away from shared UID ranges because it isn't limited to a 255 id message space. New messages should be allocated in an assigned range: https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges
If you're not using MAVLink 1 please do not allocate into the 0-255 range for any new messages.

acfloria:
Sorry for the late response from my side.

We did discuss it internally with the whole team that is using the dialect. While we see some benefits checking for collisions we also see the risks I described earlier. Our opinion is that the dialects are here to have different message definitions for the same message ID, if we don't allow that anymore then we could just directly move to one dialect. This is the reason why, at least for now, we think that not adding asluav.xml to all.xml is the way to go.

I actually did not know about the assigned MAVLink2 ranges. I will have a look at it and adjust our dialect accordingly. We are anyway refactoring some of our code and in that process might move some messages to the MAVLink2 range.

hamishwillee:
Thanks @acfloria 

> I actually did not know about the assigned MAVLink2 ranges. I will have a look at it and adjust our dialect accordingly. We are anyway refactoring some of our code and in that process might move some messages to the MAVLink2 range.

If you can move all your messages to the MAVLink 2 ranges that would be great. But if no, so be it.

@amilcarlucas I am closing this. We can't "make" people choose to accept this change.

amilcarlucas:
That is unfortunate for them. This adds barriers to GCS support. Their loss....

PR #1597, opened 2021-02-22
Add PARAM_TYPE_CHAR and PARAM_EXT_TYPE_CHAR

olliw42:
for some reason the enums MAV_PARAM_TYPE and MAV_PARAM_EXT_TYPE do not have entries for value 0. The mavgen C code however specifies MAVLINK_TYPE_CHAR = 0. I've done some searching if it is for a reason that this is so, but couldn't retrieve any. Also according to mavschema.xsd, char is a valid entry.

Hence this PR, which adds MAV_PARAM_TYPE_CHAR and MAV_PARAM_EXT_TYPE_CHAR, each with value 0.

If having omitted them was on purpose, please accept my apologies.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2340,6 +2340,9 @@
     </enum>
     <enum name="MAV_PARAM_TYPE">
       <description>Specifies the datatype of a MAVLink parameter.</description>
+      <entry value="0" name="MAV_PARAM_TYPE_CHAR">
+        <description>8-bit char</description>
+      </entry>
       <entry value="1" name="MAV_PARAM_TYPE_UINT8">
         <description>8-bit unsigned integer</description>
       </entry>
@@ -2373,6 +2376,9 @@
     </enum>
     <enum name="MAV_PARAM_EXT_TYPE">
       <description>Specifies the datatype of a MAVLink extended parameter.</description>
+      <entry value="0" name="MAV_PARAM_EXT_TYPE_CHAR">
+        <description>8-bit char</description>
+      </entry>
       <entry value="1" name="MAV_PARAM_EXT_TYPE_UINT8">
         <description>8-bit unsigned integer</description>
       </entry>


hamishwillee:
@olliw42 

I "think" that the mavgen `MAVLINK_TYPE_*` enum is an internal definition used for handling the different data types in message fields. So for example if you knew you had a char, then test code might build a string rather than an array. 

I don't think it is related to parameters/parameter use. I'm pretty sure the reason param_ext was created was to support strings - though we use MAV_PARAM_EXT_TYPE_CUSTOM for this. 

I might be missing something, but before we merged this we'd need to confirm that char is a valid param type, and that it is represented by 0.

olliw42:
That's interesting. 

I believe what can be considered facts is
* value 0 is not used in PARAM_TYPE_*, which is somewhat weird by itself
* a type char is allowed for attribute 'type' according to mavschema.xsd (https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L39-L57)
* except of char (and uint8_mavlink version and array, which do not count for the matter) the attribute 'type" in mavschema.xsd exactly matches the list of types in PARAM_TYPE_*.

So, char is missing. And 0 is unused.

I agree though that this does not strictly prove that 0 should be char.

Hence:
> we'd need to confirm that char is a valid param type

from the above logic the answer would have to be yes it is

> and that it is represented by 0.

from the above logic this indeed cannot be strictly concluded.

hamishwillee:
>from the above logic the answer would have to be yes it is

No we can't. If parameter protocol does not allow chars then the value of 0 is invalid.

olliw42:
which would imply that mavschema.xsd is incorrect in this point, wouldn't it?

hamishwillee:
Or the converse, why should it be needed? Each parameter is just one value, so do we ever need to save an "A", "B" etc, and if so,  would it be worth the implementation effort vs say just saving the value as a uint8 and converting. 

Use cases do exist for character data, which is to store strings. That got handled by the custom type in the extended parameter protocol. 

Now I am not adverse to considering this - it isn't my call. But probably I wouldn't use 0 for it.

olliw42:
well, I agree, it is not like that earth would stop rotating without a char type, and it indeed would cause a trail of effort.

I was thinking that char is established and was just forgotten to be listed, and the "evidences" and especially the missing 0 value has mislead me. I understand that now, thx for the lesson. 
I find it still strange that there is no 0 value, but that's probably for some historical reason. Maybe on could add a short comment to the description like '(0 is invalid)' to avoid that another olliw-like smart ass comes by and raises that issue LOL ...

hamishwillee:
Probably adding a comment would not hurt :-). I wouldn't add an actual value, because then someone would use it :-)

olliw42:
I close this.
Value 0 is not missing because of an oversight, as I was thinking. Neither MP, QGC, nor AP are using it. So the "evidence" I have interpreted incorrectly.
Sorry for that :).

hamishwillee:
Thanks. Glad someone is watching.

PR #1599, opened 2021-02-24
C2000 parse

Saleh-At:
Hi there
tanks alot for your mavlink_c2000 but I cant use `mavlink_parse_char(...)`  from QGC. I have a reader variable with type `char` in my tms320 for reading serial port data of  QGC but cant handle that for example MISSION PROTOCOL for send waypoints!
note: some times can handle it but more times NOT

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,9 +8,8 @@ build*
 .nfs*
 share/
 __pycache__
-pymavlink/generator/C/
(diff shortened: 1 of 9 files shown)


amilcarlucas:
This pull request belongs in the upstream Ardupilot/pymavlink github repository. Please post it there.

PR #1601, opened 2021-03-02
common.xml: added WiFi network info message

aviaks:
Added WiFi network info message and enum for WiFi network security types.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -43,12 +43,12 @@
       <field type="uint32_t" name="checksum">CRC32 checksum of current plan for specified type.</field>
     </message>
     <message id="298" name="WIFI_NETWORK_INFO">
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Detected WiFi network status information. This message is sent per each WiFi network detected in range with known SSID and general status parameters.</description>
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->      <description>Detected WiFi network status information. This message is sent per each WiFi network detected in range with known SSID and general status parameters.</description>
       <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID).</field>
       <field type="uint8_t" name="channel_id">WiFi network operating channel ID. Set to 0 if unknown or unidentified.</field>
-      <field type="uint8_t" name="signal_quality" units="%">WiFi network measured signal quality. Normalized to percentage. 0 = unknown/unset signal quality, 1 = bad signal quality, 100 = perfect signal quality.</field>
-      <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to 0 if unknown or unidentified.</field>
+      <field type="uint8_t" name="signal_quality" units="%">WiFi network signal quality.</field>
+      <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.</field>
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
     </message>
   </messages>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -44,7 +44,8 @@
     </message>
     <message id="298" name="WIFI_NETWORK_INFO">
       <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->      <description>Detected WiFi network status information. This message is sent per each WiFi network detected in range with known SSID and general status parameters.</description>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Detected WiFi network status information. This message is sent per each WiFi network detected in range with known SSID and general status parameters.</description>
       <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID).</field>
       <field type="uint8_t" name="channel_id">WiFi network operating channel ID. Set to 0 if unknown or unidentified.</field>
       <field type="uint8_t" name="signal_quality" units="%">WiFi network signal quality.</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -5,7 +5,7 @@
   <version>0</version>
   <dialect>0</dialect>
   <enums>
-      <enum name="WIFI_NETWORK_SECURITY">
+    <enum name="WIFI_NETWORK_SECURITY">
       <description>WiFi wireless security protocols.</description>
       <entry value="0" name="WIFI_NETWORK_SECURITY_UNDEFINED">
         <description>Undefined or unknown security protocol.</description>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -5,7 +5,27 @@
   <version>0</version>
   <dialect>0</dialect>
   <enums>
-    <!-- none defined (yet) -->
+      <enum name="WIFI_NETWORK_SECURITY">
+      <description>WiFi wireless security protocols.</description>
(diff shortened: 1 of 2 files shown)


hamishwillee:
Hi @aviaks, 

What's the use case/where do you plan to to deploy this/how do you plan to use this?

The reason I ask is that we're trying to move towards a model where common.xml contains things that are genuinely "common"; that means things that have a prototype implementation and a commitment from at least 2 flight stacks to use them.

@julianoes @auturgy FYI, is this something considered of use for ArduPilot or PX4.

julianoes:
Wait what, there is WPA3? Cool :).

@aviaks how would this be used? And would it be used together with WIFI_CONFIG_AP_RESPONSE and the other existing messages?

aviaks:
Hi @hamishwillee and @julianoes and thanks for your response! 

The current implementation has WIFI_CONFIG_AP ( #299 ) message which allows to setup WiFi connection on the autopilot / mission computer side from the GCS. However, the user has to know SSID in advance from somewhere else and type its name manually. The message I'm proposing allows to show the list of available WiFi networks (discovered by autopilot / mission computer) in the GCS. Then the GCS user is able to see available networks and their status and choose the one to connect to with a click, then type password only and then GCS sends WIFI_CONFIG_AP.   

I think this would be a useful addition to the existing WiFi related workflow.

hamishwillee:
@aviaks We have a mavlink meeting today - in about 8 hours if you would like to attend and discuss?

aviaks:
@hamishwillee yes, I might be able to attend. Is it 8am or 9am CET? It says 9am [here](https://mavlink.io/en/about/support.html) but marked 8am in this [calendar](https://www.dronecode.org/calendar/)?

hamishwillee:
@aviaks 8am. The calendar is always right. I have updated the docs. That's in about 3 hours from now.

LorenzMeier (approved):

aviaks:
@hamishwillee it was great to talk earlier today! As discussed, I have just added the information for the fields in case the value is unknown/unidentified.

hamishwillee:
@aviaks Thanks for coming along to the meeting! 

We're all happy for it to be in the standard, but the new process for getting things "standard" is that they only go into common.xml/standard.xml when they really are "standard". What that means is that going forward we need a prototype implementation in one stack, and at least "commitment" from another stack that they will include it.

What does that mean for you? It means that while you are doing your implementation, the messages should be in [development.xml](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/development.xml) rather than common.xml. Can you move the changes and I can then merge.

When you have a working version you would PR a move to common.xml.

Hope that makes sense? We can discuss offline if not.

aviaks:
@hamishwillee yes, I agree, this totally makes sense. Just moved the changes to development.xml and pushed.

message_definitions/v1.0/development.xml
```diff
@@ -22,5 +42,14 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
       <field type="uint32_t" name="checksum">CRC32 checksum of current plan for specified type.</field>
     </message>
+    <message id="298" name="WIFI_NETWORK_INFO">
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
```
hamishwillee, line 46: Can you also add a tag `<wip/>` here? (mirror the case above).

message_definitions/v1.0/development.xml
```diff
@@ -22,5 +42,14 @@
...
+      <description>Detected WiFi network status information. This message is sent per each WiFi network detected in range with known SSID and general status parameters.</description>
+      <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID).</field>
```
hamishwillee, line 48: Is 32 chars enough? I'm not saying it is not, I'm "asking the question". 

As general rule if info is sent rarely/on startup then it does not hurt to have longer messages (and hence fields). If it is sent more often we try to keep the size down.

message_definitions/v1.0/development.xml
```diff
@@ -22,5 +42,14 @@
...
+      <field type="uint8_t" name="channel_id">WiFi network operating channel ID. Set to 0 if unknown or unidentified.</field>
+      <field type="uint8_t" name="signal_quality" units="%">WiFi network measured signal quality. Normalized to percentage. 0 = unknown/unset signal quality, 1 = bad signal quality, 100 = perfect signal quality.</field>
```
hamishwillee, line 50: Suggest change this to "WiFi network signal quality.". The rest is all implied by using units of %.

message_definitions/v1.0/development.xml
```diff
@@ -22,5 +42,14 @@
...
+      <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to 0 if unknown or unidentified.</field>
```
hamishwillee, line 51: 0 is a valid data rate right? If so then setting it as value for "unknown/not supported" is problematic, because a GCS doesn't know if you're saying "I don't support this" vs "I just have a really awful data rate"

HOw about

> WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.

hamishwillee:
Thanks very much. I did a final detail review on this, and picked up a couple of details. Will merge after you've had a chance to look at those.

message_definitions/v1.0/development.xml
aviaks: Sure, `<wip/>` added.

message_definitions/v1.0/development.xml
aviaks: Good question. The reason why I put 32 chars is because `WIFI_CONFIG_AP` message is also using 32 chars here:

```
<message id="299" name="WIFI_CONFIG_AP">
      <description>Configure WiFi AP SSID, password, and mode. This message is re-emitted as an acknowledgement by the AP. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE</description>
      <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID). Blank to leave it unchanged when setting. Current SSID when sent back as a response.</field>
      <field type="char[64]" name="password">Password. Blank for an open AP. MD5 hash when message is sent back as a response.</field>
      <extensions/>
```

So we can extend but it will be a bit useless if we don't extend the one in `WIFI_CONFIG_AP`. We will be able to read long network names but not to connect to them. I think we should either increase the size of both then or leave both with 32 chars and maybe extend both of them at some point later.

message_definitions/v1.0/development.xml
aviaks: I agree, done. This description was originally made to be similar to this one where percentage meaning explained explicitly:
```
<message id="132" name="DISTANCE_SENSOR">
<field type="uint8_t" name="signal_quality" units="%">Signal quality of the sensor. Specific to each sensor type, representing the relation of the signal strength with the target reflectivity, distance, size or aspect, but normalised as a percentage. 0 = unknown/unset signal quality, 1 = invalid signal, 100 = perfect signal.</field>
```

message_definitions/v1.0/development.xml
aviaks: Right, I tend to agree. Probably both 0 and 65535 are not ideal but there is almost no chance of ~65 GiB/s data rate. 😊
So I changed according to your proposal.

aviaks:
Thanks @hamishwillee! All suggestions make sense. Regarding increasing SSID name size - I'm not sure, please see my explanation above.

message_definitions/v1.0/development.xml
hamishwillee: Thanks! Ignore my point - the max size is 32 bytes according to the standard that defines this field in Wifi. Phew! 

Merging now. Thanks for your patience.

hamishwillee:
PS Looking forward to when you've got an implementation. At that point you can PR into common.xml.

aviaks:
@hamishwillee Great, thanks Hamish! I will share the implementation in a few weeks time once this is ready. Many thanks again!

hamishwillee:
@aviaks I'm cleaning up MAVLink development.xml. As per https://github.com/mavlink/mavlink/pull/2164 thinking about removing this as there is no implementation.

Issue #1584, opened 2021-02-14
Support for 2 airspeed sensors


Davidsastresas:
Related to this issue from QGC: https://github.com/mavlink/qgroundcontrol/issues/9435

Right now Mission Planner and Ardupilot are able to display this second airspeed sensor. They are using a named float mavlink message.

I think the setups with 2 airspeed sensors are quite common nowadays, so it would be great to add support for this in mavlink.

hamishwillee:
Mavlink appears to assume you just want to know the current airspeed, not how it is estimated.

Davidsastresas:
I saw a couple of Ardupilot users complaining of this fact. In Ardupilot right now the second airspeed sensor is sent by a named float message, and Mission planner has support for interpreting that second airspeed by the named float message.

So it seems the users were used to mission planner, and when they switched to QGC they missed this feature.

Sorry I wasn't able to attend the call, I just saw the dev call label.

hamishwillee:
Hi @Davidsastresas 

We did discuss it, but we couldn't resolve it without more information. The information is 
- "why is it needed". That isn't really answered by "ArduPilot users complained that it isn't there".
- "what information do you want" - indicated airspeed, calibrated airspeed, true airspeed, etc. 

You're also talking about a "second" airspeed sensor - but what is the first? If you're using something like VFR_HUD for the "first" then it might not be an airspeed sensor at all. It could be a fused value from multiple sensors, or the fallback of multiple sensors, or an airspeed _calculated_ from vehicle dynamics.

On some systems where there are two airspeed sensors the fallback happens automatically. In this case there is usually a status message that is used to log the changeover and for debugging.

Upshot, can you please more precisely characterise what it is that is required of an airspeed message?

Davidsastresas:
Well, I think it is a sensor, as a GPS, as an Imu, etc and we have support for sending several of those by mavlink. Nowadays it isn't strange to have a setup with 2 GPS or 2 airspeed sensors for redundancy. In fact most serious fixed wing platforms have dual airspeed sensors because it is a sensor prone to fail due to obstructions, etc. So I really think it makes sense to have information about both airspeed sensors, don't you think?

pompecukor:
> 
> 
> Hi @Davidsastresas
> 
> We did discuss it, but we couldn't resolve it without more information. The information is
> 
>     * "why is it needed". That isn't really answered by "ArduPilot users complained that it isn't there".
> 
[...]

PR #1602, opened 2021-03-03
Common.xml: AIRSPEED message

hamishwillee:
Proposed message for AIRSPEED_SENSOR instances, as discussed in #1584. Assumption is sending rate would be low - 1 or 2 Hz, and probably something streamed on request using MESSAGE_INTERVAL rather than "by default".

Has 
- instance: id
- airspeed (IAS)
- temperature
- differential pressure
- static pressure. 

Points for discussion
- Are there any other fields needed? Sensor type?
- Are these the right units? (I copied similar fields in other messages)

FYI @dagar @sfuhrer @RomanBapst @julianoes @auturgy @LorenzMeier 

PS Under current rules I assume we don't merge until there is an implementation, but we can at least approve the design

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6337,6 +6337,7 @@
       <field type="float[4]" name="voltage" units="V">Voltage measured from each ESC.</field>
       <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
+    <!-- id 295 reserved for AIRSPEED message (development.xml). -->
     <message id="299" name="WIFI_CONFIG_AP">
       <description>Configure WiFi AP SSID, password, and mode. This message is re-emitted as an acknowledgement by the AP. The message may also be explicitly requested using MAV_CMD_REQUEST_MESSAGE</description>
       <field type="char[32]" name="ssid">Name of Wi-Fi network (SSID). Blank to leave it unchanged when setting. Current SSID when sent back as a response.</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -26,6 +26,24 @@
         <description>WPA3.</description>
       </entry>
     </enum>
+    <enum name="AIRSPEED_SENSOR_TYPE">
+      <description>Types of airspeed sensor/data. May be be used in AIRSPEED message to estimate accuracy of indicated speed.</description>
+      <entry value="0" name="AIRSPEED_SENSOR_TYPE_UNKNOWN">
+        <description>Airspeed sensor type unknown/not supplied.</description>
+      </entry>
+      <entry value="1" name="AIRSPEED_SENSOR_TYPE_DIFFERENTIAL">
+        <description>Differential airspeed sensor</description>
+      </entry>
+      <entry value="2" name="AIRSPEED_SENSOR_TYPE_MASS_FLOW">
+        <description>Mass-flow airspeed sensor.</description>
+      </entry>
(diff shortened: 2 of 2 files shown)


LorenzMeier (approved):

sfuhrer:
I would consider also having fields for calibrated (CAS) and true (TAS) airspeed, as they are what's usually used for control and navigation. In the [VFR_HUD](https://mavlink.io/en/messages/common.html#VFR_HUD) we're leaving it open if the "airspeed" field is corresponding to IAS or CAS, but for a new message I would rather have it clearly determined.

hamishwillee:
@sfuhrer The argument for IAS was that a sensor normally puts out IAS, with calculation of CAS or TAS done in the autopilot - so not "AIRSPEED_RAW". I have no opinion personally, other than to say that if we can provide both then that would be fine by me. 

@auturgy thoughts?

sfuhrer:
> @sfuhrer The argument for IAS was that a sensor normally puts out IAS, with calculation of CAS or TAS done in the autopilot - so not "AIRSPEED_RAW". I have no opinion personally, other than to say that if we can provide both then that would be fine by me.

Well I think there are also CAN airspeed sensors out there that directly output IAS and TAS. But I guess it would also be fairly easy to add fields for that later once we have a concrete implementation out.

auturgy:
A separate message, as this message is intended to reflect sensor readings not estimator output.

Regards,

James

Jaeyoung-Lim:
@auturgy I was talking about sensor readings. The ASLUAV dialect has a message that outputs the vane values in mavlink.

hamishwillee:
@Jaeyoung-Lim I'd say that if these are actually output *by the sensor* then they are possible candidates for this message. Are they (It isn't clear from [SENSOR_AIRFLOW_ANGLES](https://mavlink.io/en/messages/ASLUAV.html#SENSOR_AIRFLOW_ANGLES) if they are calculated from other data in the autopilot)?

Do we also need sideslip ? What are these values used for?

Jaeyoung-Lim:
@hamishwillee I thought the `SENSOR_` prefix means that the information was coming out of a sensor :smile: Yes, they are sensor outputs and these are important information for FW/VTOLs that are flying out of the cruising flight envelop and probably optionally used for system identification.

hamishwillee:
@Jaeyoung-Lim Ha ha. The `_RAW` tells you the data is unscaled/not via the autopilot. 

So would the addition be.
```
<field type="float" name="angle_of_attack" units="rad">Angle of attack. NaN if not supported.</field>
<field type="float" name="sidelslip" units="rad">Sideslip angle of attack. NaN if not supported.</field>
```

@auturgy Further comments?

hamishwillee:
Discussed in the devcall. In summary, angle_of_attack/sideslip will be synthesized values from an airspeed computer of some kind - so whether or not they are "sensor" they are not "raw". So if these are part of the message there is an objection to the name.

@auturgy Has agreed to propose a message for air computer data (either a replacement for this, or something that co-exists to send the additional information). This will be discussed in the next dev call and some decision made on how to proceed.

hamishwillee:
@LorenzMeier @auturgy @julianoes FYI, as discussed in dev call this updates `AIRSPEED` with fields for error and sensor type (enum). The airspeed is documented as CAS if available, otherwise IAS (it is no longer raw sensor data, so that is OK).

I added message/enum to `development.xml` (rather than common.xml) as per new process, and added a comment in common.xml just to note the id reservation.

Any objections to merging?

hamishwillee:
OK, as there are no objections I have merged as WIP into development.xml.

PR #1603, opened 2021-03-04
common: fix HDOP/VDOP unit scaling

julianoes:
I believe that the HDOP/VDOP values are generally scaled by 100 and we should therefore document it that way.

PX4:
*.01f when the UBX values are read:
https://github.com/PX4/PX4-GPSDrivers/blob/c278eb48e1b57e6b789242324e28581caf9ccf96/src/ubx.cpp#L1815-L1816

*100 before the values are sent out:
https://github.com/PX4/PX4-Autopilot/blob/5bbc66f3afe99e75852e9b3660a5de4d2fc0b902/src/modules/mavlink/mavlink_messages.cpp#L1393-L1394

ArduPilot:
It seems to be *100 everywhere without conversion from UBX:
https://github.com/ArduPilot/ardupilot/blob/ecf928d68ba497dd2502ab56037f6aaa1bc298e2/libraries/AP_GPS/AP_GPS_UBLOX.cpp#L1280-L1281
And sent out as is:
https://github.com/ArduPilot/ardupilot/blob/ecf928d68ba497dd2502ab56037f6aaa1bc298e2/libraries/AP_GPS/AP_GPS.cpp#L1238-L1239

I left GPS_INPUT unchanged as there is no scaling documented and that's
also how it is implemented by ArduPilot:
https://github.com/ArduPilot/ardupilot/blob/ecf928d68ba497dd2502ab56037f6aaa1bc298e2/libraries/AP_GPS/AP_GPS_MAV.cpp#L72-L78

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4408,8 +4408,8 @@
       <field type="int32_t" name="lat" units="degE7">Latitude (WGS84, EGM96 ellipsoid)</field>
       <field type="int32_t" name="lon" units="degE7">Longitude (WGS84, EGM96 ellipsoid)</field>
       <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up. Note that virtually all GPS modules provide the MSL altitude in addition to the WGS84 altitude.</field>
-      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
-      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
       <field type="uint16_t" name="vel" units="cm/s">GPS ground speed. If unknown, set to: UINT16_MAX</field>
       <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement) in degrees * 100, 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
       <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4408,8 +4408,8 @@
       <field type="int32_t" name="lat" units="degE7">Latitude (WGS84, EGM96 ellipsoid)</field>
       <field type="int32_t" name="lon" units="degE7">Longitude (WGS84, EGM96 ellipsoid)</field>
       <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up. Note that virtually all GPS modules provide the MSL altitude in addition to the WGS84 altitude.</field>
-      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
-      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
```
amilcarlucas, line 4412:
```suggestion
      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless*100). If unknown, set to: UINT16_MAX</field>
      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless*100). If unknown, set to: UINT16_MAX</field>
```

amilcarlucas: for consistency

amilcarlucas (changes requested):

message_definitions/v1.0/common.xml
julianoes: Oops, thanks!

amilcarlucas:
looks good to me, but I would like to ear @WickedShell's input on this

WickedShell:
That is what ArduPilot's doing, so from an ArduPilot point of view seems reasonable.

julianoes:
Thanks @amilcarlucas and @WickedShell!

amilcarlucas:
Why is this still marked as open? A couple of lines above it says the commit has been merged.

PR #1606, opened 2021-03-09
Add a new bit field to indicate position target reached

chobitsfan:
It allows a vehicle to inform GCS through POSITION_TARGET_LOCAL_NED (with type_mask 0x1DF8) that the desired position target is reached.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3529,6 +3529,9 @@
       <entry value="2048" name="POSITION_TARGET_TYPEMASK_YAW_RATE_IGNORE">
         <description>Ignore yaw rate</description>
       </entry>
+      <entry value="4096" name="POSITION_TARGET_TYPEMASK_POS_REACHED">
+        <description>vehicle reached position target</description>
+      </entry>
     </enum>
     <enum name="ATTITUDE_TARGET_TYPEMASK" bitmask="true">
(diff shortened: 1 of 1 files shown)


chobitsfan:
Thank you @hamishwillee 

The usage scenario is that GCS send SET_POSITION_TARGET_LOCAL_NED (type_mask = 0xDF8 ) to a vehicle. The vehicle keep sending POSITION_TARGET_LOCAL_NED (type_mask = 0xDF8) back to GCS when it moving toward target. The vehicle can send POSITION_TARGET_LOCAL_NED (type_mask = 0x1DF8) once it reach target.

hamishwillee:
@chobitsfan Thanks. I understand what you are trying to do. 

What I don't understand is WHY you want to do this. These setpoints are designed for sending a stream of points from a companion. The companion tracks completion - the autopilot doesn't know or care about any particular setpoint being reached and might never reach any of them before the next one comes along.

But say you do want to think of setpoints like some kind of waypoint this still doesn't make sense - because the autopilot does not know what the sender means by "reaching the setpoint". While for a waypoint it does - it means within the acceptance radius. So it should be up to the companion to work out when the setpoint has been reached.

Hope that makes sense to you.

chobitsfan:
Hi @hamishwillee You are correct and I misuse this message. Thank you very much for kindly help and detailed explanation

PR #1609, opened 2021-03-14
Added safety switch to MAV_SYS_STATUS_SENSOR enum

garfieldG:
relates to https://github.com/PX4/PX4-Autopilot/pull/17089

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -214,6 +214,9 @@
       <entry value="1073741824" name="MAV_SYS_STATUS_SENSOR_PROPULSION">
         <description>0x40000000 propulsion (actuator, esc, motor or propellor)</description>
       </entry>
+      <entry value="2147483648" name="MAV_SYS_STATUS_SENSOR_SAFETY_SWITCH">
+        <description>0x80000000 safety switch </description>
+      </entry>
     </enum>
     <enum name="MAV_FRAME">
       <entry value="0" name="MAV_FRAME_GLOBAL">


peterbarker:
ArduPilot has reported "safety" switch state in`SYS_STATUS` for a very, very long time.  About 6 years (https://github.com/ArduPilot/ardupilot/commit/9bcf9a31ed569de84bc9353bd9b3b17f8e8c5c9b)

We use `MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS`: https://github.com/ardupilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS.cpp#L209

This is the only thing we use the bit for - we do not, for example, leave the bit unset if the vehicle isn't "armed".  Solely disabling the hardware safety is sufficient to switch the bit, so we appear to be using it in exactly the fashion the proposed bit here is designed to be used.

If you're not using `MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS` (a quick grep says "no"), I suggest using it for the same information.  We have very, very few bits remaining in the fields that convey these bits to the GCS - punting that problem down the road a little further seems like a good idea to me :-)

garfieldG:
Closing this pr due to [peter comment](https://github.com/mavlink/mavlink/pull/1609#issuecomment-798995680)

hamishwillee:
@peterbarker Any thoughts on the best way to make this clear in the docs? From what I can tell we're saying that if the `SYS_STATUS.onboard_control_sensors_enabled` bit for `MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS` is set this means that the safety switch is disarmed - and that is the only thing that controls enabling/disabling the output right?

So perhaps change  `MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS`  from 

> 0x8000 motor outputs / control

TO

> 0x8000 motor outputs / control (disabled when safety switch is armed)

Reasonable?

peterbarker:
On Tue, 16 Mar 2021, Hamish Willee wrote:

> @peterbarker Any thoughts on the best way to make this clear in the 
> docs? From what I can tell we're saying that if the 
> SYS_STATUS.onboard_control_sensors_enabled bit for 
> MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS is set this means that the safety 
> switch is disarmed - and that is the only thing that controls 
> enabling/disabling the output right?
> 
> So perhaps change MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS from
>
>       0x8000 motor outputs / control
> 
> TO
>
>       0x8000 motor outputs / control (disabled when safety switch is armed)
> 
> Reasonable?

It's.... different :-)

Problems arise because the safety switch is essentially silly, and it is 
difficult to have a field like this without tying it back into the 
IOMCU architecture of fmuv2 / fmuv3 etc.

Thing is that ArduPilot (and probably PX4) also disable motor outputs via 
"soft" arming - but ArduPilot at least does *not* modify the bit based on 
soft-arming.  In fact on Plane they're orthogonal - you can be armed but 
have the safety enabled - so you end up "Armed (safe)"

Your description above, then, is accurate but insufficient - outputs are 
inhibitted in ways other than the safety switch.

0x8000 hardware safety switch state

(completely abusing "hardware" when refering to the fmuv2/3-style safety 
switch, but might be better?

Should I bring this up at a AP devcall?

hamishwillee:
@peterbarker Thanks for the clarification. PX4 also has a soft arming option (called [preaming](http://docs.px4.io/master/en/advanced_config/prearm_arm_disarm.html)) that allows non-dangerous servos to be active with motors disabled (I also don't know how safety switch impacts the sensor reporting in this case).

Does this match your interpretation:
- IF the `SYS_STATUS.onboard_control_sensors_enabled` bit for MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS is set to `1` is that the motor/servo is _capable_ of movement (if/when armed or prearmed). i.e. this means it is not safety switch or otherwise "hardware" interlocked. 
-  IF the `SYS_STATUS.onboard_control_sensors_health` bit for MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS is set to `1`  that means the motor outputs ARE software enabled. In other words we're armed. This would not be set true in a prearmed state.

So for plane you could have Armed (safe) with the enabled bit set `0` and health set `1`. This is a bit weird if you think in terms of sensors, but works if you think of them as hardware interlock and software interlock. 

I would expect `MAV_SYS_STATUS_SENSOR_MOTOR_OUTPUTS` bit is set to `0` in soft arm state - the motors are still hardware interlocked. If that is what you are saying, that is correct.

Upshot of all of that is
1. Perhaps `0x8000 motor hardware interlock state` nosoftware"
1. A user might still be interested in just the safety switch state - because safety switch may also affect servos (see that soft arm mode) and as you say, other things might affect the ability to enable motors - but you still might want some GUI visual that your safety switch is on.

Issue #1588, opened 2021-02-16
Discussion: COMPONENT_INFORMATION extension


bkueng:
I'm opening this to discuss the extension of the [COMPONENT_INFORMATION](https://mavlink.io/en/services/component_information.html) API to external components, UAVCAN devices in particular.

At the minimum, it should be possible to get (static) version information and parameter metadata for those devices.

I suggest to extend the `COMP_METADATA_TYPE_VERSION` json file with an optional list of extra components:
```
{
  extra_components: [
  { "id": x, "name": "human readable name, like ESC0", "modelName": xy, "firmwareVersion": xy, "supportedCompMetadataTypes": [1]},
   ...
  ]
}
```
Each device would have its entry, and the autopilot generates the entries during bootup, and keeps it in RAM or writes to SD card (so no hotplugging).

It allows a great deal of flexibility for the autopilot handling & hosting of metadata. From minimal to extended:
- a common (parameter) metadata file is hosted on a (mavlink) server used for all devices. The autopilot just returns that URL.
  (same limitation here as for translation files: the autopilot does not know how to set the `uid`, as the file might change on the server).
- device-specific metadata is hosted on a (mavlink) server. The autopilot sets the URL depending on the connected device type.
- the autopilot knows and serves the metadata itself (could also be a companion)
- the autopilot downloads the metadata through UAVCAN, stores it on the SD card and sets a MAVFTP URL.

### Metadata matching
Metadata needs to be matched against data, for example parameters. I see these options for that:
1) use mavlink component ID's (`id` above would be the comp_id). This is currently how it's implemented in PX4 for UAVCANv0 parameters (component id == UAVCAN node id). Going that path, it essentially means the autopilot emulates MAVLink components, with minimal functionality:
   - sending out heartbeats (so that routing works)
   - handling parameters (currently implemented in PX4)
   - handling COMPONENT_INFORMATION requests
   - handling MAVFTP requests
   - potentially events in the future
2) specifically for parameters: add namespacing via string prefix (`id` above would be the prefix). However param names are limited in length, and multiple devices of the same type need to be considered
3) extend the parameter protocol with an additional ID (a sub component id)

I tend towards 1), but we need to ensure that there's no conflicting component id's.

[...]

PR #1610, opened 2021-03-16
COMPONENT_INFORMATION extension and rework

bkueng:
From https://github.com/mavlink/mavlink/issues/1588

- move url to json file (except for the main one + external components, which are always mavlink ftp), allows to use longer URL's
- Standardize the uid to ([MAVLink's](https://mavlink.io/en/services/ftp.html#crc32-implementation)) CRC32. If compression is used, it should be over the compressed files. This has the small advantage of not having to decompress to generate it. It just means that enabling/disabling compression leads to a changed CRC and re-download, but I don't think that's an issue.
- adds a separate optional json file for external components (which is dynamically generated on bootup)
  - each can specify a list of `COMP_METADATA_TYPE`'s with uri's
  - it is explicitly not specified how to e.g. retrieve parameters, just the metadata. This could be done by adding an ID to each component in the metadata.

@DonLakeFlyer can you cross-check this, if it's what you had in mind?


message_definitions/v1.0/common.xml
```diff
@@ -6646,13 +6649,12 @@
...
-      <field type="char[70]" name="translation_uri">The translations for strings within the metadata file. If null the either do not exist or may be included in the metadata file itself. The translations specified here supercede any which may be in the metadata file itself. The uri format is the same as component_metadata_uri . Files are in Json Translation spec format. Empty string indicates no tranlsation file.</field>
+      <field type="uint32_t" name="general_metadata_file_crc">CRC32 of the TYPE_GENERAL file (can be used by a GCS for file caching).</field>
+      <field type="char[100]" name="general_metadata_uri">Component definition URI for TYPE_GENERAL. This must be a MAVLink FTP URI and the file might be compressed with xz.</field>
+      <field type="uint32_t" name="ext_comp_metadata_file_crc">CRC32 of the TYPE_EXT_COMPONENTS file (can be used by a GCS for file caching).</field>
+      <field type="char[100]" name="ext_comp_metadata_uri">(Optional) Component definition URI for TYPE_EXT_COMPONENTS. This must be a MAVLink FTP URI and the file might be compressed with xz.</field>
```
hamishwillee, line 6657: Why is this separate? My thinking was that there is just one URI which is used to request all other information about the component. The metadata that is returned would include the connected external types.

bkueng:
As discussed in the call:
- added a fallback URI
- make use of MAVLink's CRC32 used in mavftp (https://mavlink.io/en/services/ftp.html#crc32-implementation)

message_definitions/v1.0/common.xml
hamishwillee: @bkueng Your answer in devcall was "because the data is not known at build time, so cannot be supplied in the normal file".

Thinking about this, that's a reason why it can't be returned in the file, not a reason for a new URI. 

Would we be better to revert to specifying the COMP_METADATA_TYPE in the request? It could be either general (default) or COMP_METADATA_TYPE_EXT. It could even be the others too, but then we'd need to decide what to do if one of those was sent (i.e. send the URL and hash for the file still, or just ignore).

olliw42:
couldn't it maybe make sense to change language and to talk of external peripherals or just pheripherals instead of external components?
talking of components in a MAVLink context somehow triggers thoughts it has anything to do, even if only remotely, with MAVLink components...

bkueng:
> couldn't it maybe make sense to change language and to talk of external peripherals or just pheripherals instead of external components?

Yes that's a good point. Changed to peripherals.

hamishwillee (approved):
Works for me. The point of prototyping this is so that we can test whether we've seen the edge cases.

hamishwillee:
Merged. PS @bkueng Can you keep me informed of progress on implementation. I will need to update the associated docs, but I am assuming this might change a bit more going forward.

bkueng:
Thanks for merging, sure.

PR #1614, opened 2021-03-25
Add 'invalid' attribute to messages for invalid, unknown, to be not used, etc, fields

olliw42:
This PR suggests to add a new attribute 'invalid' to message fields, which indicates the value of the field when it shall be considered invalid, unknown, shall not be used, is not supplied, etc.

Quite some messages do have fields with a special value to indicate that.

The benefit of this attribute is that it allows code generators for some languages to automatically assign the fields, which simplifies coding for the user.

Lua would be such an example. Here a function can be written to send a message which is supplied with a table, which does not necessarily need to have keys for all message fields. The 'invalid' attribute now allows the function to assign the unspecified fields a proper value (or 0 if no invalid attribute is there). This can very much reduce the amount of work = code one has to write. For instance, smart battery messages have many fields which may not be supplied.

Python would be an example too, since here too one could write a generic function to send a message which is supplied with a dictionary, which does not necessarily need to have keys for all message fields.

The suggestion is not academic. I came accros it as part of my work on the Mavlink for OpenTx project, to which I have added now a generic api for sending arbitrary Mavlink messages via Lua scripts, and for resource reasons I exactly chose that approach, to provide a lua function to which a table is supplied. (https://github.com/olliw42/opentx/tree/mavtelem-dev/)

:)


olliw42:
I believe it fails because the mavschema.xsd in pymavlink needs to be updated for the new field 'invalid'. I used e.g the below and it makes mavgen to work for me.

this intertwinned dependency of mavlink on pymavgen is really unfortunate


[mavschema.zip](https://github.com/mavlink/mavlink/files/6209724/mavschema.zip)

hamishwillee:
FWIW I like the idea. I'll raise it in the dev call to see how others feel.

hamishwillee:
Sorry, we did not discuss this, but I have added to Agenda for next time. 

@amilcarlucas @julianoes @auturgy This is a proposal to explicitly add the "invalid" (not supported) value for a field that is currently in the description to the metadata. I think it is a good idea, though it will require associated changes to the XSD and to the generators (for documentation).

julianoes:
Nice @olliw42, I think that would be worthwhile to get in, once CI agrees!

olliw42:
well, if you all agree that this is good, and that calling the attribute "invalid" is appropriate, that my modification to mavschema.xsd is ok, and so on, I'll be happy to move forward

as indicated, I believe CI fails because of mavschema.xsd
so, I will raise a corresponding PR in the pymavlink repo, we will wait for it to be accepted, I then will refresh this PR to also update to the latest pymavlink, and then the CI should go away and this PR could be considered...

long journey for a simple addition ... so, let's do the next step ... https://github.com/ArduPilot/pymavlink/pull/526

hamishwillee:
@olliw42 Sounds good to me. I wanted to get broad agreement from @amilcarlucas as well as he is XSD gatekeeper - but you'll get that if you PR the change.

olliw42:
the xsd changes good merged to pymavlink now, so one could go on with the next step

I however put the pymavlink update into an extra PR here https://github.com/mavlink/mavlink/pull/1616. So this other PR would need to be merged first

:)

olliw42:
thx for merging https://github.com/mavlink/mavlink/pull/1616
so I did the last step and updated this PR here
and it at least seems to pass the checks now!! cool...

so, the PR seems ready

Some comments

* For some fields the invalid was not specified by an explicit number, like UINT8_MAX, but by an enum field. In that case I had used the enum field name and not the explicit number, like it is done in the description. I think this is reasonable. 
 
Is that ok?

* Some fields are arrays, and for arrays it is not soo obvious how to specify it's invalid state so it can be automated. That's because the procedures can be quite different for the various messages. For instance, for some it is specified that the first array field should be set to invalid for the whole array to be invalid, for others each array field can be set invalid, for others the whole array needs to be filled with the invalid values, for yet others the array needs to be filled with special values. So I did use some kind of meta format ... which might not yet be totally smooth. 
 
So I suggest this: I'm going to use the invalid field in a LUA script generator for my Mavlink for OpenTx project, so I can use this as reference to workout out and test a scheme handling all cases. I thus would suggest that we don't bother with this for the matter of this PR, and that I come up with a proposal when I did what I just said in the previous sentence. Ok?

* I think there might be also a non-functional tidy-up follow-up: Some of the ignore values are explicit, like 255 or 65535, which one might want to systematically replace by UINT8_MAX etc. pp. I am happy to do this, but would do it ion an extra PR to make review simpler. 
 
Is that ok or would you rather have me doing it now in this PR?

message_definitions/v1.0/common.xml
```diff
@@ -4420,18 +4420,18 @@
...
+      <field type="uint8_t" name="satellites_visible" invalid="255">Number of satellites visible. If unknown, set to 255</field>
       <extensions/>
       <field type="int32_t" name="alt_ellipsoid" units="mm">Altitude (above WGS84, EGM96 ellipsoid). Positive for up.</field>
       <field type="uint32_t" name="h_acc" units="mm">Position uncertainty.</field>
       <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty.</field>
       <field type="uint32_t" name="vel_acc" units="mm">Speed uncertainty.</field>
       <field type="uint32_t" name="hdg_acc" units="degE5">Heading / track uncertainty</field>
-      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="uint16_t" name="yaw" units="cdeg" invalid="65535">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
```
hamishwillee, line 4434: Hmmm should this one change to UINT16_MAX ? That is certainly the intent and better to be consistent

message_definitions/v1.0/common.xml
```diff
@@ -4793,31 +4793,31 @@
       <field type="float" name="ax" units="m/s/s">X Acceleration</field>
       <field type="float" name="ay" units="m/s/s">Y Acceleration</field>
       <field type="float" name="az" units="m/s/s">Z Acceleration</field>
-      <field type="float[45]" name="covariance">Row-major representation of position, velocity and acceleration 9x9 cross-covariance matrix upper right triangle (states: x, y, z, vx, vy, vz, ax, ay, az; first nine entries are the first ROW, next eight entries are the second row, etc.). If unknown, assign NaN value to first element in the array.</field>
+      <field type="float[45]" name="covariance" invalid="NaN">Row-major representation of position, velocity and acceleration 9x9 cross-covariance matrix upper right triangle (states: x, y, z, vx, vy, vz, ax, ay, az; first nine entries are the first ROW, next eight entries are the second row, etc.). If unknown, assign NaN value to first element in the array.</field>
```
hamishwillee, line 4796: Technically 

> assign NaN value to *first element* in the array.

It matters because presumably if all values are set as NaN then the message can't be zero truncated.

Not sure how we would represent this

message_definitions/v1.0/common.xml
```diff
@@ -4905,10 +4905,10 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the mission item.</field>
       <field type="uint8_t" name="current">Not used.</field>
       <field type="uint8_t" name="autocontinue">Not used (set 0).</field>
-      <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
-      <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
-      <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
-      <field type="float" name="param4">PARAM4, see MAV_CMD enum</field>
+      <field type="float" name="param1" invalid="NaN">PARAM1, see MAV_CMD enum</field>
```
hamishwillee, line 4908: Isn't validity dependent on the command that is actually sent for these?

message_definitions/v1.0/common.xml
olliw42: seems we were both typing at the same time:
see my 3rd bullet in https://github.com/mavlink/mavlink/pull/1614#issuecomment-819210119

message_definitions/v1.0/common.xml
```diff
@@ -4919,20 +4919,20 @@
...
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID (of command to send).</field>
       <field type="uint8_t" name="confirmation">0: First transmission of this command. 1-255: Confirmation transmissions (e.g. for kill command)</field>
-      <field type="float" name="param1">Parameter 1 (for the specific command).</field>
-      <field type="float" name="param2">Parameter 2 (for the specific command).</field>
-      <field type="float" name="param3">Parameter 3 (for the specific command).</field>
-      <field type="float" name="param4">Parameter 4 (for the specific command).</field>
-      <field type="float" name="param5">Parameter 5 (for the specific command).</field>
-      <field type="float" name="param6">Parameter 6 (for the specific command).</field>
-      <field type="float" name="param7">Parameter 7 (for the specific command).</field>
+      <field type="float" name="param1" invalid="NaN">Parameter 1 (for the specific command).</field>
```
hamishwillee, line 4922: Ditto - doesn't valid value depend on the command that is sent?

message_definitions/v1.0/common.xml
olliw42: seems we were both typing at the same time:
see my 2nd bullet in https://github.com/mavlink/mavlink/pull/1614#issuecomment-819210119

message_definitions/v1.0/common.xml
```diff
@@ -5143,7 +5143,7 @@
       <field type="float" name="pitch" units="rad">Pitch angle</field>
       <field type="float" name="yaw" units="rad">Yaw angle</field>
       <extensions/>
-      <field type="float[21]" name="covariance">Row-major representation of pose 6x6 cross-covariance matrix upper right triangle (states: x_global, y_global, z_global, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
+      <field type="float[21]" name="covariance" invalid="NaN">Row-major representation of pose 6x6 cross-covariance matrix upper right triangle (states: x_global, y_global, z_global, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
```
hamishwillee, line 5146: As above, how should invalid values be applied to array? (ie only first value is set)

message_definitions/v1.0/common.xml
```diff
@@ -5156,7 +5156,7 @@
       <field type="float" name="pitch" units="rad">Pitch angle</field>
       <field type="float" name="yaw" units="rad">Yaw angle</field>
       <extensions/>
-      <field type="float[21]" name="covariance">Row-major representation of pose 6x6 cross-covariance matrix upper right triangle (states: x, y, z, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
+      <field type="float[21]" name="covariance" invalid="NaN">Row-major representation of pose 6x6 cross-covariance matrix upper right triangle (states: x, y, z, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
```
hamishwillee, line 5159: As above - does this mean all values are NaN or just the first one?

message_definitions/v1.0/common.xml
```diff
@@ -5166,7 +5166,7 @@
       <field type="float" name="y" units="m/s">Global Y speed</field>
       <field type="float" name="z" units="m/s">Global Z speed</field>
       <extensions/>
-      <field type="float[9]" name="covariance">Row-major representation of 3x3 linear velocity covariance matrix (states: vx, vy, vz; 1st three entries - 1st row, etc.). If unknown, assign NaN value to first element in the array.</field>
+      <field type="float[9]" name="covariance" invalid="NaN">Row-major representation of 3x3 linear velocity covariance matrix (states: vx, vy, vz; 1st three entries - 1st row, etc.). If unknown, assign NaN value to first element in the array.</field>
```
hamishwillee, line 5169: As for https://github.com/mavlink/mavlink/pull/1614/files#r612921685

message_definitions/v1.0/common.xml
```diff
@@ -5179,7 +5179,7 @@
       <field type="float" name="pitch" units="rad">Pitch angle</field>
       <field type="float" name="yaw" units="rad">Yaw angle</field>
       <extensions/>
-      <field type="float[21]" name="covariance">Row-major representation of 6x6 pose cross-covariance matrix upper right triangle (states: x, y, z, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
+      <field type="float[21]" name="covariance" invalid="NaN">Row-major representation of 6x6 pose cross-covariance matrix upper right triangle (states: x, y, z, roll, pitch, yaw; first six entries are the first ROW, next five entries are the second ROW, etc.). If unknown, assign NaN value to first element in the array.</field>
```
hamishwillee, line 5182: As for https://github.com/mavlink/mavlink/pull/1614/files#r612921685

message_definitions/v1.0/common.xml
```diff
@@ -5642,16 +5642,16 @@
...
+      <field type="int32_t" name="time_remaining" units="s" invalid="0">Remaining battery time, 0: autopilot does not provide remaining battery time estimate</field>
       <field type="uint8_t" name="charge_state" enum="MAV_BATTERY_CHARGE_STATE">State for extent of discharge, provided by autopilot for warning or external reactions</field>
-      <field type="uint16_t[4]" name="voltages_ext" units="mV">Battery voltages for cells 11 to 14. Cells above the valid cell count for this battery should have a value of 0, where zero indicates not supported (note, this is different than for the voltages field and allows empty byte truncation). If the measured value is 0 then 1 should be sent instead.</field>
+      <field type="uint16_t[4]" name="voltages_ext" units="mV" invalid="0">Battery voltages for cells 11 to 14. Cells above the valid cell count for this battery should have a value of 0, where zero indicates not supported (note, this is different than for the voltages field and allows empty byte truncation). If the measured value is 0 then 1 should be sent instead.</field>
```
hamishwillee, line 5654: As per https://github.com/mavlink/mavlink/pull/1614/files#r612921685 - should it be [0...] or similar?

message_definitions/v1.0/common.xml
```diff
@@ -6428,7 +6428,7 @@
...
       <field type="uint8_t" name="sensor_type" enum="MAV_DISTANCE_SENSOR">Class id of the distance sensor type.</field>
-      <field type="uint16_t[72]" name="distances" units="cm">Distance of obstacles around the vehicle with index 0 corresponding to north + angle_offset, unless otherwise specified in the frame. A value of 0 is valid and means that the obstacle is practically touching the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
+      <field type="uint16_t[72]" name="distances" units="cm" invalid="[NaN]">Distance of obstacles around the vehicle with index 0 corresponding to north + angle_offset, unless otherwise specified in the frame. A value of 0 is valid and means that the obstacle is practically touching the sensor. A value of max_distance +1 means no obstacle is present. A value of UINT16_MAX for unknown/not used. In a array element, one unit corresponds to 1cm.</field>
```
hamishwillee, line 6431: So I think you are using [NaN] to indicate all elements are `NaN`, and `NaN` to indicate "first value". Correct?

message_definitions/v1.0/common.xml
```diff
@@ -6462,28 +6462,28 @@
...
-      <field type="float[5]" name="vel_x" units="m/s">X-velocity of waypoint, set to NaN if not being used</field>
-      <field type="float[5]" name="vel_y" units="m/s">Y-velocity of waypoint, set to NaN if not being used</field>
-      <field type="float[5]" name="vel_z" units="m/s">Z-velocity of waypoint, set to NaN if not being used</field>
-      <field type="float[5]" name="acc_x" units="m/s/s">X-acceleration of waypoint, set to NaN if not being used</field>
-      <field type="float[5]" name="acc_y" units="m/s/s">Y-acceleration of waypoint, set to NaN if not being used</field>
-      <field type="float[5]" name="acc_z" units="m/s/s">Z-acceleration of waypoint, set to NaN if not being used</field>
-      <field type="float[5]" name="pos_yaw" units="rad">Yaw angle, set to NaN if not being used</field>
-      <field type="float[5]" name="vel_yaw" units="rad/s">Yaw rate, set to NaN if not being used</field>
-      <field type="uint16_t[5]" name="command" enum="MAV_CMD">Scheduled action for each waypoint, UINT16_MAX if not being used.</field>
+      <field type="float[5]" name="pos_x" units="m" invalid="NaN">X-coordinate of waypoint, set to NaN if not being used</field>
```
hamishwillee, line 6465: So this implies that we're just setting the first value to NaN, but I think this should probably be `[NaN]`

message_definitions/v1.0/common.xml
```diff
@@ -6633,23 +6633,23 @@
...
-      <field type="uint32_t[4]" name="storage_usage" units="MiB">Amount of used storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
-      <field type="uint32_t[4]" name="storage_total" units="MiB">Total amount of storage space on the component system. A value of UINT32_MAX implies the field is unused.</field>
+      <field type="uint8_t[8]" name="cpu_cores" invalid="UINT8_MAX">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
+      <field type="uint8_t[10]" name="cpu_combined" invalid="UINT8_MAX">Combined CPU usage as the last 10 slices of 100 MS (a histogram). This allows to identify spikes in load that max out the system, but only for a short amount of time. A value of UINT8_MAX implies the field is unused.</field>
+      <field type="uint8_t[4]" name="gpu_cores" invalid="UINT8_MAX">GPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
```
hamishwillee, line 6638: `[UINT8_MAX]`?

hamishwillee:
@olliw42 I updated and tests pass. I did a partial review - we need to decide how to handle arrays - ie "all values must be X" vs "first value must be X". I added a bunch of duplicate comments before I realised it is unhelpful :-)

Further. Once we have done this we should update the generators to handle auto-adding docs for invalid, and remove the comments where possible.

message_definitions/v1.0/common.xml
olliw42: "invalid" is named in the various messages in various ways: invalid, unknown, shall not be used, is not supplied, ...
It notably is not a default value ... but may mean that a default should be used instead

in the case of the COMMAND_INT/COMMAND_LONG messages my understanding is that "invalid" means what in the command descriptions is named as 'empty' or 'reserved' ... I believe both mean the same, and I also believe that what is meant in the first paragraph of https://mavlink.io/en/messages/common.html#mav_commands is what "invalid" stands for

olliw42:
many thx for looking critically at this

> we need to decide how to handle arrays

as regards for the arrays, please see my 2nd bullet in https://github.com/mavlink/mavlink/pull/1614#issuecomment-819210119 and suggestion of how to go along with this

for the moment I kind of used "value" to indicate it's the first array field and "[value]" to indicate it's the whole array, but also noticed while going along that it doesn't cover all existing cases. It may happen that for some arrays in some very few old-ish messages one can't provide an automated ignore.

> Further. Once we have done this we should update the generators to handle auto-adding docs for invalid

you mean the doc generators or the code generators?

if doc generators: yes, I think this could much streamline the docs

message_definitions/v1.0/common.xml
olliw42: see before

hamishwillee:
@olliw42 And now I see your comments - yes, typing at the same time. Let me respond.

> For some fields the invalid was not specified by an explicit number, like UINT8_MAX, but by an enum field. In that case I had used the enum field name and not the explicit number, like it is done in the description. I think this is reasonable.
Is that ok?

IMO for enums the concept of an invalid value is invalid and it should not be defined :-)

Why? Well we use invalid to indicate "we don't support this field or supply the information that this field describes". But for enums we insist that one of the defined values is used - so we're always supply something that explicitly means something, even if that meaning is "this message is invalid or applies to invalid data".

Could be wrong

> Some fields are arrays, and for arrays it is not soo obvious how to specify it's invalid state so it can be automated. That's because the procedures can be quite different for the various messages. For instance, for some it is specified that the first array field should be set to invalid for the whole array to be invalid, for others each array field can be set invalid, for others the whole array needs to be filled with the invalid values, for yet others the array needs to be filled with special values. So I did use some kind of meta format ... which might not yet be totally smooth.
So I suggest this: I'm going to use the invalid field in a LUA script generator for my Mavlink for OpenTx project, so I can use this as reference to workout out and test a scheme handling all cases. I thus would suggest that we don't bother with this for the matter of this PR, and that I come up with a proposal when I did what I just said in the previous sentence. Ok?

I think I'd rather work this out now please. I'm happy to hold off merging until we have something that we think will work. I think if you use `[Value]` that reasonably implies "all values" must be this. I'm also not _sure_ about the "first value" case but I lean towards `[Value,]`

Thoughts?

message_definitions/v1.0/common.xml
olliw42: I should add that it would be MOST important to get this correct for COMMAND_INT/COMMAND_LONG, since these two messages would be the primary candidates there it would be used in practice...

hamishwillee:
> you mean the doc generators or the code generators?

I mean both, but specifically the _code_ generators. 

If we have the attribute it is good to remove the description text about the invalid text from the definition (in the same way we removed units info when we added attributes). 

For the doc generator, we're fine if we remove the description text for invalid value and just add a column for the attribute. However for the code generator if we take away the description text the information is lost from the in-source documentation. 
So I was proposing that if the invalid attribute is present in an entity we auto add to the description something like "Use [value] to indicate the field is unsupported/not supplied"

olliw42:
>> For some fields the invalid was not specified by an explicit number, like UINT8_MAX, but by an enum field. In that case I had used the enum field name and not the explicit number, like it is done in the description. I think this is reasonable.
>>  Is that ok?
>
>IMO for enums the concept of an invalid value is invalid and it should not be defined :-)
>
>Why? Well we use invalid to indicate "we don't support this field or supply the information that this field describes". But for >enums we insist that one of the defined values is used - so we're always supply something that explicitly means something, >even if that meaning is "this message is invalid or applies to invalid data".
>
>Could be wrong

This happened only for the message AUTOPILOT_STATE_FOR_GIMBAL_DEVICE, where we have the enum entry MAV_LANDED_STATE_UNDEFINED which is used to indicate that the landed state is unknown. So, there is an enum entry existing which is explicitly meaning "undefined".

Would you mind to look at this? MAV_LANDED_STATE is probably a very old and much used value, and I have difficulties to crasp it.

olliw42:
> I think I'd rather work this out now please. I'm happy to hold off merging until we have something that we think will work.

ok, perfectly fine with me. As said I have a specific use case to work it out and test it, so I will do this and come up with a proposal :)

message_definitions/v1.0/common.xml
hamishwillee: We don't ever use `NaN` in a command to mean "this is a value". On the other hand when you send a command then a value of NaN can mean either "this param is not defined for this command" or "this value for this defined param is not supported for this command" - a subtle difference. Also a value of 0 might also mean "invalid" for the command.

I'm not sure. My feeling though is that a command or mission message parameter has no invalid value.

olliw42:
>>  you mean the doc generators or the code generators?
> I mean both, but specifically the code generators.
>
>If we have the attribute it is good to remove the description text about the invalid text from the definition (in the same way we >removed units info when we added attributes).
>
>For the doc generator, we're fine if we remove the description text for invalid value and just add a column for the attribute. >However for the code generator if we take away the description text the information is lost from the in-source documentation.
>So I was proposing that if the invalid attribute is present in an entity we auto add to the description something like "Use [value] >to indicate the field is unsupported/not supplied"

ah

I was thinking that adding the invalid attribute would be a no-op for the code generators ... 
I must however generally say that I'm not very eager to touch the pymavlink code generators anymore (I'm using my own since a while and have given up with them) ... I hope that's not a stopper

I am btw not sure one wants to remove all text pieces like "Use UINT8_MAX to blabla" (I believe that's what you have in mind). There are some cases there where are multiple special values, and it would be nicer if all cases are listed and mentioned. Also "invalid" can mean slightly different things for some messages, i.e., it can be useful to keep saying that invalid means to use a specific other default value or a certain behavior and so on. That is, what I'm trying to say is that there might be not a totally satisfying general systematic approach which is good for all message field descriptions ...

... so I'm not sure how useful in a systematic sense the invalid attribute is for the purpose of descriptions and documentation. There are just too many cases ... and this was actually the difficult thing I had to wrap my mind around when I was first thinking abut this, to properly classify and sort the many different cases. So, I think it's main use is really for what I outlined in the first post, namely for code generators allowing to provide simplified API ... also but not so much for documentation.

hamishwillee:
>> This happened only for the message AUTOPILOT_STATE_FOR_GIMBAL_DEVICE, where we have the enum entry MAV_LANDED_STATE_UNDEFINED which is used to indicate that the landed state is unknown. So, there is an enum entry existing which is explicitly meaning "undefined".

> Would you mind to look at this? MAV_LANDED_STATE is probably a very old and much used value, and I have difficulties to crasp it.

This is ambiguous  - it could mean "we don't support this information (ever)" or "we're in an indeterminate state right now - it's a bumpy landing". @julianoes Do you know? 

If the former, then the value indicates `invalid` and you could apply it in the invalid attribute.
It it means "we don't know the state right now" or worse, this is sent when we don't support landing info AND when we don't now the state, then we can't use it for invalid IMO.

Despite that I still don't think we "need" the attribute for enums because the values are supposed to self-document. So we have a way of finding out the invalid values because it must be one of the enum values.

hamishwillee:
Re https://github.com/mavlink/mavlink/pull/1614#issuecomment-819227126

Fair enough. I guess removing the comment in description is optional. Upshot, we don't delete until we have agreement on whether we can, and what needs to be updated. Might happen or not, but we can proceed because this is useful either way.

olliw42:
> Re [#1614 (comment)](https://github.com/mavlink/mavlink/pull/1614#issuecomment-819227126)
> 
> Fair enough. I guess removing the comment in description is optional. Upshot, we don't delete until we have agreement on whether we can, and what needs to be updated. Might happen or not, but we can proceed because this is useful either way.

+1

olliw42:
> Despite that I still don't think we "need" the attribute for enums because the values are supposed to self-document. So we have a way of finding out the invalid values because it must be one of the enum values.

this would imply that the code generators read the descriptions of the enums or have some other AI kind of logic to infer that a certain enum value means invalid ... so, technically you are correct, but only from a human's perspective but maybe not so much from a code generator's perspective ...

message_definitions/v1.0/common.xml
olliw42: first off, to not confuse ourselves, for COMMAND_INT/LONG any implementation code needs to be specific for each particular MAV_CMD_xxx ... that is by construction there can't be ever a code generator which from the xml for the message alone could do anything reasonable. That is, any implementer must go through each command by hand, and thus can implement whatever the implementor thinks has to be done, independent on whatever any code generator is doing.

Every MAV_CMD has two types of parameters:

(1) It has some meaning and thus it's values can have whatever meaning, or (2) it has no meaning i.e. is empty or reserved.

For all empty/reserved parameters some value must be set.

So, what one ends up are API calls like (very generic)

send_command_long(
  MAV_CMD_xxx,
  value1,
  value2,
  value3,
  NaN,
  NaN,
  NaN,
  value4);

You may want to go through e.g. a py library, and you will find lots of such kind of calls with lots of NaN just to fill up the parameters.

What the ignore attribute allows to achieve (for some languages, like lua or py, not e.g. C) are API calls like

send_command_long(
  MAV_CMD_xxx,
  param1 = value1,
  param2 = value2,
  param3 = value3,
  param7 = value4);

That is, the code generator could produce code which realizes that some params are not specified and would fill it automatically with NaN. 

If for a given MAV_CMD_xxx param1 has a meaning, the ignore attribute would never hurt since you would have to set it's value in any case. If it happens that the default value is '0' for said MAV_CMD_xxx, you would write

send_command_long(
  MAV_CMD_xxx,
  param1 = 0,
  param2 = value2,
  param3 = value3,
  param7 = value4);

Note that you still benefit from the invalid attribute, since you don't have to also specify the values of the empty/reserved params.

Note also that in the case that the default value would be NaN for said MAV_CMD_xxx, the implementor should still write 

send_command_long(
  MAV_CMD_xxx,
  param1 = NaN,
  param2 = value2,
  param3 = value3,
  param7 = value4);

for style and readability reasons, even though she/he technically also could drop giving also the first parameter (IMHO this should be considered somewhat dirty IMHO since as a matter of fact it is somewhat fo a misuse)

Hope this helps outlining the idea.

olliw42:
so, I have worked out the array thing ... it was much easier than I initially thought

I have also implemented and tested it in my test case, the Mavlink for OpenTx project, and it works out beautifully. Unfortunately it is not easy to share this with you since it uses mavlink as submodule which isn't updated with this ...

there are only three cases which do occur:
* fill first array element with given value to indicate the array field is invalid
* fill all array elements with given value to indicate the array field is invalid
* fill any of the array elements with given value to indicate the specific array element is invalid

1st case: It would be sufficient to just specify invalid="value" as I did before but I like hamishwillee's suggestion to use invalid="[value,]" as it may scale better to future extensions and also helps with validating the xml or capturing errors in a py parser (I found one bug that way).
=> invalid="[value,]"

2nd case: I kept what I was using before
=> invalid="[value]"

3rd: case: Gladly, it boils down to same, namely that all array elements should be set to the value given by invalid. Therefore we here too are fine with
=> invalid="[value]"

Thus, we just need two versions for the moment. I have adapted the PR to this.

Although not needed now, the notation should allow extensions, such as "[1.0,NAN,,NAN]" for a [4] array, and so on, if such a thing ever should be needed.

While going over all changes in common.xml again, I found some few bugs, there I was using wrong values for invalid. I've corrected them too.

There are some few cases which might be disputable, I think (hope) that I made the right choice, but you should have a look at them. I will list them in the next post however, for better clarity. 

I have NOT changed values like 255 or 65535 to UINT8_MAX and UINT16_MAX. I think this would be better done in a small extra PR. Reason is that like now it is quite straightforward to check if the correct invalid was used by simply reading the description.

From my perspective it looks fine now and meets what I had in mind, let's see what you have to say :).

olliw42:
just for the sake of having this information available, I'd like to list the "buzz" words in the message descriptions which lead to an invalid attribute (which in turn so to say may also given an indication of what "invalid" may mean). The "invalid" attribute can indicate

- inactive
- not set by autopilot
- unknown
- nonexistent
- not available
- not supported
- invalid
- unused
- ignore
- unset
- does not measure xxx
- does not estimate xxx
- no xxx exist
- not supplied

olliw42:
as said, I here like to list some special cases which I identified as not canonical:

```
<message id="20" name="PARAM_REQUEST_READ">
<field type="int16_t" name="param_index" invalid="-1">Parameter index. Send -1 to use the param ID field as identifier (else the param id will be ignored)</field>
```
Here, the value of -1 does not only indicate that the field itself is to be ignored, but also that another field should be considered instead. I nevertheless think it makes sense. 
This is similarly so also for PARAM_EXT_REQUEST_READ

```
<message id="70" name="RC_CHANNELS_OVERRIDE">
<extensions/>
<field type="uint16_t" name="chan9_raw" units="us" invalid="UINT16_MAX">RC channel 9 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
```
Here we have two possible values to indicate that the field should be ignored. Previously I have chosen UINT16_MAX to make it similar with the non-extension fields, but have changed this now to 0 to benefit from zero byte trimming.

```
<message id="75" name="COMMAND_INT">
<message id="76" name="COMMAND_LONG">
```
We had a discussion already on these in the above, which I'm not sure has been finally settled already, so I list them here.

```
<message id="114" name="HIL_OPTICAL_FLOW">
<field type="float" name="distance" units="m" invalid="-1.0">Distance to the center of the flow field. Positive value (including zero): distance known. Negative value: Unknown distance.</field>
```
Here any negative value can mean unknown, so I have selected some "randomly". Not sure if -NaN is something reasonable.

```
<message id="124" name="GPS2_RAW">
<field type="uint16_t" name="yaw" units="cdeg" invalid="0">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
```
I think I did this incorrectly previously, I think 0 is the correct choice.

```
<message id="147" name="BATTERY_STATUS">
<field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells 1 to 10 (see voltages_ext for cells 11-14). Cells in this field above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).</field>
```
That's really a tricky one, so I have not added a invalid. I think it could be set to "[UINT16_MAX]", but I also think you should carefully check.

```
<message id="286" name="AUTOPILOT_STATE_FOR_GIMBAL_DEVICE">
<field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE" invalid="MAV_LANDED_STATE_UNDEFINED">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
```
That's the only case there invalid is an enum entry. I'm fine with it, but we had some discussion about it before which I think has not yet been finally settled, hence listed here.

```
<message id="390" name="ONBOARD_COMPUTER_STATUS">
<field type="uint8_t[8]" name="cpu_cores" invalid="[UINT8_MAX]">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
```
The description is dodgy, as the text indicate case 2 from above, while the intended use case seems to obviously be case 3 from above. This is so also for some other messages of this kind.

hamishwillee:
@olliw42 I'm finished for the day/week and my brain is toast. 

But my first impression is "excellent". I think as a post process I will try and make the text/list in https://github.com/mavlink/mavlink/pull/1614#issuecomment-820264838 consistent such that it always uses the same text for the same thing. In almost every case that should be "the sender does not support sending values for this this field". In some cases it might be the value is unknown right now OR the sender does not support sending these values.

I'm open to some common wording suggestions. I somewhat like _NaN for value unknown/not supported._

olliw42:
>  I'm finished for the day/week and my brain is toast.

:) 
so give you some rest, we want you fresh and strong

and there is no ultra urgency in this

> I think as a post process I will try and make the text/list in #1614 (comment) consistent such that it always uses the same text for the same thing. In almost every case that should be "the sender does not support sending values for this this field". In some cases it might be the value is unknown right now OR the sender does not support sending these values.

yeah ... I had the very same feeling that the wording could be much streamlined ... but as you considered it a post process
I'm not sure one will get away with just one common wording, but two or three will probably do it ...

olliw42:
Just for curiosity, is there a remote chance to see some activity on this this week?

:)

amilcarlucas:
I think this is a good addition. Would be nice to solve the conflict so that I can backport it to ardupilot's repo to see what @tridge thinks about this.

olliw42:
> Would be nice to solve the conflict

which conflict?

amilcarlucas:
Git hab says:

This branch has conflicts that must be resolved
Use the web editor or the to resolve conflicts.
Conflicting files
message_definitions/v1.0/common.xml

olliw42:
thx
I was too slow for you :)

olliw42:
conflict appears to be resolved :)

message_definitions/v1.0/common.xml
hamishwillee: Hmm. I guess what was in my mind is _by default_ parameters have a default value of `0`. This can be explicitly over-ridden to `NaN` using the `default` attribute: https://mavlink.io/en/guide/define_xml_element.html#reserved
This is needed because the "not supported" value needs to be agreed up front if you want to later reuse a reserved param for something.

Further, while we don't do so now, I'd also been thinking that someone might make a helper library for sending commands, seeded with the param defaults/invalid values. I guess though that wouldn't change anything because you'd pass in the defaults to your command sending functions "as though an implementer".

Upshot, I think I agree with you :-). Does my note here change anything for you?

hamishwillee: Actually `default` has a different meaning, but it would be  used for much the same purpose - ie. normally the default would be whatever value means "not supported".

message_definitions/v1.0/common.xml
```diff
@@ -4840,35 +4840,35 @@
...
-      <field type="uint16_t" name="chan15_raw" units="us">RC channel 15 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
-      <field type="uint16_t" name="chan16_raw" units="us">RC channel 16 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
-      <field type="uint16_t" name="chan17_raw" units="us">RC channel 17 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
-      <field type="uint16_t" name="chan18_raw" units="us">RC channel 18 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
+      <field type="uint16_t" name="chan9_raw" units="us" invalid="0">RC channel 9 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
```
hamishwillee, line 4862: Hmm "A value of 0 **or** UINT16_MAX". What you have done here is good.

message_definitions/v1.0/common.xml
```diff
@@ -5322,7 +5322,7 @@
       <field type="int16_t" name="temperature" units="cdegC">Temperature</field>
       <field type="uint8_t" name="quality">Optical flow quality / confidence. 0: no valid flow, 255: maximum quality</field>
       <field type="uint32_t" name="time_delta_distance_us" units="us">Time since the distance was sampled.</field>
-      <field type="float" name="distance" units="m">Distance to the center of the flow field. Positive value (including zero): distance known. Negative value: Unknown distance.</field>
+      <field type="float" name="distance" units="m" invalid="-1.0">Distance to the center of the flow field. Positive value (including zero): distance known. Negative value: Unknown distance.</field>
```
hamishwillee, line 5325: Hmmm. This is good. My OCD wants to change the description to be explicitly -1 as well, but we can't.

hamishwillee:
With respect to https://github.com/mavlink/mavlink/pull/1614#issuecomment-820260773 - thanks - works for me!

- As I see it  only two cases that occur: `[val]` (invalid is all elements set to val) or `[val,]` first element must be set to val. 
- I don't think we should allow "any element may be val" because that would imply software has to check every element just in case one of them is val. Easier to just define first element or all elements must be val. 
- I like that we can reasonably extend `[val,]` to allow us to specify different values should we so choose (though I don't want to "so choose".

hamishwillee:
With respect to https://github.com/mavlink/mavlink/pull/1614#issuecomment-820276363 (those which are not canonical)

```
<message id="20" name="PARAM_REQUEST_READ">
<field type="int16_t" name="param_index" invalid="-1">Parameter index. Send -1 to use the param ID field as identifier (else the param id will be ignored)</field>
```
> Here, the value of -1 does not only indicate that the field itself is to be ignored, but also that another field should be considered instead. I nevertheless think it makes sense. 

I agree. 

>This is similarly so also for PARAM_EXT_REQUEST_READ

I agree.

```
<message id="70" name="RC_CHANNELS_OVERRIDE">
<extensions/>
<field type="uint16_t" name="chan9_raw" units="us" invalid="UINT16_MAX">RC channel 9 value. A value of 0 or UINT16_MAX means to ignore this field. A value of UINT16_MAX-1 means to release this channel back to the RC radio.</field>
```
> Here we have two possible values to indicate that the field should be ignored. Previously I have chosen UINT16_MAX to make it similar with the non-extension fields, but have changed this now to 0 to benefit from zero byte trimming.

Agree - 0 is better. 

```
<message id="75" name="COMMAND_INT">
<message id="76" name="COMMAND_LONG">
```
> We had a discussion already on these in the above, which I'm not sure has been finally settled already, so I list them here.

Yes, we are still in discussion.

```
<message id="114" name="HIL_OPTICAL_FLOW">
<field type="float" name="distance" units="m" invalid="-1.0">Distance to the center of the flow field. Positive value (including zero): distance known. Negative value: Unknown distance.</field>
```
> Here any negative value can mean unknown, so I have selected some "randomly". Not sure if -NaN is something reasonable.

Agree. I think you have made the best of a bad design. This should have been just NaN, or a specific value.

```
<message id="124" name="GPS2_RAW">
<field type="uint16_t" name="yaw" units="cdeg" invalid="0">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
```
> I think I did this incorrectly previously, I think 0 is the correct choice.

Agree - 0 is the correct choice. 

As an aside I like that this field allows differentiation of  "we don't support sending out this info" and the case "we send out this info but we we don't have valid data to send right now". Generally we use invalid to mean the former and it _can_ work for either case - but it does mean that there is no way for a system to generally recognise that perhaps data is invalid now but might not be valid later.

```
<message id="147" name="BATTERY_STATUS">
<field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells 1 to 10 (see voltages_ext for cells 11-14). Cells in this field above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).</field>
```
> That's really a tricky one, so I have not added a invalid. I think it could be set to "[UINT16_MAX]", but I also think you should carefully check.

`[UINT16_MAX]` seems right. There is no case where this value would be indicate valid cell value.

```
<message id="286" name="AUTOPILOT_STATE_FOR_GIMBAL_DEVICE">
<field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE" invalid="MAV_LANDED_STATE_UNDEFINED">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
```
>That's the only case there invalid is an enum entry. I'm fine with it, but we had some discussion about it before which I think has not yet been finally settled, hence listed here.

I am happy for you to either not define this case or to use  `MAV_LANDED_STATE_UNDEFINED`. This is yet another case where it might be possible that invalid means we aren't ever going to send the data, or we just don't know what the data is right now. I think generally the invalid case is intended for the first case. The reason I accept this is that 0 is always used as a default for an enum if at all possible. 
 
```
<message id="390" name="ONBOARD_COMPUTER_STATUS">
<field type="uint8_t[8]" name="cpu_cores" invalid="[UINT8_MAX]">CPU usage on the component in percent (100 - idle). A value of UINT8_MAX implies the field is unused.</field>
```
> The description is dodgy, as the text indicate case 2 from above, while the intended use case seems to obviously be case 3 from above. This is so also for some other messages of this kind.

Note sure what you mean but yes, it is dodgy - I would expect as below - is that what you meant?
```
<field type="uint8_t[8]" name="cpu_cores" units="%" invalid="[UINT8_MAX]">CPU usage on the component in percent (0-100). UINT8_MAX: information for core not supported (supplied).</field>
```

hamishwillee:
OK, so I have been through this and it looks great (modulo the last post above). Note that I am happy to omit an invalid tag for a case or two in question to get the rest in.

My only concern is this confusion about what invalid means, and whether it is possible **or desirable**  to constrain it right now. What I want it to mean is, canonically, "this field is not supported for sending by this system" - ie you can check this once and never need to check it again. As opposed to potentially meaning "the data you are getting for this field is not valid right now". 

@julianoes Can you please sanity check.

olliw42:
many thx for all the review, this was a lot
I will now go carefully through what is left over, but want to directly comment to 

> My only concern is this confusion about what invalid means, and whether it is possible or desirable to constrain it right now. What I want it to mean is, canonically, "this field is not supported for sending by this system" - ie you can check this once and never need to check it again. As opposed to potentially meaning "the data you are getting for this field is not valid right now".

I here would disagree, I think one does not want to separate these two cases out into two, and in fact rather wants it to be like it is now. I see various reasons:

* logically: If it would be separated into two cases one also would have to assign two different values to indicate the two different cases, which would really make it clumsy for many messages. While you could argue for ints that we have enough of them at our disposal, like UNIT16_MAX, UNIT16_MAX-1, UNIT16_MAX-2, ... it is far less obvious so for float fields ... do we want to use now NaN, Inf, and when?

* conceptually: For many messages it would be 1&2. For instance consider a gimbal message where you now want to control it with attitude and later with velocity and then yet later again with attitude, and so on. So, it would is not only logically required to introduce two values but also factually.
 
* practically: I can't see a case where it would ever be an issue with not being able to distinguish. To me it seems that for each message it was absolutely clear so far what this special value was supposed to mean and what one is supposed to do and code (except of course the possibility that for few messages the docu might not be perfect, but that's not a fault of that special value). I hence that what has worked for until now also should work fine for until later.

* code wise: There is really no difference between the two since either way you have to do an if () in order to jump into the right action. One actually might argue that code-wise it is more convenient to not have two (I went through some codes in ArduPilot and me and it looks to me that two values would rather make coding more ugly).

So, I definitely would go with one special value for both cases and not two special values.

The only thing which indeed is "confusing" is that the word of the attribute itself doesn't directly reveal what it means, but that one needs to understand that it can mean various and that one needs to read the description in order to figure out what it actually is supposed to mean. But that's a limitation of language.

As I mentioned somewhere up it was actually the part I found most "difficult" to get along with the language thing and it took me a while to condense it down and come up with "invalid" as a summary of the cases I had listed in https://github.com/mavlink/mavlink/pull/1614#issuecomment-820264838. But note, the very same language ambiguity exists e.g. also for "not supported", "unknown", "not available", "not supplied", and so on ... in each case one wonders momentarily or never?

my view on this :)

hamishwillee:
> "invalid" I'm of course most happy to change 

Invalid it pretty good general word - it clearly means "this is not the data you were hoping for", without implying whether we don't support the data type or we just aren't sending it yet. I think "unknown" does the same thing. Either way, unless we refine the meaning this is a good choice of name.

If we were to change this to unsupported we would be implying this is a field that will never contain the data you are looking for.

Let me see what others think. I lean towards your view.

olliw42:
I went now through your responses.

* As much as I can see only one of them needed additional change, namely 

```
<message id="147" name="BATTERY_STATUS">
<field type="uint16_t[10]" name="voltages" units="mV">Battery voltage of cells 1 to 10 (see voltages_ext for cells 11-14). Cells in this field above the valid cell count for this battery should have the UINT16_MAX value. If individual cell voltages are unknown or not measured for this battery, then the overall battery voltage should be filled in cell 0, with all others set to UINT16_MAX. If the voltage of the battery is greater than (UINT16_MAX - 1), then cell 0 should be set to (UINT16_MAX - 1), and cell 1 to the remaining voltage. This can be extended to multiple cells if the total voltage is greater than 2 * (UINT16_MAX - 1).</field>
```
>>That's really a tricky one, so I have not added a invalid. I think it could be set to "[UINT16_MAX]", but I also think you should carefully check.

>[UINT16_MAX] seems right. There is no case where this value would be indicate valid cell value.

So I've added this. Thx for the advice.

* One point seems to be still open, namely

```
<message id="75" name="COMMAND_INT">
<message id="76" name="COMMAND_LONG">
```

>> We had a discussion already on these in the above, which I'm not sure has been finally settled already, so I list them here.

>Yes, we are still in discussion.

Besides the last bullet, all else seems to be settled to me. I'd like to add some irrelevant comments to few points:

> * I don't think we should allow "any element may be val" because that would imply software has to check every element just in case one of them is val. Easier to just define first element or all elements must be val.

Just for clarity, some messages require exactly this, so it's a valid case. Examples would be the voltage fields in BATTERY_STATUS, but also the cpu_cores field in ONBOARD_COMPUTER_STATUS. I.e., arrays exist and are in use there individual fields can be invalid, and the code accordingly has to check each field individually. I don't see any issue with this however, it's just a matter of documentation and not of principle.

Since you didn't disagree with keeping `invalid="MAV_LANDED_STATE_UNDEFINED"`, I did so :)

The description for the cpu_cores field in ONBOARD_COMPUTER_STATUS I found dodgy since the language doesn't exactly match what common sense suggests, namely that each array field can be set invalid or not. But that's just a matter of improving description, which shouldn't be part of this PR.

message_definitions/v1.0/common.xml
```diff
@@ -5264,11 +5264,11 @@
     </message>
     <message id="109" name="RADIO_STATUS">
       <description>Status generated by radio and injected into MAVLink stream.</description>
-      <field type="uint8_t" name="rssi">Local (message sender) recieved signal strength indication in device-dependent units/scale. Values: [0-254], 255: invalid/unknown.</field>
-      <field type="uint8_t" name="remrssi">Remote (message receiver) signal strength indication in device-dependent units/scale. Values: [0-254], 255: invalid/unknown.</field>
+      <field type="uint8_t" name="rssi" invalid="255">Local (message sender) recieved signal strength indication in device-dependent units/scale. Values: [0-254], 255: invalid/unknown.</field>
```
julianoes, line 5267: Couldn't all `255` be `UINT8_MAX`?

message_definitions/v1.0/common.xml
```diff
@@ -5297,14 +5297,14 @@
...
+      <field type="uint16_t" name="epv" invalid="UINT16_MAX">GPS VDOP vertical dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="vel" units="cm/s" invalid="65535">GPS ground speed. If unknown, set to: 65535</field>
       <field type="int16_t" name="vn" units="cm/s">GPS velocity in north direction in earth-fixed NED frame</field>
       <field type="int16_t" name="ve" units="cm/s">GPS velocity in east direction in earth-fixed NED frame</field>
       <field type="int16_t" name="vd" units="cm/s">GPS velocity in down direction in earth-fixed NED frame</field>
-      <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement), 0.0..359.99 degrees. If unknown, set to: 65535</field>
-      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint16_t" name="cog" units="cdeg" invalid="65535">Course over ground (NOT heading, but direction of movement), 0.0..359.99 degrees. If unknown, set to: 65535</field>
```
julianoes, line 5306: That's `UINT16_MAX`, right?

message_definitions/v1.0/common.xml
```diff
@@ -5356,7 +5356,7 @@
       <field type="int16_t" name="ymag" units="mgauss">Y Magnetic field</field>
       <field type="int16_t" name="zmag" units="mgauss">Z Magnetic field</field>
       <extensions/>
-      <field type="int16_t" name="temperature" units="cdegC">Temperature, 0: IMU does not provide temperature values. If the IMU is at 0C it must send 1 (0.01C).</field>
+      <field type="int16_t" name="temperature" units="cdegC" invalid="0">Temperature, 0: IMU does not provide temperature values. If the IMU is at 0C it must send 1 (0.01C).</field>
```
julianoes, line 5359: These crack me up :smile:. But all correct.

message_definitions/v1.0/common.xml
```diff
@@ -5617,10 +5617,10 @@
       <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
       <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
       <field type="float" name="alt" units="m">Altitude (MSL)</field>
-      <field type="float[3]" name="vel" units="m/s">target velocity (0,0,0) for unknown</field>
-      <field type="float[3]" name="acc" units="m/s/s">linear target acceleration (0,0,0) for unknown</field>
-      <field type="float[4]" name="attitude_q">(1 0 0 0 for unknown)</field>
-      <field type="float[3]" name="rates">(0 0 0 for unknown)</field>
+      <field type="float[3]" name="vel" units="m/s" invalid="[0]">target velocity (0,0,0) for unknown</field>
+      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">linear target acceleration (0,0,0) for unknown</field>
+      <field type="float[4]" name="attitude_q" invalid="[0]">(0 0 0 0 for unknown)</field>
```
julianoes, line 5622: Is that change on purpose? Before it was `1 0 0 0` for unknown.

message_definitions/v1.0/common.xml
```diff
@@ -6723,15 +6723,15 @@
...
-      <field type="int32_t" name="longitude" units="degE7">Current longitude of the unmanned aircraft. If unknown: 0 (both Lat/Lon).</field>
-      <field type="float" name="altitude_barometric" units="m">The altitude calculated from the barometric pressue. Reference is against 29.92inHg or 1013.2mb. If unknown: -1000 m.</field>
-      <field type="float" name="altitude_geodetic" units="m">The geodetic altitude as defined by WGS84. If unknown: -1000 m.</field>
+      <field type="int32_t" name="latitude" units="degE7" invalid="0">Current latitude of the unmanned aircraft. If unknown: 0 (both Lat/Lon).</field>
+      <field type="int32_t" name="longitude" units="degE7" invalid="0">Current longitude of the unmanned aircraft. If unknown: 0 (both Lat/Lon).</field>
+      <field type="float" name="altitude_barometric" units="m" invalid="-1000">The altitude calculated from the barometric pressue. Reference is against 29.92inHg or 1013.2mb. If unknown: -1000 m.</field>
+      <field type="float" name="altitude_geodetic" units="m" invalid="-1000">The geodetic altitude as defined by WGS84. If unknown: -1000 m.</field>
```
julianoes, line 6732: Wow, we're not consistent :facepalm:.

julianoes (commented):
Thanks for all that work @olliw42!

Couple of small comments for improvements. Didn't find any mistake.

message_definitions/v1.0/common.xml
olliw42: yes. I however didn't wanted to do this also in this PR to not get it to confusing, it find it is easier to see what's up. So I suggested to do it later in a separate PR.

message_definitions/v1.0/common.xml
olliw42: dito

message_definitions/v1.0/common.xml
olliw42: yeah, this must be a very very old message ... probably before I was born ...

message_definitions/v1.0/common.xml
olliw42: oohhaa ... THX for spotting this ! 

it seems to have slipped my attention and I certainly should have had added it to my list of "todos" in https://github.com/mavlink/mavlink/pull/1614#issuecomment-820276363 .... my applogies

you are right, this would be a change to the definition 

I can't retrace exactly how it came in, but I find it highly likely that I did it at a very early stage, i.e., the answer would be: yes on purpose.

Anyway, I believe that it is meanigfull, 1,0,0,0 is not really an impossible attitude ... so I guess when I did that I figured this is rather a bug or flaw than a feature. But I should have highlighted this change !

message_definitions/v1.0/common.xml
julianoes: Haha!

message_definitions/v1.0/common.xml
julianoes: That's true. 1, 0, 0, 0 is really just the default. So 0, 0, 0, 0 does make more sense to mark invalid.

message_definitions/v1.0/common.xml
julianoes: Ah, ok makes sense.

julianoes (approved):

hamishwillee:
@amilcarlucas @peterbarker FYI only, this adds `invalid` attribute to all fields that already specify a value that means "invalid" (i.e. data can't be sent). This was already discussed as part of the PR that pushed the invalid attribute into the XSD.

A couple of main things to note:
- The syntax for array values being invalid. `[value]` indicates that `value` is used to indicate each element of the array that is invalid, while `[value,]` indicates that the whole array is invalid if the first element is value.
- There are some values that are numerical 255 say, when we mean UINT_MAX8 (or whatever). These reflect what the docs currently say in the description. We would fix up both the docs and invalid value as a post process.

Thanks for following all this up @olliw42 , and for reviewing @julianoes

PR #1616, opened 2021-04-13
Update pymavlink to @ 4dac715

olliw42:
this is for https://github.com/mavlink/mavlink/pull/1614

I've separated this update out into an extra PR, since I'm not so sure I do the things correct with the git submodules... (so it's a bit of a learning PR for me)


hamishwillee:
Looks good to me. Thanks!

PR #1618, opened 2021-04-14
common.xml: add fields to smart_battery_info

hendjoshsr71:
This adds a few fields to the smart battery info. message that characterize the pack.

- `num_cells`
- `charging_maximum_voltage`
- `discharge_maximum_current`
- `discharge_burst_current`
- `manufacture_date`

Downstream PR https://github.com/ArduPilot/mavlink/pull/181

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6601,10 +6601,7 @@
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
-    <!-- Smart battery messages -->
     <message id="370" name="SMART_BATTERY_INFO">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Smart Battery information (static/infrequent update). Use for updates from: smart battery to flight stack, flight stack to GCS. Use BATTERY_STATUS for smart battery frequent updates.</description>
       <field type="uint8_t" name="id" instance="true">Battery ID</field>
       <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
@@ -6613,11 +6610,17 @@
       <field type="int32_t" name="capacity_full" units="mAh" invalid="-1">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
       <field type="uint16_t" name="cycle_count" invalid="UINT16_MAX">Charge/discharge cycle count. UINT16_MAX: field not provided.</field>
       <field type="char[16]" name="serial_number" invalid="[0]">Serial number in ASCII characters, 0 terminated. All 0: field not provided.</field>
-      <field type="char[50]" name="device_name">Static device name. Encode as manufacturer and product names separated using an underscore.</field>
+      <field type="char[50]" name="device_name" invalid="[0]">Static device name in ASCII characters, 0 terminated. All 0: field not provided. Encode as manufacturer name then product name separated using an underscore.</field>
       <field type="uint16_t" name="weight" units="g" invalid="0">Battery weight. 0: field not provided.</field>
       <field type="uint16_t" name="discharge_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="charging_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when charging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="resting_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when resting. If not supplied set to UINT16_MAX value.</field>
+      <extensions/>
+      <field type="uint16_t" name="charging_maximum_voltage" units="mV" invalid="0">Maximum per-cell voltage when charged. 0: field not provided.</field>
+      <field type="uint8_t" name="cells_in_series" invalid="0">Number of battery cells in series. 0: field not provided.</field>
+      <field type="uint32_t" name="discharge_maximum_current" units="mA" invalid="0">Maximum pack discharge current. 0: field not provided.</field>
+      <field type="uint32_t" name="discharge_maximum_burst_current" units="mA" invalid="0">Maximum pack discharge burst current. 0: field not provided.</field>
+      <field type="char[11]" name="manufacture_date" invalid="[0]">Manufacture date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
     </message>
     <message id="373" name="GENERATOR_STATUS">
       <description>Telemetry of power generation system. Alternator or mechanical generator.</description>


message_definitions/v1.0/common.xml
```diff
@@ -6585,6 +6585,11 @@
       <field type="uint16_t" name="discharge_minimum_voltage" units="mV">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="charging_minimum_voltage" units="mV">Minimum per-cell voltage when charging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="resting_minimum_voltage" units="mV">Minimum per-cell voltage when resting. If not supplied set to UINT16_MAX value.</field>
+      <field type="uint8_t" name="num_cells"> Number of battery cells. 0: field not provided.</field>
+      <field type="uint32_t" name="charging_maximum_voltage" units="mV">Maximum pack voltage when charging. 0: field not provided.</field>
+      <field type="uint32_t" name="discharge_maximum_current" units="mA">Maximum pack discharge current. 0: field not provided.</field>
+      <field type="uint32_t" name="discharge_maximum_burst_current" units="mA">Maximum pack discharge burst current. 0: field not provided.</field>
+      <field type="char[10]" name="manufacture_date">Manufacture date in ASCII characters, 0 terminated. All 0: field not provided.</field>
```
hamishwillee, line 6592: What format? YYYYMMDD ? If so, why 10 chars?

message_definitions/v1.0/common.xml
```diff
@@ -6585,6 +6585,11 @@
       <field type="uint16_t" name="discharge_minimum_voltage" units="mV">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="charging_minimum_voltage" units="mV">Minimum per-cell voltage when charging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="resting_minimum_voltage" units="mV">Minimum per-cell voltage when resting. If not supplied set to UINT16_MAX value.</field>
+      <field type="uint8_t" name="num_cells"> Number of battery cells. 0: field not provided.</field>
```
hamishwillee, line 6588: This can be inferred from [BATTERY_STATUS.voltages](https://mavlink.io/en/messages/common.html#BATTERY_STATUS) and `voltages_ext`. Note I have no particular issue with including it as this is sent at low low rate.

message_definitions/v1.0/common.xml
```diff
@@ -6585,6 +6585,11 @@
...
+      <field type="uint32_t" name="charging_maximum_voltage" units="mV">Maximum pack voltage when charging. 0: field not provided.</field>
+      <field type="uint32_t" name="discharge_maximum_current" units="mA">Maximum pack discharge current. 0: field not provided.</field>
```
hamishwillee, line 6590: What is the maximum current we might reasonably see FMI? Would it ever be more than 65A?

hamishwillee:
These proposed fields for [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO) all seem sensible to me (though the cell number can be inferred, no point quibbling about a few bytes here an there for a low rate message).

My only concern is that this is a breaking change. Are there any implementations we need to inform or does this need to be done as an extension field? 

If there is an implementation we need to remove the WIP tag. 

@cmic0 @AlexKlimaj @LorenzMeier @auturgy @WickedShell Thoughts?

message_definitions/v1.0/common.xml
hendjoshsr71: There are batteries for commercial drone that are rated for 550A continuous draw. I'm sure someone with a big system could be pulling more than. (Not common of course.)

Yes, above 65A current draw is common. I thought uint32 was best for this as we would have to make it dA to be in a reasonable range.

message_definitions/v1.0/common.xml
hendjoshsr71: Originally, I had thought about DDMMMYYYY where month was JAN, FEB to be less ambiguous. But to limit implementation flash size I think something like the following might be better, "DD/MM/YYYY" ?

message_definitions/v1.0/common.xml
hendjoshsr71: This was requested to cover the major battery parameters in this without inference.

AlexKlimaj:
These changes look okay to me. So far I don't know who has implemented this message.

message_definitions/v1.0/common.xml
hamishwillee: Seems reasonable.

message_definitions/v1.0/common.xml
hamishwillee: The only requirement is that we make it unambiguous (i.e. document it). To me YYYYMMDD is the most appropriate format because it has no separators (compact) and does away with expectation of day/month order.

So something like

> Manufacture date as a zero-terminated string. Format is YYYYMMDD, using ASCII numeric characters (e.g. 20330225). All 0: field not provided.

hamishwillee:
Looks good to me other than comment on the date format. I'll get some feedback from MAVLink dev call this evening. Feel free to [attend](https://mavlink.io/en/about/support.html#schedule) (7pm in Melbourne/Canberra today).

message_definitions/v1.0/common.xml
hendjoshsr71: The reason I added the delimiters was to not need the additional string parsing code to ensure the leading zeros get filled. IE to ensure 2021121 becomes 20210121. It is only 72 bytes though to add that vs 2 bytes for delimiters.

hendjoshsr71:
Thanks for the invite! I would attend, but that is middle of the night here in the US :)

hamishwillee:
@hendjoshsr71 We didn't get around to this during the meeting sorry. I will seek more direct review. 

@cmic0 If you do not comment on this I will assume that it is OK to break this WIP message. 

@julianoes Do you have thoughts on these additions, and more specifically on https://github.com/mavlink/mavlink/pull/1618#discussion_r621790992 ?

message_definitions/v1.0/common.xml
julianoes: I'm confused by the fact that some of the voltage fields are per cell while other s like these ones are over the whole pack but I suppose it makes more sense for these that way?

Also, the other existing fields are `uint16_t` while these ones `uint32_t`.

julianoes (commented):
I can see that this message is already implemented in PX4. By omitting `</extensions>` we will break existing implementations, if they are out there.

hendjoshsr71:
The uint32s for the current has to do with the expected range if we keep mA as the base unit in keeping with mAh (0.001A to 65.536A). If dA was used, then uint16 would be fine range of (0.1A to 6553.6A). Same goes for volts as 65V isn't very high for future systems or large systems.

My thought for not doing per-cell volts was it can always be computed if the number of cells is available. Also I was thinking this message could be used for systems where there is a monitor for each pack and monitors on groups of packs or the entire pack. Then per-cell makes a little less sense. but not a lot. Possibly that would be better captured by some other future message.

hendjoshsr71:
Made these new fields extensions and switched to `Maximum per-cell voltage when charged` using an uint16_t

message_definitions/v1.0/common.xml
```diff
@@ -6585,6 +6582,12 @@
       <field type="uint16_t" name="discharge_minimum_voltage" units="mV">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="charging_minimum_voltage" units="mV">Minimum per-cell voltage when charging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="resting_minimum_voltage" units="mV">Minimum per-cell voltage when resting. If not supplied set to UINT16_MAX value.</field>
+      <extensions/>
+      <field type="char[11]" name="manufacture_date">Manufacture date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
```
julianoes, line 6586: I would move this to the end, so that it can be trimmed if it is not used.

julianoes (commented):
Thanks for the fixes. I have one more comment and then this is perfect.

message_definitions/v1.0/common.xml
hamishwillee: @julianoes Do you see @cmic0 ? I know he had some feedback. For example PX4 already has a manufacturer data + serial number defined here: https://github.com/PX4/PX4-Autopilot/blob/4a7b2c490a2fea9d2f0a9a5805a1dbea3c0a3562/src/modules/mavlink/streams/SMART_BATTERY_INFO.hpp#L86-L87 so we need to make sure there is a smooth migration path.

julianoes: I don't understand how this was working. Was this already in this mavlink message?

hendjoshsr71: Manufacture date was not already in this message. I've moved it to the end for trimming.

hendjoshsr71:
@WickedShell Do you have any comments?

message_definitions/v1.0/common.xml
hamishwillee: Not sure I agree with this, but not my call. If no one else complains works for me.

message_definitions/v1.0/common.xml
hamishwillee: @julianoes @hendjoshsr71 There is something very broken about this PR. I can see by syncing the head that the message already has some of these fields.
```
    <!-- Smart battery messages -->
    <message id="370" name="SMART_BATTERY_INFO">
      <wip/>
      <description>Smart Battery information (static/infrequent update). Use for updates from: smart battery to flight stack, flight stack to GCS. Use BATTERY_STATUS for smart battery frequent updates.</description>
      <field type="uint8_t" name="id" instance="true">Battery ID</field>
      <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
      <field type="uint8_t" name="type" enum="MAV_BATTERY_TYPE">Type (chemistry) of the battery</field>
      <field type="int32_t" name="capacity_full_specification" units="mAh" invalid="-1">Capacity when full according to manufacturer, -1: field not provided.</field>
      <field type="int32_t" name="capacity_full" units="mAh" invalid="-1">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
      <field type="uint16_t" name="cycle_count" invalid="UINT16_MAX">Charge/discharge cycle count. UINT16_MAX: field not provided.</field>
      <field type="char[16]" name="serial_number" invalid="[0]">Serial number in ASCII characters, 0 terminated. All 0: field not provided.</field>
      <field type="char[50]" name="device_name">Static device name. Encode as manufacturer and product names separated using an underscore.</field>
      <field type="uint16_t" name="weight" units="g" invalid="0">Battery weight. 0: field not provided.</field>
      <field type="uint16_t" name="discharge_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>
      <field type="uint16_t" name="charging_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when charging. If not supplied set to UINT16_MAX value.</field>
      <field type="uint16_t" name="resting_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when resting. If not supplied set to UINT16_MAX value.</field>
    </message>
```
I'll try capture this in a fresh one.

hamishwillee: > I don't understand how this was working. Was this already in this mavlink message?

@julianoes @hendjoshsr71 My note above is for the same reason as you. I can see fields used by PX4 that I can't see in the message.

This appears to be some problem with github (at first I thought this PR).

The file view of the PR is missing many of the fields that are in master
image

When I [view file on this file in this PR](https://github.com/mavlink/mavlink/blob/0b3f3f0855d4ac897546d63dd430664ebcc498e4/message_definitions/v1.0/common.xml) the file looks screwed up - it is truncated before  SMART_BATTERY_INFO.

Only when I view the [raw file](https://raw.githubusercontent.com/mavlink/mavlink/0b3f3f0855d4ac897546d63dd430664ebcc498e4/message_definitions/v1.0/common.xml) do I see the "missing fields".
image

When I created my own fresh version of this I get the same problem of fields not being shown - see https://github.com/mavlink/mavlink/compare/pr_smart_message_extensions

Any advice on how this might be fixed/tidied. It is impossible to properly review a PR that doesn't reflect the underlying data?

hamishwillee:
FYI, we just added invalid fields in master - would look like this:
```
      <extensions/>
      <field type="uint16_t" name="charging_maximum_voltage" units="mV" invalid="0">Maximum per-cell voltage when charged. 0: field not provided.</field>
      <field type="uint8_t" name="cells_in_series" invalid="0">Number of battery cells in series. 0: field not provided.</field>
      <field type="uint32_t" name="discharge_maximum_current" units="mA" invalid="0">Maximum pack discharge current. 0: field not provided.</field>
      <field type="uint32_t" name="discharge_maximum_burst_current" units="mA" invalid="0">Maximum pack discharge burst current. 0: field not provided.</field>
      <field type="char[11]" name="manufacture_date" invalid="[0]">Manufacture date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
```

hendjoshsr71:
Rebased and added the invalid fields.  I'm not sure what was up with github... weird. Hopefully, the push I just did fixes the issue.

message_definitions/v1.0/common.xml
hamishwillee: Fixed by https://github.com/mavlink/mavlink/pull/1618#issuecomment-833251435

hamishwillee:
Thanks @hendjoshsr71 - looks good to me. Thanks. @julianoes @cmic0 @auturgy Any further comments? Are you happy for this to be merged?

message_definitions/v1.0/common.xml
```diff
@@ -6613,11 +6610,17 @@
       <field type="int32_t" name="capacity_full" units="mAh" invalid="-1">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
       <field type="uint16_t" name="cycle_count" invalid="UINT16_MAX">Charge/discharge cycle count. UINT16_MAX: field not provided.</field>
       <field type="char[16]" name="serial_number" invalid="[0]">Serial number in ASCII characters, 0 terminated. All 0: field not provided.</field>
-      <field type="char[50]" name="device_name">Static device name. Encode as manufacturer and product names separated using an underscore.</field>
+      <field type="char[50]" name="device_name" invalid="[0]">Static device name. Encode as manufacturer and product names separated using an underscore.</field>
```
julianoes, line 6613: Can we agree that this one should also be 0 terminated?

julianoes (approved):
One question, otherwise all good!

message_definitions/v1.0/common.xml
hendjoshsr71: Certainly! Added that and clarified the wording for the encoding as well, `ManufacuterName_ProductName`.

julianoes (approved):

message_definitions/v1.0/common.xml
julianoes: Thanks @hendjoshsr71!

julianoes:
Thanks a lot @hendjoshsr71.

olliw42:
@hendjoshsr71 @hamishwillee 
just a comment please concerning the invalids here. they don't hurt but

* I think that for char arrays they do not make much sense since it's - AFAIK - a pretty general convention of MAVLink that char arrays are null terminated except when completely full, which implies that all null is empty. If one would want to add invalid to char arrays, which one could if one wants to, this would affect a number of other messages too which accordingly one also would want to change consistently.

* I also think that one rather should use [0,] instead of [0]. First, that's exactly what it was before, it was sufficient that only the first byte is null. And moreover it is much easier to test just one byte than all bytes. So, requiring that all bytes have to be null is a change in behavior and also a change in the wrong direction (towards worse)
:)

julianoes:
@olliw42 I agree with your second suggestion. I'd merge a PR for it.
For the first point, I'm not entirely sure what you mean. I just prefer strings to be null terminated because it makes it easier to handle on the other side. You can just set the last char to 0 and then safely use `strcpy` or `strlen` without having to make copies.
I guess you are saying nothing/zero is not invalid? And hence you would not mark it as invalid? Isn't it the same thing? :smile: Or what am I missing?

olliw42:
IMHO this char array is not different to e.g. the param_id char array or the one in the camera information and so on

the convention for these char arrays is that the entered string is null terminated, unless the string has exactly as many bytes as the length of the char array is. The C like null termination for shorter strings allows us to infer how much of the data comprises the information. 

=> this is not just a "prefer it", it is a "must do it" since otherwise one couldn't infer how many of the bytes comprise the information !!

For the  a param_id of "HELLOMYFRIEND" the char array would be filled as 

H, E, L, L, O, M, Y, F, R, I, E, N, D, must be \0, could be 0 but doesn't matter, could be 0 doesn't matter, could be 0 doesn't matter

If the char array would not hold any information it would be 

must be \0, 15 times could be 0 but doesn't matter ...

=>

1) it is sufficient to specify [0,], and not [0] (which you already agreed to)

2) invalid=[0,] is unambiguously implied by the fabric of the whole thing ... and it is thus superfluous to specify it

3) if one nevertheless would want to make it explicit and specify it, it should be done consistently so for all char arrays in all other messages too

:)

hendjoshsr71:
I can change it to whatever decision is made. I was following the example on the line above that was previously done. :)
`<field type="char[16]" name="serial_number" invalid="[0]"> ......`

I get the idea that it is implied. I will say for folks not as versed in MAVLINK, like me, being explicit is nice. Somewhere that says the general convention for char fields.
1. Encode in ASCII
2.  First value \0: field not provided
3. And the nice explanation below that is in the param fields where it should apply
    -  `Terminated by NULL if the length is less than 50 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 50 chars - applications have to provide 50+1 bytes storage if the ID is stored as string`
    - This changes my implementation a bit as I presumed char[49 + \0] was the norm

In my ideal there would a link in the generated documentation that links to this info. and it wouldn't need to be stated everywhere. And my quick look through common.xml is that many `char[]` do not specify any of this info. 

1. and 2. are good assumptions but 3. ....?

julianoes:
Some fields like param_name are allowed to be "filled" without zero termination. I would like to avoid creating more of these as they are harder to deal with. Often you end up doing a memcpy to a bigger field to make sure it's safe. I would like to avoid that (and instead transmit one more byte).

For the invalid tag, I leave that to @olliw42.

PR #1621, opened 2021-04-20
added bitmask

python36:


hamishwillee:
@amilcarlucas @peterbarker This is an ardupilotmega dialect only addition. It looks good to me. You OK to merge?

peterbarker:
@hamishwillee strong feelings from some DevTeam members that changes to `ardupilotmega.xml` should only ever come from https://github.com/ardupilot/mavlink - so could we get the equivalent patch there first, please?

hamishwillee:
@python36 Thanks for this. Can you please make these changes in https://github.com/ardupilot/mavlink - they will then be pushed here by the ArduPilot team. Which is not to say that the changes are not good - they appear to be.  But we use this process to ensure that changes to the dialect are only made/approved by the dialect's owners. 

Thanks @peterbarker .

python36:
@hamishwillee Ok. For the files `autoquad.xml` and` uAvionix.xml` make pull request to this repository or to ArduPilot too?

hamishwillee:
Not sure. @julianoes Do you happen to know who owns the dialects autoquad.xml and uAvionix.xm

peterbarker:
@magicrub knows something of uAvionix.xml and who should be consulted on poking it.

hamishwillee:
@python36 For uAvionix there are contact details in the header of the XML file https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/uAvionix.xml#L2

@mpaperno Owns the autoquad.xml. I'd accept a pull from https://github.com/AutoQuad/mavlink/
Max, are you still using this dialect? If so, can I please have a PR that adds contact details as per uAvionix above. If not, can you confirm this is no longer used and we will remove it

mpaperno:
@hamishwillee I went to update and submit a PR, but apparently someone has already been modifying our file, and in some cases breaking functionality.  For example, defined enum values?  Defeats the point of being able to extend enums.  (I haven't been keeping up with any changes, so I dunno, maybe that's somehow a good thing.)

Anyway, since the AQ project is in (possibly permanent) stasis, I'd say just delete the file. 

Thanks,
-Max

hamishwillee:
@mpaperno Thanks very much - I will delete the autoquad.xml file.

> For example, defined enum values? Defeats the point of being able to extend enums. (I haven't been keeping up with any changes, so I dunno, maybe that's somehow a good thing.)

Apologies if the changes to enum broke your files. 
The way enum value generation works is that the generators automatically assign an enum value if none has been assigned. If you explicitly assign the same value that would otherwise be generated (which is what we did) nothing can break.
Why did we do this? Well, if you extend your enum by putting a new value at the end nothing happens - you just get the new ID. However if you were to decide to extend it by putting your enum in the middle then you'd offset every single enum value after that point - fine if you're regenerating all the libraries used on all systems at the same time, but will break anything that uses a value that you didn't update. 

Upshot, it is always better to explicitly assign enum values, unless you know that you will always be updating all firmware that uses the libraries at the same time.

PR #1622, opened 2021-04-21
Add simple FSM support to ORBIT_EXECUTION_STATUS message in common.xml

tahsinkose:
This is a simple feature extension on orbit mode that allows the implementation of a very simple finite state machine structure in autopilot software. Normally orbit executes indefinitely, however the use-case I'm currently working on requires a finite orbit behavior for a certain amount of angles. 

Following is a diagram of the added states when the orbit is requested to happen only for 360 degrees.
OrbitFSM

Based on use-cases, this state machine can be populated with more fine-grained states. Nevertheless this alone enables many capabilities.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -704,6 +704,21 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
     </enum>
+    <enum name="ORBIT_EXECUTION_STATE">
+      <description>State of execution during orbit flight.</description>
+      <entry value="0" name="ORBIT_EXECUTION_STATE_UNKNOWN">
+        <description>Vehicle state is unknown according to orbit action.</description>
+      </entry>
+      <entry value="1" name="ORBIT_EXECUTION_STATE_APPROACH_TO_CIRCLE">
+        <description>Vehicle is approaching to orbit track.</description>
+      </entry>
+      <entry value="2" name="ORBIT_EXECUTION_STATE_STARTED">
+        <description>Vehicle started following orbit track.</description>
+      </entry>
+      <entry value="3" name="ORBIT_EXECUTION_STATE_FINISHED">
+        <description>Vehicle finished following orbit track.</description>
+      </entry>
+    </enum>
     <enum name="WIFI_CONFIG_AP_RESPONSE">
       <description>Possible responses from a WIFI_CONFIG_AP message.</description>
       <entry value="0" name="WIFI_CONFIG_AP_RESPONSE_UNDEFINED">
@@ -6572,6 +6587,7 @@
       <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
+      <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
     </message>
     <!-- Smart battery messages -->
     <message id="370" name="SMART_BATTERY_INFO">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -943,14 +943,14 @@
       <entry value="34" name="MAV_CMD_DO_ORBIT" hasLocation="true" isDestination="true">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Start orbiting on the circumference of a circle defined by the parameters. Setting any value NaN results in using defaults.</description>
-        <param index="1" label="Radius" units="m">Radius of the circle. positive: Orbit clockwise. negative: Orbit counter-clockwise.</param>
-        <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Vehicle configuration default.</param>
+        <description>Start orbiting on the circumference of a circle defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
+        <param index="1" label="Radius" units="m">Radius of the circle. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
+        <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
         <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
-        <param index="4" label="Orbits" units="rad" minValue="0" default="0">Orbit centre point for this many radians (i.e. for a three-quarter orbit set 270*Pi/180). 0: Orbit forever.</param>
-        <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
-        <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
-        <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
+        <param index="4" label="Orbits" units="rad" minValue="0" default="0">Orbit around the centre point for this many radians (i.e. for a three-quarter orbit set 270*Pi/180). 0: Orbit forever. NaN: Use vehicle default, or current value if already orbiting.</param>
+        <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
+        <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
+        <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -962,7 +962,7 @@
         <param index="1" label="Radius" units="m">Radius of the circle. positive: Orbit clockwise. negative: Orbit counter-clockwise.</param>
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Vehicle configuration default.</param>
         <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
-        <param index="4">Reserved (e.g. for dynamic center beacon options)</param>
+        <param index="4" label="Turn in Radians" units="rad" minValue="0" default="0">Orbital turn in radians.</param>
         <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle position or current center if already orbiting.</param>
@@ -6588,6 +6588,7 @@
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
       <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
+      <field type="float" name="progress">Progress of orbit in percentage, only meaningful if the orbit is finite.</field>
     </message>
     <!-- Smart battery messages -->
     <message id="370" name="SMART_BATTERY_INFO">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6588,7 +6588,7 @@
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
       <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
-      <field type="float" name="progress">Progress of orbit in percentage, only meaningful if the orbit is finite.</field>
+      <field type="uint8_t" units="%" name="progress">Progress of orbit. UINT_MAX8 if orbit is indefinite.</field>
     </message>
     <!-- Smart battery messages -->
     <message id="370" name="SMART_BATTERY_INFO">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -704,21 +704,6 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
     </enum>
-    <enum name="ORBIT_EXECUTION_STATE">
-      <description>State of execution during orbit flight.</description>
-      <entry value="0" name="ORBIT_EXECUTION_STATE_UNKNOWN">
-        <description>Vehicle state is unknown according to orbit action.</description>
-      </entry>
-      <entry value="1" name="ORBIT_EXECUTION_STATE_APPROACH_TO_CIRCLE">
-        <description>Vehicle is approaching to orbit track.</description>
-      </entry>
-      <entry value="2" name="ORBIT_EXECUTION_STATE_STARTED">
-        <description>Vehicle started following orbit track.</description>
-      </entry>
-      <entry value="3" name="ORBIT_EXECUTION_STATE_FINISHED">
-        <description>Vehicle finished following orbit track.</description>
-      </entry>
-    </enum>
     <enum name="WIFI_CONFIG_AP_RESPONSE">
       <description>Possible responses from a WIFI_CONFIG_AP message.</description>
       <entry value="0" name="WIFI_CONFIG_AP_RESPONSE_UNDEFINED">
@@ -6587,8 +6572,6 @@
       <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
-      <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
-      <field type="uint8_t" units="%" name="progress">Progress of orbit. UINT_MAX8 if orbit is indefinite.</field>
     </message>
     <!-- Smart battery messages -->
     <message id="370" name="SMART_BATTERY_INFO">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6572,6 +6584,7 @@
       <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
+      <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
```
hamishwillee, line 6587: This is in `WIFI_CONFIG_AP_RESPONSE_UNDEFINED`, which is definitely an error. Where did you want it to be?

message_definitions/v1.0/common.xml
```diff
@@ -704,6 +704,18 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
     </enum>
+    <enum name="ORBIT_EXECUTION_STATE">
+      <description>State of execution during orbit flight.</description>
+      <entry value="0" name="ORBIT_EXECUTION_STATE_APPROACH_TO_CIRCLE">
```
hamishwillee, line 709: The first value should probably be value ORBIT_EXECUTION_STATE_UNKNOWN, with value 0. Otherwise a system that gets 0 will not know if this value means the message is being sent from an old system that simply does not use the field

message_definitions/v1.0/common.xml
```diff
@@ -704,6 +704,18 @@
...
+        <description>Vehicle is approaching to orbit track.</description>
+      </entry>
+      <entry value="1" name="ORBIT_EXECUTION_STATE_STARTED">
+        <description>Vehicle started following orbit track.</description>
+      </entry>
+      <entry value="2" name="ORBIT_EXECUTION_STATE_FINISHED">
```
hamishwillee, line 715: So the ORBIT_EXECUTION_STATUS is only sent out while the orbit is in progress. So this will be sent out at most one time (and of course the recipient might miss it). Is the inference that ORBIT_EXECUTION_STATUS  will always be sent out at the point it leaves the orbit? IF so the protocol would need to state this.

hamishwillee:
This is addition of a state enum to [ORBIT_EXECUTION_STATUS](https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS) to indicate whether you're approaching or "in" the orbit, from the perspective of the flight stack. The use case is to make it easier for a ROS system (linked PR) to work out the point at which the orbit started, in order to choose an exit angle.

Any opinions? I'm not opposed in that a GCS can now also know the autopilot thinks it is circling, rather than just guessing based on it overlaying the planned path.

message_definitions/v1.0/common.xml
tahsinkose: Fair :+1:

message_definitions/v1.0/common.xml
tahsinkose: > So the ORBIT_EXECUTION_STATUS is only sent out while the orbit is in progress

Yes.

> So this will be sent out at most one time (and of course the recipient might miss it). Is the inference that ORBIT_EXECUTION_STATUS will always be sent out at the point it leaves the orbit?

A proper implementation should send telemetry of orbit status as long as the vehicle is conducting orbit action. Therefore it will not be sent out at most one time. The vehicle should start hovering on wherever it is once the orbiting action finishes, i.e. in `ORBIT_EXECUTION_STATE_FINISHED`, but orbit flight task would be still active. Between FINISHED and flight task update time window, telemetry is still kept published with the state info.

tahsinkose:
@hamishwillee any updates on this?

message_definitions/v1.0/common.xml
hamishwillee: @tahsinkose The problem is that there is no such thing/concept as the orbit "finishing" - at least not related to these messages or PX4 (and I am pretty sure ArduPilot too). The only concept I know in mavlink of an orbit finishing is the loiter command for a fixed wing vehicle, which will finish after either time or a certain number of turns. 

What we have is two WIP messages:
- [MAV_CMD_DO_ORBIT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_ORBIT) - This **starts** the orbit. There is no mechanism to stop the orbit. This does not mention [ORBIT_EXECUTION_STATUS](https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS)
- [ORBIT_EXECUTION_STATUS](https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS) - says it is emitted while orbit is happening.

I would expect `ORBIT_EXECUTION_STATUS` to stop being emitted when you stop orbiting. When that happens would depend on how the orbit is implemented in a particular flight stack - so if the flight stack supports the idea of loitering after one turn (say) then yes, perhaps what you are describing might happen. But that's not how [PX4 orbit mode](http://docs.px4.io/master/en/flight_modes/orbit.html) works (for example) and we need to try think of messages that are agnostic of a particular stack.

Thoughts?

hamishwillee:
@tahsinkose Sorry for delay - I typically address the queue on Weds/Thurs every week but am sporadic between. 

I added a comment and will discuss in the dev call (feel free to attend - it is in about 5 hours from now - see [dronecode calendar](https://mavlink.io/en/about/support.html#schedule)). 

We should probably remove the WIP on these messages and add info about ORBIT_EXECUTION_STATUS to the orbit command. I THINK this has to be an extension field too - depends on whether there is a plan in place to allow breaking the message?

message_definitions/v1.0/common.xml
tahsinkose: @hamishwillee PX4 orbit mode is a time-uncapped flight mode that would execute as long as autopilot is not requested to transition into another flight mode. It doesn't prevent a stateful finite orbit flight mode to be added such as `FlightTaskFiniteOrbit`. Actually this is the name of flight task I'm currently using in my project and this extension on message makes both flight tasks (`FlightTaskOrbit` and `FlightTaskFiniteOrbit`) gracefully work without any interference.

I can't see how these messages are not stack-agnostic. The particular flight modes that use this message who don't care about the state of orbit can simply neglect those fields.

tahsinkose:
@hamishwillee thanks for the due diligence. Unfortunately I can't commit to removing WIP on the messages yet (not before Sept 2021). Nevertheless this can wait until then - or someone else finishes WIP statuses on them and this would become ready for free.

hamishwillee:
Hi @tahsinkose ,

You could already achieve what (we think) you want using the existing messages. Simply set a target on the point on the radius that you want to enter, fly there, start orbit, then track the position on the radius. Then send new command when you reach the point on the radius you want to exit. This way you know the starting point and end point because you set rather than get them.

We had a look at the command and message in the dev call. The general feeling was that perhaps
- there may be better ways of achieving what you want than exposing this state. 
- the messages are a bit limited, and while this might address your case, they aren't a good generalization of the constraints that we might want to set on an orbit.
- We don't really understand your use case well enough to provide a good design. Can you please provide more background on the flight you envisage. Is this for use in a mission or is it command driven. How is it planned? Where would you be coming from or going to - does the path matter?

I don't want to go into much detail on command design without a better view on your use case, but our thinking was:
- Desirable to command orbit forever, for time, for number of orbits (in the same way as for plane loiter commands (
- [MAV_CMD_NAV_LOITER_UNLIM](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_UNLIM), [MAV_CMD_NAV_LOITER_TURNS](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TURNS), [MAV_CMD_NAV_LOITER_TIME](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TIME)). 
- We might want to specify that number of turns in radians to exit at a particular angle, as you need.
- We might further want to command the exit path https://mavlink.io/en/services/mission.html#mission-command-detail or an entry path.
- An orbit in a mission would progress to the next waypoint on completion. An orbit sent as a command might simply loiter on completion.  Which is where your state would potentially be useful to emit.
- For mission estimation it is better to separate your commands for getting to the orbit location and your commands for doing the orbit itself [This was someone elses point - I don't understand why].
- For a smart shot, usually you want to fly to orbit location, do the orbit, then fly back to original position where orbit shot was done.
- Desirable to be able to specify centre, radius, direction and speed of rotation, direction of vehicle yaw, direction of yaw on exit in a mission.

Anything else?

To progress this can you outline a little more about your setup. It would help to have a discussion at the dev call. I know that might be hard for you - is there a good time that overlaps with Zurich work hours? If so, then I can arrange a discussion with someone in the team.

hamishwillee:
> It is really tentative, but I can attend a meeting of 1-2 hrs in May 2 between CEST 12.00-5 PM 🤔

Sorry, I missed your response!

@julianoes Can you liase with @tahsinkose to discuss the needs/requirements here?

message_definitions/v1.0/common.xml
```diff
@@ -704,6 +704,21 @@
...
+        <description>Vehicle state is unknown according to orbit action.</description>
+      </entry>
+      <entry value="1" name="ORBIT_EXECUTION_STATE_APPROACH_TO_CIRCLE">
+        <description>Vehicle is approaching to orbit track.</description>
+      </entry>
+      <entry value="2" name="ORBIT_EXECUTION_STATE_STARTED">
+        <description>Vehicle started following orbit track.</description>
+      </entry>
+      <entry value="3" name="ORBIT_EXECUTION_STATE_FINISHED">
+        <description>Vehicle finished following orbit track.</description>
```
MaEtUgR, line 719: Finished needs to be clearly defined. One complete circle fixed? Should we support commanding a certain number of turns before it's finished? Should we share a [0,1] progress value?

message_definitions/v1.0/common.xml
MaEtUgR: I thought the same. This is a quite bad GitHub bug 😕 We should send a bug report with this example.

MaEtUgR (commented):
Good idea.

In the minimum, we should clearly define finished as one turn. Optionally we support commanding multiple turns and reporting a progress. I think that would make it much more flexible.

What do you think? If we see the use case I think we should extend right away.

message_definitions/v1.0/common.xml
tahsinkose: > Should we support commanding a certain number of turns before it's finished

I believe radians would be more general to command. e.g. 4*pi for 2 turns, 3*pi for 1.5 turns and etc. Reporting progress between STARTED-FINISHED in percentage seems a nice idea.

See https://mavlink.io/en/messages/common.html#MAV_CMD_DO_ORBIT. 4th field is already reserved for such improvements. We may simply rename it as `turn_in_radians`, which would default to NaN (means infinite orbit -> back-compatibility). 

By the way I'm only referring PX4's `FlightTaskOrbit` as the implementation. Maybe there are other autopilot software who already started using that field and such a change might break them. Nevertheless there is an explicit message that says `WORK IN PROGRESS: Do not use in stable production environments (it may change)`, so this is a fair change :slightly_smiling_face:

hamishwillee:
> In the minimum, we should clearly define finished as one turn. Optionally we support commanding multiple turns and reporting a progress. I think that would make it much more flexible.

@MaEtUgR So it depends a bit on what happens after finishing. Right now I think this is a command and the orbit basically goes on forever. So finished would mean "I got another command that made me leave this mode". There isn't value unless you can be in the mode and not exiting.

In that case I see it as useful, but again, it would mean "we stopped orbiting". I wouldn't base that on complete turns since we might want to exit on partial turns - and we'd still be finished. Hope that makes sense. I plan on looking at the responses on this tomorrow. But I really think you guys should all have a meeting to nut out the different use cases we'd like an orbit to address.

message_definitions/v1.0/common.xml
hamishwillee: FWIW This sounds practical except you will [have to use 0](https://mavlink.io/en/guide/define_xml_element.html#reserved) to indicate unlimited turns. This is not a problem. 

Other than that. 
- No so concerned about not having sufficient fields to specify "orbit for time". If we need that we could add another command.
- Are you considering a separate field for progress of rotation in the ORBIT_EXECUTION_STATUS or a replacement for the state machine field? I see both as somewhat useful. 
- Is OK always OK to change a command by using a reserved field, provided the default value doesn't change. Orbit status not so sure. Depends on what is using it.
- This would be OK in a mission too. If no number of turns was specified the vehicle would orbit forever. Otherwise it would just go to the next waypoint and you would get the normal waypoint change message as well on the switch.
- @MaEtUgR What would the PX4 mode be before, during and after orbit? I.e. right now I _think_ the task executes in the context of HOLD mode right? I'm just trying to understand if you'd consider just shifting to hold mode after the orbit completes or do something else?

@julianoes Any thoughts? I was hoping you and @MaEtUgR and @tahsinkose could discuss this so we can move forward. Otherwise we converge here on a solution that suits us all.

tahsinkose: > FWIW This sounds practical except you will have to use 0 to indicate unlimited turns

NaN seems better IMHO.

> Are you considering a separate field for progress of rotation in the ORBIT_EXECUTION_STATUS or a replacement for the state machine field

Separate filed is better I think.

> This would be OK in a mission too. If no number of turns was specified the vehicle would orbit forever. Otherwise it would just go to the next waypoint and you would get the normal waypoint change message as well on the switch.

Agreed.

tahsinkose:
@MaEtUgR @hamishwillee [Here is a video](https://drive.google.com/file/d/1Zw99k6o96xzR-vB87dNFLajm9bdI1p7q/view?usp=sharing) in Gazebo where I command the robot to orbit for `3pi` around each plant. As soon as `3pi` radian is achieved, orbit execution state machine transition into `FINISHED`, effectively causing the robot to hover at its current pose. Since my ROS application subscribes to orbit execution status telemetry topic (which I added in mavros with https://github.com/mavlink/mavros/pull/1569), it commands the robot to start the next orbit if there are more plants to monitor, otherwise it commands "Return Home". 

> In the minimum, we should clearly define finished as one turn

I can't see any particular reason for such a constraint. I currently hardcode `3pi` radians to my custom flight task implementation, but that can be easily made an argument. One can even specify, e.g. `pi/9` to have a small arched motion around a center point for just showing off :slightly_smiling_face:

message_definitions/v1.0/common.xml
hamishwillee: >> FWIW This sounds practical except you will have to use 0 to indicate unlimited turns

> NaN seems better IMHO.

I appreciate it is not as logical, but this is a convention with a reasonable justification. The default "do nothing value" has to be set when you create a command - and by default this is 0. Any GCS that has not been updated to understand NaN will be sending 0 for this field, which means that a recipient that does support the new command will by default immediately exit the orbit. That's a break in behaviour. 

>> Are you considering a separate field for progress of rotation in the ORBIT_EXECUTION_STATUS or a replacement for the state machine field

> I'm considering to add a separate field.

Makes sense to me. Both fields have potential merit.

>> In the minimum, we should clearly define finished as one turn

> I can't see any particular reason for such a constraint. I currently hardcode 3pi radians to my custom flight task implementation, but that can be easily made an argument. One can even specify, e.g. pi/9 to have a small arched motion around a center point for just showing off 🙂

I agree. There is no need for a minimum if you specify the number in radians.

----

My only concern is still "what marks the end of orbit", and it is mostly informed by the fact that I know on PX4 the orbit is a task not a mode.  Now if orbit was a mode, then you would know that orbit has three phases - getting to the orbit, orbiting, whatever it does once the orbit is complete - probably loitering. Then you would emit the `ORBIT_EXECUTION_STATUS` for all three phases. The chances of getting an ORBIT_EXECUTION_STATUS indicating finished is high, because you're sitting there in loiter say afterwards, continuing to emit it.

As it is now, you'll finish the orbit and I suspect drop out of the flight task. My concern is that at this point you will at most emit just one ORBIT_EXECUTION_STATUS indicating you're finished and the progress - easy to miss. Not a bit deal in a mission because you can see the waypoint change or query the waypoint number. I think that perhaps we should not have this finished state at all because we can't reliably emit it - and perhaps we no longer need it. 

@tahsinkose Could you update the PR to reflect all this thinking (or at least the bits we agree on?). I don't have edit rights.

hamishwillee: > Finished needs to be clearly defined

Agree. I think it means "we have competed our orbiting phase and are now doing something else". That might mean we're still in an orbit mode, but we're loitering because our orbit finished. Or it might mean we transitioned to some other mode. 

I think it isn't particularly useful in the "general" case because often it will get sent only once or never because you've moved to a new mode. Too easy to miss, so you'd probably end up looking for absence of the status message or a mode change as well. I guess it does not harm.

tahsinkose: @hamishwillee I now enabled edit rights. 

> As it is now, you'll finish the orbit and I suspect drop out of the flight task. My concern is that at this point you will at most emit just one ORBIT_EXECUTION_STATUS indicating you're finished and the progress - easy to miss

I don't think there is an inclusive _drop out of flight task_ behavior embedded in flight tasks, but again I'm referring to https://docs.px4.io/master/en/concept/flight_tasks.html. I must say that this is pretty autopilot specific and another implementation might simply prefer transitioning into hovering/loitering state. In that case, I agree with your concern.

However PX4 executes the generated setpoints for the orbit indefinitely until another flight task takes control. In this setting, this can only happen in a higher level -e.g. a composite FlightTask written in autopilot, from the offboard application that uses autopilot externally or directly via GCS. Let's assume such a mode change triggered by some external component and the application is unaware of it for some finite time. Even in that case, mode-change monitoring could serve the same purpose in the absence of status message as you suggest. Consequently it boils down to the application implementation to handle such redundancy issues IMO.

tahsinkose: @hamishwillee Addressed. PTAL :pray:

message_definitions/v1.0/common.xml
```diff
@@ -6572,6 +6587,8 @@
       <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
+      <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
+      <field type="uint8_t" units="%" name="progress">Progress of orbit. UINT_MAX8 if orbit is indefinite.</field>
```
hamishwillee, line 6591: TYI I turned this into a uint8 - much smaller and granularity of 1 is fine for %

hamishwillee:
Thanks @tahsinkose . Looking good. There are probably a couple of other minor things I should do to it (E.g. allow default value for radius, velocity etc if called when in already in orbit). Will discuss with dev call tonight and hopefully merge tomorrow.

message_definitions/v1.0/common.xml
```diff
@@ -962,7 +962,7 @@
         <param index="1" label="Radius" units="m">Radius of the circle. positive: Orbit clockwise. negative: Orbit counter-clockwise.</param>
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Vehicle configuration default.</param>
         <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
-        <param index="4" label="Turn in Radians" units="rad" minValue="0" default="0">Orbital turn in radians.</param>
+        <param index="4" label="Orbits" units="rad" minValue="0" default="0">Orbit centre point for this many radians (i.e. for a three-quarter orbit set 270*π/180). 0: Orbit forever.</param>
```
julianoes, line 965: @hamishwillee Do we support non ascii chars like `π` in xml?
I would not be surprised if it breaks tools down the stream.

hamishwillee: @julianoes Fair point. I'll change to pi.

hamishwillee:
@tahsinkose We had a discussion in the dev call on this. There was concern that the progress update is something we need to have more generically - ie all commands should report their progress in some way. As a result I have pulled that out of this request (and will submit attempt at generic message shortly in a separate PR).

We are broadly agreed on the turns addition though. I will do some other minor fixes on this and seek final review before merging.

This is what I removed.
```
    <enum name="ORBIT_EXECUTION_STATE">
      <description>State of execution during orbit flight.</description>
      <entry value="0" name="ORBIT_EXECUTION_STATE_UNKNOWN">
        <description>Vehicle state is unknown according to orbit action.</description>
      </entry>
      <entry value="1" name="ORBIT_EXECUTION_STATE_APPROACH_TO_CIRCLE">
        <description>Vehicle is approaching to orbit track.</description>
      </entry>
      <entry value="2" name="ORBIT_EXECUTION_STATE_STARTED">
        <description>Vehicle started following orbit track.</description>
      </entry>
      <entry value="3" name="ORBIT_EXECUTION_STATE_FINISHED">
        <description>Vehicle finished following orbit track.</description>
      </entry>
    </enum>

      <field type="uint8_t" name="state" enum="ORBIT_EXECUTION_STATE">State of execution during the orbit action.</field>
      <field type="uint8_t" units="%" name="progress">Progress of orbit. UINT_MAX8 if orbit is indefinite.</field>
```

tahsinkose:
Orbit execution status is already a a published telemetry. How the new generic method will apply to that then without the explicit `state` field?

hamishwillee:
@tahsinkose That is a very good question - or at least as I am interpreting it "Should the new mesage also send positional information, if present". If not, what do you mean? The new message will has state information in it.

tahsinkose:
Your interpretation seems correct AFAIU. I like discussing over real examples though. Consider https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS. My proposal was to add `state` field in it so that it can be published over MAVLink in the context of `ORBIT_EXECUTION_STATUS`. But you have deleted it for the sake of `COMMAND_PROGRESS` message. I suppose, this `COMMAND_PROGRESS` message will be a field of `ORBIT_EXECUTION_STATUS`. Otherwise how `COMMAND_PROGRESS` should be published?

hamishwillee:
@tahsinkose The COMMAND_PROGRESS is still in discussion (as you know) and I hope we can answer those questions in that PR. 

I felt it useful to merge this because it was the part of the command we were generally in agreement on. It is possible that the `COMMAND_PROGRESS` doesn't work as a general command and we end up with a new PR on ORBIT_EXECUTION_STATUS. 

Just FMI, if you know you can command a particular arc what do you actually _need_ in the way of additional information for your use case? i.e. you no longer need the entry and exit status to calculate your arc. Perhaps you need to know when the arc is finished to send the next command?

PR #1623, opened 2021-04-28
Fix external links and dev/all dialect docs

hamishwillee:
This improves the generated mavlink docs:
- The addition of the external dialects was throwing an error as the dialect doc was not being generated. My take is that we should not build these non-standard dialects, so the update changes the include from a link to just plain text.
- The development.xml and all.xml are new. This also adds header docs explaining what they are for.

### DIFF
--- a/doc/mavlink_gitbook.py
+++ b/doc/mavlink_gitbook.py
@@ -66,6 +66,18 @@ def fix_content_in_tags(input_html):
     input_html=re.sub(r'\>(\s+?\w+?.*?)\<', remove_space_between_content_tags, input_html,flags=re.DOTALL)
     return input_html
     
+def fix_external_dialect_link(input_html):
+    #print("fix_external_dialect_link was called")
+    def fixupexternaldialecturls(matchobj):
+        stripped_string=matchobj.group(1).strip()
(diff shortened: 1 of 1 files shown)


doc/mavlink_gitbook.py
```diff
@@ -66,6 +66,18 @@ def remove_space_between_content_tags(matchobj):
     input_html=re.sub(r'\>(\s+?\w+?.*?)\<', remove_space_between_content_tags, input_html,flags=re.DOTALL)
     return input_html
     
+def fix_external_dialect_link(input_html):
+    #print("fix_external_dialect_link was called")
+    def fixupexternaldialecturls(matchobj):
+        stripped_string=matchobj.group(1).strip()
+        return '%s' % stripped_string
```
julianoes, line 73: Isn't this the same as `return stripped_string`?

And then you could just make it:
```
return matchobj.group(1).strip()
```

hamishwillee: Yes. Fixed.

hamishwillee:
I'm going to self merge since this got reviewed.

julianoes (approved):

PR #1626, opened 2021-04-30
Revert "Command to allow geofence action to be set per-fence (#1482)"

bkueng:
This reverts commit 66520310ec21ba55dcd006e8f5d8f433bc2f3fcf.

There's a problem with the C library generator:
The addition of MAV_CMD to development.xml led to the removal
of MAV_CMD in common.h. This breaks build for everyone using that enum
and including 'common.h' or 'standard.h' (and probably others).
See https://github.com/mavlink/c_library_v2/commit/992871ba04b721dac6ef8a851dbe9df3b6c22e2d

@hamishwillee agreed to revert this until we have a correct fix.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -314,30 +314,20 @@
     </enum>
     <!-- fenced mode enums -->
     <enum name="FENCE_ACTION">
-      <description>Actions following geofence breach.</description>
       <entry value="0" name="FENCE_ACTION_NONE">
-        <description>Disable fenced mode. If used in a plan this would mean the next fence is disabled.</description>
+        <description>Disable fenced mode</description>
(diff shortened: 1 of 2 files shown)


olliw42:
@bkueng
just in case it might be helpful
this is ultimately a problem of the generator, you can however often circumvent it by careful arrangement of the order of dialects here https://github.com/mavlink/mavlink/blob/master/scripts/update_c_library.sh#L63-L69
you need to choose the order such that "higher" dialects don't overwrite "lower" dialects
it can be logically understood but it is a bit weird, so a bit testing might be needed
I have not tested, but I guess I would try to swap the calls to standard and development

(few months ago I did a fix in mavgen but threw it away)(this issue was btw one reason I went away from mavgen LOL)

PR #1627, opened 2021-05-05
Invalid attribute: Replace 255 by UINT8_MAX and 65535 by UINT16_MAX

olliw42:
so here now the clean up PR
I've only touched standard.xml and includes, which boils down to only common.xml

* replaces 255 by UINT8_MAX in message fields which have an 'invalid' attribute
* replaces 65535 by UINT16_MAX in message fields which have an 'invalid' attribute
* in two cases the invalid is 0, but in the message description 65535 appears for another situation, I have replaced these also by UINT16_MAX (see messages GPS_RAW_INT and GPS2_RAW)

there are now no 65535's anymore

there are still a couple of 255's in message field descriptions, but these are valid values. I generally found the description more readable with the 255's, so I systematically left those as they are.

This PR should be totally NFC

Not done: As discussed in the original PR, the attribute 'invalid' refers to a longer list of possibilities which I had listed all here: https://github.com/mavlink/mavlink/pull/1614#issuecomment-820264838, and one might want to modify the descriptions to only some few wordings to be more consistent, see https://github.com/mavlink/mavlink/pull/1614#issuecomment-820310664. Since among us two hamishwillee is the language expert, I didn't jump forward here :)


hamishwillee (approved):
Thanks @olliw42 - looks great to me. @julianoes I have approved, if you are also OK with this then please merge.

As Ollie indicated, we can defer making the text for describing what invalid means.

julianoes (approved):
Nice, looks correct!

olliw42:
hups, this was fast ... 
:)

hamishwillee:
Yes, well I see it as non controversial. I've been wrong before though.

PR #1628, opened 2021-05-05
external STorM32.xml: Add invalid attribute, add QSHOT home targeting

olliw42:
the title says it

updates external dialect STorM32.xml to make use of the invalid attribute, and adds a new enum field

THX


hamishwillee:
Done. Hope you're having a great day!

olliw42:
many thx
sure I do :):)

PR #1632, opened 2021-05-11
minimal: add compid for safety

julianoes:
It would be nice to have a component ID for a safety device such as a parachute module.

FYI @MaEtUgR

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -581,6 +581,9 @@
       <entry value="160" name="MAV_COMP_ID_FLARM">
         <description>FLARM collision alert component.</description>
       </entry>
+      <entry value="161" name="MAV_COMP_ID_PARACHUTE">
+        <description>Parachute component.</description>
+      </entry>
       <entry value="171" name="MAV_COMP_ID_GIMBAL2">
         <description>Gimbal #2.</description>
       </entry>
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -179,6 +179,9 @@
       <entry value="36" name="MAV_TYPE_BATTERY">
         <description>Battery</description>
       </entry>
+      <entry value="37" name="MAV_TYPE_PARACHUTE">
+        <description>Parachute</description>
+      </entry>
     </enum>
     <enum name="MAV_MODE_FLAG" bitmask="true">
       <description>These flags encode the MAV mode.</description>


hamishwillee:
@julianoes You have a mavlink-enabled parachute? Tell me more?

If you need have a parachute that needs to be a discrete node then sure, but [for component identity](https://mavlink.io/en/services/heartbeat.html#component-identity) you will also need a https://mavlink.io/en/messages/common.html#MAV_TYPE

message_definitions/v1.0/minimal.xml
```diff
@@ -581,6 +581,9 @@
       <entry value="160" name="MAV_COMP_ID_FLARM">
         <description>FLARM collision alert component.</description>
       </entry>
+      <entry value="161" name="MAV_COMP_ID_SAFETY">
```
hamishwillee, line 584: This also feels a little broad - you could call FLARM above it a safety too. Perhaps this is a mavlink safety switch? :-). Is there a better alternative? Maybe ID_EMERGENCY_FAILSAFE ?

hamishwillee: I'm not advocating, just asking the question. The main reason to allocate an id is to make clashes rare if you only have one instance of a particular thing. If you potentially have 2 instances then you have to have a way for an integrator to change the ids. Upshot, if you use it for many possible things then you are more likely to have clashes than if you use if for just one thing.

julianoes: FAILSAFE could mean many things, I'd say. PARACHUTE would be an option.

hamishwillee: I'd prefer parachute. But as per my other note, if there will only ever by "one" of whatever this is on the vehicle then it can be as general as you like. 

What we really need to do is start making sure the MAV_TYPE is used as it is supposed to be and invent a protocol to auto-allocate all system and component IDs. The problem now is that people still configure code based on the presumed type of the component based on this ID. We'd prefer they didn't.

julianoes: Hmm, also right.

julianoes: The problem is that when you send a command somewhere in your code you need to fill in the fields sysid/compid of where your command should be sent to. However, in that location you don't know the map of all the mavlink components which have been discovered and what their MAV_TYPE is, so it's easier to just say "send my parachute command to the parachute component".

hamishwillee: Yes, I understand it is a shit. That is why I objected so hard to the change to MAV_TYPE - it makes things easier for standalone component authors who would quite like to be able to hard code the type and somewhat hard code the id, but it makes it harder for things that need to address, and now need to maintain a mapping of id to type. 

That said, I am still not sure that our current way is not better. I think at some point we will run out of pre-allocated IDs. Then we're going to want to have auto allocation, and this approach allows it.

hamishwillee:
If you're going to add this I think we do need a TYPE as well.

julianoes:
> If you're going to add this I think we do need a TYPE as well.

Done.

hamishwillee:
Thanks :-)

PR #1633, opened 2021-05-11
common: make PARACHUTE_ACTION enum consistent

julianoes:
I think that's more consistent and easier to grep this way.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3700,13 +3700,13 @@
     <!-- parachute action enum -->
     <enum name="PARACHUTE_ACTION">
       <description>Parachute actions. Trigger release and enable/disable auto-release.</description>
-      <entry value="0" name="PARACHUTE_DISABLE">
+      <entry value="0" name="PARACHUTE_ACTION_DISABLE">
         <description>Disable auto-release of parachute (i.e. release triggered by crash detectors).</description>
       </entry>
-      <entry value="1" name="PARACHUTE_ENABLE">
+      <entry value="1" name="PARACHUTE_ACTION_ENABLE">
         <description>Enable auto-release of parachute.</description>
       </entry>
-      <entry value="2" name="PARACHUTE_RELEASE">
+      <entry value="2" name="PARACHUTE_ACTION_RELEASE">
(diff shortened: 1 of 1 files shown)


hamishwillee:
FWIW it's a code change in a couple of places like https://github.com/ArduPilot/ardupilot/blob/master/ArduCopter/GCS_Mavlink.cpp#L828-L844 and the GCS, but not any kind of compatibility change or break in what is sent (ie it's a break that you can quickly fix next time you update the library).

julianoes:
In my opinion it should be ok to break this because it's caught at compile time which is easy to fix. Plus, the data stays the same on the wire.

Unless people disagree, then I'll close this.

auturgy:
I don't disagree the intent, but don't think a cosmetic tidy up warrants breaking safety critical systems.

hamishwillee:
> I don't disagree the intent, but don't think a cosmetic tidy up warrants breaking safety critical systems.

Both the change and the risk are trivial. Personally I think it is worth making the change if the risk really is trivial and the effort minimal, which is true in this case. I would object to the same change in [MAV_COMPONENT](https://mavlink.io/en/messages/common.html#MAV_COMPONENT), because there are so many instances of these things that the risk and effort are increased.

auturgy:
I don't see the risk as trivial: this is implemented in a number of ways external to the flight controller and gcs. I agree the intent but introducing risk into a safety critical system for what is a cosmetic tidy up does not make sense to me. The benefit is trivial compared to the consequence.

julianoes:
Just to understand: @auturgy what is your safety concern? This does not change the meaning or values on the wire.

auturgy:
I understand that: but we don’t have visibility of how this is implemented in external micros, scripts, devices etc. If we can prove that this isn’t going to break anything anywhere, great. But I don’t think we can. We’d be relying on implementers to test and update all components and I don’t think we should expose our communities to that for a cosmetic change.
If this wasn’t for parachute I’d be less concerned, but when you need a chute to deploy, you need it to deploy.

LorenzMeier:
I think those concerns are fair and the risk / reward doesn't in favor of this change. I think we need to solve versioning in order to be able to make changes like this.

PR #1634, opened 2021-05-13
common.xml - COMMAND_INT - add note about using Nan/INT32_MAX for default params

hamishwillee:
Copied this note out of MISSION_ITEM_INT:

> NaN or INT32_MAX may be used in float/integer params (respectively) to indicate optional/default values (e.g. to use the component's current latitude, yaw rather than a specific value)

This essentially exists for the same reason - to remind people that they can't use NaN for param 5,6 if sending in the COMMAND_INT.


julianoes (approved):
Nice, thanks.

PR #1635, opened 2021-05-13
COMMAND_PROGRESS message proposal 

hamishwillee:
This PR adds proposal for a new message `COMMAND_PROGRESS` that can be streamed to indicate progress and completion of the action associated with a command. For example, if you send the `MAV_CMD_NAV_TAKEOFF` command it may complete immediately with a `COMMAND_ACK.result=MAV_RESULT_ACCEPTED`. You can then monitor COMMAND_PROGRESS to understand whether it is close to reaching the takeoff altitude, and when it has reached the altitude.

While it is possible to detect command completion in this case by other means (e.g. checking altitude and comparing to the expected target), `COMMAND_PROGRESS` provides a _generic_ solution to detecting completion of any command.

> **Note** Further, this is better than the [long running commands protocol](https://mavlink.io/en/services/command.html#long_running_commands) (and can in theory replace it) because many commands that are actually long running (like takeoff) predate the protocol and cannot be updated. This allows a non-breaking update path. 

This proposal originated from discussion of https://github.com/mavlink/mavlink/pull/1622

--- 

Notes:
- if desired, we could add a configurable field or two of zero-default "data" that could be used for whatever purpose is required and documented in a particular command. E.g. additional error values that make sense for that system, current count of turns, etc. FYI @MaEtUgR 
- This states this is only for commands. My thinking is that for commands uses in missions you already have the notification of changed waypoint to know you reached end of particular item.
- I have stated this is streamed after command accepted, and until it is complete + at least 3 messages. Reasonable?
- Note this is directed to the vehicle that sent the command via the target system/command id.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -119,7 +119,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">Mission command being reported.</field>
-      <field type="uint8_t" name="progress" units="%">Progress of main command action. 0 if has not started. UINT16_MAX if progress cannot be estimated.</field>
+      <field type="uint8_t" name="progress" units="%">Progress of main command action. 0 if has not started. UINT8_MAX if progress cannot be estimated.</field>
       <field type="uint8_t" name="state" enum="COMMAND_STATE">State of command execution.</field>
     </message>
   </messages>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -44,6 +44,33 @@
         <description>Synthetic/calculated airspeed.</description>
       </entry>
     </enum>
+    <enum name="COMMAND_STATE">
+      <description>Command execution state. This is used in COMMAND_PROGRESS message to indicate progress of the main action performed by a command.</description>
+      <entry value="0" name="COMMAND_STATE_UNKNOWN">
+        <description>Command state unknown.</description>
+      </entry>
+      <entry value="1" name="COMMAND_STATE_PENDING">
+        <description>Action has not yet started.</description>
+      </entry>
+      <entry value="3" name="COMMAND_STATE_PRE_ACTION">
+        <description>Command is executing/doing stuff but the action has not yet started (e.g. flying to orbit position).</description>
+      </entry>
+      <entry value="4" name="COMMAND_STATE_STARTED">
+        <description>Action has started.</description>
+      </entry>
+      <entry value="5" name="COMMAND_STATE_COMPLETE">
+        <description>Action has completed normally (e.g. takeoff altitude reached, commanded number of orbits reached).</description>
+      </entry>
+      <entry value="6" name="COMMAND_STATE_FAILED">
+        <description>Commanded action has failed (e.g takeoff, but did not reach altitude, format error on file write).</description>
+      </entry>
+      <entry value="7" name="COMMAND_STATE_INTERRUPTED">
+        <description>Commanded action has been replaced with some other operation (e.g. a goto command has been received half way through a takeoff).</description>
+      </entry>
+      <entry value="8" name="COMMAND_STATE_CANCELLED">
+        <description>Commanded action has been excplicitly cancelled (e.g. by COMMAND_CANCEL message).</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="53" name="MISSION_CHECKSUM">
@@ -81,6 +108,19 @@
       <field type="uint8_t" name="signal_quality" units="%">WiFi network signal quality.</field>
       <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.</field>
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
+    </message>
+    <message id="78" name="COMMAND_PROGRESS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Command protocol progress message.
+      This is used to indicate progress and completion of the action associated with a command (i.e. completion of takeoff), even if the command itself completes immediately with MAV_RESULT_ACCEPTED.
+      It should be streamed from the point the command has been accepted until the main action has completed or otherwise ceased (e.g. if the action is superseded by another action or). The final state should be streamed at least three times following the action ending.
+      The streaming rate is nominally 5Hz, but is expected to be appropriate for the operation being reported, and configurable using normal MAVLink methods. </description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint16_t" name="command" enum="MAV_CMD">Mission command being reported.</field>
+      <field type="uint8_t" name="progress" units="%">Progress of main command action. 0 if has not started. UINT16_MAX if progress cannot be estimated.</field>
+      <field type="uint8_t" name="state" enum="COMMAND_STATE">State of command execution.</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -68,7 +68,7 @@
         <description>Commanded action has been replaced with some other operation (e.g. a goto command has been received half way through a takeoff).</description>
       </entry>
       <entry value="8" name="COMMAND_STATE_CANCELLED">
-        <description>Commanded action has been excplicitly cancelled (e.g. by COMMAND_CANCEL message).</description>
+        <description>Commanded action has been explicitly cancelled (e.g. by COMMAND_CANCEL message).</description>
       </entry>
     </enum>
   </enums>


message_definitions/v1.0/development.xml
```diff
@@ -44,6 +44,33 @@
...
     </enum>
+    <enum name="COMMAND_STATE">
+      <description>Command execution state. This is used in COMMAND_PROGRESS message to indicate progress of the main action performed by a command.</description>
+      <entry value="0" name="COMMAND_STATE_UNKNOWN">
+        <description>Command state unknown.</description>
+      </entry>
+      <entry value="1" name="COMMAND_STATE_PENDING">
+        <description>Action has not yet started.</description>
+      </entry>
+      <entry value="3" name="COMMAND_STATE_PRE_ACTION">
```
hamishwillee, line 55: Note, this is to allow easy reporting of things like flying to orbit. I am not sure if it should be used for (say) arming when taking off, but my leaning is that it should not. It is just for cases where you do something else before you can start the thing you actually want to do, and where people might care.

message_definitions/v1.0/development.xml
```diff
@@ -44,6 +44,33 @@
...
+        <description>Command is executing/doing stuff but the action has not yet started (e.g. flying to orbit position).</description>
+      </entry>
+      <entry value="4" name="COMMAND_STATE_STARTED">
+        <description>Action has started.</description>
+      </entry>
+      <entry value="5" name="COMMAND_STATE_COMPLETE">
+        <description>Action has completed normally (e.g. takeoff altitude reached, commanded number of orbits reached).</description>
+      </entry>
+      <entry value="6" name="COMMAND_STATE_FAILED">
```
hamishwillee, line 64: So these three fail states exist because they allow this message to replace the long running command protocol

julianoes:
> .. because many commands that are actually long running (like takeoff) predate the protocol and cannot be updated.

I'd like to challenge that. Why can these commands not be updated?

message_definitions/v1.0/development.xml
```diff
@@ -44,6 +44,33 @@
...
+      <entry value="1" name="COMMAND_STATE_PENDING">
+        <description>Action has not yet started.</description>
+      </entry>
+      <entry value="3" name="COMMAND_STATE_PRE_ACTION">
+        <description>Command is executing/doing stuff but the action has not yet started (e.g. flying to orbit position).</description>
+      </entry>
+      <entry value="4" name="COMMAND_STATE_STARTED">
+        <description>Action has started.</description>
+      </entry>
+      <entry value="5" name="COMMAND_STATE_COMPLETE">
```
tahsinkose, line 61: typo -> COMPLETED

tahsinkose (changes requested):
Please explicitly add `COMMAND_PROGRESS` to orbit execution status at https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS

hamishwillee:
>> .. because many commands that are actually long running (like takeoff) predate the protocol and cannot be updated.

> I'd like to challenge that. Why can these commands not be updated?

Of course anything is possible, but  I really believe "will never be". ArduPilot do not support the long running protocol, partially because it was merged with some other bits they don't agree with. I think this approach is something they might want to do anyway because it could be appended to the existing process in a non-breaking way. I suspect little change needed for state machines, and if you have a script that currently works, it will still work. That is not completely certain if takeoff (for example) were to start returning progress updates.

To me it just feels like a better way to extend command protocol to support long running commands than what we did. Of course the long running stuff already is defined ... Upshot, there are some questions here.

message_definitions/v1.0/development.xml
hamishwillee: That's not a typo - command is complete, command has completed

hamishwillee:
Should this message include current position information, if relevant to the message? I.e. consider [ORBIT_EXECUTION_STATUS](https://mavlink.io/en/messages/common.html#ORBIT_EXECUTION_STATUS). 

My gut feeling is that it is _relevant_ to a bunch of commands but I still wouldn't include it - there are other ways to get this information and it would be overloading the purpose of this message too greatly.

tahsinkose:
@hamishwillee if `COMMAND_PROGRESS` is not made a field of `ORBIT_EXECUTION_STATUS`, how could we send that as a telemetry in the respective task in the autopilot? Meaningfully, there would be 2 distinct telemetry informations (`COMMAND_PROGRESS` and `ORBIT_EXECUTION_STATUS`) that could only be published into different channels unless the former is added as a field in the latter. See the orbit telemetry publish implementation at https://github.com/PX4/PX4-Autopilot/blob/5f14ee2a9705d746de7a87034f347db1c624fab2/src/modules/flight_mode_manager/tasks/Orbit/FlightTaskOrbit.cpp#L108-L128. If they are left isolated, an additional publish that sends `COMMAND_PROGRESS` needs to be done in here. Besides `COMMAND_PROGRESS` might seem as a command-agnostic component, but in practice it is not. The progress could only be determined by the custom logic of the particular command. In essence, it needs to be computed at `FlightTask` level, which enforces the telemetry `publish` to be done at that level.

I think any such telemetry message -e.g. `***_STATUS` - that _needs_ it should include as a field. If it is not needed, then there is no point of flooding the communication layer with never-used progress messages.

julianoes:
> partially because it was merged with some other bits they don't agree with.

Right, that's unfortunate.

TSC21:
So what happens with the `IN_PROGRESS` ACKs?

julianoes:
@hamishwillee I think this is yet again a problem of versioning and having flags to move from one implementation to a better one. We have progress in commands but we can't use it because it requires ground stations and autopilots to move together (which they don't). So we need a way to flag (or version mechanism) that either side supports progress and then use it depending on that.

hamishwillee:
The main driver for this message being generic was (for me) to have a way to get around the fact that things like takeoff and orbit are long-running commands that are implemented (by every stack) as though they were instant (i.e. they complete with accepted as soon as the command is accepted).

Is there any reason we couldn't implement them as long running commands? [other than the fact that the API hasn't been used by ArduPilot]?

The [COMMAND_ACK](https://mavlink.io/en/messages/common.html#COMMAND_ACK) already has progress, so most of what we add here could be done as a long running command.

Either way I think perhaps we should consider changing the completion ACK of a [long running command](https://mavlink.io/en/services/command.html#long_running_commands) to a new value "completed" rather than accepted. That is more accurate use of both completed and accepted.

>  I think this is yet again a problem of versioning and having flags to move from one implementation to a better one.  ...

@julianoes Yes. Though since there are few things actually using the long running command (just the OS update as far as I know) presumably this is something that can still happen without breaking more than one stack.

> So what happens with the IN_PROGRESS ACKs?

@TSC21 The simple answer is until we have broad agreement that we want to change then who knows. Perhaps (as above) we iterate our existing protocol. 

@tahsinkose You make some good points. Generally the case where you want progress and status is the long running commands. The idea there is that the progress updates are sent at a rate that is appropriate for the particular command.

message_definitions/v1.0/development.xml
```diff
@@ -82,5 +109,18 @@
...
+      <description>Command protocol progress message.
+      This is used to indicate progress and completion of the action associated with a command (i.e. completion of takeoff), even if the command itself completes immediately with MAV_RESULT_ACCEPTED.
+      It should be streamed from the point the command has been accepted until the main action has completed or otherwise ceased (e.g. if the action is superseded by another action or). The final state should be streamed at least three times following the action ending.
+      The streaming rate is nominally 5Hz, but is expected to be appropriate for the operation being reported, and configurable using normal MAVLink methods. </description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint16_t" name="command" enum="MAV_CMD">Mission command being reported.</field>
+      <field type="uint8_t" name="progress" units="%">Progress of main command action. 0 if has not started. UINT16_MAX if progress cannot be estimated.</field>
```
julianoes, line 122:
```suggestion
      <field type="uint8_t" name="progress" units="%">Progress of main command action. 0 if has not started. UINT8_MAX if progress cannot be estimated.</field>
```

hamishwillee: Nice catch.

hamishwillee:
General consensus seems to be that something like this is only broadly attractive to stakeholders because it offers the opportunity to add backwards-compatible reporting about progress of commands to the existing command protocol - ie we can add this and things that want to rely on them do.

There were some questions about how you would "know" about this. Suggestion is that anything that emits this should document in the associated command. Personally I think this is another case where we need component information to indicate support for the protocol feature.

Here is a list of things that would definitely benefit 
- MAV_CMD_NAV_TAKEOFF - in particular because there is no guarantee that a vehicle will reach exactly the target height specified, so there is no way that is both reliable and efficient to know that takeoff has really completed. 
- MAV_CMD_NAV_LAND - Generally you land then disarm (automatically). Not sure how useful this is as there is. Is there some other "next step" you are waiting on? For example, can this fail, and if so is it useful to have a generic way to monitor for such a failure such as this message?
- MAV_CMD_DO_ORBIT - useful to know when you have actually started orbiting, how long it is likely to take to finish the orbiting, how long to complete a certain number of orbits. If you're orbiting forever, useful to know that you've started orbiting.
- MAV_CMD_OVERRIDE_GOTO - presumably useful to know when you got to the point you wanted to go to
- MAV_CMD_NAV_RETURN_TO_LAUNCH - ditto - can this fail? Would you use this message to find out about the failure or is a more specific approach useful?
- MAV_CMD_DO_START_MAG_CAL - knowing when it is really done?

I guess I'm asking, is this  really worth solving generically?

hamishwillee:
Closing. 1. No stakeholder, 2. We can use long-running commands for this.

PR #1637, opened 2021-05-17
Removed entries with 0 values

python36:
Only `ESC_FAILURE_NONE` and `CAMERA_TRACKING_TARGET_DATA_NONE` has 0 value. Although similar records do not have 0 value. For example `MAV_BATTERY_FAULT` do not have `MAV_BATTERY_FAULT_NONE`.


hamishwillee:
Thanks, but I'm closing this as invalid. 
- [MAV_BATTERY_FAULT](https://mavlink.io/en/messages/common.html#MAV_BATTERY_FAULT) has no "0" case because it is a bitmask - ie. if the zero bit is 0 this means there is no `MAV_BATTERY_FAULT_DEEP_DISCHARGE` fault.
- If we were to remove the ESC_FAILURE_NONE case then we would have to report an ESC error in https://mavlink.io/en/messages/common.html#ESC_INFO even if there was no error.

More generally, we try not to remove things that people might be using unless we absolutely know that no one is relying on them. Generally that means we grow this file, but do not cut things out without great care.

python36:
But ESC_FAILURE_FLAGS and CAMERA_TRACKING_TARGET_DATA are bitmasks too.

 The problem is the lack of unification.
 Some zero values are described in the description. For example:
```
    <enum name="ATTITUDE_TARGET_TYPEMASK" bitmask="true">
      <description>Bitmap to indicate which dimensions should be ignored by the vehicle: a value of 0b00000000 indicates that none of the setpoint dimensions should be ignored.</description>
      <entry value="1" name="ATTITUDE_TARGET_TYPEMASK_BODY_ROLL_RATE_IGNORE">
...
    <enum name="GIMBAL_DEVICE_ERROR_FLAGS" bitmask="true">
      <description>Gimbal device (low level) error flags (bitmap, 0 means no error)</description>
      <entry value="1" name="GIMBAL_DEVICE_ERROR_FLAGS_AT_ROLL_LIMIT">
```
 Some are described in the text in the fields where bitmasks are used. For example:
 ```
      <field type="uint32_t" name="onboard_control_sensors_present" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are present. Value of 0: not present. Value of 1: present.</field>
...
 ```
  Some zero values are described as a entry. For example:
  ```
      <entry value="0" name="ESC_FAILURE_NONE">
        <description>No ESC failure.</description>
      </entry>
...
      <entry value="0" name="CAMERA_TRACKING_TARGET_DATA_NONE">
        <description>No target data</description>
      </entry>
  ```

If we cannot be sure of compatibility, then we need to add a zero value for all bitmasks. What do you think about this?

hamishwillee:
There may be some inconsistency in the way that various cases are presented. But none of this is ambiguous or a problem, and there is no issue of "not being sure of compatibility". 

For these I'd probably accept that "Value of 0: not present. Value of 1: present." is not needed (but nor does it do any particular harm):
```
 <field type="uint32_t" name="onboard_control_sensors_present" enum="MAV_SYS_STATUS_SENSOR" display="bitmask" print_format="0x%04x">Bitmap showing which onboard controllers and sensors are present. </field>
```

I am also concerned that these values for bitmasks propagate to GCS. What that means is that if you add a value for 0 it will (probably) appear as an option in the UI. 0 makes no sense for a flag option to present to a user.

PR #1638, opened 2021-05-17
adding new onboard computer component ids

frederictaillandier:
We need to be able to address differently to all the component from the onboard computer


hamishwillee (approved):
@julianoes Now that we rely on MAV_TYPE for component identity there isn't all that much value in reserving component ids for different instances. You shouldn't hard code any of your code to point to those ids, but instead discover components that have the right MAV_TYPE and assign them some kind of internal mapping.

If you can only ever likely to have one instance of particular component it does makes sense to reserve a component ID because it is an easy way of knowing you can integrate the component without having to explicitly set its component id.

Anyway, what I'm thinking is that perhaps by allowing this kind of allocation we're encouraging misuse of component ids. 

I have approved but not merged - feel free to do so if you think it is a good idea.


PS Though it won't do any other particular harm - we still have plenty of free ids.

auturgy:
Doesn't harm anything.

hamishwillee:
Merging as no one is arguing against. Thanks @frederictaillandier

PR #1642, opened 2021-05-27
common.xml: COMP_METADATA_TYPE - tidy up - not used in messages

hamishwillee:
@bkueng This tidies up COMP_METADATA_TYPE to reflect, in particular, that it is not used in `COMPONENT_INFORMATION.comp_metadata_typ2`

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -769,21 +769,21 @@
       </entry>
     </enum>
     <enum name="COMP_METADATA_TYPE">
-      <description>Possible values for COMPONENT_INFORMATION.comp_metadata_type.</description>
+      <description>Supported component metadata types. These are used in the "general" metadata file returned by COMPONENT_INFORMATION to provide information about supported metadata types. The types are not used directly in MAVLink messages.</description>
       <entry value="0" name="COMP_METADATA_TYPE_GENERAL">
-        <description>General information which also includes information on other optional supported COMP_METADATA_TYPE's. Must be supported. Only downloadable from vehicle.</description>
+        <description>General information about the component. General metadata includes information about other COMP_METADATA_TYPE's supported by the component. This type must be supported and must be downloadable from vehicle.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -769,21 +769,21 @@
       </entry>
     </enum>
     <enum name="COMP_METADATA_TYPE">
-      <description>Possible values for COMPONENT_INFORMATION.comp_metadata_type.</description>
+      <description>Supported component metadata types. These are used in the "general" metadata file returned by COMPONENT_INFORMATION to provide information about supported metadata types. The types are not used directly in MAVLink messages.</description>
       <entry value="0" name="COMP_METADATA_TYPE_GENERAL">
```
hamishwillee, line 773: @bkueng I guess technically we don't even need COMP_METADATA_TYPE_GENERAL, since we don't/cant reference it separately. I'd leave it in though.

message_definitions/v1.0/common.xml
```diff
@@ -769,21 +769,21 @@
...
-        <description>General information which also includes information on other optional supported COMP_METADATA_TYPE's. Must be supported. Only downloadable from vehicle.</description>
+        <description>General information about the component. General metadata includes information about other COMP_METADATA_TYPE's supported by the component. This type must be supported and must be downloadable from vehicle.</description>
```
hamishwillee, line 774: @bkueng Is it still true that this must be downloadable from the vehicle? What about on resource constrained vehicles?

message_definitions/v1.0/common.xml
bkueng: Right, I think it's used in QGC for managing all metadata.

message_definitions/v1.0/common.xml
bkueng: Yes, it's also on the vehicle there

bkueng (approved):

Issue #1641, opened 2021-05-21
Addition enum for HIL_SENSOR.fields_updated and HIGHRES_IMU.fields_updated


python36:
I suggest adding enumerations for `HIL_SENSOR.fields_updated` and  `HIGHRES_IMU.fields_updated`.

hamishwillee:
Hmmm, so 
1 - HIL_SENSOR_UPDATED_XACC
4 - HIL_SENSOR_UPDATED_YACC
8 - HIL_SENSOR_UPDATED_ZACC
...
?- HIL_SENSOR_UPDATED_TEMP

It's definitely not "wrong", but I'm not sure how useful it would be either. @julianoes got an opinion on this?

python36:
This is much clearer. For example now like this:
```
hil_sensor_send(
    time_usec=0, xacc=0, yacc=0, zacc=1, xgyro=0, ygyro=0, zgyro=0, xmag=1,
    ymag=0, zmag=0, abs_pressure=0, diff_pressure=0, pressure_alt=0, temperature=1,
    fields_updated=4164)
```
when there will be enums:
```
hil_sensor_send(
    time_usec=0, xacc=0, yacc=0, zacc=1, xgyro=0, ygyro=0, zgyro=0, xmag=1,
    ymag=0, zmag=0, abs_pressure=0, diff_pressure=0, pressure_alt=0, temperature=1,
    fields_updated=(HIL_SENSOR_UPDATED_ZACC + HIL_SENSOR_UPDATED_XMAG + HIL_SENSOR_UPDATED_TEMPERATURE))
```

hamishwillee:
Good example. I'm not opposed. Can you please create a PR for these?

PR #1643, opened 2021-05-27
Added enums HIGHRES_IMU_UPDATED_FLAGS and HIL_SENSOR_UPDATED_FLAGS

python36:
Proposed enums for HIL_SENSOR and HIGHRES_IMU, as discussed in #1641.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4339,6 +4339,93 @@
         <description>A sequence reset has happened (e.g. vehicle reboot).</description>
       </entry>
     </enum>
+    <enum name="HIL_SENSOR_UPDATED_FLAGS" bitmask="true">
+      <description>Flags in the HIL_SENSOR message indicate which fields have updated since the last message</description>
+      <entry value="1" name="HIL_SENSOR_UPDATED_XACC">
+        <description>The value in the xacc field has been updated since the last message</description>
+      </entry>
+      <entry value="2" name="HIL_SENSOR_UPDATED_YACC">
+        <description>The value in the yacc field has been updated since the last message</description>
+      </entry>
+      <entry value="4" name="HIL_SENSOR_UPDATED_ZACC">
+        <description>The value in the zacc field has been updated since the last message</description>
+      </entry>
+      <entry value="8" name="HIL_SENSOR_UPDATED_XGYRO">
+        <description>The value in the xgyro field has been updated since the last message</description>
+      </entry>
+      <entry value="16" name="HIL_SENSOR_UPDATED_YGYRO">
+        <description>The value in the ygyro field has been updated since the last message</description>
+      </entry>
+      <entry value="32" name="HIL_SENSOR_UPDATED_ZGYRO">
+        <description>The value in the zgyro field has been updated since the last message</description>
+      </entry>
+      <entry value="64" name="HIL_SENSOR_UPDATED_XMAG">
+        <description>The value in the xmag field has been updated since the last message</description>
+      </entry>
+      <entry value="128" name="HIL_SENSOR_UPDATED_YMAG">
+        <description>The value in the ymag field has been updated since the last message</description>
+      </entry>
+      <entry value="256" name="HIL_SENSOR_UPDATED_ZMAG">
+        <description>The value in the zmag field has been updated since the last message</description>
+      </entry>
+      <entry value="512" name="HIL_SENSOR_UPDATED_ABS_PRESSURE">
+        <description>The value in the abs_pressure field has been updated since the last message</description>
+      </entry>
+      <entry value="1024" name="HIL_SENSOR_UPDATED_DIFF_PRESSURE">
+        <description>The value in the diff_pressure field has been updated since the last message</description>
+      </entry>
+      <entry value="2048" name="HIL_SENSOR_UPDATED_PRESSURE_ALT">
+        <description>The value in the pressure_alt field has been updated since the last message</description>
+      </entry>
+      <entry value="4096" name="HIL_SENSOR_UPDATED_TEMPERATURE">
+        <description>The value in the temperature field has been updated since the last message</description>
+      </entry>
+      <entry value="2147483648" name="HIL_SENSOR_UPDATED_ALL">
+        <description>Full reset of attitude/position/velocities/etc was performed in sim</description>
+      </entry>
+    </enum>
+    <enum name="HIGHRES_IMU_UPDATED_FLAGS" bitmask="true">
+      <description>Flags in the HIGHRES_IMU message indicate which fields have updated since the last message</description>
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
Makes sense to me (so I have approved). Thanks for taking the time to post @python36 !

@auturgy @julianoes Any objection? Discussion on this was in https://github.com/mavlink/mavlink/issues/1641#issuecomment-848524921

julianoes:
@TSC21, @Jaeyoung-Lim any objection?

TSC21:
Where's the reset?

hamishwillee:
> Why is reset == all updated? They should serve different purposes.

@TSC21 What do you mean? Are you talking about `HIL_SENSOR_UPDATED_ALL`: Full reset of attitude/position/velocities/etc was performed in sim ?

message_definitions/v1.0/common.xml
```diff
@@ -4339,6 +4339,93 @@
...
+      </entry>
+      <entry value="1024" name="HIGHRES_IMU_UPDATED_DIFF_PRESSURE">
+        <description>The value in the diff_pressure field has been updated the since last message</description>
+      </entry>
+      <entry value="2048" name="HIGHRES_IMU_UPDATED_PRESSURE_ALT">
+        <description>The value in the pressure_alt field has been updated the since last message</description>
+      </entry>
+      <entry value="4096" name="HIGHRES_IMU_UPDATED_TEMPERATURE">
+        <description>The value in the temperature field has been updated the since last message</description>
+      </entry>
```
Jaeyoung-Lim, line 4427: Any reason why the UPDATED_ALL missing for the `HIGHRES_IMU_UPDATED_FLAGS`?

message_definitions/v1.0/common.xml
```diff
@@ -4339,6 +4339,93 @@
         <description>A sequence reset has happened (e.g. vehicle reboot).</description>
       </entry>
     </enum>
+    <enum name="HIL_SENSOR_UPDATED_FLAGS" bitmask="true">
+      <description>Flags in the HIL_SENSOR message indicate which fields have updated since the last message</description>
+      <entry value="1" name="HIL_SENSOR_UPDATED_XACC">
```
Jaeyoung-Lim, line 4344: Also add the case when entry value=0 e.g. `HIL_SENSOR_UPDATED_NONE`?

hamishwillee: My understanding is that UIs often dynamically generate UI from the info here - we wouldn't want a checkbox for the none case. Could be wrong though - and arguably the UI should be smart enough to know that a bitmask value of 0 would need special handling.

message_definitions/v1.0/common.xml
python36: In the old description of `HIL_SENSOR.fields_updated` it was written:
`Bitmap for fields that have updated since last message, bit 0 = xacc, bit 12: temperature, bit 31: full reset of attitude/position/velocities/etc was performed in sim.`

In the old description of `HIGHRES_IMU.fields_updated` it was written:
`Bitmap for fields that have updated since last message, bit 0 = xacc, bit 12: temperature`

TSC21:
> > Why is reset == all updated? They should serve different purposes.
> 
> @TSC21 What do you mean? Are you talking about `HIL_SENSOR_UPDATED_ALL`: Full reset of attitude/position/velocities/etc was performed in sim ?

@hamishwillee full reset of the sensors (position, velocity, etc are states). Example: https://github.com/PX4/jMAVSim/blob/master/src/me/drton/jmavsim/MAVLinkHILSystem.java#L208-L210.

TSC21:
@MaEtUgR @bresch FYI.

hamishwillee:
@TSC21 re https://github.com/mavlink/mavlink/pull/1643#issuecomment-860055615

Ah, I get it - you're not happy with the description because it mentions reset, which is not the same as update. Would this be better? (if not, propose alternative please)"

`HIL_SENSOR_UPDATED_ALL: All sensors in HIL_SENSOR have been updated.`

message_definitions/v1.0/common.xml
hamishwillee: @python36 That said can you please add "none" values? They are specified inconsistently but I think this is more explicit.

hamishwillee:
>>> Why is reset == all updated? They should serve different purposes.

>> @TSC21 What do you mean? Are you talking about HIL_SENSOR_UPDATED_ALL: Full reset of attitude/position/velocities/etc was performed in sim ?

> @hamishwillee yes. Example: https://github.com/PX4/jMAVSim/blob/master/src/me/drton/jmavsim/MAVLinkHILSystem.java#L208-L210. What I am saying is that is that all sensors updated and all sensors reset should be defined differently because conceptually a "reset to all" entry means that all got updated to a 0/null/NaN state, but "all updated" entry means we are expecting them to all be changed to a value different than 0/null/NaN.

@Jaeyoung-Lim @TSC21 @python36  OK, Apologies - I get this now from reading https://github.com/mavlink/mavlink/pull/1643/files#r649708994

So bit 31 is now named `HIL_SENSOR_UPDATED_RESET` and has the description _Full reset of attitude/position/velocities/etc was performed in sim (Bit 31)._ This now 

I am not sure whether it is valid/makes senset to  `HIL_SENSOR_UPDATED_ALL` which is the value for when all the bits are set - since it isn't clear whether if HIL_SENSOR_UPDATED_RESET was set you would always set all the other bits. My assumption would be not. It might be worth defining `HIL_SENSOR_UPDATED_ALL` as the value if all the field values are set, but I propose we leave this as is for now.

I added "none" values, and a the "ALL" value for the IMU enum.

Is this OK now for everyone?

message_definitions/v1.0/common.xml
python36: @hamishwillee Sorry, I read this right now. Is there anything else I can do to get this PR accepted?

message_definitions/v1.0/common.xml
hamishwillee: I agree with @python36 here. HIGHRES_IMU_UPDATED_ALL makes sense but HIL_SENSOR_UPDATED_ALL might be a little on when you consider `HIL_SENSOR_UPDATED_RESET`. So I am happy with this as is.

hamishwillee:
@TSC21 @Jaeyoung-Lim I have re-reviewed the various comments above.

I think this is good to merge because it precisely reflects the original text as an email with the following changes
- 0-value flag is defined to ease checking "nothing set"
- ALL value flag is defined for the IMU case to ease this test. All value is NOT defined for HIL_SENSOR_UPDATED_ALL because there is one value ("bit 31") that I think might make this value pointless to test against.

Any object to merging?

@peterbarker FYI this modifies some definitions that we defined in a string in the description into enum values.

hamishwillee:
Merging, as discussed. Thanks for your patience @python36 !

PR #1647, opened 2021-06-08
Add MAV_CMD_DELETE_IMAGE command

MatejFranceskin:
Add a command for deleting image from the image log.

Related messages and commands:

CAMERA_IMAGE_CAPTURED contains index in image log
CAMERA_CAPTURE_STATUS  contains total number of images in the image log
MAV_CMD_STORAGE_FORMAT has an option to reset image log
MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE to request CAMERA_IMAGE_CAPTURED from image log

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1772,6 +1772,10 @@
         <param index="4" reserved="true" default="NaN"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
+      <entry value="533" name="MAV_CMD_DELETE_IMAGE" hasLocation="false" isDestination="false">
+        <description>Delete image from image log. This operation will decrement all image indexes after deleted image and number of images in CAMERA_CAPTURE_STATUS.image_count. Camera responds with CAMERA_CAPTURE_STATUS message with updated image count.</description>
+        <param index="1" label="Number" minValue="0" increment="1">Sequence number for image log entry to be deleted.</param>
+      </entry>
       <entry value="600" name="MAV_CMD_JUMP_TAG" hasLocation="false" isDestination="false">
         <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
         <param index="1" label="Tag" minValue="0" increment="1">Tag.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1772,7 +1772,7 @@
         <param index="4" reserved="true" default="NaN"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
-      <entry value="533" name="MAV_CMD_UNLOG_IMAGE" hasLocation="false" isDestination="false">
+      <entry value="533" name="MAV_CMD_REMOVE_IMAGE_RECORD" hasLocation="false" isDestination="false">
         <description>Remove image from the image log (no longer accessible using the MAVLink camera API). Note that the image may still be present on actual storage media.</description>
         <param index="1" label="Number" minValue="0" increment="1">Sequence number for image log entry to be removed.</param>
       </entry>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1772,9 +1772,9 @@
         <param index="4" reserved="true" default="NaN"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
-      <entry value="533" name="MAV_CMD_DELETE_IMAGE" hasLocation="false" isDestination="false">
-        <description>Delete image from image log.</description>
-        <param index="1" label="Number" minValue="0" increment="1">Sequence number for image log entry to be deleted.</param>
+      <entry value="533" name="MAV_CMD_UNLOG_IMAGE" hasLocation="false" isDestination="false">
+        <description>Remove image from the image log (no longer accessible using the MAVLink camera API). Note that the image may still be present on actual storage media.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
So the use case is that you have some UI for deleting images and you want to keep the other camera data consistent? Rather than just deleting it through some other UI? Sounds sensible to me. @dogmaphobic @julianoes any comments?

If this goes though, could you provide associated update in https://mavlink.io/en/services/camera.html#still-image-capture docs?

MatejFranceskin:
Yes, this is the use case. When deleting images we want to keep the other data consistent.
I will update the docs too.

message_definitions/v1.0/common.xml
```diff
@@ -1772,6 +1772,10 @@
         <param index="4" reserved="true" default="NaN"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
+      <entry value="533" name="MAV_CMD_DELETE_IMAGE" hasLocation="false" isDestination="false">
+        <description>Delete image from image log. This operation will decrement all image indexes after deleted image and number of images in CAMERA_CAPTURE_STATUS.image_count.</description>
```
julianoes, line 1776: So for a camera implementation you would then rename all images to match the new indices? This doesn't sound intuitive to me. Usually, with cameras, when I delete a camera it would leave everything else untouched.

julianoes: I guess I connected the two in my mind, I guess my assumption was not right then.

However, @hamishwillee raised a very good point:
If you subscribe to the CAMERA_IMAGE_CAPTURED on multiple MAVLink devices, then you don't know that a picture has been deleted and you will be off when requesting one. So basically, if you move the indices of all following images, it creates a synchronization problem.

To me it would be easier to leave the index as is and avoid these problems. I don't see what the advantage would be of moving them all.

MatejFranceskin: What if we add a hash or crc of the whole photo log so that we could detect that it has to be resynced?

MatejFranceskin: The implementation should rely on photo names (URLs) and not indexes. So that it one index is deleted and others shift it doesn't create a problem. However I don't know how to specify this correctly in the protocol definition.

MatejFranceskin: I made a change so that I removed the part where indexes should be shifted after deleting image log entry. It would be better to keep the index "empty" so that if that entry is queries it would return DENIED or  CAMERA_IMAGE_CAPTURED with empty file_url.

hamishwillee:
@MatejFranceskin @julianoes That change makes me more comfortable - and less work for all concerned.

Question, the wording is "Delete image from image log." and "number for image log entry to be deleted". 
I read that as. 
1. I have a folder full of images
2. I have a log of those images
3. I call this message to delete the record of the image. The index is now empty.
4. The image itself might still be present in that folder.

Can you confirm this interpretation. I.e. this is an instruction that actions on the log - there is no assumption that anything has happened to the underlying image.

julianoes (approved):

MatejFranceskin:
@hamishwillee Yes, this is an instruction for the log of images. The image can remain on the storage, but if it is not in the log then GCS will not be able to get its URL.

hamishwillee:
@MatejFranceskin @julianoes I have renamed and edited this to make it clear that it is a removal from log and not a deletion of file. I think that is necessary because
- this wasn't absolutely clear to me
- it is quite possible and reasonable you might also want to have an API to actually delete images from media - that can now exist in parallel

Are you OK with the new message name and text? If so I think we are good to merge.

message_definitions/v1.0/common.xml
```diff
@@ -1772,9 +1772,9 @@
         <param index="4" reserved="true" default="NaN"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
-      <entry value="533" name="MAV_CMD_DELETE_IMAGE" hasLocation="false" isDestination="false">
-        <description>Delete image from image log.</description>
-        <param index="1" label="Number" minValue="0" increment="1">Sequence number for image log entry to be deleted.</param>
+      <entry value="533" name="MAV_CMD_UNLOG_IMAGE" hasLocation="false" isDestination="false">
+        <description>Remove image from the image log (no longer accessible using the MAVLink camera API). Note that the image may still be present on actual storage media.</description>
+        <param index="1" label="Number" minValue="0" increment="1">Sequence number for image log entry to be removed.</param>
```
julianoes: I don't understand why you would want that. Deleting a picture file makes sense. Deleting a line from a log, that's not intuitive to me.

julianoes (changes requested):

message_definitions/v1.0/common.xml
hamishwillee: @MatejFranceskin This question is for you.

MatejFranceskin: But what would happen if you remove the image and keep it referenced in a log? How can GCS know that image was removed? We don't want to try to download the images all the time just to get negative ACK to figure out that it was deleted. We would like to figure it out only from the image log.
Let's go back to what I need: I need a command that would remove the image file on the vehicle and from the log. The log count wouldn't be decremented, just the deleted image URL in the log would be set to empty.

MatejFranceskin: Actually what I need is delete image and this can be done via HTTP server that is used to download actual images. In this case what happens to get image log after that is left to implementation. Requesting log entry for deleted image can return "denied" or empty URL.
In this case we can close this PR.

julianoes: I think I agree, yes.

PR #1648, opened 2021-06-10
common.xml: Deprecate MAV_CMD_NAV_PAYLOAD_* [Proposal]

hamishwillee:
@LorenzMeier These messages are problematic: [MAV_CMD_PAYLOAD_CONTROL_DEPLOY](https://mavlink.io/en/messages/common.html#MAV_CMD_PAYLOAD_CONTROL_DEPLOY), [MAV_CMD_PAYLOAD_PREPARE_DEPLOY](https://mavlink.io/en/messages/common.html#MAV_CMD_PAYLOAD_PREPARE_DEPLOY), [MAV_CMD_NAV_PAYLOAD_PLACE](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_PAYLOAD_PLACE)

- It is not clear how they would be used.
- They would be hard to implement or use in a "generic" way because this refers to an arbitrary payload.
- They aren't designed as we would now as "primitives" but are instead somewhat monolithic - ie in the same way you did not particularly like the orbit command.
- They do not appear to have been used in ArduPilot or PX4. They were added by you in https://github.com/mavlink/mavlink/pull/235 (7 years ago) 

I would like to delete or deprecate them. If we are not doing either of those things, then document them (in which case can you tell me who can help me understand their intent).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1041,6 +1041,7 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="94" name="MAV_CMD_NAV_PAYLOAD_PLACE" hasLocation="true" isDestination="true">
+        <deprecated since="2021-06"/>
         <description>Descend and place payload. Vehicle moves to specified location, descends until it detects a hanging payload has reached the ground, and then releases the payload. If ground is not detected before the reaching the maximum descent value (param1), the command will complete without releasing the payload.</description>
         <param index="1" label="Max Descent" units="m" minValue="0">Maximum distance to descend.</param>
         <param index="2">Empty</param>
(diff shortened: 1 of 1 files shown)


auturgy:
ArduPilot is using CMD_NAV_PAYLOAD_PLACE here: https://github.com/ArduPilot/ardupilot/blob/591c7d55c7528c5f578eaf5510ce893bebcbd9a4/ArduCopter/mode_auto.cpp#L498 but not the other two, so suggest we could deprecate PAYLOAD_PREPARE_DEPLOY and CONTROL_DEPLOY.  Obviously the use case they were designed for is important, so figuring out why they're not used might have value.
@peterbarker thoughts?

hamishwillee:
Thanks @auturgy , I have removed `MAV_CMD_NAV_PAYLOAD_PLACE` from the PR.

It looks like  `MAV_CMD_NAV_PAYLOAD_PLACE` was created many years ago and intended to be used as part of a mission - based on it's ID of 94 and the fact IT is a `NAV` COMMAND. 

The other two commands look more like command protocol commands:  MAV_CMD_PAYLOAD_CONTROL_DEPLOY, MAV_CMD_PAYLOAD_PREPARE_DEPLOY
- they don't have `NAV` in the name. 
- their ids are high 30001, 2 indicating they were added a lot later.
- `MAV_CMD_PAYLOAD_PREPARE_DEPLOY` includes the text _if used in MISSION_ITEM (deprecated)_

The intent isn't very clear, but I think they were allowing creating of a kind of semi-manual payload deployment mode where you could batch up a bunch of deployments, and abort them either individually or as a whole. A bit like a mission but recognising that aborting a particular step makes a lot of sense. 

Do you want me to hold off merging while you guys discuss? My leaning is that @LorenzMeier is right - we need more input from people actually doing deployments. I'd also be much happier if the needs were met within a mission.

peterbarker:
@hamishwillee I can't speak for the two you're deprecating here - but covering the payload-place in a mission is problematic as completing the mission item is dependent on sensor input.  In ArduPilot we look for our throttle requirement dropping, indicating our load is on the ground - the vehicle is in a commanded descent while we're looking for that throttle change.  It would be difficult to encapsulate all of the logic here in multiple items.

Now-adays we'd actually probably suggest people create custom LUA mission items to do this sort of thing - moving us even further from a mavlink standard :-(

hamishwillee:
Thanks @auturgy , @peterbarker 

OK, so what I take from this is that the two marked as deprecated aren't used by PX4 or ArduPilot, probably aren't used by anyone else, and would be to ambiguous to implement consistently anyway with their current description. 

The PAYLOAD-PLACE is supported in ArduPilot. It should perhaps be deprecated because it is problematic and there are better ways, but it is supported now. 

So I'll merge this. We should probably still discuss the cargo use case longer term, but not something I'm going to drive without an interested stakeholder.

PR #1649, opened 2021-06-11
common: specify how to deactivate termination

julianoes:
We specify the positive action but not the negative/default action.
Given the implementation in PX4, I suggest to add this to the spec.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1202,7 +1202,7 @@
       </entry>
       <entry value="185" name="MAV_CMD_DO_FLIGHTTERMINATION" hasLocation="false" isDestination="false">
         <description>Terminate flight immediately</description>
-        <param index="1" label="Terminate" minValue="0" maxValue="1" increment="1">Flight termination activated if &gt; 0.5</param>
+        <param index="1" label="Terminate" minValue="0" maxValue="1" increment="1">Flight termination activated if &gt; 0.5, deactivated if &lt; 0.5</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>


auturgy:
How do you recover from a termination event?
This needs a bit of thought and discussion, as in my view the inability to reverse a termination command is a feature, not a bug. A bug would be getting down that codepath accidentally/incorrectly.
There is the possibility/probably we’ve interpreted the scope of “termination” differently.

julianoes:
@auturgy that would be a valid discussion. I would argue there are situations where an "un-termination" makes sense, and others where it's not.

Looking at ardupilot, I see that the param is validated but I'm not exactly sure what would happen if it was not set to 1:

https://github.com/ArduPilot/ardupilot/blob/f09abc705bb8624f1c3817d838f3e213ccba3c1b/libraries/GCS_MAVLink/GCS_Common.cpp#L2845-L2849

hamishwillee:
Added to agenda here https://github.com/mavlink/mavlink/wiki/20210623-Dev-Meeting

My understanding is that often we initiate flight termination via a kill switch, and sometimes this might be triggered accidentally. So on PX4 it is possible to undo/revert the flight termination within 5 seconds, as per https://docs.px4.io/master/en/config/safety.html#kill-switch . So @julianoes I think you are proposing that this undo-case be mirrored for MAVLInk-based flight termination. Is that so that the same use case can be addressed for someone who has a joystick-based controller and wants to assign a kill switch?

We can chat about this in the dev call, just wanted to get might thoughts down

julianoes:
> My understanding is that often we initiate flight termination via a kill switch, and sometimes this might be triggered accidentally.

For the kill *switch* that's correct, however, I think that should be fixed in hardware or process. The kill switch must never be triggered accidentally! (Of course most companies have to learn this the hard way which is why it is the way it is in PX4.)

For a mavlink command the chance to trigger it accidentally should not exist, so un-termination should not be a thing either.
So I would like kill/termination to be what it says which is terminal (pun intended?).

However, my suggestion would reflect what PX4 implements. Right now the behavior seems at least undefined when param1 is not set to 1, so we could document what that means.

hamishwillee:
> However, my suggestion would reflect what PX4 implements. Right now the behavior seems at least undefined when param1 is not set to 1, so we could document what that means.

Fair enough. I've never particularly liked the using param 1 to enable. If you got an enable message that only supports enable then the intent is pretty clear.

I'd be leaning then towards "command is ignored" for other values. Anyway, as you say, let's clarify in the dev call.

julianoes:
As discussed on the call. Let's leave it as is.

PR #1650, opened 2021-06-17
common: add target for timesync

julianoes:
This would be an attempt to enable timesync for more than two nodes.
I'm not sure if this is enough though because the timesync spec/comments are fairly confusing.

FYI @potaito @JonasVautherin

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5321,6 +5321,9 @@
       <description>Time synchronization message.</description>
       <field type="int64_t" name="tc1">Time sync timestamp 1</field>
       <field type="int64_t" name="ts1">Time sync timestamp 2</field>
+      <extensions/>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
     </message>
     <message id="112" name="CAMERA_TRIGGER">
       <description>Camera-IMU triggering and synchronisation message.</description>


TSC21:
So in a multi-component system, who starts the timesync? Who has the "master" clock? I this a multicast message, and a single entity starts the time synchronization with all the others? In a system where all the components communicate with each other, do they synchronize the timestamps between each other, or do we store a master clock + an offset per component?

julianoes:
@TSC21 I have no idea, you tell me.

potaito:
@TSC21 the way we understood the implementation in PX4 autopilot is that both ways work. Either PX4 only fills in its own timestamp and returns the message to the sender: 

https://github.com/PX4/PX4-Autopilot/blob/500c45e79820a852a57a2ebfc0eecc0b1463f344/src/modules/mavlink/mavlink_timesync.cpp#L62-L71

Or it calculates the offset itself:

https://github.com/PX4/PX4-Autopilot/blob/500c45e79820a852a57a2ebfc0eecc0b1463f344/src/modules/mavlink/mavlink_timesync.cpp#L73-L153

The issue is, that the message is currently not only returned to the sender, but sent to every component without the components being able to differentiate who the message is meant for.

TSC21:
> @TSC21 I have no idea, you tell me.

Well I don't have the answer either. I just wanted to start an healthy argument and maybe come-up with a solution that can be added to complement and extend the spec.

julianoes:
Sorry, my answer was a bit blunt, and I was hoping you have the answers :). Let me try again.
What if every component just does a time sync with every other component that it sees? So it would send it out one by one to all components that it knows (from having received heartbeats).

TSC21:
> What if every component just does a time sync with every other component that it sees? So it would send it out one by one to all components that it knows (from having received heartbeats).

So each component stores it's own table with components and offsets? I think that might be beneficial, in the sense that there's no system time base, and every new ad-hoc "node" will timesync with other nodes independently, But I also see this as problematic, because it means that each node needs to be capable of doing this and storing this table. A centralized node where this happen would mitigate this, but this also means that there needs to be an initial mechanism that chooses the node that sets a "master" clock and stores the offsets between all nodes.

JonasVautherin:
> The issue is, that the message is currently not only returned to the sender, but sent to every component without the components being able to differentiate who the message is meant for.

Yep, that was my understanding when I looked into it. Which IMO means that it cannot be done without changing MAVLink, and therefore it means that it's a good moment to think about the ideal way (instead of being limited to building on top of the existing one).

> A centralized node where this happen would mitigate this, but this also means that there needs to be a mechanism that chooses the node that sets a "master" clock and stores the offsets between all nodes.

Maybe naive, but isn't there always one central component as the guy who sends the signals to the actuators? If you have only one instance of the autopilot, that's this guy, and if you have multiple instances of it... well does anything work at all with two instances of an autopilot? I would bet it completely explodes in that case :sweat_smile:. And I would guess that this kind of redundancy should be transparent (i.e. if you have 2 autopilot instances, they should "look" like 1 instance on the mavlink network, so QGC only ever receives heartbeats from one unified autopilot).

TL;DR: why not synchronizing everyone with the autopilot? I don't really get the point of decentralizing that, given that every node has to know who is the autopilot.

potaito:
> > The issue is, that the message is currently not only returned to the sender, but sent to every component without the components being able to differentiate who the message is meant for.
> 
> Yep, that was my understanding when I looked into it. Which IMO means that it cannot be done without changing MAVLink, and therefore it means that it's a good time to think about the ideal way (instead of being limited to building on top of the existing one).
> 
> > A centralized node where this happen would mitigate this, but this also means that there needs to be a mechanism that chooses the node that sets a "master" clock and stores the offsets between all nodes.
> 
> Maybe naive, but isn't there always one central component as the guy who sends the signals to the actuators (a.k.a. autopilot)? If you have only one instance of the autopilot, that's this guy, and if you have multiple instances of it... well does anything work at all with two instances of an autopilot? I would bet it completely explodes in that case sweat_smile. And I would guess that this kind of redundancy should be transparent (i.e. if you have 2 autopilot instances, they should "look" like 1 instance on the mavlink network, so QGC only ever receives heartbeats from one unified autopilot).
> 
> TL;DR: why not synchronizing everyone with the autopilot? I don't really get the point of decentralizing that, given that every node has to know who is the autopilot.

Sounds good. I can't really think of any arguments against having the autopilot as the time master, except perhaps when you want to sync two components that are wired together, but not wired to the autopilot. Because then the two components would have less delay when just talking to each other rather than going through the autopilot. But I can't really think of a hardware setup where the autopilot is not wired to other components like gimbal, camera, companion computer etc.

hamishwillee:
FYI how it works now by @mhkabir is here: https://github.com/mavlink/mavlink-devguide/issues/194
@mhkabir Given your experience with the original design anything to add?

Great to kick off discussion. We should involve @auturgy and before this is merged, have a full description on the interactions.

julianoes:
@JonasVautherin so if it is the autopilot doing the syncing with everyone, what message and fields would that require? Would it work with the target that I added here?

JonasVautherin:
> @JonasVautherin so if it is the autopilot doing the syncing with everyone, what message and fields would that require? Would it work with the target that I added here?

I think it would (given my understanding from https://github.com/mavlink/mavlink-devguide/issues/194 and @AlexisTM drawing in https://github.com/mavlink/MAVSDK/issues/1170).

It also seems to me that the one initiating the protocol is the one computing the time offset, and therefore both sides have to initiate it (or am I missing something?). Not sure what it implies :thinking:... maybe that's fine, and with your addition a component could actually synchronize with all the others if it wants to?

mhkabir:
> the one initiating the protocol is the one computing the time offset, and therefore both sides have to initiate it

Yes this was the intent

JonasVautherin (approved):
> Yes this was the intent

In that case, I think that this PR solves the limitation of the timesync message: by having a target sysid:compid, it should work in a system with more than 2 components :+1:.

peterbarker:
ArduPilot currently encodes the system ID in the bottom bits of the timestamp.

```
    // we add in our own system id try to ensure we only consider
    // responses to our own timesync request messages
    return AP_HAL::micros64()*1000LL + mavlink_system.sysid;
```

julianoes:
@peterbarker so would you be in favor of this change then?

peterbarker:
On Mon, 21 Jun 2021, Julian Oes wrote:

> @peterbarker so would you be in favor of this change then?

At first glance I thought this a much better solution that what I did in 
ArduPilot.  I can't recall now whether I considered doing this myself :-)

However.

The ArduPilot encode-sysid-in-timestamp thing works even across 
non-mavlink2-pure links.  Without having that hack in place you'd have to 
reject any return timesync message that have the extension fields as zero 
- meaning a loss of functionality.  So we'd have to retain the existing 
hack (probably corrected for not having the component ID in there...).  If 
you *didn't* hve the hack *and were working across a non-mavlink2 link 
then you might end up consuming other's responses.

So I'm not sure this *can* be retrofitted....

julianoes:
> So I'm not sure this *can* be retrofitted....

Ok fair. So this sounds like a separate new message would be better than.

hamishwillee:
Yes. I think a new message is better in this case. 

Want to update the PR @julianoes?

Is there anything else required for timesync? Presumably we are requiring this always to be targeted to a particular node, or would it be allowable to broadcast and process multiple returned responses?

julianoes:
I will open a PR to create a new timesync message.

PR #1651, opened 2021-06-17
common.xml: add MAV_CMD_RUN_PREARM_CHECKS

peterbarker:
Allows other mavlink nodes to request that prearm checks be run immediately.

Closes a problem where to work out whether a system is armable you have to try to arm it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1640,6 +1640,9 @@
         <param index="1" label="Arm" minValue="0" maxValue="1" increment="1">0: disarm, 1: arm</param>
         <param index="2" label="Force" minValue="0" maxValue="21196" increment="21196">0: arm-disarm unless prevented by safety checks (i.e. when landed), 21196: force arming/disarming (e.g. allow arming to override preflight checks and disarming in flight)</param>
       </entry>
+      <entry value="401" name="MAV_CMD_RUN_PREARM_CHECKS" hasLocation="false" isDestination="false">
+        <description>Instructs system to run pre-arm checks.  This command should return MAV_RESULT_TEMPORARILY_REJECTED in the case the system is armed, otherwse MAV_RESULT_ACCEPTED.  Note that the return value from executing this command does not indicate whether the vehicle is armable or not, just whether the system has successfully run/is currently running the checks.  The result of the checks is reflected in the SYS_STATUS message.</description>
+      </entry>
       <entry value="405" name="MAV_CMD_ILLUMINATOR_ON_OFF" hasLocation="false" isDestination="false">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1641,7 +1641,7 @@
         <param index="2" label="Force" minValue="0" maxValue="21196" increment="21196">0: arm-disarm unless prevented by safety checks (i.e. when landed), 21196: force arming/disarming (e.g. allow arming to override preflight checks and disarming in flight)</param>
       </entry>
       <entry value="401" name="MAV_CMD_RUN_PREARM_CHECKS" hasLocation="false" isDestination="false">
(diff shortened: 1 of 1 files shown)


hamishwillee:
> Instructs system to run pre-arm checks. This command should return MAV_RESULT_TEMPORARILY_REJECTED in the case the system is armed, otherwse MAV_RESULT_ACCEPTED. Note that the return value from executing this command does not indicate whether the vehicle is armable or not, just whether the system has successfully run/is currently running the checks. The result of the checks is reflected in the SYS_STATUS message.

So to be clear, you're about to ARM and you want to know if prearm checks would succeed. Is this to improve the user experience because it looks like they can arm based on GPS but they can't really?

So you send this command. If you get back MAV_RESULT_TEMPORARILY_REJECTED you know you're armed (you probably did anyway). If you get back ACCEPTED you know that the prearm checks are being run, so you start monitoring SYS_STATUS. Once you see that `MAV_SYS_STATUS_PREARM_CHECK` is true you know arming will succeed?

Some questions/statements:
- How do you know when the checks have completed - ie what if they fail?
- Can the checks succeed but the user go to press the arm button and it still fails?
- This would be better done as a long running command or the proposed COMMAND_PROGRESS https://github.com/mavlink/mavlink/pull/1635 -

Upshot, I am very happy with the idea of the command, but I think we should use an explicit and standardised method to indicate actual command completion. For compatibility reasons, that would probably best be COMMAND_PROGRESS.

@julianoes FYI

peterbarker:
On Thu, 17 Jun 2021, Hamish Willee wrote:

> So to be clear, you're about to ARM and you want to know if prearm checks
> would succeed. Is this to improve the user experience because it looks like
> they can arm based on GPS but they can't really?

Actually, this is more like, "SYS_STATUS.<PREARM_CHECKS>" tells me I can't 
arm.  So run the checks *now* so I get my messages *now*.

> So you send this command. If you get back MAV_RESULT_TEMPORARILY_REJECTED
> you know you're armed (you probably did anyway). If you get back ACCEPTED
> you know that the prearm checks are being run, so you start monitoring
> SYS_STATUS. Once you see that MAV_SYS_STATUS_PREARM_CHECK is true you know
> arming will succeed?

Well, you always know the latter.  This just makes tha autopilot check 
right now.  So if you're only running your prearm checks on your autopilot 
at once/30 seconds, an impatient user can poke it straight away to get any 
details.

>  *  How do you know when the checks have completed - ie what if they fail?

You don't - but if they fail you could expect the PREARM flag to stay 
false.

>  *  Can the checks succeed but the user go to press the arm button and it
>     still fails?

Yes, this is always going to be a race condition.  For example, between 
sending a SYS_STATUS with the prearm health set to healthy and receiving 
an arm command the vehicle's velocity estimate might have dropped out of 
acceptable ranges.

>  *  This would be better done as a long running command or the proposed
>     COMMAND_PROGRESS #1635 -
> 
> Upshot, I am very happy with the idea of the command, but I think we should
> use an explicit and standardised method to indicate actual command
> completion. For compatibility reasons, that would probably best be
> COMMAND_PROGRESS.

OK, but here's the thing.  You've asked the vehicle to run its prearm 
checks - I was trying to be careful about the naming there.  If it doesn't 
support that command it will return you an unrecognised-command failure 
code saying it doesn't know about that command.  If it does recognise it 
then (IMO) the return code should revolve around the success or failure of 
the command, rather than success or failure of the prearm checks.  That's 
why this is "MAV_CMD_RUN_PREARM_CHECKS" rather than 
"MAV_CMD_PREARM_VEHICLE" - the latter I could see as a long-running 
command as you describe.

julianoes (approved):
I think this makes sense.

However, I would like to get @bkueng's opinion as well.

LorenzMeier (approved):

bkueng:
It's a possible approach, but personally I'd rather run the checks at ~1Hz and report immediately on change.

hamishwillee:
> It's a possible approach, but personally I'd rather run the checks at ~1Hz and report immediately on change.

FWIW me too. I'm also far more interested in what is blocking it that getting it to try to arm. Still, Lorenz was OK with this.

peterbarker:
On Mon, 21 Jun 2021, Beat Küng wrote:

> It's a possible approach, but personally I'd rather run the checks at ~1Hz and report immediately on change.

In fact ArduPilot vehicles do run the checks at some interval - but some 
of them don't bombard the user with all of the failure message strings 
(the latter being a different argument I think we've had elsewhere ;-).

We're reviewing this inconsistency between the AP vehicle types, but in 
the interim this allows a user to get those failure strings from the 
less-wordy vehicle in a safer manner than attempting to arm the vehicle!

PR #1659, opened 2021-07-21
commom.xml: add FUEL_TANK

PonomarevDA:
According to the discussion opened in [PX4 issue #17861](https://github.com/PX4/PX4-Autopilot/issues/17861), there is no any mavlink message that has info about current level of fuel tank (only fuel_consumed and fuel_flow from [EFI_STATUS](https://mavlink.io/en/messages/common.html#EFI_STATUS)). This PR added new mavlink message `FUEL_TANK` similar to [ uavcan.equipment.ice.FuelTankStatus](https://legacy.uavcan.org/Specification/7._List_of_standard_data_types/#fueltankstatus).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6767,6 +6767,16 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
+    <message id="420" name="FUEL_TANK">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description></description>
+      <field type="uint8_t" name="available_fuel_volume_percent">Available fuel volume percent, from 0% to 100%</field>
+      <field type="float" name="available_fuel_volume_cm3">Available fuel volume, centimeter^3</field>
+      <field type="float" name="fuel_consumption_rate_cm3pm">Estimate of the current fuel consumption rate, (centimeter^3)/minute</field>
+      <field type="float" name="fuel_temperature">Fuel temperature, kelvin</field>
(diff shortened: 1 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6767,6 +6767,16 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
+    <message id="420" name="FUEL_TANK">
```
hamishwillee, line 6770: Thanks. IF accepted
- message should go in development.xml until it is implemented and verified
- Description required.
- [message fields should have units](https://mavlink.io/en/guide/xml_schema.html#messages). The names should not include the units and nor should the description. So below you would have
   ```
  <field type="uint8_t" name="fuel_remaining" units="%">Available fuel volume</field>
   ```
- Typically you would not have remaining fuel both as a percent and as a volume. Instead you'd provide the maximum value and the remaining volume - i.e. send the minimum and allow people to calculate the form they want. 
- We typically have the id as the first value, but this is not "required"

PonomarevDA: fixed

hamishwillee:
Thanks for the fixes. 

We discussed this in the MAVLink call. ArduPilot [handled this case](https://github.com/ArduPilot/ardupilot/pull/11951) using the BATTERY_STATUS message as a "proxy". The option of having a dedicated message is definitely more clear, but there is a whole lot of code around warnings and safety that did not have to be re-written my doing this hack. It MIGHT be that PX4 will do similarly, in which case we should instead update the docs around battery to explain how you address fuel tank cases.

I'll add a comment on the original issue to see what Daniel thinks. This should be left open until that discussion converges.

hamishwillee:
Thanks @PonomarevDA 

This has been superseded by https://github.com/mavlink/mavlink/pull/2107. I'm sorry we did not progress this as the time. I turns out that the there is now more general support for using this kind of approach rather than battery status.

PR #1661, opened 2021-07-28
common: MAV_CMD_DO_SET_ACTUATOR - remove WIP

hamishwillee:
This was originally agreed in dev call and is now supported in a released version of PX4. I have therefore removed the WIP tagging.

@auturgy Any comments before merging?


auturgy:
Go ahead.

PR #1662, opened 2021-07-29
add MAV_CMD_DO_ADSB_OUT_IDENT 

magicrub:
Add new MAV_CMD_DO_ADSB_OUT_IDENT to trigger an ADSB/ModeA/C/S Transponder IDENT.

The number 10001 was chosen because the uAvionix.xml file uses 10000-10099 but there's no MAV_CMDs in there. This was originally in that file but I pulled it out to common.xml. I think the same number is still OK to use though since it's not a msg_id.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2065,6 +2065,16 @@
         <param index="7">Reserved (set to 0)</param>
       </entry>
       <!-- END of UAVCAN range -->
+      <entry value="10001" name="MAV_CMD_DO_ADSB_OUT_IDENT" hasLocation="false" isDestination="false">
+        <description>Trigger the start of an ADSB-out IDENT. This should only be used when requested to do so by an Air Traffic Controller in controlled airspace. This starts the IDENT which is then typically held for 18 seconds by the hardware per the Mode A, C, and S transponder spec.</description>
+        <param index="1">Reserved (set to 0)</param>
+        <param index="2">Reserved (set to 0)</param>
+        <param index="3">Reserved (set to 0)</param>
+        <param index="4">Reserved (set to 0)</param>
+        <param index="5">Reserved (set to 0)</param>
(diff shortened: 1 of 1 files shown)


magicrub:
Ardupilot/mavlink PR is https://github.com/ArduPilot/mavlink/pull/215

hamishwillee:
@magicrub So from a MAVLink point of view message is fine 
- there might have been a near-clash with slugs.xml dialect, but that was recently removed. (it had range 10001 - 11999)
- You don't need to define empty parameters - these are automatically then created as "Reserved" with a default value of 0. If you ever thing you will need to assign a parameter that might have a value of `NaN` you should explicitly define that now - see https://mavlink.io/en/guide/define_xml_element.html#reserved

FMI ...
It is not clear to me what  `ADSB-out IDENT` means from the linked docs. My understanding is that an ADSB broadcasts position and identify information all the time when turned on. Is this IDENT separate to that? What is the flow? - i.e. what is sending this message - GCS, ADSB transponder? What is receiving it? Are there any other mavlink messages/interactions involved?

auturgy:
> information all the time when turned on. Is this IDENT separate to that? What

Closely related.  You may be directed by ATC, for instance, to "Squawk IDENT" if they have you on primary radar or are in voice comms, but don't have a corresponding transponder appearing on SSR / ADSB / IFF.

magicrub:
@hamishwillee would you like me to include this URL in the description? https://en.wikipedia.org/wiki/Transponder_(aeronautics)

hamishwillee:
Just to summarise the dev call. This makes perfect sense. No need to add additional information, but would be better replace "Empty" with "Reserved (set to 0)" in the unused parameters as "Reserved" has mavlink meaning, and Empty does not.

magicrub:
@hamishwillee updated per our conversation. Ready to merge!

hamishwillee:
@magicrub Done. Thanks :-)

PR #1671, opened 2021-08-12
Add Mission Manager Component ID

MatejFranceskin:
Mission manager is an optional component on the vehicle that implements mission upload and download protocol.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -635,6 +635,9 @@
       <entry value="198" name="MAV_COMP_ID_PAIRING_MANAGER">
         <description>Component that manages pairing of vehicle and GCS.</description>
       </entry>
+      <entry value="199" name="MAV_COMP_ID_MISSION_MANAGER">
+        <description>Component that manages missions on the vehicle.</description>
+      </entry>
       <entry value="200" name="MAV_COMP_ID_IMU">
         <description>Inertial Measurement Unit (IMU) #1.</description>
       </entry>


message_definitions/v1.0/minimal.xml
```diff
@@ -635,6 +635,9 @@
       <entry value="198" name="MAV_COMP_ID_PAIRING_MANAGER">
         <description>Component that manages pairing of vehicle and GCS.</description>
       </entry>
+      <entry value="199" name="MAV_COMP_ID_MISSION_MANAGER">
```
hamishwillee, line 638: 1. What do you mean by manage, and how does this differ from [MAV_COMP_ID_MISSIONPLANNER](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_MISSIONPLANNER)?
2. Why do you need a specific component id for this? Is it because you are going to have a separate heartbeat-emitting component on the flight controller that will process commands addressed to this component?

I'm not trying to complicate things - just want to make sure I know the answers if this every comes up in discussion. 

Personally I don't think we should have any allocated IDs - they should be assigned to nodes dynamically. The only benefit to allocating is that it is a simple way to avoid clashes if there can only ever be one instance of something on a vehicle.

hamishwillee: PS Mav call on tonight if you want to come and chat about it quickly.

MatejFranceskin: Sorry for late reply - I was absent last time. Will join for the next meeting.
Idea behind mission manager is to have a component on the vehicle that can take over mission control and handle more complex mission tasks then the flight controller. If ground control station detects its heartbeat then the mission would be uploaded to mission manager instead of flight controller.

hamishwillee: Item added to agenda: https://github.com/mavlink/mavlink/wiki/20210901-Dev-Meeting

Just a few things that occur to me:
- you will need corresponding MAV_TYPE, and you should use the MAV_TYPE to identify the component.
- How does GCS know whether to send to flight stack or companion?
- Conflict resolution - ie. if multiple things advertise themselves as the particular component or TYPE
- What extra can the manager do? How does it work? Can it interact with obstacle avoidance etc also running on the companion?

hamishwillee: @julianoes Any comments on this one. Essentially moves mission control to the companion. I suspect some of those other things I comment on will also be relevant and I think I'd like to see a bit more of a design.

julianoes: I see this is a bigger change and a component ID alone will probably not be enough to enable that use case.
And the term mission manager is also not clear to a reader in my opinion as it's not clear what the difference to a mission planner is.

MatejFranceskin:
Since we're not going to work on this soon (this was related to mavlink FTP mission upload to the vehicle which is postponed) I am going to close it for now. We will come back to it when we will actually implement it.

hamishwillee:
Thanks @MatejFranceskin - FWIW this sounds like a cool project.

PR #1673, opened 2021-08-16
represent BASE_MODE field in messages SET_MODE and MAV_CMD_DO_SET_MODE

asmyasnikov:
Attribute bitmask used for my tool [go-mavgen](https://github.com/asmyasnikov/go-mavlink) for serialise message to string as bitmask array of MAV_MODE_FLAG bits. But current definition interpret BASE_MODE as enum with switch-case rules. Ardupilot and PX4 interprets BASE_MODE as bitmap value

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1112,7 +1112,7 @@
       </entry>
       <entry value="176" name="MAV_CMD_DO_SET_MODE" hasLocation="false" isDestination="false">
         <description>Set system mode.</description>
-        <param index="1" label="Mode" enum="MAV_MODE">Mode</param>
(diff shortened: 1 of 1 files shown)


hamishwillee:
In order to accept this we would need **verification** that this would not break any of the deployed systems. Minimally that means perhaps building the library and slotting it into PX4 and Mission planner builds. If you can verify that no changes at all had to be made in PX4 or Mission Planner builds corresponding to this message change then we'd consider this further.

In addition, this fails the tests because the `display` attribute is not allowed on parameters - see https://mavlink.io/en/guide/xml_schema.html#param . Before this could be accepted we'd have to modify the XSD. 

I'm closing for now, but feel free to reopen if you can provide the following changes and verification.

asmyasnikov:
Ok, I revert appending display attribute to param - you are write

PR #1674, opened 2021-08-16
Add additional axis to Manual control, support additional buttons.

Williangalvani:
Certain vehicles are capable of simultaneous control of 6dofs, but are unable to use the functionality because the existing form of the MANUAL_CONTROL message limits us to 4 axes with joystick control.

Support for additional buttons has been requested by several QGC users, particularly since some existing game controllers support more than 16 buttons (e.g. PS4, PS5, and [Logitech X56 H.O.T.A.S.](https://www.logitechg.com/pt-br/products/space/x56-space-flight-vr-simulator-controller.945-000058.html)). We also have users developing custom controllers for industrial clients, who frequently want to include several additional inputs beyond the base 16.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4877,18 +4877,18 @@
       <field type="uint8_t" name="on_off">1 stream is enabled, 0 stream is stopped.</field>
     </message>
     <message id="69" name="MANUAL_CONTROL">
-      <description>This message provides an API for manually controlling the vehicle using standard joystick axes nomenclature, along with a joystick-like input device. Unused axes can be disabled an buttons are also transmit as boolean values of their </description>
+      <description>This message provides an API for manually controlling the vehicle using standard joystick axes nomenclature, along with a joystick-like input device. Unused axes can be disabled and buttons states are transmitted as individual on/off bits of a bitmask</description>
       <field type="uint8_t" name="target">The system to be controlled.</field>
       <field type="int16_t" name="x" invalid="INT16_MAX">X-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to forward(1000)-backward(-1000) movement on a joystick and the pitch of a vehicle.</field>
       <field type="int16_t" name="y" invalid="INT16_MAX">Y-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to left(-1000)-right(1000) movement on a joystick and the roll of a vehicle.</field>
       <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a separate slider movement with maximum being 1000 and minimum being -1000 on a joystick and the thrust of a vehicle. Positive values are positive thrust, negative values are negative thrust.</field>
       <field type="int16_t" name="r" invalid="INT16_MAX">R-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a twisting of the joystick, with counter-clockwise being 1000 and clockwise being -1000, and the yaw of a vehicle.</field>
-      <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
+      <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' 0-15 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
       <extensions/>
       <field type="uint16_t" name="buttons2">A bitfield corresponding to the joystick buttons' 16-31 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 16.</field>
       <field type="uint8_t" name="enabled_extensions">Set bits to 1 to indicate which of the following extension fields contain valid data: bit 0: pitch, bit 1: roll.</field>
-      <field type="int16_t" name="s" invalid="INT16_MAX">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Set to zero if invalid.</field>
-      <field type="int16_t" name="t" invalid="INT16_MAX">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom.  Set to zero if invalid.</field>
+      <field type="int16_t" name="s" invalid="0">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Valid if bit 0 of enabled_extensions is set.</field>
+      <field type="int16_t" name="t" invalid="0">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom. Valid if bit 1 of enabled_extensions is set.</field>
     </message>
     <message id="70" name="RC_CHANNELS_OVERRIDE">
       <description>The RAW values of the RC channels sent to the MAV to override info received from the RC radio. The standard PPM modulation is as follows: 1000 microseconds: 0%, 2000 microseconds: 100%. Individual receivers/transmitters might violate this specification.  Note carefully the semantic differences between the first 8 channels and the subsequent channels</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4887,8 +4887,8 @@
       <extensions/>
       <field type="uint16_t" name="buttons2">A bitfield corresponding to the joystick buttons' 16-31 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 16.</field>
       <field type="uint8_t" name="enabled_extensions">Set bits to 1 to indicate which of the following extension fields contain valid data: bit 0: pitch, bit 1: roll.</field>
-      <field type="int16_t" name="s" invalid="0">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Valid if bit 0 of enabled_extensions is set.</field>
-      <field type="int16_t" name="t" invalid="0">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom. Valid if bit 1 of enabled_extensions is set.</field>
+      <field type="int16_t" name="s">Pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Valid if bit 0 of enabled_extensions field is set. Set to 0 if invalid.</field>
+      <field type="int16_t" name="t">Roll-only-axis, normalized to the range [-1000,1000]. Generally corresponds to roll on vehicles with additional degrees of freedom. Valid if bit 1 of enabled_extensions field is set. Set to 0 if invalid.</field>
     </message>
     <message id="70" name="RC_CHANNELS_OVERRIDE">
       <description>The RAW values of the RC channels sent to the MAV to override info received from the RC radio. The standard PPM modulation is as follows: 1000 microseconds: 0%, 2000 microseconds: 100%. Individual receivers/transmitters might violate this specification.  Note carefully the semantic differences between the first 8 channels and the subsequent channels</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4884,6 +4884,11 @@
       <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a separate slider movement with maximum being 1000 and minimum being -1000 on a joystick and the thrust of a vehicle. Positive values are positive thrust, negative values are negative thrust.</field>
       <field type="int16_t" name="r" invalid="INT16_MAX">R-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a twisting of the joystick, with counter-clockwise being 1000 and clockwise being -1000, and the yaw of a vehicle.</field>
       <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
+      <extensions/>
+      <field type="uint16_t" name="buttons2">A bitfield corresponding to the joystick buttons' 16-31 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 16.</field>
+      <field type="uint8_t" name="enabled_extensions">Set bits to 1 to indicate which of the following extension fields contain valid data: bit 0: pitch, bit 1: roll.</field>
+      <field type="int16_t" name="s" invalid="INT16_MAX">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Set to zero if invalid.</field>
+      <field type="int16_t" name="t" invalid="INT16_MAX">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom.  Set to zero if invalid.</field>
     </message>
     <message id="70" name="RC_CHANNELS_OVERRIDE">
       <description>The RAW values of the RC channels sent to the MAV to override info received from the RC radio. The standard PPM modulation is as follows: 1000 microseconds: 0%, 2000 microseconds: 100%. Individual receivers/transmitters might violate this specification.  Note carefully the semantic differences between the first 8 channels and the subsequent channels</description>


message_definitions/v1.0/common.xml
```diff
@@ -4877,13 +4877,17 @@
...
       <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a separate slider movement with maximum being 1000 and minimum being -1000 on a joystick and the thrust of a vehicle. Positive values are positive thrust, negative values are negative thrust.</field>
       <field type="int16_t" name="r" invalid="INT16_MAX">R-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a twisting of the joystick, with counter-clockwise being 1000 and clockwise being -1000, and the yaw of a vehicle.</field>
-      <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
+      <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' 0-15 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
```
hamishwillee, line 4886: This line is a good changed, irrespective of whatever else happens.

hamishwillee:
Thanks @Williangalvani , A few thoughts ...
1. New fields add to the size of a message unless they can be truncated when not needed. 
2. Extensions that are not understood by the sender arrive at the recipient as 0 filled values. For that reason by default extensions should generally use 0 to mean unused or unchanged. You can't use 0 for a meaningful value unless a GCS has some other way to know that the sender does not support the new form of the message. 
3. To handle that, where 0 is a valid value, we sometimes have to do some crappy workaround like saying "0 means not supported. Any value other than 0 and < 0.1 (say) means zero". Sometimes we have another field that must be set as non-zero to indicate that the other fields have valid values - so the GCS knows that it is working with new style mesage.
4. Where messages can't be guaranteed to be truncated we may be better off with a second message or a replacement message.
5. Best to think about future proofing now - what other DOF might come up? Will we need more buttons in future? Are there other types of manual control data?

Upshot
1. If we extend this, 
   - I would put the buttons2 first, since that is more likely to be used than the extra DOF.
   - I would make 0 mean "not supported" for the other two new fields - even though it is inconvenient since I assumed these will be sent/required much less regularly. 
2. If we don't extend this we need to think about either a or multiple messages. Would it matter to a recipient if it got the information in different messages. i.e. what if we did:

```xml
<message id="69" name="MANUAL_CONTROL">
      <description>This message provides an API for manually controlling the vehicle using standard joystick axes nomenclature, along with a joystick-like input device. Unused axes can be disabled an buttons are also transmit as boolean values of their </description>
      <field type="uint8_t" name="target">The system to be controlled.</field>
      <field type="int16_t" name="x" invalid="INT16_MAX">X-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to forward(1000)-backward(-1000) movement on a joystick and the pitch of a vehicle.</field>
      <field type="int16_t" name="y" invalid="INT16_MAX">Y-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to left(-1000)-right(1000) movement on a joystick and the roll of a vehicle.</field>
      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a separate slider movement with maximum being 1000 and minimum being -1000 on a joystick and the thrust of a vehicle. Positive values are positive thrust, negative values are negative thrust.</field>
      <field type="int16_t" name="r" invalid="INT16_MAX">R-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a twisting of the joystick, with counter-clockwise being 1000 and clockwise being -1000, and the yaw of a vehicle.</field>
      <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' 0-15 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
      <extensions/>
      <field type="uint16_t" name="buttons2">A bitfield corresponding to the joystick buttons' 16-31 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 16.</field>
    </message>
```
and
```xml
<message id="?" name="MANUAL_CONTROL_EXT">
      <description>This message provides ...</description>
      <field type="int16_t" name="pitch" invalid="INT16_MAX">pitch-only-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to pitch on vehicles with additional degrees of freedom</field>
      <field type="int16_t" name="roll" invalid="INT16_MAX">roll-only-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to roll on vehicles with additional degrees of freedom</field>
    </message>
```

ES-Alexander:
> 5\. Best to think about future proofing now - what other DOF might come up? Will we need more buttons in future? Are there other types of manual control data?

I'm yet to see a controller with more than I think 36 individual buttons, but that doesn't include shift functionality. Taking the logitech X56 as an example of the extreme it says it has 5 hats (x4 positions each) and 31 programmable buttons, which is a minimum of 51 button controls, and if you wanted to include shift functionality for that it'd be double or triple. In saying that, presumably some controls would be better placed doing things with the general computer than necessarily going to the autopilot. I can't imagine it would be useful having more than about 50 buttons sent to the autopilot. You could get 48 in with a `uint32 buttons2`, or 80 with a `uint64`.

For control axes it's a bit more vague. The X56 has 13, including several which are rotating nobs. That kind of thing can be useful for adjusting servo-type controls absolutely rather than relatively, which includes components like lights and grippers (or an articulating vehicle for the really adventurous), and also things like control gain (which in an ideal world might have independent values for different axes). Perhaps worth noting that a board like the Pixhawk 4 has 16 servo outputs (8 main, 8 aux), so if 4 were used in a quad-rotor setup that would leave 12 for external control purposes. 

For an extreme control scenario I suppose then you could have 12 external axes, 6 for 6dof motion control, and another 6 for gain, which ends up at 24. That definitely seems a bit over the top, but perhaps half or more could be lower resolution (e.g. 256 control points (8-bits) for some lights is well above what's necessary, so one 16-bit value could potentially be used to control four 4-bit axes, which still gives 16 control points for those). Having a smallish set of 16-bit values is likely fine - perhaps something like a `uint8` for target and then 2 `int16`s for additional axis control, and 4 `uint16`s for less well-defined purposes. Extra axes that are sharing those fields is then outside what mavlink cares about, and the minor extra processing required to extract the values isn't unreasonable for 'secondary' functionality.

From a programming standpoint it's perhaps marginally easier if everything is in the one message, but it definitely doesn't seem unreasonable to split out a second one, especially if there will be several extra axes that many users won't require. People that need significant extra control will already need to be doing a bunch of programming for it, so can likely deal with the overhead of handling an additional message.

hamishwillee:
@Williangalvani We discussed this in the dev call last night. The points raised by @ES-Alexander seem really relevant. 

In particular sending button presses to the flight stack to resolve into functions is not very efficient. We might instead have the button-to-function mapping managed in the GCS and simply send the appropriate commands. This would prevent some other system doing the mapping using the manual control message, but then button presses could be sent in a separate message that gets sent "IFF needed".

Further we should minimise the size of these messages as far as we can because they do need to be sent at quite high rate on some vehicle. So it might make sense (as discussed) to split out extra DOF's if we can. 

Did you guys already do an analysis of the alternatives? We decided in the dev call that @auturgy will get in contact with you @Williangalvani to discuss next steps.

patrickelectric:
@hamishwillee, just to make some points clear.

The idea of the extra two axis is to have a full 6dof control over the vehicle attitude, allowing directly 
thruster or control input for all 6 degrees, maybe this is already a good extension for MANUAL_CONTROL and the only problem here is how to deal with buttons. 

I agree that moving the button mapping to GCS makes more sense, for some reasons:
- It makes the architecture simpler since it will only be handle by GCS
- The shift functionality will also be done by the GCS (if it sends a message for this specific command)
- It makes easy to control multiple vehicles with the same GCS/computer (since the control map will be done by GCS and not the vehicle)
- GCS should be able to export the control mapping for others computers

Maybe a message with:
```xml
<message id="?" name="FUNCTION_CONTROL">
      <description>Set the state of the a desired function</description>
      <field type="uint8_t" name="target">The system to be controlled.</field>
      <field type="int16_t" name="control_id" invalid="INT16_MAX">Identification number for the control</field>
      <field type="int16_t" name="value" invalid="INT16_MAX">Value for the desired control</field>
</message>
```
But..

For the GCS to know all the possible control options: it'll be necessary to have a fetch mechanism.
For the GCS to know the description of such controls, if it's a button, slider, and etc: a xml (like parameters) may be necessary.

So, the complexity of such approach starts to grow, but I believe that can makes everything easier in the long run.
We need to see the tradeoff of each approach and discuss that more.

auturgy:
I think that moving the mapping to the GCS is the longer term better approach.  In the interim, a simple extension for "instance" as a bool or int might achieve what this message seeks, without adding extensions we can't trim - ie instance 0 = current message, instance 1, 2, etc = the next range for each field.  This would mean zero impact on most users, but would have a bandwidth cost for those that do implement, rather than making everyone pay for fields they don't use.  Thoughts?

message_definitions/v1.0/common.xml
```diff
@@ -4877,13 +4877,17 @@
...
+      <extensions/>
+      <field type="int16_t" name="pitch" invalid="INT16_MAX">pitch-only-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to pitch on vehicles with additional degrees of freedom</field>
```
tridge, line 4888: could add a flags field with 2 bits that indicate that pitch and roll are valid. That way we don't need INT16_MAX to be special

tridge (changes requested):

message_definitions/v1.0/common.xml
auturgy: so if the flag correlating to a particular extension field is 0, the field is invalid and hence a zero value can be sent and trimmed without precluding the use of zero in a valid control message.  Makes sense.

ES-Alexander:
> I think that moving the mapping to the GCS is the longer term better approach. In the interim, a simple extension for "instance" as a bool or int might achieve what this message seeks, without adding extensions we can't trim - ie instance 0 = current message, instance 1, 2, etc = the next range for each field. This would mean zero impact on most users, but would have a bandwidth cost for those that do implement, rather than making everyone pay for fields they don't use. Thoughts?

To me this seems really clean. An "instance" `bool` extension would double the existing possibilities, and a `uint8` would add up to 255x the current options without requiring further modifications to mavlink itself, just an update of docs when one of the instance values gets reserved.

auturgy:
An issue was identified during the ArduPilot devcall with that approach: if sent over mavlink1, the extensions wouldn’t get sent but the message would still be valid, leading to undesirable behaviour. Tridge’s suggestion overcomes that: if the extensions aren’t sent you lose the additional fields, but original behaviour is maintained/predictable.

Regards,

James

message_definitions/v1.0/common.xml
hamishwillee: So the "interim" proposal is something like (?):
```xml
<message id="69" name="MANUAL_CONTROL">
 <description>This message provides an API for manually controlling the vehicle using standard joystick axes nomenclature, along with a joystick-like input device. Unused axes can be disabled and buttons states are transmitted as individual on/off bits of a bitmask</description>
 <field type="uint8_t" name="target">The system to be controlled.</field>
 <field type="int16_t" name="x" invalid="INT16_MAX">X-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to forward(1000)-backward(-1000) movement on a joystick and the pitch of a vehicle.</field>
 <field type="int16_t" name="y" invalid="INT16_MAX">Y-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to left(-1000)-right(1000) movement on a joystick and the roll of a vehicle.</field>
 <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a separate slider movement with maximum being 1000 and minimum being -1000 on a joystick and the thrust of a vehicle. Positive values are positive thrust, negative values are negative thrust.</field>
 <field type="int16_t" name="r" invalid="INT16_MAX">R-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a twisting of the joystick, with counter-clockwise being 1000 and clockwise being -1000, and the yaw of a vehicle.</field>
 <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' 0-15 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
 <extensions/>
 <field type="uint16_t" name="buttons2">A bitfield corresponding to the joystick buttons' 16-31 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 16.</field>
 <field type="uint8_t" name="enabled_extensions">Set bits to 1 to indicate which of the following extension fields contain valid data: 
  bit 0: pitch, bit 1: roll.</field>
 <field type="int16_t" name="pitch" invalid="0">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Set to zero if invalid.</field>
 <field type="int16_t" name="roll" invalid="INT16_MAX">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom.  Set to zero if invalid.</field>
</message>
```
I guess there is no way for a GCS to know if the other end supports this - so if it sends additional button presses and axes they will just get ignored.

hamishwillee: @auturgy P.S. I kind of liked the instance approach more. I'm not so concerned about MAVLink 1 (should we be?). However I would be concerned if the extended message got sent to a flight stack that had not been updated - so it was getting information for the extra buttons and DOF but thought it was getting the original info

hamishwillee:
I've put in request for confirmation of interim proposal in https://github.com/mavlink/mavlink/pull/1674#discussion_r695322935

What's the best way to continue the longer term discussion?

auturgy:
I’d suggest parking the longer term discussion in an issue for now.
Also, I think we need to consider whether the extensions should be roll and pitch, or x and y. The extant message conflates these: but for multicopters and fixed wing they’re essentially the same. So currently almost all vehicles using the existing message are XYZR == RPTY / AETR. Which was the intent of the message. It would follow that for consistent behaviour across vehicle types the extensions should really be X / Y lateral commands, not rotational. Repurposing 2 of the existing DoF’s when adding the extra 2 isn’t ideal. I’m guessing this is how Sub has implemented though, so open to discussion.

message_definitions/v1.0/common.xml
Williangalvani: @hamishwillee that looks correct.

I have two concerns about going with the instances approach. First one is that we would need to check per-vehicle if it supports instances, otherwise it will just interpret contradicting data (ignoring instance field and taking the values anyway). My second concern is how that would be handled at the GCS side. Should we have a per-vehicle list of controllable axis and a way to map these instances+fields into these axis? While it is more flexible, it feels a lot more cumbersome to implement and maintain.

hamishwillee:
@julianoes @auturgy Any counter-arguments? Would be good to take a proposal for general confirmation to the dev call tomorrow.

auturgy:
I'd prefer consistency across vehicles, but accept that it's a long standing issue - so may cause more confusion than it solves if we fix it now. No blocker from me.

message_definitions/v1.0/common.xml
```diff
@@ -4877,13 +4877,18 @@
...
+      <field type="uint16_t" name="buttons">A bitfield corresponding to the joystick buttons' 0-15 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 1.</field>
+      <extensions/>
+      <field type="uint16_t" name="buttons2">A bitfield corresponding to the joystick buttons' 16-31 current state, 1 for pressed, 0 for released. The lowest bit corresponds to Button 16.</field>
+      <field type="uint8_t" name="enabled_extensions">Set bits to 1 to indicate which of the following extension fields contain valid data: bit 0: pitch, bit 1: roll.</field>
+      <field type="int16_t" name="s" invalid="INT16_MAX">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Set to zero if invalid.</field>
+      <field type="int16_t" name="t" invalid="INT16_MAX">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom.  Set to zero if invalid.</field>
```
julianoes, line 4891: "Set to zero if invalid" doesn't match the `invalid="INT16_MAX"` field. I assume invalid should really be zero, so:

```
```suggestion
      <field type="int16_t" name="s" invalid="0">pitch-only-axis, normalized to the range [-1000,1000]. Generally corresponds to pitch on vehicles with additional degrees of freedom. Set to zero if invalid.</field>
      <field type="int16_t" name="t" invalid="0">roll-only-axis, normalized to the range [-1000,1000].  Generally corresponds to roll on vehicles with additional degrees of freedom.  Set to zero if invalid.</field>
```
```

julianoes (commented):
Other than my comment this looks ok.

auturgy:
We probably need to clarify that 0 is invalid iff the validity bit is also 0. If the validity bit is 1, 0 is a valid command value.

Williangalvani:
I'm not sure of the working and of what to use as an invalid value.
as 0 is in the valid range and we have a flag to tell if this is valid or not, do we even need the "invalid" field?

hamishwillee:
@Williangalvani @auturgy @julianoes I have removed the `invalid` flag as it is not useful in this context. Instead the description notes that the field must be set to zero if invalid. 
If there are any further changes to this, let me know. Otherwise will discuss this version in devcall this evening.

auturgy:
Merged.  Please provide documentation to reflect the update

hamishwillee:
> Please provide documentation to reflect the update

@Williangalvani obviously this is needed. If you need help, then happy to discuss. 

I was imaging we would add a page covering manual control protocol in https://mavlink.io/en/services/ - explaining that it is used to map joysticks. 

Not that we might well adjust that in future. It was pointed out in dev call that if you know about microservices, then probably you will look for it in https://mavlink.io/en/services/ . However if you don't know about it then you will probably be looking for something like "Joystick" or whatever, and not know where to start. This is also true for things like Camera, Gimbal and other payload integration. 

Upshot, for now docs can live in the services section, but I hope to make them more discoverable.

PR #1677, opened 2021-08-19
all.xml: add matrixpilot

peterbarker:
ArduPilot can remove SENSOR_OFFSETS

I initially expect this PR to fail because `SENSOR_OFFSETS` has not been removed from `ardupilotmega.xml`

Please do not merge this PR until I've cleared killing these messages with the ArduPilot project.


peterbarker:
OK, non-trivial as ArduPilot definitely needs `MEMINFO` and I've no idea if matrixpilot needs `FLEXIFUNCTION_BUFFER_FUNCTION`

PR #1679, opened 2021-08-22
Adding some missing MAV_TYPEs

olliw42:
In the heartbeat/connection protocol (https://mavlink.io/en/services/heartbeat.html) it is very nicely described that the type of the component should be identified through MAV_TYPE and not through its component id.

A necessary implication is that any type present in the MAV_COMP_ID list must also be present in the MAV_TYPE enum (i.e., that for each default MAV_COMP_ID there must be a suitable type available in the MAV_TYPE enu), but not vice versa.

For some types reflected in MAV_COMP_ID there is however no matching MAV_TYPE. This PR attempts to resolve this. (I actually wonder how it could happen that e.g. GPS isn't there).

For some MAV_COMP_ID it appears pretty clear what the MAV_TYPE should be, but for some it is actually not so clear to me. Hence this PR may not be the final answer and may have to be followed up by one adding what still turns out to be missing.

These MAV_COMP_IDs do not appear to have an obvious match in the MAV_TYPE enum:

MAV_COMP_ID_LOG
MAV_COMP_ID_OSD
MAV_COMP_ID_PERIPHERAL
MAV_COMP_ID_PATHPLANNER
MAV_COMP_ID_OBSTACLE_AVOIDANCE
MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY
MAV_COMP_ID_PAIRING_MANAGER
MAV_COMP_ID_IMU
MAV_COMP_ID_GPS
MAV_COMP_ID_UDP_BRIDGE
MAV_COMP_ID_UART_BRIDGE
MAV_COMP_ID_TUNNEL_NODE

Let's go through them.

For these MAV_COMP_IDs it seems to be obvious what their match should be, and these I have added in this PR

MAV_COMP_ID_LOG
MAV_COMP_ID_OSD
MAV_COMP_ID_IMU
MAV_COMP_ID_GPS

For these MAV_COMP_IDs I suspect that this would be the appropriate mapping, but I don't know, so here your input/checking is required:

MAV_COMP_ID_PATHPLANNER -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_OBSTACLE_AVOIDANCE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_UDP_BRIDGE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_UART_BRIDGE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_TUNNEL_NODE -> MAV_TYPE_GCS

For these MAV_COM_IDs I simply have no idea that their actual use is and what their types could be, you need to fill in your knowledge:

MAV_COMP_ID_PERIPHERAL -> ???
MAV_COMP_ID_PAIRING_MANAGER -> ???

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -182,6 +182,18 @@
       <entry value="37" name="MAV_TYPE_PARACHUTE">
         <description>Parachute</description>
       </entry>
+      <entry value="38" name="MAV_TYPE_LOG">
+        <description>Log</description>
+      </entry>
+      <entry value="39" name="MAV_TYPE_OSD">
+        <description>OSD</description>
+      </entry>
+      <entry value="40" name="MAV_TYPE_IMU">
+        <description>IMU</description>
+      </entry>
+      <entry value="41" name="MAV_TYPE_GPS">
+        <description>GPS</description>
+      </entry>
     </enum>
     <enum name="MAV_MODE_FLAG" bitmask="true">
       <description>These flags encode the MAV mode.</description>


olliw42:
this seems to be still valid to me :)

hamishwillee:
Thanks @olliw42 . It definitely deserves proper consideration. 

@julianoes if we don't have types to match components then we perpetuate using component IDs as type identifiers - because it is impossible/a hassle for people to use them as we recommend.

I'm pretty clear on the fact that you have a standalone mavlink component it needs a mav_type that matches what it does. So this PR adds MAV_TYPE_LOG, MAV_TYPE_OSD, MAV_TYPE_IMU, MAV_TYPE_GPS - I'm happy for those to exist, except possibly the IMU case, if it can't ever be a stand-alone thingy.

hamishwillee:
I am not still not clear on the rules for services that are inherently _part_ of an autopilot or onboard controller, and hence not at all sure what the "theory" is for this mapping:
MAV_COMP_ID_PATHPLANNER -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_OBSTACLE_AVOIDANCE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_UDP_BRIDGE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_UART_BRIDGE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_TUNNEL_NODE -> MAV_TYPE_GCS

Assume that an autopilot has 3 cameras attached (not standalone). In an ideal world the autopilot would emit a separate heartbeat for each. A GCS would get three heartbeats with the camera type and hence know the component IDs for the cameras, and that they exist. All very consistent, albeit with a bit more congestion. If you want to send a message to capture an image you send it to the autopilot system with the component ID of your desired target.

But autopilots don't do this. My understanding is that autopilots just emit the type of the vehicle and the component ID for the autopilot in their HEARTBEAT. A GCS therefore can't determine what standalone cameras are connected. To trigger a camera it would have to be hardcoded to understand that the vehicle had multiple cameras and it would have to use the specific component ids allocated for camera numbers to send a command. 

That sits badly with me - we're saying both that component IDs are just for addressing, and that they must also infer type, if you're sending them to an autopilot. Further, it implies that if we aren't going to use heartbeats we always need to create a separate discovery and identity mechanism.

The same kind of discussion could also be applied to the MAV_COMP_ID_PATHPLANNER etc above.

What am I missing?

olliw42:
> So this PR adds MAV_TYPE_LOG, MAV_TYPE_OSD, MAV_TYPE_IMU, MAV_TYPE_GPS - I'm happy for those to exist, except possibly the IMU case, if it can't ever be a stand-alone thingy.

Ardupilot supports external imus. (not saying a mavlink imu ever will exist, but it could)

olliw42:
> Assume that an autopilot has 3 cameras attached (not standalone)

IMHO only two situations exist then: 
(a) the autopilot mimics the camera protocol on behalf of each the cameras and would send heartbeats - as you said, but then it also would **have to** handle the messages targeted at the camera id's it is managing., i.e., do the full camera protocol for each camera.
(b) the autopilot would not. In this case there is **NO** established/defined mavlink procedure !!! 

Therefore the concern you raise does not exist from a mavlink perspective. The fact that folks use mavlink messages in all sorts of ways (which they do) does **NOT** mean that what all they do is within or even defines the mavlink concept.

Ergo, it only can be the 1st bullet. This architecture had been defined kind of in day 1 of mavlink.

Note that the same "problem" also exists for **ANY** other type, e.g., you could make the same line of words for gimbals, servos, escs, osds, gpses, ...

There is just no concept in mavlink for that. One could change mavlink, and close that seeming gap, but, hey, what is easier, to make an autopilot doing the existing messages or to define a whole bunch of new messages and make the autopilot handle these?

> which seems ... a technical challenge

not really. That's exactly what STorM32 does. It provides you both a gimbal and a camera and does that fully along mavlink, i.e. sends out two heartbeats, handles both ids appropriately, and everything else. To mavlink this one piece of hardware looks like two, mavlink couldn't distinguish it from two separate.

It sure may take some work for existing autopilots to be able to mimic more than one component, but that's just because of laziness, and not a real technical challenge. The question is if this is sufficient reason to giver up the existing mavlink concept. IMHO it is perfectly fine. In fact, if autopilots would gain that ability, other issues like the identification of "external components", which you try to "solve" by all sorts of meta data, would just evaporate away.

The only disadvantage of the identification via heartbeat is that it is send regularly and thus adds to the permanent traffic if there are really many components, as compared to a request&response approach. The advantage however is that it is much much easier to implement, as compared to a request&response approach (doing all these requests and waiting for the response and then the erroring based on it can be a bit of a pain)

I think mavlink should just take its concept seriously ;)

> I am ... not at all sure what the "theory" is for this mapping:

I'm neither sure about that mapping. However, neither of what I just said in the above is relevant here (IMHO, i.e., for said arguments).

To me the problem is rather that it is not clear if these beasts do or can or potentially can exist as standalone mavlink components. That is, is it warranted to introduce a MAV_TYPE for each of them or can one be more economically and use existing ones

hamishwillee:
@olliw42 Yes mavlink would all be much better if autopilots implemented support for connected components as though they were normal mavlink components. I doubt that will happen.

I do agree that this PR makes sense in an of itself. Waiting on @julianoes to confirm (away this week). I propose we ignore the problem of the other component ID mappings unless Julian has any ideas.

julianoes:
@hamishwillee you mention an autopilot which has three camera connected to it. I agree with the issue you are raising but I don't think cameras connected to an autopilot are a good example because it's not something that I have seen in the wild.

However, what does exist are companion computers which "contain" all sorts of mavlink "components", e.g. multiple cameras, gimbals, obstacle avoidance, and a path planner. For this case, I would argue that the companion computer should send out multiple heartbeats. Yes that's overhead but as companion computers would mostly be connected via an IP link (rather than low bandwidth serial), I would be ok with that.

And components should only answer to commands targeted to all, or only their component ID, so that would have to be a component id that they advertise using the heartbeat.

Now for this open component IDs/MAV_TYPEs:

MAV_COMP_ID_PATHPLANNER -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_OBSTACLE_AVOIDANCE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_VISUAL_INERTIAL_ODOMETRY -> MAV_TYPE_ONBOARD_CONTROLLER

The component IDs are maybe a bit too granular. I would probably argue that they should have just been MAV_COMP_ID_COMPANION_COMPUTER instead. Of course, we can't change that now, so the mapping @olliw42 suggests is ok, in my opinion.

MAV_COMP_ID_UDP_BRIDGE -> MAV_TYPE_ONBOARD_CONTROLLER
MAV_COMP_ID_UART_BRIDGE -> MAV_TYPE_ONBOARD_CONTROLLER

UDP and UART bridges are basically implemented by mavlink-router, I'd say (which also can do logging but let's ignore that for now), so I would use something like MAV_TYPE_BRIDGE, or even MAV_TYPE_UDP_BRIDGE, etc. I would not use MAV_TYPE_ONBOARD_CONTROLLER because these bridges can be on the vehicle or on the ground, so they are not tied to "onboard".

MAV_COMP_ID_TUNNEL_NODE -> MAV_TYPE_GCS

And in the same way the TUNNEL_NODE is not tied to the GCS, I think. Here my question would be if the TUNNEL_NODE actually sends messages and needs to be advertised in a heartbeat, or if it could actually be transparent, and doesn't require a component ID and MAV_TYPE at all.

olliw42:
> MAV_TYPE_BRIDGE

good idea

> TUNNEL_NODE ...  Here my question would be if the TUNNEL_NODE actually sends messages and needs to be advertised in a heartbeat, or if it could actually be transparent, and doesn't require a component ID and MAV_TYPE at all.

it does
fundamentally, the tunnel message is exchanged between two things which necessarily are mavlink components and hence each need their ids and type. It is correct however that it would not necessarily need a special tunnel node id, but it is helpful.

Maybe I describe (one of) my use cases of tunnel, as it points to a rather general problem with "GCS"es: 

I use it to communicate between a STorM32 gimbal and it's STorM32 specific native GUI, through any available mavlink port. That's super practical since one doesn't need to access the gimbal itself, but can do the config through any port, e.g. the autopilot usb or telemetry (I had done a video some time ago which I posted somewhere here to show to @hamishwillee its practical use). 
So, the GUI needs some system and component ids, and of course it would be possible to choose any which is not yet used. That's however the point: If a user would first had to set these values before starting using the GUI it would seriously diminish the practical value of this (I'm a fan of plug&play for such things). 
So the next would be to use ids which hopefully will never be used by any other. This works, but is not nice. And this was the reason I once suggested a generic id for this other end of the tunnel message, to reduce chances of conflict.
It also could do some auto detection - and the GUI does do quite some to figure out what's going on - but this also can't work always since eventually someone needs to make a decision and start sending.
The GUI could be seen as a GCS however, and that's why I use the GCS type for it. Functionally, it would be better to have a specific STorM32-gui type, but that's not a concept either as it implies a flurry of types.
That's the background of the above assignment MAV_COMP_ID_TUNNEL_NODE -> MAV_TYPE_GCS

The underlying issue is IMHO broader since it generally occurs when multiple GCSes can be in the game. In my case it are 4: MissionPlanner, QGC, OpenTx transmitter running MAVLink for OpenTx firmware, and said STorM32 GUI. One can hard-wire the ids and hope for the best. One can make them user-configurable, but that's super annoying. Note that for other components as those on a vehicle the issue doesn't exist so much, since usually these are not "hot-swapable".

What one really would like to have is as sort of dynamic id allocation, or a DNS server for ids, or such.

if it should happen that MAV_COMP_ID_TUNNEL_NODE is the only id which causes troubles, i.e, where one can't find a suitable type, I would not strongly object to removing it ... I would simply return to what I did before to just pick a value ... not nice conceptually, but the overall fabric might be more important

hamishwillee (approved):
I can see you're continuing the discussion. I'm going to approve and merge this set because we agree on them. 

You guys can continue the discussion on the companion case separately -might be good to open a new issue?

PR #1687, opened 2021-08-27
Mission item grouping commands and messages

MatejFranceskin:
This PR adds mission items grouping commands. Groups are defined by START/STOP items and 64bit unsigned integer ID. A group can be nested inside another group. When control reaches MAV_CMD_GROUP_START or MAV_CMD_GROUP_END it emits GROUP_START message with group ID and current timestamp.

Task requirements and what kind of problem are we trying to solve:
We want to be able to associate data acquired by the vehicle (f.e. photos) with groups of mission items.
For example: user wants to inspect 2 bridges and creates a flight plan with several way points in proximity of each bridge.
With the proposed solution we would group mission items in 2 groups. For each photo we will add current group id and group timestamp to image metadata. This metadata will be later used to f.e. filter out only photos taken at "bridge 2".

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1610,6 +1610,14 @@
         <param index="1" label="First Item" minValue="0" increment="1">first_item: the first mission item to run</param>
         <param index="2" label="Last Item" minValue="0" increment="1">last_item:  the last mission item to run (after this item is run, the mission ends)</param>
       </entry>
+      <entry value="301" name="MAV_CMD_GROUP_START" hasLocation="false" isDestination="false">
+        <description>Define beginning of a group of mission items. When control reaches this command a GROUP_START message is emitted. Groups can be nested.</description>
+        <param index="1" label="Group ID" minValue="0" increment="1">Unique group id</param>
+      </entry>
+      <entry value="302" name="MAV_CMD_GROUP_END" hasLocation="false" isDestination="false">
+        <description>Define ending of a group of mission items. When control reaches this command a GROUP_END message is emitted. Groups can be nested.</description>
+        <param index="1" label="Group ID" minValue="0" increment="1">Unique group id</param>
+      </entry>
       <entry value="400" name="MAV_CMD_COMPONENT_ARM_DISARM" hasLocation="false" isDestination="false">
         <description>Arms / Disarms a component</description>
         <param index="1" label="Arm" minValue="0" maxValue="1" increment="1">0: disarm, 1: arm</param>
@@ -4690,6 +4698,20 @@
       <field type="float" name="p2x" units="m">x position 2 / Latitude 2</field>
       <field type="float" name="p2y" units="m">y position 2 / Longitude 2</field>
       <field type="float" name="p2z" units="m">z position 2 / Altitude 2</field>
+    </message>
+    <message id="56" name="GROUP_START">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Emitted when control reaches MAV_CMD_GROUP_START with unique group id and time stamp of the event.</description>
+      <field type="uint32_t" name="group_id">Unique group id.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+    </message>
+    <message id="57" name="GROUP_END">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Emitted when control reaches MAV_CMD_GROUP_END with unique group id and time stamp of the event.</description>
+      <field type="uint32_t" name="group_id">Unique group id.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
     </message>
     <message id="61" name="ATTITUDE_QUATERNION_COV">
       <description>The attitude in the aeronautical frame (right-handed, Z-down, X-front, Y-right), expressed as quaternion. Quaternion order is w, x, y, z and a zero rotation would be expressed as (1 0 0 0).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1610,14 +1610,6 @@
         <param index="1" label="First Item" minValue="0" increment="1">first_item: the first mission item to run</param>
         <param index="2" label="Last Item" minValue="0" increment="1">last_item:  the last mission item to run (after this item is run, the mission ends)</param>
       </entry>
-      <entry value="301" name="MAV_CMD_GROUP_START" hasLocation="false" isDestination="false">
-        <description>Define start of a group of mission items. When control reaches this command a GROUP_START message is emitted. The end of a group is marked using MAV_CMD_GROUP_END with the same group id. Groups can be nested or overlap.</description>
-        <param index="1" label="Group ID" minValue="0" increment="1">Mission-unique group id</param>
-      </entry>
-      <entry value="302" name="MAV_CMD_GROUP_END" hasLocation="false" isDestination="false">
-        <description>Define end of a group of mission items. When control reaches this command a GROUP_END message is emitted. The start of the group is marked is marked using MAV_CMD_GROUP_START with the same group id. Groups can be nested or overlap.</description>
-        <param index="1" label="Group ID" minValue="0" increment="1">Mission-unique group id</param>
-      </entry>
       <entry value="400" name="MAV_CMD_COMPONENT_ARM_DISARM" hasLocation="false" isDestination="false">
         <description>Arms / Disarms a component</description>
         <param index="1" label="Arm" minValue="0" maxValue="1" increment="1">0: disarm, 1: arm</param>
@@ -6663,20 +6655,6 @@
       <field type="uint16_t" name="sequence">Sequence number.</field>
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
-    </message>
-    <message id="414" name="GROUP_START">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Emitted during mission execution when control reaches MAV_CMD_GROUP_START.</description>
-      <field type="uint32_t" name="group_id">Mission-unique group id (from MAV_CMD_GROUP_START).</field>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
-    </message>
-    <message id="415" name="GROUP_END">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Emitted during mission execution when control reaches MAV_CMD_GROUP_END.</description>
-      <field type="uint32_t" name="group_id">Mission-unique group id (from MAV_CMD_GROUP_END).</field>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
     </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1611,12 +1611,12 @@
         <param index="2" label="Last Item" minValue="0" increment="1">last_item:  the last mission item to run (after this item is run, the mission ends)</param>
       </entry>
       <entry value="301" name="MAV_CMD_GROUP_START" hasLocation="false" isDestination="false">
-        <description>Define beginning of a group of mission items. When control reaches this command a GROUP_START message is emitted. Groups can be nested.</description>
-        <param index="1" label="Group ID" minValue="0" increment="1">Unique group id</param>
+        <description>Define start of a group of mission items. When control reaches this command a GROUP_START message is emitted. The end of a group is marked using MAV_CMD_GROUP_END with the same group id. Groups can be nested or overlap.</description>
+        <param index="1" label="Group ID" minValue="0" increment="1">Mission-unique group id</param>
       </entry>
       <entry value="302" name="MAV_CMD_GROUP_END" hasLocation="false" isDestination="false">
-        <description>Define ending of a group of mission items. When control reaches this command a GROUP_END message is emitted. Groups can be nested.</description>
-        <param index="1" label="Group ID" minValue="0" increment="1">Unique group id</param>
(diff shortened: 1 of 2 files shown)


hamishwillee:
Is this ready for review?

message_definitions/v1.0/common.xml
```diff
@@ -4794,6 +4802,20 @@
       <field type="float" name="p2y" units="m">y position 2 / Longitude 2</field>
       <field type="float" name="p2z" units="m">z position 2 / Altitude 2</field>
     </message>
+    <message id="56" name="GROUP_START">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Emited when control reaches MAV_CMD_GROUP_START with unique group id and time stamp of the event.</description>
```
julianoes, line 4808:
```suggestion
      <description>Emitted when control reaches MAV_CMD_GROUP_START with unique group id and time stamp of the event.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -4794,6 +4802,20 @@
...
+      <field type="uint32_t" name="group_id">Unique group id.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+    </message>
+    <message id="57" name="GROUP_END">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Emited when control reaches MAV_CMD_GROUP_END with unique group id and time stamp of the event.</description>
```
julianoes, line 4815:
```suggestion
      <description>Emitted when control reaches MAV_CMD_GROUP_END with unique group id and time stamp of the event.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -4794,6 +4802,20 @@
       <field type="float" name="p2y" units="m">y position 2 / Longitude 2</field>
       <field type="float" name="p2z" units="m">z position 2 / Altitude 2</field>
     </message>
+    <message id="56" name="GROUP_START">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Emited when control reaches MAV_CMD_GROUP_START with unique group id and time stamp of the event.</description>
+      <field type="uint32_t" name="group_id">Unique group id.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
```
julianoes, line 4810: Do we need microsecond resolution for this? Or would millisecond be enough?

julianoes (commented):
I think that's overall ok. The one sub-optimal thing is that we set the ID as a float param, so we don't have the full 32bit unsigned range available.

message_definitions/v1.0/common.xml
MatejFranceskin: @julianoes We don't need microsecond resolution. I used it because it is used everywhere as Unix epoch time, while uint32_t millisecond timestamp usually means time since boot.
Should I keep 64bit time, but change the resolution. Like milliseconds from 1.1.1970?

julianoes: Oh, of course! Leave it as is then.

LorenzMeier:
@MatejFranceskin CI tests are failing.

message_definitions/v1.0/common.xml
```diff
@@ -4794,6 +4802,20 @@
       <field type="float" name="p2y" units="m">y position 2 / Longitude 2</field>
       <field type="float" name="p2z" units="m">z position 2 / Altitude 2</field>
     </message>
+    <message id="56" name="GROUP_START">
```
hamishwillee, line 4805: Does this message _really_ need to use one of the very few low id values?

hamishwillee: Ditto for the other one?

hamishwillee:
What if messages are lost
1. Is the intent that the start/end will be logged flight-stack side so that they can't be missed
2. Is the intent that the metadata is applied to images on the flight-stack side?
3. You say groups can be nested, but can they overlap? I.e. can I define "group to bridge 1" and "group to bridge1 then bridge"?
4. Assuming overlapping is not allowed assume the case where a GCS gets group1_start, group2_start, group_1_end. That indicates a dropped group 2_end. Is there anything can be done about that/what assumptions are made?

RomanBapst:
@hamishwillee @MatejFranceskin I'm not seeing any field specifying the actual group name, thus my conclusion is that there is going to be metadata stored somewhere which maps the ID to a name.
If the above assumption is correct, does it then really make sense to introduce this grouping command, why is it not enough to just read the current sequence number (MISSION_CURRENT) and then just read all required info from the matadata file (e.g. json)?

julianoes:
> What if messages are lost

They should not get lost as part of a mission as the autopilot will just go through them all one by one and doesn't/shouldn't lose anything on the way.

If used as commands, they would individually get ack'd but I don't understand why you would use them as commands.

> 2. Is the intent that the metadata is applied to images on the flight-stack side?

That's a good question. Yes that's possible, or if required items can be specced that contain metadata, although putting strings into floats would not be nice.

> 3. You say groups can be nested, but can they overlap? I.e. can I define "group to bridge 1" and "group to bridge1 then bridge"?

I would say, they can nest like a tree but it needs to be hierarchical:
Group1-start, Group2-start, Group2-end, Group1-end -> OK
Group1-start, Group2-start, Group1-end, Group2-end -> NOT OK

> 4. That indicates a dropped group 2_end

No drops in my opinion.

message_definitions/v1.0/common.xml
MatejFranceskin: @hamishwillee No, it can be any value. What should I change it to?

MatejFranceskin:
> 
> 
> @hamishwillee @MatejFranceskin I'm not seeing any field specifying the actual group name, thus my conclusion is that there is going to be metadata stored somewhere which maps the ID to a name.
> If the above assumption is correct, does it then really make sense to introduce this grouping command, why is it not enough to just read the current sequence number (MISSION_CURRENT) and then just read all required info from the matadata file (e.g. json)?

This is a good remark. The idea was that group id is a key for retrieving string information about the groups and timestamp would be used to distinguish between multiple executions of the same mission.

message_definitions/v1.0/common.xml
hamishwillee: The ids you have picked seem good :-)

hamishwillee:
> What if messages are lost

@julianoes 

Commands are never missed but the message emitted in response travel over a radio link, so they most definitely can be lost :-)

Hence my second question. If the logging of the group information is done on the PX4-side then the fact that a message can be lost is irrelevant. However if the assumption is that the GCS is going to process the images based on a timestamp in the **received** message (which seems to be the intent) then the loss of a message is very relevant.

Which leads to my third/fourth question, if there is a loss is there anything we can assume? From your response you're saying that groups must be nested, so we can at least assume that a missing end tag will be bounded by the following nest group closure. 

> The idea was that group id is a key for retrieving string information about the groups and timestamp would be used to distinguish between multiple executions of the same mission.

@MatejFranceskin Makes sense. That would be something for "the docs".

> Development.xml

Has someone mentioned the fact that things need to be prototyped before they can be accepted into common.xml - usually this should be in development.xml.

I personally don't mind how this is managed, but we can't merge into common.xml until after there is a prototype and broad agreement from at least one other flight stack to support this. At that point the wip tags would all be removed too. Thoughts?

MatejFranceskin:
@hamishwillee I have a simple working prototype with changes in PX4, QGC & vehicle camera manager.

hamishwillee:
@auturgy This is probably coming close to the point where you will want to have a look. It's a way to add more detailed tracking of what images should be grouped in a survey than just timestamping. Essentially a mission command that wraps around images captured in a particular survey, each which results in an emitted message. 

It has a prototype in PX4. I'm hoping it can go into common.xml.

auturgy:
This is forcing a specific workflow methodology into Mavlink.
That might be valid, might not, but definitely needs a more robust and detailed argument than that provided

hamishwillee:
@auturgy What you mean by "This is forcing a specific workflow methodology into Mavlink"? Because I don't see that at all, so I don't know what extra information you might require from @MatejFranceskin that would be more "robust".

All this does is allow a mission to be tagged with metadata that a GCS can interpret if it so wishes to timestamp and associate particular parts of a mission with artifacts of the mission. It forces no particular workflow, and as with other commands, a GCS or Autopilot is not required to implement the command.

The benefit to me is that this is a mavlink-compatible way to allow someone running better identify a subset of a survey. That use case comes up again and again, primarily from people stitching images, and this is the first time I've seen an approach that will work and which won't have a reliance on the camera doing anything (a good thing). What I'd like to see is more information about the prototype - logs, screenshots etc.

Perhaps a chat offline if you have time to help me understand the concern? Otherwise I'll add it to the devcall agenda.

MatejFranceskin:
@julianoes
"I would say, they can nest like a tree but it needs to be hierarchical:
Group1-start, Group2-start, Group2-end, Group1-end -> OK
Group1-start, Group2-start, Group1-end, Group2-end -> NOT OK"

This is how it was meant to be used, but do we have to limit it like this in mavlink standard? Isn't it enough that we just define behaviour of how flight controller executes these commands when it is their turn in the flight plan. If someone would want to use overlapping groups instead of just nested groups then why not.

hamishwillee:
> If someone would want to use overlapping groups instead of just nested groups then why not.

I have no problem with that. The point is to remove as much potential ambiguity as possible - i.e. explicitly state that groups can nest or overlap - the only requirement is that groups start with a start marker and end with an end marker.

Further, the implication is that you can't make any assumptions if you don't get an end marker other than it was omitted. So either we'd 
- say something in our docs to the effect that "start and end messages may not be received; a GCS would need to use an on-vehicle log of messages to ensure that this information is available
- Require flight stack to store the last message and resend on request (with old timestamp etc).

hamishwillee:
@MatejFranceskin 

I've done a subedit:
- cross reference commands for start and end (mention them in each others descriptions
- Note that the messages are emitted when executing missions - if you were just reading the messages it isn't obvious (though it should be)
- Note that nesting and overlap are allowed
- Note that the id is mission-unique, not globally unique. Note that in the message it comes from the command.

Thanks for attending the dev call. The summarise the discussion, we're all on board with this. However we need an implementation before we can merge. That means this should really be pushed into development.xml (without the WIP tagging). Then tested. When there is a validation and you're sure it is good we can move back into common.xml in a separate PR.

Hope that matches your understanding.

hamishwillee:
@MatejFranceskin 
1. I fixed a trivial typo
2. Why such a huge ID? It says "mission unique", are you now assuming some kind of global hash? If not, then I can assure you that yo udo not need a 64 bit number :-)
3. My assumption is that even though we're using these for missions, there is nothing in theory to stop us using the same feature outside for some kinds of grouping in the log. So name is good. If we want to restrict this to missions forever we'd name `MISSION_GROUP_START` etc.
4. It might be good to update common.xml with a comment at the places where the ids are inserted. For example see https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L1760
    This is not a "requirement" but it makes it less likely someone else will try use the same ids while you're working on this.

If you're otherwise good with this, let me know when you're ready to merge.

MatejFranceskin:
I changed the definitions so that we're able to have 64bit group ids. We cannot store more then 24bit in the floats, so I decided to use 4 parameters and limit them to 16bits.

Created a draft PR in PX4 repo that uses these new messages.

hamishwillee:
> I changed the definitions so that we're able to have 64bit group ids. We cannot store more then 24bit in the floats, so I decided to use 4 parameters and limit them to 16bits.

@MatejFranceskin Sorry, I don't think that answers the question "why do you need 64 bit group ids"? For your described use case I was envisaging 5 or 6 groups per mission, which you could do with a u_int8. Let's say a u_int16 to allow more than 256 - if you need 65K of these you're probably doing something wrong.

MatejFranceskin:
Group IDs will be randomly generated and should be unique among multiple ground control stations.
This is why the team working on ground station group / image metadata database wants them to be 64bit.

MatejFranceskin:
> > I changed the definitions so that we're able to have 64bit group ids. We cannot store more then 24bit in the floats, so I decided to use 4 parameters and limit them to 16bits.
> 
> @MatejFranceskin Sorry, I don't think that answers the question "why do you need 64 bit group ids"? For your described use case I was envisaging 5 or 6 groups per mission, which you could do with a u_int8. Let's say a u_int16 to allow more than 256 - if you need 65K of these you're probably doing something wrong.

So it looks like we can live with one parameter that gives us 24bit resolution of group IDs. I will change back from 4 parameters to 1.

auturgy:
17 million group ID’s?
You still need to justify that. I get that it’s only an additional byte, but I’m really struggling to see how you could possibly consume 16 bits of unique identifier, as suggested by Hamish, in the context of this message.
Can you better align your implementation to your presented explanation?

hamishwillee:
Hi @MatejFranceskin,

> Group IDs will be randomly generated and should be unique among multiple ground control stations.

It absolutely makes sense to use a larger random number if the value needs to be unique among ground stations. But the use case you shared was that this would be closely tied to the GCS or other system using the data. Under that assumption the value could be a byte and uniqueness would really only be of interest to the current GCS.

So I think @auturgy is asking why it needs to be unique among ground stations. 

If this is a random number, we will need to change the text from "mission unique" - it will become a globally unique.

auturgy:
The use case/implementation needs to be explicitly described.
I see the general utility of groups, but as per my first comment, this seems to be designed for a very specific use case/workflow.
Based on the (obviously incomplete) information provided, it appears that in order to simplify some element of your specific case you are complicating the general case.

The group should not need to be globally unique. The receiver knows the SYSID of the sender, and the time stamp. For any operator with basic fleet management practices, the combination of the group id, SYSID and time stamp will be unique even with a small range of group ids, and provides for easier filtering/post processing: you’ll almost certainly need that additional metadata ass with the group id anyway.

message_definitions/v1.0/development.xml
```diff
@@ -155,6 +155,16 @@
         <description>SIM is available, but not usuable for connection</description>
       </entry>
     </enum>
+    <enum name="MAV_CMD">
+      <entry value="301" name="MAV_CMD_GROUP_START" hasLocation="false" isDestination="false">
+        <description>Define start of a group of mission items. When control reaches this command a GROUP_START message is emitted. The end of a group is marked using MAV_CMD_GROUP_END with the same group id. Groups can be nested or overlap.</description>
```
julianoes, line 160: I don't understand why that would be required. But if that's required, ok we can leave it in.

julianoes (approved):

message_definitions/v1.0/development.xml
LorenzMeier: Groups should NOT overlap because otherwise we end up with a non-tree structure, which would be hard to map to "normal" data structures. No need to get creative like that without a good reason / use case.

MatejFranceskin: Removed "overlap".

julianoes (approved):

hamishwillee:
@MatejFranceskin Looks excellent. I am going to merge.
- I added a note "Group ids are expected, but not required, to iterate sequentially." (which is something that Lorenz stated)
- I added some layout changes. Ignore those - there is no "policy" on layout other than it can't break the parser. But I find it much easier to read in PRs and changes if sentences are on different lines.

The only thing that caught my eye is this "CRC32 checksum of current plan for `MAV_MISSION_TYPE_ALL`". I assume that for the purposes of this "All" is better than just for the mission - ie that fences and rally points have to be part of it?

hamishwillee:
@RomanBapst What would it mean if you put this around a fence? 

In a mission the command is executed when the mission reaches the correct point of the sequence, emiting the corresponding group message. But a fence definition isn't executed sequentially, so I don't understand how you would want this to behave.

If you just wanted to bookend a part of a fence, then I'd want to know "for what purpose". Note we have got an approach (unimplemented) to allow per-fence actions to be defined.

PR #1688, opened 2021-08-28
MAV_LANDED_STATE: add state for preparing for take off

olliw42:
this PR adds a new, additional, but optional, state to MAV_LANDED_STATE, in order to indicate that the vehicle will soon take off.

ArduPilot for instance introduces (for copters) a 2 sec period after arming and before taking off (=motors engange and spool up). 

This period can be very useful for components, such as e.g. gimbals. A gimbal for instance could run its gyro bias calculations and other initialization or estimations up to this point, and then has enough time to safely apply them while the vehicle is still at rest, before the vehicle is actually taking off and thus not ensured to be at rest anymore. Since there is usually an significant longer time between power up and actual takeoff, this can yield much better results than the normal procedure of doing these estimations shortly after power up. For instance, the gimbal's IMUs might have better adjusted to the temperature, or "boat" modes which typically require longer estimation times become better feasible. 

For ArduPilot it in principle would be possible to infer this pre-takeoff period from the messages which yield the arming state and the messages which yield the landed state. However, these are emitted at relatively low frequency (1 Hz), and not synchronized to the arming nor taking-off, nor to each other, which means quite some ambiguity and cut-down of the safe time. Also, this is obviously very flight-stack specific.

Note that the landed_state is (can be) communicated to the gimbal by the AUTOPILOT_STATE_FOR_GIMBAL_DEVICE message, which is typically a high-frequency message, and adding the new state to the landed_state makes this info available to the gimbal in a very timely and synchronized fashion. Currently this is not so.

Also note that this state does not have to be supported by a vehicle, which should mitigate incompatibility concerns, i.e., vehicle codes do not have to be modified. Things would then just work as before.

One could consider adding as requirement for this state that a vehicle must set it only if the pre-takeoff period is guaranteed to be longer than e.g. 1 sec, or 2 sec, or whatever one finds reasonable to specify. This would have the advantage that a component could rely on a certain time span independent on implementation/flight-stack. But I would not make this a part of this PR, as it might need more discussion, so just wanted to mention it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3564,6 +3564,9 @@
       <entry value="4" name="MAV_LANDED_STATE_LANDING">
         <description>MAV currently landing</description>
       </entry>
+      <entry value="5" name="MAV_LANDED_STATE_PREPARING_FOR_TAKEOFF">
+        <description>MAV will soon take off</description>
+      </entry>
     </enum>
     <enum name="ADSB_ALTITUDE_TYPE">
       <description>Enumeration of the ADSB altimeter types</description>


hamishwillee:
we should be very clear on what we mean by `MAV_LANDED_STATE_PREPARING_FOR_TAKEOFF` and `MAV_LANDED_STATE_TAKEOFF`, and the differences/overlaps between them. Right now I "imagine" that `MAV_LANDED_STATE_TAKEOFF` means that the state machine has started the process of taking off, so it would include what you mean by `MAV_LANDED_STATE_PREPARING_FOR_TAKEOFF`. 

What implications would it have to things using the enum to effectively split the state into two?

olliw42:
that's at least not so for ardupilot (on copter)
MAV_LANDED_STATE_TAKEOFF means that the motors are spooling up
but there is a minimaly 2 sec period between when the copter has been command to take off 
I can't describe exactly now what "command to take off" means since this can occur in various ways, involving arm and takeoff triggers (commands/messages/sticks...)
so the "state machine" is more complicated, but ardupilot (on copter) changes to MAV_LANDED_STATE_TAKEOFF when it spools up the motors i.e. actually takes off

the problem is a bit language here and what "takeoff" means ... I mean the physical takeoff, there the motors start to spool up and the copter starts to stop being at rest ... I do not mean any internal states or whatever

I think how ardupilot is interpreting landed state is the most meaning full way

but I of course fully agree that it should be as clearly defined as possible, e.g., to avoid confusion with internal state machine's states

I guess I define by purpose, i.e., to give components & peripherals a moment of time to make themselves ready for takeoff (I believe this is why ardupilot has this 2 secs itself, to get the EKFs and what else ready)

auturgy:
We need to discuss this a little: it was agreed at the mavlink devcall that this is a good idea, but we need to achieve three things:
1.  Look at second order effects.  This is a core, stable message so the barrier to change is a little higher than perhaps other messages.
2. Implementation: @LorenzMeier will look into PX4 implementation.  If @olliw42 could look at ArduPilot implementation, it would be appreciated.  If that hits a hurdle, ping me
3. Landing state.  Should this be replecated during landing as well?  Our gut feel is "yes", but we need to have a proper think about it.  Your perspective would be appreciated.

Thanks,

James

olliw42:
many thx for looking at this, and so swiftly, much appreciated

I surely agree it needs careful discussion

I'd like to mention that this seems to be so also for the already EXISTING states. Hamishwillee considered the transition to MAV_LANDED_STATE_TAKEOFF to correspond to the start of a state machine, which may do whatever things (e.g. waiting), while ArduCopter/me regard this as the point in time there motors are spooled up (actually, it's a bit more complicated in ArduCopter since there are a number of timers here, but I don't know how to better formulate it). So, even the precise meaning of MAV_LANDED_STATE_TAKEOFF appears a bit ambigious.

In order to better describe what I have in mind (the wording in the dev call minutes I find somewhat confusing if not misleading), I'd like to say that the new state I have in mind is a **substate of MAV_LANDED_STATE_ON_GROUND** !!!

That is, a flight stack which does not want to support the new state would report MAV_LANDED_STATE_ON_GROUND. Like-wise, a recipient which does not understand or does not want to handle the new state would read it to mean MAV_LANDED_STATE_ON_GROUND.


@auturgy

I do not completely understand what I should do concerning point (2). Should I suggest a implementation of the new state, or you "just" have a review of the current implementation in mind? ???

Concerning (3), the landing state was not part of my original thinking, i.e., from my perspective I would not see changes here.


Lastly, I'd like to mention that there are of course alternative approaches to achieve what this new state aims at. I'd like to briefly mention two, to widen the perspective a bit. For instance, one could add another field/flag to AUTOPILOT_STATE_FOR_GIMBAL_DEVICE, which could report certain events, or one does not use MAV_LANDED_STATE but a gimbal specific MAV_GIMBAL_LANDED_STATE. The historical reason why MAV_LANDED_STATE was used was ergonomics, i.e., trying to reuse as much as possible. If it now turns out that this is a bit too constraining or too infectious to stable messages, it might be better/easier/cleaner to let lose of MAV_LANDED_STATE.

olliw42:
I'm not sure this is helpful or rather confusing, but nevertheless two pictures of two flights with the "original" ArduCopter behavior and "my" modified behavior. It may be confusing because for "my" case it does NOT do what it is supposed to do for landing, but that's just because I did it the simple way and the wrong state for landing doesn't matter to me !! So pl just look at around takeoff.

the pics shows the landed state (red), the rangefinder distance (green) and a motor output (blue)


* original ArduCopter:

the motors spool up and the copter gains height at (ca) the moment when the state transitions form STATE_ONGROUND to STATE_TAKEOFF




* "my" modified 

here the motors spool up and the copter gains height at (ca) the moment when the state transitions to STATE_TAKEOFF, but here the states transitions to STATE_PREPARING_FOR_TAKEOFF at minimum 2 seconds before the motors spool up and the copter gains height. The transition from STATE_ON_GROUND to STATE_PREPARING_FOR_TAKEOFF corresponds here in fact to when the copter is armed. Again, the behavior for landing is incorrect solely because of my lazy coding.

hamishwillee:
@olliw42 Is this something you still want, and think you can get into a flight stack? I'm doing cleanup - if there is no stakeholder who wants and needs this I'll close it.

hamishwillee:
OK. Closing. Feel free to reopen if we can get a meaningful discussion from someone who wants to implement.

PR #1695, opened 2021-09-02
common: add motor_info message

cmic0:
As per discussion in https://github.com/mavlink/mavlink/pull/1663, this PR introduces a new Mavlink message to handle Motor information in a different message than `ESC_INFO`.

@hamishwillee @auturgy

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -627,7 +627,7 @@
       <!-- This enum is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Flags to report ESC failures.</description>
       <entry value="0" name="ESC_FAILURE_NONE">
-        <description>No ESC failure.</description>
+        <description>No ESC failures.</description>
       </entry>
       <entry value="1" name="ESC_FAILURE_OVER_CURRENT">
         <description>Over current failure.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes @auturgy you know more about motors than me - would you like to add your comments first? I will look at this tomorrow.

auturgy:
I've made a couple of suggestions. Biggest comment is that this should go into development.xml until implemented.

message_definitions/v1.0/common.xml
```diff
@@ -6385,6 +6420,18 @@
       <field type="float[4]" name="voltage" units="V">Voltage measured from each ESC.</field>
       <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
+    <message id="292" name="MOTOR_INFO">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Motor information for lower rate streaming. Recommended streaming rate 1Hz.</description>
+      <field type="uint8_t" name="index" instance="true">Index of the first Motor in this message. minValue = 0, maxValue = 60, increment = 4.</field>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number.</field>
+      <field type="uint16_t" name="counter">Counter of data packets received.</field>
```
hamishwillee, line 6429: What is this for?

message_definitions/v1.0/common.xml
```diff
@@ -6385,6 +6420,18 @@
...
+      <field type="uint8_t" name="count">Total number of Motors in all messages of this type. Message fields with an index higher than this should be ignored because they contain invalid data.</field>
```
hamishwillee, line 6430: Would it be better to make this "Count of motors in this message"? The reason you have something like this is so that you can identify the end point of a partially filled array.

If this is count of all motors, then to work out whether values in these arrays are valid you need to compare `count` and `index` etc. Is there some other reason to keep max count?

hamishwillee:
@auturgy YOur suggestions aren't showning up. 

@cmic0 I assume you understand the workflow. We can't merge a PR into common.xml until there is working implementation. At that point the changes merge *without* wip tags (and you would also do the "removals" from the ESC fault enum, and remove its WIP tagging).

hamishwillee:
Closing. This went stale. Got revisited in https://github.com/mavlink/mavlink/pull/1895. TO progress need a new PR and stakeholder.

PR #1697, opened 2021-09-09
Move CELLULAR_CONFIG to development.xml

hamishwillee:
This moves WIP CELLULAR_CONFIG message and associated enums to development.xml

I **think* this was agreed in the dev call [20210901](https://github.com/mavlink/mavlink/wiki/20210901-Dev-Meeting) as there appears to be no public implementation. @julianoes @auturgy Was that what you remember?

FYI @DanielePettenuzzo


julianoes (commented):
Yes, that's what I remember.

hamishwillee:
Moving then as it what we both recall. Thanks Julian.

PR #1700, opened 2021-09-16
MAV_FRAME - better match descriptions to RFC

hamishwillee:
This attempts to better match MAV_FRAME descriptions to the conventions used in the [proposed RFC](https://github.com/mavlink/rfcs/pull/2), as discussed in #1685. This requires careful review to ensure I have not changed the meaning of any of the frames from their intent.

There are notes inline about some sections.

Anyone else we can/should get to review? The RFC authors? @Jaeyoung-Lim ? @TSC21 ?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -216,11 +216,27 @@
       </entry>
     </enum>
     <enum name="MAV_FRAME">
+      <description>Co-ordinate frames used by MAVLink. Not all frames are supported by all commands, messages, or vehicles.
+      
+      Global frames use the following naming conventions:
+      - `GLOBAL`: Global co-ordinate frame with WGS84 latitude/longitude and altitude positive over mean sea level (MSL) by default. 
+        The following modifiers may be used with `GLOBAL`:
+        - `RELATIVE_ALT`: Altitude is relative to the vehicle home position rather than MSL
+        - `TERRAIN_ALT`: Altitude is relative to ground level rather than MSL
+        - `INT`: Latitude/longitude (in degrees) are scaled by multiplying by 1E7  
+
+      Local frames use the following naming conventions:
+      - `LOCAL`: Origin of local frame is fixed relative to earth. Unless otherwise specified this origin is the origin of the vehicle position-estimator ("EKF").
+      - `BODY`: Origin of local frame travels with the vehicle.
+      - `OFFSET`: Deprecated synonym for `BODY` (origin travels with the vehicle). Not to be used for new frames.
+
+      Some deprecated frames do not follow these conventions (e.g. MAV_FRAME_BODY_NED and MAV_FRAME_BODY_OFFSET_NED).
+ </description>
       <entry value="0" name="MAV_FRAME_GLOBAL">
         <description>Global (WGS84) coordinate frame + MSL altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="1" name="MAV_FRAME_LOCAL_NED">
-        <description>Local coordinate frame, Z-down (x: North, y: East, z: Down).</description>
+        <description>NED local tangent frame (x: North, y: East, z: Down) with origin fixed relative to earth.</description>
       </entry>
       <entry value="2" name="MAV_FRAME_MISSION">
         <description>NOT a coordinate frame, indicates a mission command.</description>
@@ -229,24 +245,24 @@
         <description>Global (WGS84) coordinate frame + altitude relative to the home position. First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
       </entry>
       <entry value="4" name="MAV_FRAME_LOCAL_ENU">
-        <description>Local coordinate frame, Z-up (x: East, y: North, z: Up).</description>
+        <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1.0e-7, second value / y: longitude in degrees*1.0e-7, third value / z: positive altitude over mean sea level (MSL).</description>
+        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
         <description>Global (WGS84) coordinate frame (scaled) + altitude relative to the home position. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
       </entry>
       <entry value="7" name="MAV_FRAME_LOCAL_OFFSET_NED">
-        <description>Offset to the current local frame. Anything expressed in this frame should be added to the current local frame position.</description>
+        <description>NED local tangent frame (x: North, y: East, z: Down) with origin that travels with the vehicle.</description>
       </entry>
       <entry value="8" name="MAV_FRAME_BODY_NED">
         <deprecated since="2019-08" replaced_by="MAV_FRAME_BODY_FRD"/>
-        <description>Setpoint in body NED frame. This makes sense if all position control is externalized - e.g. useful to command 2 m/s^2 acceleration to the right.</description>
+        <description>Same as MAV_FRAME_LOCAL_NED when used to represent position values. Same as MAV_FRAME_BODY_FRD when used with velocity/accelaration values.</description>
       </entry>
       <entry value="9" name="MAV_FRAME_BODY_OFFSET_NED">
         <deprecated since="2019-08" replaced_by="MAV_FRAME_BODY_FRD"/>
-        <description>Offset in body NED frame. This makes sense if adding setpoints to the current flight path, to avoid an obstacle - e.g. useful to command 2 m/s^2 acceleration to the east.</description>
+        <description>This is the same as MAV_FRAME_BODY_FRD.</description>
       </entry>
       <entry value="10" name="MAV_FRAME_GLOBAL_TERRAIN_ALT">
         <description>Global (WGS84) coordinate frame with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees, second value / y: longitude in degrees, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -216,11 +216,27 @@
       </entry>
     </enum>
     <enum name="MAV_FRAME">
+      <description>Co-ordinate frames used by MAVLink. Not all frames are supported by all commands, messages, or vehicles.
```
hamishwillee, line 219: Note, I figure describing the convention up front is helpful. This might not render properly.

message_definitions/v1.0/common.xml
```diff
@@ -216,11 +216,27 @@
...
+      
+      Global frames use the following naming conventions:
+      - `GLOBAL`: Global co-ordinate frame with WGS84 latitude/longitude and altitude positive over mean sea level (MSL) by default. 
+        The following modifiers may be used with `GLOBAL`:
+        - `RELATIVE_ALT`: Altitude is relative to the vehicle home position rather than MSL
+        - `TERRAIN_ALT`: Altitude is relative to ground level rather than MSL
+        - `INT`: Latitude/longitude (in degrees) are scaled by multiplying by 1E7  
+
+      Local frames use the following naming conventions:
+      - `LOCAL`: Origin of local frame is fixed relative to earth. Unless otherwise specified this origin is the origin of the vehicle position-estimator ("EKF").
```
hamishwillee, line 229: Note here that I've said "Unless otherwise specified this origin is the origin of the vehicle position-estimator ("EKF")". In the actual description of each frame I've just said "origin is fixed to earth,".

This allows us to use the frame with a different origin if we so wanted. Could simply fix the origin to EKF origin in each `LOCAL` frame if that makes more sense to others.

message_definitions/v1.0/common.xml
```diff
@@ -216,11 +216,27 @@
...
+      - `OFFSET`: Deprecated synonym for `BODY` (origin travels with the vehicle). Not to be used for new frames.
+
+      Some deprecated frames do not follow these conventions (e.g. MAV_FRAME_BODY_NED and MAV_FRAME_BODY_OFFSET_NED).
+ </description>
       <entry value="0" name="MAV_FRAME_GLOBAL">
         <description>Global (WGS84) coordinate frame + MSL altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="1" name="MAV_FRAME_LOCAL_NED">
-        <description>Local coordinate frame, Z-down (x: North, y: East, z: Down).</description>
+        <description>NED local tangent frame (x: North, y: East, z: Down) with origin fixed relative to earth.</description>
```
hamishwillee, line 239: I have used the term "local tangent frame" deliberately. This gives user a good chance of finding https://en.wikipedia.org/wiki/Local_tangent_plane_coordinates if they search :-)

message_definitions/v1.0/common.xml
```diff
@@ -229,33 +245,33 @@
         <description>Global (WGS84) coordinate frame + altitude relative to the home position. First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
       </entry>
       <entry value="4" name="MAV_FRAME_LOCAL_ENU">
-        <description>Local coordinate frame, Z-up (x: East, y: North, z: Up).</description>
+        <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1.0e-7, second value / y: longitude in degrees*1.0e-7, third value / z: positive altitude over mean sea level (MSL).</description>
+        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
```
hamishwillee, line 251: NOTE, this should go in whatever we do with the rest of this PR. I am pretty sure we multiply by 1E7 to end up with a big nubber rather than 1E-7 :-). This typo was everywhere here.

message_definitions/v1.0/common.xml
```diff
@@ -229,33 +245,33 @@
...
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + altitude relative to the home position. First value / x: latitude in degrees*10e-7, second value / y: longitude in degrees*10e-7, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
+        <description>Global (WGS84) coordinate frame (scaled) + altitude relative to the home position. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
       </entry>
       <entry value="7" name="MAV_FRAME_LOCAL_OFFSET_NED">
-        <description>Offset to the current local frame. Anything expressed in this frame should be added to the current local frame position.</description>
+        <description>NED local tangent frame (x: North, y: East, z: Down) with origin that travels with the vehicle.</description>
```
hamishwillee, line 257: FYI, this should be named `MAV_FRAME_BODY_NED` but unfortunately that is already defined prior to the convention. I have however spelled out the meaning consistent with the other ones. 

The frame currently named `MAV_FRAME_BODY_NED` is deprecated, but possibly still in use. We should work to really remove it and eventually reuse the name.

message_definitions/v1.0/common.xml
```diff
@@ -216,11 +216,27 @@
...
+      Global frames use the following naming conventions:
+      - `GLOBAL`: Global co-ordinate frame with WGS84 latitude/longitude and altitude positive over mean sea level (MSL) by default. 
+        The following modifiers may be used with `GLOBAL`:
+        - `RELATIVE_ALT`: Altitude is relative to the vehicle home position rather than MSL
+        - `TERRAIN_ALT`: Altitude is relative to ground level rather than MSL
+        - `INT`: Latitude/longitude (in degrees) are scaled by multiplying by 1E7  
+
+      Local frames use the following naming conventions:
+      - `LOCAL`: Origin of local frame is fixed relative to earth. Unless otherwise specified this origin is the origin of the vehicle position-estimator ("EKF").
+      - `BODY`: Origin of local frame travels with the vehicle.
```
Jaeyoung-Lim, line 230: I don't think this is correct. The body frame is not a local frame that travels with the vehicle, since the orientation changes with the vehicle also. I think the description is what `OFFSET` was supposed to be describing

olliw42: @all
you see guys, even with trying to define in words what "body" is supposed to mean it's hard to get it out of human's minds what they expect it to mean ... and misunderstanding and discussions take place ...
why not replacing it with e.g. "vehicle"

Jaeyoung-Lim: @olliw42 IMHO, the word body frame is something quite commonly used in aerospace / robotics and especially flight control for aerial vehicles - which defines the frame attached to the body of the aircraft.(It looks like it is stationary if you are to look at the frame while you are on the vehilce) If we change this to another word IMO this would cause even more confusion.

olliw42: the point is that the word "body" does NOT mean here the body frame ... and that's exactly what tripped you off !!! (and still seems to trip you off)

if you look into the other thread you will learn that "BODY" is supposed to relate to the ORIGIN of e.g. the NED frame, so it is NOT attached to the body in the sense it would change it's orientation as the real body frame would do, only the origin is attached to the body

the description "`BODY`: Origin of local frame travels with the vehicle." is technically absolutely correct

my point here and in this other thread was exactly what you just said, namely that there is a hardwired understanding of "body frame" ...

@all
I repeat, think that's a beautiful example of what happens if one tries to override hardwired definitions :):)

olliw42: if you want to stick with "BODY" I think you at least must make it MUCH clearer that one must not confuse it with the normal body frame ... e.g. by a big red warning comment ... (didn't I mentioned this hihi)

(I think I demonstrate here a better understanding of human limitations than you do LOL)

message_definitions/v1.0/common.xml
```diff
@@ -216,11 +216,27 @@
...
+      Global frames use the following naming conventions:
+      - `GLOBAL`: Global co-ordinate frame with WGS84 latitude/longitude and altitude positive over mean sea level (MSL) by default. 
+        The following modifiers may be used with `GLOBAL`:
+        - `RELATIVE_ALT`: Altitude is relative to the vehicle home position rather than MSL
+        - `TERRAIN_ALT`: Altitude is relative to ground level rather than MSL
+        - `INT`: Latitude/longitude (in degrees) are scaled by multiplying by 1E7  
+
+      Local frames use the following naming conventions:
+      - `LOCAL`: Origin of local frame is fixed relative to earth. Unless otherwise specified this origin is the origin of the vehicle position-estimator ("EKF").
+      - `BODY`: Origin of local frame travels with the vehicle. NOTE, `BODY` does NOT indicate alignment of frame axis with vehicle attitude.
```
hamishwillee, line 230: To continue the conversation in https://github.com/mavlink/mavlink/pull/1700/files#r709871027

@Jaeyoung-Lim  It is well worth having a read of the RFC to understand the problems with the old conventions and the solution offered by the new ones: https://github.com/mavlink/rfcs/pull/2

The new-style convention has a format like `MAV_FRAME` `ORIGIN` `FRAME`, where `ORIGIN` indicates the frame origin and is LOCAL or BODY, and `FRAME` is the frame alignment - NED, ENU, FRD etc. No frame alignments with vehicle are currently defined, but that would be done using RPY (according to the RFC) rather than BODY.

The problem I am trying to fix here is that the current definitions are ambiguous. The docs say:

> `MAV_FRAME_BODY_FRD`: Body fixed frame of reference, Z-down (x: Forward, y: Right, z: Down).

But I know that this was written to the new RFC conventions, which would more accurately be described as:

> `MAV_FRAME_BODY_FRD`: NED local tangent frame (x: North, y: East, z: Down) with origin that travels with the vehicle.

Note, "local tangent frame" indicates alignment with the earth plane.

@olliw42 Is quite right that the use of `BODY` for the origin here is confusing and unhelpful. I argued in the dev call that we should change the terminology for this origin to `COG` (or similar) - ie change the name from `MAV_FRAME_BODY_FRD` to `MAV_FRAME_COG_FRD`. However @auturgy pointed out that while this is desirable it would be a huge break.

I am open to suggestions that move us from a position of ambiguity to one of clarity. My preference would be one that removes the confusion from the naming convention in a way that is non-breaking and eventually leads to a happy place.

message_definitions/v1.0/common.xml
hamishwillee: Discussion moved to https://github.com/mavlink/mavlink/pull/1700#discussion_r713501137

message_definitions/v1.0/common.xml
hamishwillee: Note, I moved this to another PR.

auturgy:
Let's move this forward, but open a new issue with the outstanding clarification re BODY

hamishwillee:
Dev meeting - this is an improvement so let's get it in. As auturgy says above, will create separate issue/PR to address issues with naming of "BODY"

PR #1702, opened 2021-09-21
Global int frames - degrees*1E7 not E-7

hamishwillee:
Fix bug in global frames definition - integer representation is multiplied by degrees*1E7 not degrees*1E-7


julianoes (approved):
Right!

PR #1705, opened 2021-09-23
common: add storage usage flags

julianoes:
This solves a problem that I have with a drone which has internal memory
as well as a Micro SD card. I basically need a way to:
- Select which storage should be used by the camera.
- Be able to find out which storage to use to display the remaining space in the ground station.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -696,6 +696,27 @@
         <description>Storage type is other, not listed type.</description>
       </entry>
     </enum>
+    <enum name="STORAGE_USAGE_FLAG">
+      <description>Flags to indicate how a storage is being used.</description>
+      <entry value="1" name="STORAGE_USAGE_SET">
+        <description>To signal that this flag is set at all (if used as an extension).</description>
+      </entry>
+      <entry value="2" name="STORAGE_USAGE_UNUSED">
+        <description>Storage is currently unused.</description>
+      </entry>
+      <entry value="4" name="STORAGE_USAGE_USED_PRIMARILY">
+        <description>Storage is used primarily as main storage.</description>
+      </entry>
+      <entry value="8" name="STORAGE_USAGE_USED_SECONDARILY">
+        <description>Storage is used secondarily for redundancy.</description>
+      </entry>
+      <entry value="16" name="STORAGE_USAGE_USED_FOR_PHOTO">
+        <description>Storage is used to save photos.</description>
+      </entry>
+      <entry value="32" name="STORAGE_USAGE_USED_FOR_VIDEO">
+        <description>Storage is used to save videos.</description>
+      </entry>
+    </enum>
     <enum name="ORBIT_YAW_BEHAVIOUR">
       <description>Yaw behaviour during orbit flight.</description>
       <entry value="0" name="ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER">
@@ -1747,6 +1768,11 @@
         <param index="3" reserved="true" default="NaN"/>
         <param index="4" reserved="true" default="NaN"/>
         <param index="7" reserved="true" default="NaN"/>
+      </entry>
+      <entry value="533" name="MAV_CMD_SET_STORAGE_USAGE" hasLocation="false" isDestination="false">
+        <description>Specify storage usage, e.g. if a camera should save to internal memory or an external SD card.</description>
+        <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
+        <param index="2" label="Usage" enum="STORAGE_USAGE_FLAG">Usage flags</param>
       </entry>
       <entry value="600" name="MAV_CMD_JUMP_TAG" hasLocation="false" isDestination="false">
         <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
@@ -4218,7 +4244,7 @@
       </entry>
       <entry value="1" name="NAV_VTOL_LAND_OPTIONS_FW_DESCENT">
         <description>Descend in fixed wing mode, transitioning to multicopter mode for vertical landing when close to the ground.
-          The fixed wing descent pattern is at the discretion of the vehicle (e.g. transition altitude, loiter direction, radius, and speed, etc.).        
+          The fixed wing descent pattern is at the discretion of the vehicle (e.g. transition altitude, loiter direction, radius, and speed, etc.).
         </description>
       </entry>
       <entry value="2" name="NAV_VTOL_LAND_OPTIONS_HOVER_DESCENT">
@@ -5998,6 +6024,7 @@
       <extensions/>
       <field type="uint8_t" name="type" enum="STORAGE_TYPE">Type of storage</field>
       <field type="char[32]" name="name">Textual storage name to be used in UI (microSD 1, Internal Memory, etc.) This is a NULL terminated string. If it is exactly 32 characters long, add a terminating NULL. If this string is empty, the generic type is shown to the user.</field>
+      <field type="uint8_t" name="storage_usage" enum="STORAGE_USAGE_FLAG">Flags about usage of a storage</field>
     </message>
     <message id="262" name="CAMERA_CAPTURE_STATUS">
       <description>Information about the status of a capture. Can be requested with a MAV_CMD_REQUEST_MESSAGE command.</description>


message_definitions/v1.0/common.xml
```diff
@@ -696,6 +696,27 @@
...
+      </entry>
+      <entry value="2" name="STORAGE_USAGE_UNUSED">
+        <description>Storage is currently unused.</description>
+      </entry>
+      <entry value="4" name="STORAGE_USAGE_USED_PRIMARILY">
+        <description>Storage is used primarily as main storage.</description>
+      </entry>
+      <entry value="8" name="STORAGE_USAGE_USED_SECONDARILY">
+        <description>Storage is used secondarily for redundancy.</description>
+      </entry>
```
JonasVautherin, line 712: Not sure I get that one. If the `STORAGE_USAGE_USED_PRIMARILY` is not set, doesn't that mean that it is used secondarily? What would it mean if both are not set, for instance?

julianoes: If it's not used PRIMARILY or SECONDARILY it's not used, I'd say. Happy to change it if you have suggestions.

JonasVautherin: Oh, I see. Thanks!

JonasVautherin (approved):

MatejFranceskin (approved):
I think it is ok!

hamishwillee:
@julianoes So we just have  [STORAGE_INFORMATION](https://mavlink.io/en/messages/common.html#STORAGE_INFORMATION) which tells us what storage is available, it's type and gives us a flight-stack specific `name` for each unit - e.g. memory card 1, internal storage.

You should be able to use the `name` field to iterate what storage the camera has available in a way that is camera-relevant (internal, card 1, etc), but there does not appear to be any way to say _where_ you want data to be stored. The implication is that your mavlink camera can store the information on any storage it likes that has space available.

So agree something like this is needed. 

1. This should be in development.xml until prototyped.
2. Enum names are used as prefix by convention so STORAGE_USAGE_FLAG_UNUSED not STORAGE_USAGE_UNUSED. Wording is oriented as querying, not setting (best to make generic if possible). I'd suggest dropping `Flag` from enum name - but can/should we shove "CAM" in there to namespace this? Suggestion based purely on current design (... further questions to follow)

- `CAM_STORAGE_USAGE` - Flags for setting/getting how a storage device is used by the camera
  - `CAM_STORAGE_USAGE_ENABLE`: Enable if storage usage can be set. If 0 the camera uses its own (arbitrary) algorithm for data storage.
  - `CAM_STORAGE_USAGE_UNUSED` - Storage not used.
  - `CAM_STORAGE_USAGE_PRIMARY` - Primary storage.
  - `CAM_STORAGE_USAGE_SECONDARY` - Secondary (fallback) storage if primary storage is not available.
  - `CAM_STORAGE_USAGE_PHOTO` - Photos storage
  - `CAM_STORAGE_USAGE_VIDEO` - Video storage.

 3. My problem with the above is "assumptions on usage". This kind of assumes you will set up one thing as primary, one as secondary. But it isn't clear what a system should do if there are more than two storage, or if multiples are set up as primary or secondary. Here are my assumptions for your confirmation:
 
   - Storage marked as primary will be used before any storage marked as secondary.
   - If there are multiple things marked as primary the camera can use either as primary. It is up to the implementation to decide how. Should we specify that it can't use both/alternate?
   - if there are multiple things marked as secondary then any secondary can be used in any order (after the primary)
   - Upshot - we're happy that if we have multiples, we only have one first preference. I think this is OK - mostly I just care about internal and external.
   - if CAM_STORAGE_USAGE_PHOTO is not set it can be saved anywhere - up to the camera
   - if CAM_STORAGE_USAGE_PHOTO is set on secondary, should this be saved to the primary or the secondary?
   - if CAM_STORAGE_USAGE_PHOTO is set on primary but not secondary, can I save my photos to the secondary if primary runs out?

I suspect that we're really setting preferences. It might be that we should dump CAM_STORAGE_USAGE_PHOTO and CAM_STORAGE_USAGE_VIDEO?

julianoes:
@hamishwillee thanks for your comments. I'll go through it point by point and update the PR.

> 1. This should be in development.xml until prototyped.

I don't think that's possible because it's an extension to an existing message.

> 2. Enum names are used as prefix by convention so STORAGE_USAGE_FLAG_UNUSED not STORAGE_USAGE_UNUSED. Wording is oriented as querying, not setting (best to make generic if possible). I'd suggest dropping Flag from enum name - but can/should we shove "CAM" in there to namespace this? Suggestion based purely on current design (... further questions to follow)

Good catch, the prefix was missing.
I think whenever we have bits/flags instead of options/enum, we should call it FLAG, no?
I don't think we should add CAM to the name as it is not tied to a camera, neither with the existing storage_information nor with the updated one.

> 3. My problem with the above is "assumptions on usage".

I was trying to make the flags in a way that various use cases can be covered, of course inspired by the ones I know which are:
- One camera with one storage, all simple.
- A camera with internal storage and an SD card. The primary storage can be selected, and the free space should be presented accordingly in a ground station.
- A camera with two SD card slots for redundancy (primary and secondary).
- A camera that saves pictures via USB on a companion computer but saves video files on the internal SD card due to speed constraints.

So, I agree that UNUSED might be confusing and it's best to drop it. For the rest, I would argue it allows to mix and match the flags for these use cases.



> -  Storage marked as primary will be used before any storage marked as secondary.
I'd say so.
> -  If there are multiple things marked as primary the camera can use either as primary. It is up to the implementation to decide how. Should we specify that it should fill one and then the next (i.e. not arbitrary spread?)
I would say there is only one primary one.
> -  if there are multiple things marked as secondary then any secondary can be used in any order (after the primary)
But I guess there can be multiple secondary ones, as they are just mirroring the primary one.
> -  If something is not primary or secondary then it is unused. Do we then need unused? What if unused and primary are set. Upshot - options in a flag shouldn't be mutually exclusive if at all possible.
Ok, remove unused.
> -  Upshot - we're happy that if we have multiples, we only have one first preference. I think this is OK - mostly I just care about internal and external.
> -  if CAM_STORAGE_USAGE_PHOTO is not set it can be saved anywhere - up to the camera
Makes sense.
> -  if CAM_STORAGE_USAGE_PHOTO is set on secondary, should this be saved to the primary or the secondary?
If the primary one does not set STORAGE_USAGE_PHOTO, then photos will be stored somewhere else.
> -  if CAM_STORAGE_USAGE_PHOTO is set on primary but not secondary, can I save my photos to the secondary if primary runs out?
That's all somewhat implementation specific. Maybe the PHOTO/VIDEO flags are too complex and we should just drop them.

message_definitions/v1.0/common.xml
```diff
@@ -696,6 +696,24 @@
         <description>Storage type is other, not listed type.</description>
       </entry>
     </enum>
+    <enum name="STORAGE_USAGE_FLAG">
+      <description>Flags to indicate how a storage is being used.</description>
+      <entry value="1" name="STORAGE_USAGE_FLAG_SET">
+        <description>To signal that this flag is set at all (if used as an extension).</description>
+      </entry>
+      <entry value="2" name="STORAGE_USAGE_FLAG_USED_PRIMARILY">
```
hamishwillee, line 704: I would just call _PRIMARY and _SUMMARY, 

Similarly, STORAGE_USAGE_FLAG_VIDEO clearly means the same as STORAGE_USAGE_FLAG_USED_FOR_VIDEO and is a lot more succinct.

hamishwillee:
I think you can't look at your preferred use case and assume people will set it the same way. YOu have to assume someone will come along with 20 cameras and clarify your expectation because it is a shit to clean up later.

So if  only one storage can be primary, then the message should say that, and the command should reject any enum that sets a second primary storage.  Personally I think you're better off saying that primary just means it will be used before any secondaries. That essentially means that you can set just one primary if you like, but if someone else want two that's OK - you don't have to check that there are no duplicates. In other words you can always say "save internal" to mean just one particular drive if you want.

What about if you have just one storage - does that mean the storage is set primary all the time, or would you omit the flag to indicate that primary/secondary don't matter?

I think the Photo/Video might be too complex - mostly because they are orthogonal to primary/secondary. How about 
- used for storage = if 1 then it is used, if not then it is not used ever.
- primary for photos = it is a primary for photos (otherwise is secondary)
- primary for videos = it is a primary for videos (otherwise secondary)

If primary is set on more than one storage then it is used on that storage before going to secondaries, but which one is up to the implementation. 

I think that allows to set up all the sensible cases, and also any odd cases. What if we add new things to store though - is it still generic? Just a thought.

auturgy:
PRIMARILY and SECONDARILY are a bit of a mouthful.
Why not just primary/secondary or main/alternate?
Not a big deal, but looks neater with fewer letters :)

julianoes:
@hamishwillee I don't quite follow what you mean but feel free to add a commit to demonstrate it.

auturgy (approved):

hamishwillee:
I have simplified this a bit to minimal text and we can build on that.

What I think you want (?) is:
- USAGE_PHOTO and USAGE_VIDEO indicate primary location for those file types irrespective of other flags.
- PRIMARY indicates primary for other file types or if photo and video not specified.
- SECONDARY indicates a fallback for anything - if any other flag is specified you'd ignore this - so it might be "enabled"
- You have indicated that there should only be one primary, but IMO you don't care - all you care is that any primary are used before any fallback. 

A possible rework for you to consider is below. 

```xml
<enum name="STORAGE_USAGE_FLAG">
      <description>Flags to indicate how a storage is being used.
       Use STORAGE_USAGE_FLAG_ENABLED to indicate storage can be used, 
       and STORAGE_USAGE_FLAG_OTHER_PRIMARY to indicate the primary storage location for all files
       (if set on multiple storage the order in which the storage will be used is implementation dependent).
      </description>
      <entry value="1" name="STORAGE_USAGE_FLAG_SET">
        <description>Always set true (indicates the associated extension field is supported).</description>
      </entry>
      <entry value="2" name="STORAGE_USAGE_FLAG_ENABLED">
        <description>This storage may be used for images, video, or other media. If this is disabled the primary flags must be ignored.</description>
      </entry>
      <entry value="4" name="STORAGE_USAGE_FLAG_PHOTO_PRIMARY">
        <description>Primary storage for photos storage.</description>
      </entry>
      <entry value="8" name="STORAGE_USAGE_FLAG_VIDEO_PRIMARY">
        <description>Primary storage for videos.</description>
      </entry>
      <entry value="16" name="STORAGE_USAGE_FLAG_DEFAULT_PRIMARY">
        <description>Primary storage for files where no primary file storage has been set (including photos/videos).</description>
      </entry>
    </enum>
```

Is that closer? The only one I don't like is `STORAGE_USAGE_FLAG_ENABLED` because it has side effects - ie if you disable it then the primary flags are meaningless.

message_definitions/v1.0/common.xml
```diff
@@ -696,6 +696,21 @@
...
     </enum>
+    <enum name="STORAGE_USAGE_FLAG">
+      <description>Flags to indicate how a storage is being used.</description>
+      <entry value="1" name="STORAGE_USAGE_FLAG_SET">
+        <description>Always set true (indicates this extension field is supported).</description>
+      </entry>
+      <entry value="2" name="STORAGE_USAGE_FLAG_USED_FOR_PHOTO">
+        <description>Storage to be used to save photos.</description>
+      </entry>
+      <entry value="4" name="STORAGE_USAGE_FLAG_USED_FOR_VIDEO">
```
hamishwillee, line 707: Why are you so keen on `STORAGE_USAGE_FLAG_USED_FOR_VIDEO` over `STORAGE_USAGE_FLAG_VIDEO` ?

hamishwillee:
Much better. 
- What if nothing has these flags? My thinking is that if the flag is not set on a storage then that storage cannot be used for storing the file type. I.e. you'd have an error if you tried to save a picture and no storage type was specified for photos.
- What if multiple storage have these flags? 
  - My feeling is that is fine. The system will just store on any of the storage with the flags, in whatever order it likes.

  From a UI perspective though you might choose to only allow one. I.e invert the option you have say:
  - Drive for photos: 1, 2, 3
  - Drive for videos: 1, 2, 3
  - Drive for logs: 1, 2, 3.

julianoes:
> The difference being that allowing multiples to be selected means that you can allow overflow if a drive gets full.

I would just swap the flag as you overflow, in the camera implementation. So that way at least the ground station can always show how much is left. It would be slightly confusing because it would jump from 1 GB left (internal let's say) to 32 GB left when it switches to the SD card.

auturgy:
If the flags aren’t set I expect that the device would use its defaults.
I would be surprised and disappointed if a mavlink enabled camera (a) didn’t set a default value and/or (b) refused to operate because the storage flag wasn’t set.

julianoes:
> If the flags aren’t set I expect that the device would use its defaults.

Yes, agreed. Or a device could just announce its defaults anyway, and reject any configuration anyway. My main goal was really that the GCS can display the correct storage in the UI, so that the "space remaining" number is correct.

hamishwillee:
@julianoes @auturgy OK, I think the current update I just made catches all that. It basically says:
- Each flag can only be set on one storage.
- If flag is set that will succeed, clearing the flag if set on other storages.
- camera can ignore the command to set the preference - ACK the message as not supported
- camera can ignore the command to set the preference on a particular storage (ack failed).
- camera uses its own defaults if no flag set. 

I am assuming that a camera that chooses to use its own default save locations is NOT required to publish the preferred locations. It probably just returns 0 for the storage info?

julianoes (commented):
Yes, that's good now.

hamishwillee:
@julianoes Before I merge can you confirm whether a camera using default storage _may_, _must_, _should_, or _will not_ report a preferred storage?

In other words, say I didn't set the location for video OR the camera doesn't support setting defaults. The camera will still store the images somewhere. Since you said that you want to display the remaining storage how will you work out which drive to do that on if there is more than one.

My assumption is that the camera may or may not support telling you what its default storage is. So the answer to the above is probably that you say something like "The camera _should_ publish the default storage it is using if at all possible, even if the flags are not explicitly set by the user". And if it can't, your GUI will just have to live with it.

Right?

hamishwillee:
Perfect. I have added update to storage info. If you are happy with that, can you please merge. If not, update and I will recheck.

julianoes (commented):
Ok!

PR #1706, opened 2021-09-27
common: allow any compid to be rebooted

julianoes:
Instead of having hard-coded params to reboot a camera or gimbal/mount specifically, this change allows to reboot any (or all) components based on the MAVLink component ID.

Previously param3 and param4 were marked WIP, and I doubt they have been implemented anywhere.


hamishwillee:
@julianoes 

My understanding is that you send `MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN` to the system that manages updates/reboots on behalf of others - e.g. the companion. Is that right. I guess in theory you could have components acting on this themselves. Is that right?

Yes. This is a much better design. 
- Means you can use it for systems with multiple cameras or "things", albeit at the cost of having to map component IDs to types.
- I assume you are choosing to keep the Autopilot and Companion fields (even though we could use the new component setting) because these are the main use case, there are only one each of those and it would be a pain to map the component id) and because they are not marked WIP?
- At some point in future perhaps make the options into enums.

@auturgy Are you OK with this to merge? It is untested but it does make sense. Further, it will remove some more WIPs from common.xml :-). No need to respond now, will add to dev call for confirmation.

julianoes:
I assume you are choosing to keep the Autopilot and Companion fields (even though we could use the new component setting) because these are the main use case, there are only one each of those and it would be a pain to map the component id) and because they are not marked WIP?

Correct, because they might be implemented and are not marked WIP.

auturgy:
lgtm

PR #1708, opened 2021-09-29
development: add GIMBAL_DEVICE_DEBUG_STATUS

julianoes:
This is in addition to the usual GIMBAL_DEVICE_ATTITUDE_STATUS to provide more debugging information.

@cmic0 what else should we add here? Temperature? Torque?


hulk-ethan:
I've been thinking about this.  I think this should be included in the attitude message, like your previous PR.  In my mind, this is not a debugging message at all.  In the future, the aircraft will need to know everything about the kinematics of the gimbal at the same moment in time as the attitude information - not simply as a test message.  I don't think this should be added to temperature/torque/whatever.

hamishwillee:
@hulk-ethan @jeremyzff The main concern at the dev call was not that the joint angles were "wrong", but that:
- we might be missing a lot more useful information.
- this points to possible shortcoming in the bigger design - as per @jeremyzff comment in https://github.com/mavlink/mavlink/pull/1704#issuecomment-930429622 a more generalised approach might be better.

So the thinking was that we separate out the fields into their own message and evolve them in development.xml. Following prototyping we can then assess whether we have a workable solution that incorporates more of the edge cases. If at that time it makes sense to have it as part of the original message we would happily reconsider.

Hope that makes sense? (@julianoes please correct me if I am misrepresenting this)

julianoes:
@hulk-ethan see https://github.com/mavlink/mavlink/pull/1704#issuecomment-930836918.

julianoes:
@hulk-ethan @jeremyzff I would appreciate a response to https://github.com/mavlink/mavlink/pull/1704#issuecomment-930836918.

PR #1709, opened 2021-09-29
Clarify how MAV_MISSION_TYPE_ALL checksum is calculated

hamishwillee:
This updates the `MISSION_CHECKSUM` definition to clarify exactly how the `MAV_MISSION_TYPE_ALL` is calculated - my understanding is that the order in which you parse the mission, geofence and rally point part of the plan matters!

I also took the opportunity to make it _more_ clear that if you update just the mission (say) then the checksum you get is the mission checksum, not the `MAV_MISSION_TYPE_ALL` checksum. 

@IamPete1 @MatejFranceskin @julianoes This makes sense? Pete, I am assuming you did not do the "all" case in another order? If so, we can modify to match you.

Guys, this also counts as an introduction. Pete is implementing this on ArduPilot/Mission Planner. Matej will do on PX4 and (presumably) QGC and Julian will do on MAVSDK. If you have concerns about this, or suggestions for improvement then please get together and discuss them. You're all in Europe!

Pete, FYI there has been some work done to start integrating MAVSDK with ArduPilot. This means that with some luck soon we may be able to use MAVSDK as a first class test for this kind of change.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -195,13 +195,15 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="53" name="MISSION_CHECKSUM">
-      <description>Checksum for the current mission, rally points or geofence plan (a GCS can use this checksum to determine if it has a matching plan definition).
-        This message must be broadcast following any change to a plan (immediately after the MISSION_ACK that completes the plan upload sequence).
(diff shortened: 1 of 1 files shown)


MatejFranceskin (approved):
Makes sense to me. The order is the same as in the MAV_MISSION_TYPE enum.

IamPete1:
@peterbarker has a PR for the AP implementation before I got round it. https://github.com/ArduPilot/ardupilot/pull/17370

Looks like it does not yet handle the `MAV_MISSION_TYPE_ALL` case, so I guess we can do what ever.

hamishwillee:
OK. In it goes then.

PR #1710, opened 2021-09-30
common: add not feasible mission results

julianoes:
It came up that it should be possible for the flight controller to communicate with the ground station if an uploaded mission is not actually feasible.

Hence I'm adding two cases for where the upload is not feasible. This is probably not the full list. And I'm wondering if we should have only one `NOT_FEASIBLE` result and then communicate the exact reason via statustext/event.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2486,6 +2486,12 @@
       <entry value="15" name="MAV_MISSION_OPERATION_CANCELLED">
         <description>Current mission operation cancelled (e.g. mission upload, mission download).</description>
       </entry>
+      <entry value="16" name="MAV_MISSION_NOT_FEASIBLE_TOO_LONG">
+        <description>Mission is not feasible because it's too long for this airframe.</description>
+      </entry>
+      <entry value="17" name="MAV_MISSION_NOT_FEASIBLE_GEOFENCE">
+        <description>Mission is not feasible because it's outside of a geofence.</description>
+      </entry>
     </enum>
     <enum name="MAV_SEVERITY">
       <description>Indicates the severity level, generally used for status messages to indicate their relative urgency. Based on RFC-5424 using expanded definitions at: http://www.kiwisyslog.com/kb/info:-syslog-message-levels/.</description>


message_definitions/v1.0/common.xml
```diff
@@ -2486,6 +2486,12 @@
       <entry value="15" name="MAV_MISSION_OPERATION_CANCELLED">
         <description>Current mission operation cancelled (e.g. mission upload, mission download).</description>
       </entry>
+      <entry value="16" name="MAV_MISSION_NOT_FEASIBLE_TOO_LONG">
+        <description>Mission is not feasible because it's too long for this airframe.</description>
```
hamishwillee, line 2490: This means you can't fly the distance because the frame with a full battery would crash before the end? Does it take account of partial battery?

message_definitions/v1.0/common.xml
```diff
@@ -2486,6 +2486,12 @@
...
+      </entry>
+      <entry value="17" name="MAV_MISSION_NOT_FEASIBLE_GEOFENCE">
+        <description>Mission is not feasible because it's outside of a geofence.</description>
```
hamishwillee, line 2493: Mission is not feasible because it will cause a geofence violation - essentially catering for exclusion and inclusion geofences

hamishwillee:
If we include not feasible at all then I think having individual reasons at this point is better than status messages. 

My concern about feasibility is that some of these things only work with atomic upload of the whole plan. i.e I upload a mission then a geofence then a rally points (say). When I upload my new mission it will be compared against the existing geofence and can potentially fail.

I sit on the fence a bit for this one. Consider the too long test - 
- should be able to upload a mission when my battery has low battery - I could just add more batteries.
- What if my mission allows landing for battery recharge.
- It is not "certain" changing vehicle airframe and reboot clears a mission. So you've rejected the mission that might be valid if I just had my vehicle type set wrong?

My "gut" feeling is that we are always going to have to a pre-flight or pre-mission validity check anyway, and getting that right and informing the user is actually more important than failing here. Just my 2 bits.

message_definitions/v1.0/common.xml
julianoes: I guess the question is if a mission should have battery stops programmed in or not. To this day I have not seen multi battery missions being executed correctly, so that's a bigger problem to solve, in my opinion.

julianoes:
> When I upload my new mission it will be compared against the existing geofence and can potentially fail.

Yes, that's a good point.

> My "gut" feeling is that we are always going to have to a pre-flight or pre-mission validity check anyway, and getting that right and informing the user is actually more important than failing here. Just my 2 bits.

Yes, agreed.

julianoes:
Takway: needs more thought, and is probably not so easy. Closing for now. We can easily pick that back up once it comes up.

PR #1711, opened 2021-10-05
common: allow video recording to toggle

julianoes:
Some devices don't keep track of the video recording state but should nevertheless be able to toggle the state. By adding a parameter to toggle the recording, this can be accomplished.


hamishwillee:
@julianoes So to summarise/paraphrase a GCS
- would previously send `MAV_CMD_VIDEO_START_CAPTURE` to start video recording and MAV_CMD_VIDEO_STOP_CAPTURE to stop recording.
- can now send MAV_CMD_VIDEO_START_CAPTURE to toggle capture on and off. 

Can a toggle be implemented using the previous commands/without a toggle?
If not:
- how does a GCS know that the camera supports toggle and not on/off. 
- What should the camera do if it gets the wrong command?

Just feels like the GCS and the camera must know whether they are recording or not. So a toggle can be thought of as two different switches.

julianoes:
@hamishwillee correct.

> how does a GCS know that the camera supports toggle and not on/off.

It does not, that's a good point.

> Just feels like the GCS and the camera must know whether they are recording or not.

The problem I have is that in my setup the autopilot sends the command given RC inputs. Now the autopilot (so far) does not keep track of the camera's recording state, and in my opinion it should not have to. All it does is basically say "trigger a recording start/stop when pressed".

julianoes:
>It is easier to send a single message to toggle than to store state and send separate on/off messages.

I keep track of the state but it might change if a GCS also starts/stops a recording.

> If this is required, then I lean towards having a separate MAV_CMD_VIDEO_TOGGLE_CAPTURE command

That would be the alternative. And doesn't have the backward compatibility problem.

hamishwillee:
> With respect to the toggle message ...

So if you attach your vehicle to a GCS I am guessing it will display on the GCS as normal but nothing will happen if you press the record button. I am guessing with some luck that that the camera will get the command to start/stop recording and reject it as "NOT SUPPORTED". That way the GCS UI will be sub optimal - i.e. indicate it can record, but at least you'll be told it isn't going to. 

1. Is that right?

Longer term we'd like GCS to be able to handle a camera that can support either or both of the activation/deactivation messages. 
I think the only way we can do that is to tell the camera what is supported. So perhaps a new flag `CAMERA_CAP_FLAGS_SUPPORTS_VIDEO_CAPTURE_TOGGLE`. The assumption being that either you support toggle or on/off (is that reasonable?)

Anyway, I think that this should be a new message in development.xml, because clearly we have some thinking to do :-)

PR #1712, opened 2021-10-05
Move CELLULAR_STATUS back to common dialect

Jaeyoung-Lim:
**Problem Description**
The `CELLULAR_STATUS` message was removed from the common dialect in https://github.com/mavlink/mavlink/pull/1697 as it was considered that there were no public implementation. @hamishwillee 

However, the message is used in PX4(https://github.com/PX4/PX4-Autopilot/blob/6d78054f5001548ea2f1d48bb59966e2352bc9fe/src/modules/mavlink/mavlink_receiver.cpp#L206), and this change seems to break updating mavlink.

The message is quite useful to monitor the cellular status e.g. LTE module status of the vehicle in case it has cellular connections

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3618,6 +3618,49 @@
         <description>The fields next_lat, next_lon and next_alt contain valid data.</description>
       </entry>
     </enum>
+    <!-- cellular status information -->
+    <enum name="CELLULAR_STATUS_FLAG">
+      <description>These flags encode the cellular network status</description>
+      <entry value="0" name="CELLULAR_STATUS_FLAG_UNKNOWN">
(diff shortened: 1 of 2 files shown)


hamishwillee:
@Jaeyoung-Lim That's logging, not a proposed implementation as described when this was added to the spec. Without the history if you were to try add that as "vanilla" to common.xml I would say put it in development.xml until you can prove it actually works.

My assumption is that this is just convenience logging - the actual implementation is on a companion computer? We have a bit of a problem that it's hard to show these kinds of implementations because they aren't part of the autopilot code.

Anyway, I wouldn't have removed it had I know it would break you, so back in it goes. Note that it differs now from when it came out that it is no longer WIP. If you need to make breaking changes from now on, you can't.

PR #1713, opened 2021-10-05
Include ASLUAV into all dialect

Jaeyoung-Lim:
**Problem Description**
The ASLUAV was not included in the `all.xml` dialect due to a message id conflict with ardupilot `GIMBAL_CONTROL` messages.

- This PR moves the ASLUAV dialect into `all` dialect by moving conflicting message IDs in the ASLUAV dialect
- `SATCOM_LINK_STATUS` message is using one of the ardupilot dialect's mavlink 1 IDs as was discussed in the mavlink dev call some weeks ago

@tstastny @acfloria

### DIFF
--- a/message_definitions/v1.0/ASLUAV.xml
+++ b/message_definitions/v1.0/ASLUAV.xml
@@ -270,6 +270,7 @@
       <field type="uint8_t" name="sinr_ecio">SINR (LTE) or ECIO (WCDMA) as reported by modem (unconverted)</field>
       <field type="uint8_t" name="rsrq">RSRQ (LTE only) as reported by modem (unconverted)</field>
     </message>
+    <!-- Using mavlink 1 msg ID from the arudpilot reserved space, in order to save header size -->
     <message id="224" name="SATCOM_LINK_STATUS">
       <description>Status of the SatCom link</description>
       <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
--- a/message_definitions/v1.0/ASLUAV.xml
+++ b/message_definitions/v1.0/ASLUAV.xml
@@ -54,7 +54,7 @@
     </enum>
   </enums>
   <messages>
-    <message id="78" name="COMMAND_INT_STAMPED">
+    <message id="8000" name="COMMAND_INT_STAMPED">
       <description>Message encoding a command with parameters as scaled integers and additional metadata. Scaling depends on the actual command value.</description>
       <field type="uint32_t" name="utc_time">UTC time, seconds elapsed since 01.01.1970</field>
       <field type="uint64_t" name="vehicle_timestamp">Microseconds elapsed since vehicle boot</field>
@@ -72,7 +72,7 @@
       <field type="int32_t" name="y">PARAM6 / local: y position in meters * 1e4, global: longitude in degrees * 10^7</field>
       <field type="float" name="z">PARAM7 / z position: global: altitude in meters (MSL, WGS84, AGL or relative to home - depending on frame).</field>
     </message>
-    <message id="79" name="COMMAND_LONG_STAMPED">
+    <message id="8001" name="COMMAND_LONG_STAMPED">
       <description>Send a command with up to seven parameters to the MAV and additional metadata</description>
       <field type="uint32_t" name="utc_time">UTC time, seconds elapsed since 01.01.1970</field>
       <field type="uint64_t" name="vehicle_timestamp">Microseconds elapsed since vehicle boot</field>
@@ -88,14 +88,14 @@
       <field type="float" name="param6">Parameter 6, as defined by MAV_CMD enum.</field>
       <field type="float" name="param7">Parameter 7, as defined by MAV_CMD enum.</field>
     </message>
-    <message id="201" name="SENS_POWER">
+    <message id="8002" name="SENS_POWER">
       <description>Voltage and current sensor data</description>
       <field type="float" name="adc121_vspb_volt" units="V"> Power board voltage sensor reading</field>
       <field type="float" name="adc121_cspb_amp" units="A"> Power board current sensor reading</field>
       <field type="float" name="adc121_cs1_amp" units="A"> Board current sensor 1 reading</field>
       <field type="float" name="adc121_cs2_amp" units="A"> Board current sensor 2 reading</field>
     </message>
-    <message id="202" name="SENS_MPPT">
+    <message id="8003" name="SENS_MPPT">
       <description>Maximum Power Point Tracker (MPPT) sensor data for solar module power performance tracking</description>
       <field type="uint64_t" name="mppt_timestamp" units="us"> MPPT last timestamp </field>
       <field type="float" name="mppt1_volt" units="V"> MPPT1 voltage </field>
@@ -111,7 +111,7 @@
       <field type="uint16_t" name="mppt3_pwm" units="us"> MPPT3 pwm </field>
       <field type="uint8_t" name="mppt3_status"> MPPT3 status </field>
     </message>
-    <message id="203" name="ASLCTRL_DATA">
+    <message id="8004" name="ASLCTRL_DATA">
       <description>ASL-fixed-wing controller data</description>
       <field type="uint64_t" name="timestamp" units="us"> Timestamp</field>
       <field type="uint8_t" name="aslctrl_mode"> ASLCTRL control-mode (manual, stabilized, auto, etc...)</field>
@@ -139,7 +139,7 @@
       <field type="float" name="uAil"> </field>
       <field type="float" name="uRud"> </field>
     </message>
-    <message id="204" name="ASLCTRL_DEBUG">
+    <message id="8005" name="ASLCTRL_DEBUG">
       <description>ASL-fixed-wing controller debug data</description>
       <field type="uint32_t" name="i32_1"> Debug data</field>
       <field type="uint8_t" name="i8_1"> Debug data</field>
@@ -153,7 +153,7 @@
       <field type="float" name="f_7"> Debug data</field>
       <field type="float" name="f_8"> Debug data</field>
     </message>
-    <message id="205" name="ASLUAV_STATUS">
+    <message id="8006" name="ASLUAV_STATUS">
       <description>Extended state information for ASLUAVs</description>
       <field type="uint8_t" name="LED_status"> Status of the position-indicator LEDs</field>
       <field type="uint8_t" name="SATCOM_status"> Status of the IRIDIUM satellite communication system</field>
@@ -161,7 +161,7 @@
       <field type="float" name="Motor_rpm"> Motor RPM </field>
       <!-- to be extended-->
     </message>
-    <message id="206" name="EKF_EXT">
+    <message id="8007" name="EKF_EXT">
       <description>Extended EKF state estimates for ASLUAVs</description>
       <field type="uint64_t" name="timestamp" units="us"> Time since system start</field>
       <field type="float" name="Windspeed" units="m/s"> Magnitude of wind velocity (in lateral inertial plane)</field>
@@ -171,7 +171,7 @@
       <field type="float" name="beta" units="rad"> Sideslip angle</field>
       <field type="float" name="alpha" units="rad"> Angle of attack</field>
     </message>
-    <message id="207" name="ASL_OBCTRL">
+    <message id="8008" name="ASL_OBCTRL">
       <description>Off-board controls/commands for ASLUAVs</description>
       <field type="uint64_t" name="timestamp" units="us"> Time since system start</field>
       <field type="float" name="uElev"> Elevator command [~]</field>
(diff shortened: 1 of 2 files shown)


hamishwillee (approved):
@Jaeyoung-Lim The ardupilot ID is one you agreed with @auturgy ?

If you can confirm that, this looks great to merge.

peterbarker:
@Jaeyoung-Lim I might be missing something here, but I don't see *any* `SATCOM_STATUS` message here.

I think we might need more comments floating around in the .xml to indicate what's going on.  If there's a mavlink1 id ostensibly in ArduPilot's court but only used by the ASLUAV dialect, I think we can swing that (I'll raise it at DevCall tonight) - but we need comments explaining what's going on :-)

Is there also a comment somewhere indicating that ASLUAV now owns some space @ 8000+?

hamishwillee:
@peterbarker I didn't even think of the ranges - good catch. 

The ranges that I "know about" are documented in 
- Messages: https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges
- Commands: https://mavlink.io/en/guide/define_xml_element.html#command_values

That seems overly hidden. Wouldn't the best place for this information to live be in the respective XML? For now a comment, but maybe as XML so that you could generate a warning if an incorrect value is used?

With respect to the id, I never really understood why a MAVLink 1 id is needed. Something to do with a reduced size header - but I thought all messages sent using MAVLink 2 used the MAVLink 2 serialisation - so I don't really "get" the benefit.

peterbarker:
On Tue, 5 Oct 2021, Hamish Willee wrote:

> The ranges that I "know about" are documented in
>  *  Messages: https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges
>  *  Commands: https://mavlink.io/en/guide/define_xml_element.html#command_values

> That seems overly hidden. Wouldn't the best place for this information to live be in the respective XML? For now a comment, but maybe as XML so that you could generate a warning if an incorrect value is used?

Yes, that sounds useful!

I was also thinking just as a comment at the end of each line in all.xml - 
as a kind of quick-peek as to what's where, too.  Perhaps too many places 
potentially getting out-of-sync?

> With respect to the id, I never really understood why a MAVLink 1 id is needed. Something to do with a reduced size header - but I thought all messages sent using MAVLink 2 used the MAVLink 2 serialisation - so I don't really "get" the benefit.

I'm not going to argue with them on the point today :-)

mavlink2 headers are larger - they've got compat/incompat flags at the 
least.

Perhaps the message is designed to be used over a satellite link using 
only mavlink1 and every bit is precious?

It might pay to move it up as well straight away if this change is as 
breaking as it looks - more input from PR author required :-)

hamishwillee:
@peterbarker I have a slight preference for ranges in the the dialect file itself because I think that is slightly more discoverable for users coming into the community. It also makes more sense if we add tooling, because mavgen runs on individual dialect files.

The benefit of having ranges in one place is that it makes it easier to see all the ranges together and allocate new ranges for people. That could be all.xml. If we were doing tool checks, it might even be a separate ranges.xml file we use for tracking this stuff (and import in every build by default). Overall though, least churn is in the dialect file.

As you said though, must be in one place only. Do you have a preference?

@Jaeyoung-Lim Can you confirm that the satellite link is a MAVLink 1 communication channel? If not then it could be a MAVLink 2 ID.

peterbarker:
ArduPilot DevCall is OK with a mavlink1 ID from "our" reserved space - but I'd really like more information first :-)

Jaeyoung-Lim:
> @Jaeyoung-Lim I might be missing something here, but I don't see any SATCOM_STATUS message here.

Sorry, I was supposed to write `SATCOM_LINK_STATUS` 

> I think we might need more comments floating around in the .xml to indicate what's going on. If there's a mavlink1 id ostensibly in ArduPilot's court but only used by the ASLUAV dialect, I think we can swing that (I'll raise it at DevCall tonight) - but we need comments explaining what's going on :-)

I thought that this was already agreed upon with @auturgy , but I will be happy to add comments explaining this. The reason is that we want to save the amount of bytes sended over through the satcom by using mavlink 1 messages.

Jaeyoung-Lim:
@peterbarker @hamishwillee I have stated the ID range and also reason of the `SATCOM_LINK_STATUS` message using mavlink 1 ID range.

On our side we are open to moving message definitions that is commonly found useful to the more standard dialects. For example it is quite useful to use `SENSOR_AIRFLOW_ANGLES` to display airflow angles on the GCS. 

However, I believe this is a discussion that is separate from this PR

peterbarker:
> Sorry, I was supposed to write `SATCOM_LINK_STATUS`

All good, I see it now :-)
 
> I thought that this was already agreed upon with @auturgy , but I will be happy to add comments explaining this. The reason is that we want to save the amount of bytes sended over through the satcom by using mavlink 1 messages.

Cool.  We also want a comment in `ardupilotmega.xml` saying that the ID is used elsewhere - it will save questions later when it conflicts for someone... (there's an existing comment which says the ID is reserved for GOPRO which needs to be modified too)

I'm curious about the size-savings.  Not the shape of message I'd use for byte-minimisation if I was going all-out - the 64-bit timestamps are almost certainly overkill, and what looks like entire bytes used for a "pending" flags?!  Still, it's pre-existing and presumably a bunch of stuff depends on the current shape, making that shape non-negotiable.

FWIW, this is what I'd be doing:
```
    <message id="214" name="SATCOM_LINK_STATUS">
      <description>Status of the SatCom link</description>
      <field type="uint32_t" name="timestamp" units="ms">Timestamp</field>
      <field type="uint32_t" name="last_heartbeat" units="ms">Timestamp of the last successful sbd session</field>
      <field type="uint8_t" name="failed_sessions">Number of failed sessions</field>
      <field type="uint8_t" name="successful_sessions">Number of successful sessions</field>
      <field type="uint8_t" name="signal_quality">Signal quality</field>
      <field type="uint8_t" name="flags" enum=""SATCOM_LINK_STATUS_FLAGS">Flags</field>
    </message>
```

The size of a message is dependent on the encoding used, of course - 6 byte header for mavlink1, 10-byte header for mavlink2.  However, mavlink2 also includes "trailing null trimming".  My guess is that those last three bytes in `SATCOM_LINK_STATUS` will be zero the vast majority of the time (and will this be trimmed), so you might consider making this making a mavlink2 message-id for simplicity.  Your call, `224` is yours if you want it.

peterbarker:
> On our side we are open to moving message definitions that is commonly found useful to the more standard dialects. For example it is quite useful to use `SENSOR_AIRFLOW_ANGLES` to display airflow angles on the GCS.

That sounds good.  If we'd spotted that we might not have created our own message (rather more recently than yours, I think):
```
    <!-- Angle of Attack and Side Slip Angle message -->
    <message id="11020" name="AOA_SSA">
      <description>Angle of Attack and Side Slip Angle.</description>
      <field type="uint64_t" name="time_usec" units="us">Timestamp (since boot or Unix epoch).</field>
      <field name="AOA" type="float" units="deg">Angle of Attack.</field>
      <field name="SSA" type="float" units="deg">Side Slip Angle.</field>
    </message>
```

I believe we would transport "invalid" by sending a NaN in the relevant field rather than a separate `uint8_t` flag.

> However, I believe this is a discussion that is separate from this PR

Indeed :-)

In particular,     `<message id="210" name="FW_SOARING_DATA">` caught my eye.  Is there a GCS out there that renders that?  I'll bring it to the attention of our soaring maintainer.

Jaeyoung-Lim:
@peterbarker We have our custom QGC build that displays that information. You can see it working in https://youtu.be/8m76Mx9m2nM?t=161

hamishwillee:
@Jaeyoung-Lim I am very confused. Your last change moved `SATCOM_LINK_STATUS` **out** of the mavlink 1 range and command_long_stamped and command_int_stamped **into** mavlink 1 range. So I would have merged this but now I need an explanation/justification and consistent note for the ID usage as it is not in mavlink 1 range.

image

Before we merge, @Jaeyoung-Lim is there any chance of migrating to `SATCOM_LINK_STATUS` as suggested by @peterbarker (or some negotiated compromise)? My thinking is that if this is a shared use case that we might want in common.xml, I'd rather move towards that now while you're breaking your dependencies anyway. @peterbarker I assume you would consider this for common if it was changed as you suggest?
```
    <message id="214" name="SATCOM_LINK_STATUS">
      <description>Status of the SatCom link</description>
      <field type="uint32_t" name="timestamp" units="ms">Timestamp</field>
      <field type="uint32_t" name="last_heartbeat" units="ms">Timestamp of the last successful sbd session</field>
      <field type="uint8_t" name="failed_sessions">Number of failed sessions</field>
      <field type="uint8_t" name="successful_sessions">Number of successful sessions</field>
      <field type="uint8_t" name="signal_quality">Signal quality</field>
      <field type="uint8_t" name="flags" enum=""SATCOM_LINK_STATUS_FLAGS">Flags</field>
    </message>
```

With respect to the other messages it sounds like you've found some common use cases - if so, then can we move the versions of messages that are common and useful into common.xml? If there is conflict, the design of the ardupilot message sounds better to me.

peterbarker:
On Wed, 6 Oct 2021, Hamish Willee wrote:

> With respect to the other messages it sounds like you've found some common use cases - if so, then can we move the versions of messages that are common and useful into common.xml? If there is conflict, the design of the ardupilot message sounds better to me.

Messages should be in-common.xml when they're in-common between autopilots 
:-)

hamishwillee:
> Messages should be in-common.xml when they're in-common between autopilots

@peterbarker Yes. The second rule of MAVLink is that common functionality should be made common in autopilots :-)

hamishwillee:
@peterbarker Question for you arising from https://github.com/mavlink/mavlink/pull/1713#issuecomment-940883653: 

ASLUAV need TWO ids for the messages that will be sent over MAVLink 1 Satcom link. Are you happy for 223 and 224 in the ArduPilot range to be used for `COMMAND_INT_STAMPED` and `COMMAND_LONG_STAMPED`? (rather than just the one for the satcom message which was originally requested)

> I am not opposed to moving the SATCOM_LINK_STATUS to common in a slightly adjusted message definition.

@acfloria The alternative version Peter is proposing here for inclusion in common is the same as yours but with changes as marked up below. You said you would accept  _a slightly adjusted message definition_ to go into common. Can you confirm that you mean you are OK for **this** proposal to go in?:

```xml
    <message id="214" name="SATCOM_LINK_STATUS">
      <description>Status of the SatCom link</description>
      <field type="uint64_t" name="timestamp" units="us">Timestamp</field>
      <field type="uint64_t" name="last_heartbeat" units="us">Timestamp of the last successful sbd session</field>
      <field type="uint16_t" name="failed_sessions">Number of failed sessions</field>
      <field type="uint16_t" name="successful_sessions">Number of successful sessions</field>
      <field type="uint8_t" name="signal_quality">Signal quality</field>
      <!-- <field type="uint8_t" name="ring_pending">Ring call pending</field> -->
      <!-- <field type="uint8_t" name="tx_session_pending">Transmission session pending</field> -->
      <!-- <field type="uint8_t" name="rx_session_pending">Receiving session pending</field> -->
      + <field type="uint8_t" name="flags" enum=""SATCOM_LINK_STATUS_FLAGS">Flags</field>
    </message>
```

@peterbarker I am assuming here that you would be happy for this to go into common if it were changed as you suggested? Would be good for you to confirm because if this is not "common enough" then let's not force ASLUAV to change it.

hamishwillee:
@acfloria Thanks. OK, so let's see what @peterbarker says. I need him to confirm the allocation of the 2 IDs before this can merge.

I would highly recommend you update the message to match his version of `SATCOM_LINK_STATUS`. I wpuld then later move this into common.xml if two autopilots agree to integrate it.

hamishwillee:
@acfloria OK, so ArduPilot have agreed the two mavlink IDs can be used for this purpose. 

With respect to the `SATCOM_LINK_STATUS` there is need for a common message, but they are not sure that Peter's design is sufficient and do not have bandwidth to discuss at this time. 
So can modify the `SATCOM_LINK_STATUS` **if you want** but there is no guarantee that it might not have to change again at some point.

Upshot, if you don't want to modify the `SATCOM_LINK_STATUS` at this time I can merge this, otherwise I can wait until you've made whatever changes you want. Let me know if you want me to merge now?

tridge:
> ASLUAV need TWO ids for the messages that will be sent over MAVLink 1 Satcom link. Are you happy for 223 and 224 in the ArduPilot range to be used for `COMMAND_INT_STAMPED` and `COMMAND_LONG_STAMPED`? (rather than just the one for the satcom message which was originally requested)

I'm happy to lose two of the unused IDs below 256 to help get us all onto all.xml

hamishwillee:
Thanks @tridge !

hamishwillee:
OK. I am merging this. @acfloria @Jaeyoung-Lim Happy to take an updated `SATCOM_LINK_STATUS` in your dialect at any time, if you so wish.

PR #1718, opened 2021-10-13
Add MAV_AUTOPILOT_REFLEX.

m-ou-se:
This allocates MAV_AUTOPILOT number 20 for our flight controller.


hamishwillee (approved):
@julianoes @auturgy @LorenzMeier Any objections to this ID allocation? I see no reason not to merge (and I have approved it). I'm only asking because this is an update to minimal.xml.

If there is no response I will merge next week.

PR #1725, opened 2021-10-20
common: add target for SERIAL_CONTROL

julianoes:
It turns out the SERIAL_CONTROL messages did not have a target which means they are broadcast even though they are really meant for one system and component only.

I know adding this prevents truncating the message. The only workaround for that would be to create a new message to replace it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5563,6 +5563,9 @@
       <field type="uint32_t" name="baudrate" units="bits/s">Baudrate of transfer. Zero means no change.</field>
       <field type="uint8_t" name="count" units="bytes">how many bytes in this transfer</field>
       <field type="uint8_t[70]" name="data">serial data</field>
+      <extensions/>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
     </message>
     <message id="127" name="GPS_RTK">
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes this is sent from a GCS to a flight stack, and the flight stack then uses it to set the telemetry port (say) - is that correct?

So your concern is that if a GCS sends this into a network with multiple vehicles, they might all be configured using the message?
My further assumption being that you would call this relatively rarely - if its a configuration thing you probably do once, maybe when connected by USB and then forever ignore. 

If so, then IMO this seems like a very good solution - non-updated systems will send zero or  ignore these fields which has the same meaning. This can't break anything existing and while there will be wasteful messages they're not streamed continuously. 
If not, then our only option right now would be to add a new message and deprecate this.

@peterbarker As @auturgy is out of office, could you provide your "ArduPilot" opinion on this?

peterbarker (approved):
This looks right to me.

I imagine that an autopilot should return to `0, 0` if-and-only-if they received from `0,0`.

bkueng (approved):
> The only workaround for that would be to create a new message to replace it.

That would make sense anyway. There are some things I miss, like retransmission and sessions. That would require more work though.

julianoes:
> Ah, suspicions confirmed, than :-)
> I mean, yeah, bugs in other people's code, but we do need to be cognizant.

Yes but I think this is broken for extensions and non-extension fields, right? Or do only extensions get trimmed?

PR #1726, opened 2021-10-21
[WIP] Remove MAV_CMD_DO_UPGRADE from development.xml

hamishwillee:
`MAV_CMD_DO_UPGRADE` was removed from common.xml to development.xml in #1720 .

This is proposal to delete altogether, as it does not appear to be used anywhere (at all) and there are better message designs that could be used - e.g. generic component operations including upgrade, and also capturing everything done by https://mavlink.io/en/messages/common.html#MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN

Not merging yet, because want final confirmation before doing so.


hamishwillee:
Closing. Already removed

PR #1727, opened 2021-10-26
common.xml: use top 16 bits of AUTOPILOT_VERSION.board_version for board-id

peterbarker:
Concept: include a board ID from the `board_types.txt` in the mavlink stream to help GCSs out.  Can be useful when verifying parameters are correct.  Can be useful when working out whether there's a firmware upgrade for a board.

The WIP ArduPilot PR to use this is here: https://github.com/ArduPilot/ardupilot/pull/19001/files

I'm a bit leery of tying mavlink definitions to specifications not-actually-mavlink, but this would be really rather useful, I think.  OTOH, the existing "silicon-id" kind of already makes assumptions which might not be valid depending on implementation.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5802,7 +5802,7 @@
       <field type="uint32_t" name="flight_sw_version">Firmware version number</field>
       <field type="uint32_t" name="middleware_sw_version">Middleware version number</field>
       <field type="uint32_t" name="os_sw_version">Operating system version number</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@peterbarker I changed this from "last 8 bytes" to "last 4 bytes" as suggested by @bkueng. Please confirm that you're happy to merge with this change.

hamishwillee (approved):

peterbarker (commented):
Updated patch looks good.

hamishwillee:
@dagar - merged, so you can make the PX4 changes.

PR #1729, opened 2021-10-27
Component information: add actuators

bkueng:
This adds a new metadata file for actuator and geometry/mixer configuration and actuator testing.
The PR contains an example metadata file.

It also adds a new actuator test command. I'm aware of the existing `MAV_CMD_DO_MOTOR_TEST` message. The one I'm adding here is different in that it operates on an output function level. It allows to specify 'Motor 1' independent from the output where Motor 1 is physically connected.
There is a set of predefined output functions, however the UI in QGC will use the custom range >=1000, as it learns the actually available functions dynamically from the metadata.

QGC implementation: https://github.com/mavlink/qgroundcontrol/pull/9952

This adds to the `development.xml`, but QGC is using `ardupilotmega.xml` by default. @DonLakeFlyer @hamishwillee how do you want to proceed with this?

TODO:
- [x] add a schema
- [ ] add documentation


hamishwillee:
@bkueng I've had a bit of a look at ArduPilot control allocation. They essentially have two approaches. The first/original model essentially hard-codes what PX4 would call the geometry but does allow mapping of outputs to functions using parameters.

The second way uses [lua scripting](https://ardupilot.org/copter/docs/common-lua-scripts.html) to script the geometry and mixing. I think this also abstracts the outputs. It is very powerful in that the script can effectively dynamically modify geometry on the fly - supporting [complex geometries like this](https://github.com/ArduPilot/ardupilot/blob/master/libraries/AP_Scripting/examples/quadruped.lua) which you just can't do with a static definition. It can also set motor test orders etc.

The stuff to map motors and servos to outputs might therefore be generically useful to ArduPilot for the "original" case. It would be less useful for the lua scripts because these can be added and removed from the frame at will. I don't think the geometry stuff would be particular useful to them, though I guess I could be wrong.

I'm not sure what that means for the component stuff because we don't have explicit "rules" on that. 

Can you split the messages out into their own PR? I think we can merge that pretty easily and then discuss the component definition separately.

hamishwillee:
Thanks for making the change. I've requested review of the other PR and sent you some further offline discussion of this one.

hamishwillee:
Just to summarize the discussion of the [dev call](https://github.com/mavlink/mavlink/wiki/20211110-Dev-Meeting) this is "broadly approved" as consistent with the existing schema and design of component information.
It is waiting on minor changes from Beat and then I can merge (once pinged).

It should be useful for any flight stack that can assign functions and output mappings using parameters. It can also be useful for any flight stack that can define a geometry in terms of parameters, and that also applies to both PX4 and ArduPilot.
ArduPilot allows geometry/mixing outputs to be dynamically defined using lua scripts. Provided the lua scripts use parameters to define the geometry this approach still works but the information file could not be hard coded and there might need to be a discussion on how to get the CRC to work properly.

hamishwillee:
@bkueng With respect to the image used for the geometry settings, I know there is a fallback to "no image" and that is "enough". 

However the general concept we try for is that a GCS can work with any flight stack for standard kinds of things without having to have hard coded. So it would be cool if we can provide some mechanisms to allow configuration of standard frames at least - e.g. quad-X, etc.

Do you think it is _possible_/reasonable we provide a standard set of UIs that different flight stacks might use - e..g. a key  to say this geometry is for a Quad-X  with motors numbered like this so that if the GCS knows what that is, it can use the image type?

bkueng:
> Do you think it is possible/reasonable we provide a standard set of UIs that different flight stacks might use - e..g. a key to say this geometry is for a Quad-X with motors numbered like this so that if the GCS knows what that is, it can use the image type?

Yes I intended to add that once it's needed (it's a pure addition, non-breaking change), but I don't want to specify this w/o having at least one image in a GCS for this.
Different flight stacks typically have different actuator numberings.
And for the Quad-X example, this is already possible with the generic rendering: the metadata geometry would just contain the fixed motor positions (something I tested already), which allows any motor ordering to correctly display.

hamishwillee:
Thank you! Waiting on you to ping me or Julian to merge, because you need it in.

hamishwillee:
Merging, as agreed in dev call.

Issue #1730, opened 2021-10-28
docs, common: remove references to the removed GLOBAL_POSITION


dayjaby:
There are some references to GLOBAL_POSITION, which seems to be removed in the docs, e.g. `See message GLOBAL_POSITION for the global position estimate.` Clicking on `GLOBAL_POSITION` in this case does nothing.

hamishwillee:
Thanks for the PR.

PR #1731, opened 2021-10-29
common: remove references to GLOBAL_POSITION

dayjaby:
As mentioned in #1730 , there is no GLOBAL_POSITION anymore, so it makes sense to remove any references to it or replace them with GLOBAL_POSITION_INT.


hamishwillee (approved):
Thanks very much @dayjaby !

PR #1733, opened 2021-11-01
gimbalv2: improved documentation

dayjaby:
Following changes:

- corrected the identical amount of bits for `GIMBAL_[MANAGER|DEVICE]_CAP_FLAGS` and `GIMBAL_[MANAGER|DEVICE]_FLAGS`.
- made the documentation for the reference frame in `GIMBAL_DEVICE_ATTITUDE_STATUS` more explicit

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -449,7 +449,7 @@
       </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_CAP_FLAGS" bitmask="true">
-      <description>Gimbal manager high level capability flags (bitmap). The first 16 bits are identical to the GIMBAL_DEVICE_CAP_FLAGS which are identical with GIMBAL_DEVICE_FLAGS. However, the gimbal manager does not need to copy the flags from the gimbal but can also enhance the capabilities and thus add flags.</description>
+      <description>Gimbal manager high level capability flags (bitmap). The first 12 bits are identical to the GIMBAL_DEVICE_CAP_FLAGS which are identical with GIMBAL_DEVICE_FLAGS. However, the gimbal manager does not need to copy the flags from the gimbal but can also enhance the capabilities and thus add flags.</description>
       <entry value="1" name="GIMBAL_MANAGER_CAP_FLAGS_HAS_RETRACT">
         <description>Based on GIMBAL_DEVICE_CAP_FLAGS_HAS_RETRACT.</description>
       </entry>
@@ -512,7 +512,7 @@
       </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_FLAGS" bitmask="true">
-      <description>Flags for high level gimbal manager operation The first 16 bytes are identical to the GIMBAL_DEVICE_FLAGS.</description>
+      <description>Flags for high level gimbal manager operation The first 5 bytes are identical to the GIMBAL_DEVICE_FLAGS.</description>
       <entry value="1" name="GIMBAL_MANAGER_FLAGS_RETRACT">
         <description>Based on GIMBAL_DEVICE_FLAGS_RETRACT</description>
       </entry>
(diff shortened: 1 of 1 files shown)


dayjaby:
`GIMBAL_DEVICE_ATTITUDE_STATUS` is still a bit ambiguous: Is only the yaw in global frame, if the flag is set? How to send yaw in global frame, pitch/roll in body frame? Are ROLL_LOCK and PITCH_LOCK flags completely irrelevant here?

message_definitions/v1.0/common.xml
```diff
@@ -449,7 +449,7 @@
       </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_CAP_FLAGS" bitmask="true">
-      <description>Gimbal manager high level capability flags (bitmap). The first 16 bits are identical to the GIMBAL_DEVICE_CAP_FLAGS which are identical with GIMBAL_DEVICE_FLAGS. However, the gimbal manager does not need to copy the flags from the gimbal but can also enhance the capabilities and thus add flags.</description>
+      <description>Gimbal manager high level capability flags (bitmap). The first 12 bits are identical to the GIMBAL_DEVICE_CAP_FLAGS which are identical with GIMBAL_DEVICE_FLAGS. However, the gimbal manager does not need to copy the flags from the gimbal but can also enhance the capabilities and thus add flags.</description>
```
julianoes, line 452: I disagree. The first 16 are supposed to be the same. It's just that bit 13-16 are not used/populated yet.

message_definitions/v1.0/common.xml
```diff
@@ -512,7 +512,7 @@
       </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_FLAGS" bitmask="true">
-      <description>Flags for high level gimbal manager operation The first 16 bytes are identical to the GIMBAL_DEVICE_FLAGS.</description>
+      <description>Flags for high level gimbal manager operation The first 5 bytes are identical to the GIMBAL_DEVICE_FLAGS.</description>
```
julianoes, line 515: Same here. 16 are reserved to be the identical.

message_definitions/v1.0/common.xml
```diff
@@ -6407,12 +6407,12 @@
     <message id="285" name="GIMBAL_DEVICE_ATTITUDE_STATUS">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Message reporting the status of a gimbal device. This message should be broadcasted by a gimbal device component. The angles encoded in the quaternion are in the global frame (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right). This message should be broadcast at a low regular rate (e.g. 10Hz).</description>
+      <description>Message reporting the status of a gimbal device. This message should be broadcasted by a gimbal device component. The angles encoded in the quaternion are in the global frame if the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right) or in body frame if the flag is not set. This message should be broadcast at a low regular rate (e.g. 10Hz).</description>
```
julianoes, line 6410: It's not really the "body" frame but just relative to the vehicle heading rather than North.

What about?
```suggestion
      <description>Message reporting the status of a gimbal device. This message should be broadcasted by a gimbal device component. The angles encoded in the quaternion are relative to absolute North if the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set (roll: positive is rolling to the right, pitch: positive is pitching up, yaw is turn to the right) or relative to the vehicle heading if the flag is not set. This message should be broadcast at a low regular rate (e.g. 10Hz).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -6407,12 +6407,12 @@
...
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS">Current gimbal flags set.</field>
-      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation, the frame is depends on whether the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set)</field>
+      <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation)</field>
```
julianoes, line 6415: I would repeat it again down here, actually. It's easy to be missed.

julianoes (commented):
Thanks for the corrections. I don't agree with all of them, see inline comments.

message_definitions/v1.0/common.xml
dayjaby: So we expect most gimbals to have their roll/pitch based on an accelerometer sensor? There is no gimbal that provides roll/pitch in a position encoded manner?

julianoes: If a gimbal has only encoders, it can use the attitude sent by the autopilot in [AUTOPILOT_STATE_FOR_GIMBAL_DEVICE](https://mavlink.io/en/messages/common.html#AUTOPILOT_STATE_FOR_GIMBAL_DEVICE) and then add its angles on top of it.

message_definitions/v1.0/common.xml
dayjaby: "are identical to the GIMBAL_DEVICE_CAP_FLAGS which are identical with GIMBAL_DEVICE_FLAGS" seems to be a wrong statement though?

dayjaby:
@julianoes Updated the PR. Thanks for your review :)

julianoes (approved):
Great, thank you @dayjaby!

PR #1735, opened 2021-11-03
common: request multiple CAMERA_IMAGE_CAPTURED

julianoes:
This adds a way to request multiple CAMERA_IMAGE_CAPTURED messages. This solves the problem where it takes a long time to request all of the messages if there are many pictures in storage.

For instance to request around 1000 pictures, it can easily take several minutes depending on the link roundtrip time.

This PR is mostly to start the discussion around how to deal with that problem. Maybe extending this request/message is not the right way but there and this should be done differently.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6205,7 +6205,13 @@
       <field type="int32_t" name="image_count">Total number of images captured ('forever', or until reset using MAV_CMD_STORAGE_FORMAT).</field>
     </message>
     <message id="263" name="CAMERA_IMAGE_CAPTURED">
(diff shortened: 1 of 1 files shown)


julianoes:
1. Ok makes sense.
2. send all images with sequence number > param1"
3. Ok that would make sense.

hamishwillee:
@julianoes OK, updated. @auturgy Look OK to you?

julianoes (commented):
I think this is good like that. Should we check in the dev call before merging this

hamishwillee:
Sure. Added to agenda. Probably just a rubber stamp as ArduPilot don't implement this.

auturgy:
Just merge it I think

hamishwillee:
Thanks!

PR #1737, opened 2021-11-03
DISCUSS: MAV_FRAME_BODY_ to MAV_FRAME_MAV_

hamishwillee:
The `MAV_FRAME` naming convention is that `_BODY_` means "a frame with origin that is centered on the vehicle". So `MAV_FRAME_BODY_FRD` means the "FRD tangent frame with origin centered on the vehicle".

This is confusing because  `_BODY_` frames are more generally used to refer to frames that aligned with the vehicle body/attitude rather than the earth plane.

The proposal is to change the convention to use `MAV` (or something else other than `BODY`) to indicate a frame with the origin centered on the vehicle. In practice this means deprecating `MAV_FRAME_BODY_FRD` and adding `MAV_FRAME_MAV_FRD`.

This is a bit of a mouthful but "MAV" is synonymous with a "mavlink vehicle" and not a frame so it is a reasonable option. Other options:
- MAV_FRAME_IMU_FRD 
- MAV_FRAME_AHRS_FRD - [AHRS](https://en.wikipedia.org/wiki/Attitude_and_heading_reference_system) similar to IMU but a little more general
- MAV_FRAME_COG_FRD - apparently the COG is not a good choice for this.
- ?

Follows on from https://github.com/mavlink/mavlink/pull/1700.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -227,8 +227,8 @@
 
       Local frames use the following naming conventions:
       - `LOCAL`: Origin of local frame is fixed relative to earth. Unless otherwise specified this origin is the origin of the vehicle position-estimator ("EKF").
-      - `BODY`: Origin of local frame travels with the vehicle. NOTE, `BODY` does NOT indicate alignment of frame axis with vehicle attitude.
-      - `OFFSET`: Deprecated synonym for `BODY` (origin travels with the vehicle). Not to be used for new frames.
+      - `MAV`: Origin of local frame travels with the vehicle.
+      - `OFFSET`: Deprecated synonym for `MAV` (origin travels with the vehicle). Not to be used for new frames.
 
       Some deprecated frames do not follow these conventions (e.g. MAV_FRAME_BODY_NED and MAV_FRAME_BODY_OFFSET_NED).
  </description>
@@ -257,12 +257,12 @@
         <description>NED local tangent frame (x: North, y: East, z: Down) with origin that travels with the vehicle.</description>
       </entry>
       <entry value="8" name="MAV_FRAME_BODY_NED">
-        <deprecated since="2019-08" replaced_by="MAV_FRAME_BODY_FRD"/>
-        <description>Same as MAV_FRAME_LOCAL_NED when used to represent position values. Same as MAV_FRAME_BODY_FRD when used with velocity/accelaration values.</description>
+        <deprecated since="2019-08" replaced_by="MAV_FRAME_MAV_FRD"/>
+        <description>Same as MAV_FRAME_LOCAL_NED when used to represent position values. Same as MAV_FRAME_MAV_FRD when used with velocity/accelaration values.</description>
       </entry>
       <entry value="9" name="MAV_FRAME_BODY_OFFSET_NED">
-        <deprecated since="2019-08" replaced_by="MAV_FRAME_BODY_FRD"/>
-        <description>This is the same as MAV_FRAME_BODY_FRD.</description>
+        <deprecated since="2019-08" replaced_by="MAV_FRAME_MAV_FRD"/>
+        <description>This is the same as MAV_FRAME_MAV_FRD.</description>
       </entry>
       <entry value="10" name="MAV_FRAME_GLOBAL_TERRAIN_ALT">
         <description>Global (WGS84) coordinate frame with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees, second value / y: longitude in degrees, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
@@ -271,7 +271,8 @@
         <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
       </entry>
       <entry value="12" name="MAV_FRAME_BODY_FRD">
-        <description>FRD local tangent frame (x: Forward, y: Right, z: Down) with origin that travels with vehicle. The forward axis is aligned to the front of the vehicle in the horizontal plane.</description>
+        <deprecated since="2021-11" replaced_by="MAV_FRAME_MAV_FRD"/>
+        <description>Alias of MAV_FRAME_MAV_FRD (Deprecated because _BODY_ implies attitude alignment not origin).</description>
       </entry>
       <entry value="13" name="MAV_FRAME_RESERVED_13">
         <deprecated since="2019-04" replaced_by=""/>
@@ -306,6 +307,10 @@
       </entry>
       <entry value="21" name="MAV_FRAME_LOCAL_FLU">
         <description>FLU local tangent frame (x: Forward, y: Left, z: Up) with origin fixed relative to earth. The forward axis is aligned to the front of the vehicle in the horizontal plane.</description>
+      </entry>
+      <entry value="22" name="MAV_FRAME_MAV_FRD">
+        <description>FRD local tangent frame (x: Forward, y: Right, z: Down) with origin that travels with vehicle.
+          The forward axis is aligned to the front of the vehicle in the horizontal plane.</description>
       </entry>
     </enum>
     <enum name="MAVLINK_DATA_STREAM_TYPE">
@@ -6584,7 +6589,7 @@
       <extensions/>
       <field type="float" name="increment_f" units="deg">Angular width in degrees of each array element as a float. If non-zero then this value is used instead of the uint8_t increment field. Positive is clockwise direction, negative is counter-clockwise.</field>
       <field type="float" name="angle_offset" units="deg">Relative angle offset of the 0-index element in the distances array. Value of 0 corresponds to forward. Positive is clockwise direction, negative is counter-clockwise.</field>
-      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame of reference for the yaw rotation and offset of the sensor data. Defaults to MAV_FRAME_GLOBAL, which is north aligned. For body-mounted sensors use MAV_FRAME_BODY_FRD, which is vehicle front aligned.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame of reference for the yaw rotation and offset of the sensor data. Defaults to MAV_FRAME_GLOBAL, which is north aligned. For body-mounted sensors use MAV_FRAME_MAV_FRD, which is vehicle front aligned.</field>
     </message>
     <message id="331" name="ODOMETRY">
       <description>Odometry message to communicate odometry information with an external interface. Fits ROS REP 147 standard for aerial vehicles (http://www.ros.org/reps/rep-0147.html).</description>


olliw42:
great

MAV_FRAME_VEHICLE_FRD
MAV_FRAME_DRONE_FRD
MAV_FRAME_UAV_FRD
MAV_FRAME_BOARD_FRD
MAV_FRAME_AUTOPILOT_FRD
MAV_FRAME_SYSTEM_FRD
MAV_FRAME_NAV_FRD

hendjoshsr71:
Love that this is getting fixed! It a GNC person's nightmare of nomenclature abuse.

I'd add as a possibility MAV_FRAME_INS_FRD

Any more explanation on why MAV_COG_FRD is a bad idea? (Sure COG can change during a flight, but those changes are computable from a reference point.) If you put _MAV, _VEHICLE, etc a GNC person will likely assume those are center of gravity referenced?

In the text you say the origin travels with the vehicle, but don't give an idea of where that origin is with respect to a position on the vehicle? (Seems intentional to do so, but the next question is going to be where is this origin with respect to one of either the COG or INS usually.) That will then need to reconciled somehow. 

Basically looking at the lists the origin is either:
    - COG (sort of):  (MAV, VEHICLE, etc.)
    - INS: IMU, AHRS, BOARD, AUTOPILOT, etc.

hamishwillee:
Thanks @hendjoshsr71 - I like INS too, because then you don't have to discuss the actual location. Hoping to discuss in the next mav call.

korigod:
I agree that users often associate `BODY` with vehicle-aligned axes, in the same way they associate `LOCAL` with some Earth-aligned ones. It's not just the problem of `BODY` — `LOCAL_FRD` frame would cause the same confusion as `BODY_NED` does. Yet designing mavlink/rfcs#2 we decided to keep `LOCAL` and `BODY` designations to minimize the need to deprecate the existing frames (I really don't like to deprecate things if there is a possibility not to) and to keep the frame names readable  — `MAV_FRAME_BODY_NED` is something that at least looks understandable, and `MAV_FRAME_INS_FRD` looks like gibberish. On the other hand, I agree that sometimes it's better to deprecate an entity and to introduce a new one instead of changing the meaning, but the meaning of `MAV_FRAME_BODY_NED` is already aligned with RFC 2 by #1700.

Considering options other than `BODY`:

- It seems semantically more correct to me to bind the frame origin to some kind of the center of the vehicle, not the location of the IMU. The autopilot might be off-center, IMU might be external, etc., but if the drone e.g. is in the middle between two walls it would be confusing if the distances to these walls were different because the IMU is placed off-center. So I think that `IMU`, `INS`, `BOARD` options are inferior in this regard — they are correct if someone really wants to use such an origin, but I believe it's rarely the case.
- MAVLink is used not only on aerial vehicles, so I don't think it's a best idea to use `DRONE`/`UAV`/`MAV` designations.

Which leaves us with either `VEHICLE` or `COG`. Although `VEHICLE` is longer, it's definitely more understandable and also more open to different implementations — often we want the origin to be at some kind of geometrical center, not the center of gravity, which can be difficult to determine and which can change.

So if there is a consensus on the need to replace `BODY` with something else, I think `VEHICLE` is a nice option, although maybe someone would come up with something even better. But personally, I'll be just as happy if we keep existing `BODY` designation, the main thing here is the clear and consistent documentation and aligning of the PX4 implementation with it.

hamishwillee:
Thanks @korigod.

I understand that local_frd and body_frd _can_ both be misinterpreted; the difference is that people _might_ misinterpret the local case, but they _often_ misinterpret the body case. I think that is because "Body" can actually impy "aligned with the frame pose" while local just means "not global" (at least to most people).

Right now we have just one non-deprecated body frame. So if we can agree a different word than Body for that, I think we're on a path where one day we can have the bigger problem fixed.

I am not proposing to change local for the same reasons as you in the RFC - there are too many of them (and generally no one is reporting confusion).

With respect to the actual term used, I proposed COG but apparently that moves with the vehicle orientation. I see you get my point though - we want something that says "moves with the vehicle" but not aligned with the vehicle pose. DRONE is OK IMO because it doesn't have to be aerial. Any of them are better than "BODY".

You're right that the important thing is to align the flight stacks with these. Are there any non-deprecated frame definitions that you know PX4 or ArduPilot do not match?

hamishwillee:
We discussed this in the dev call last night. We ended up looking at `MAV_FRAME_BODY_FRD`, which is defined as "FRD _tangent frame_ with origin centered on the vehicle" (so like MAV_FRAME_BODY_NED but aligned with the front of the vehicle). However one member pointed out that FRD is often vehicle-pose aligned too, so changing to  `MAV_FRAME_VEHICLE_FRD` is still potentially confusing.

Further, we need both the current tangent aligned and vehicle pose aligned frames, and our convention doesn't support this. The spec talks about RPY for the frame, but that implies angles, and all the messages which use frames take vectors.

So we need to work out a convention that supports the FRD in the frame of the vehicle attitude. Any thoughts?

Proposal was that we look at this in two ways:
1. Look at existing frame conventions used in industry and move to one (or more) of those. 
2. We try and extend the current frame convention.

Any suggestions?

hamishwillee:
For "2" my thinking is that if we weren't already using BODY for an origin, we could use it to indicate a body alignment. 
So maybe use MOV to indicate moving origin and BODY to indicate a body aligned with moving origin. 
- `MAV_FRAME_MOV_FRD`
- `MAV_FRAME_BODY_FRD`

Although I think I prefer we are explicit about the frame being tangent or body aligned if it is ever ambiguous.
- `MAV_FRAME_MOV_TANGENT_FRD`
- `MAV_FRAME_MOV_BODY_FRD`
-  `MAV_FRAME_MOV_NED` (because it can't be BODY aligned by definition)

kapacheuski:
@hamishwillee I'm actually happy with MAV_FRAME_{ORIGIN_LOCATION}_{FRAME_ORIENTATION} Frames format.
By the way, I think that needs to be added additional orientation options like NAV. 

So with the origin of the body (vehicle): 

- MAV_FRAME_BODY_NED - origin from the body but orientation North East Down (body actually can be a vehicle, boat, etc - no matter ) actually now it  MAV_FRAME_LOCAL_OFFSET_NED but it is not fit in Frame Format.
- MAV_FRAME_BODY_FRD - attitude depended on frame related to Front, Right, Down of body, or vehicle etc.
- MAV_FRAME_BODY_NAV (or MAV_FRAME_BODY_NAVFRD) related to navigation frame forward, right in horizon plane, down to earth. NAV because it is simple to use for navigation of all kinds of UV.  

Same methods for naming local frames (origin with local):

- MAV_FRAME_LOCAL_NED - no questions
- MAV_FRAME_LOCAL_FRD - Origin Local and Front, Right, Down, of LOCAL object (Groundstation) - not of the UxV. Depended on the attitude of Groudstation.
- MAV_FRAME_LOCAL_NAV (NAVFRD) - Origin local and front -> heading of the ground station, front and right in a horizontal plane.

And actually, LOCAL and BODY can be replaced with Other origin points for example IMU or some POI (Point of interest) and it will fit in the frames naming conversion before.

hamishwillee:
@kapacheuski Yes, much this has been discussed before. The [RFC](https://github.com/mavlink/rfcs/pull/2/files) makes it even more clear - all the frames are are horizontally aligned! So for FRD, say, it is the forward and right components in the horizontal to earth (tangential) frames, and the D is always down.

The RFC defines one frame aligned to the vehicle attitude, which is "RPY". Not added to the standard though, because I won't do it unless someone commits to implementing.

Upshot is that this is all good and consistent, but because "BODY" means "vehicle attitude" to so many people  it is confusing. Adding the "tangential" word in there does not seem to help.

kapacheuski:
@hamishwillee Yes, but I think "BODY" should not confuse in terms of frame "origin", It just points to the center of mass of an object. 

> The [RFC](https://github.com/mavlink/rfcs/pull/2/files) makes it even more clear - all the frames are are horizontally aligned!

But how to work in case you need to send LANDING_TARGET message with the information provided by a camera on vehicle? It is attitude aligned. The question is because for now we working on ArduPilot Companion message processing and it needs to be clear to me whether am I doing all right or not. For reference, you have a look on [PrecLand: support LANDING_TARGET ext field - by ](https://github.com/ArduPilot/ardupilot/pull/21329) @chobitsfan and [AC_PrecLand: Mavlink V2 messages support](https://github.com/ArduPilot/ardupilot/pull/21556) by me.
According [RFC](https://github.com/mavlink/rfcs/pull/2/files) Maybe it is just only my problem but I am unable to realize the difference between MAV_FRAME_LOCAL_OFFSET_NED and MAV_FRAME_BODY_OFFSET_NED ? If you don't mind can I prepare some pictures with different frames and try to name them and you will comment on them? Thanks.

hamishwillee:
> @hamishwillee Yes, but I think "BODY" should not confuse in terms of frame "origin", It just points to the center of mass of an object.

It doesn't matter whether you or I think that it "should not" confuse people. The fact is that the term body was previously used to mean frame at attitude of vehicle. That remains confusing for people - i.e. it isn't a "guess" or a "supposed problem", it has resulted in incorrect implementation.

But I can live with it. The distinction is clear if you read the docs. 

> According [RFC](https://github.com/mavlink/rfcs/pull/2/files) Maybe it is just only my problem but I am unable to realize the difference between MAV_FRAME_LOCAL_OFFSET_NED and MAV_FRAME_BODY_OFFSET_NED ? 

Both `MAV_FRAME_LOCAL_OFFSET_NED` and `MAV_FRAME_BODY_OFFSET_NED` travel with the vehicle. However the frame for MAV_FRAME_LOCAL_OFFSET_NED  is "NED" - it always aligns North, while the frame for MAV_FRAME_BODY_OFFSET_NED is FRD (vehicle aligned in the horizontal plane).

It is not just you. This is confusing, partially because we have a mixture of frames that still use the old convention. I would like to actually remove some of them to reduce the mess, but that is hard with mavlink.

- [MAV_FRAME_LOCAL_OFFSET_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_LOCAL_OFFSET_NED) - NED local tangent frame (x: North, y: East, z: Down) with origin that travels with the vehicle.
  - The fact it has "offset" in the name tells you that it was an "old style" definition that might not obey the new convention. Further, the old style often didn't follow any convention. They are mixed.
  - Always read the precise definition at the link (i.e. this is NED with origin that travels with the vehicle).
  - So if we were to name this now it would be MAV_FRAME_BODY_NED
  - But we can't renamed it, because there is already a definition [MAV_FRAME_BODY_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_NED) with a different meaning.

- [MAV_FRAME_BODY_OFFSET_NED](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_OFFSET_NED) - This is the same as [MAV_FRAME_BODY_FRD](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_FRD).
  - The fact this has offset in the name and is deprecated tells you it is an old-style name :-(
  - So the definition at [MAV_FRAME_BODY_FRD](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_FRD) is "FRD local tangent frame (x: Forward, y: Right, z: Down) with origin that travels with vehicle. The forward axis is aligned to the front of the vehicle in the horizontal plane."
  - Would it help to move the definition into the right place in the message rather than linking to MAV_FRAME_BODY_FRD?

Does that help?

hamishwillee:
With respect to PRECLAND, there is also a PR for precision hold that it might be worth you looking at (for interest): https://github.com/mavlink/mavlink/pull/1858 

With respect to [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) I have no particular experience. I don't know the problems you will encounter. @dakejahl is a good person to talk to as he has experience with the PX4 implementation. @dakejahl what frames do you use?

chobitsfan:
Hi @hamishwillee Thank you for detailed explanation. 
I agree with @kapacheuski that a vehicle attitude aligned MAV_FRAME may be required for vision based navigation. Because a camera is usually fixed to vehicle.

dakejahl:
>@dakejahl what frames do you use?

I technically misuse the LANDING_TARGET message WRT to the "landing target protocol." I am sending target angle_x/angle_y/q relative to the camera. Using LOCAL_NED makes no sense for precision landing because it adds an additional requirement to the sensor system to know the origin and pose of the vehicle at all times.

In the future I do plan to make a PR to expand the usage of the LANDING_TARGET such that the data transmitted is target XYZQ with the origin being the drone itself (so maybe MAV_FRAME_BODY_FRD)

kapacheuski:
@dakejahl Hi. About LOCAL_NED - I think it has some advantages to use it as a case of target movement. For example, a landing system can have an internal filter (built-in landing system) to calculate the target LOCAL_NED position depending on the target velocity. Yes, it requires information from the vehicle but I think it does not matter to depreciate it. More of all it only frame format that is supported by PX4 stack for companion (to be honest I think it did not work on PX4 now - see my PR [Landing position updates for precision landing.](https://github.com/PX4/PX4-Autopilot/pull/19873) ).

> In the future I do plan to make a PR to expand the usage of the LANDING_TARGET such that the data transmitted is target XYZQ with the origin being the drone itself (so maybe MAV_FRAME_BODY_FRD)

Yes, it's what I'm working on too.  So if you have some PR code to test I will be happy to join our efforts!

dakejahl:
I don't have any PR code yet, and yes I also discovered precision land XY control did not working during descent. I have a fix for it as well that I have not pushed upstream. My plan is to take a shot at refactoring navigator and moving these things into flight tasks, hopefully will find some time for that in the next month or two.

hamishwillee:
I hope you guys feel the questions are resolved/you are resolving these with each other. Often I am not the right person to tell you what to do - I'm here to help you work out what the options are, and if you can't agree them, to find a consensus.

>  By the way what about @chobitsfan question? So we should use MAV_FRAME_BODY_FLU as fixed with attitude (for the camera-based sensor) and MAV_FRAME_BODY_FRD as a horizontally aligned frame. Am I right?

MAV_FRAME_BODY_FLU is removed - its ID used to be at the point MAV_FRAME_RESERVED_13 but there were no implementations.
If you need a frame that is aligned to vehicle attitude you will need to define one.
That would probably be MAV_FRAME_BODY_RPY.

tridge:
> That would probably be MAV_FRAME_BODY_RPY.

maybe  That would probably be MAV_FRAME_BODY_REALLY_MEAN_IT_THIS_TIME ?

hamishwillee:
@julianoes Can you check what PX4 does with MAV_FRAME_BODY_FRD as @tridge has indicated in https://github.com/mavlink/mavlink/pull/1737#issuecomment-1246403704 and the post immediately above.

@tridge I'm fine with this being changed if everyone does the same thing. It doesn't break the convention for the origin of the frame being referred to by "BODY" and to my mind FRD is completely fine as a vehicle aligned frame. Note, I don't "make this shit up" - I spent a long time trying to build consensus on the definitions and thought we had it.

hamishwillee:
@dakejahl Can you please help me?
 
- Tridge [in comment above](https://github.com/mavlink/mavlink/pull/1737#issuecomment-1246403704)  wants to interpret `MAV_FRAME_BODY_FRD` as vehicle-origin and vehicle attitude aligned frame.
- This makes more sense than what the docs define, which is vehicle origin frame aligned with the vehicle FR in the horizontal plane above earth, D always down.
- Tridge says he thinks PX4 already is doing the right thing - see [comment above](https://github.com/mavlink/mavlink/pull/1737#issuecomment-1246413685)

Can you confirm how PX4 handles this frame - does it FRD align with with the vehicle as the thinks? If so I will change the MAVLink definition.

I am not good with frames or code.

dakejahl:
In PX4 MAV_FRAME_BODY_FRD is body fixed frame, so yes Tridge is correct it is aligned with the vehicles attitude (which is a rotation between the body fixed frame and world frame

hamishwillee:
THanks. @tridge et al, please confirm you're OK with the specific words in https://github.com/mavlink/mavlink/pull/1894

kapacheuski:
@hamishwillee according to the discussion maybe it is possible to make as I proposed before in the [comment](https://github.com/mavlink/mavlink/pull/1737#issuecomment-1233784507)? Perhaps it has the sense to replace BODY with UXV and LOCAL to HOME (BASE) and add it as new frames. And make a decision to describe orientation, not from the type of BODY, LOCAL etc but from FRD, NED, NAV. So in total, we have all frames going according MAV_FRAME_{ORIGIN}_{ORIENTATION and coordinate system type}. So we get something like this: 

- MAV_FRAME_UXV_FRD - means that origin on UxV orientation in body frame (attitude)[m,m,m]  
- MAV_FRAME_UXV_HED - means for example Origin on UXV but orientation is Heading (from North) Elevation(from horizontal plane) Distance [rad,rad,m].(it is useful for navigation to hold course and descending angle)
- MAV_FRAME_UXV_YPD - Yaw pitch distance same as head but fixed with attitude (it is useful for onboard gimbal or antenna tracker)
- MAV_FRAME_UXV_NAV(NFRD) - means Origin on UXV , Forward Right Down in horizontal plane [m,m,m].
- MAV_FRAME_UXV_NED .... etc

FOR antenna teacher on the ground it can be like MAV_FRAME_ATS_HED, MAV_FRAME_ATS_YPD.

For the LOCAL frame same way but to avoid collision with the previous naming, it is possible to change LOCAL to BASE or ZERO. Possibly it allows adding different frames without misunderstanding. If you don't mind I can prepare PR related to these changes - and we will discuss them.

hamishwillee:
@kapacheuski The dev call think these are not the right changes - i.e. there is no appetite to add more frames unless there is a use case that actually needs them, and the BODY and LOCAL are well enough understood /accurate enough that it isn't worth changing.

I'm therefore closing this discussion. If you wish to reopen I suggest coming along to the MAV dev call sometime.

kapacheuski:
@hamishwillee Hi.

> The dev call think these are not the right changes - i.e. there is no appetite to add more frames unless there is a use case that actually needs them, and the BODY and LOCAL are well enough understood /accurate enough that it isn't worth changing.

Yes, I agree it is not time to change but it needs to be discussed. The situation with MAV_FRAME_BODY_FRD shows that it requires some changes in the future. I will connect to the next dev call if you don't mind and will try to explain my point of view.

hamishwillee:
> The situation with MAV_FRAME_BODY_FRD shows that it requires some changes in the future

Not really. We fixed that and now the definitions are all consistent. Changes to existing definitions should not be needed in future, and I'd be resistant to adding a new frame unless it was really needed.

But you are most welcome to attend the call. (in two weeks)

kapacheuski:
@hamishwillee 

> and I'd be resistant to adding a new frame unless it was really needed

You absolutely right - it is just the only way to keep all in order.

> But you are most welcome to attend the call. (in two weeks)

I will be happy to talk with the community online! (I hope no changes will be in 2 weeks).

PR #1739, opened 2021-11-04
Auto generate docs for reserved params if explicitly specified

hamishwillee:
The definition file format allows you to specify that [parameters are reserved](https://mavlink.io/en/guide/define_xml_element.html#reserved) and have a default value like this:
```xml
<param index="3" reserved="true" default="NaN" />
```

This change adds docs for the parameter like `Reserved (set to NaN)` for the above, where reserved is set to to true. 

Note, that if _no parameter_ is defined for a particular index, MAVLink defines this to mean the same as
```xml
<param index="3" reserved="true" default="0" />
```
This works properly for generated libraries but not for the docs - the parameter will simply be omitted from the docs. The reason is that the generator uses XSLT to process nodes. IF there is no node, then nothing is done.

### DIFF
--- a/doc/mavlink_to_html_table_gitbook.xsl
+++ b/doc/mavlink_to_html_table_gitbook.xsl
@@ -189,7 +189,7 @@
         <xsl:if test='@label'>: <xsl:value-of select="@label" /></xsl:if>
         </td> <!-- mission_param -->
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
Self merging. Extensively tested.

PR #1740, opened 2021-11-10
Update invalid attribute for arrays where first element sets whole array as invalid

hamishwillee:
The terminology we were using to indicate the invalid value for an array where the first element value sets that the whole array is invalid was inconsistent. The terminology for this case has changed to `[value:]` in https://github.com/mavlink/mavlink-devguide/pull/404


olliw42:
btw, to just have dropped it

for char arrays one could also use [0:] instead of [0]. 
This technically would be a change of behavior, but one could argue it would be an improvement, and in fact might most likely reflect what is actually done in implementations (which one can presume will check the length of the string which is equal to just testing the first element :))

not saying one should, but one could

hamishwillee:
Thanks. I'm not changing it now. What we have works.

hamishwillee:
Self merging. This invalid not used by anyone else yet AFAIK so this change should do no harm.

PR #1744, opened 2021-11-16
docs: catch more hyperlinks

dayjaby:
@hamishwillee Some hyperlinks are not being created if the mavlink message/command is not preceded by a space char, e.g. `(STORAGE_STATUS_READY)`. With this, we catch at least some additional cases, like parentheses and grave accents.

### DIFF
--- a/doc/mavlink_gitbook.py
+++ b/doc/mavlink_gitbook.py
@@ -127,16 +127,16 @@ def fix_add_implicit_links_items(input_html):
     #print("fix_add_implicit_link was called")
     def make_text_to_link(matchobj):
         #print("make_entry_to_link was called: %s" % matchobj.group(0))
-        item_string = matchobj.group(1)
+        item_string = matchobj.group(2)
         item_url=item_string
         if item_string == 'MAV_CMD':
             item_url='mav_commands'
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @dayjaby . Looks good - in particular catching the `(MAV_CMD_SET_CAMERA_MODE)` cases. I am not so sure about the backtick case as they really shouldn't be present at all - they don't render in HTML anyway.

Did you test rendering? FYI generated files with and without your patch attached.

[compare_doc.zip](https://github.com/mavlink/mavlink/files/7550622/compare_doc.zip)

hamishwillee:
OK, so IMO this is definitely worth going in and I am going to merge it. Thank you.

But as a post process, would welcome your comments on how we deal with case like  TERRAIN_ALT below. This is valid text for the example but it is not an actual entity. Now the backticks are invalid HTML and should be removed from source for that reason, but we might be best off modifying this so that backticks are not parsed, but are used by us to indicate something we don't want the parser to mess with in rendering.

Though perhaps better to add a disallow list of things to mark up?

image

dayjaby:
Let's just replace the backticks with `""` for the MAV_FRAME documentation?

hamishwillee:
Good idea :-). Fixed in https://github.com/mavlink/mavlink/pull/1745

PR #1747, opened 2021-11-21
16-cell support

muramura:
Devices that support up to 16 cells have been shipped.
QGC can be selected up to 16 cells.


hamishwillee:
We can't do this as it is an incompatible change to the message. I'll add a discussion in the dev call today.

amilcarlucas:
I just added support for 14 cells to mavros, and there is a mavros PR to support upto 10 batteries.
Breaking the message is not a good idea, but better now than later.

hamishwillee:
@amilcarlucas The PR that was proposed here was a compatibility break, so we will never use this approach. 

In theory we might add another extension field to the end of the message with yet more cells. The problem with that approach is that this message is sent a LOT - but in most cases the cells aren't used and can't be truncated. Causing problems for real-world users.

Last time we discussed this we agreed that we wouldn't just extend the message without looking at alternatives. I'll put a placeholder in the devcall to discuss - happy to take a proposal that is more extensible.

Issue #1678, opened 2021-08-20
Integrated Camera/Gimbal Systems Documentation 


dlwalter:
With the Gimbal V2 Protocol getting mainlined we should look at updating the documentation for vendors implementing an "all-in-one" standalone integrated camera/gimbal video system to ease integration into PX4/QGroundControl based systems.  Examples of these COTS gimbals would be Nextvision's Colibri and Nighthawk, and Trillium HD40 with more coming soon.  

The current Gimbal V2 documentation (https://mavlink.io/en/services/gimbal_v2.html) comes from the perspective of pointing a gimbal. The Camera Protocol Documentation comes from the perspective of a camera.  However, an all-in-one implementation demands communication using both, and then a few more messages from the rest of the mavlink message set.

We suggest a _section of documentation is needed to explain how to manage an integrated smart gimbal from the QGC interface_ (or other base station).  To get started, we suggest outlining the intended user actions, and then the mavlink messages that support those actions.

Simply having this mapping table of actions to messages would have helped us digest and implement these new protocols much faster into new products..

A new section could cover the specific use case of a fully integrated camera/gimbal.

**A standalone gimbaled video system has the following features:**
- Video Streaming and Recording from 1 or more cameras.
- Still Image Capture from 1 or more cameras.
- 3 Axis stabilization creating a camera view that is level to the horizon
- Accepts Pan/Tilt controls to slew the camera view around while stabilized in the Navigation frame.

It may also include:
- Retract into a safe position during landing
- Switching between visible, IR, or dual views
- Zoom using mechanical, digital, or lens changes
- Slew to a global position (given global position of UAS)
- Track a point or area on the ground
- Nudge the camera while tracking
- Stream global position of camera view center for map display along with slant range
- Stream global position of tracked point/area
- Allow stabilization turned off for specific axes

	
**If we frame the problem from the viewpoint of the user we can assume the following intentions/actions:**

- Use a joystick or other interface device to pan and tilt a stabilized gimbal to control a camera while keeping the horizon level.  When released the joystick will point at the same attitude relative to north/horizon despite aircraft movements.
- Zoom in and out a gimbaled camera or switch FOV from narrow to wide
- Start/Stop Recording Video
- Capture single/multiple still images
- Select between IR and visible video on a gimbaled camera or view IR and visible side by side
- Change the Thermal palette (LUT) if supported.
- Retract the Gimbal while landing
- Select a point on a map and have the gimbaled camera payload look at that point.
[...]

PR #1749, opened 2021-11-22
Add MAV_CMD_SELECT_TRIGG_CAM command.

MatejFranceskin:
MAV_CMD_SELECT_TRIGG_CAM as a mission item added to the mission plan selects the camera and stream that will be used for capturing images in case where the vehicle has multiple cameras and/or camera has multiple sensors available.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -119,6 +119,11 @@
         <param index="1" label="Group ID" minValue="0" maxValue="16777216" increment="1">Mission-unique group id.
           Group id is limited because only 24 bit integer can be stored in 32 bit float.</param>
       </entry>
+      <entry value="303" name="MAV_CMD_SELECT_TRIGG_CAM" hasLocation="false" isDestination="false">
+        <description>Select camera device and its sensor that will be used for capturing images.</description>
+        <param index="1" label="Camera device ID">Component ID of selected camera device (0 to address all cameras)</param>
+        <param index="2" label="Sensor ID">Selected sensor ID, related to stream_id field in VIDEO_STREAM_INFORMATION. (0 to address all sensors)</param>
+      </entry>
     </enum>
   </enums>
   <messages>


hamishwillee:
Sounds sensible but I'll have a proper look tomorrow. The things I'm interested in checking/questioning are (and you might want to clarify, but you might not since it is a stream of muddy thinking ...):
- Whether this can be used as a command as well
- Confirming whether this applies for the next call on the camera, or all subsequent calls on the camera. Does it outlast the mission. So normally you might use this as a command to set the default behaviour and as a mission item to override the default behaviour.
- Will I ever want to trigger 2 but not all cameras - ie. this can trigger all and just one, but not necessarily specific ones.
- Is there any other camera triggering that this might clash with, in which case how is that resolved.
- Does/should this select anything for video as well? 
- Is there any overriding that can happen from commands etc during mission, and is there any conflict resolution.
- Is there any other configuration that you might want to do in a mission other than triggering - let's think about that quickly now.

MatejFranceskin:
1. MAV_CMD_SELECT_TRIGG_CAM is used as command as well. When it is reached in the mission the flight controller sends the same command to all cameras. This is because also those that are not addressed have to "unselect" itself.
2. Yes, the command outlasts the mission.
3. The way it is currently designed we can trigger only one or all the cameras.
For other questions, we have to discuss it. In current implementation this command selects the camera for all commands that are related to camera triggering not just f.e. MAV_CMD_IMAGE_START_CAPTURE but also MAV_CMD_DO_SET_CAM_TRIGG_DIST and others. It could be extended also to MAV_CMD_VIDEO_START_CAPTURE which already contains information about which stream to record but if used as part of the mission plan then it doesn't address particular camera.

hamishwillee:
Thanks @MatejFranceskin ! There is a dev call this evening (my time tonight) if you want to come and discuss. Julian is on holiday and I'm not sure who will show up. Here are my thoughts anyway.

As I understand it, when the autopilot executes this command it would broadcast the message to all cameras (vehicle system ID and component id =0). Each camera on the vehicle gets the message and checks whether it has a component id that matches the message (and stream ID). If it does, then it sets it's state to either ignore or accept future broadcast camera commands [the _ignore_ is the interesting case, since presumably by default a camera will accept any broadcast command].
It will ignore further commands until it gets a new command to enable itself (should it persist the setting across reboots?)

Is that about right?

I think I prefer a design where  - ie the vehicle reverts to the user's preferences for normal flight. However as specified I'm not sure you can do that.

My first concern is what the GCS/user should do for cameras that were set disabled in the mission.

For example, in the current model when flying _outside_ of a mission the GCS allows you to switch between cameras, and commands are directed to the current camera.
- If the current camera was disabled by the mission, should it be shown as active in the UI?
- What should a disabled camera do if it gets a camera trigger command later that is targeted to its component ID (not broadcast)? Should it decide it is still disabled, or should it accept the command? 
- If it can't act on commands, then I THINK that we need to expose the enabled/disabled state of the cameras to the GCS. 

My second concern is about the next mission. Say I run a mission and it sets the camera up. I then run a new mission and it does nothing with cameras - it just users whatever is currently set. 
Generally it is much better if a setting in a mission does not persist beyond the mission. 

Lastly, I wonder if the mission should in some way display which cameras are active and which are inactive? 

Do we perhaps need to have a camera enabled/disabled state and publish that as well?

Lots of questions sorry.

auturgy:
https://github.com/mavlink/mavlink/issues/1678

hamishwillee:
THanks for closing this @MatejFranceskin 

I don't think it is a solved problem yet. If we ever need to then I think we'd lean towards the gimbal approach and just add a target component id to an empty param (like [MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE)) and get the flight stack or mavlink camera to use this if it is non-zero.
Not "perfect" but consistent.

PR #1751, opened 2021-11-24
SYS_STATUS: Extend system status flags, and add flag for recovery system

ThomasDebrunner:
`MAV_SYS_STATUS_SENSOR_*` bit field last bit to be used to indicate the health of a parachute system.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -214,6 +214,9 @@
       <entry value="1073741824" name="MAV_SYS_STATUS_SENSOR_PROPULSION">
         <description>0x40000000 propulsion (actuator, esc, motor or propellor)</description>
       </entry>
+      <entry value="2147483648" name="MAV_SYS_STATUS_RECOVERY_SYSTEM">
+        <description>0x80000000 Recovery system (parachute, balloon, retracts etc)</description>
+      </entry>
     </enum>
     <enum name="MAV_FRAME">
       <description>Co-ordinate frames used by MAVLink. Not all frames are supported by all commands, messages, or vehicles.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -214,10 +214,18 @@
       <entry value="1073741824" name="MAV_SYS_STATUS_SENSOR_PROPULSION">
         <description>0x40000000 propulsion (actuator, esc, motor or propellor)</description>
       </entry>
-      <entry value="2147483648" name="MAV_SYS_STATUS_RECOVERY_SYSTEM">
-        <description>0x80000000 Recovery system (parachute, balloon, retracts etc)</description>
+      <entry value="2147483648" name="MAV_SYS_STATUS_EXTENSION_USED">
+        <description>0x80000000 Extended bit-field are used for further sensor status bits</description>
       </entry>
     </enum>
+
+    <enum name="MAV_SYS_STATUS_SENSOR_EXTENDED" bitmask="true">
+      <description>These encode the sensors whose status is sent as part of the SYS_STATUS message in the extended fields.</description>
+      <entry value="1" name="MAV_SYS_STATUS_RECOVERY_SYSTEM">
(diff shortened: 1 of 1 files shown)


julianoes:
@hamishwillee @auturgy this change is interesting because it can't be prototyped in another dialect first. And it's tricky as it's the very last bit :smile:

message_definitions/v1.0/common.xml
```diff
@@ -214,6 +214,9 @@
       <entry value="1073741824" name="MAV_SYS_STATUS_SENSOR_PROPULSION">
         <description>0x40000000 propulsion (actuator, esc, motor or propellor)</description>
       </entry>
+      <entry value="2147483648" name="MAV_SYS_STATUS_PARACHUTE">
```
auturgy, line 217: Suggest:
<entry value="2147483648" name="MAV_SYS_STATUS_RECOVERY_SYSTEM">
<description>0x80000000 Recovery system (parachute, balloon, retracts etc)</description>

auturgy: Excuse dodgy markup: unfortunately github parses it in comments so I had to remove a couple of tags

auturgy (changes requested):
I think it's as good a use as any for the last bit: it does force us to think about what we do next though.  See suggestion to generalise from parachute to recovery system.

peterbarker:
What are the suggested semantics of the relevant fields, present/enabled/healthy?

message_definitions/v1.0/common.xml
ThomasDebrunner: Agree, that's a good option

ThomasDebrunner: Changed the PR

ThomasDebrunner:
> What are the suggested semantics of the relevant fields, present/enabled/healthy?

Present: System is fitted to the vehicle and detected
Enabled: System is set to trigger in an emergency
Healthy: There are no issues detected with the system

Is that what you mean?

hamishwillee (commented):
>> And it's tricky as it's the very last bit 😄

> I think it's as good a use as any for the last bit: it does force us to think about what we do next though. See suggestion to generalise from parachute to recovery system.

@auturgy I think it's best for us to think about what we do next before we merge this.

Specifically, if we decided that we would extend [SYS_STATUS](https://mavlink.io/en/messages/common.html#SYS_STATUS) with a second triplet of fields then we would want to use this last flag to indicate that the following flag set is in use, and the first flag of the new set for the recovery system.

I'm not advocating, but I do think it is worth asking the question before we remove the choice. What are the options?

(otherwise yes, this looks like a good use of the last flag).

ThomasDebrunner:
@hamishwillee I like the idea of adding another triplet for status bits and using the last bit to indicate the presence of the second triplet. 
I pushed another commit for that. They can be squashed on merge

julianoes (commented):
I think this looks correct.

tridge:
no objections from me, seems like the best option

hamishwillee (commented):
@ThomasDebrunner Thanks very much. Looks like we have broad consensus on this. Can you run format on the file (it fails the test). On ubuntu you can do this from the mavlink repo folder using the command:
```
./scripts/format_xml.sh ./message_definitions/v1.0/common.xml
```

message_definitions/v1.0/common.xml
```diff
@@ -214,7 +214,18 @@
       <entry value="1073741824" name="MAV_SYS_STATUS_SENSOR_PROPULSION">
         <description>0x40000000 propulsion (actuator, esc, motor or propellor)</description>
       </entry>
+      <entry value="2147483648" name="MAV_SYS_STATUS_EXTENSION_USED">
```
hamishwillee, line 217: To avoid any ambiguity on which field(s) should this be set? I.e. do we require it set on all the `onboard_control_sensors_xxx` fields, or just (say) `onboard_control_sensors_present` (I propose just onboard_control_sensors_present)

julianoes: Good point. `onboard_control_sensors_present` sounds sensible.

PR #1761, opened 2021-12-29
Move "external" dialects to be normal dialects

peterbarker:
Easier to see what's going on.  No good reason to separate these AFAICS, apart from making life difficult for people.  Will make it easier to re-use messages from other places.

The big one is "simplifies tooling", which the "external" directory just makes awful.  There are four open PRs against pymavlink to fix issues related to external dialects; this PR just makes those problems disappear.

@Thomas-Li-AVSS  @olliw42 

... as an aside, having a relevant github username we can `@` in the github UI required in the headers of these xml files would be nice....


hamishwillee:
@auturgy @julianoes @LorenzMeier I have no particular problem with this other than it differs from our original discussion/agreement on what external means. Let's discuss in our next meeting.

hamishwillee:
@peterbarker Personally I think the limitation of having to only build in one folder is rubbish. But we chatted in the dev call and we generally agreed that if we can't have that we'd prefer this files moved into the normal location than we loose the ability to build all.xml with the external files (i.e. it is more valuable for us to help avoid clashes).

In summary, I aim to merge next week, provided I do not hear strong objections from @LorenzMeier in the meantime.

dayjaby:
Seriously, this is nothing but a pymavlink issue. I provided a very simple solution for that issue (https://github.com/ArduPilot/pymavlink/pull/601), at least for the python generator. You can use mavlink/mavlink as it is with this, or even get some deeply nested external dialects to work: https://github.com/mavlink/mavlink/pull/1779.

> There are four open PRs against pymavlink to fix issues related to external dialects; this PR just makes those problems disappear.

That's a weak statement. PRs are not problems, but problem solvers. Maybe put in some effort to at least try out solutions that others have poured many hours into, especially if the repository is not python setup.py-installable for over 6 months.

auturgy:
David, I hear your point but you're missing some context. The external dialects were introduced/accepted due to an intended change in mavlink governance. That governance change hasn't eventuated, so the justification for the external dialects doesn't exist.

hamishwillee:
Historically we have expected official dialects to be in a particular place and only include other official dialects. External dialects are expected to be out of tree and be able to include the official dialects from the official location.

Where we broke things was adding *all.xml* as an official dialect that includes external dialects - outside expectation. So we can update the toolchain to allow this behivour so it can build xml, move all.xml to be external, move the external dialects into the same dialect folder as the others (as per this PR), or decide not to build all.xml. 

As we want to build all.xml and there aren't very strong reasons to keep the external dialects separate, the pragmatic solution is to merge this PR.

hamishwillee:
@olliw42 I just realised that you will be impacted by this. I plan to merge this tomorrow. It will move the external dialects into the same folder as common.xml etc.

olliw42:
@hamishwillee, and others

yes, I am impacted too, many thx for noticing this, very kind :) As the culprit of the external dialects idea I may make some comments:

from a general perspective I'm (too ?) surprised that such a thing like different folders could cause such unsurmountable issues, it somewhat comes across like a surprising show up of incapability. However, given that:

I very much like the idea to provide github @ names, very good idea indeed

I'd like to recall that then the external dialects were installed we agreed that it might be evaluated at a later time, with the implication that if you mavlink guys feel it didn't pay out that it might be removed. I'm not sure where you are on this, but technically if you feel it is rather a failure, then removing would also be a "solution".

I think one of the biggest advantage of the external dialects idea for the respective dialect owners is that it helps avoid ID clashes. I'm glad to see that this aspect has been given quite some significance in the above. However, this could be - technically - achieved by other means too. E.g., there is no need to compile them in all.xml. There is not even a need to have the dialect xml's stored in the mavlink repo. All it minimally would need would be a file somewhere there the covered ID ranges would be listed, with some pointers to the dialects. Could be a xml with only comments in it. This also would be a "solution".

Personally I think that there is an advantage of having the external dialects in a separate location, as it just makes it clear that this is not supported part of mavlink. But that's of course just a matter of taste.

I'm not convinced that just moving the external dialects over will be a proper pragmatic solution. As much as I know - pl correct me if I'm wrong - the tool chain still cannot handle arbitrary includes, even if the xml's are in one folder, but one needs to order them specifically (buzz word: development.xml). External dialects, since they likely tend to have includes, may easily make this harder and possibly impossible. I think a related PR, which too is just a half-solution, is open since a while now. That is: As much as I can see you are not going to get around having to get your toolchain working properly concerning includes.

So, the current proposal looks to me like one of these "quick pragmatic solutions" which just is going to cause more ongoing issues later on. I think that instead of wasting your time on debating and stepping forward and backwards one also just could use the time to just get it right.

Besides these comments, I'd like to say that I will be ok with whatever you will do, it will work for me.

hamishwillee:
Thanks @olliw42. Yes, this is a "pragmatic" fix. The ideal fix would be to support arbitrary include locations. Even though the work has been done to propose a fix, getting this reviewed and included is significantly more time consuming than just moving the dialects.

With unlimited resource the decision might have been different. I hope this doesn't bite us later, but if it does we'll deal with that then. In the meantime I understand from @auturgy that Peter is going to look at the problematic build order thing - perhaps adapting the PR I put in to fix this.

peterbarker:
@olliw42 https://github.com/ArduPilot/mavlink/pull/237 - can't tag you over there....

PR #1763, opened 2021-12-30
MAV_CMD_NAV_PUNCTUAL_WAYPOINT - Adding support for punctual waypoint commands

mustafa-gokce:
Hi.

This is my PR to support punctual waypoint commands to MAVLink enabled vehicles.
The reason behind that is I was working on the ArduPilot codebase to support punctual navigation to a certain waypoint but there were no related MAVLink commands.

This command is something like a mix-up of MAV_CMD_NAV_WAYPOINT and MAV_CMD_NAV_DELAY but instead of delaying to go to a certain waypoint, I would like to navigate to a waypoint at a certain time.

For example, if the first field (delta time) is set to an integer value > 0, the vehicle will try to be at the specified location (param 5, 6, and 7) exactly after delta time seconds neither earlier nor later.

If the first field (delta time) is set to -1, then UTC fields (param 2, 3, and 4) will be enabled and the vehicle will try to be at the specified location (param 5, 6, and 7) exactly at the specified time neither earlier nor later (Hour:Minute:Second).

If the first field (delta time) is set to 0, vehicle will go to this location like MAV_CMD_NAV_WAYPOINT.

Firstly I will implement this in GUIDED mode but later I will add support to this command in AUTO mode (of ArduPilot).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2422,6 +2422,16 @@
         <param index="7">User defined</param>
       </entry>
       <!-- END of user range (31000 to 31999) -->
+      <entry value="26009" name="MAV_CMD_NAV_PUNCTUAL_WAYPOINT" hasLocation="true" isDestination="true">
+        <description>Navigate to a location at certain time.</description>
+        <param index="1" label="Delta Time" units="s" minValue="-1" increment="1">Delta time in seconds (-1 to enable time-of-day fields)</param>
+        <param index="2" label="Hour" minValue="-1" maxValue="23" increment="1">hour (24h format, UTC, -1 to ignore)</param>
+        <param index="3" label="Minute" minValue="-1" maxValue="59" increment="1">minute (24h format, UTC, -1 to ignore)</param>
+        <param index="4" label="Second" minValue="-1" maxValue="59" increment="1">second (24h format, UTC, -1 to ignore)</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
+      </entry>
     </enum>
     <enum name="MAV_DATA_STREAM">
       <deprecated since="2015-06" replaced_by="MESSAGE_INTERVAL"/>


hamishwillee:
The behaviour of this mission item is not well enough defined to make an assessment. Specifically how it works with relation to other mission items. Consider this mission.
```
TAKEOFF
MISSION_WAYPOINT
PUNCTUAL_WAYPOINT 3pm
MISSION_WAYPOINT
PUNCTUAL_WAYPOINT 4pm
LAND
```
At 2pm you're landed and you go to auto mode. The mission takes off and goes to the first mission item. It gets there at 2.15pm and is 20 metres from the first punctual waypoint. What should it then do? Should it fly the 20 meters in 45 minutes. Should it wait  until nearly the time and fly at full speed. If there is a speed waypoint setting in injected before the waypoint should that be respected? 

I dislike this because it forces the autopilot to evaluate the whole mission in the context of future waypoints, rather than the current setup where the evaluation is done at the current waypoint. Further, the mission feasibility changes very radically depending on when you run the mission.

A better way to do this, if it meets your use-case, is to define this in terms of when the vehicle will leave the waypoint. So MAV_CMD_CONDITION_WAIT_UNTIL that waits at a waypoint until a specific time. You would apply this at the waypoint and it can get there whenever it likes, but will not leave until the time condition is met. If it gets there "late" it just continues on. 
This allows you to plan a route with approximate times, but the condition is evaluated only when the mission gets to it.

hamishwillee:
@mustafa-gokce Thanks. You're right, what I was proposing already exists and is [MAV_CMD_NAV_DELAY](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_DELAY). 

As above, this sounds like it makes the autopilot's job of working out how to fly the vehicle a whole lot more complicated - because it has to know about future waypoints before it gets to them. The stack would have to know which commands it should ignore or respect, including ones that might come along in future - i.e. you would have to keep on evaluating how this command works every time something new is added.

If I wanted to plan a mission in this way I would do it on the GCS or a companion computer, dynamically constructing the route at runtime.

@auturgy @julianoes Is this something you think ardupilot and PX4 would want?

julianoes:
My hunch is that we would do this using MAV_CMD_NAV_DELAY. What would be the use case?

hamishwillee:
I think this will be very hard to define and get right in the flight stack, and better implemented in a GCS or companion using existing commands at mission time.

I'm therefore going to close. If you can get clear use cases and support from a flight stack to prototype in their flight stack we would be happy to reconsider. 

@auturgy @julianoes FYI.

mustafa-gokce:
@hamishwillee OK, I understand. I am working on the implementation with MAV_CMD_WAYPOINT_USER_1.

PR #1766, opened 2022-01-03
Extend WINCH_STATUS_FLAG and WINCH_ACTION

dakejahl:
- Added additional flags to `status` of the **WINCH_STATUS** message. 
- Added additional actions to the **WINCH_ACTIONS** enum for the **MAV_CMD_DO_WINCH** command.
- Renamed **MAV_WINCH_STATUS_*** to **WINCH_STATUS_*** and actions **WINCH_*** to **WINCH_ACTION_***

This PR is specifically to enable support for a smart winch which implements the new winch status flags and actions. I worked with @julianoes on this solution as to avoid using custom messages/commands in our product. The only quirk with this solution is that states are being encoded into the field meant for status flags, which aren't quite the same thing (states are mutually exclusive while flags are not) -- but I feel this is only a minor flaw.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -594,7 +594,7 @@
         <description>Perform the locking sequence to relieve motor while in the fully retracted position. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="4" name="WINCH_DELIVER">
-        <description>Sequence of drop, slow down, touch down, reel up, lock. Only Action and Instance command parameters are used, others are ignored.</description>
+        <description>Sequence of drop, slow down, touch down, reel up, lock. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="5" name="WINCH_HOLD">
         <description>Engage motor and hold current position. Only Action and Instance command parameters are used, others are ignored.</description>
@@ -602,13 +602,10 @@
       <entry value="6" name="WINCH_RETRACT">
         <description>Return the reel to the fully retracted position. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
-      <entry value="7" name="WINCH_DISENGAGE">
-        <description>Disengage clutch. Motor will become inactive. Only Action and Instance command parameters are used, others are ignored.</description>
-      </entry>
-      <entry value="8" name="WINCH_LOAD_LINE">
+      <entry value="7" name="WINCH_LOAD_LINE">
         <description>Load the reel with line. The winch will calculate the total loaded length and stop when the tension exceeds a threshold. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
-      <entry value="9" name="WINCH_ABANDON_LINE">
+      <entry value="8" name="WINCH_ABANDON_LINE">
         <description>Spool out the entire length of the line. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -581,14 +581,32 @@
     <!-- winch action enum -->
     <enum name="WINCH_ACTIONS">
       <description>Winch actions.</description>
-      <entry value="0" name="WINCH_RELAXED">
-        <description>Relax winch.</description>
+      <entry value="0" name="WINCH_ACTION_FREEWHEEL">
+        <description>Allow motor to freewheel.</description>
       </entry>
-      <entry value="1" name="WINCH_RELATIVE_LENGTH_CONTROL">
+      <entry value="1" name="WINCH_ACTION_RELATIVE_LENGTH_CONTROL">
         <description>Wind or unwind specified length of cable, optionally using specified rate.</description>
       </entry>
-      <entry value="2" name="WINCH_RATE_CONTROL">
+      <entry value="2" name="WINCH_ACTION_RATE_CONTROL">
         <description>Wind or unwind cable at specified rate.</description>
+      </entry>
+      <entry value="3" name="WINCH_ACTION_LOCK">
+        <description>Perform the locking sequence.</description>
+      </entry>
+      <entry value="4" name="WINCH_ACTION_DELIVER">
+        <description>Sequence of drop, slow down, touch down, reel up, lock.</description>
+      </entry>
+      <entry value="5" name="WINCH_ACTION_HOLD">
+        <description>Engage motor and hold current position.</description>
+      </entry>
+      <entry value="6" name="WINCH_ACTION_RETURN">
+        <description>Return the reel to the home position.</description>
+      </entry>
+      <entry value="7" name="WINCH_ACTION_DISENGAGE">
+        <description>Disengage clutch. Motor will become inactive.</description>
+      </entry>
+      <entry value="8" name="WINCH_ACTION_LOAD">
+        <description>Load the reel with line.</description>
       </entry>
     </enum>
     <!-- UAVCAN node health enumeration -->
@@ -4471,19 +4489,34 @@
         <description>Land in multicopter mode on reaching the landing co-ordinates (the whole landing is by "hover descent").</description>
       </entry>
     </enum>
-    <enum name="MAV_WINCH_STATUS_FLAG" bitmask="true">
+    <enum name="WINCH_STATUS_FLAG" bitmask="true">
       <description>Winch status flags used in WINCH_STATUS</description>
-      <entry value="1" name="MAV_WINCH_STATUS_HEALTHY">
+      <entry value="1" name="WINCH_STATUS_HEALTHY">
         <description>Winch is healthy</description>
       </entry>
-      <entry value="2" name="MAV_WINCH_STATUS_FULLY_RETRACTED">
+      <entry value="2" name="WINCH_STATUS_FULLY_RETRACTED">
         <description>Winch thread is fully retracted</description>
       </entry>
-      <entry value="4" name="MAV_WINCH_STATUS_MOVING">
+      <entry value="4" name="WINCH_STATUS_MOVING">
         <description>Winch motor is moving</description>
       </entry>
-      <entry value="8" name="MAV_WINCH_STATUS_CLUTCH_ENGAGED">
-        <description>Winch clutch is engaged allowing motor to move freely</description>
+      <entry value="8" name="WINCH_STATUS_CLUTCH_ENGAGED">
+        <description>Winch clutch is engaged. Motor is active.</description>
+      </entry>
+      <entry value="16" name="WINCH_STATUS_LOCKED">
+        <description>Winch is locked by locking mechanism.</description>
+      </entry>
+      <entry value="32" name="WINCH_STATUS_DROPPING">
+        <description>Winch is dropping payload.</description>
+      </entry>
+      <entry value="64" name="WINCH_STATUS_ARRESTING">
+        <description>Winch is arresting payload descent.</description>
+      </entry>
+      <entry value="128" name="WINCH_STATUS_GROUND_SENSE">
+        <description>Winch is using torque measurements to sense the ground.</description>
+      </entry>
+      <entry value="256" name="WINCH_STATUS_RETURNING_TO_HOME">
+        <description>Winch is returning to the home position.</description>
       </entry>
     </enum>
     <enum name="MAG_CAL_STATUS">
@@ -7041,7 +7074,7 @@
       <field type="float" name="voltage" units="V" invalid="NaN">Voltage of the battery supplying the winch. NaN if unknown</field>
       <field type="float" name="current" units="A" invalid="NaN">Current draw from the winch. NaN if unknown</field>
       <field type="int16_t" name="temperature" units="degC" invalid="INT16_MAX">Temperature of the motor. INT16_MAX if unknown</field>
-      <field type="uint32_t" name="status" display="bitmask" enum="MAV_WINCH_STATUS_FLAG">Status flags</field>
+      <field type="uint32_t" name="status" display="bitmask" enum="WINCH_STATUS_FLAG">Status flags</field>
     </message>
     <message id="12900" name="OPEN_DRONE_ID_BASIC_ID">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -591,25 +591,25 @@
         <description>Wind or unwind line at specified rate.</description>
       </entry>
       <entry value="3" name="WINCH_LOCK">
-        <description>Perform the locking sequence to relieve motor while in the fully retracted position. All other command parameters are ignored.</description>
+        <description>Perform the locking sequence to relieve motor while in the fully retracted position. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="4" name="WINCH_DELIVER">
-        <description>Sequence of drop, slow down, touch down, reel up, lock. All other command parameters are ignored.</description>
+        <description>Sequence of drop, slow down, touch down, reel up, lock. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="5" name="WINCH_HOLD">
-        <description>Engage motor and hold current position. All other command parameters are ignored.</description>
+        <description>Engage motor and hold current position. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="6" name="WINCH_RETRACT">
-        <description>Return the reel to the fully retracted position. All other command parameters are ignored.</description>
+        <description>Return the reel to the fully retracted position. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="7" name="WINCH_DISENGAGE">
-        <description>Disengage clutch. Motor will become inactive. All other command parameters are ignored.</description>
+        <description>Disengage clutch. Motor will become inactive. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="8" name="WINCH_LOAD_LINE">
-        <description>Load the reel with line. The winch will calculate the total loaded length and stop when the tension exceeds a threshold. All other command parameters are ignored.</description>
+        <description>Load the reel with line. The winch will calculate the total loaded length and stop when the tension exceeds a threshold. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="9" name="WINCH_ABANDON_LINE">
-        <description>Spool out the entire length of the line. All other command parameters are ignored.</description>
+        <description>Spool out the entire length of the line. Only Action and Instance command parameters are used, others are ignored.</description>
       </entry>
     </enum>
     <!-- UAVCAN node health enumeration -->
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -591,22 +591,22 @@
         <description>Wind or unwind line at specified rate.</description>
       </entry>
       <entry value="3" name="WINCH_LOCK">
-        <description>Perform the locking sequence to relieve motor while in the fully retracted position. Only Action and Instance command parameters are used, others are ignored.</description>
+        <description>Perform the locking sequence to relieve motor while in the fully retracted position. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="4" name="WINCH_DELIVER">
         <description>Sequence of drop, slow down, touch down, reel up, lock. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="5" name="WINCH_HOLD">
-        <description>Engage motor and hold current position. Only Action and Instance command parameters are used, others are ignored.</description>
+        <description>Engage motor and hold current position. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="6" name="WINCH_RETRACT">
-        <description>Return the reel to the fully retracted position. Only Action and Instance command parameters are used, others are ignored.</description>
+        <description>Return the reel to the fully retracted position. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="7" name="WINCH_LOAD_LINE">
-        <description>Load the reel with line. The winch will calculate the total loaded length and stop when the tension exceeds a threshold. Only Action and Instance command parameters are used, others are ignored.</description>
+        <description>Load the reel with line. The winch will calculate the total loaded length and stop when the tension exceeds a threshold. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
       <entry value="8" name="WINCH_ABANDON_LINE">
-        <description>Spool out the entire length of the line. Only Action and Instance command parameters are used, others are ignored.</description>
+        <description>Spool out the entire length of the line. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
     </enum>
     <!-- UAVCAN node health enumeration -->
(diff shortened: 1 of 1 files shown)


auturgy:
The re-naming will break existing implementations. I agree that it is cleaner, but the juice isn't worth the squeeze imho.

dakejahl:
Good point! I can change it back.

message_definitions/v1.0/common.xml
```diff
@@ -581,15 +581,33 @@
     <!-- winch action enum -->
     <enum name="WINCH_ACTIONS">
       <description>Winch actions.</description>
-      <entry value="0" name="WINCH_RELAXED">
-        <description>Relax winch.</description>
+      <entry value="0" name="WINCH_ACTION_FREEWHEEL">
+        <description>Allow motor to freewheel.</description>
```
hamishwillee, line 585: This change might be OK, if freewheel and relax mean the same thing? i.e. you can change descriptions that are semantically the same, but not enum names.

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
     <enum name="WINCH_ACTIONS">
       <description>Winch actions.</description>
       <entry value="0" name="WINCH_RELAXED">
-        <description>Relax winch.</description>
+        <description>Allow motor to freewheel.</description>
```
hamishwillee, line 585: @rmackay9 This PR adds some actions and status flags to the WINCH commands that you moved into common in #1432.

The changes are all non-breaking (additions). Would welcome adding additional feedback or comments from your experience with the original message.

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
...
       </entry>
       <entry value="1" name="WINCH_RELATIVE_LENGTH_CONTROL">
         <description>Wind or unwind specified length of cable, optionally using specified rate.</description>
       </entry>
       <entry value="2" name="WINCH_RATE_CONTROL">
         <description>Wind or unwind cable at specified rate.</description>
       </entry>
+      <entry value="3" name="WINCH_LOCK">
+        <description>Perform the locking sequence.</description>
```
hamishwillee, line 594: What is a locking sequence?

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
...
+      </entry>
+      <entry value="4" name="WINCH_DELIVER">
+        <description>Sequence of drop, slow down, touch down, reel up, lock.</description>
```
hamishwillee, line 597: I presume the winch should respect the length and rate for the drop and reel up phases (?). Does it also respect the length or is the expectation that it will have its own sense of touching down and will use whatever length is required for that? OR is the length in this case a maximum length - you didn't touch down, so you abort?

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
...
+      </entry>
+      <entry value="5" name="WINCH_HOLD">
+        <description>Engage motor and hold current position.</description>
+      </entry>
+      <entry value="6" name="WINCH_RETURN">
+        <description>Return the reel to the home position.</description>
```
hamishwillee, line 603: What is the home position? Does this mean fully retracted?

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
...
+      </entry>
+      <entry value="7" name="WINCH_DISENGAGE">
+        <description>Disengage clutch. Motor will become inactive.</description>
```
hamishwillee, line 606: So is this different from freewheeling? I.e. how/when would this be used?

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
...
+      </entry>
+      <entry value="8" name="WINCH_LOAD">
+        <description>Load the reel with line.</description>
```
hamishwillee, line 609: So this is like a maintenance mode? What is the expectation on the winch when this is called?

message_definitions/v1.0/common.xml
```diff
@@ -4483,7 +4501,22 @@
...
+        <description>Winch is dropping payload.</description>
+      </entry>
+      <entry value="64" name="MAV_WINCH_STATUS_ARRESTING">
+        <description>Winch is arresting payload descent.</description>
+      </entry>
+      <entry value="128" name="MAV_WINCH_STATUS_GROUND_SENSE">
+        <description>Winch is using torque measurements to sense the ground.</description>
+      </entry>
+      <entry value="256" name="MAV_WINCH_STATUS_RETURNING_TO_HOME">
+        <description>Winch is returning to the home position.</description>
```
hamishwillee, line 4519: Do we need a status for when we are doing WINCH_LOAD?

message_definitions/v1.0/common.xml
dakejahl: You don't want to hold the weight of a package using just the motor stall current -- therefore you need some form of locking mechanism to hold the weight of the package and allow the motor to go inactive. The locking sequence would perform this action.

hamishwillee:
Assuming these commands are intended for a GCS how would it know whether the winch is supported on a particular vehicle, and whether this additional set of actions is supported? 

@julianoes Could/should could use yet another protocol flag or two?

message_definitions/v1.0/common.xml
dakejahl: For our product, none of these added commands respect the length or rate values. For example, in our product **Deliver** is an entirely self contained procedure. We have a settings interface enabled by the Extended Parameter Protocol that allows configuring things like DROP_ARREST_PCT, SENSE_SPEED, and RETURN_SPEED. Same thing for reel up/down we have REEL_UP_SPEED and REEL_DOWN_SPEED. These parameters are backed up to a file system so that they persist through reboots -- the idea is that this is a "smart winch" and most of its actions are self contained and any changes to behavior is done via parameters.

message_definitions/v1.0/common.xml
dakejahl: Yes the home position is the fully retracted position which allows the locking mechanism to be engaged.

message_definitions/v1.0/common.xml
dakejahl: Yes this is freewheeling. This would be used in emergency situations. I suppose I could renamed this to **WINCH_ABANDON_LINE**. Right now I am using it in development for allowing the operator to ditch a package if for example it gets snagged in a tree.

message_definitions/v1.0/common.xml
dakejahl: I actually have not written the code for this yet. The idea is that the winch zeroes its estimated spooled length and will begin reeling in at a fixed speed and calculate the length spooled. It will stop when torque exceeds LOCKING_TORQUE. It uses this maximum length to determine if the deliver distance is out of range as well as adjust the cm/rotation for estimating the spooled out length during DROP/ARREST/RETURN (spool diameter changes with length loaded)

dakejahl: This would be used by an operator who needs to load fresh line

message_definitions/v1.0/common.xml
dakejahl: Good catch! Yes I need to add a state for this.

dakejahl:
Initially I thought that we might define a "smart winch" protocol. Especially since we use the Extended Parameter Protocol. We'd be able to better define all of this stuff rather than make due with the existing messages.

message_definitions/v1.0/common.xml
rmackay9: yes, it makes sense to be able to tell the winch to lock.  The [Daiwa Winch](https://ardupilot.org/copter/docs/common-daiwa-winch.html) doesn't directly support a "lock" state but it does have a "clutch" that allows controlling the tension so to lock it you just set the tension to "engaged hard" and set the rate to zero and it will hold.

message_definitions/v1.0/common.xml
```diff
@@ -582,14 +582,32 @@
...
         <description>Wind or unwind cable at specified rate.</description>
       </entry>
+      <entry value="3" name="WINCH_LOCK">
+        <description>Perform the locking sequence.</description>
+      </entry>
+      <entry value="4" name="WINCH_DELIVER">
+        <description>Sequence of drop, slow down, touch down, reel up, lock.</description>
+      </entry>
+      <entry value="5" name="WINCH_HOLD">
+        <description>Engage motor and hold current position.</description>
```
rmackay9, line 600: I wonder if WINCH_HOLD is perhaps the same as the existing length control or rate control.

message_definitions/v1.0/common.xml
rmackay9: Maybe we should rename "RETURN" to "RETRACT"

message_definitions/v1.0/common.xml
rmackay9: I'm also not clear on what the difference is between "DISENGAGE" and the existing "RELAXED"

message_definitions/v1.0/common.xml
rmackay9: I think we should rename "LOAD" to something more specific.  "LOAD" can be either a verb or a noun so on it's own it's not clear what it's for.  maybe "CALIBRATE" would be better or perhaps we should skip this until the code has been written.

auturgy:
Bit off topic but use of param_ext should be discouraged. Unless there's no way to use the normal param protocol I'd steer well clear of it.

message_definitions/v1.0/common.xml
dakejahl: Semantics. We could use the term relaxed however it's not clear to me if the motor is on or off in the relaxed state. Relaxed could just mean zero tension (zero torque) which isn't the same as freewheeling. Originally I had this named WINCH_FREEWHEEL.

message_definitions/v1.0/common.xml
dakejahl: Yes it could be. Our winch doesn't actually use any of the fields other than `action` in **MAV_CMD_DO_WINCH** since all of our actions are self contained and all variables are controlled via parameters (extended parameter protocol)

message_definitions/v1.0/common.xml
rmackay9: Do we actually need these flags? Would any autopilot, ground station or user actually use these flags somehow?

message_definitions/v1.0/common.xml
```diff
@@ -4483,7 +4501,22 @@
         <description>Winch motor is moving</description>
       </entry>
       <entry value="8" name="MAV_WINCH_STATUS_CLUTCH_ENGAGED">
-        <description>Winch clutch is engaged allowing motor to move freely</description>
+        <description>Winch clutch is engaged. Motor is active.</description>
```
rmackay9, line 4504: I'd prefer if we left the description for MAV_WINCH_STATUS_CLUTCH_ENGAGED as it was..

dakejahl:
QGC uses it for cameras and I was under the impression this was what all components other than flight controllers use but now that I am reading the docs closer I realize this is not the case! Hahaha I can change it to the normal parameter protocol, I guess it doesn't really matter at the moment. Thanks for the insight.

rmackay9:
It's great that another smart winch is being developed.  Will this be readily available for purchase?  I'm interested in writing a driver so ArduPilot can support the winch if I can get my hands on one somehow.  Here in Japan everyone uses the Daiwa winch but they're very expensive and difficult to purchase outside of Japan.

message_definitions/v1.0/common.xml
dakejahl: I realize I left out two additional states that I have planned (but have not yet implemented) and that is:

1. Redeliver -- package doesn't release, winch detects this from torque measurements during the RETURN state
2. Abandon line -- after X redeliver attempts the package/line would be abandoned

But to answer your question -- no, other than situational awareness. This is kind of what telemetry is all about though, so an operator has some situational awareness about what their machine is up to if/when things don't look right.

dakejahl: This screenshot is a little outdated, but this is a control panel I added to QGC
image

message_definitions/v1.0/common.xml
dakejahl: The "move freely" part was what confused me. "clutch engaged" tells me the motor is active and if I stick my hand into it I might lose a finger whereas "move freely" makes me think the motor is inactive and I could move it freely with my hand

dakejahl:
>Will this be readily available for purchase? I'm interested in writing a driver so ArduPilot can support the winch if I can get my hands on one somehow.

Yes however it will only work with our proprietary solution (PRISM Sky). My driver runs on our companion computer and interfaces to our hardware directly. Eventually we could easily move this to something like a raspberry pi with an ethernet or serial interface -- it would still be mavlink interface.

message_definitions/v1.0/common.xml
rmackay9: The flags were not really intended to provide detailed visibility of all the winch's internal state so I don't really think all these are necessary.

We should avoid synonyms.. e.g. why name the action "RETURN" and the flag "RETURN_TO_HOME"?

message_definitions/v1.0/common.xml
hamishwillee: @dakejahl That's fine, but we should make it clear that this is the intent. Maybe something similar to below for the smart actions where it is not obvious the parameters will be ignored.

```suggestion
        <description>Sequence of drop, slow down, touch down, reel up, lock.
          The action is self contained (winch command parameters for line length, rate, etc.  are ignored).</description>
```

message_definitions/v1.0/common.xml
hamishwillee: Sounds good. We'll also need corresponding change in the status.

message_definitions/v1.0/common.xml
dakejahl: I agree, I'll try and consolidate all of the terminology

message_definitions/v1.0/common.xml
hamishwillee: The difference should be precisely captured in the description. @rmackay9 what does the winch you are using mean by "RELAXED"?

message_definitions/v1.0/common.xml
hamishwillee: I'm with @rmackay9 on this one. When this goes in it should explain in some detail the preconditions and detail.

message_definitions/v1.0/common.xml
```diff
@@ -4483,7 +4501,22 @@
...
+      <entry value="32" name="MAV_WINCH_STATUS_DROPPING">
+        <description>Winch is dropping payload.</description>
+      </entry>
+      <entry value="64" name="MAV_WINCH_STATUS_ARRESTING">
+        <description>Winch is arresting payload descent.</description>
+      </entry>
+      <entry value="128" name="MAV_WINCH_STATUS_GROUND_SENSE">
+        <description>Winch is using torque measurements to sense the ground.</description>
+      </entry>
+      <entry value="256" name="MAV_WINCH_STATUS_RETURNING_TO_HOME">
```
hamishwillee, line 4518: Rename to retract or whatever

message_definitions/v1.0/common.xml
hamishwillee: @rmackay9 I think I disagree about needing some, or possibly all of these. Generally it is useful to know what something is doing, and when you have sent a ommand, to know that it has completed. In PX4 we could know the completion status via a long running command, but without this will be hard for a great GCS experience on ArduPilot. 

isn't it useful for a user to know when the winch is locked? Ditto if the winch is spooling in, out, calibrating. etc. Not so sure about ground sensing, but certainly "releasing" might be useful.

hamishwillee:
> Bit off topic but use of param_ext should be discouraged. Unless there's no way to use the normal param protocol I'd steer well clear of it.

I agree with this. The extended protocol was created to support the camera protocol by allowing arbitrary data (most commonly strings) to be used. However it is not supported by ArduPilot and unless we actually need to send arbitrary data we should not design an API that requires it.

message_definitions/v1.0/common.xml
```diff
@@ -582,13 +582,34 @@
...
       <entry value="1" name="WINCH_RELATIVE_LENGTH_CONTROL">
-        <description>Wind or unwind specified length of cable, optionally using specified rate.</description>
+        <description>Wind or unwind specified length of line, optionally using specified rate.</description>
       </entry>
       <entry value="2" name="WINCH_RATE_CONTROL">
-        <description>Wind or unwind cable at specified rate.</description>
+        <description>Wind or unwind line at specified rate.</description>
+      </entry>
+      <entry value="3" name="WINCH_LOCK">
+        <description>Perform the locking sequence to relieve motor while in the fully retracted position. All other command parameters are ignored.</description>
```
hamishwillee, line 594: Well, except the index that identifies the particular target winch. But good enough :-)

hamishwillee:
> I updated the descriptions and consolidated the naming. I added two more states (redeliver and abandon_line) and one more command (abandon_line). Let me know your thoughts or if I missed something.

I'd like to hear @rmackay9 say that he doesn't think this will break anything and that he is OK with the new definitions, in particular with respect to the status. If not, please propose an alternative and acceptable way that a GCS might determine minimally completion the current smart operation and its completion.

- I don't really understand the difference between freewheeling (relaxed) and disengaged. I think that perhaps 
  -  disengaged means you have removed the clutch and stopped the motor. Because the clutch is disengaged there is no connection to the motor at at all, the line can just drop out if there is some load. Right?
  - relaxed/freewheeling means that the clutch is engaged but the motor is not turned on - so this might turn the motor - it will effectively brake using the motor.

dakejahl:
Our winch does not have a clutch, so the term isn't quite right for our use case anyway. We have an electric motor which can be turned on/off and a mechanical lock for the fully retracted position. So yes if the motor is off it will freewheel but the resistance is minimal, it does not brake while freewheeling. If the motor is turned on, it will be controlled from the setpoints which come from the states.

julianoes:
> @julianoes Could/should could use yet another protocol flag or two on the component meta? Or would you propose a better way?

I suppose we could/should do that.

> I don't really understand the difference between freewheeling (relaxed) and disengaged. I think that perhaps

My understanding was that there are two things at play:
1. locking using a physical lock.
2. a motor with a clutch

Both of these can be "used":
1. Locked, or unlocked (relaxed?)
2. Clutch engaged (connected), clutch disengaged (disconnected)

If unlocked and clutch disengaged, it's freewheeling.

julianoes:
> Bit off topic but use of param_ext should be discouraged. Unless there's no way to use the normal param protocol I'd steer well clear of it.

I guess it's time to come up with param version 3 to avoid this.

auturgy:
Yeah, that's been raised a few times, but I think we're all lacking the emotional energy to fight through the challenges of doing that. There be dragons, etc.
Perhaps we should take it offline and have a chat about it.

message_definitions/v1.0/common.xml
hamishwillee: @rmackay9 Either the clutch is engaged and the motor is driving it or the clutch is disengaged and the motor is effectively freewheeling. So as it is now you have two mutually exclusive things in one sentence. Given the name is "engaged" I would suggest the new text is correct. Are you interpreting this as "motor is freewheeling" or not?

rmackay9: @hamishwillee, hmm.. This may be a result of a Japanese to English translation issue.  I'm afraid though that in the AP implementation when the clutch is "engaged" the motor moves freely.

The AP implementation is first, open source and is already being used though so I don't think we should change this for now.

dakejahl: I hate to be this guy but I actually didn't know what was technically correct until I looked it up. I think the confusion might have come from the typical way we think about using a clutch in a car. When you push the "clutch" (foot pedal) in it disengages the actual clutch -- so you're engaging the disengagement mechanism :sweat_smile:
https://mechanics.stackexchange.com/questions/18552/how-does-a-clutch-work

image
Selection_430

hamishwillee: @dakejahl Thanks for "being that guy". I'd much rather be wrong about this than try to explain an incorrect naming convention.

hamishwillee:
@rmackay9 Fortunately I was wrong about the clutch thing. The issue with duplicate actions for commanding a freewheel has also been addressed by removal of "disengage". 

The dev call are happy for this to merge, I plan to do this towards the end of the day (about 5 hours) to give you time to make any final suggestions/objections.

message_definitions/v1.0/common.xml
```diff
@@ -4477,13 +4495,34 @@
         <description>Winch is healthy</description>
       </entry>
       <entry value="2" name="MAV_WINCH_STATUS_FULLY_RETRACTED">
-        <description>Winch thread is fully retracted</description>
```
rmackay9, line 4480: It's nice to see this removal of synonyms.  I clearly wasn't following my own advice when I originally wrote the message.

rmackay9 (approved):

hamishwillee:
Thanks everyone. Merged now.

message_definitions/v1.0/common.xml
julianoes: So, hang on. The description was wrong and "fixed" now. However, as @rmackay9 writes, this breaks the existing implementation, so we shouldn't just do that, right?

rmackay9: oh shoot, I thought that line was backed out.  I'd prefer if it was left as it was until I'm able to reverse the AP implementation.

hamishwillee: I thought it was reverted too. Very sorry. Will fix now.

hamishwillee: Arggh. Reverted, but in a commit not a PR: https://github.com/mavlink/mavlink/commit/2180611b70f4295f76f313211454da321510d46e

My understanding @rmackay9 from the above is that if the clutch is engaged the motor is driving the winch. I think we should use words like that once you have reverted the implementation. 

Do you have an approximate timeframe.

rmackay9: @hamishwillee I don't have a timeline I'm afraid... but I guess that dakejahl can just go against the spec in the meantime.  I don't think there's any other choice really..

dakejahl: It doesn't really matter since I have a custom UI for all of this anyway. If a GCS implements a UI I would expect it to use the same terms as in the mavlink spec, so if anything the user would just be confused when they see "CLUTCH ENGAGED" and the thing is freewheeling (I use different terms in our UI). The point of this PR is to formalize the concept of a smart winch into the mavlink spec which I think we succeeded at. Correct descriptions would be nice to have but aren't required in this case.

julianoes: Or we can document that this is phrased incorrectly but that's how it is.

hamishwillee: I prefer we highlight the inconsistency - @rmackay9 would any of the options below suit you (I like the first).

```xml
        <description>Winch clutch is disengaged. The motor is moving freely and not driving the winch.
          Note that this flag is incorrectly named ("clutch engaged" should mean that the motor is driving the winch).</description>
```
```xml
        <description>Winch clutch is engaged allowing motor to move freely.
          Note that this flag is incorrectly named ("clutch engaged" should mean that the motor is driving the winch).</description>
```
Going forward, changing the meaning of the enum is a semantic break. We can do that if @rmackay9 and @dakejahl are happy to manage it in such a way that we know we are not causing safety issues among potential users. We can live with that, though changing is desirable if we can.

julianoes: I think the first suggestion is good.

hamishwillee: Cool. PR created in https://github.com/mavlink/mavlink/pull/1781#issue-1099927256

PR #1768, opened 2022-01-05
MAV_PROTOCOL_CAPABILITY_MISSION_FLOAT - deprecate

hamishwillee:
`MAV_PROTOCOL_CAPABILITY_MISSION_FLOAT` indicates support for the deprecated `mission_item` message. This update marks it as deprecated.

I believe the plan is actually to delete these flags at some point, when they are not being set by active systems.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2930,14 +2930,14 @@
     <enum name="MAV_PROTOCOL_CAPABILITY" bitmask="true">
       <description>Bitmask of (optional) autopilot capabilities (64 bit). If a bit is set, the autopilot supports this capability.</description>
       <entry value="1" name="MAV_PROTOCOL_CAPABILITY_MISSION_FLOAT">
-        <deprecated since="2022-05" replaced_by="NA">The associated message type (MISSION_ITEM) is deprecated and should no longer be used.</deprecated>
+        <deprecated since="2022-05" replaced_by="">The associated message type (MISSION_ITEM) is deprecated and should no longer be used.</deprecated>
         <description>Autopilot supports MISSION_ITEM float message type.</description>
       </entry>
       <entry value="2" name="MAV_PROTOCOL_CAPABILITY_PARAM_FLOAT">
         <description>Autopilot supports the new param float message type.</description>
       </entry>
       <entry value="4" name="MAV_PROTOCOL_CAPABILITY_MISSION_INT">
-        <deprecated since="2020-06" replaced_by="NA">This flag must always be set if missions are supported, because missions must always use MISSION_ITEM_INT (rather than MISSION_ITEM, which is deprecated).</deprecated>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Note, @auturgy @julianoes  Both PX4 and ArduPilot need to stop indicating support for this in current versions. Neither actually do send it or receive it.

auturgy:
I don't see the need here.  MISSION_FLOAT has been deprecated, not removed.  So if the flight stacks still support it, it should still be indicated.

hamishwillee:
> I don't see the need here. MISSION_ITEM (float) has been deprecated, not removed. So if the flight stacks still support it, it should still be indicated.

I think you're arguing that ArduPilot should not stop setting this bit right, not the PR itself?

As discussed in https://github.com/mavlink/mavlink/issues/1285 the bit does not need to be set because it will always be ignored. If the GCS receives a mission_item it is expected to return a mission_int. This was shown by Don to work with both ArduPilot and PX4.

So the "next steps" would be that if the protocol bit and the message are being used we should stop doing so. After all, the point of deprecation is to indicate that something better/more flexible/more recent :-) has come along.

julianoes:
> Eventually we can reuse the bit for something else.

That's the question. Can we ever do that.

hamishwillee:
>> Eventually we can reuse the bit for something else.

> That's the question. Can we ever do that.

Sure we could if we had the will. For example, we could update DroneKit, the biggest dependency, to behave like QGC. Within a few years everyone would be on that version and the flag could be removed everywhere. Later on we could remove the messages themselves.

Here's the rub. No one is willing to maintain DroneKit, and yet we're treating it as an important dependency that can't be broken. No protocol or standard can evolve if it is forced to carry abandonware forever.

hamishwillee:
@auturgy @julianoes In spite of our discussion in the dev call, I have updated this to _remove_ the deprecation on both `MAV_PROTOCOL_CAPABILITY_MISSION_INT` and  `MAV_PROTOCOL_CAPABILITY_MISSION_FLOAT`  (and added a note that the float protocol is deprecated).

If you mark the bit as deprecated you're not signaling that the protocol is deprecated: you're signaling that the bit isn't needed.  That's not the case for  `MAV_PROTOCOL_CAPABILITY_MISSION_FLOAT` - as indicated by the "DroneKit" discussion. 

This was the original logic for marking `MAV_PROTOCOL_CAPABILITY_MISSION_INT` as deprecated - the theory is that it does not matter if you set this bit or not because it will always be supported. 
This is short term thinking. In the general case of a protocol you should never assume it is "finished". If we introduce a new message, or more likely  `MAV_PROTOCOL_CAPABILITY_MISSION_V3`, then eventually it will not be possible to make this inference.

Un-deprecating this shouldn't break anything

Can you sanity check that logic?

hamishwillee:
@julianoes @auturgy Does this make sense ^^^

julianoes (approved):

julianoes:
I think that makes sense, yes.

hamishwillee:
Merging. @auturgy agreed offline that this makes sense too.

PR #1771, opened 2022-01-06
MAV_PROTOCOL_CAPABILITY_PARAM_UNION - never implemented

hamishwillee:
[MAV_PROTOCOL_CAPABILITY_PARAM_UNION](https://mavlink.io/en/messages/common.html#MAV_PROTOCOL_CAPABILITY_PARAM_UNION) was described as :

> Autopilot supports the new param union message type.

- There does not appear to be any message that might conceivably be this "param union" message.
- There is no mention or appearance it on search of mavlink/mavlink, other than this issue raised by OllieW #1153 which I should have properly addressed at the time.
- Grep shows this bit is not set in the PX4 or ArduPilot codebase.

My assumption is that whatever it was intended for never happened. Now I could just deprecate this, but I think it can actually be safely reused. I have therefore indicated that in this change.

@auturgy @julianoes You OK with this?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2960,8 +2960,8 @@
       <entry value="8" name="MAV_PROTOCOL_CAPABILITY_COMMAND_INT">
         <description>Autopilot supports COMMAND_INT scaled integer message type.</description>
       </entry>
-      <entry value="16" name="MAV_PROTOCOL_CAPABILITY_PARAM_UNION">
-        <description>Autopilot supports the new param union message type.</description>
+      <entry value="16" name="MAV_PROTOCOL_CAPABILITY_RESERVED1">
+        <description>Reserved for future use.</description>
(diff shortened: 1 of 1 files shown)


julianoes (approved):
I think that's indeed not used in PX4

auturgy:
It’s a C helper thing. ArduPilot has never used it to the best of my knowledge.

https://mavlink.io/en/services/parameter.html#mavgen-c-api

Regards,

James

hamishwillee:
I'll take that as "can be merged".

auturgy:
This needs to be reverted and used by px4, as it describes the parameter protocol used by px4.

hamishwillee:
Note, reverted in https://github.com/mavlink/mavlink/commit/b724aab56ad74695d446bf0d4239ad5ad2b7198c

Will take a shot at updating in a separate PR, because the current docs are also wrong/unhelpful.

PR #1772, opened 2022-01-06
[Discuss.] MAV_PROTOCOL_CAPABILITY_FTP - supports FTP protocol not specific message

hamishwillee:
[MAV_PROTOCOL_CAPABILITY_FTP](https://mavlink.io/en/messages/common.html#MAV_PROTOCOL_CAPABILITY_FTP) currently says:

> Autopilot supports the new FILE_TRANSFER_PROTOCOL message type.

This changes that to mean "supports FTP protocol v1 as documented in the devguide" (and now supported by PX4, ArduPilot, QGC, MAVSDK and possibly more). If you don't support this version you don't set it.

The justification for this is that it provides us a clear link to the actual protocol implemented by all these cases, so that if we make this better in future (please!) we have an upgrade path. We should define a bit for this, and I'd prefer it was this bit so that none of the existing systems need to do anything to be compliant. 

**Further, this removes the discussion of "opaque/private" implementations.** I don't think there are any or that they would be useful, and this is a much clearer story. I will have to make corresponding change to docs if this is accepted.

@julianoes @auturgy @LorenzMeier

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2964,7 +2964,7 @@
         <description>Autopilot supports the new param union message type.</description>
       </entry>
       <entry value="32" name="MAV_PROTOCOL_CAPABILITY_FTP">
-        <description>Autopilot supports the new FILE_TRANSFER_PROTOCOL message type.</description>
+        <description>Autopilot supports the File Transfer Protocol v1: https://mavlink.io/en/services/ftp.html.</description>
       </entry>
       <entry value="64" name="MAV_PROTOCOL_CAPABILITY_SET_ATTITUDE_TARGET">
         <description>Autopilot supports commanding attitude offboard.</description>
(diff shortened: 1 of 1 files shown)


julianoes (approved):

hamishwillee:
This is waiting on @auturgy to confirm that the ArduPilot implementation is indeed compatible/standardized with respect to the implementations in QGC, PX4, MAVSDK. 

Note, this was updated to remove the mention of opaque private FTP protocols.

julianoes (approved):

hamishwillee:
@auturgy Gentle ping. Waiting on you to confirm ardupilot uses compatible MAVFTP so could talk to MAVSDK or QGC seamlessly, all things being equal.

hamishwillee:
Merging. Pretty sure this accurately reflects discussion in dev call.

PR #1774, opened 2022-01-06
MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION - improve linking to command

hamishwillee:
`MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION` this explicitly names the supported command. That both removes any ambiguity about the command (were there to be one in future) and also creates a link in the documentation. 

Note that we probably shouldn't have wasted a protocol bit on this. It isn't needed - you could just sent the command with param1=0 and the vehicle should not terminate, but should ACK to indicate that it understand the command.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1379,7 +1379,9 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="185" name="MAV_CMD_DO_FLIGHTTERMINATION" hasLocation="false" isDestination="false">
-        <description>Terminate flight immediately</description>
+        <description>Terminate flight immediately.
+          Support for this command can be tested by sending the command with param1=0 or using the protocol bit: MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION.
+        </description>
         <param index="1" label="Terminate" minValue="0" maxValue="1" increment="1">Flight termination activated if &gt; 0.5</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1380,8 +1380,12 @@
       </entry>
       <entry value="185" name="MAV_CMD_DO_FLIGHTTERMINATION" hasLocation="false" isDestination="false">
         <description>Terminate flight immediately.
-          Support for this command can be tested by sending the command with param1=0 (&lt; 0.5); the ACK should be either MAV_RESULT_FAILED or MAV_RESULT_UNSUPPORTED.
-          You can also test support using the protocol bit: MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION.
+          Flight termination immediately and irreversably terminates the current flight, returning the vehicle to ground.
+          The vehicle will ignore RC or other input until it has been power-cycled.
+          Termination may trigger safety measures, including: disabling motors and deployment of parachute on multicopters, and setting flight surfaces to initiate a landing pattern on fixed-wing).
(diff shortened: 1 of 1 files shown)


hamishwillee:
We should define what flight termination means:
- ArduPlane https://ardupilot.org/plane/docs/advanced-failsafe-configuration.html#afs-termination
- PX4: https://docs.px4.io/master/en/advanced_config/flight_termination.html#flight-termination-configuration
- ArduCopter?

On PX4 the FC sets failsafe values, which for FW attempt a glide land, and for MC just turn off all motors. You can also trigger things like a chute off unused ports. On ArduPlane Plane it looks like this does similar, but not clear what Copter does

I think the common story here is that the vehicle will immediately land and that this operation cannot be reset until the vehicle has landed? Anything much else that can be said? I'd like to mention turning off motors if both platforms do that.

@auturgy @julianoes Otherwise OK?

julianoes:
> I think the common story here is that the vehicle will immediately land

I would not call it land although it will eventually "land" on the ground. :trollface:  In my mind termination is something like parachute, fixed bank angle, or drop like a stone. Otherwise you could just tell it to "land" which would be more graceful.

message_definitions/v1.0/common.xml
```diff
@@ -1379,7 +1379,9 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="185" name="MAV_CMD_DO_FLIGHTTERMINATION" hasLocation="false" isDestination="false">
-        <description>Terminate flight immediately</description>
+        <description>Terminate flight immediately.
+          Support for this command can be tested by sending the command with param1=0 or using the protocol bit: MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION.
```
julianoes, line 1383: Not so sure about this. We have to check if this is the current implementation. Otherwise dangerous and unexpected things can happen.

hamishwillee: Good point. 
- PX4 does not appear to support this! 
- Ardupilot does this - only returns MAV_RESULT_ACCEPTED if termination is supported on the frame AND value is over 0.5. Returns MAV_RESULT_FAILED for other value. Pedantically that is probably incorrect, but I'll update documentation to reflect it.

message_definitions/v1.0/common.xml
```diff
@@ -1379,8 +1379,15 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="185" name="MAV_CMD_DO_FLIGHTTERMINATION" hasLocation="false" isDestination="false">
-        <description>Terminate flight immediately</description>
-        <param index="1" label="Terminate" minValue="0" maxValue="1" increment="1">Flight termination activated if &gt; 0.5</param>
+        <description>Terminate flight immediately.
+          Flight termination immediately and irreversably terminates the current flight, returning the vehicle to ground.
+          The vehicle will ignore RC or other input until it has been power-cycled.
```
hamishwillee, line 1384: @auturgy @julianoes I have updated `MAV_CMD_DO_FLIGHTTERMINATION` as discussed in the dev call. Can it be merged?

NOTE, that this line is the only one that concerns me. It states that flight termination is irreversible (as agreed) BUT it also says that the vehicle will _ignore RC or other input until it has been power-cycled_.
Is that true? i.e. is it a power cycle that is required for restarting? Can other inputs be made during the crash?

julianoes: I would say: yes that's how it should be.

hamishwillee:
Agreed by @auturgy too (offline). Merging. 

@julianoes I think I assigned you a bug in PX4 to match to this behaviour (last week).

PR #1775, opened 2022-01-06
MAV_PROTOCOL_CAPABILITY_FLIGHT_INFORMATION - mark as reserved capability

hamishwillee:
`MAV_PROTOCOL_CAPABILITY_FLIGHT_INFORMATION` indicates support for the flight information protocol but that is not documented. Looking into this, all it means is that [FLIGHT_INFORMATION](https://mavlink.io/en/messages/common.html#FLIGHT_INFORMATION) an be emitted on request, using `MAV_CMD_REQUEST_MESSAGE`

That's a waste of a capability bit. You can just send `MAV_CMD_REQUEST_MESSAGE` to determine if it is supported, which has no problematic side-effects. Further, it is just as likely that if you trust this capability bit you will be wrong. For example, I believe PX4 emits this message but does not set the bit. 

As no one sets the bit and it serves no useful purpose, I have marked it as reserved.

@auturgy @julianoes OK with this? If not I can make the text to "Autopilot supports the requesting the FLIGHT_INFORMATION message.". But I'd rather just delete it since we can without consequence.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2996,8 +2996,8 @@
       <entry value="32768" name="MAV_PROTOCOL_CAPABILITY_MISSION_RALLY">
         <description>Autopilot supports mission rally point protocol.</description>
       </entry>
-      <entry value="65536" name="MAV_PROTOCOL_CAPABILITY_FLIGHT_INFORMATION">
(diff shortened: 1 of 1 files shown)


auturgy:
Lgtm

julianoes (approved):

PR #1776, opened 2022-01-06
pymavlink: update submodule to latest master

julianoes:
It looks like there were improvements that we should make use of.


hamishwillee:
I never know how to assess this kind of PR so I am just trusting that everyone has being doing the right thing.

julianoes:
@hamishwillee me too. Try it out and if we find regressions we could always go back.

PR #1778, opened 2022-01-11
add extended sensor status bits to MAVLink UDP example

tatsuy:
This fixes the compile error of mavlink_udp.c.
I added extended bits added by #1751.


hamishwillee:
The instructions here https://mavlink.io/en/mavgen_c/example_c_udp.html#buildingrunning-the-example and here https://github.com/mavlink/mavlink/blob/master/examples/linux/README.md assume MAVLink 1, for which this still works.
I am happy to accept this but we need to update the instructions too.

So have updated in the devguide here: https://github.com/mavlink/mavlink-devguide/pull/427. Will update the other in a post-process.

Thanks very much.

PR #1784, opened 2022-01-13
MAV_PROTOCOL_CAPABILITY_SET_ACTUATOR_TARGET for ACTUATOR_CONTROL_TARGET

hamishwillee:
According to archeology research by @peterbarker , `MAV_PROTOCOL_CAPABILITY_SET_ACTUATOR_TARGET` indicates support for [ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#ACTUATOR_CONTROL_TARGET). That is not a very useful message because it has no sysid or component id - and you would not want to broadcast this. 

There is another message one id earlier [SET_ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#SET_ACTUATOR_CONTROL_TARGET) that does have the id. But you couldn't rely on it being supported. 

This is not set by PX4 or ArduPilot. I'm tempted to mark it reserved as not used, because it is not useful. This change is the minimum for it not to be ambiguous.

@julianoes @auturgy OK, or should we just mark it reserved?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2982,7 +2982,7 @@
         <description>Autopilot supports terrain protocol / data handling.</description>
       </entry>
       <entry value="1024" name="MAV_PROTOCOL_CAPABILITY_SET_ACTUATOR_TARGET">
-        <description>Autopilot supports direct actuator control.</description>
+        <description>Autopilot supports the ACTUATOR_CONTROL_TARGET message.</description>
       </entry>
       <entry value="2048" name="MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION">
         <description>Autopilot supports the flight termination command.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2981,8 +2981,8 @@
       <entry value="512" name="MAV_PROTOCOL_CAPABILITY_TERRAIN">
         <description>Autopilot supports terrain protocol / data handling.</description>
       </entry>
-      <entry value="1024" name="MAV_PROTOCOL_CAPABILITY_SET_ACTUATOR_TARGET">
-        <description>Autopilot supports the ACTUATOR_CONTROL_TARGET message.</description>
+      <entry value="1024" name="MAV_PROTOCOL_CAPABILITY_RESERVED3">
+        <description>Reserved for future use.</description>
       </entry>
       <entry value="2048" name="MAV_PROTOCOL_CAPABILITY_FLIGHT_TERMINATION">
         <description>Autopilot supports the flight termination command.</description>


julianoes (approved):

julianoes:
We could also mark it reserved, or use it for a general support for offboard/guided capability which was probably the intent.

hamishwillee:
@auturgy @julianoes So, since this is not used in ArduPilot or PX4, and the associated API looks like something you shouldn't use because it is a bit broken, I have marked it as reserved. We all good with that?

auturgy (approved):

Jaeyoung-Lim:
> I think marking it as reserved makes sense in this case because I can't see anyone using this message.

> IMO best to never use a capability bit to indicate general support for anything - it's only useful it what it stands for is specific and versioned and well documented. Otherwise a GCS can't make any assumptions.

@hamishwillee @julianoes Too bad, I was using the `SET_ACTUATOR_TARGET` message for generating system identification maneuvers. Also, I have seen this message be used a lot for control applications, where they want access to direct actuator(motor) control

peterbarker:
On Mon, 17 Jul 2023, JaeyoungLim wrote:

> @hamishwillee @julianoes Too bad, I was using the SET_ACTUATOR_TARGET message for generating system identification maneuvers. Also, I have seen this message be used a lot for control applications, where they
> want access to direct actuator(motor) control

I'm a little surprised identification works well given latencies involved. 
I expect that doesn't matter if you're using the results in post-flight 
analysis.

I've seen various on-board solutions for system identification - some 
scripting-based things like the Plane couplets script, but also Bill's 
"chirping" stuff for Heli and Copter's "system id" mode.

Jaeyoung-Lim:
> I'm a little surprised identification works well given latencies involved.
I expect that doesn't matter if you're using the results in post-flight
analysis.

@peterbarker Exactly, the latencies are between the mavlink link, not the actuator delays, and therefore not a problem for running identification with these messages.

hamishwillee:
> Given that said, does this PR remove support for this message? or can I still use it? Since it seems that the message definition still exists. (What is the actual difference between MAV_PROTOCOL_CAPABILITIES* and the message itself?)

The `MAV_PROTOCOL_CAPABILITIES_` is used to provide protocol bits that you can use to check if some feature is supported - the theory being that if you want to use a feature such as "mission protocol" you can know in advance. The theory is a bit broken because historically these have been misapplied. Further they are a limited number of bits to work with - so we haven't been using them out of fear. We should use them more or find a better way, but every attempt at that has been thwarted by someone.

This particular case the meaning of the bit might be inferred to be support for [ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#ACTUATOR_CONTROL_TARGET) in the flight stack. Since no one is setting it, the protocol bit is less than useless. 

In any case, no, the associated message is not deprecated. I just don't understand what it is for.

If it is useful to know that this message is supported, and you know that it is supported by PX4, we can add this bit back, but you would need to also set the bit in PX4 so that it actually is useful.

----

PS The command sets actuator values to specific values right? What does it mean to "using the SET_ACTUATOR_TARGET message for generating system identification maneuvers". Moving a vehicle in such a way you can tell what it is if you are watching the vehicle?

Jaeyoung-Lim:
@hamishwillee Sorry for the late reply,

> PS The command sets actuator values to specific values right? What does it mean to "using the SET_ACTUATOR_TARGET message for generating system identification maneuvers". Moving a vehicle in such a way you can tell what it is if you are watching the vehicle?

yes exactly. This is an example of commanding an actuator input and observing the response to workout what the system dynamics model is

137283988-0c76629f-6f1d-4982-bc23-90b335a941e7
137282663-c178796b-e741-4609-8ebb-a9359ba9d724

hamishwillee:
@Jaeyoung-Lim Hmmm. 
- This was actually removed entirely from PX4 in https://github.com/PX4/PX4-Autopilot/pull/20837 in March and will not be in 1.14.
- The messages are not "general"- they refer to an actuator mapping that assumes the PX4 mixer actuator mapping is "standard"
- The actuator_control_target and set_actuator_control_target have the same description.

My thinking is we should actually deprecate these and work towards removal. Thoughts?

hamishwillee:
@peterbarker How do ArduPilot do direct actuator control on arbitrary actuator? Just use RC_OVERRIDES?

peterbarker:
> @peterbarker How do ArduPilot do direct actuator control on arbitrary actuator? Just use RC_OVERRIDES?

Mmmm... You can't *just* use RC_OVERRIDES as the servo scaling might be weird; a PWM-in-from-RC_OVERRIDES of 2000 might yield a servo output of 1700 if `SERVOn_MAX` is 1700 and the RC channel's MAX was 2000.

RC overrides can work, but you need to set `SERVOn_FUNCTION` to 1 to say "copy the 1-1 mapped rc channel to the servo output (roughly speaking; there are also functions for each rc input channel).

> Context is that we can't remove these messages immediately as they are in use. However if no one is going to use them eventually we should deprecate.

The intent of the message in the description just doesn't seem to match the description of the controls field; one would have to assume descriptions of both messages are just wrong (one obviously is).  It's unclear how this really corresponds to `RC_CHANNELS` as the controls seem to be allowing you t o say "give me actuator outputs that correspond to 20%  maximum pitch and 40% maximum roll actuator outputs).  Presumably some sort of weird multi-axis mxiing has to be in play given actuator coupling for e.g. yaw on a quadcopter).  This isn't how you described them above, as "direct actuator control"....
 
They don't look particularly useful to me; I'm sure they were for the original author at the time!

> Further, if the use case is not met elsewhere it would be good to have messages. I'm not sure these ones are fit for purpose, as they appear to match the PX4 mixer design not "generic actuator control".

These don't appear to allow for direct control of single actuators.  Even "direct control" is a bit misleading, as the range can be different based on the actuator.  For example, we could have a `SET_SERVO_OUTPUT_RAW` message, which would be quite useful for people doing system identification work, for introducing step-jumps in servo output to try to tickle hardware bugs and for calibrating ESCS to specific ranges.  Note, however, that this is *not* "actuator control" in that many actuators aren't PWM-based, e.g. relays, stepper motors, Torqeedo, d-shot....

peterbarker:
... perhaps `SERVO_RAW_OVERRIDE` would be more appropriate.

hamishwillee:
Thanks @peterbarker - much appreciated. I think by "direct" @Jaeyoung-Lim just means "bypassing estimators, controllers, etc".

Based on the naming and description I'd say that this was way to specify normalized outputs for roll pitch yaw etc using the same definitions as used by old PX4 mixer architecture https://docs.px4.io/v1.13/en/concept/mixing.html#control-group-0-flight-control
In other words, a way to identify the servo by purpose and set a normalised output, without having to supply a mapping to define what the actual output you are mapping to is, or its type. 

So not "direct" control but you can do the same thing with RC_OVERRIDES but you need to know about what your ESCs/servos are and how the RC is mapped - it would be less "generic" to set up.

If I had to do this I'd suggest a new _message_ like [MAV_CMD_DO_SET_ACTUATOR](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ACTUATOR) where the values are normalized and the purpose of each output is mapped by the flight stack using a parameter. This would be a bit like an RC_OVERRIDE but sent with normalized values, and as a user you could easily define the output mapping rather than it depending on physical writing and mapped channels.

W.r.t. the identification use case, I don't know much about this. Are there other ways? Such as just sending a message which provides the required info?

@Jaeyoung-Lim No urgency, but if you have suggestions on a message that would suit your needs ... (if the one above does not).

julianoes:
Not sure if that came up above but for what it's worth, PX4 no longer supports [SET_ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#SET_ACTUATOR_CONTROL_TARGET).

https://github.com/PX4/PX4-Autopilot/issues/21058

It's still in MAVSDK though.

hamishwillee:
FYI @julianoes I just added that to the PX4 v1.14 project board. I can't see any sane reason for its removal.

Jaeyoung-Lim:
@julianoes Yes, this broke when control allocation replaced the mixer

julianoes:
@hamishwillee I agree that it's something fundamental to have but I had a discussion with @bkueng in some thread and he argued that the uORB/ROS 2 interface should be used instead for this use case. Anyway, that's a PX4 discussion.

Jaeyoung-Lim:
@julianoes I agree with @bkueng , but this would make a big part of the research community move away from PX4, which using ros2 is not an option.

hamishwillee:
@julianoes Without discussing with @bkueng I couldn't say. But fundamentally removing support for any MAVLink API, which is our stated integration API without any notification or plan is screwed up. 

FYI it does sound like [SET_ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#SET_ACTUATOR_CONTROL_TARGET) might still work but I realised I don't have hardware to test this.

julianoes:
@hamishwillee I doubt it.

v1.12: https://github.com/PX4/PX4-Autopilot/blob/release/1.12/src/modules/mavlink/mavlink_receiver.cpp#L1117

v1.13:
https://github.com/PX4/PX4-Autopilot/blob/release/1.13/src/modules/mavlink/mavlink_receiver.cpp#L1259

v1.14: nowhere to be seen

hamishwillee:
Boo hiss.

eng-86:
> Boo hiss.

I am a PhD student. I am working on a distributive fault tolerant control for multi-UAVs. Each UAV can estimate its local actuator fault and neighbors fault relying on torques and thrust control inputs to each UAV in addition to UAVs' states. I was using a one-year-recent version PX4 with ROS  to verify my proposed algorithms. It was working except that the real time scale in GAZEBO persists at 0.2 when trying to simulate four vehicles. This caused an accuracy problem (the algorithm was also validated through matlab simulation). When I upgraded the code to the latest one, the real time problem was solved, but a new problem appeared. My algorithm depends on desired mavros actuator target controls for each UAV to estimate the actuator fault, and it is not published in new versions (I checked it during simulation and according to your discussion also). I think it is important to publish such message. Any ideas to solve this issue?

eng-86:
hello again, I compared also two log data samples. I found that in old version, the actuator control log was found (Roll, Pitch, Yaw, Thrust (up), Thrust (forward). @Jaeyoung-Lim these parameters as very important in research domain.
bokeh_plot

sfuhrer:
> @julianoes I agree with @bkueng , but this would make a big part of the research community move away from PX4, in which using ros2 is not an option.
> 
> I have already seen quite some groups(including the one I am in) either move to betaflight or just create something completely custom due to incompatible breaking changes.

@Jaeyoung-Lim are you referring to the same as @eng-86 ? We likely need to bring back support for sending mavlink messages for torque and thrust setpoinst (what ACTUATOR_CONTROL_TARGET was doing, but was tied to the old mixing system).

Issue #1686, opened 2021-08-26
Concerns with component info protocol


hamishwillee:
Following on from some discussions here https://github.com/mavlink/mavlink/issues/1669#issuecomment-903393762

This copies (and removes) a discussion by @olliw42 that was formerly here: https://github.com/mavlink/mavlink/issues/1669#issuecomment-905133926

----

ok, I do what I was hesitant to do, let me try to bring across some thoughts ... it will be a longer text and if you decide to read it I like to ask to read it carefully since I may not be able to perfectly describe my thoughts in full in words ... also I'm not as deep into the component_info business as I was 1-2 years ago so I may have forgotten some key points I would have made otherwise, but this is what is obvious to me:

first off, although it's interesting and important to understand why some things are as they are, let's put history aside.

**1. Have a clear concept or "Have a clear mind what should be in and what should not be in COMP_INFO"**

That's a key point IMHO, and I see lot's of meandering here. My key argument will be:

The data which is accessible via the COMP_INFO definition file MUST NOT be crucial for the functioning of the MAVLink services.

**A corollary is:** Component specific XXX_INFO messages and COMP_INFO not only can coexist but also may HAVE TO coexist (since they serve fundamentally different purposes)!

As often, a simple example helps to clear up minds. Let's consider GIMBAL_DEVICE_INFORMATION. Besides all sorts of blabla it has a field cap_flags (and other similar fields), which carries most important information which any party, be it an autopilot, a companion, a gcs, or whatever potentially needs to know in order to operate properly. It would be - IMHO - simply a disaster if all these parties would have to fetch that info through the heavy weight and cumberness of mavftp ...

Ergo, here there is - IMHO - a clear need for the existence of a GIMBAL_DEVICE_INFORMATION message. The parts I just called "blabla", like the name and version fields, are in contrast indeed parts which could go into a COMPONENT_INFORMATION message - because this is info which makes sense for ANY component, be it a gimbal, camera, servo, gps.

Similarly so for CAMERA_INFORMATION, it has also a capabilities flag and other crucial info, which e.g. a companion would need and it would be a disaster if that would have to be done by mavftp. But also similarly so it has this "blabla".

Here, btw, you also have seen a new argument/point coming up, namely that mavftp is heavy weight and cumberfull, and http may not be even available ... which will lead to point 2 below.

**Another corollary is:** Don't use it as a general purpose bag for everything.

[...]

PR #1785, opened 2022-01-14
Component information: add note

julianoes:
~The component information has been implemented in the latest QGC 4.2.0, therefore it makes sense to stabilize this and remove the WIP tag.~ We'll leave it there a bit longer until we are confident this is good to go.

Also, this adds an explanation of why we have general and peripherals URLs, as well as that there can be more files inside the general metadata file.


bkueng (approved):

auturgy:
Do we need to work through https://github.com/mavlink/mavlink/issues/1686 prior to removing wip?

PR #1786, opened 2022-01-14
component_info: re-use protocol capabilities

julianoes:
I think it might make sense to re-use the existing capability flags from AUTOPILOT_VERSION also for COMPONENT_INFORMATION_BASICS. Otherwise, we have weird parallel flags that conflict with the existing ones.

If/when we run out of these 64 flags, we can extend both messages, AUTOPILOT_VERSION, and COMPONENT_INFORMATION_BASICS with another 64 bits of MAV_PROTOCOL_CAPABILITY_2.

@hamishwillee,@olliw42, @auturgy what do you think? Good idea or bad idea?

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -120,33 +120,6 @@
           Group id is limited because only 24 bit integer can be stored in 32 bit float.</param>
       </entry>
     </enum>
-    <enum name="COMPONENT_CAP_FLAGS1" bitmask="true">
-      <description>Component capability flags 1 (Bitmap)</description>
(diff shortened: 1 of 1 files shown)


LorenzMeier (approved):
Makes sense, dev call suggests to merge.

olliw42:
@julianoes 
many thx for pinging me, much appreciated
unfortunately I do not have an opinion and would have to spend time to think about it which I currently don't have, i.e., I'm of no help here, sorry sir

hamishwillee:
Yes, this makes lots of sense. You might end up with a capability or two that don't make sense for a particular component, but that is much better than duplicating all the other cases, and keeping things in sync.

PR #1790, opened 2022-02-02
COMPONENT_INFORMATION: dynamic updates

bkueng:
A new use-case for component information peripherals has come up: keeping track of attached smart batteries (name, serial number, etc.). This means we need to handle dynamic updates.
For that a new message is added that is broadcast at regular intervals (after an update), containing the minimal state (the CRC's) with which clients can detect updates and re-request the metadata.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -182,6 +182,11 @@
       <field type="char[24]" name="hardware_version">Hardware version. The version format can be custom, recommended is SEMVER 'major.minor.patch'.</field>
       <field type="uint64_t" name="capabilities" enum="MAV_PROTOCOL_CAPABILITY" display="bitmask">Component capability flags</field>
     </message>
+    <message id="397" name="COMPONENT_INFORMATION_STATE">
+      <description>Contains the current checksums from COMPONENT_INFORMATION and is intended for dynamic updates. For hosts: this message is optional and only required when metadata changes dynamically after bootup. In that case the host broadcasts this message at regular intervals with the updated CRC's. For clients: if a new CRC is detected, re-request the metadata via COMPONENT_INFORMATION.</description>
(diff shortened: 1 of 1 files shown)


julianoes (approved):
I think this makes sense.

hamishwillee:
@auturgy Thoughts on this addition? 

My first reaction was use an enum and broadcast for just the affected file. Was discussed in the [dev call](https://github.com/mavlink/mavlink/wiki/20220202-Dev-Meeting) and this is not needed or desirable.

hamishwillee:
As discussed in  Dev call [20220216](https://github.com/mavlink/mavlink/wiki/20220216-Dev-Meeting) 
- this seems a reasonable proposal for making component information dynamic. 
- this might not be the best way of supplying this information though. The use case is smart batteries, so we could include the info in [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO).
- I am going to make this into Draft. @bkueng please either move back into active or close when you have made a decision on best approach.

@bkueng Just a thought, but is there any way that asset tracking like this could instead be done with [COMPONENT_INFORMATION_BASIC](https://mavlink.io/en/messages/development.html#COMPONENT_INFORMATION_BASIC)? I guess you lose the association with the specific battery being attached but it feels like a more general approach.

bkueng:
> @bkueng Just a thought, but is there any way that asset tracking like this could instead be done with [COMPONENT_INFORMATION_BASIC](https://mavlink.io/en/messages/development.html#COMPONENT_INFORMATION_BASIC)? I guess you lose the association with the specific battery being attached but it feels like a more general approach.

The challenge with that is that a single mavlink component needs to emit multiple sub-components and handle disconnects.

hamishwillee:
Propose you close this, or use it for all the updates. As discussed in https://github.com/mavlink/mavlink/pull/1806#issuecomment-1069746491 we're using a slightly different approach for dynamic updates.

PR #1791, opened 2022-02-03
common.xml: deprecate SET_HOME_POSITION message

peterbarker:
MAVProxy doesn't use this.

MissionPlanner doesn't use this.

QGroundControl doesn't use this

PX4Firmware doesn't use this.

apm_planner2 doesn't use this(!)

The command equivalent, MAV_CMD_DO_SET_HOME allows for the command to be
rejected, so everyone can save flash by removing support for this
message.


hamishwillee (approved):
Sounds very sensible to me. Any objection @julianoes ?

I guess we can't just delete it.

julianoes (approved):
I think that makes sense, thanks @peterbarker.

hamishwillee:
Merged. Yes, thanks @peterbarker

PR #1792, opened 2022-02-07
common: added MAV_CMD_CAN_FORWARD and CAN_FRAME

dagar:
- https://github.com/ArduPilot/mavlink/pull/244

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2456,7 +2456,7 @@
       </entry>
       <!-- END of user range (31000 to 31999) -->
       <entry value="32000" name="MAV_CMD_CAN_FORWARD" hasLocation="false" isDestination="false">
-        <description>Request forwarding of CAN packets from the given CAN bus to this interface. CAN Frames are sent using CAN_FRAME messages</description>
+        <description>Request forwarding of CAN packets from the given CAN bus to this component). CAN Frames are sent using CAN_FRAME messages</description>
         <param index="1" label="bus">Bus number (0 to disable forwarding, 1 for first bus, 2 for 2nd bus, 3 for 3rd bus).</param>
         <param index="2">Empty.</param>
         <param index="3">Empty.</param>
@@ -7012,7 +7012,7 @@
       <description>A forwarded CAN frame as requested by MAV_CMD_CAN_FORWARD.</description>
       <field type="uint8_t" name="target_system">System ID.</field>
       <field type="uint8_t" name="target_component">Component ID.</field>
-      <field type="uint8_t" name="bus">bus number</field>
+      <field type="uint8_t" name="bus">Bus number</field>
       <field type="uint8_t" name="len">Frame length</field>
       <field type="uint32_t" name="id">Frame ID</field>
       <field type="uint8_t[8]" name="data">Frame data</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2455,6 +2455,16 @@
         <param index="7">User defined</param>
       </entry>
       <!-- END of user range (31000 to 31999) -->
+      <entry value="32000" name="MAV_CMD_CAN_FORWARD" hasLocation="false" isDestination="false">
+        <description>Request forwarding of CAN packets from the given CAN bus to this interface. CAN Frames are sent using CAN_FRAME messages</description>
+        <param index="1" label="bus">Bus number (0 to disable forwarding, 1 for first bus, 2 for 2nd bus, 3 for 3rd bus).</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4">Empty.</param>
+        <param index="5">Empty.</param>
+        <param index="6">Empty.</param>
+        <param index="7">Empty.</param>
+      </entry>
     </enum>
     <enum name="MAV_DATA_STREAM">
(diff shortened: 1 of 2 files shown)


message_definitions/v1.0/minimal.xml
```diff
@@ -626,6 +626,9 @@
       <entry value="181" name="MAV_COMP_ID_BATTERY2">
         <description>Battery #2.</description>
       </entry>
+      <entry value="189" name="MAV_COMP_ID_MAVCAN">
```
hamishwillee, line 629: Can there be multiples? Will the component ever emit a heartbeat? If so you also need a MAV_TYPE

message_definitions/v1.0/common.xml
```diff
@@ -2455,6 +2455,16 @@
         <param index="7">User defined</param>
       </entry>
       <!-- END of user range (31000 to 31999) -->
+      <entry value="32000" name="MAV_CMD_CAN_FORWARD" hasLocation="false" isDestination="false">
+        <description>Request forwarding of CAN packets from the given CAN bus to this interface. CAN Frames are sent using CAN_FRAME messages</description>
```
hamishwillee, line 2459: When you say interface, do you mean this system and component?

message_definitions/v1.0/common.xml
```diff
@@ -6998,6 +7008,15 @@
       <field type="uint8_t" name="payload_length">Length of the data transported in payload</field>
       <field type="uint8_t[128]" name="payload">Variable length payload. The payload length is defined by payload_length. The entire content of this block is opaque unless you understand the encoding specified by payload_type.</field>
     </message>
+    <message id="386" name="CAN_FRAME">
+      <description>A forwarded CAN frame as requested by MAV_CMD_CAN_FORWARD.</description>
+      <field type="uint8_t" name="target_system">System ID.</field>
+      <field type="uint8_t" name="target_component">Component ID.</field>
+      <field type="uint8_t" name="bus">bus number</field>
```
hamishwillee, line 7015:
```suggestion
      <field type="uint8_t" name="bus">Bus number</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -2455,6 +2455,16 @@
         <param index="7">User defined</param>
       </entry>
       <!-- END of user range (31000 to 31999) -->
+      <entry value="32000" name="MAV_CMD_CAN_FORWARD" hasLocation="false" isDestination="false">
+        <description>Request forwarding of CAN packets from the given CAN bus to this interface. CAN Frames are sent using CAN_FRAME messages</description>
+        <param index="1" label="bus">Bus number (0 to disable forwarding, 1 for first bus, 2 for 2nd bus, 3 for 3rd bus).</param>
+        <param index="2">Empty.</param>
```
hamishwillee, line 2461: You could omit all these. By default the MAVLink spec says they will be generated as Reserved with default value of `0`.

hamishwillee (commented):
Seems like this has been agreed by main stakeholders. @julianoes @auturgy Any comments on a MAVLink/ technical level?

@dagar @tridge 
- My understanding is that this is so a GCS can understand what is going on with CAN. So perhaps it will have some way to present this CAN traffic for users - is that right? The reason I ask is that it would be good to document the scope of this.
- Do we need to have some way to check whether the API is supported? E.g. a protocol bit?
- Is there a need to also inject CAN messages from a MAVLink component?

message_definitions/v1.0/common.xml
julianoes: I'd agree with component.

julianoes (commented):
Would be nice to add that indeed.

hamishwillee:
FYI, I won't merge this without an answer to https://github.com/mavlink/mavlink/pull/1792#pullrequestreview-876727435. Experience tells me that if we don't explain why we're doing something at least in the PR part, there is no way to get the information later.

tridge:
> * My understanding is that this is so a GCS can understand what is going on with CAN. So perhaps it will have some way to present this CAN traffic for users - is that right? The reason I ask is that it would be good to document the scope of this.

it is already in the dronecan_gui_tool and in pydronecan. It is fully working now with the ArduPilot mavlink repo

> Do we need to have some way to check whether the API is supported? E.g. a protocol bit?

its a MAV_CMD, so it already has an UNSUPPORTED result

> Is there a need to also inject CAN messages from a MAVLink component?

that is in the implementation already, using the same CAN_FRAME messages

tridge (approved):

tridge:
@dagar I've added CANFD_FRAME and CAN_FILTER_MODIFY to this PR. Both are in the ArduPilot mavlink repo already. 
CANFD over mavlink has been tested, and is supported in pydronecan and dronecan_gui_tool

message_definitions/v1.0/minimal.xml
tridge: these comp IDs are just a convention, not required. I've made the dronecan_gui_tool use it, but I expect the CAN_FRAME message will be implemented in all GCS over time.

message_definitions/v1.0/common.xml
tridge: personally I think it is clearer leaving them in

dagar:
Are we good to go here?

tridge:
> Are we good to go here?

I think so, but maybe you could approve?

tridge:
i'll merge now .... seems pointless waiting on you to approve your own PR

PR #1794, opened 2022-02-11
add ftp.xml containing enums for ftp service

muexxl:
Hi all, 
I added the enums for the ftp service in a seprate XML file according to https://mavlink.io/en/services/ftp.html
This is my first attempt to contribute. not sure if this is how its being done. 
//Muexxl

### DIFF
--- a/message_definitions/v1.0/ftp.xml
+++ b/message_definitions/v1.0/ftp.xml
@@ -1,0 +1,102 @@
+<?xml version="1.0"?>
+<mavlink>
+	<version>3</version>
+	<dialect>2</dialect>
+	<enums>
+		<enum name="MAV_FTP_ERR">
+			<description>MAV FTP Error Codes</description>
+			<entry value="0" name="MAV_FTP_ERR_NONE">
+				<description>No error </description>
+			</entry>
+			<entry value="1" name="MAV_FTP_ERR_FAIL">
+				<description>Unknown failure</description>
+			</entry>
+			<entry value="2" name="MAV_FTP_ERR_FAILERRNO">
+				<description> Command failed, Err number sent back in PayloadHeader.data[1]. This is a file-system error number understood by the server operating system </description>
+			</entry>
+			<entry value="3" name="MAV_FTP_ERR_INVALIDDATASIZE">
+				<description> Payload size is invalid </description>
+			</entry>
+			<entry value='4' name="MAV_FTP_ERR_INVALIDSESSION">
+				<description> Session is not currently open </description>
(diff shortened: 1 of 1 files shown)


peterbarker:
Why the separate file?  I think by convention we'd lump everything together.

I know @hamishwillee has been thinking about trying to separate things a bit, but it should probably be all-or-nothing, not half-here-half-there for logically collected things like ftp messages and their enumerations.

hamishwillee:
Thanks @muexxl.
As @peterbarker says, IF we want to do this, we should put this into common.xml. I do want to make the protocols more modular, but have not had broad agreement.

But I do not want to do this now. The benefit of enums in MAVLink is that they document and enforce allowed values. However the [FILE_TRANSFER_PROTOCOL](https://mavlink.io/en/messages/common.html#FILE_TRANSFER_PROTOCOL) is an opaque field. So there is no way to link these enums to the message, and hence IMO not worth it.

What would be good is to design a replacement message/service that does this right.
The simple approach would be one that just redefines the fields currently defined as being in the payload into the message itself. That would allow enforcement of the structure, and easy packing and unpacking. 
If we were going to do that, we'd also take the time to think about whether this might be further improved by breaking it down into several messages. 

@peterbarker Do you think there is any interest ArduPilot side in improving the message?

hamishwillee:
PS, closing because we can't accept this. Thanks again @muexxl .

peterbarker:
[10:03 AM] Peter Barker: @Muexxl while @Hamish Willee did close that initial pull request, I would actually like to see us get rid of those magic numbers somehow.  If we were to move to an FTP message which was less convoluted any such enumeration should be able to be re-used, so I'm really not sure why we couldn't create it now.  Even if we were not to add it to the mavlink standard, even having an internal ArduPilot enumeration in place of the damned fixed numbers would be nice!  Having them in the standards document would at least make people implementing or adjusting the standard consider what it might mean for other implementors!  Cross-posting this back to the closed PR, too 🙂

(cross-posted from ArduPilot's Discord pymavlink channel)

hamishwillee:
FYI only, the general feeling in the dev call was that there is value in the current message design. Specifically that it mirrors the official FTP spec data structure and sequences in mavlink. So we probably won't change it. I added a note in https://mavlink.io/en/services/ftp.html to capture this:
image

That said, I would not personally be opposed to having an enum instead of the magic numbers if that would help developers. Someone else would have to put that in common.xml. Whether or not it went in would then depend on all stakeholders to the standard, but I expect it would be accepted.

PR #1795, opened 2022-02-11
AVSS dialect version 2

Thomas-Li-AVSS:
1,deleted some obsolete command
2,added 2 enum data
3,added 3 new message for avss drone
4,modified AVSS_PRS_SYS_STATUS

### DIFF
--- a/message_definitions/v1.0/AVSSUAS.xml
+++ b/message_definitions/v1.0/AVSSUAS.xml
@@ -3,16 +3,16 @@
 <!-- AVSS is first commercially available products are parachute recovery systems for commercial drones. -->
 <!-- AVSS contact info: -->
 <!-- company URL: https://www.avss.co -->
-<!-- email contact: josh.boudreau@avss.co thomas.li@avss.co-->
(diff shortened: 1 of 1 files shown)


hamishwillee:
@Thomas-Li-AVSS As this is a dialect and you are one of the stated owners, there is no issue with me merging. 

Just a few notes "FYI only" on usual conventions (in case you're interested):
1. usually enum values are named with a prefix. This helps avoid namespace clashes when the values are generated in many languages. So if you have:
    ```
   <enum name="MAV_AVSS_COMMAND_FAILURE_REASON">
      <entry name="PRS_NOT_STEADY" value="1">
   ```
   You might change it normally to 
    ```
   <enum name="MAV_AVSS_CMD_FAIL_REASON">
      <entry name="MAV_AVSS_CMD_FAIL_REASON_PRS_NOT_STEADY" value="1">
   ```
2. A command like `MAV_CMD_PRS_GET_ARM_ALTI` gets something, but it isn't clear how. Normally in docs I'd say something like "This causes message XXXXX to be emitted:.

Thomas-Li-AVSS:
> 
Thanks for your suggestion, I will revise it in the next PR.

PR #1797, opened 2022-02-15
PREFLIGHT_STORAGE: use lowercase, don't shout

julianoes:
And change "current" to "current ... data" to make that clearer.


hamishwillee (approved):
Thanks @julianoes. Merging - zero risk and some benefit in this change.

PR #1800, opened 2022-02-17
fix common.xml include in AVSSUAS

aler9:
4ee1eeb changed the `<include>` directive of AVSSUAS and added a directory prefix that is now useless, since all dialects were moved into the same folder.

This is causing problems with generation of external libraries.

This PR fixes the issue.


hamishwillee:
Thanks, I've merged. Note that "as a rule" I wouldn't as you're not in the contact list on the file - but since this is unusable otherwise ....

PR #1804, opened 2022-02-23
Add `WATER_CURRENT` and `DVL` messages

bazfp:
There is a need for more marine-based sensors in the MAVLink standard. Some common ones are the DVL + ADCP.

ADCP measures water currents in a column of water. It can be onboard a vehicle or standalone. Measurements can be taken at a range of distances from the device. If measurements are captured from a moving vehicle the measured current **MUST** be subtracted from the vehicle velocity in order to get _actual_ current. For this I have included vehicle velocity and measured readings into the message as both are required. ADCPs are often combined with DVLs in order to get accurate vehicle velocities.

A DVL is used to track vehicle position across the seafloor. There has been prior work on this using ROS and https://www.ardusub.com/developers/dvl-integration.html and `[VISION_POSITION_DELTA](https://mavlink.io/en/messages/ardupilotmega.html#VISION_POSITION_DELTA)`. This enables EKF position updating but this message doesn't sit right with me as it's not vision tracking and it completely hides the sensor type and data from an onboard computer.

The intention for these messages is to expose the sensor data via MAVLink but also to allow DVL to contribute to positioning data.

Telemarine DVL: http://www.teledynemarine.com/Lists/Downloads/Tasman_DVL.pdf

Happy to expand these. I would also like a nice way to incorporate the sensor location (up-look vs down-look). I thought about including "Measurement Depth" in the `WATER_CURRENT` message but am still unsure on how best to tackle these as the measurement depth will be dependent on orientation of the hardware itself. For now I have omitted it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7143,6 +7143,28 @@
       <field type="uint8_t" name="num_ids">number of IDs in filter list</field>
       <field type="uint16_t[16]" name="ids">filter IDs, length num_ids</field>
     </message>
+    <message id="390" name="WATER_CURRENT">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Measured water current values. Note: Actual Water current = (Vehicle Velocity - Measured Current) </description>
+      <field type="float" name="measured_current_x" units="m/s">Measured water current in X (NED) direction</field>
+      <field type="float" name="measured_current_y" units="m/s">Measured water current in Y (NED) direction</field>
+      <field type="float" name="measured_current_z" units="m/s">Measured water current in Z (NED) direction</field>
+      <field type="float" name="distance" units="m">Measurement distance from the sensor</field>
+      <field type="float" name="vehicle_velocity_x" units="m/s">Vehicle velocity when measurement was taken in X (NED)</field>
+      <field type="float" name="vehicle_velocity_y" units="m/s">Vehicle velocity when measurement was taken in Y (NED)</field>
+      <field type="float" name="vehicle_velocity_z" units="m/s">Vehicle velocity when measurement was taken in Z (NED)</field>
+    </message>
+    <message id="391" name="DVL">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Doppler Velocity Log Sensor. Used to provide vehicle velocity </description>
+      <field type="uint8_t" name="has_lock">Has bottom lock to the floor (1 = true)</field>
+      <field type="float[16]" name="altitude_water" units="m">Vehicle altitude to the floor of body of water for each beam</field>
+      <field type="float" name="vehicle_velocity_x" units="m/s">Vehicle velocity in X (NED)</field>
+      <field type="float" name="vehicle_velocity_y" units="m/s">Vehicle velocity in Y (NED)</field>
+      <field type="float" name="vehicle_velocity_z" units="m/s">Vehicle velocity in Z (NED)</field>
+    </message>
     <!-- Rover specific messages -->
     <message id="9000" name="WHEEL_DISTANCE">
       <description>Cumulative distance traveled for each reported wheel.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7143,7 +7143,7 @@
       <field type="uint8_t" name="num_ids">number of IDs in filter list</field>
       <field type="uint16_t[16]" name="ids">filter IDs, length num_ids</field>
     </message>
-    <message id="390" name="WATER_CURRENT">
+    <message id="391" name="WATER_CURRENT">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Measured water current values. Note: Actual Water current = (Vehicle Velocity - Measured Current) </description>
@@ -7155,7 +7155,7 @@
       <field type="float" name="vehicle_velocity_y" units="m/s">Vehicle velocity when measurement was taken in Y (NED)</field>
       <field type="float" name="vehicle_velocity_z" units="m/s">Vehicle velocity when measurement was taken in Z (NED)</field>
     </message>
-    <message id="391" name="DVL">
+    <message id="392" name="DVL">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
       <description>Doppler Velocity Log Sensor. Used to provide vehicle velocity </description>


message_definitions/v1.0/common.xml
```diff
@@ -7143,6 +7143,28 @@
...
+      <field type="float" name="vehicle_velocity_x" units="m/s">Vehicle velocity when measurement was taken in X (NED)</field>
+      <field type="float" name="vehicle_velocity_y" units="m/s">Vehicle velocity when measurement was taken in Y (NED)</field>
+      <field type="float" name="vehicle_velocity_z" units="m/s">Vehicle velocity when measurement was taken in Z (NED)</field>
+    </message>
+    <message id="392" name="DVL">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Doppler Velocity Log Sensor. Used to provide vehicle velocity </description>
+      <field type="uint8_t" name="has_lock">Has bottom lock to the floor (1 = true)</field>
+      <field type="float[16]" name="altitude_water" units="m">Vehicle altitude to the floor of body of water for each beam</field>
```
hamishwillee, line 7163: I do not know what this means - i.e. what each float would be/what a beam is. Is there are link we could use?

message_definitions/v1.0/common.xml
```diff
@@ -7143,6 +7143,28 @@
...
+      <field type="float" name="vehicle_velocity_x" units="m/s">Vehicle velocity in X (NED)</field>
```
hamishwillee, line 7164: What resolution do you need? Generally we try reduce the traffic, so this might be better as an int16_t with units cm/s (and similar for other cases)

hamishwillee:
Thanks very much. Generally we now push messages into development.xml first. When there is a prototype in at least one autopilot and broad intent to take in another we then push into common.xml. 

Are you working with a particular flight stack/do you have a timeframe for implementation and prototying?

If there are a bunch of other sensors you are thinking of adding it might be good to discuss them all together, whether we add all of them or not. Do you have any others you're interested in?

@auturgy @julianoes Anyone you can suggest we get to review these messages?

auturgy:
I’m going to have a look at them and perhaps a chat with Seb also.

hamishwillee:
Thanks. Do we have a git handle for Seb?

auturgy:
This is his PR

hamishwillee:
I don't approve of people who don't use their names as a github handle :-)

message_definitions/v1.0/common.xml
```diff
@@ -7143,6 +7143,28 @@
...
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Measured water current values. Note: Actual Water current = (Vehicle Velocity - Measured Current) </description>
+      <field type="float" name="measured_current_x" units="m/s">Measured water current in X (NED) direction</field>
+      <field type="float" name="measured_current_y" units="m/s">Measured water current in Y (NED) direction</field>
+      <field type="float" name="measured_current_z" units="m/s">Measured water current in Z (NED) direction</field>
+      <field type="float" name="distance" units="m">Measurement distance from the sensor</field>
+      <field type="float" name="vehicle_velocity_x" units="m/s">Vehicle velocity when measurement was taken in X (NED)</field>
+      <field type="float" name="vehicle_velocity_y" units="m/s">Vehicle velocity when measurement was taken in Y (NED)</field>
+      <field type="float" name="vehicle_velocity_z" units="m/s">Vehicle velocity when measurement was taken in Z (NED)</field>
+    </message>
```
auturgy, line 7157: Whilst ADCP will provide velocity over seabed, we're calling this a WATER_CURRENT message, not an ADCP message, so I'm not convinced that the vessel velocities belong.  Perhaps timestamp instead to enable correlation?

message_definitions/v1.0/common.xml
auturgy: pretty sure we can squeeze altitude_water as cm into the float...

message_definitions/v1.0/common.xml
auturgy: int16 cm/s will be fine here. 32000 cm/s ~ 640kts, which is about twice the water speed record.  If @bazfp can think of a reason we'd need more than that, happy to chat.

auturgy:
A few things, but basically we need to have a chat about this to clarify the immediate use case, and determine whether (or how much) we generalise this for relevant future applications.
1.  We should consider adding the sensors to the mav_sys_status_sensor enum so that they can be reported via SYS_STATUS
2.  Depending how the sensors are integrated, if they are their own mavlink component (or virtual component), we should consider a MAV_TYPE for their heartbeat.
3. Are we using these messages to provide data to the autopilot (ie the mission computer has the driver, and is feeding via mavlink, or the sensor is mavlink native), or to stream data from the autopilot (sensor driver is in the autopilot code) to a mission computer or GCS?  The two approaches need slightly different fields, but it should be possible to cover both.
4. Need instance numbers unless the sensor has a component ID.  We can't assume a vessel will only have one DVL or ADCP.
5. Orientation could either be in a message or in the vehicle config (ie parameters), but unless there is a guarantee that the consumer will know the orientation based on instance number or component id, i'd put it in a message.  Obviously it's a one shot thing though, so we don't need to be streaming it (although DISTANCE_SENSOR does).  Need to discuss.
6. Sensor location same (xyz offset).
7. Need to be clearer on frame.

auturgy:
@stephendade fyi

message_definitions/v1.0/common.xml
Williangalvani: @hamishwillee [this](https://www.waterlinked.com/dvl#hs_cos_wrapper_widget_1628665701042_) is a superficial explanation of how DVLs work

message_definitions/v1.0/common.xml
Williangalvani: DVLs are pretty precise in my experience, Waterlinked's DVL A50 claims to have a precision of 0.1mm/s. This will likely matter for station-keeping.

I also think that bandwidth is less of an issue for vehicles that could use a DVL. Most subs have high bandwidth links.
It could be important for boats, but  I worry about loosing resolution (plus boats are likely to just use a regular gps instead).

message_definitions/v1.0/common.xml
```diff
@@ -7143,6 +7143,28 @@
...
+    <message id="392" name="DVL">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Doppler Velocity Log Sensor. Used to provide vehicle velocity </description>
+      <field type="uint8_t" name="has_lock">Has bottom lock to the floor (1 = true)</field>
```
Williangalvani, line 7162: maybe add a "confidence" field?

hamishwillee:
@auturgy All good points. Minor pedantic points.
1. Need instance numbers **even IF** the sensor has a component ID. We use widely and historically component ids as identifiers, but we shouldn't - they are not fixed to a particular purpose and should be considered as dynamic allocations.
1. We should consider a MAV_TYPE for their heartbeat if it can EVER be an independent mavlink component.
1. FYI only updated your comment to refer to [MAV_SYS_STATUS_SENSOR_EXTENDED](https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_SENSOR_EXTENDED)

@Williangalvani Thanks for joining. I feel stupid for not inviting you directly.

message_definitions/v1.0/common.xml
bazfp: I agree, I originally set out to do a `WATER_CURRENT` but realise it's a somewhat useless metric by itself without vehicle velocity (or someway to retrieve accurate vehicle velocity, which a DVL provides)

message_definitions/v1.0/common.xml
hamishwillee: Thanks @Williangalvani ! More than enough for me.

message_definitions/v1.0/common.xml
hamishwillee: Not my call, just asking the questions.

hamishwillee:
Thanks for coming to the Mavlink call @bazfp . Looking forward to your proposals - ground, sea, and underwater vehicles are all somewhat neglected.

There are some comments inline. As discussed, better to push these into development.xml.

message_definitions/v1.0/common.xml
bazfp: How do you feel if `vehicle_velocity_x` is removed and `corrected_current_x/y/z` values are added that takes into account vehicle velocity?

hamishwillee: @auturgy ^^^^

ES-Alexander:
Late to the party, but noting that ArduPilotMega has a [`WATER_DEPTH`](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/ardupilotmega.xml#L1996) message for managing echosounder/rangefinder measurements in a way that's useful for generating bathymetry maps and the like. It seems to fit in with these to some extent, and is already implemented in ArduRover for boats, with [ArduSub integration on the way](https://github.com/ArduPilot/ardupilot/pull/31622).

Not sure if this is the best place to raise it, but other marine-relevant data that is currently not included in MAVLink would be water temperature and density / speed of sound, the latter of which is quite relevant for the accuracy of sonar distance measurements and profiling, and (in density form) could adjust the sophistication ceiling of dynamic buoyancy control mechanisms for submarines and towfishes and the like.

hamishwillee:
Thanks @ES-Alexander - we appear to have lost our stakeholder on this discussion three years ago. @bazfp are you still interested in this?

If this progresses it should certainly consider `WATER_DEPTH`. But if it does not continue we woudn't look at this topic any more. ... unless someone felt strongly that `WATER_DEPTH` should be used on PX4 too (say) or a message with some overlap was proposed.

Hope that makes sense? I'd like to standardize on marine sensors and messages, but there isn't any impetus if only one stack is interested and the other is happy with its messages.

ES-Alexander:
@hamishwillee I only found this because @clydemcqueen has been investigating performance improvements for sea-floor surface tracking, from considering individual beams of a DVL rather than just the DVL's velocity estimate. The group he's working with are using ArduPilot, so I suppose if the PX4 side has lost interest in standardisation we should pursue adding this kind of message via the ArduPilotMega message set instead?

Or do you think it makes sense to add via development so it's available if and when PX4 (or some other stack?) want to use it?

bazfp:
We actually have a much heftier set of changes to the MAVlink spec (mavlinkmarine) we have been using since this was put up and are looking to get upstreamed.

Currently looking at resourcing over the next few months to see if we can dedicate some time to getting this across the line.

ES-Alexander:
> We actually have a much heftier set of changes to the MAVlink spec (mavlinkmarine) ...

From some snooping - [here's a link](https://github.com/mavlinkmarine/mavlink/blob/master/message_definitions/v1.0/marine.xml) for the interested :-)

hamishwillee:
> > We actually have a much heftier set of changes to the MAVlink spec (mavlinkmarine) ...
> 
> From some snooping - [here's a link](https://github.com/mavlinkmarine/mavlink/blob/master/message_definitions/v1.0/marine.xml) for the interested :-)

Cool. I'm off until next week but on quick snoop that appears to mix payload, companion computer control, and water sensor stuff. Would great to split these so they can be discussed separately. The sensor stuff should definitely be discussed as part of this.

The rest we'll have to see. Everyone has an opinion on payload and companion stuff. It's a little harder for me to work with because standardization rules are a little more complicated - i.e.  a companion interface doesn't necessarily appear in a flight stack, so "in a flight stack and strong interest from another flight stack" isn't relevant.

hamishwillee:
Closing this. Stakeholder has left the building.

PR #1807, opened 2022-03-01
camera/component information: URI zero termination

julianoes:
This way it's much easier to deal with these message as the last char can just be set to '\0' before handing it. Otherwise the string needs to be copied first.


bkueng (approved):

hamishwillee:
@julianoes Would there be any implications to existing implementations? My thought was no, but want to make sure before I merge.

hamishwillee:
Merging. As discussed in dev call, only implication is if a particular camera uses the whole URL, and that is unlikely. Otherwise better for all.

PR #1808, opened 2022-03-03
VTOL - define MAV_TYPE for belly-sitting non tiltrotor

hamishwillee:
There is no MAV_TYPE for the vehicle known as a quadplane or standard vtol - a vehicle that takes off and lands on its belly, which has fixed separate rotors for vertical flight.  Commonly `MAV_TYPE_VTOL_RESERVED2` has been used for this.

This PR un-reserves `MAV_TYPE_VTOL_RESERVED2` (22) replacing with:
>  `MAV_TYPE_VTOL_FIXEDROTOR` - VTOL with separate fixed rotors for hover flight. Takes off and lands on belly.

While doing this I realised that naming is pretty adhoc, so I propose a naming convention: 
- `MAV_TYPE_VTOL_<TYPE>[_ROTORS]`, where:
  - `TYPE`: TAILSITTER | FIXEDROTOR | TILTROTOR:
  - `ROTORS` is string for numbers of "propulsion units" - rotors, fans, whatever.
    This is optional - my assumption is that a GCS doesn't do anything special based on the number of rotors. If it did you define a more specific type by adding a string like "DUOROTOR"

Due to this I have also replaced `MAV_TYPE_VTOL_RESERVED3` (23) with:
> `MAV_TYPE_VTOL_TAILSITTER` - Tailsitter VTOL. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.

Assumptions/Questions:
- For rotors I have assumed a string like "DUOROTOR" is better than "DUO" - so `MAV_TYPE_VTOL_TAILSITTER_DUOROTOR` 
- The main thing about the VTOL types is that it tells a GCS how to orient UI elements/horizon in the different states. So "really" you only need to know TAIL or BELLY orientation. BUT this is more scalable and in line with current types.
- There are two existing frames `VTOL MAV_TYPE_VTOL_DUOROTOR` and `MAV_TYPE_VTOL_QUADROTOR` that should be named `MAV_TYPE_VTOL_TAILSITTER_QUAD`/`MAV_TYPE_VTOL_TAILSITTER_DUO` or  `MAV_TYPE_VTOL_TAILSITTER_QUAD`/`MAV_TYPE_VTOL_TAILSITTER_QUADROTOR`. I assume that it would break too much to change them.

@meee1 @DonLakeFlyer As GCS authors, does this make sense to you? My assumption is basically GSC uses the MAV_TYPE for GCS layout based on the vehicle, and primarily the thing of interest is the orientation of the vehicle. 
The change for you guys would be configuring the UI on MAV_TYPE_VTOL_FIXEDROTOR and accepting MAV_TYPE_VTOL_TAILSITTER as a proxy for either or MAV_TYPE_VTOL_TAILSITTER_DUO or MAV_TYPE_VTOL_QUADROTOR.

FYI @sfuhrer @auturgy @julianoes

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -136,15 +136,15 @@
       <entry value="21" name="MAV_TYPE_VTOL_TILTROTOR">
         <description>Tiltrotor VTOL. Fuselage and wings stay (nominally) horizontal in all flight phases. It able to tilt (some) rotors to provide thrust in cruise flight.</description>
       </entry>
-      <!-- Entries up to 25 reserved for other VTOL airframes -->
-      <entry value="22" name="MAV_TYPE_VTOL_RESERVED2">
-        <description>VTOL reserved 2</description>
-      </entry>
-      <entry value="23" name="MAV_TYPE_VTOL_FIXEDROTOR">
+      <entry value="22" name="MAV_TYPE_VTOL_FIXEDROTOR">
         <description>VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
       </entry>
-      <entry value="24" name="MAV_TYPE_VTOL_TAILSITTER">
+      <entry value="23" name="MAV_TYPE_VTOL_TAILSITTER">
         <description>Tailsitter VTOL. Fuselage and wings orientation changes depending on flight phase: vertical for hover, horizontal for cruise. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
+      </entry>
+      <!-- Entries 24/25 reserved for other VTOL airframe -->
+      <entry value="24" name="MAV_TYPE_VTOL_RESERVED4">
+        <description>VTOL reserved 4</description>
       </entry>
       <entry value="25" name="MAV_TYPE_VTOL_RESERVED5">
         <description>VTOL reserved 5</description>
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -128,20 +128,20 @@
         <description>Onboard companion controller</description>
       </entry>
       <entry value="19" name="MAV_TYPE_VTOL_DUOROTOR">
-        <description>Two-rotor VTOL using control surfaces in vertical operation in addition. Tailsitter.</description>
+        <description>Two-rotor Tailsitter VTOL that additionally uses control surfaces in vertical operation. Note: should be named MAV_TYPE_VTOL_TAILSITTER_DUOROTOR.</description>
       </entry>
       <entry value="20" name="MAV_TYPE_VTOL_QUADROTOR">
-        <description>Quad-rotor VTOL using a V-shaped quad config in vertical operation. Tailsitter.</description>
+        <description>Quad-rotor Tailsitter VTOL using a V-shaped quad config in vertical operation. Note: should be named MAV_TYPE_VTOL_TAILSITTER_QUADROTOR.</description>
       </entry>
       <entry value="21" name="MAV_TYPE_VTOL_TILTROTOR">
-        <description>Tiltrotor VTOL</description>
+        <description>Tiltrotor VTOL. Takes off and lands on belly.</description>
       </entry>
       <!-- Entries up to 25 reserved for other VTOL airframes -->
-      <entry value="22" name="MAV_TYPE_VTOL_RESERVED2">
-        <description>VTOL reserved 2</description>
+      <entry value="22" name="MAV_TYPE_VTOL_FIXEDROTOR">
+        <description>VTOL with separate fixed rotors for hover flight. Takes off and lands on belly.</description>
       </entry>
-      <entry value="23" name="MAV_TYPE_VTOL_RESERVED3">
-        <description>VTOL reserved 3</description>
+      <entry value="23" name="MAV_TYPE_VTOL_TAILSITTER">
+        <description>Tailsitter VTOL. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
       </entry>
       <entry value="24" name="MAV_TYPE_VTOL_RESERVED4">
         <description>VTOL reserved 4</description>
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -137,8 +137,8 @@
         <description>Tiltrotor VTOL. Fuselage and wings stay (nominally) horizontal in all flight phases. It able to tilt (some) rotors to provide thrust in cruise flight.</description>
       </entry>
       <!-- Entries up to 25 reserved for other VTOL airframes -->
-      <entry value="22" name="MAV_TYPE_VTOL_FIXEDROTOR">
-        <description>VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
+      <entry value="22" name="MAV_TYPE_VTOL_RESERVED2">
+        <description>MAV_TYPE_VTOL_FIXEDROTOR: VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
       </entry>
       <entry value="23" name="MAV_TYPE_VTOL_TAILSITTER">
         <description>Tailsitter VTOL. Fuselage and wings orientation changes depending on flight phase: vertical for hover, horizontal for cruise. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -134,14 +134,14 @@
         <description>Quad-rotor Tailsitter VTOL using a V-shaped quad config in vertical operation. Note: should be named MAV_TYPE_VTOL_TAILSITTER_QUADROTOR.</description>
       </entry>
       <entry value="21" name="MAV_TYPE_VTOL_TILTROTOR">
-        <description>Tiltrotor VTOL. Takes off and lands on belly.</description>
+        <description>Tiltrotor VTOL. Fuselage and wings stay (nominally) horizontal in all flight phases. It able to tilt (some) rotors to provide thrust in cruise flight.</description>
       </entry>
       <!-- Entries up to 25 reserved for other VTOL airframes -->
       <entry value="22" name="MAV_TYPE_VTOL_FIXEDROTOR">
-        <description>VTOL with separate fixed rotors for hover flight. Takes off and lands on belly.</description>
+        <description>VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
       </entry>
       <entry value="23" name="MAV_TYPE_VTOL_TAILSITTER">
-        <description>Tailsitter VTOL. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
+        <description>Tailsitter VTOL. Fuselage and wings orientation changes depending on flight phase: vertical for hover, horizontal for cruise. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
       </entry>
       <entry value="24" name="MAV_TYPE_VTOL_RESERVED4">
         <description>VTOL reserved 4</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/minimal.xml
```diff
@@ -128,20 +128,20 @@
...
-        <description>Two-rotor VTOL using control surfaces in vertical operation in addition. Tailsitter.</description>
+        <description>Two-rotor Tailsitter VTOL that additionally uses control surfaces in vertical operation. Note: should be named MAV_TYPE_VTOL_TAILSITTER_DUOROTOR.</description>
       </entry>
       <entry value="20" name="MAV_TYPE_VTOL_QUADROTOR">
-        <description>Quad-rotor VTOL using a V-shaped quad config in vertical operation. Tailsitter.</description>
+        <description>Quad-rotor Tailsitter VTOL using a V-shaped quad config in vertical operation. Note: should be named MAV_TYPE_VTOL_TAILSITTER_QUADROTOR.</description>
       </entry>
       <entry value="21" name="MAV_TYPE_VTOL_TILTROTOR">
-        <description>Tiltrotor VTOL</description>
+        <description>Tiltrotor VTOL. Takes off and lands on belly.</description>
```
sfuhrer, line 137: When I read "belly landing" a fixed-wing style landing comes to my mind.
How about:
```suggestion
        <description>Tiltrotor VTOL. Fuselage and wings stay horizontal and is able to tilt (some) rotors to provide thrust in cruise flight.</description>
```

message_definitions/v1.0/minimal.xml
```diff
@@ -128,20 +128,20 @@
...
       </entry>
       <!-- Entries up to 25 reserved for other VTOL airframes -->
-      <entry value="22" name="MAV_TYPE_VTOL_RESERVED2">
-        <description>VTOL reserved 2</description>
+      <entry value="22" name="MAV_TYPE_VTOL_FIXEDROTOR">
+        <description>VTOL with separate fixed rotors for hover flight. Takes off and lands on belly.</description>
```
sfuhrer, line 141:
```suggestion
        <description>VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay horizontal in all flight phases.</description>
```

message_definitions/v1.0/minimal.xml
```diff
@@ -128,20 +128,20 @@
...
       </entry>
-      <entry value="23" name="MAV_TYPE_VTOL_RESERVED3">
-        <description>VTOL reserved 3</description>
+      <entry value="23" name="MAV_TYPE_VTOL_TAILSITTER">
+        <description>Tailsitter VTOL. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
```
sfuhrer, line 144:
```suggestion
        <description>Tailsitter VTOL. Fuselage and wings orientation changes depending on flight phase: vertical for hover, horizontal for cruise. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
```

sfuhrer (commented):
Thanks for taking care of that @hamishwillee ! I have some proposals on how we define the separate types.

DonLakeFlyer:
QGC doesn't do anything really special with these VTOL types other than use them to identify the vehicle as a generic vtol. QGC already uses the reserved values for this determination as well. So changed a reserved value name to a different name but keeping the same value means QGC will continue to work as expected.

message_definitions/v1.0/minimal.xml
hamishwillee: @sfuhrer I added the word "nominally" because of course the vehicle isn't horizontal if it chooses to pitch or roll. 
```suggestion
        <description>VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
```

message_definitions/v1.0/minimal.xml
hamishwillee: @sfuhrer Above I modified to add the "nominally" because the vehicle has a natural horizontal position, but it doesn't "stay" perfectly horizontal.

message_definitions/v1.0/minimal.xml
```diff
@@ -128,20 +128,20 @@
...
       </entry>
       <entry value="21" name="MAV_TYPE_VTOL_TILTROTOR">
-        <description>Tiltrotor VTOL</description>
+        <description>Tiltrotor VTOL. Fuselage and wings stay (nominally) horizontal in all flight phases. It able to tilt (some) rotors to provide thrust in cruise flight.</description>
       </entry>
       <!-- Entries up to 25 reserved for other VTOL airframes -->
-      <entry value="22" name="MAV_TYPE_VTOL_RESERVED2">
-        <description>VTOL reserved 2</description>
+      <entry value="22" name="MAV_TYPE_VTOL_FIXEDROTOR">
+        <description>VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
```
hamishwillee, line 141: @sfuhrer For _now_, I have reverted the name back to MAV_TYPE_VTOL_RESERVED2, because I don't want to break the behaviour for things that are using this value - see https://github.com/mavlink/mavlink/pull/1808#issuecomment-1058245358
```suggestion
      <entry value="22" name="MAV_TYPE_VTOL_RESERVED2">
        <description>MAV_TYPE_VTOL_FIXEDROTOR: VTOL with separate fixed rotors for hover and cruise flight. Fuselage and wings stay (nominally) horizontal in all flight phases.</description>
```

hamishwillee:
Re https://github.com/mavlink/mavlink/pull/1808#issuecomment-1058245358

> So changed a reserved value name to a different name but keeping the same value means QGC will continue to work as expected.

@DonLakeFlyer ( and @sfuhrer) **No one should be using anything named "Reserved" for anything.** MAVLink assumes these are free for us to break that at will. 

Can we co-ordinate merging of this (now updated PR) so that QGC:
- treats the new values `MAV_TYPE_VTOL_FIXEDROTOR` and  `MAV_TYPE_VTOL_TAILSITTER` as VTOL (replacing `MAV_TYPE_VTOL_RESERVED3` and `MAV_TYPE_VTOL_RESERVED4`)
- stops treating `MAV_TYPE_VTOL_RESERVED5` as VTOL (so that there is no reason for people to use it).
- don't do anything with ` MAV_TYPE_VTOL_RESERVED2` (which I have reverted)

The theory behind this is that 
- `MAV_TYPE_VTOL_RESERVED2` (22) is commonly used for quadplane/standardplane (what we are now calling "MAV_TYPE_VTOL_FIXEDROTOR"). With the above change those users will not be broken, but over time will come to use MAV_TYPE_VTOL_FIXEDROTOR.
- no one is using the other reserved cases except QGC, so we can break them and move on.

Make sense?

message_definitions/v1.0/minimal.xml
```diff
@@ -128,23 +128,23 @@
         <description>Onboard companion controller</description>
       </entry>
       <entry value="19" name="MAV_TYPE_VTOL_DUOROTOR">
-        <description>Two-rotor VTOL using control surfaces in vertical operation in addition. Tailsitter.</description>
+        <description>Two-rotor Tailsitter VTOL that additionally uses control surfaces in vertical operation. Note: should be named MAV_TYPE_VTOL_TAILSITTER_DUOROTOR.</description>
```
sfuhrer, line 131: Would we break anything by renaming it directly instead of just remarking that it should be renamed?

DonLakeFlyer:
"No one should be using anything named "Reserved" for anything. MAVLink assumes these are free for us to break whenever we want."

I would tend to disagree with that. For example the changes proposed to move some reserved values to an actual named VTOL type means that current QGC code (which references these reserved types) will work correctly with never firmwares without needing to be updated. It provides a good path to forward compat. I would assume RESERVED_VTOL means that this value although not currently a named VTOL type will still end up being a VTOL something at some point. It's not going to change to be a non-VTOL value. If that is the case why was it reserved for VTOL in the first place.

hamishwillee:
@DonLakeFlyer Re https://github.com/mavlink/mavlink/pull/1808#issuecomment-1063146796

No no no no no :-)

You can't rely on those values marked as reserved for VTOL being reserved for VTOL. They are only reserved so we have a convenient slot to put new VTOL types adjacent to the others - i.e. a readability tool. If the mavlink team need to use them for something else we will, without considering GCS.

If you provide any support for a reserved value doing anything at all then we need to start considering who might be broken. Case in point - MAV_TYPE_VTOL_RESERVED2 cannot be renamed now until all the vehicles that are using it to mean "Standard VTOL" are updated or discontinued - 10 years?

If QGC had ignored reserved values then users would get no VTOL formatting if they chose that value - so there would be no reason for them to do so. They would be forced to come to MAVLink and asked for a VTOL type that does what they need. At that point you add specific support for the type and you know that it matches what the type does.
Best of all I would not need to talk to QGC, Mission Planner, AMC, etc etc to manage the change. You'd either support the type or not but you wouldn't be broken and have to do any updates.

message_definitions/v1.0/minimal.xml
hamishwillee: @sfuhrer Sadly yes. Once something has a fixed meaning then it is very hard to change,  because if you do then  any GCS that updates to the new value will not understand the messages from any vehicle that has not. We know that the update cycle for vehicles is of the order of 5 - 10 years across various flight stacks, so that's not viable.

That is why I am so adamant that we don't assign any meaning to reserved values in the other discussion https://github.com/mavlink/mavlink/pull/1808#issuecomment-1063518075 - because if we do then we can't rename them or do much else with them.

What we could do is create the new value as well and mark the old value as deprecated. I'm in two minds about that approach. On one hand it makes the pattern clear and means that "one day" you might be able to remove the inconsistent value. On the other hand the value is a bit ugly, and still has to exist until that "one day" comes. 
I lean towards this approach, but I'd be interested in what both you and @DonLakeFlyer think.

DonLakeFlyer: If the value remains the same for the new name. And the name MAV_TYPE_VTOL_DUOROTOR is always moved to the dustbin forever and never reused. Then there is no problem renaming. You can note the previous name in the test if you want. It will just cause a GCS to change names as well and recompile. This sort of thing happens all the time when new fields are added to a mavlink message. That causes the need for a code change and recompile as well. All of this continues to work in a forward compat without the code change as well.

DonLakeFlyer:
It seems to be that you are saying that this comment "   " is a load of crap. If that is the case then a GCS should not relay on these reserved values for any reason wahatsoever. If that comment is actually true then it is perfectly fine for a GCS to use all named and reserved values for VTOL as a deterministic mechanism to know when a vehicle is a VTOL class vehicle. This has nothing to do with knowing the actual type of the VTOL. It's just disambiguates from fixed wing, rover, multi-rotor and vtol as high level vehicle classes. That is something QGC needs to do.

hamishwillee:
> It seems to be that you are saying that this comment  is not true.

I'm saying the same thing as for `MAV_TYPE_VTOL_RESERVED2` - this comment is a placeholder so you don't fill in the space that I might want to use for VTOL airframes. Or I might want to use for something else. You can't possibly know, so don't do it.

Essentially we do this when we're too lazy to fill in the `MAV_TYPE_VTOL_RESERVEDn` values.

> If that is the case then a GCS should not rely on these reserved values for any reason whatsoever. 

Correct. Don't. 

>  It's just disambiguates from fixed wing, rover, multi-rotor and vtol as high level vehicle classes. That is something QGC needs to do.

No it doesn't. Because no one should be emitting those you don't need to disambiguate - the are nothing! What you're doing is creating a self fulfilling prophesy - you provide a meaning, so people use them.

In most cases it really won't make too much difference. In this case it has because I can't ever use `MAV_TYPE_VTOL_RESERVED2`

message_definitions/v1.0/minimal.xml
hamishwillee: Sounds OK to me, but will address in a separate PR because this **does** force a rebuilt for anyone using the item.

hamishwillee:
> I'll raise it in the MAVLink call, but I'm in no doubt of the result.

@DonLakeFlyer I was wrong - the dev call had a range of views. The summary is that we'll still change them as we need to and not feel the need to consult anyone. We recommend that you don't give them any meaning, but if you do then you need to manage that change. I suspect that means business as usual :-). I'm unconvinced, but my apologies for being so dogmatic.

I would like to change MAV_TYPE_VTOL_DUOROTOR et al to match the pattern. However I want to get more eyes on that because it does involve "work" for more than just QGC. Separate PR to follow.

Merging.

Just FYI only:
- @julianoes interprets the reserved values as you do, and sees no harm in anyone using them to represent a VTOL provided they are willing to understand that the values may well change and if they do if you use it you will have to do work.
- @auturgy Sees no particular problem in this case, but thinks the "law on unintended consequences" may apply - i.e. he wouldn't be surprised if applying any meaning to reserved values will end up stuffing someone. He thinks "on balance" it would probably be better to not give the reserved values meaning.



I have therefore shifted the ids back to fill up the first cases.

junwoo091400:
I believe this PR deleted the "MAV_TYPE_VTOL_RESERVED2". Which is causing issues when building in PX4 like here. Does this need fix in PX4 side @hamishwillee ?

image

hamishwillee:
@junwoo091400 Yes, this is the cause, I'll comment there.

PR #1812, opened 2022-03-08
common: remove option for relative speed change

julianoes:
This removes the functionality to do a speed change in a relative way (e.g. 5 m/s slower) rather than an absolute way (10 m/s).

This option in param4 had been [added in 2016](https://github.com/mavlink/mavlink/commit/320f616179d5088ce3a37147aa79a81d07ea848a) but neither implemented in PX4 nor ArduPilot from what I can tell.

While this option might be a good idea, there is - at this point - no safe option to add the functionality. If we were to start using the
option, then it might not work correctly if the autopilot does not actually check param4.

Additionally to the fact that we can't start using it, the API is also odd in that a negative speed (or at least -1) is generally ignored, so a relative change that is slower (negative) would likely not work either.

I think the safe bet is to reserve this to 0, and add a new command for relative changes, if required.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1313,10 +1313,10 @@
         <param index="1" label="Speed Type" minValue="0" maxValue="3" increment="1">Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)</param>
         <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change)</param>
         <param index="3" label="Throttle" units="%" minValue="-1">Throttle (-1 indicates no change)</param>
-        <param index="4">Reserved, no longer used, set to 0</param>
-        <param index="5">Empty</param>
-        <param index="6">Empty</param>
-        <param index="7">Empty</param>
+        <param index="4" reserved="true" default="0" />
+        <param index="5" reserved="true" default="0" />
+        <param index="6" reserved="true" default="0" />
+        <param index="7" reserved="true" default="0" />
       </entry>
       <entry value="179" name="MAV_CMD_DO_SET_HOME" hasLocation="true" isDestination="false">
         <description>Changes the home location either to the current location or a specified location.</description>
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
@julianoes Makes sense to me. 
- This change does not appear to have an associated PR so not clear where it was discussed or the use case it was designed to meet.
- Yes, I think it is unsafe - too easy to set your speed to something that will make a plane stall. 
- In theory it should not be unsafe with a system that does not understand it, because such a system should be rejecting the message if it has invalid values. But then I doubt this is enforced by anyone. I wonder if we could push that into the library generation "empty or reserved values" with non zero values automatically reject?
- I doubt anyone is using this. Good reason to have a release process though.

I won't merge though. Let's see if @auturgy has an opinion.

hamishwillee:
@dagar @LorenzMeier Any objection to reverting "relative do_speed" commands that you guys merged in 2016. They do not appear to be used anywhere, and look a bit "risky".

hamishwillee:
I pinged Dan offline. No response. Let's add to next dev call and hope Lorenz is present.

julianoes:
> This change does not appear to have an associated PR so not clear where it was discussed or the use case it was designed to meet.

There isn't. I was just going through MAVSDK checking for MAVLink compatibility as I was looking into pure mode.

> I wonder if we could push that into the library generation "empty or reserved values" with non zero values automatically reject?

We could add mechanisms for it but at this point it seems like an awful lot of work.

> Good reason to have a release process though.

Of course, MAVLink should have a version and/or release mechanism.

hamishwillee:
>> I wonder if we could push that into the library generation "empty or reserved values" with non zero values automatically reject?

> We could add mechanisms for it but at this point it seems like an awful lot of work.

Yeah. Something to think about for a test suite :-).

We'll look at this in dev call tonight.

hamishwillee:
[20220317-Dev-Meeting](https://github.com/mavlink/mavlink/wiki/20220317-Dev-Meeting) discussion agreed this is a good cleanup. If relative changes are needed then those should be done as a separate message.

@julianoes FYI only, I did change this to use `<param index="n" reserved="true" default="0"/>` for all the unused values. I intend to do this "globally" at some point soon. This does generate proper documentation - and I checked that it does same thing for Python and C.

PR #1818, opened 2022-03-16
MAV_TYPE_VTOL _DUOROTOR and _QUADROTOR renamed to new format

hamishwillee:
This PR update the names of the enum values:
 - `MAV_TYPE_VTOL_DUOROTOR` to `MAV_TYPE_VTOL_TAILSITTER_DUOROTOR`
- `MAV_TYPE_VTOL_QUADROTOR` to `MAV_TYPE_VTOL_TAILSITTER_QUADROTOR`

This update matches the values to the naming convention discussed in https://github.com/mavlink/mavlink/pull/1808#issue-1157921604 

The change follows on from #1808 and in particular the discussion https://github.com/mavlink/mavlink/pull/1808/files#r823318681

It is in a separate PR because anyone who uses these values will need to update their code on next build with new library. Note however that code that is not rebuilt will continue to work (the change is forward compatible.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -127,11 +127,11 @@
       <entry value="18" name="MAV_TYPE_ONBOARD_CONTROLLER">
         <description>Onboard companion controller</description>
       </entry>
-      <entry value="19" name="MAV_TYPE_VTOL_DUOROTOR">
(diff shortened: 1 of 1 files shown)


hamishwillee:
FYI @sfuhrer @auturgy @meee1 Don is happy to force all users of these enum values to update next time they do a rebuild (though in no way requiring them to do a rebuild). I think this is fine and I would like to do this.

Anyway this is an FYI to give you a chance to agree/disagree on behalf of our respective communities.

hamishwillee:
This has also been agreed/approved by @auturgy (offline). Merging. 

@meee1 @DonLakeFlyer Note the break on next compile.

PR #1822, opened 2022-03-22
Deprecate MAV_CMD_GET_MESSAGE_INTERVAL for generic MAV_CMD_REQUEST_MESSAGE

hamishwillee:
We're moving to the generic requester for most commands. This adds deprecation for this case too.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1861,8 +1861,10 @@
         <param index="2" label="RC Type" enum="RC_TYPE">RC type.</param>
       </entry>
       <entry value="510" name="MAV_CMD_GET_MESSAGE_INTERVAL" hasLocation="false" isDestination="false">
-        <deprecated since="2022-03" replaced_by="MAV_CMD_REQUEST_MESSAGE"/>
-        <description>Request the interval between messages for a particular MAVLink message ID. The receiver should ACK the command and then emit its response in a MESSAGE_INTERVAL message.</description>
+        <description>
+          Request the interval between messages for a particular MAVLink message ID.
+          The receiver should ACK the command and then emit its response in a MESSAGE_INTERVAL message.
+        </description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
       </entry>
       <entry value="511" name="MAV_CMD_SET_MESSAGE_INTERVAL" hasLocation="false" isDestination="false">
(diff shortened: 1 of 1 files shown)


julianoes (approved):

auturgy:
@peterbarker

peterbarker:
Uh.  This doesn't look right.

One requests the message.

The other requests the current message interval for that message.

So send a `MAV_CMD_GET_MESSAGE_INTERVAL` and you will get back a `MESSAGE_INTERVAL` message (and a COMMAND_ACK)

Send a `MAV_CMD_REQUEST_MESSAGE` message specifying 242 in P1 and you will receive a `HOME_POSITION` message (and a `COMMAND_ACK`).

hamishwillee:
@peterbarker Thank you! 
You're right, I omitted documenting how you specify the message id of the message for which the interval is required. Updated. 

So now

- Send `MAV_CMD_GET_MESSAGE_INTERVAL` with `param1=<message_id of message for which interval is required>`  and you will get back a `MESSAGE_INTERVAL` message for that message_id (and a COMMAND_ACK)
- Send `MAV_CMD_REQUEST_MESSAGE ` with `param1=244` (id of `MESSAGE_INTERVAL`) message and `param2=<message_id of message for which interval is required>` and you will get back a `MESSAGE_INTERVAL` message for that message_id (and a COMMAND_ACK)

message_definitions/v1.0/common.xml
```diff
@@ -6384,7 +6385,11 @@
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
     </message>
     <message id="244" name="MESSAGE_INTERVAL">
-      <description>The interval between messages for a particular MAVLink message ID. This message is the response to the MAV_CMD_GET_MESSAGE_INTERVAL command. This interface replaces DATA_STREAM.</description>
+      <description>244
```
hamishwillee, line 6388: Typo
```suggestion
      <description>
```

peterbarker:
> @peterbarker Thank you! You're right, I omitted documenting how you specify the message id of the message for which the interval is required. Updated.
> 
> So now
> 
>     * Send `MAV_CMD_GET_MESSAGE_INTERVAL` with `param1=<message_id of message for which interval is required>`  and you will get back a `MESSAGE_INTERVAL` message for that message_id (and a COMMAND_ACK)

Yep.

> 
>     * Send `MAV_CMD_REQUEST_MESSAGE ` with `param1=244` (id of `MESSAGE_INTERVAL`) message and `param2=<message_id of message for which interval is required>` and you will get back a `MESSAGE_INTERVAL` message for that message_id (and a COMMAND_ACK)

Hmmm...  That's not sometihng ArduPilot would currently do.  ArduPilot doesn't currently honour param2.  It wasn't around when  I first implemented `MAV_CMD_REQUEST_MESSAGE`... not saying it isn't a good addition, just that we don't support it.  A quick glance shows that it won't actually be all that nice to implement, either, as actually sending the message is done aynchronously so we'd need to store `param2` somewhere.... and now you've got the problem of servicing multiple `MAV_CMD_REQUEST_MESSAGE`s at the same time, so you may need to temporarily reject requests until you've serviced the previous one :-/

We do, of course, handle `MAV_CMD_GET_MESSAGE_INTERVAL` just fine.

peterbarker:
On Wed, 23 Mar 2022, Hamish Willee wrote:

> What do you think we should do about this?

Don't deprecate the original mechanism for now.

At some stage we'll find some way of doing that param2 instance-id thing.

hamishwillee:
OK, deprecation reverted. So now this basically updates `MESSAGE_INTERVAL` to state that either method can be used to retrieve it. Of course that depends on what is implemented, and now we have two ways of doing the same thing :-(

@peterbarker, OK to merge?

peterbarker (approved):
LGTM - thanks @hamishwillee

PR #1823, opened 2022-03-23
add COMPONENT_METADATA, deprecate COMPONENT_INFORMATION

bkueng:
- the peripherals is added to the general json with optional CRC, and can use the (future) dynamic update mechanism, this cleaning up the message a bit
- makes the CRC optional in general
- removes the translation CRC
- rename from COMPONENT_INFORMATION to COMPONENT_METADATA
- keeping the existing COMPONENT_INFORMATION for a transition period

I think this is all that we discussed, @auturgy @julianoes @hamishwillee , let me know if I missed something.


component_metadata/general.schema.json
```diff
@@ -43,7 +43,7 @@
                         "type":         "string"
```
hamishwillee, line 43: Should this be mandated to https? We agreed that was what we wanted, but not sure all implications were discussed.

hamishwillee: Ditto in other instances mentioned below.

message_definitions/v1.0/common.xml
```diff
@@ -7063,23 +7063,32 @@
...
-      <field type="uint32_t" name="peripherals_metadata_file_crc">CRC32 of peripherals metadata file (peripherals_metadata_uri).</field>
-      <field type="char[100]" name="peripherals_metadata_uri">(Optional) MAVLink FTP URI for the peripherals metadata file (COMP_METADATA_TYPE_PERIPHERALS), which may be compressed with xz. This contains data about "attached components" such as UAVCAN nodes. The peripherals are in a separate file because the information must be generated dynamically at runtime. The string needs to be zero terminated.</field>
+      <field type="uint32_t" name="file_crc">CRC32 of the general metadata file.</field>
+      <field type="char[100]" name="uri">MAVLink FTP URI for the general metadata file (COMP_METADATA_TYPE_GENERAL), which may be compressed with xz. The file contains general component metadata, and may contain URI links for additional metadata (see COMP_METADATA_TYPE). The information is static from boot, and may be generated at compile time. The string needs to be zero terminated.</field>
```
hamishwillee, line 7091: Last chance to decide you might want more than 100 chars?

hamishwillee:
Looks great to me. I made these minor changes https://github.com/mavlink/mavlink/pull/1823/commits/13a554e41722f153af05713e62cb40e7c2e49690 and a couple of questions about HTTP / HTTPS inline.

component_metadata/general.schema.json
bkueng: Yes let's change that as well then.

message_definitions/v1.0/common.xml
bkueng: I was considering it but as it has to come from the vehicle, it's plenty. You could also use an alias.

component_metadata/general.schema.json
```diff
@@ -39,11 +39,11 @@
                         "type":         "integer"
                     },
                     "uri": {
-                        "description":  "http[s] or MAVLink FTP uri",
+                        "description":  "https or MAVLink FTP uri",
                         "type":         "string"
                     },
                     "fileCrc": {
-                        "description":  "CRC32 checksum of the file",
+                        "description":  "CRC32 checksum of the file. This may not present for dynamically updated metadata.",
```
hamishwillee, line 46:
```suggestion
                        "description":  "CRC32 checksum of the file. This should not be set for dynamically updated metadata.",
```

component_metadata/general.schema.json
```diff
@@ -55,23 +55,15 @@
                         "type":         "integer"
                     },
                     "translationUri": {
-                        "description":  "http[s] or MAVLink FTP url",
+                        "description":  "https or MAVLink FTP url",
```
hamishwillee, line 58: FWIW probably will always be https - if you make this mavlink you're hard coding the translation set until you next rebuild. I guess it is possible.

message_definitions/v1.0/common.xml
```diff
@@ -7063,23 +7063,32 @@
       <field type="uint32_t[6]" name="link_rx_max" units="KiB/s" invalid="[UINT32_MAX]">Network capacity to the component system. A value of UINT32_MAX implies the field is unused.</field>
     </message>
     <message id="395" name="COMPONENT_INFORMATION">
+      <deprecated since="2022-03" replaced_by="COMPONENT_METADATA"/>
```
hamishwillee, line 7066:
```suggestion
      <deprecated since="2022-04" replaced_by="COMPONENT_METADATA"/>
```

hamishwillee:
@auturgy @julianoes Do you want to sanity check this? If not, I plan to merge this later today.

I have verified that it meets the requirements for the API changes agreed in the dev call:
- COMPONENT_INFORMATION is deprecated, replaced by wip COMPONENT_METADATA that just has the general metadata url and file crc
- CRCs have been made optional, and removed where can never be useful (translation). 

@bkueng I assume you are planning to add the dynamic update message separately. That would need to happen prior to us removing WIP.

component_metadata/general.schema.json
```diff
@@ -39,11 +39,11 @@
                         "type":         "integer"
                     },
                     "uri": {
-                        "description":  "http[s] or MAVLink FTP uri",
+                        "description":  "https or MAVLink FTP uri",
                         "type":         "string"
                     },
                     "fileCrc": {
-                        "description":  "CRC32 checksum of the file",
+                        "description":  "CRC32 checksum of the file. This should not be set for dynamically updated metadata.",
```
julianoes, line 46: So, if thse files doesn't have a CRC anymore, then we can't cache them between flights or when restarting a ground station or rebooting a drone, or can we?

hamishwillee: @julianoes Let's see what @bkueng says, but my take on this...

Yes, but you don't need to/not with these messages.

Things to consider
- this is the general metadata file - it is statically built and stored in firmware. It can only update when the firmware is rebuilt. 
- Anything this contains that has a CRC must conform to that same rule. That is true for normal parameters etc.

Upshot, in the normal case you'll download this first time and store the CRC, then download all the contained files that have CRCs and cache their CRCs. On next connection you will not have to download again. In fact you won't download anything else from here until the firmware changes, because the CRC is fixed.

So what about the ones that don't have CRC? These are dynamic - for example "parameters dynamic". It is correct that there is nothing in the file to let you cache them.

The idea is that the component will stream (at an appropriate rate) a message with CRC and file type for the dynamic file types. So on startup you'd wait until you saw this message, check the CRC and download if needed.

Hope that makes sense.

bkueng: Yes, if no crc => don't download.
The general file could be on the SD card and updated on boot, but it's not relevant for a client.
And independently it could be dynamically updated. I'm not sure if this should be allowed though.

hamishwillee: > And independently it could be dynamically updated. I'm not sure if this should be allowed though.

I think that would make things challenging. For a start, the file CRC might be hard coded - in which case you'd have to build the firmware anyway. Not sure it would be necessary either - what change might you want in the SDK card to static behaviour that would not also require a firmware rebuild?

julianoes: Sorry, I misread the diff. I can see that the CRC for static data is still included, so this would change the general CRC32 and hence correctly trigger a re-download. That's all good then. Thanks for explaining that to me.

bkueng: > I think that would make things challenging. For a start, the file CRC might be hard coded - in which case you'd have to build the firmware anyway. Not sure it would be necessary either - what change might you want in the SDK card to static behaviour that would not also require a firmware rebuild?

I also tend towards not supporting it. The use-case would be changing url's. For example when updating the metadata on a companion. There's also other ways to achieve that though.

hamishwillee: Your call. But we discussed in the dev call and we're happy to merge this.

PR #1824, opened 2022-03-24
AVSUAS: correct MODE_M300_FORCE_AUTO_LANDING enum entry

peterbarker:
This space breaks pymavlink generation


hamishwillee (approved):
@peterbarker Thanks, I'm going to immediately merge this as a no-brainer/non breaking change.

I will also look to add some checking in the XSD.
However I can't see any reason for this to break the toolchain, so can you automatically strip whitespace from either side of parsed-in names?

hamishwillee:
Done in https://github.com/ArduPilot/pymavlink/pull/652

PR #1826, opened 2022-03-29
Option to not reset non-configurable params in preflight storage

ThomasDebrunner:
[MAV_CMD_PREFLIGHT_STORAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_PREFLIGHT_STORAGE) only has an option to reset all parameters. While some parameters hold "config" values, other params may contain values that are not meant to be configured manually. 
For example in PX4, there is  "LND_FLIGHT_T_" which containst the accumulated flight time of the system, or "COM_FLIGHT_UUID" which is a monotonically increasing flight number for a specific vehicle. These values are set by the system and are not meant for a user to configure.

This PR changes the meaning of the existing command (2) to only reset "config" values, while adding a new option (4) to completely wipe everything. 

PX4 implementation: https://github.com/PX4/PX4-Autopilot/pull/19409

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -985,6 +985,42 @@
         <description>Autotune yaw axis.</description>
       </entry>
     </enum>
+    <enum name="PREFLIGHT_STORAGE_PARAMETER_ACTION">
+      <description>
+        Actions for reading/writing parameters between persistent and volatile storage when using MAV_CMD_PREFLIGHT_STORAGE.
+        (Commonly parameters are loaded from persistent storage (flash/EEPROM) into volatile storage (RAM) on startup and written back when they are changed.)
+      </description>
+      <entry value="0" name="PARAM_READ_PERSISTENT">
+        <description>Read all parameters from persistent storage. Replaces values in volatile storage.</description>
+      </entry>
+      <entry value="1" name="PARAM_WRITE_PERSISTENT">
+        <description>Write all parameter values to persistent storage (flash/EEPROM)</description>
+      </entry>
+      <entry value="2" name="PARAM_RESET_CONFIG_DEFAULT">
+        <description>Reset all user configurable parameters to their default value. Does not reset values that contain operation counters and vehicle computed statistics.</description>
+      </entry>
+      <entry value="3" name="PARAM_RESET_SENSOR_DEFAULT">
+        <description>Reset only sensor calibration parameters to factory defaults (or firmware default if not available)</description>
+      </entry>
+      <entry value="4" name="PARAM_RESET_ALL_DEFAULT">
+        <description>Reset all parameters, including operation counters, to default values</description>
+      </entry>
+    </enum>
+    <enum name="PREFLIGHT_STORAGE_MISSION_ACTION">
+      <description>
+        Actions for reading and writing plan information (mission, rally points, geofence) between persistent and volatile storage when using MAV_CMD_PREFLIGHT_STORAGE.
+        (Commonly missions are loaded from persistent storage (flash/EEPROM) into volatile storage (RAM) on startup and written back when they are changed.)
+      </description>
+      <entry value="0" name="MISSION_READ_PERSISTENT">
+        <description>Read current mission data from persistent storage</description>
+      </entry>
+      <entry value="1" name="MISSION_WRITE_PERSISTENT">
+        <description>Write current mission data to persistent storage</description>
+      </entry>
+      <entry value="2" name="MISSION_RESET_DEFAULT">
+        <description>Erase all mission data stored on the vehicle (both persistent and volatile storage)</description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
@@ -1761,8 +1797,8 @@
       </entry>
       <entry value="245" name="MAV_CMD_PREFLIGHT_STORAGE" hasLocation="false" isDestination="false">
         <description>Request storage of different parameter values and logs. This command will be only accepted if in pre-flight mode.</description>
-        <param index="1" label="Parameter Storage" minValue="0" maxValue="3" increment="1">Parameter storage: 0: Read from flash/EEPROM, 1: Write current parameter data to flash/EEPROM, 2: Reset to defaults, 3: Reset sensor calibration parameter data to factory default (or firmware default if not available)</param>
-        <param index="2" label="Mission Storage" minValue="0" maxValue="2" increment="1">Mission storage: 0: Read from FLASH/EEPROM, 1: Write current data to flash/EEPROM, 2: Reset to defaults</param>
+        <param index="1" label="Parameter Storage" enum="PREFLIGHT_STORAGE_PARAMETER_ACTION">Action to perform on the persistent parameter storage</param>
+        <param index="2" label="Mission Storage" enum="PREFLIGHT_STORAGE_MISSION_ACTION">Action to perform on the persistent mission storage</param>
         <param index="3" label="Logging Rate" units="Hz" minValue="-1" increment="1">Onboard logging: 0: Ignore, 1: Start default rate logging, -1: Stop logging, &gt; 1: logging rate (e.g. set to 1000 for 1000 Hz logging)</param>
         <param index="4">Reserved</param>
         <param index="5">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -997,7 +997,7 @@
         <description>Write all parameter values to persistent storage (flash/EEPROM)</description>
       </entry>
       <entry value="2" name="PARAM_RESET_CONFIG_DEFAULT">
-        <description>Reset all user configurable parameters to their default value. Does not reset values that contain operation counters and vehicle computed statistics.</description>
+        <description>Reset all user configurable parameters to their default value (including airframe selection, sensor calibration data, safety settings, and so on). Does not reset values that contain operation counters and vehicle computed statistics.</description>
       </entry>
       <entry value="3" name="PARAM_RESET_SENSOR_DEFAULT">
         <description>Reset only sensor calibration parameters to factory defaults (or firmware default if not available)</description>


message_definitions/v1.0/common.xml
```diff
@@ -1761,7 +1761,7 @@
       </entry>
       <entry value="245" name="MAV_CMD_PREFLIGHT_STORAGE" hasLocation="false" isDestination="false">
         <description>Request storage of different parameter values and logs. This command will be only accepted if in pre-flight mode.</description>
-        <param index="1" label="Parameter Storage" minValue="0" maxValue="3" increment="1">Parameter storage: 0: Read from flash/EEPROM, 1: Write current parameter data to flash/EEPROM, 2: Reset to defaults, 3: Reset sensor calibration parameter data to factory default (or firmware default if not available)</param>
+        <param index="1" label="Parameter Storage" minValue="0" maxValue="4" increment="1">Parameter storage: 0: Read from flash/EEPROM, 1: Write current parameter data to flash/EEPROM, 2: Reset all configurable parameters to defaults, 3: Reset sensor calibration parameter data to factory default (or firmware default if not available), 4: Reset all (including non-configurable) parameters to defaults</param>
```
hamishwillee, line 1764: 1. Does "0" mean "reset the current params from stored param values on external memory - i.e. not "live memory"? Or something else?
2. Is it worth splitting some of these out into an enum? The main reason I suggest this is that it would be good to have this definition of configurable clearly stated in the message.

hamishwillee:
Sounds sensible to me. @auturgy @julianoes ?

message_definitions/v1.0/common.xml
hamishwillee: My assumption is that 2 is currently taken to mean "reset ALL" parameters, so this is a break. Can you switch 2 and 4?

hamishwillee:
@ThomasDebrunner We discussed in the mav call. Logically this is all fine - it should have no real consequences. We'd like you to replace the text strings for the Parameter Storage with an enum though, and more precisely outline what they mean/how they are used. Does that make sense?

ThomasDebrunner:
@hamishwillee Thanks! I changed the fields to enum, lmn if you'd like a different formulation of the strings

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
         <description>Autotune yaw axis.</description>
       </entry>
     </enum>
+    <enum name="PREFLIGHT_STORAGE_PARAMETER_ACTION">
+      <description>Action to take on the persistent parameter storage</description>
+      <entry value="0" name="PARAM_READ_PERSISTENT">
+        <description>Read all parameters from persistent storage (flash/EEPROM)</description>
```
hamishwillee, line 991:
```suggestion
        <description>Read all parameters from persistent storage (flash/EEPROM), clearing any updated values in RAM.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
...
+      </entry>
+      <entry value="1" name="PARAM_WRITE_PERSISTENT">
+        <description>Write all parameter values to persistent storage (flash/EEPROM)</description>
+      </entry>
+      <entry value="2" name="PARAM_RESET_CONFIG_DEFAULT">
+        <description>Reset all user configurable parameters to their default value. Does not reset values that contain operation counters and vehicle computed statistics.</description>
```
hamishwillee, line 997: Does this include sensors calibration?

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
...
+      </entry>
+      <entry value="3" name="PARAM_RESET_SENSOR_DEFAULT">
+        <description>Reset only sensor calibration parameters to factory defaults (or firmware default if not available).</description>
```
hamishwillee, line 1000:
```suggestion
        <description>Reset only sensor calibration parameters to factory defaults (or firmware default if not available)</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
         <description>Autotune yaw axis.</description>
       </entry>
     </enum>
+    <enum name="PREFLIGHT_STORAGE_PARAMETER_ACTION">
+      <description>Action to take on the persistent parameter storage</description>
```
hamishwillee, line 989: This is fine, but let's create a cross link in the docs to where it is used.
```suggestion
      <description>
        Actions for reading/writing parameters between persistent and volatile storage when using MAV_CMD_PREFLIGHT_STORAGE.
        (Commonly parameters are loaded from persistent storage (flash/EEPROM) into volatile storage (RAM) on startup and written back when they are changed.)
      </description>
```

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
...
+        <description>Reset only sensor calibration parameters to factory defaults (or firmware default if not available).</description>
+      </entry>
+      <entry value="4" name="PARAM_RESET_ALL_DEFAULT">
+        <description>Reset all parameters, including operation counters, to default values</description>
+      </entry>
+    </enum>
+    <enum name="PREFLIGHT_STORAGE_MISSION_ACTION">
+      <description>Action to take on the mission storage</description>
+      <entry value="0" name="MISSION_READ_PERSISTENT">
+        <description>Read current mission data from persistent storage (flash/EEPROM)</description>
```
hamishwillee, line 1009:
```suggestion
        <description>Read current mission data from persistent storage</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
...
+      </entry>
+      <entry value="1" name="MISSION_WRITE_PERSISTENT">
+        <description>Write current mission data to persistent storage (flash/EEPROM)</description>
```
hamishwillee, line 1012:
```suggestion
        <description>Write current mission data to persistent storage</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
...
+      </entry>
+      <entry value="2" name="MISSION_RESET_DEFAULT">
+        <description>Erase all mission data stored on the vehicle</description>
```
hamishwillee, line 1015:
```suggestion
        <description>Erase all mission data stored on the vehicle (both persistent and volatile storage)</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,36 @@
...
+      </entry>
+      <entry value="3" name="PARAM_RESET_SENSOR_DEFAULT">
+        <description>Reset only sensor calibration parameters to factory defaults (or firmware default if not available).</description>
+      </entry>
+      <entry value="4" name="PARAM_RESET_ALL_DEFAULT">
+        <description>Reset all parameters, including operation counters, to default values</description>
+      </entry>
+    </enum>
+    <enum name="PREFLIGHT_STORAGE_MISSION_ACTION">
+      <description>Action to take on the mission storage</description>
```
hamishwillee, line 1007: Suggest following modification - is it correct? Note, the use of "Mission" is widely used, because it predates geofence and rally points. We now tend to use Plan where we can. So perhaps (what do you think?) PREFLIGHT_STORAGE_PLAN_ACTION?


```suggestion
      <description>
        Actions for reading and writing plan information (mission, rally points, geofence) between persistent and volatile storage when using MAV_CMD_PREFLIGHT_STORAGE.
        (Commonly missions are loaded from persistent storage (flash/EEPROM) into volatile storage (RAM) on startup and written back when they are changed.)
      </description>
```

message_definitions/v1.0/common.xml
ThomasDebrunner: For `PARAM_RESET_CONFIG_DEFAULT `, all of the above will get reset. The only thing that does not get re-set is operation counters (like total flight time, or number of flights taken) and computed statistics, which could include things like "average distance per flight" or similar. Nothing else really. It doesn't really read from persistent storage, it resets the values to their default values, e.g. writes them

message_definitions/v1.0/common.xml
```diff
@@ -985,6 +985,42 @@
...
+        (Commonly parameters are loaded from persistent storage (flash/EEPROM) into volatile storage (RAM) on startup and written back when they are changed.)
+      </description>
+      <entry value="0" name="PARAM_READ_PERSISTENT">
+        <description>Read all parameters from persistent storage. Replaces values in volatile storage.</description>
+      </entry>
+      <entry value="1" name="PARAM_WRITE_PERSISTENT">
+        <description>Write all parameter values to persistent storage (flash/EEPROM)</description>
+      </entry>
+      <entry value="2" name="PARAM_RESET_CONFIG_DEFAULT">
+        <description>Reset all user configurable parameters to their default value. Does not reset values that contain operation counters and vehicle computed statistics.</description>
```
hamishwillee, line 1000:
```suggestion
        <description>Reset all user configurable parameters to their default value (including airframe selection, sensor calibration data, safety settings, and so on). Does not reset values that contain operation counters and vehicle computed statistics.</description>
```

message_definitions/v1.0/common.xml
hamishwillee: Thanks, I added a note to that. Mostly just because I wanted to confirm that sensor calibration is included. 

> It doesn't really read from persistent storage, it resets the values to their default values, e.g. writes them

OK. FWIW just trying to capture that the data will be default values after this. Presumably the result is that both static and volatile data ends up being default values (being written to) but not a detail we need to capture.

hamishwillee (approved):
@julianoes Can you give this a quick sanity check before I merge?

julianoes (approved):

peterbarker:
OK, so I'm 4 years late to this party.  But:

> This PR changes the meaning of the existing command (2) to only reset "config" values, while adding a new option (4) to completely wipe everything.

I'm trying to put this politely.

What the fuck?

julianoes:
@peterbarker try harder :smile:.

I think/hope the PR description is not actually reflecting what the PR diff changes. Looking at it, it does seem to be backwards compatible, or am I missing something? It depends a bit on the terminology around default, default config, etc..

hamishwillee:
> Looking at it, it does seem to be backwards compatible, or am I missing something? It depends a bit on the terminology around default, default config, etc..

This is a clarification that `2: Reset to defaults` means do not reset internal params - I suspect for example you might not want to delete some data that needs to persist for vehicle lifetime and so on. If ArduPilot really resets everything on value 2 this does look like a break to me. I'm not sure a critical one though, since as you say @julianoes ArduPilot could well consider every parameter to be user configurable.

hamishwillee:
IMO this is breaking - option 2 should have been left alone and the new behaviour been done in a new enum value. Unfortunately reverting needs to be managed.

I've created #2418 to look at our options.

PR #1830, opened 2022-04-05
Precision XY hold with MAV_CMD_NAV_LOITER_TO_ALT 

dakejahl:
Use previously empty param3 in MAV_CMD_NAV_LOITER_TO_ALT to specify usage of a target for XY position hold during altitude adjustment.

Ardupilot already has the concept of "Precision Loiter" as a mode.
https://ardupilot.org/copter/docs/precision-landing-with-irlock.html

PX4 could use this during a mission to perform a precise altitude adjustment above a target.


dakejahl:
Actually this should probably be MAV_CMD_DO_CHANGE_ALTITUDE. I didn't realize LOITER_TO_ALT was mostly a fixed wing thing

hamishwillee:
@dakejahl FWIW you can use [MAV_CMD_NAV_LOITER_TO_ALT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TO_ALT) in multicopter's, it's just that loiter means "hold position in the horizontal plane", and all the other params are ignored. You can use `MAV_CMD_DO_CHANGE_ALTITUDE` on either vehicle too - its just that the loiter command on fixed wing allows you to better plan the path out of the loiter on fixed wing. For  `MAV_CMD_DO_CHANGE_ALTITUDE` you're just changing the setpoint, so all other behaviour is "system default".

There is a little bit about what the loiters mean in https://mavlink.io/en/services/mission.html#loiter_commands

dakejahl:
>You can use MAV_CMD_DO_CHANGE_ALTITUDE on either vehicle too 

This command isn't actually implemented in either Ardupilot or PX4 so it could be a good candidate to extend.

>For MAV_CMD_DO_CHANGE_ALTITUDE you're just changing the setpoint, so all other behaviour is "system default".

Or would you prefer that a vehicle must first be placed into some kind of "XY lock" sub-mode, and this command stays as is? Ardupilot does it this way, PX4 could possibly do it this way as well if we added a new mode

hamishwillee:
> This command isn't actually implemented in either Ardupilot or PX4 so it could be a good candidate to extend.

I did not know that. I suspect that means the message is redundant - GCS just use [MAV_CMD_NAV_WAYPOINT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_WAYPOINT) or one of the loiter commands and specify the target altitude of interest. I would look at what QGC does on change altitude.

> Or would you prefer that a vehicle must first be placed into some kind of "XY lock" sub-mode, and this command stays as is? Ardupilot does it this way, PX4 could possibly do it this way as well if we added a new mode

Well, we should first confirm what people generally use. If this command is used by no one I'd rather deprecate it and document the "generally supported way".

If it is a useful message to have then I think my answer is "probably no". 
- The message explicitly states setting the altitude setpoint and not X, Y. 
- I **think** that is more flexible - I imagine this as being like set_speed - I call it to in parallel with other things. 
- In any case, would require some discussion about the use cases before it could be locked down.

hamishwillee:
If you want to use this one let me know and I'll raise it in the next dev call. I'll raise it anyway.

hamishwillee:
FYI apparently [MAV_CMD_DO_CHANGE_ALTITUDE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CHANGE_ALTITUDE) is implemented and used in PX4.

dakejahl:
image

dakejahl:
I think we should just define a new MAV_CMD though. I added a "precision loiter" mode into navigator so a corresponding command might make more sense.

dakejahl:
also I will probably hold off on upstreaming until we do a rewrite of navigator and navigator sub-modes in general

PR #1831, opened 2022-04-05
Figure eight MAV_CMD

KonradRudin:
Added figure eight MAV_CMD to command a figure eight loitering pattern and its related figure eight execution feedback message.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1143,22 +1143,6 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
-      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Fly a figure eight path as defined by the parameters.
-          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
-          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
-          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
-        </description>
-        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
-        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
-        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
-        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
-        <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
-        <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
-        <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
-      </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
         <description>Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicle's control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
@@ -7075,19 +7059,6 @@
       <description>Vehicle status report that is sent out while orbit execution is in progress (see MAV_CMD_DO_ORBIT).</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
       <field type="float" name="radius" units="m">Radius of the orbit circle. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
-      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
-      <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
-      <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
-      <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
-    </message>
-    <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
-      <wip/>
-      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
-      <description>Vehicle status report that is sent out while figure eight execution is in progress (see MAV_CMD_DO_FIGUREEIGHT).</description>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
-      <field type="float" name="major_radius" units="m">Major axis radius of the figure eight. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
-      <field type="float" name="minor_radius" units="m">Minor axis radius of the figure eight.</field>
-      <field type="float" name="orientation" units="rad">Orientation of the major axis of the figure eight.</field>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
       <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1150,7 +1150,7 @@
         <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
         <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
         <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
-        <param index="4" label="Orientation" units="rad"> Orientation of the figure eight major axis with respect to north. NaN: use default value of 0°.</param>
+        <param index="4" label="Orientation" units="rad"> Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
         <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1143,14 +1143,18 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
-      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
+      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
         <wip/>
         <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-        <description>Start flying on the outline of a figure eight defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
+          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
+        </description>
         <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
         <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
         <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
-        <param index="4" label="Orientation" units="rad"> Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
+        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
         <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1143,6 +1143,18 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Start flying on the outline of a figure eight defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
+        <param index="4" label="Orientation" units="rad"> Orientation of the figure eight major axis with respect to north. NaN: use default value of 0°.</param>
+        <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
+        <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
+        <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
+      </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
         <description>Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicle's control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
@@ -7059,6 +7071,19 @@
       <description>Vehicle status report that is sent out while orbit execution is in progress (see MAV_CMD_DO_ORBIT).</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
       <field type="float" name="radius" units="m">Radius of the orbit circle. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
+      <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
+      <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
+      <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
+    </message>
+    <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Vehicle status report that is sent out while figure eight execution is in progress (see MAV_CMD_DO_FIGUREEIGHT).</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="float" name="major_radius" units="m">Major axis radius of the figure eight. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
+      <field type="float" name="minor_radius" units="m">Minor axis radius of the figure eight.</field>
+      <field type="float" name="orientation" units="rad">Orientation of the major axis of the figure eight.</field>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
       <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -157,6 +157,28 @@
     <!-- ALL the entries in the MAV_CMD enum have a maximum of 7 parameters -->
     <enum name="MAV_CMD">
       <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data. NaN and INT32_MAX may be used in float/integer params (respectively) to indicate optional/default values (e.g. to use the component's current yaw or latitude rather than a specific value). See https://mavlink.io/en/guide/xml_schema.html#MAV_CMD for information about the structure of the MAV_CMD entries</description>
+      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
+          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
+        </description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise. 
+        NaN: The radius will be set to 2.5 times the minor radius. 
+        Must be greater or equal to two times the minor radius for feasible values.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.
+        NaN: The radius will be set to the default loiter radius.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already loitering.</param>
+        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to true north in [-pi,pi). NaN: use default orientation aligned to true north.</param>
+        <param index="5" label="Latitude/X">Center point latitude/X coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle position, or current center if already loitering.</param>
+        <param index="6" label="Longitude/Y">Center point longitude/Y coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle position, or current center if already loitering.</param>
+        <param index="7" label="Altitude/Z">Center point altitude MSL/Z coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle altitude.</param>
+      </entry>
       <entry value="900" name="MAV_CMD_PARAM_TRANSACTION" hasLocation="false" isDestination="false">
         <description>Request to start or end a parameter transaction. Multiple kinds of transport layers can be used to exchange parameters in the transaction (param, param_ext and mavftp). The command response can either be a success/failure or an in progress in case the receiving side takes some time to apply the parameters.</description>
         <param index="1" label="Action" enum="PARAM_TRANSACTION_ACTION">Action to be performed (start, commit, cancel, etc.)</param>
@@ -269,6 +291,19 @@
       <field type="uint8_t" name="signal_quality" units="%">WiFi network signal quality.</field>
       <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.</field>
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
+    </message>
+    <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Vehicle status report that is sent out while figure eight execution is in progress (see MAV_CMD_DO_FIGURE_EIGHT).</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="float" name="major_radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise.</field>
+      <field type="float" name="minor_radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles.</field>
+      <field type="float" name="orientation" units="rad">Orientation of the figure eight major axis with respect to true north in [-pi,pi).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
+      <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field.</field>
+      <field type="int32_t" name="y">Y coordinate of center point. Coordinate system depends on frame field.</field>
+      <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
     <message id="396" name="COMPONENT_INFORMATION_BASIC">
       <description>Basic component information data.</description>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -163,7 +163,7 @@
         <description>Fly a figure eight path as defined by the parameters.
           Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
           The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode), allowing POI tracking for gimbals that don't support infinite rotation.
-          This command only defines the flight path. Speed should be set indepentently (use e.g. MAV_CMD_DO_CHANGE_SPEED).
+          This command only defines the flight path. Speed should be set independently (use e.g. MAV_CMD_DO_CHANGE_SPEED).
           Yaw and other degrees of freedom are not specified, and will be flight-stack specific (on vehicles where they can be controlled independent of the heading).
         </description>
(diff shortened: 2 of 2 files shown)


tstastny:
@MaEtUgR

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
...
       </entry>
+      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Start flying on the outline of a figure eight defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
+        <param index="4" label="Orientation" units="rad"> Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
```
MaEtUgR, line 1153:
```suggestion
        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Start flying on the outline of a figure eight defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
```
MaEtUgR, line 1150: Maybe a stupid question but how is clockwise defined in a figure-eight of arbitrary orientation? Is it if you orient it back to 0 and the more north part is the clockwise one?

MaEtUgR (commented):
The general setup makes perfect sense but the flight/drive path is not exactly defined or am I mistaken? Is that by design?

Just to clarify intuitively what might be intended:
- major radius >> minor radius: flying back and forth on a line just turning on one end to the left on the other to the right
- major radius = 2 * minor radius: flying two loiters next to each other switching always in the tangential point
- major radius < 2 * minor radius: unfeasible tangentially, what to do exactly?

hamishwillee:
This would have to be defined in development.xml until there is a proven implementation and intent for another flight stack to implement.

message_definitions/v1.0/common.xml
hamishwillee, line 1150: I don't like using sign on the radius to indicate direction, if only because a negative radius is confusing.

A negative velocity is more intuitive.

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
...
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
+        <param index="4" label="Orientation" units="rad"> Orientation of the figure eight major axis with respect to north. NaN: use default value of 0.</param>
+        <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
```
hamishwillee, line 1154: If no MAV_frame is specified then do you mean a WGS84 latitude scaled by 1e7 , or unscaled or what. I am aware that this copies the ORBIT command, but I think that is broken too.

If a frame is specified I'd expect both to be according to the frame. And I'd expect the default frame to be global - i.e. 1E7 scaled

So for this, something like:

> Center point latitude/X co-ordinate. If no MAV_FRAME specified, `MAV_FRAME_GLOBAL` is assumed. etc ....

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
...
+        <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
+        <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
```
hamishwillee, line 1156: As above - the frame should define this, and you define a default frame - just my two bits. @MaEtUgR did you do ORBIT? Can we change that?

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
...
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Start flying on the outline of a figure eight defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
```
hamishwillee, line 1152: As above, direction as a negative/positive velocity makes more sense.

hamishwillee:
I think I need some diagrams showing the orientation and the major and minor radius. 

Are there any requirements on the entry and exit path? This may make a difference on missions - to give you and idea of what I mean see the loiter commands: https://mavlink.io/en/services/mission.html#loiter_commands

Can you confirm:
1. Path is intended for FW vehicles, but you'd expect any vehicle capable of flying the path to obey it? 
2. If not, what would an MC do? Ignore it? Where would MC face if it got this message? Direction of travel, center, what?
3. Is this intended to work as both a command an in missions?
4. Why would you prefer to do this instead of just circle loiter (just FMI - what's the "need").

message_definitions/v1.0/common.xml
KonradRudin: Yes, the clockwise rotation refers to the direction of the north circle.

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Start flying on the outline of a figure eight defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
```
hamishwillee, line 1149:
```suggestion
        <description>Fly a figure eight path as defined by the parameters. Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values. The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode). Multicopters and other unsupported vehicles s must NAK the command as unsupported.  </description>
```

message_definitions/v1.0/common.xml
hamishwillee: Yes. Good reason, but all those other commands suck :-) @MaEtUgR @julianoes What do you think?

hamishwillee:
Thanks @KonradRudin !

I added a proposal to state that MC vehicles should simply ignore the command (NACK with unsupported). 

@MaEtUgR I think we should either define expected behaviour in a mission or that it can only be used in a command. Ditto for Orbit. Which approach do you think is best. My leaning is to say it can only be used in a command.

To be a very clear, orbit and loiter already have a bit of overlap. So does this. I want to make sure we can give users clear direction on which command you should use and when and our design should reflect that.
Right now it looks like you use LOITER in missions, and that orbit and FIGURE8 are explicitly intended for commands (and should be rejected in missions?). We should think in each case what will happen on different vehicles and in commands vs mission.

The above doesn't answer:

> Are there any requirements on the entry and exit path? This may make a difference on missions - to give you and idea of what I mean see the loiter commands: https://mavlink.io/en/services/mission.html#loiter_commands

My assumption is no - because the loiter commands are in missions where you KNOW you have a new waypoint to go to. For a command the behaviour is up to you "what comes next". So here you might define the inward track to the figure 8, but the user decides how they exit.

hamishwillee:
PS I still want images :-)

tstastny:
> I added a proposal to state that MC vehicles should simply ignore the command (NACK with unsupported).

@hamishwillee @KonradRudin Any reason why we should explicitly tell MC *not* to follow these? I believe MC can currently follow orbit commands in PX4, and it wouldn't be hard to implement the same for figure 8. Of course depends on the given flight controller and if they have this logic implemented.

If the behavior is well enough described, I don't see why MC cannot also follow a figure eight.

message_definitions/v1.0/common.xml
MaEtUgR: Great, I think that's common. 👍  Possibly we should state it more explicit for the less familiar (like me 😇 ).

MaEtUgR:
> To be a very clear, orbit and loiter already have a bit of overlap.

@hamishwillee I'm aware there's sometimes confusion but from my perspective: Loiter tells the vehicle to stop and wait in a location. A fixed-wing happens to require flying in a circle to keep up the airspeed and not stall. Orbit tells the vehicle to circle in an exactly specified way. Am I wrong? Do we need to clarify further?

> Right now it looks like you use LOITER in missions, and that orbit and FIGURE8 are explicitly intended for commands

That's not by design but more about what's currently implemented in PX4. The implementation to make it all available in mission is desired but currently not as straightforward. How do we reflect that current state without explicitly disallowing the mission use in the MAVLink standard?

> I added a proposal to state that MC vehicles should simply ignore the command

I do not see a reason why multicopters, rovers, boats, ... can not execute the command just because it's initially added for a fixed-wing use case. Like above there's at the moment no implementation. How should we reflect that? I could do the multicopter one if it's desired.

sfuhrer:
>Orbit tells the vehicle to circle in an exactly specified way

I'm not sure if using "Orbit" for "flying an arbitrary patter" (in our case circular or figure8) is intuitive, [as it by standard definition is supposed to be called this way only if on a circular or elliptical path](https://www.vocabulary.com/dictionary/orbit). I would avoid references here to "orbit" and rather use "Loiter". I know that PX4 at least isn't consistent in the Loiter vs Orbit usage, but this sounds at least to me the right way forward is to have [LOITER_UNLIMITED](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_UNLIM) as the circular pendant to the here proposed unlimited figure8.

message_definitions/v1.0/common.xml
```diff
@@ -1143,6 +1143,18 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="35" name="MAV_CMD_DO_FIGUREEIGHT" hasLocation="true" isDestination="true">
```
julianoes, line 1146: What about `MAV_CMD_DO_FIGURE_EIGHT`, I think it might be more readable.

message_definitions/v1.0/common.xml
julianoes: I think I'm used to the negative radius convention, so I'd be ok with it.

message_definitions/v1.0/common.xml
```diff
@@ -7064,6 +7076,19 @@
...
+      <field type="float" name="major_radius" units="m">Major axis radius of the figure eight. Positive values orbit clockwise, negative values orbit counter-clockwise.</field>
+      <field type="float" name="minor_radius" units="m">Minor axis radius of the figure eight.</field>
+      <field type="float" name="orientation" units="rad">Orientation of the major axis of the figure eight.</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the fields: x, y, z.</field>
+      <field type="int32_t" name="x">X coordinate of center point. Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
+      <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
+      <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
+    </message>
```
julianoes, line 7091: :+1:

message_definitions/v1.0/common.xml
julianoes: I don't understand what the "outline" means either.

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
```
I'm just going to change that.

message_definitions/v1.0/common.xml
hamishwillee: The meaning is fairly clear - it means "fly the line defined by". But it is redundant.

message_definitions/v1.0/common.xml
hamishwillee: It must be stated, and I think we need diagrams (still).

hamishwillee: I hate you all.

RomanBapst: > NaN: Use vehicle default radius (for the major radius)

This is not clear to me because there is not default major radius. For the minor radius it makes more sense I think because that one can be set to the default loiter radius. So maybe it should state that if major radius is NAN then it will be set to x * minor_radius?

message_definitions/v1.0/common.xml
RomanBapst, line 1153: Could add "true north"

RomanBapst: Also we should add the range, no? E.g. [-pi, pi]

message_definitions/v1.0/common.xml
KonradRudin: Good catch @roman, i will change that.

message_definitions/v1.0/common.xml
KonradRudin: @tstastny @MaEtUgR  i really don't see any use case for MC to execute the figure of eight, so i would agree with @hamishwillee proposed text. I know, they would technically be able to fly it, but for what purpose? Also for the MC, the figure of eight is not completely defined, as we don't specify what to do with yaw? We don't have any parameters left to define that. Do you still want to put it in anyways?

tstastny: I'm just saying that boats or multi-copters or auto gryos or helicopters or etc could definitely do the pattern, and who are we to say what an operator wants to do with any given pattern? it only depends on what the flight controller's limitations are (which is always the case for any mavlink command). E.g. what to do with the gimbal is also not defined in this message for the fixed-wing, but we assume that is handled elsewhere, and it is even the specific use case this command is being developed for. Similarly, there are many mavlink commands that don't explicitly define every aspect of *how a vehicle does something, but rather the general desired *behavior. (so what to do with yaw maybe doesnt need to be explicitly defined in this message for MC?)

If that was a sticky point, could also switch out speed for yaw. Especially for the fixed-wing case, does anyone really use the speed command in mavlink for a specific command? Or wouldn't we rather have the vehicle decide what safe cruise speed it can fly at. We also have the "set speed" commands, right? Which would be a more modular way of independently setting desired behaviors for the vehicle without needing to limit ourselves to 4 params for every possible capability the vehicle could possibly have.

I just dont see a reason to arbitrarily limit use cases because we can't see the need for it now? @hamishwillee am I off base on the principles of mavlink? These are clearly just my opinions, but maybe I'm missing some fundamental guidelines mavlink messages are meant to adhere to. Please let me know.

KonradRudin:
Figure8

I guess i would need to put this into the documentation? Also, @hamishwillee there is no requirement for entry exit path, For now it will also loiter unlimited

hamishwillee:
Excellent diagram. I will think about this properly next week.

message_definitions/v1.0/common.xml
hamishwillee: You're not off-base. My position on this is just that the spec is clear and unambiguous for/about the cases it supports. If someone is ever potentially going to want to do this on copter then we'd be best thinking this out now.

MaEtUgR: At this moment the only purpose I see to support it on multirotor and other vehicle types is an autonomous reproducable test for varying curvature not 360° turning paths. This can be done to test position control tuning and also to do endurance, high speed, ... vehicle tests autonomously. Orbit already gets used for similar use cases and I honestly did not really expect that at the beginning. But Orbit is a steady state constant thrust, constant tilt, turning the heading >360° which has advantages for some tests and disadvantages for others. I agree with @tstastny 's points.

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
+          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
+        </description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise. 
+        NaN: The radius will be set to 2.5 times the minor radius. 
+        Must be greater or equal to two times the minor radius for feasible values.</param>
```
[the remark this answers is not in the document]
KonradRudin: But if the major axis is <2*minor axis, there is no line which is tangential to the circles. Then, you would need to implement two different logics on how to connect those elements. Also if you add another logic, there is at least the restriction that the major axis >= 1 * minor axis. Isn't that too much to define a separate behavior just for this region? At least for the intended use case it is implemented it does not make any sense.

message_definitions/v1.0/common.xml
KonradRudin: OK, to move forward, should we allow the motion for all vehicle, but state that it only defines the linear/position motion. If the vehicle has more degrees of freedom (as yaw  for MC for example) this would need to be defined in another MAVLINK message, as we don't have any parameters left to define it. Should we even define a default behavior then?

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
+          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
+        </description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise. 
+        NaN: The radius will be set to 2.5 times the minor radius.
```
hamishwillee, line 169: @KonradRudin You need to specify direction for this case. I have assumed clockwise (positive).

```suggestion
        NaN: The radius will be set to 2.5 times the minor radius and direction is clockwise.
```

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
+          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
+        </description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise. 
+        NaN: The radius will be set to 2.5 times the minor radius. 
+        Must be greater or equal to two times the minor radius for feasible values.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.
+        NaN: The radius will be set to the default loiter radius.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already loitering.</param>
+        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to true north in [-pi,pi). NaN: use default orientation aligned to true north.</param>
```
hamishwillee, line 174:
```suggestion
        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to true north in range: [-pi,pi]. NaN: use default orientation aligned to true north.</param>
```

hamishwillee: P.S. Note, I'd put the range in with `minValue`, `maxValue` but not sure how to specify pi.

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode).
```
hamishwillee, line 165:
```suggestion
          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode), allowing POI tracking for gimbals that don't support infinite rotation.
```

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+          Multicopters and other unsupported vehicles s must NAK the command as unsupported.
```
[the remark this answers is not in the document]
hamishwillee: @MaEtUgR @KonradRudin OK then, let's fully spec it for hover-vehicles then. The main decision is where it should yaw? I'd like to do [ORBIT_YAW_BEHAVIOUR](https://mavlink.io/en/messages/common.html#ORBIT_YAW_BEHAVIOUR) but we have no slots.

One alternative is to specify a priority, but I don't know if it is flexible enough.

```suggestion
          Multicopters (and other vehicles that can yaw separately from heading) should yaw towards the nearest POI if set, otherwise towards the center.
```

message_definitions/v1.0/common.xml
hamishwillee: Not sure, I have proposed this alternative: https://github.com/mavlink/mavlink/pull/1831#discussion_r849074501 for you to think about.

hamishwillee:
@MaEtUgR 

>> To be a very clear, orbit and loiter already have a bit of overlap.

> @hamishwillee I'm aware there's sometimes confusion but from my perspective: Loiter tells the vehicle to stop and wait in a location. A fixed-wing happens to require flying in a circle to keep up the airspeed and not stall. Orbit tells the vehicle to circle in an exactly specified way. Am I wrong? Do we need to clarify further?

Good point. You are correct, but I never quite "connected the dots". Reading again it is clear this is indeed a circle that you follow. We could make it clear in the loiter commands that the radius is approximate? Or in this message we could clarify by adding this to the description?

```
Unlike when loitering, fixed wing should attempt to closely follow the defined circumference.
```

>> Right now it looks like you use LOITER in missions, and that orbit and FIGURE8 are explicitly intended for commands

> That's not by design but more about what's currently implemented in PX4. The implementation to make it all available in mission is desired but currently not as straightforward. How do we reflect that current state without explicitly disallowing the mission use in the MAVLink standard?

There isn't any way. What would normally happen is that the first person to implement it would do so and we'd all feel annoyed with the suboptimal implementation. Ditto with what happens if an FW tries to fly the orbit.
That's why we spec it out.

I think it all makes sense for FW except for the yaw. We should explicitly state that FW always ignore specified yaw value and follows the vehicle trajectory. Initially it will head towards the orbit center but will then follow the flight path. This would be true in both missions and commands.

I think it all makes sense for missions except for yaw values . The values you have now are:

value | name | description
-- | -- | --
0 | ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TO_CIRCLE_CENTER | Vehicle front points to the center (default).
1 | ORBIT_YAW_BEHAVIOUR_HOLD_INITIAL_HEADING | Vehicle front holds heading when message received.
2 | ORBIT_YAW_BEHAVIOUR_UNCONTROLLED | Yaw uncontrolled.
3 | ORBIT_YAW_BEHAVIOUR_HOLD_FRONT_TANGENT_TO_CIRCLE | Vehicle front follows flight path (tangential to circle).
4 | ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED

For normal waypoints in a mission the MC yaw is, as I recall system defined. On PX4 that's [MPC_YAW_MODE](http://docs.px4.io/master/en/advanced_config/parameter_reference.html#MPC_YAW_MODE):
  - 0: towards waypoint
  - 1: towards home
  - 2: away from home
  - 3: along trajectory
  - 4: towards waypoint (yaw first)

1. Should we add a yaw option that points to a/nearest POI, if set?
3. Should we change ORBIT_YAW_BEHAVIOUR_UNCONTROLLED to "Undefined = system default" (i.e. what does uncontrolled mean?)
4. We will need to explicitly omit ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED when in missions right?

Note sure how to specify this. Depends on what you want it to do.

>> I added a proposal to state that MC vehicles should simply ignore the command

> I do not see a reason wh...

Answered elsewhere. Better to spec it

auturgy:
why not degrees for orientation? conversion to/from radians needs to occur at one end either way, and I think that's better on the consumer (autopilot) end than the generator (user interface).

KonradRudin:
> why not degrees for orientation? conversion to/from radians needs to occur at one end either way, and I think that's better on the consumer (autopilot) end than the generator

@auturgy I'm more of a fan of defining the interfaces in SI units and to convert non SI units right away, but i have seen that the other MAVLINK messages are not consistent in either using radians or degrees (e.g. for yaw). So i will leave that decision to the reviewers to decide. @hamishwillee @MaEtUgR @tstastny whats your opinion on this?

hamishwillee:
>> why not degrees for orientation? conversion to/from radians needs to occur at one end either way, and I think that's better on the consumer (autopilot) end than the generator

> @auturgy I'm more of a fan of defining the interfaces in SI units and to convert non SI units right away, but i have seen that the other MAVLINK messages are not consistent in either using radians or degrees (e.g. for yaw). So i will leave that decision to the reviewers to decide. @hamishwillee @MaEtUgR @tstastny whats your opinion on this?

Yuk:
- Personally I prefer degrees - I don't think in radians and my assumption is that most people don't. 
- @LorenzMeier has previous stated a preference for SI units/radians
- Based on the `units` attribute there are 61 instances of `rad` and 131 of degree-like things (16 `cdeg` and 47 `deg` 68 `degE7` (GPS lat/log). `degE7` is also used implicitly 36 MAV_CMDs for lat/log - so 72 more implicit instances.
- Consistency is nice, but:
  - we will not be changing existing messages. 
  - plenty of the above cases are places where you would never use radians - e.g. wherever we have used degE7 and cdeg
- It really doesn't matter from a conversion perspective - someone has to do this.

Given that there is no right answer I'm not going to ask for you to change this. Would anyone else like to provide an opinion? @LorenzMeier ?

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
KonradRudin: Thanks for the proposition @hamishwillee , however it is not a good idea to define a default behavior to point towards the center. In the figure eight flight, the vehicle will be flying through the center, which would result in sudden yaw changes of 180°, which is not desired. Only the initial heading or tangent to path makes sense. Maybe its stil best to explicitly state, that the message only defines 3 degrees of freedom on the position, if the vehicle has more degrees of freedom (as the yaw behavior)) this should be defines elsewhere (we still can propose a default behavior).

message_definitions/v1.0/development.xml
KonradRudin: Yeah, i was not sure either on how to put pi as maxValue, thats why i did it this way.

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
KonradRudin: Hey @hamishwillee , i have update the description. We can't use the yaw to center, and hold the initial heading or hold front tangential to track is epending on the airframe. Thus i have stated that this degree of freedom is uncontrolled (and the default behavior should be defined directly onboard).

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode), allowing POI tracking for gimbals that don't support infinite rotation.
+          This message defines only the 3D positional degree of freedom. Multicopter and other vehicles with additional degrees of freedom (e.g. yaw separately from course) will have those degrees of freedom uncontrolled.
+        </description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise. 
+        NaN: The radius will be set to 2.5 times the minor radius and direction is clockwise.
+        Must be greater or equal to two times the minor radius for feasible values.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles. Negative value has no effect.
```
hamishwillee, line 171:
```suggestion
        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the two circles that make up the figure. Negative value has no effect.
```

message_definitions/v1.0/development.xml
```diff
@@ -270,6 +292,19 @@
...
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
     </message>
+    <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
+      <wip/>
+      <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
+      <description>Vehicle status report that is sent out while figure eight execution is in progress (see MAV_CMD_DO_FIGURE_EIGHT).</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="float" name="major_radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise.</field>
+      <field type="float" name="minor_radius" units="m">Minor axis radius of the figure eight. Defines the radius of the circles.</field>
```
hamishwillee, line 301:
```suggestion
      <field type="float" name="minor_radius" units="m">Minor axis radius of the figure eight. Defines the radius of two circles that make up the figure.</field>
```

message_definitions/v1.0/development.xml
hamishwillee: What I'm trying to do here is describe what the minor and major radius are in words, so that you'd have a chance of understanding this without images.
Once this is accepted into common I can add the image and link to it though.

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,28 @@
...
+      <entry value="35" name="MAV_CMD_DO_FIGURE_EIGHT" hasLocation="true" isDestination="true">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+        <description>Fly a figure eight path as defined by the parameters.
+          Set parameters to NaN/INT32_MAX (as appropriate) to use system-default values.
+          The command is intended for fixed wing vehicles (and VTOL hybrids flying in fixed-wing mode), allowing POI tracking for gimbals that don't support infinite rotation.
+          This message defines only the 3D positional degree of freedom. Multicopter and other vehicles with additional degrees of freedom (e.g. yaw separately from course) will have those degrees of freedom uncontrolled.
```
hamishwillee, line 166: Minimally we should change to refer to this as a command rather than a message. Here is a slightly simpler version.
```suggestion
          This command only defines the flight path.
          Vehicles with additional degrees of freedom (e.g. that can have different yaw and heading) will have those degrees of freedom uncontrolled.
```

hamishwillee: Actually, that is precise, but there is the slight niggle in my mind that "uncontrolled" here means "uncontrolled by this command" not "uncontrolled at all". Might this be better?:

```suggestion
          This command only defines the flight path.
          Yaw and other degrees of freedom are not specified, and will be flight-stack specific (on vehicles where they can be controlled independent of the heading).
```

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
hamishwillee: @KonradRudin Thanks. I take your point. I've suggested an alternative to your wording that is just a little simpler: https://github.com/mavlink/mavlink/pull/1831/files#r859306083

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
KonradRudin: Also already at major radius = 2 * minor radius, those are two full circle of exactly 360° changes. if you make it smaller, it you can't make the vehicle turn less suddenly. If the space is of concern, you would need to define some other logic (as for example always make half a circle and then connect with non tangential lines ;)). I mark this as resolved unless @MaEtUgR  has some other opinion.

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
KonradRudin: Thanks, looks good for me.

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,29 @@
...
+        NaN: The radius will be set to 2.5 times the minor radius and direction is clockwise.
+        Must be greater or equal to two times the minor radius for feasible values.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the two circles that make up the figure. Negative value has no effect.
+        NaN: The radius will be set to the default loiter radius.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already loitering.</param>
+        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to true north (range: [-pi,pi]). NaN: use default orientation aligned to true north.</param>
+        <param index="5" label="Latitude/X">Center point latitude/X coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle position, or current center if already loitering.</param>
```
Jaeyoung-Lim, line 177: I am not sure if we need to define the behavior of the flight stack on how the message should be consumed from the mavlink message. Wouldn't it be already enough that `INT32_MAX` would invalidate the position field?

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,29 @@
...
+          This command only defines the flight path.
+          Yaw and other degrees of freedom are not specified, and will be flight-stack specific (on vehicles where they can be controlled independent of the heading).
+        </description>
+        <param index="1" label="Major Radius" units="m">Major axis radius of the figure eight. Positive: orbit the north circle clockwise. Negative: orbit the north circle counter-clockwise. 
+        NaN: The radius will be set to 2.5 times the minor radius and direction is clockwise.
+        Must be greater or equal to two times the minor radius for feasible values.</param>
+        <param index="2" label="Minor Radius" units="m">Minor axis radius of the figure eight. Defines the radius of the two circles that make up the figure. Negative value has no effect.
+        NaN: The radius will be set to the default loiter radius.</param>
+        <param index="3" label="Velocity" units="m/s">Cruising Velocity. NaN: Use vehicle default velocity, or current velocity if already loitering.</param>
```
Jaeyoung-Lim, line 174: While this is a reference path, why is it necessary to define the velocity of the figure 8 trajectory? Also, would this be airspeed or groundspeed. For a fixed wing, shouldn't it always fly at the specified mission airspeed?

message_definitions/v1.0/development.xml
```diff
@@ -157,6 +157,29 @@
...
+        <param index="4" label="Orientation" units="rad">Orientation of the figure eight major axis with respect to true north (range: [-pi,pi]). NaN: use default orientation aligned to true north.</param>
+        <param index="5" label="Latitude/X">Center point latitude/X coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle position, or current center if already loitering.</param>
+        <param index="6" label="Longitude/Y">Center point longitude/Y coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle position, or current center if already loitering.</param>
+        <param index="7" label="Altitude/Z">Center point altitude MSL/Z coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
+        INT32_MAX or NaN: Use current vehicle altitude.</param>
```
Jaeyoung-Lim, line 181: Same comment as above, wouldn't using the current vehicle altitude a flight stack specific behavior?

Jaeyoung-Lim (commented):
Probably too late, but some minor comments / questions on the message definition

 - It seems that the message itself is not safe guarded against degenerate figure 8s. Would the plan be that the flight stack is responsible of checking whether the message has a valid figure 8 trajectory?

message_definitions/v1.0/development.xml
KonradRudin: The cruising speed should refer to airspeed for fixed wing and ground speed for multicopters, this convention was also used in other mavlink messages afaik. Formulating it like this works for both vehicles then. Also i would keep the velocity in the message, the mission airspeed can be used for optimal traveling speed (least energy consumption for distance), here you can specify a different speed for best loitering (optimal speed to stay airborne).

message_definitions/v1.0/development.xml
KonradRudin: INT32_MAX and NaN are specified depending on the Frame type. if defined as latitude longitude, you should set NaN to invalidate (floating point value), for local coordinate in meters its INT32_MAX. For your second comment, i guess you refer to the specification to use the current vehicle position if it is invalid? Its just easier for the onboard side to determine the current position instead of the user-side to fill in the current vehicle position if it should start the figure of eight immediately at the current position. And its more consistent with the other messages.

Jaeyoung-Lim:
> You referring to the major axis which must be greater than twice the minor axis? Yes this must be ultimately checked by the flight stack. I don't think you can put constraints on value in MAVLINK based on other values. But also the user side should check this as well in the GUI, as you shouldn't be able to draw the figure of 8 with tangential lines, if you violate the constraint.

It would have been nicer if the parameterization of the figure 8 path could have been inherently feasible, but if not probably there are no better way than currently implemented. Thanks for clarifying!

message_definitions/v1.0/development.xml
tstastny: https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CHANGE_SPEED
^i also think speed should be set independently

KonradRudin: ok, but MAV_CMD_DO_ORBIT has. Seems to be inconsistent there as well. I would also be in favor of setting different things with different messages, in the sense of " do this path using this speed and this heading" combining three messages. But how do you handle that with other messages, who have already defined it inside? Seems many have already defined a yaw behavior for example and not using different messages.

Jaeyoung-Lim: @KonradRudin I think your right, I wasn't aware of the orbit message and it seems like it is inconsistent. `MAV_CMD_DO_ORBIT` also seems to be `MAV_CMD_NAV_LOITER_*` + behavior related commands combined into one :man_facepalming: However as far as I am concerned the speed command in `MAV_CMD_DO_ORBIT` are all ignored in fixedwing for PX4. Is the plan then to add this functionality in PX4?

message_definitions/v1.0/development.xml
KonradRudin: I think, you should not define flight stack behavior, if it could change for different vehicle or vehicle types. In this case of the center of the figure of 8 however, it can be advantageous to force a default for different flight stacks so it is directly clear from the mavlink message what the behavior is instead of consulting the flight stack implementation as well..

message_definitions/v1.0/development.xml
KonradRudin: And @tstastny there is also a MAV_CMD_DO_CHANGE_ALTITUDE, We could also use this to set the altitude and use the figure of 8 mavlink message only as a lateral definition. There is no clear distinction in the other mavlink messages to define only a minimal set. Here a 3D trajectory is chosen.

tstastny: @KonradRudin, good catch with the DO_CHANGE_ALTITUDE, but IMO a "loiter" seems like something that should have a fully defined position. Otherwise, where is the loitering defined? at current altitude? Keeping the xyz coordinate for it seems more appropriate. Whereas speed is something that generally already has a default in FCU's, and would mean we just require a "change" in speed from our current operating condition (hence the DO_CHANGE_SPEED).

This is just my interpretation though

KonradRudin: @hamishwillee would you be ok, to remove the airspeed. It is fine with me to use this message to define only the path and use other more generic commands to tweak the vehicles behavior like the airspeed or yaw.

hamishwillee: @KonradRudin Yes I would be fine with you removing the airspeed. I would also be happy with you leaving it in. The "pro" of removing it is that you can decide to use that field for something else later.
- You might want to add a comment in the description that the airspeed may be changed independently using DO_CHANGE_SPEED
- If you say that, the you should verify it in the implementation (I would expect it to just work) because the command changes the setpoint.
- Consider that if you might want to use param 3 for this in the future. To do that you should redefine it with a default value of NaN: `<param index="3" reserved="true" default="NaN" />` (see [this](https://mavlink.io/en/guide/define_xml_element.html#reserved) for an explanation of why you need to decide if the default value is 0 or NaN)


DO_CHANGE_SPEED is safe to use in parallel because it just changes the setpoint - it doesn't have other semantics.[MAV_CMD_DO_CHANGE_ALTITUDE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CHANGE_ALTITUDE) should be safe for the same reason - but I do worry that someone might have implemented this assuming a circular pattern.

There is this (stalled) PR for setting yaw independently: https://github.com/mavlink/mavlink/pull/1297

message_definitions/v1.0/development.xml
hamishwillee: > For example if a flight stack using mavlink decides not to use the current position in case of an invalid position is passed then it is breaking the mavlink compatibility. Shouldn't this be more decided by the flight stack itself rather than defined in the message?

@Jaeyoung-Lim 
- We should absolutely define flight stack behaviour if we can do so sensibly and unambiguously. 
- We commonly want to tell the flight stack to do an orbit, figure 9, or whatever "at your current position". This avoids having to read the current position and send the command with a position that will inevitably be out of date and require a correction by the time it is received.

Note that here we mean a specific invalid value not "any invalid value". I would expect the command to reject a command if a value was sent that was genuinely invalid in the context of the command.

Perhaps I misunderstand the question.

message_definitions/v1.0/development.xml
KonradRudin: THanks @hamishwillee for hinting about defining the default on a reserved parameter. I updated it accordingly.

hamishwillee: You're welcome. It's a bit of a frustration, because you kind of have to "know in advance" if your unchanged value would be better as a 0 or a NaN. The default is 0.

hamishwillee:
@KonradRudin Thanks very much. IMO many people have had an opportunity to comment now and we won't learn much else except by prototyping it. I will therefore merge.

Note: 
1. Once you have a prototype and can demonstrate that you're happy with the behaviour we create a PR to move this to common.xml
2. Until it is in common.xml it is still subject to change by you or others.

As an aside (non blocking), should we perhaps add a note in the status message about the rate that it should be sent?

hamishwillee:
P.S. When you eventually submit a PR to put this in master can you please cross reference this PR. That will remind me of the context and that I need to add documentation for this. At that point we might at an URL link in the description as well.

KonradRudin:
> As an aside (non blocking), should we perhaps add a note in the status message about the rate that it should be sent?

Yes, we could add that as we send it at a low update rate atm (i think its 2Hz).

hamishwillee:
Cool @KonradRudin - I added a note and merged: https://github.com/mavlink/mavlink/pull/1841

PR #1836, opened 2022-04-28
Clarify and cross reference home position everywhere

hamishwillee:
The home position is set at/after takeoff (and may also be set later using MAV_CMD_DO_SET_HOME - though if this is done before it is set on takeoff the value will be over-written). It is defined as the default position that the vehicle will return to.

This was explained primarily in deprecated messages (not current messages) and also there were cases where the term "home location" was used.

This PR updates all the HOME_POSITION related commands and messages to clearly explain what the position is and cross link to each other. It also turns instances of home location into home position.

This is a consistency change across the related content, and should have no impact. @auturgy @julianoes

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -251,7 +251,10 @@
         <description>NOT a coordinate frame, indicates a mission command.</description>
       </entry>
       <entry value="3" name="MAV_FRAME_GLOBAL_RELATIVE_ALT">
-        <description>Global (WGS84) coordinate frame + altitude relative to the home position. First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
+        <description>
+          Global (WGS84) coordinate frame + altitude relative to the home position.
+          First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home position.
+        </description>
       </entry>
       <entry value="4" name="MAV_FRAME_LOCAL_ENU">
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -260,7 +263,10 @@
         <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + altitude relative to the home position. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude with 0 being at the altitude of the home location.</description>
+        <description>
```
julianoes, line 266: Does this whitespace work like that?

julianoes (commented):
I'm not sure if the white space changes work well in generators but other than that it looks ok.

message_definitions/v1.0/common.xml
hamishwillee: Yes it works fine like this. Annoyingly the text here gets pulled in and treated as HTML so any whitespace is ignored from a rendering perspective. What it does do is make the source much more readable, particularly in change requests/diffs. That's why I have been doing this.

hamishwillee: @auturgy You OK with this? Easy enough to revert to all one line descriptions. But I LIKE the source being readable, and while a bit odd, this breaks nothing.

hamishwillee: Discussed in dev call https://github.com/mavlink/mavlink/wiki/20220525-Dev-Meeting

This is required in order to improve our docs (via markdown) and is tested to be non-breaking. We can fix up the crappy layout in generated code if we wish. @auturgy OK with it. I am merging.

PR #1843, opened 2022-05-11
HOME_POSITION - quaternion can have an invalid value

### LINKED ISSUE #1839: Clarification: Home Position Q
I'm writing code for my [MAVLink library](https://github.com/bolderflight/mavlink) for my [company's flight stack](https://github.com/bolderflight/spaaro) to send the home position when the vehicle is armed. This is actually to facilitate a companion path planning computer to know the origin of the NED frame to convert mission items from a global to a local NED frame.

The [home position message](http://mavlink.io/en/messages/common.html#HOME_POSITION) specifies *q* as a normal vector perpendicular to the surface and aligned with the heading. This makes sense and I assume that q is a quaternion. Is there more information or an example for how q should be computed? Of course quaternions can come in multiple orders (i.e. rotation first, rotation last, etc), so I think a little more clarity would help with general applicability of this message across flight stacks.

Also the *approach_x*, *approach_y*, and *approach_z* variables are a little confusing and seem to imply that for a fixed-wing aircraft, takeoff and landing should happen in opposite directions? Which doesn't make sense to me.

hamishwillee:
[HOME_POSITION](http://mavlink.io/en/messages/common.html#HOME_POSITION)  includes a quaternion that specifies the slope and heading of the vehicle on takeoff. 

Both PX4 and ArduPilot provide false values.
- PX4 sets roll and pitch as (level vehicle) but specifies the yaw.
- ArduPilot sets it to [1,0,0,0] (level vehicle facing North).

This change states that if the values are not known to be true then they should be set to `NaN`. I assume this is reasonable, but may not be mathematically. If it is OK , then we should raise bugs against the flight stacks to output the invalid value.

If not, we need alternative way to tell recipients that the data is at least partially bogus.

Fixes #1839


flybrianfly:
Would it be better to just include the heading / yaw instead of using a quaternion? I'm not sure how I would use the slope information in real life - is anyone operating out of areas where the slope would actually be needed?

hamishwillee:
The message has been used in the wild for a long time, so whether some alternative would be better is kind of irrelevant in the context of this message. We have to assume that it has been and that quaternion is useful. At that point the question for this message is how we make sure that it is unambiguous and ensure that flight stacks can provide sensible information (even if that might be to say "I'm not providing this info"). 

I don't know if the slope is actually needed - the fact that neither PX4 or ardupilot provide it indicates perhaps not. Perhaps a new message might make sense as well.

julianoes:
What if we add roll and pitch to the two reserved fields in [MAV_CMD_DO_SET_HOME](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_HOME) (param2, param3)?

message_definitions/v1.0/common.xml
```diff
@@ -6415,7 +6415,11 @@
       <field type="float" name="x" units="m">Local X position of this position in the local coordinate frame</field>
       <field type="float" name="y" units="m">Local Y position of this position in the local coordinate frame</field>
       <field type="float" name="z" units="m">Local Z position of this position in the local coordinate frame</field>
-      <field type="float[4]" name="q">World to surface normal and heading transformation of the takeoff position. Used to indicate the heading and slope of the ground</field>
+      <field type="float[4]" name="q">
+        Quaternion indicating world-to-surface-normal and heading transformation of the takeoff position.
+        Used to indicate the heading and slope of the ground.
+        Use NaN in fields to indicate values that are not supplied.
```
hamishwillee, line 6421: As discussed in dev call, you can't just indicate one field in a quaternion as NaN without making the whole thing meaningless. So either all info should be supplied or none. 
```suggestion
        All fields should be set to NaN if an accurate quaternion for both heading and surface slope cannot be supplied.
```

hamishwillee:
> What if we add roll and pitch to the two reserved fields in [MAV_CMD_DO_SET_HOME](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_HOME) (param2, param3)?

Not sure. I guess you could argue that if someone is willing to specify that the home position has a particular attitude then we should take that. I lean towards agree with James that this isn't actually of interest to people, and we should avoid adding anything until a use case that someone really needs is  provided.

auturgy (approved):

auturgy:
Doesn't break anything, aids clarity and hopefully will flow through into the flight stacks

julianoes (approved):

message_definitions/v1.0/common.xml
```diff
@@ -6545,7 +6545,11 @@
       <field type="float" name="x" units="m">Local X position of this position in the local coordinate frame (NED)</field>
       <field type="float" name="y" units="m">Local Y position of this position in the local coordinate frame (NED)</field>
       <field type="float" name="z" units="m">Local Z position of this position in the local coordinate frame (NED: positive "down")</field>
-      <field type="float[4]" name="q">World to surface normal and heading transformation of the takeoff position. Used to indicate the heading and slope of the ground</field>
+      <field type="float[4]" name="q">
```
hamishwillee, line 6548:
```suggestion
      <field type="float[4]" name="q" invalid="[NaN]">
```

hamishwillee:
I've updated to use `invalid=['Nan']` tag too. 

@auturgy I know that ArduPilot are already thinking about this. I figure we might have other values that should be set invalid values to but we can wait on that.

@julianoes Who would we point this to in PX4? I.e. lets make sure that there is at least an issue report so we can make sure that people are aware PX4 is non compliant with this change (probably).

Merging now. Thanks.

PR #1845, opened 2022-05-15
WIND_COV: clarify description and unknown value

julianoes:
This should clarify the meaning of the wind variability and wind accuracy a bit.
Also, it describes the values to signal when a value is unknown.

In my opinion it is fine to change the meaning of the variability slightly as it is hard-coded in PX4, and not implemented in ArduPilot (since ArduPilot uses WIND, not WIND_COV).

As suggested in https://github.com/ArduPilot/ardupilot/issues/4230#issuecomment-222572110.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6294,14 +6294,14 @@
     <message id="231" name="WIND_COV">
       <description>Wind estimate from vehicle.</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
-      <field type="float" name="wind_x" units="m/s">Wind in North (NED) direction (NAN if unknown)</field>
-      <field type="float" name="wind_y" units="m/s">Wind in East (NED) direction (NAN if unknown)</field>
-      <field type="float" name="wind_z" units="m/s">Wind in down (NED) direction (NAN if unknown)</field>
-      <field type="float" name="var_horiz" units="m/s">Variability of wind in XY, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
-      <field type="float" name="var_vert" units="m/s">Variability of wind in Z, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
-      <field type="float" name="wind_alt" units="m">Altitude (MSL) that this measurement was taken at (NAN if unknown)</field>
-      <field type="float" name="horiz_accuracy" units="m/s">Horizontal speed 1-STD accuracy (NAN if unknown)</field>
-      <field type="float" name="vert_accuracy" units="m/s">Vertical speed 1-STD accuracy (NAN if unknown)</field>
+      <field type="float" name="wind_x" units="m/s" invalid="NaN">Wind in North (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_y" units="m/s" invalid="NaN">Wind in East (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_z" units="m/s" invalid="NaN">Wind in down (NED) direction (NAN if unknown)</field>
+      <field type="float" name="var_horiz" units="m/s" invalid="NaN">Variability of wind in XY, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
+      <field type="float" name="var_vert" units="m/s" invalid="NaN">Variability of wind in Z, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
+      <field type="float" name="wind_alt" units="m" invalid="NaN">Altitude (MSL) that this measurement was taken at (NAN if unknown)</field>
+      <field type="float" name="horiz_accuracy" units="m/s" invalid="0">Horizontal speed 1-STD accuracy (0 if unknown)</field>
+      <field type="float" name="vert_accuracy" units="m/s" invalid="0">Vertical speed 1-STD accuracy (0 if unknown)</field>
     </message>
     <message id="232" name="GPS_INPUT">
       <description>GPS sensor input message.  This is a raw sensor value sent by the GPS. This is NOT the global position estimate of the system.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6292,16 +6292,16 @@
       <field type="float" name="pos_vert_accuracy" units="m">Vertical position 1-STD accuracy relative to the EKF local origin</field>
     </message>
     <message id="231" name="WIND_COV">
-      <description>Wind covariance estimate from vehicle.</description>
+      <description>Wind estimate from vehicle.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6292,16 +6292,16 @@
...
-      <field type="float" name="horiz_accuracy" units="m/s">Horizontal speed 1-STD accuracy</field>
-      <field type="float" name="vert_accuracy" units="m/s">Vertical speed 1-STD accuracy</field>
+      <field type="float" name="wind_x" units="m/s">Wind in North (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_y" units="m/s">Wind in East (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_z" units="m/s">Wind in down (NED) direction (NAN if unknown)</field>
+      <field type="float" name="var_horiz" units="m/s">Variability of wind in XY, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
+      <field type="float" name="var_vert" units="m/s">Variability of wind in Z, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
+      <field type="float" name="wind_alt" units="m">Altitude (MSL) that this measurement was taken at (NAN if unknown)</field>
+      <field type="float" name="horiz_accuracy" units="m/s">Horizontal speed 1-STD accuracy (NAN if unknown)</field>
+      <field type="float" name="vert_accuracy" units="m/s">Vertical speed 1-STD accuracy (NAN if unknown)</field>
```
hamishwillee, line 6304: @WickedShell Suggested that we might use 0 for the accuracy, since a value of 0 is suspect. That would allow these last two to zero-byte truncate. Is it worth it?

message_definitions/v1.0/common.xml
```diff
@@ -6292,16 +6292,16 @@
...
-      <field type="float" name="wind_z" units="m/s">Wind in Z (NED) direction</field>
-      <field type="float" name="var_horiz" units="m/s">Variability of the wind in XY. RMS of a 1 Hz lowpassed wind estimate.</field>
-      <field type="float" name="var_vert" units="m/s">Variability of the wind in Z. RMS of a 1 Hz lowpassed wind estimate.</field>
-      <field type="float" name="wind_alt" units="m">Altitude (MSL) that this measurement was taken at</field>
-      <field type="float" name="horiz_accuracy" units="m/s">Horizontal speed 1-STD accuracy</field>
-      <field type="float" name="vert_accuracy" units="m/s">Vertical speed 1-STD accuracy</field>
+      <field type="float" name="wind_x" units="m/s">Wind in North (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_y" units="m/s">Wind in East (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_z" units="m/s">Wind in down (NED) direction (NAN if unknown)</field>
+      <field type="float" name="var_horiz" units="m/s">Variability of wind in XY, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
```
hamishwillee, line 6300: @WickedShell says "I suspect removing the 1Hz lowpass wind requirement is probably better, but I don't feel that strongly about it". 

Thoughts? (I don't know enough to understand the implications of this requirement)

message_definitions/v1.0/common.xml
julianoes: Yer I was wondering about that too. Then the question is if these fields are really at the end?
Yes, they are: https://github.com/mavlink/c_library_v2/blob/241907e288b43513b28f83595d0de3b2088bce0f/common/mavlink_msg_wind_cov.h#L15-L16

So, yes, setting to 0 for unknown would be fine with me.

hamishwillee, line 6304: OK, changing to zero and setting invalid flag for the same thing on all.

```suggestion
      <field type="float" name="wind_x" units="m/s" invalid="NaN">Wind in North (NED) direction (NAN if unknown)</field>
      <field type="float" name="wind_y" units="m/s" invalid="NaN">Wind in East (NED) direction (NAN if unknown)</field>
      <field type="float" name="wind_z" units="m/s" invalid="NaN">Wind in down (NED) direction (NAN if unknown)</field>
      <field type="float" name="var_horiz" units="m/s" invalid="NaN">Variability of wind in XY, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
      <field type="float" name="var_vert" units="m/s" invalid="NaN">Variability of wind in Z, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
      <field type="float" name="wind_alt" units="m" invalid="NaN">Altitude (MSL) that this measurement was taken at (NAN if unknown)</field>
      <field type="float" name="horiz_accuracy" units="m/s" invalid="0">Horizontal speed 1-STD accuracy (0 if unknown)</field>
      <field type="float" name="vert_accuracy" units="m/s" invalid="0">Vertical speed 1-STD accuracy (0 if unknown)</field>
```

julianoes: As far as I'm aware there is no change that would make anything PX4 sends incompatible.

message_definitions/v1.0/common.xml
```diff
@@ -6292,16 +6292,16 @@
...
-      <field type="float" name="var_vert" units="m/s">Variability of the wind in Z. RMS of a 1 Hz lowpassed wind estimate.</field>
-      <field type="float" name="wind_alt" units="m">Altitude (MSL) that this measurement was taken at</field>
-      <field type="float" name="horiz_accuracy" units="m/s">Horizontal speed 1-STD accuracy</field>
-      <field type="float" name="vert_accuracy" units="m/s">Vertical speed 1-STD accuracy</field>
+      <field type="float" name="wind_x" units="m/s" invalid="NaN">Wind in North (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_y" units="m/s" invalid="NaN">Wind in East (NED) direction (NAN if unknown)</field>
+      <field type="float" name="wind_z" units="m/s" invalid="NaN">Wind in down (NED) direction (NAN if unknown)</field>
+      <field type="float" name="var_horiz" units="m/s" invalid="NaN">Variability of wind in XY, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
+      <field type="float" name="var_vert" units="m/s" invalid="NaN">Variability of wind in Z, 1-STD estimated from a 1 Hz lowpassed wind estimate (NAN if unknown)</field>
```
Jaeyoung-Lim, line 6301: Is this the standard deviation or the variance? The units and description imply that `var_vert` should be standard deviation, but the fact that it has `var_` makes it confusing. Or if it is variance, the unit should be `m^2/s^2`

julianoes: It is confusing indeed but `var_` stands for variability :joy: as far as I can tell.

julianoes: Exactly 🤦‍♂️

hamishwillee:
@julianoes Given that there is so much confusion, is it perhaps worth saying in the top description "none of these values are actually covariances!"? 

Actually, given there is so much confusion, could we start again with `WIND_INFO` and give each of the fields sensible names and values? 
If so, now is the right time before ArduPilot move to WIND_COV?

julianoes:
I don't want to start over. That just opens the discussion up even further, delays it further, and then doesn't get implemented.

hamishwillee:
As discussed in dev call, this is much more clear than it was, and about as far as we can go without a breaking change. Merging.

hamishwillee:
Thanks Julian!

PR #1847, opened 2022-05-23
Change  current_battery type in BATTERY_STATUS and SYS_STATUS 

feemi:
This pr fixes overflow problem on high current values. 
When current is over around 320Amps, gcs shows negative values on telemetry feedback.


hamishwillee:
@feemi THanks, but this is a breaking change. It changes the [CRC_EXTRA](https://mavlink.io/en/guide/serialization.html#crc_extra), which means that any GCS (say) that hasn't been updated with the new version of mavlink will drop both messages (not get them), and any GCS that has been updated will not be able to process the message from any older autopilots that have not been updated.

I have proposed a new battery status message in https://github.com/mavlink/rfcs/pull/19/files that will effectively supersede BATTERY_STATUS. This already proposes a bigger field. Would welcome comments on that RFC.

The `SYS_STATUS` battery indications needs to be revisited (yet again). The fields, which are shown below, were created a long time ago and assumed a single battery (and as you say, lower currents than are now sometimes used).

field | size | unit | Desc.
--- | --- | --- | ---
voltage_battery | uint16_t | mV |   | Battery voltage, UINT16_MAX: Voltage not sent by autopilot
current_battery | int16_t | cA |   | Battery current, -1: Current not sent by autopilot
battery_remaining | int8_t | %

What I'd like to do is deprecate these. However they are much used by OSDs and other simpler systems. For a large current system like yours I would definitely set these to -1.

hamishwillee:
Closing, as we can't do this.

I have updated docs to make issues with these fields more obvious: https://github.com/mavlink/mavlink/pull/1850 . Also note the RFC will fix this issue with the current.

PR #1849, opened 2022-05-26
MAV_CMD_DO_SET_HOME - add home position roll/pitch

hamishwillee:
The vehicle outputs a quaternion of vehicle roll/pitch/yaw  in the [HOME_POSITION](https://mavlink.io/en/messages/common.html#SET_HOME_POSITION) message.
As discussed in https://github.com/mavlink/mavlink/pull/1843, the expectation is that either the quaternion outputs accurate values or NaNs for all value (even if you have a value for the yaw, you can't output it in the form of a quaternion unless you also have "some" roll and pitch value). 

The proposal here is that we allow the `MAV_CMD_DO_SET_HOME` to also specify a roll and pitch of the home position.

Note, there is no way to say "I don't know the pitch, roll" of the surface at home - if you don't know you're saying it's flat. Is that OK?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1368,8 +1368,8 @@
           Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2" label="Roll" units="deg">Roll angle (of surface). Range: -180..180 degrees. NAN or 0 means value not set. 0.01 indicates zero roll.</param>
-        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). Range: -90..90 degrees. NAN or 0 means value not set. 0.01 means zero pitch.</param>
+        <param index="2" label="Roll" units="deg" minValue="-180" maxValue="180">Roll angle (of surface). Range: -180..180 degrees. NAN or 0 means value not set. 0.01 indicates zero roll.</param>
+        <param index="3" label="Pitch" units="deg" minValue="-90" maxValue="90">Pitch angle (of surface). Range: -90..90 degrees. NAN or 0 means value not set. 0.01 means zero pitch.</param>
         <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading. Range: -180..180 degrees.</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1368,8 +1368,8 @@
           Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).      
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2" label="Roll" units="deg">Roll angle (of surface). NaN is equivalent to 0.</param>
-        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). NaN is equivalent to 0.</param>
+        <param index="2" label="Roll" units="deg">Roll angle (of surface). 0 means value not set. 0.01 indicates zero roll.</param>
+        <param index="3" label="Pitch" units="deg">Pitch angle (of surface).  0 means value not set. 0.01 means zero pitch.</param>
         <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1368,8 +1368,8 @@
           Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).      
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2" label="Roll" units="deg">Roll angle (of surface). NaN to use default roll (0)</param>
-        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). NaN to use default roll (0)</param>
+        <param index="2" label="Roll" units="deg">Roll angle (of surface). NaN is equivalent to 0.</param>
+        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). NaN is equivalent to 0.</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1368,8 +1368,8 @@
           Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).      
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2">Empty</param>
-        <param index="3">Empty</param>
+        <param index="2" label="Roll" units="deg">Roll angle (of surface). NaN to use default roll (0)</param>
+        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). NaN to use default roll (0)</param>
```
hamishwillee, line 1372: Being more clear. NaN is zero for these cases - flat surface. There is no way to say "I don't know this surface angle". 
```suggestion
        <param index="2" label="Roll" units="deg">Roll angle (of surface). NaN is equivalent to 0.</param>
        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). NaN is equivalent to 0.</param>
```

hamishwillee:
This follows on from https://github.com/mavlink/mavlink/pull/1843#issuecomment-1136915800
The affected PX4 PR is https://github.com/PX4/PX4-Autopilot/pull/19717

julianoes:
I'd like to talk about this at the next meeting and either close it or get it in.

hamishwillee:
Makes sense. @auturgy FYI. I have added to the next agenda https://github.com/mavlink/mavlink/wiki/20231122-Dev-Meeting . If you guys have an opinion on this would be good to present that before the meeting

julianoes:
My vote is to get this in.

message_definitions/v1.0/common.xml
```diff
@@ -1368,8 +1368,8 @@
           Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).      
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2">Empty</param>
-        <param index="3">Empty</param>
+        <param index="2" label="Roll" units="deg">Roll angle (of surface). NaN is equivalent to 0.</param>
+        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). NaN is equivalent to 0.</param>
```
hamishwillee, line 1372: From description of this issue:

> Note, there is no way to say "I don't know the pitch, roll" of the surface at home - if you don't know you're saying it's flat. Is that OK?

`0` is the default that is always sent currently so it has to mean "we don't know".
The question is, do we to differentiate between 0 and not sent? 
If we do, how about we do something like this?

```suggestion
        <param index="2" label="Roll" units="deg">Roll angle (of surface). 0 means value not set. 0.01 indicates zero roll.</param>
        <param index="3" label="Pitch" units="deg">Pitch angle (of surface).  0 means value not set. 0.01 means zero pitch.</param>
```

hamishwillee:
Mine too.

auturgy:
Need to define ranges

message_definitions/v1.0/common.xml
julianoes: Note to self: add ranges

hamishwillee:
[20231122-Dev-Meeting ](https://github.com/mavlink/mavlink/wiki/20231122-Dev-Meeting) - agreed to get this in but need to define ranges on the fields (at least the new ones). OK for it to be zero and NaN for "unknown. Think the 0.1 value should be used to represent 0.

Ping me to look when you're done.

message_definitions/v1.0/common.xml
```diff
@@ -1365,12 +1365,12 @@
...
           The position is set automatically by the system during the takeoff (and may also be set using this command).
-          Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).      
+          Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2">Empty</param>
-        <param index="3">Empty</param>
-        <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading</param>
+        <param index="2" label="Roll" units="deg">Roll angle (of surface). Range: -180..180 degrees. NAN or 0 means value not set. 0.01 indicates zero roll.</param>
+        <param index="3" label="Pitch" units="deg">Pitch angle (of surface). Range: -90..90 degrees. NAN or 0 means value not set. 0.01 means zero pitch.</param>
```
hamishwillee, line 1372: We can specify range in metadata too. I am OK with having in both as long as they match.

```suggestion
        <param index="2" label="Roll" units="deg" minValue="-180" maxValue="180">Roll angle (of surface). Range: -180..180 degrees. NAN or 0 means value not set. 0.01 indicates zero roll.</param>
        <param index="3" label="Pitch" units="deg" minValue="-90" maxValue="90">Pitch angle (of surface). Range: -90..90 degrees. NAN or 0 means value not set. 0.01 means zero pitch.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1365,12 +1365,12 @@
...
-          Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).      
+          Note: the current home position may be emitted in a HOME_POSITION message on request (using MAV_CMD_REQUEST_MESSAGE with param1=242).
         </description>
         <param index="1" label="Use Current" minValue="0" maxValue="1" increment="1">Use current (1=use current location, 0=use specified location)</param>
-        <param index="2">Empty</param>
-        <param index="3">Empty</param>
-        <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading</param>
+        <param index="2" label="Roll" units="deg" minValue="-180" maxValue="180">Roll angle (of surface). Range: -180..180 degrees. NAN or 0 means value not set. 0.01 indicates zero roll.</param>
+        <param index="3" label="Pitch" units="deg" minValue="-90" maxValue="90">Pitch angle (of surface). Range: -90..90 degrees. NAN or 0 means value not set. 0.01 means zero pitch.</param>
+        <param index="4" label="Yaw" units="deg">Yaw angle. NaN to use default heading. Range: -180..180 degrees.</param>
```
hamishwillee, line 1373: @auturgy This looks like what we asked for except possibly this Yaw.  Does ardupilot allow this constraint to be applied?

hamishwillee: @auturgy Julian is going to add a range. Can you look at this tomorrow - otherwise I will merge this as "what we were talking about AFAIK"

hamishwillee:
I am self merging following the discussions on this in MAV call etc.

julianoes:
Thanks @hamishwillee

PR #1851, opened 2022-05-26
Update storm32.xml

olliw42:
This updates the storm32.xml dialect file. The changes are non functional.

1. Add the github user name to the contacts, as had been suggested/recommended somewhere I cannot find anymore
2. Removes a comment which has become obsolete.

THX!
Olli

### DIFF
--- a/message_definitions/v1.0/storm32.xml
+++ b/message_definitions/v1.0/storm32.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0"?>
 <!--
 Contact:
+  - github user name: @olliw42 (to ping)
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
Thanks @olliw42 - I think it is useful to have a github handle. The only requirement is "some reliable mechanism" for us to know who owns it.

olliw42:
It's me who has to thank you guys, as it was suggested by someone of you :)
I also have to thank ardupilot, since they finally made it possible that the comment could be removed.
thx for merging
:)

PR #1852, opened 2022-05-30
ardupilotmega: add nav_attitude_time command

rmackay9:
This adds a new ArduPilot specific NAV_ATTITUDE_TIME command which allows the user to specify an attitude they would like the vehicle to maintain for a specified period of time during.  This can be useful if a user wants to, for example, specify that a copter should drift with the wind for 10 seconds (they could set roll, pitch and climb_rate arguments to zero).

This has [already been merged into the AP specific mavlink](https://github.com/ArduPilot/mavlink/pull/272) but if there's interest in other teams using this message I'm happy to move it to common.xml instead.

This has been tested successfully in SITL using AP.

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -291,6 +291,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="42703" name="MAV_CMD_NAV_ATTITUDE_TIME" hasLocation="false" isDestination="false">
+        <description>Maintain an attitude for a specified time.</description>
+        <param index="1" label="time" units="s">Time to maintain specified attitude and climb rate</param>
+        <param index="2" label="roll" units="deg">Roll angle in degrees (positive is lean right, negative is lean left)</param>
+        <param index="3" label="pitch" units="deg">Pitch angle in degrees (positive is lean back, negative is lean forward)</param>
+        <param index="4" label="yaw" units="deg">Yaw angle</param>
+        <param index="5" label="climb_rate" units="m/s">Climb rate</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
       <entry value="43000" name="MAV_CMD_GUIDED_CHANGE_SPEED" hasLocation="false" isDestination="false">
         <description>Change flight speed at a given rate. This slews the vehicle at a controllable rate between it's previous speed and the new one. (affects GUIDED only. Outside GUIDED, aircraft ignores these commands. Designed for onboard companion-computer command-and-control, not normally operator/GCS control.)</description>
         <param index="1" label="speed type" enum="SPEED_TYPE">Airspeed or groundspeed.</param>


message_definitions/v1.0/ardupilotmega.xml
```diff
@@ -291,6 +291,16 @@
...
+      <entry value="42703" name="MAV_CMD_NAV_ATTITUDE_TIME" hasLocation="false" isDestination="false">
+        <description>Maintain an attitude for a specified time.</description>
+        <param index="1" label="time" units="s">Time to maintain specified attitude and climb rate</param>
+        <param index="2" label="roll" units="deg">Roll angle in degrees (positive is lean right, negative is lean left)</param>
+        <param index="3" label="pitch" units="deg">Pitch angle in degrees (positive is lean back, negative is lean forward)</param>
+        <param index="4" label="yaw" units="deg">Yaw angle</param>
+        <param index="5" label="climb_rate" units="m/s">Climb rate</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
```
hamishwillee, line 303: @rmackay9 As this is a dialect and you're a known/trusted owner of the dialect I'll just merge this.

That said, some comments that I would have if this was going into common.
1. What's the climb rate for if you're maintaining altitude? Is this the rate to climb to the altitude if you're not there already? 
2. So does time start from when the command is sent (that's implied) or from when you reach altitude?
3. What happens if this is used on an FW? 
4. Are there any limits: maxValue, minValue, increment on any of these? 
5. IN particular param 5 is likely to be an integer, so are you expecting only whole numbers of m/s climb rate?

hamishwillee:
@bkueng @julianoes @dagar  Is this something that might be useful in common.xml? I.e. does PX4 want this?

IF so, can you provide suggestions for things that would be "required" here?

message_definitions/v1.0/ardupilotmega.xml
rmackay9: @hamishwillee,

Txs very much for merging.  So this command is for **attitude** not altitude (easy to mix up).

I don't think this will be very useful on fixed wing or ground vehicles.  For those we might need an alternative command called something like ATTITUDE_THRUST_TIME which allow providing a thrust rather than a climb rate.

The roll, pitch and yaw are all Euler angles in degrees so they have these limits:

- roll: +- 180 deg
- pitch: +- 90 deg
- yaw: -180 to +360 (to allow for people to specify in the -180 to +180 range or in the 0 to 360)

You know, I don't think I've ever used the "minValue" and "maxValue" tags but I could add those. Txs for that suggestion.

hamishwillee: > So this command is for attitude not altitude (easy to mix up).

@rmackay9 You're very kind. I should learn to read.

For max, min, etc. these are supposed to make it easier for a GCS to create a sensible editor control. Don wanted them for QGC but I have not checked that it uses them. They could also auto-add limits for libraries too.
Docs on the schema here: https://mavlink.io/en/guide/xml_schema.html#param (or of course you can just check the schema doc). Note, [fields](https://mavlink.io/en/guide/xml_schema.html#messages) don't have these, but IMO should.

If it were me I would definitely add a comment that this is intended for MCs. Perhaps something like:

> This is intended for MC use. A vehicle that cannot independently control the specified degrees of freedom should reject the command with [MAV_RESULT_TEMPORARILY_REJECTED](https://mavlink.io/en/messages/common.html#MAV_RESULT_TEMPORARILY_REJECTED).

hamishwillee: PS This is still valid
- param 5 is sent in an integer. You need to move this is you expect users to specify floats, and otherwise add `increment` if you expect integers.

hamishwillee: If it were me I'd just push it to param 7 :-). Param 6 and 5 can just be empty. Yes the break isn't lovely, but not really a big deal. Your call!

Oh, and you're effectively forcing all of the values to be supplied, which is fine. You might choose to use NaN to indicate "current value".

rmackay9: The NaN idea is an interesting one.  I had thought that it would be a good idea to be able to use the current yaw instead of always having to provide it but couldn't think of a way to do that.  I'd forgotten that floats can carry a NaN.  txs!

hamishwillee:
@rmackay9 Had a chat. Interested on where you generally would use? Bluff body calculations?

General feeling was that we _might_ like to take it, but would it be possible to move climb_rate to param7 because it is clearly broken in this respect.

PR #1854, opened 2022-06-03
common.xml: add ignition_voltage to EFI_STATUS

peterbarker:
Supply voltage to EFI sparking system.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6289,6 +6289,8 @@
       <field type="float" name="exhaust_gas_temperature" units="degC">Exhaust gas temperature</field>
       <field type="float" name="throttle_out" units="%">Output throttle</field>
       <field type="float" name="pt_compensation">Pressure/temperature compensation</field>
+      <extensions/>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6289,6 +6289,8 @@
       <field type="float" name="exhaust_gas_temperature" units="degC">Exhaust gas temperature</field>
       <field type="float" name="throttle_out" units="%">Output throttle</field>
       <field type="float" name="pt_compensation">Pressure/temperature compensation</field>
+      <extensions/>
+      <field type="float" name="ignition_voltage" units="V">Supply voltage to EFI sparking system.  Zero in this value means "unknown", so if the supply voltage really is zero volts use 0.0001 instead.</field>
```
hamishwillee, line 6293: I'm OK with this. I guess the "usual" question - why not a uint_16_t and units of cV. You could still use 0 for the "not supplied".

@julianoes @auturgy Any opinion?

I guess my questions are the normal - is a float the right type? If this can never be negative you could use

julianoes (approved):

hamishwillee:
I had a chat to @peterbarker . This is sent at relatively low rate, and is already inefficient - proportionally having just this field being smaller will make no difference, except a negative one for any implementer (who will need to think differently about handling the particular field).

So technically would be the right thing to do, but not worth the hassle for anyone to modify this. Agreed to merge.

PR #1855, opened 2022-06-03
Improve payload deploy specific messages

junwoo091400:
## Background
This is a PR to improve the definitions of [payload specific messages](https://mavlink.io/en/services/payload.html#payload-specific-commands) that are related to payload **deployment**. Which covers:
1. [MAV_CMD_DO_GRIPPER](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GRIPPER)
2. [MAV_CMD_DO_WINCH](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_WINCH)

### Use case
This PR is needed for supporting a Missions where exact execution flow of the payload deployment actions are important.

Previous implementation which has no concept of timeout has always blindly trusted the payload deployments to work instantaneously, which is not the case in a lot of applications. The winch can get stuck, the gripper may not open in time, etc.

Also, without any time delay specifications, the previous behavior was to trigger the deployment of payload, and continue on the mission. This is not desirable since ideally, we want to be sure that we can deliver / actuate the payload precisely at a certain spot & verify that the deployment has been successful.

### Additions
With this PR, I am adding the Timeout & Definite Execution time definitions, which means the following:
1. Timeout: "Maximum time the vehicle should wait while waiting for the acknowledgement of the successful gripper action"
   - This is used for payload deployment mechanisms with feedback sensor
2. Definite Execution Time: "For how long the vehicle must wait after start commanding the gripper, regardless of the acknowledgement of a successful deployment"
   - Used for payload deployment mechanisms without feedback sensor

## Discussion

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1677,8 +1677,8 @@
         <description>Mission command to operate a gripper.</description>
         <param index="1" label="Instance" minValue="1" increment="1">Gripper instance number.</param>
         <param index="2" label="Action" enum="GRIPPER_ACTIONS">Gripper action to perform.</param>
-        <param index="3">Empty</param>
-        <param index="4">Empty</param>
+        <param index="3" label="Timeout" units="s">Maximum time the vehicle should wait while waiting for the acknowledgement of the successful gripper action. Used for gripers with feedback sensor</param>
+        <param index="4" label="Definite Execution Time" units="s">For how long the vehicle must wait after start commanding the gripper, regardless of the acknowledgement of a successful deployment. Used for grippers without feedback sensor</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
@@ -2356,10 +2356,10 @@
         <description>Command to operate winch.</description>
         <param index="1" label="Instance" minValue="1" increment="1">Winch instance number.</param>
         <param index="2" label="Action" enum="WINCH_ACTIONS">Action to perform.</param>
-        <param index="3" label="Length" units="m">Length of line to release (negative to wind).</param>
-        <param index="4" label="Rate" units="m/s">Release rate (negative to wind).</param>
-        <param index="5">Empty.</param>
-        <param index="6">Empty.</param>
+        <param index="3" label="Length" units="m">Length of line to release</param>
+        <param index="4" label="Rate" units="m/s">Release rate</param>
+        <param index="5" label="Timeout" units="s">Maximum time the vehicle should wait while waiting for the acknowledgement of the successful winch deployment. Used for winches with feedback sensor</param>
+        <param index="6" label="Definite Execution Time" units="s">For how long the vehicle must wait after start deploying the winch, regardless of the acknowledgement of a successful deployment. Used for winches without feedback sensor</param>
         <param index="7">Empty.</param>
       </entry>
       <!-- END of payload range (30000 to 30999) -->


TSC21:
🚀

junwoo091400:
I find it a bit odd that "MAV_CMD_DO_WINCH" MAVLink ID is like so out of random (42600), is that fixable? Do we need to consider compatibility to pre-existing Groundstation / Flight controllers?

hamishwillee:
> I find it a bit odd that "MAV_CMD_DO_WINCH" MAVLink ID is like so out of random (42600), is that fixable? Do we need to consider compatibility to pre-existing Groundstation / Flight controllers?

Yes.

It's not broken, but changing it will break things.

message_definitions/v1.0/common.xml
```diff
@@ -1677,8 +1677,8 @@
         <description>Mission command to operate a gripper.</description>
         <param index="1" label="Instance" minValue="1" increment="1">Gripper instance number.</param>
         <param index="2" label="Action" enum="GRIPPER_ACTIONS">Gripper action to perform.</param>
-        <param index="3">Empty</param>
-        <param index="4">Empty</param>
+        <param index="3" label="Timeout" units="s">Maximum time the vehicle should wait while waiting for the acknowledgement of the successful gripper action. Used for gripers with feedback sensor</param>
+        <param index="4" label="Definite Execution Time" units="s">For how long the vehicle must wait after start commanding the gripper, regardless of the acknowledgement of a successful deployment. Used for grippers without feedback sensor</param>
```
hamishwillee, line 1681: Why not use a https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_DELAY for this? That's what it is for.

hamishwillee, line 1681: So gripper sends an ack somehow and this triggers next waypoint. But if you don't get this trigger  you will wait for this amount of time?

For both cases would 0 mean as it is now? If so, that should be stated.

hamishwillee:
@rmackay9 could you please comment on the use case, and if you see alternative solutions? I don't think we need the fixed timeout, but the concept of wait until triggered or timeout seems reasonable enough - though you might not want to just continue in this case but definite an action.

julianoes:
I have the intuition that the specific information about a gripper or winch would be internal to the implementation. Therefore a user planning the mission would not have to know or care about these specific timeouts but the application would deal with it internally.

hamishwillee:
Thanks @julianoes I _think_ I agree.

When something is used in a mission there is no options for feedback, so generally the flight stack has to do a "sensible thing" - in this case not complete the operation until a timeout or it get's a signal from the gripper that it is ready to proceed.

My concern is that this _might_ be a special case. Normally in a mission no step can fail in such a way that you can't just continue to the next step. I.e. if the photo doesn't get taken or the orbit was too small, or whatever, continuing the mission should not end in tragedy. 
But is that true for this case? I mean if you are running a mission with a cargo drone and it fails to unload the cargo that might result in it not having enough juice to go home, or dropping the cargo half way. I don't know enough about what you'd "normally do" in those cases to know if this is worth thinking about.

Anyway, it might be interesting to think about what would happen if this was used manually first - i.e. what would a pilot do, what feedback would they get, what would the winch/gripper do "by default".

message_definitions/v1.0/common.xml
junwoo091400: Since that would mean the user would need to have the mission item sequence as : "GRIPPER" -> "NAV_DELAY", which isn't ideal as it increases the complexity of planning the mission (increasing the number of mission items)

junwoo091400: Exactly, the "Timeout" means the "Ack Timeout", which means exactly what you have described. And 0 would have the same behavior as now, yes!

junwoo091400:
> user planning the mission would not have to know or care about these specific timeouts but the application would deal with it internally.

Then the problem arises that the mission plan becomes ambiguous. How long should it wait at that waypoint while deploying the payload? Does it always have to wait for the ack? etc.

For example for FixedWing pacakge deployment: It is preferrable to not wait at the payload deployment stage for an ack, but rather go on to the next waypoint. How would the user be able to configure this? While keeping the ack functionality?

And sometimes, for Multicopter doing package deployment: It is preferrable to wait for the ack, so that it will be deployed with precision.

If we don't have the information in the mission file, I am not sure how we can differentiate the two cases above, which is why the setting was included in the MAVLink messages :thinking:

rmackay9:
It doesn't seem to me like these are values that a user would want to set each time.  They seem like they should be held in parameters.

hamishwillee:
@junwoo091400 In a mission there is no "ack". Whatever is executing the mission item decides whether it is done, and then goes to the next waypoint. I suspect that at the moment the implementation probably just triggers the winch/gripper and then immediately heads off, but as Randy says, you could have a parameter that says delay this time or wait for this return trigger. This _could_ be set differently for the FW or MC case, and appropriate for the actual type of gripper.

It is true that some of this could be configured in the message, but that would make it harder to write generic missions.

hamishwillee:
NOte, I'm very open to looking at payload deployment. There are other messages like https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_PAYLOAD_PLACE that seem to be unimplemented because they are too generic. But the whole area needs some dedicated thinking.

rmackay9:
@hamishwillee, of course [NAV_PAYLOAD_PLACE is implemented in ArduPilot](https://ardupilot.org/copter/docs/mission-command-list.html#package-place).

hamishwillee:
@rmackay9 Thanks. I'm aware. @auturgy indicates it is a bit of a pain to "get right".

message_definitions/v1.0/common.xml
julianoes: The design of mission items is composition of simple items/commands forming a more complex task combined. Putting to much complexity into one item is generally not preferred.

hamishwillee:
@junwoo091400 I think there is broad agreement that this is not the right way to iterate these messages. Closing.

PR #1858, opened 2022-06-22
Add MAV_CMD_DO_PRECISION_HOLD

dakejahl:
This adds an additional MAV_CMD to initiate what I am dubbing a "Precision Hold". This is exactly the same thing as Precision Land except instead of landing the vehicle will control XY (and Z if specified) to hover precisely above a target. Ardupilot already has something similar called [Precision Loiter](https://ardupilot.org/copter/docs/precision-landing-with-irlock.html). I chose the term Hold because Loiter usually implies fixed wing aircraft and this feature cannot be used with a fixed wing. I've already implemented this in a fork of PX4 and will merge into upstream once there's proper mavlink support.

We must look at what already exists for the mavlink [Landing Target protocol](https://mavlink.io/en/services/landing_target.html) and [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) message.

**Minimum requirements of the command**
- Specify whether we wish to enable or disable the Precison Hold behavior
- Specify which target to use if multiple targets are present
- Specify a height offset above target to maintain or use current altitude
- Specify if the vehicle should follow the target if it moves
- Specify if the vehicle should align its heading with the target heading

I chose to exclude optional Latitude/Longitude since this would make the command coordinate based, which would then require an Altitude in AMSL and makes the command more complex to use in missions (as well as we are limited to 7 parameters -- Altitude would make it 8)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1187,13 +1187,13 @@
       </entry>
       <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
         <description>
-          Vehicle moves to specified location (lat/lon/alt) and engages Precision Hold mode (NaN/INT32_MAX may be used to set the current location when the message is received).
-          If it gets LANDING_TARGET messages for the indicated landing target it: adjusts its horizontal position to align precisely above the target, adjusts its altitude to the Hold altitude (param3), and then holds the position. 
+          Vehicle moves to location (if specified) and descends above a target. It will hold its position above the target at a fixed distance, specified in the param "Hold distance".
+          If it gets LANDING_TARGET messages for the indicated landing target it: adjusts its horizontal position to align precisely above the target, descends to above the target, and then hold the position at "Hold distance" from the target.
           If the specified target is not found the vehicle holds the specified lat/lon/alt and enters normal Position Hold mode (the timeout for entering position hold is flight-stack specific).
         </description>
-        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode. 0 to use first target found.</param>
         <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
-        <param index="3" label="Hold altitude" units="m">Altitude above target to hold position. Set to NaN to use current altitude as the Hold altitude.</param>
+        <param index="3" label="Hold distance" units="m">Distance above target to hold position. NaN to use current altitude.</param>
         <param index="4">Empty</param>
         <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
         <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1185,7 +1185,7 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
-      <entry value="35" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
         <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a target.</description>
         <param index="1" label="Enable">Enable or disable Precision Hold.</param>
         <param index="2" label="Target number">The ID of the target if multiple targets are present.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1186,11 +1186,11 @@
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
       <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
-        <description>Vehicle moves to specified location and enables or disables Precision Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If any of the 3 fields Latitude/Longitude/Altitude are NAN the current position will be used instead.</description>
-        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
-        <param index="2" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found, the lowest ID target available will be used.</param>
-        <param index="3" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
-        <param index="4" label="Height above target" units="m">Set to zero to maintain current altitude.</param>
+        <description>Vehicle moves to specified location and engages Precison Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If Latitude/Longitude are FLT_MAX (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current position will be used instead. If Altitude is FLT_MAX (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current altitude will be used instead </description>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
+        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="3" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
+        <param index="4">Empty</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
         <param index="7" label="Altitude">Altitude</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1187,19 +1187,17 @@
       </entry>
       <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
         <description>
-          Vehicle moves to specified location (lat/lon/alt) and engages Precision Hold mode.
-          If it gets LANDING_TARGET messages for the indicated landing target it adjusts its horizontal position to align precisely above the target (and maintains altitude).
+          Vehicle moves to specified location (lat/lon/alt) and engages Precision Hold mode (NaN/INT32_MAX may be used to set the current location when the message is received).
+          If it gets LANDING_TARGET messages for the indicated landing target it: adjusts its horizontal position to align precisely above the target, adjusts its altitude to the Hold altitude (param3), and then holds the position. 
           If the specified target is not found the vehicle holds the specified lat/lon/alt and enters normal Position Hold mode (the timeout for entering position hold is flight-stack specific).
-          If Latitude/Longitude are NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current position will be used.
-          If Altitude is NaN the vehicle altitude when the message is received will be used.
         </description>
         <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
         <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
-        <param index="3" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
+        <param index="3" label="Hold altitude" units="m">Altitude above target to hold position. Set to NaN to use current altitude as the Hold altitude.</param>
         <param index="4">Empty</param>
-        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME</param>
-        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME</param>
-        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME</param>
+        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
+        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
+        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN to use current altitude.</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1186,12 +1186,12 @@
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
       <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
-        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a target.</description>
+        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a LANDING_TARGET.</description>
         <param index="1" label="Enable">Enable or disable Precision Hold.</param>
         <param index="2" label="Target number">The ID of the target if multiple targets are present.</param>
         <param index="3" label="Staionary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
         <param index="4" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
-        <param index="5" label="Height above target">Set to zero to maintain current altitude.</param>
+        <param index="5" label="Height above target" units="m">Set to zero to maintain current altitude.</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1186,7 +1186,13 @@
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
       <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
-        <description>Vehicle moves to specified location and engages Precison Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If Latitude/Longitude are FLT_MAX (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current position will be used instead. If Altitude is FLT_MAX (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current altitude will be used instead </description>
+        <description>
+          Vehicle moves to specified location (lat/lon/alt) and engages Precision Hold mode.
+          If it gets LANDING_TARGET messages for the indicated landing target it adjusts its horizontal position to align precisely above the target (and maintains altitude).
+          If the specified target is not found the vehicle holds the specified lat/lon/alt and enters normal Position Hold mode (the timeout for entering position hold is flight-stack specific).
+          If Latitude/Longitude are NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current position will be used.
+          If Altitude is NaN the vehicle altitude when the message is received will be used.
+        </description>
         <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
         <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
         <param index="3" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1186,8 +1186,14 @@
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
       <entry value="35" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
-        <description>Initiate a precision hold at the current position and maintain a distance above a target.</description>
-        <param index="1" label="Height above target">Set to zero to maintain current altitude</param>
+        <description>Enable or disable a Precision Hold. Optional Latitude/Longitude and distance above target.</description>
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
+        <param index="2" label="Height above target">Set to zero to maintain current altitude.</param>
+        <param index="3" label="Target number">The ID of the target if multiple targets are present.</param>
+        <param index="4" label="Staionary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
+        <param index="5" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="6" label="Latitude">Latitude to travel to before initiating Precision Hold. Set to NAN to use current position.</param>
+        <param index="7" label="Longitude">Longitude to travel to before initiating Precision Hold. Set to NAN to use current position.</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1185,6 +1185,10 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="35" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+        <description>Initiate a precision hold at the current position and maintain a distance above a target.</description>
+        <param index="1" label="Height above target">Set to zero to maintain current altitude</param>
+      </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
         <description>Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicle's control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1185,13 +1185,15 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
-      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
-        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a LANDING_TARGET.</description>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
+        <description>Vehicle moves to specified location and enables or disables Precision Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If any of the 3 fields Latitude/Longitude/Altitude are NAN the current position will be used instead.</description>
         <param index="1" label="Enable">Enable or disable Precision Hold.</param>
         <param index="2" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found, the lowest ID target available will be used.</param>
-        <param index="3" label="Stationary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
-        <param index="4" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
-        <param index="5" label="Height above target" units="m">Set to zero to maintain current altitude.</param>
+        <param index="3" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="4" label="Height above target" units="m">Set to zero to maintain current altitude.</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude">Altitude</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
         <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1197,9 +1197,9 @@
         <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
         <param index="3" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
         <param index="4">Empty</param>
-        <param index="5" label="Latitude">Latitude</param>
-        <param index="6" label="Longitude">Longitude</param>
-        <param index="7" label="Altitude">Altitude</param>
+        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME</param>
+        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME</param>
+        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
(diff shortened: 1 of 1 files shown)


dagar (approved):

dagar:
Duplicate enums with development.xml.
Screen Shot 2022-06-29 at 6 49 11 PM

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a target.</description>
```
hamishwillee, line 1189: I'd like to cross link to the LANDING_TARGET because that is what defines the ID. 
```suggestion
        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a LANDING_TARGET.</description>
```

You mentioned other things that it can do - including allowing manual control of position. That should be mentioned here.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
...
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
+        <param index="2" label="Target number">The ID of the target if multiple targets are present.</param>
+        <param index="3" label="Staionary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
+        <param index="4" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="5" label="Height above target">Set to zero to maintain current altitude.</param>
```
hamishwillee, line 1194: Add units please. My assumption is that this must be positive too, so put reasonable limits if you know them.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a target.</description>
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
+        <param index="2" label="Target number">The ID of the target if multiple targets are present.</param>
+        <param index="3" label="Staionary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
+        <param index="4" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
```
hamishwillee, line 1193: These look like good candidates for a bitmask enum.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a target.</description>
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
+        <param index="2" label="Target number">The ID of the target if multiple targets are present.</param>
+        <param index="3" label="Staionary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
```
hamishwillee, line 1192: So the two cases are:
1. This is set to 1 so the first landing target message you get with the right ID is used to provide the position you use going forward?
2. This is set to 0, so you continue to track new LANDING_TARGET data.

Is that right?

message_definitions/v1.0/common.xml
dakejahl: ahh good idea! This would allow using a LLA to make this coordinate based. eg travel to the LLA before performing the action. Thoughts?

hamishwillee: LLA is?

hamishwillee: Assuming you mean setting a broad target location first then yes.

message_definitions/v1.0/common.xml
dakejahl: **Target Number** corresponds to LANDING_TARGET.target_num
**Stationary** just says whether or not we should follow the `target_num` if it starts to move... but this implies that the autopilot has logic to track the targets motion over time... which might not be true or feasible

message_definitions/v1.0/common.xml
dakejahl: latitude/longitude/altitude. Yes the idea is you would travel to a location (LLA) and then initiate the Precision Hold. The Precision Hold will behave similar to Precision Land where it will wait for X seconds to see if it starts receiving LANDING_TARGET messages and then after those X seconds elapse it will "search". In PX4 the "search" for precision land just involves adjusting the altitude, but in the future it could do some kind of spiral pattern or something. This would all be defined by parameters in the flight stack.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a LANDING_TARGET.</description>
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
+        <param index="2" label="Target number">The ID of the target if multiple targets are present.</param>
```
hamishwillee, line 1191:
```suggestion
        <param index="2" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present.</param>
```

message_definitions/v1.0/common.xml
hamishwillee: Re "stationary" - I get it. For some reason this makes me a little uncomfortable, but I can't articulate it so no need to change.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
...
+        <param index="3" label="Staionary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
```
hamishwillee, line 1192:
```suggestion
        <param index="3" label="Stationary">Set to 1 if the target is expected to remain stationary. The vehicle should not follow the target if it moves.</param>
```

message_definitions/v1.0/common.xml
hamishwillee: I would prefer this because it aligns the cases more closely and allows us to clarify common handling. We should still in THIS message note how we handle the case of the specified ID not being found. That would go in the description.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
+        <description>Enable or disable a Precision Hold which allows the vehicle to hover precisely above a LANDING_TARGET.</description>
```
hamishwillee, line 1189: So we need more info here about what precision hold means.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,14 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="false" isDestination="false">
```
hamishwillee, line 1188: PS I prefer "HOLD" over loiter too. Loiter means "hang out in an area", so if you named this loiter you'd be opening up the semantics that maybe an MC might simply be holding in the approximate area and not attempting to hold a precise position.

Very much an aside but we went through this kind of thinking when we were deciding the standard flight modes - so we don't use "position mode" for FW but instead https://mavlink.io/en/messages/development.html#MAV_STANDARD_MODE_CRUISE - if you see what I mean.

message_definitions/v1.0/common.xml
dakejahl: Honestly it makes me uncomfortable too. I think this expects too much from the autopilot. For many cases you probably do want to follow a moving target, I am feeling inclined to remove the field...

message_definitions/v1.0/common.xml
dakejahl: >I would prefer this because it aligns the cases more closely and allows us to clarify common handling

To clarify, you would prefer having this command be coordinate based? So adding Lat/Lon/Alt fields?

message_definitions/v1.0/common.xml
hamishwillee: It's OK to ask impossible things of the autopilot. The autopilot is supposed to reject those things it can't handle with the appropriate ack.

I think that if we want to specify this we should be very precise "Stationary target. 0 (default): track moving targets. 1: target latched to the position in the first LANDING_TARGET it detects with the required target_num."

message_definitions/v1.0/common.xml
hamishwillee: Yes. (sorry)

hamishwillee: That is a preference. You'll have to implement and then you'd find out whether it made sense or not.

message_definitions/v1.0/common.xml
dakejahl: >At what altitude will it hold? Is that fixed

See parameter usage

>Can it be moved from that position using RC? You indicated it could in your description in this issue.

I did not. Where did I indicate that? This is an AUTO mode just like Land/RTL/Reposition. But I would probably say the behavior is autopilot specific anyway

>What if it can't find the indicated target - what will it do?

See parameter usage

>What if you disable precision hold when in precision hold mode? Does it go into normal hold?

Yes I would say go into Hold mode but that is probably autpilot specific.

I guess I am confused about how verbose we want to be in the description, since all of the info is covered if you read the entire message definition. If you look at the descriptions for many other commands they are very short

```
      <entry value="192" name="MAV_CMD_DO_REPOSITION" hasLocation="true" isDestination="true">
        <description>Reposition the vehicle to a specific WGS84 global position.</description>
        
      <entry value="20" name="MAV_CMD_NAV_RETURN_TO_LAUNCH" hasLocation="false" isDestination="false">
        <description>Return to launch location</description>
        
      <entry value="21" name="MAV_CMD_NAV_LAND" hasLocation="true" isDestination="true">
        <description>Land at location.</description>               
```

message_definitions/v1.0/common.xml
dakejahl: I think this might be better left to the autopilot to decide, rather than the message telling it what to do, right? There is a parameter in PX4 for the landing target estimator which specifies whether the target is expected to be stationary or moving, so this might suffice. Honestly this is outside of my specific use case so I haven't given it a lot of thought.
http://docs.px4.io/master/en/advanced_config/parameter_reference.html#LTEST_MODE

dakejahl: I removed it for now to make room for Lat/Lon/Alt. If we decide to keep it I like your suggestion of a bitmask for binary choices like stationary/enable/use_heading/etc

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,16 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
+        <description>Vehicle moves to specified location and enables or disables Precision Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If any of the 3 fields Latitude/Longitude/Altitude are NAN the current position will be used instead.</description>
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
```
hamishwillee, line 1190: So 1 is enable and 0 is disable?

message_definitions/v1.0/common.xml
hamishwillee: Re

> Can it be moved from that position using RC? You indicated it could in your description in this issue.

Sorry - I completely misinterpreted "Land except instead of landing the vehicle will control XY (and Z if specified) to hover precisely above a target.". Dunno how.

hamishwillee: > I guess I am confused about how verbose we want to be in the description, since all of the info is covered if you read the entire message definition. 

If you look at all these examples they are unambiguous references to well defined terms. But a "Precision Hold" is a new term so it kind of needs a little more. Perhaps I am overthinking this.

message_definitions/v1.0/common.xml
hamishwillee: Note, it should perhaps be up to the autopilot to decide what disabling means. But consider normally you would not disable mode - you just set another mode. Further, if something exits a mode and does switch mode, the default mode it tends to go into is Hold mode.

So to me it seems a valid question to ask "what do you think the autopilot will do". My guess is this would put it into Hold mode. I wouldn't even provide a disable option, freeing up that parameter too.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,16 @@
...
+        <param index="2" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found, the lowest ID target available will be used.</param>
+        <param index="3" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="4" label="Height above target" units="m">Set to zero to maintain current altitude.</param>
```
MaEtUgR, line 1193: Wouldn't NAN already be the way to specify holding the current altitude?

message_definitions/v1.0/common.xml
MaEtUgR: I'd interpret it as 0 keeps the heading at the time of command 🤔

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,16 @@
...
+        <param index="5" label="Latitude">Latitude</param>
```
MaEtUgR, line 1194: Even though most initial use cases are global coordinates wouldn't you want to allow specifying the frame as in e.g. the Orbit command or does that not make sense to support?

message_definitions/v1.0/common.xml
hamishwillee: Yes, if 0 is a valid value for a target height.

message_definitions/v1.0/common.xml
hamishwillee: @MaEtUgR Did you mean this for param 3? Ideally anything to do with headings should all be set in one param.

I _think_ this param (1) should be removed because if you have precision hold I think you want it used, and if you don't then you just turn off the target signal and the system should revert to normal hold behaviour. Or you change the mode.

IF there is a reason to have the ability to disable the mode then I would do it like this:

```suggestion
        <param index="1" label="Enable">Enable Precision Hold mode. 1: Enable mode, 0: disable mode. If precision hold mode is disabled when active the vehicle should enter (normal) hold mode. If the mode is inactive, the command should be temporarily rejected.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,16 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
+        <description>Vehicle moves to specified location and enables or disables Precision Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If any of the 3 fields Latitude/Longitude/Altitude are NAN the current position will be used instead.</description>
```
hamishwillee, line 1189: Lat, Lon, can't be NaN - use max value for field. Is it possible you will want to use the current altitude as default, but still specify position?

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
        <param index="4" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,16 @@
...
+        <param index="1" label="Enable">Enable or disable Precision Hold.</param>
+        <param index="2" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found, the lowest ID target available will be used.</param>
```
hamishwillee, line 1191: I think "If the specified target is not found, the lowest ID target available will be used." is risky in theory and kind of negates the point of precision hold (though perhaps not in practice). 

How about:
```suggestion
        <param index="2" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
```

message_definitions/v1.0/common.xml
MaEtUgR: > Yes, if 0 is a valid value for a target height.

Independently of that I would just assume from reading the beginning of the message description that NAN is used to maintain the current altitude. I see you got it.

message_definitions/v1.0/common.xml
MaEtUgR: Ah now I see because of COMMAND_INT limitations it has to be integer representable 💡

message_definitions/v1.0/common.xml
MaEtUgR: @hamishwillee Yes, for heading I misread the line. I agree with your first comment that switching out is done by another command. I would be interested in what the advantage of this dedicated parameter would be other than just being able to send the same command instead of a pause or hold position.

dakejahl: I don't disagree, I think falling back to regular Hold if no target is present is fine, I also think sending a mode change to Hold is a perfectly acceptable way to turn it off. I will remove the parameter.

message_definitions/v1.0/common.xml
dakejahl: >@dakejahl My understanding of this, - just to be clear, based on precision land:

Yes you've got it 100% correct

>Is there some notion of a timeout at which it will stop looking?

This will probably need to come in the form of a parameter like PLD_SRCH_TOUT

message_definitions/v1.0/common.xml
dakejahl: >Is it possible you will want to use the current altitude as default, but still specify position?

Yes you may want to just descend first (so that you can see your target) before activating the mode.

message_definitions/v1.0/common.xml
dakejahl: If you can suggest a more comprehensive explanation I am all for it :grinning:

message_definitions/v1.0/common.xml
dakejahl: This is a bit of a tangent, but what's up with command int and command long? Why are there two? And why do they both use MAV_CMD but one has additional fields (such as mav_frame)? Can some commands only be used with long/int? How do you know if you need to use long or need to use int?

hamishwillee: @dakejahl This predates me, but my understanding is that the command_long came first. Then someone realised that if you just shove a float in the lat/lon there are scaling errors of the order of several metres. So they invented COMMAND_INT using E7 scaled lat/lon and added a frame field too, to make it more flexible. 

If you're sending a position, use the COMMAND_INT. Otherwise its up to you, but you do need to think about not defining anything in param 5,6 that might not make sense in the other type of message. I always use COMMAND_INT.

Yes this is crap.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,16 @@
...
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
+        <description>Vehicle moves to specified location and engages Precison Hold mode which allows the vehicle to hover precisely above a LANDING_TARGET. If Latitude/Longitude are FLT_MAX (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current position will be used instead. If Altitude is FLT_MAX (COMMAND_LONG) or INT32_MAX (COMMAND_INT) the current altitude will be used instead </description>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
+        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="3" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
```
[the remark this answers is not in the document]
dakejahl: It's different. Altitude (param7) would be the altitude you adjust to before engaging the mode -- you may need to descend a bit before engaging to ensure the target is close enough to see. Height above target is the distance offset above the target at which you hold. So it's just like precision land, but instead of landing you hover at this distance above the target.

message_definitions/v1.0/common.xml
dakejahl: I think we should add XYZ as alternate definitions for Param5/6/7 since COMMAND_INT has MAV_FRAME and this is a position based MAV_CMD. I don't have a use for it but others in the future may.

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: Apologies, did not get to review this. It is still in draft - does that mean this can wait a little?

I also want to re-engage on batteries.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,22 @@
...
+          If Altitude is NaN the vehicle altitude when the message is received will be used.
+        </description>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
+        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="3" label="Height above target" units="m">Set to NaN to maintain current altitude.</param>
+        <param index="4">Empty</param>
+        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME</param>
+        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME</param>
+        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME</param>
+      </entry>
```
hamishwillee, line 1203: @dakejahl I think this is what you mean right?:

```suggestion
      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
        <description>
          Vehicle moves to specified location (lat/lon/alt) and engages Precision Hold mode (NaN/INT32_MAX may be used to set the current location when the message is received).
          If it gets LANDING_TARGET messages for the indicated landing target it: adjusts its horizontal position to align precisely above the target, adjusts its altitude to the Hold altitude (param3), and then holds the position. 
          If the specified target is not found the vehicle holds the specified lat/lon/alt and enters normal Position Hold mode (the timeout for entering position hold is flight-stack specific).
        </description>
        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
        <param index="3" label="Hold altitude" units="m">Altitude above target to hold position. Set to NaN to maintain current altitude.</param>
        <param index="4">Empty</param>
        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN to use current altitude.</param>
      </entry>
```

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: OK, not clear from the docs. MOdified  in https://github.com/mavlink/mavlink/pull/1858/files#r953377991

message_definitions/v1.0/common.xml
hamishwillee: PS Let me know when you want this out of draft so we can get more eyes on it.

hamishwillee: @dakejahl Does above suggestion ^^^ do what you meant?

dakejahl:
@hamishwillee Yes this looks good to go from my end

hamishwillee:
@dakejahl I presume you intend to implement in PX4/QGC? Note that we don't take things into common until there is an implementation in one flight stack and intent to adopt in another. So this will have to move into development.xml. But we can iterate here for now.

@julianoes @auturgy @MaEtUgR This PR is now IMO in a good state for review. Can you please have a look?

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,20 @@
...
+        </description>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
+        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="3" label="Hold altitude" units="m">Altitude above target to hold position. Set to NaN to use current altitude as the Hold altitude.</param>
+        <param index="4">Empty</param>
+        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
+        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
+        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN to use current altitude.</param>
```
julianoes, line 1200: So this is the altitude of the target? Maybe this could be clarified:

```suggestion
        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME of the target. NaN to use current altitude.</param>
```

julianoes (approved):

message_definitions/v1.0/common.xml
dakejahl: No this is just the LLA or XYZ position to move to before performing the action

julianoes: So why is there also a hold altitude? I'm confused.

dakejahl: Okay I updated the description. Does it make more sense now?

dakejahl:
> Note that we don't take things into common until there is an implementation in one flight stack and intent to adopt in another

There's already an implementation in PX4 I just need to make a PR. There's also already an implementation in Ardupilot ([precision loiter](https://ardupilot.org/copter/docs/precision-landing-with-irlock.html)) but it's missing mavlink support.

message_definitions/v1.0/common.xml
hamishwillee: No. Much less. Let me have a go.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,20 @@
...
+        </description>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num) if multiple targets are present. If the specified target is not found the vehicle will use normal Position Hold mode. 0 to use first target found.</param>
+        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="3" label="Hold distance" units="m">Distance above target to hold position. NaN to use current altitude.</param>
+        <param index="4">Empty</param>
+        <param index="5" label="Latitude/X">Latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
+        <param index="6" label="Longitude/Y">Longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
+        <param index="7" label="Altitude/Z">Altitude (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN to use current altitude.</param>
```
hamishwillee, line 1200: This is attempt to address @julianoes confusion. The description makes it more clear that the target of the hold is contained in the LANDING_TARGET, not param 5, 6, 7. Param 5, 6, 7 renamed too.

```suggestion
          Precision hold over a LANDING_TARGET (param1) at target-relative hold altitude (param3)
          The vehicle first moves to the initial location (param5, 6, 7) and then engages Precision Hold mode (Set initial location to NaN/INT32_MAX to engage Precision Hold at the current position).
          The vehicle waits at the initial location for LANDING_TARGET messages with the indicated id (param1).
          If found, it then aligns with the horizontal position defined in the LANDING_TARGET at a target-relative altitude specified in the Hold altitude field (param3).
          If the specified target is not found within a timeout the vehicle enters normal Position Hold mode and holds its current position (the timeout is flight-stack specific).
        </description>
        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num). 0 to use first target found. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
        <param index="3" label="Hold altitude" units="m">Altitude above target to hold position. NaN to hold at current altitude.</param>
        <param index="4">Empty</param>
        <param index="5" label="Initial Latitude/X">Latitude before entering Position Hold mode (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
        <param index="6" label="Longitude/Y">Longitude before entering Position Hold mode (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
        <param index="7" label="Initial Altitude/Z">Altitude before entering Position Hold mode (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN to use current altitude.</param>
```

hamishwillee: @julianoes Better?

message_definitions/v1.0/common.xml
julianoes: Ok, so it wasn't just me.

message_definitions/v1.0/common.xml
julianoes: So, I guess my question is this: when a vehicle gets this command, does it know the altitude of the target?

Because from what I read here, this command gives you the altitude to fly at, and the altitude to hold once found but not the altitude of the target.

dakejahl: >So, I guess my question is this: when a vehicle gets this command, does it know the altitude of the target?

It's specified in the [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) message
image

dakejahl: It can also be dervied from the **z** field in the message
```
Z Position of the landing target in MAV_FRAME
```

julianoes: I see. That makes sense then.

hamishwillee: Cool. To further reduce the confusion, I made the first sentence very very very clear :-)

> Precision hold precisely over the 3d position defined in a particular LANDING_TARGET message (param1) at a hold altitude (param3) relative to the target.

And what will make this even clearer is that when this renders it will link to the LANDING_TARGET message

Better @julianoes ?

julianoes: I think that's clearer now, thanks!

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,22 @@
...
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
+        <description>
+          Precision hold precisely over the 3d position defined in a particular LANDING_TARGET message (param1) at a hold altitude (param3) relative to the target.
+          The vehicle first moves to the initial location (param5, 6, 7) and then engages Precision Hold mode (Set initial location to NaN/INT32_MAX to engage Precision Hold at the current position).
+          The vehicle waits at the initial location for LANDING_TARGET messages with the indicated id (param1).
+          If found, it then aligns with the horizontal position defined in the LANDING_TARGET at a target-relative altitude specified in the Hold altitude field (param3).
+          If the specified target is not found within a timeout the vehicle enters normal Position Hold mode and holds its current position (the timeout is flight-stack specific).
```
hamishwillee, line 1194:
```suggestion
          If the specified target is not found within a (flight-stack specific) timeout the vehicle enters normal Position Hold mode and holds its current position.
```

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,22 @@
...
+          If the specified target is not found within a (flight-stack specific) timeout the vehicle enters normal Position Hold mode and holds its current position.
+        </description>
+        <param index="1" label="Target number">The ID of the target (LANDING_TARGET.target_num). 0 to use first target found. If the specified target is not found the vehicle will use normal Position Hold mode.</param>
+        <param index="2" label="Use Heading">Set to 1 if the drone should align it's heading with the target heading.</param>
+        <param index="3" label="Hold altitude" units="m">Altitude above target to hold position. NaN to hold at current altitude.</param>
+        <param index="4">Empty</param>
+        <param index="5" label="Initial Latitude/X">Latitude before entering Position Hold mode (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current latitude.</param>
+        <param index="6" label="Longitude/Y">Longitude before entering Position Hold mode (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
```
dakejahl, line 1201:
```suggestion
        <param index="6" label="Initial Longitude/Y">Longitude before entering Position Hold mode (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. NaN (COMMAND_LONG) or INT32_MAX (COMMAND_INT) to use current longitude.</param>
```
Why was "initial" left out here?

hamishwillee: Because I suck? Good catch.

message_definitions/v1.0/common.xml
```diff
@@ -1185,6 +1185,22 @@
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="36" name="MAV_CMD_DO_PRECISION_HOLD" hasLocation="true" isDestination="true">
+        <description>
+          Precision hold precisely over the 3d position defined in a particular LANDING_TARGET message (param1) at a hold altitude (param3) relative to the target.
```
hamishwillee, line 1190:
```suggestion
          Precision hold over the 3d position defined in a particular LANDING_TARGET message (param1) at a hold altitude (param3) relative to the target.
```

dakejahl:
@akkawimo

PR #1861, opened 2022-06-28
common: ODOMETRY add optional quality field

dagar:
This adds an optional quality percentage to the ODOMETRY message. In some cases we don't have usable covariances (or any at all).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7037,7 +7037,7 @@
       <extensions/>
       <field type="uint8_t" name="reset_counter">Estimate reset counter. This should be incremented when the estimate resets in any of the dimensions (position, velocity, attitude, angular speed). This is designed to be used when e.g an external SLAM system detects a loop-closure and the estimate jumps.</field>
       <field type="uint8_t" name="estimator_type" enum="MAV_ESTIMATOR_TYPE">Type of estimator that is providing the odometry.</field>
-      <field type="uint8_t" name="quality" units="%" invalid="0">Optional odometry quality metric as a percentage. 0 = unknown/unset quality, 1 = worst quality, 100 = best quality</field>
+      <field type="int8_t" name="quality" units="%" invalid="0">Optional odometry quality metric as a percentage. -1 = odometry has failed, 0 = unknown/unset quality, 1 = worst quality, 100 = best quality</field>
     </message>
     <message id="332" name="TRAJECTORY_REPRESENTATION_WAYPOINTS">
       <description>Describe a trajectory using an array of up-to 5 waypoints in the local frame (MAV_FRAME_LOCAL_NED).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7037,6 +7037,7 @@
       <extensions/>
       <field type="uint8_t" name="reset_counter">Estimate reset counter. This should be incremented when the estimate resets in any of the dimensions (position, velocity, attitude, angular speed). This is designed to be used when e.g an external SLAM system detects a loop-closure and the estimate jumps.</field>
       <field type="uint8_t" name="estimator_type" enum="MAV_ESTIMATOR_TYPE">Type of estimator that is providing the odometry.</field>
+      <field type="uint8_t" name="quality" units="%" invalid="0">Optional odometry quality metric as a percentage. 0 = unknown/unset quality, 1 = worst quality, 100 = best quality</field>
     </message>
     <message id="332" name="TRAJECTORY_REPRESENTATION_WAYPOINTS">
       <description>Describe a trajectory using an array of up-to 5 waypoints in the local frame (MAV_FRAME_LOCAL_NED).</description>


Jaeyoung-Lim (commented):
Could you elaborate on what we would be normalizing the covariance value into?(what would we be normalizing?)

 I am not sure if there are meaningful metrics when the covariance metric is not finite

dagar:
> Could you elaborate on what we would be normalizing the covariance value into?(what would we be normalizing?)

It doesn't necessarily have anything to do with the covariances, it's for systems that only give you some kind of quality percentage or worse (eg T265 reported 3 quality levels).

Does that make sense? Happy to change the wording to remove ambiguity.

james-modalai:
This would be an excellent feature. I believe VINS Fusion does not report a covariance, I'm sure other algorithms do not too. Furthermore, I've found covariance is not a reliable predictor of when an MSCKF based VIO is likely to fail. Number of tracked feature points and their distribution in the frame is much better predictor of health. Having a user-defined scalar quality metric for VIO in flight logs would be extremely helpful to review and track VIO health during flight.

hamishwillee (approved):
LGTM. @auturgy any comments/objections? The invalid case is 0 so nothing changes for anyone who chooses not to adopt this.

auturgy:
No issues

hamishwillee (approved):
@Jaeyoung-Lim Are you happy with responses? If so this can be merged.

james-modalai:
I may request that the type is int8_t instead of uint8_t. This would allow a value of -1 to indicate that the odometry algorithm has failed entirely and indicate to the subscriber to perform any required behaviors such as falling back to a different flight mode and removing odometry as an input source entirely without needing to wait for a timeout. That then allows a value of 0 to remain as indicating the field is unused, and 1-100 indicating the algorithm is functioning but at various health levels. Otherwise I don't see a clean way in this message to indicate that an odometry algorithm has failed but is still actively running and may recover in the future.

hamishwillee:
@james-modalai Well, you could use any value from 101 to uint_max8 to indicate that state. But arguably -1 is more obviously a "not valid" indicator. Upshot, doesn't matter as long as it is clear.

james-modalai:
@hamishwillee completely agree, UINT8_MAX serves the purpose just as well. Either way the field description should indicate either -1 or UINT8_MAX should be used in that scenario. Should I submit a new merge request? @dagar I don't mean to hijack your merge request :)

hamishwillee:
@james-modalai
1. In general, MAVLink doesn't differentiate between the case of "always invalid" and "currently invalid" for supplied data. So in theory you could just set 0 when your algorithm is failing. I wouldn't be surprised though if some systems latch on the first value they get and if it is "invalid" ignore the field subsequently. So perhaps it is best to be precise.
2. I would do this by adding a github suggestion to the line - then @dagar can merge it if he is happy.

Jaeyoung-Lim:
@dagar I think it makes a lot more sense now, so probably makes sense. Since actual stake holders such as @james-modalai find it useful I have no objections

Just a bit concerned that the quality metric will be something inconsistent between systems and there is no way from the end user to figure out what it means (probably also true for the T265)

message_definitions/v1.0/common.xml
```diff
@@ -7037,6 +7037,7 @@
       <extensions/>
       <field type="uint8_t" name="reset_counter">Estimate reset counter. This should be incremented when the estimate resets in any of the dimensions (position, velocity, attitude, angular speed). This is designed to be used when e.g an external SLAM system detects a loop-closure and the estimate jumps.</field>
       <field type="uint8_t" name="estimator_type" enum="MAV_ESTIMATOR_TYPE">Type of estimator that is providing the odometry.</field>
+      <field type="uint8_t" name="quality" units="%" invalid="0">Optional odometry quality metric as a percentage. 0 = unknown/unset quality, 1 = worst quality, 100 = best quality</field>
```
james-modalai, line 7040:
```suggestion
      <field type="int8_t" name="quality" units="%" invalid="0">Optional odometry quality metric as a percentage. -1 = odometry has failed, 0 = unknown/unset quality, 1 = worst quality, 100 = best quality</field>
```

james-modalai (commented):
Clarify how user should indicate odometry has failed through the quality field.

dagar:
> Just a bit concerned that the quality metric will be something inconsistent between systems and there is no way from the end user to figure out what it means (probably also true for the T265)

I share this concern, but I'm not sure what else we could do?

hamishwillee:
>> Just a bit concerned that the quality metric will be something inconsistent between systems and there is no way from the end user to figure out what it means (probably also true for the T265)

> I share this concern, but I'm not sure what else we could do?

@dagar It depends on whether there is some broad guidance/calibration you can ever give to end users - e.g. "A value over 60% generally indicates that the system is performing well. Values below 30% indicate poor performance and you should do Y.

If not, then I'd just make it clear how the field can be used:

> Note: the interpretation of this value is system-specific. It is provided primarily for log analysis and would not generally be displayed to end users.

PR #1864, opened 2022-07-09
Add HEXADECAROTOR

muramura:
Add 16 rotor definition for OCTA-X16 COPTER.
18 rotor to be added for VOLOCOPTER.

https://github.com/ArduPilot/ardupilot/issues/21154

HexadecaOcta ( OCTA-X16) Copter
Screenshot from 2022-07-14 06-05-45

VOLOCOPTER:
Screenshot from 2022-07-11 07-03-03

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -200,6 +200,12 @@
       <entry value="42" name="MAV_TYPE_WINCH">
         <description>Winch</description>
       </entry>
+      <entry value="43" name="MAV_TYPE_HEXADECAROTOR">
+        <description>Hexadecarotor</description>
+      </entry>
+      <entry value="44" name="MAV_TYPE_OCTADECAROTOR">
+        <description>Octadecarotor</description>
+      </entry>
     </enum>
     <enum name="MAV_MODE_FLAG" bitmask="true">
       <description>These flags encode the MAV mode.</description>


hamishwillee:
Discussed in the dev call. How can this work when neither ArduPilot or PX4 currently support more than 16 outputs?

rmackay9:
The latest versions of AP do support up to 32 outputs but we don't have support for this vehicle type in AP yet.

hamishwillee:
Thanks @rmackay9 I guess it is chicken and egg - you'll need this before it goes in. Any particular ETA?

peterbarker:
> Thanks @rmackay9 I guess it is chicken and egg - you'll need this before it goes in. Any particular ETA?

Well, I'm rebasing the [relevant PR](https://github.com/ArduPilot/ardupilot/pull/21157) ATM, and can't build it due to the lack of the frame definition.

Has somebody nutted out the taxonomy for these frames?

In MAVLink ArduPilot says "MAV_TYPE_OCTOROTOR" for all vehicles with 8 rotors (unless specifically specified by the user).

Internally we call a vehicle with 8 co-planar motors an "OCTA", and a vehicle with motors in two parallel  planes  an "OCTAQUAD".

Our dodecahex support seems to be limited to two planes of 6 rotors each.

So it looks like the MAVLINK MAV_TYPE standard really only cares about *number* of rotors for mutlicopters, not layout.

This is both disappointing in that it limits how well the GCS can customise the vehicle icon (and perhaps setup screens), but also makes things simpler - adding hexadecarotor as 16 seems pretty straight-forward, it really is just the rotor count, like for octo

hamishwillee:
@peterbarker No one has done further work on the naming conventions since we tidied up the convention for `MAV_TYPE_VTOL_TAILSITTER_` variants.

I think you're right - the GCS can currently only use the type for very broad differentiation - i.e. VTOL vs MC. This might then used for icons and displaying things like a VTOL transition button on VTOLS.
Looking at QGC though, it doesn't display the icon indicating vehicle type when it connects (though it may if connecting to multiples).
The only place the GCS cares about the type seems to be in airframe selection, and that is entirely a parameter/parameter metadata thing (no MAVLink direct understanding).

So my guess it that **all** these broader divisions in type may not actually be used. But we need to be sure. 

@meee1 @DonLakeFlyer 
- can you comment on exactly what a GCS uses the `MAV_TYPE_` of a vehicle for, if anything? 
- How do QGC/Mission planner differentiate between a frame configurations with the same number of rotors - e.g. an 8 rotor as an OCTO configuration vs coaxial (4 rotors aligned over 4 other rotors)?
- If we provided a convention for names of the different alignments, would that be useful? Who by?

DonLakeFlyer:
QGC mainly uses `MAV_TYPE_` to distinguish between vehicle classes. For example which types relate to a multi-rotor. This then drives differences in the UI.

As far as frame configurations the only comes into play in two places:
* Actuator setup: But this is all done through the new actuator mavlink spec stuff which includes meta data and not map types.
* Old style Motor testing: Only comes into place if new Actuator stuff is not supported. This uses motor counts which is determined from mav type. But for this octo versus coax doesn't matter. It's just physical motors.

hamishwillee:
Thanks @DonLakeFlyer 

@peterbarker This seems to confirm the likelihood that GCS do very little with this information, probably because it is hard to do much more than count motors. Upshot, you could reasonably use [MAV_TYPE_GENERIC_MULTIROTOR](https://mavlink.io/en/messages/common.html#MAV_TYPE_GENERIC_MULTIROTOR) for this case.

If you want to add an entry then I would like to use a convention, at least "going forward". This is the current set:

MAV_TYPE_QUADROTOR	Quadrotor
MAV_TYPE_COAXIAL	Coaxial helicopter
MAV_TYPE_HELICOPTER	Normal helicopter with tail rotor.
MAV_TYPE_HEXAROTOR	Hexarotor
MAV_TYPE_OCTOROTOR	Octorotor
MAV_TYPE_TRICOPTER	Tricopter
MAV_TYPE_DODECAROTOR	Dodecarotor
MAV_TYPE_DECAROTOR	Decarotor
MAV_TYPE_GENERIC_MULTIROTOR

For discussion "going forward" would be good to have
-  `mc_` and `heli` prefix
- followed by signifier for "rotors in parallel plane".  - so this would be 6 for a hexarotor and 6 for a coaxial hexarotor (even though it has twelve rotors). . I'd prefer a number for larger cases because dodeca etc are confusing
- followed by something to indicate number of layers if more than one - so coaxial? I presume that there aren't 4 layer vehicles?

MAV_TYPE_HELI  (normal, with tail)
MAV_TYPE_HELI_COAXIAL  (2 rotors aligned coaxially, counter-rotating)
MAV_TYPE MC_TRIROTOR
MAV_TYPE MC_TRIROTOR_COAXIAL (6 rotors, mounted on three points)
MAV_TYPE_MC_QUAD
MAV_TYPE_MC_QUAD_COAXIAL (8 rotor)
MAV_TYPE_MC_HEXA or MAV_TYPE_MC_6ROTOR
MAV_TYPE_MC_OCTO or MAV_TYPE_MC_8ROTOR
MAV_TYPE_MC_10ROTOR	// Note, after octa it becomes hard to work out what you're talking, so this is better.
MAV_TYPE_MC 10ROTOR_COAXIAL


Don't know if that makes sense? Are there other factors?

hamishwillee:
@peterbarker I'm going to close this based on the conclusion that it isn't needed- from the last post. 

If @DonLakeFlyer or another GCS vendor representative comes back to us and says they are going to use these values for something, then we should at that point use a convention for naming them that makes sense based on usage.

peterbarker:
On Thu, 11 Jun 2026, Hamish Willee wrote:

> @peterbarker I'm going to close this based on the conclusion that it isn't needed- from the last post.
> 
> If @DonLakeFlyer or another GCS vendor representative comes back to us and says they are going to use these values for something, then we should at that point use a convention for naming
> them that makes sense based on usage.

Well, we've moved on to dotriaconta_octaquad_x now.

Probably just need WAYTOOMANYROTORS now.

hamishwillee:
Probably. But until we have a consumer that going to use the information. This is parked.

PR #1868, opened 2022-07-21
Mission item - document values for autocontinue

hamishwillee:
Autocontinue field can be set true or false. If true when the mission item is "achieved" it will start executing the next item. If false, it will pause until explicitly commanded to proceed using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO.

This makes it clear how the command should be used - since it wasn't clear what values would cause what behaviour.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5160,7 +5160,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the mission item completes.</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
@@ -5419,7 +5419,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the mission item completes.</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true.Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5160,7 +5160,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
@@ -5419,7 +5419,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true.Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5184,7 +5184,7 @@
       <description>
         Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
         The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
-        If the mission is paused, the system will continue to the new item after it is restarted.
+        If the mission is paused, the mission may need to be restarted before continuing to the new item.
       </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5181,11 +5181,7 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>
-        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
-        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
-        If the mission is paused, it may need to be resumed before continuing.
-      </description>
+      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint16_t" name="seq">Sequence</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5182,7 +5182,7 @@
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
       <description>
-        Set the mission item with sequence number seq as the current item.
+        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
         The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
         If the mission is paused, the system will continue to the new item after it is restarted.
       </description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5160,7 +5160,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes.</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
@@ -5184,7 +5184,7 @@
       <description>
         Set the mission item with sequence number seq as the current item.
         The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
-        If the mission is paused, the system will continue to the new item after it is restarted (MAV_CMD_DO_PAUSE_CONTINUE).
+        If the mission is paused, the system will continue to the new item after it is restarted.
       </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
@@ -5423,7 +5423,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes.</field>
       <field type="float" name="param1">PARAM1, see MAV_CMD enum</field>
       <field type="float" name="param2">PARAM2, see MAV_CMD enum</field>
       <field type="float" name="param3">PARAM3, see MAV_CMD enum</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5184,7 +5184,7 @@
       <description>
         Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
         The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
-        If the mission is paused, it may need to be restarted before continuing.
+        If the mission is paused, it may need to be resumed before continuing.
       </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5184,7 +5184,7 @@
       <description>
         Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
         The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
-        If the mission is paused, the mission may need to be restarted before continuing to the new item.
+        If the mission is paused, it may need to be restarted before continuing.
       </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item.
+        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+        If the mission is paused, the system will continue to the new item after it is restarted (MAV_CMD_DO_PAUSE_CONTINUE).
+      </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint16_t" name="seq">Sequence</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5160,7 +5160,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the mission item completes.</field>
```
hamishwillee, line 5163:
```suggestion
      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5419,7 +5419,7 @@
...
```
hamishwillee, line 5422:
```suggestion
      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true.Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
```

auturgy:
I think DO_JUMP would work too

hamishwillee:
@auturgy @julianoes 

[MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) sets the current mission item to whatever you want but it should not restart a paused mission. So it isn't something you can use if autocontinue was previously set as `false`. Note you can even call this on a completed mission and it just changes the number. I am certain of this on PX4 and 99% sure on ArduPilot.

That is also probably true for  [MAV_CMD_DO_JUMP](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_JUMP). It is hard to be sure because the description implies it is used in missions. 

I will fix both those to make it more clear ^^^.

Only [MAV_CMD_OVERRIDE_GOTO](https://mavlink.io/en/messages/common.html#MAV_CMD_OVERRIDE_GOTO) and [MAV_CMD_DO_PAUSE_CONTINUE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PAUSE_CONTINUE) are correct here because they are the only ones that have the semantics of "restart" or "continue".

Make sense? If so, this OK to merge?

hamishwillee:
PS I know about the Pause already and I found override by searching when doing due diligence for this issue @julianoes :-).

julianoes:
> Note you can even call this on a completed mission and it just changes the number.

And will continue from there, right?

I'm pretty sure this is how it worked in QGC around 2013 :smile:. There was a checkbox in front of every mission item, and it would switch as it progressed, and you could move the current item by clicking on it, and from my recollection it would send [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT).

image2
(more screenshots here: https://diydrones.com/profiles/blogs/qgroundcontrol-update-mission)


> [MAV_CMD_OVERRIDE_GOTO](https://mavlink.io/en/messages/common.html#MAV_CMD_OVERRIDE_GOTO)

This is not implemented by PX4 nor ArduPilot, from what I can tell, so not sure we should be mention it. It is also sometimes confused with `DO_REPOSITION` in the  PX4 world.

> [MAV_CMD_DO_JUMP](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_JUMP)

DO_JUMP - from my understanding - is very much a command you use while doing a mission as you go through the mission items and jump around within them. It's new to me that you would use it "from the outside" to move along a mission. It can probably be implemented that way but I would probably rather not do that, given there is [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) for exactly that purpose.

hamishwillee:
>> [MAV_CMD_OVERRIDE_GOTO](https://mavlink.io/en/messages/common.html#MAV_CMD_OVERRIDE_GOTO)

> This is not implemented by PX4 nor ArduPilot, from what I can tell, so not sure we should be mention it. It is also sometimes confused with DO_REPOSITION in the PX4 world.

Can we remove it? At least deprecate? @auturgy

hamishwillee:
>> Note you can even call this on a completed mission and it just changes the number.

> And will continue from there, right?

Depends. All this should do is update the mission number. It shouldn't change the pause/running state. So if you call the method while the the mission is running it will continue at the new item. If you call it when paused it will move to the new item, but still be paused.

That's clearly the mavlink intent. If you can show that the implementation is not doing that then we have a problem.

hamishwillee:
And yes, for DO_JUMP it doesn't make so much sense out of missions, though it could used. We should be explicit.

julianoes:
> If you call it when paused it will move to the new item, but still be paused.

That makes sense. Now the question is what does it mean to have a "paused mission". I don't think there is a notion of this in PX4, it's just that we're either in mission mode or not. If we're not, then that's usually Hold/Loiter mode. QGC actually sends the command DO_REPOSITION with lat/lon NAN if I remember it right to go into Hold mode and "pause" the mission. MAVSDK just switches to Hold/Loiter mode directly.

Now the interesting part is when the mission is "finished". I think PX4 stays in mission mode, so if you then click on a previous item, it will go there instead. However, it would also be intuitive to switch to Hold mode once done which means that sending a SET_CURRENT would then not have an effect.

hamishwillee:
@julianoes "hold-as-pause" is a valid alternative implementation, provided the system can comply with the mavlink interface. 

My take is that SET_CURRENT on PX4 should set the current mission item but not change the mode. If it does something different it would be non-compliant.
- Going into hold mode for pause always going to be a bit complex because it complicates some edge cases.

I think going into hold on mission completion is fine too, and might be more consistent. But even with this what I'm not so fine with is the fact that:
- I don't think we signal mission completeness:
- I don't think we can query for mission completeness
- We don't have a way to command to fully reset a mission. Currently this is done on disarm/arm. 

So going into hold immediately at the end of a mission might be obvious, but the reason that trying to put a complete mission back to running fails might not be obvious to a user. So we would need to provide a reason for this.

The discussion in https://github.com/mavlink/mavlink/pull/1867 about how to implement in-mission land, disarm, arm, continue is relevant to this.

Does any of this change the PR being correct? (genuine question)?

julianoes:
> I don't think we signal mission completeness:

We somewhat do by sending out the current index. If the current index is equal to the size, or -1, that can be used as a signal that it's done.

> We don't have a way to command to fully reset a mission. Currently this is done on disarm/arm.

You can reset it by setting it to item 0. However, this does not reset DO_JUMPS with counters that go down every time they are executed.

You can also reset it by uploading it again. Disarm only sets it back to 0, in the same way as when you do that manually.

Regarding the PR: I think it's fine except that MAV_CMD_OVERRIDE_GOTO is not implemented, so I wouldn't recommend it. MAV_CMD_DO_PAUSE_CONTINUE might work but then I'd also mention MISSION_SET_CURRENT.

hamishwillee:
Hmmm. So MAV_CMD_OVERRIDE_GOTO  is not implemented, but you never know what is implemented in mavlink. I can't recommend MISSION_SET_CURRENT because it **shouldn't** work for continuing (as per this discussion). If it does it is a bug.

I think it is worth having a simpler way to reset the mission. Uploading a whole mission is costly, and resetting to item 0 is not resetting if it doesn't fix the counter.

julianoes:
> [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) sets the current mission item to whatever you want but it should not restart a paused mission. So it isn't something you can use if autocontinue was previously set as false. Note you can even call this on a completed mission and it just changes the number. I am certain of this on PX4 and 99% sure on ArduPilot.

I don't think you're right here. So we have to try it out for both, in my opinion. In PX4 there is no notion of a paused mission. You're either in mission mode and so changing the set current, would jump to that item and go there, or you're not in mission but Hold and yes, it will switch it but not do anything.

For ArduPilot I don't know.

hamishwillee:
@julianoes Fortunately PX4 does as MAVLink implies - [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) just changes the waypoint number - it doesn't continue-by-default if paused. Verified on current master using QGC and also using Pymavlink to change the mission number.

I'll remove overridegoto though. We can confirm this is OK with James this evening.

message_definitions/v1.0/common.xml
```diff
@@ -5160,7 +5160,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
```
hamishwillee, line 5163:
```suggestion
      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5419,7 +5419,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true.Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE or MAV_CMD_OVERRIDE_GOTO).</field>
```
hamishwillee, line 5422:
```suggestion
      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
```

julianoes:
@hamishwillee and how did you pause the mission? Using `autocontinue` set to `false` or clicking Pause in QGC?

hamishwillee:
@julianoes So now we are in violent agreement, this makes sense as documented right?

julianoes:
@hamishwillee yes and no. The comment is now technically not wrong in terms of MAVLink spec, however, it's confusing when you would apply it to PX4 as MAV_CMD_DO_PAUSE_CONTINUE is not implemented and the way to get it to "continue" is to use SET_MISSION_CURRENT to the next one.

hamishwillee:
@julianoes 

What I think you are saying is that 
- when PX4 is put into "Pause" QGC is actually calling [MAV_CMD_DO_REPOSITION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION) (presumably specifying [MAV_DO_REPOSITION_FLAGS_CHANGE_MODE](https://mavlink.io/en/messages/common.html#MAV_DO_REPOSITION_FLAGS_CHANGE_MODE) flag to set into Hold mode?)
- To continue QGC just moves PX4 back into into mission mode (presumably calling [MAV_CMD_DO_SET_MODE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MODE)?). At this point the system continues.
- [MAV_CMD_DO_PAUSE_CONTINUE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PAUSE_CONTINUE) is not implemented
- [MAV_CMD_DO_SET_MISSION_CURRENT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MISSION_CURRENT) is also not implemented FYI.

Then you said 

> the way to get it to "continue" is to use SET_MISSION_CURRENT to the next one.

Testing shows this does not work. If you're in hold mode you can change the current waypoint without changing into mission mode by calling SET_MISSION_CURRENT. So SET_MISSION_CURRENT  does not cause things to continue. I "think" only changing back into the mission mode will restart it. Unfortunately I'm unable to verify the info flow on window because QGC isn't letting me upload a mission. 

Once I understand how PX4 works I can comment on how this might be updated. However my gut feeling is that PX4 might be technically compliant with the spec but is woefully implementing it. Who would expect do_reposition to be the method you use to pause?

julianoes:
> when PX4 is put into "Pause" QGC is actually calling [MAV_CMD_DO_REPOSITION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION) (presumably specifying [MAV_DO_REPOSITION_FLAGS_CHANGE_MODE](https://mavlink.io/en/messages/common.html#MAV_DO_REPOSITION_FLAGS_CHANGE_MODE) flag to set into Hold mode?)

Correct.

> To continue QGC just moves PX4 back into into mission mode (presumably calling [MAV_CMD_DO_SET_MODE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MODE)?). At this point the system continues.

That's my guess, yes.

> [MAV_CMD_DO_PAUSE_CONTINUE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PAUSE_CONTINUE) is not implemented

Correct.

> [MAV_CMD_DO_SET_MISSION_CURRENT](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MISSION_CURRENT) is also not implemented FYI.

Oh, I missed this one. Yes, it doesn't look like it's implemented. I guess that's almost the same as [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) except that it has proper acks/nacks. So we should probably deprecate [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) which is of course the one implemented everywhere.

Here is where it was added: https://github.com/mavlink/mavlink/pull/940 We basically added it without discouraging or deprecating the old one that does the same thing.

> Testing shows this does not work. If you're in hold mode you can change the current waypoint without changing into mission mode by calling SET_MISSION_CURRENT. So SET_MISSION_CURRENT does not cause things to continue. I "think" only changing back into the mission mode will restart it. Unfortunately I'm unable to verify the info flow on window because QGC isn't letting me upload a mission.

This does work for "pause with autocontinue=false" but not for paused to a different mode which makes sense to me.

To test it:
1. Create a mission.
2. Save the .plan file
3. Edit the .plan file and change one autoContinue to false
4. Load the .plan file
5. Upload the mission.

Oh, you can't upload missions at all :astonished:, hmmm :thinking:... You could use mavsdk :)

> Who would expect do_reposition to be the method you use to pause?

That's wrong, I think, but it doesn't matter because otherwise "pause" would just translate to "switch mode to Loiter/Hold", same outcome.

hamishwillee:
Thanks @julianoes 

> This does work for "pause with autocontinue=false" but not for paused to a different mode which makes sense to me.

How can that make sense???? The semantics everywhere else are that this just sets the sequence - independent of all other settings - i.e. you can set the current index when you're landed, or when you're completed and that works fine. Why should the one case of `autocontinue=false` have additional semantics such that setting the value also does something unexpected.
Further, if you accept that pausing something in one case should explicitly required restarting, then how is the other case different ?

I can't verify what ArduPilot does - it appears to ignore the autocontinue setting altogether.

Perhaps we can let the old version behave oddly and fix up the new command?

Another question on this. What should happen if the value is set to the same sequence ID? 
- it could follow the same codepaths but there would be no side effects - no change to the current number so MISSION_CURRENT not emitted etc.
- It could reject the command immediately.
- 
What I'm kind of getting at is that I expect MISSION_CURRENT to be emitted when the number changes. But it could be that it also gets emitted when you call this command.

julianoes:
> How can that make sense???? The semantics everywhere else are that this just sets the sequence

It does "just set the sequence". But when you're in an active mission stuck at autocontinue=false, then changing the sequence (or in this case it could be "manually setting it to next", it means the mission will now continue again.

> What I'm kind of getting at is that I expect MISSION_CURRENT to be emitted when the number changes. But it could be that it also gets emitted when you call this command.

It gets emitted at a regular rate, so yes it's a bit odd. Which is why the command would be more appropriate.

hamishwillee:
> It does "just set the sequence". But when you're in an active mission stuck at autocontinue=false, then changing the sequence (or in this case it could be "manually setting it to next", it means the mission will now continue again.

I understand what it does thanks. I just think it's inconsistent and therefore likely unexpected. Not technically a spec violation though, since the spec was not written clearly enough. I will think on this now we have a more full picture of what things are doing.

> It gets emitted at a regular rate, so yes it's a bit odd. Which is why the command would be more appropriate.

I think it probably should be emitted after the command is sent too. But not at all sure if this should emit only if the number changed. Also it is a spec change to say this so I think I'll leave that aspect for now.

julianoes:
> I think it probably should be emitted after the command is sent too.

That as well. I think that's the behavior.

message_definitions/v1.0/common.xml
```diff
@@ -5160,7 +5160,7 @@
       <field type="uint8_t" name="frame" enum="MAV_FRAME">The coordinate system of the waypoint.</field>
       <field type="uint16_t" name="command" enum="MAV_CMD">The scheduled action for the waypoint.</field>
       <field type="uint8_t" name="current">false:0, true:1</field>
-      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint</field>
+      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes (after which it can be restarted using MAV_CMD_DO_PAUSE_CONTINUE).</field>
```
hamishwillee, line 5163: Removing suggestion on _how_ it can be restarted. At the moment PX4 will also restart if you call `SET_MISSION_CURRENT` following an auto-continue pause. There is no clear story, so for now removing.
```suggestion
      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5419,7 +5423,7 @@
...
```
hamishwillee, line 5426:
```suggestion
      <field type="uint8_t" name="autocontinue">Autocontinue to next waypoint. 0: false, 1: true. Set false to pause mission after the item completes.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item.
+        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+        If the mission is paused, the system will continue to the new item after it is restarted (MAV_CMD_DO_PAUSE_CONTINUE).
```
hamishwillee, line 5187: This is the problematic case. I don't want to say calling `MISSION_SET_CURRENT` will restart the the mission because:
- ArduPilot will accept this to restart a mission in Plane, but not in other frames.
- PX4 will accept this to restart a mission from auto-continue = false pause but not from a change mode pause. 

So the implementations are inconsistent. What I'd like to do is deprecate this message and make the behaviour consistent in the set current mission command.

For now, the suggestion is:

```suggestion
        If the mission is paused, the system will continue to the new item after it is restarted.
```

hamishwillee:
@julianoes OK, I have removed the mention of how this might be paused/restarted.

I'm still concerned about this, but perhaps we can agree consistent behaviour in https://github.com/mavlink/mavlink/pull/1870 and deprecate SET_MISSION_CURRENT.

IN any case this should be "OK" now.

message_definitions/v1.0/common.xml
```diff
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item.
```
hamishwillee, line 5185:
```suggestion
        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
```

message_definitions/v1.0/common.xml
```diff
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+        If the mission is paused, the system will continue to the new item after it is restarted.
```
hamishwillee, line 5187:
```suggestion
        If the mission is paused, the mission may need to be restarted before continuing to the new item.
```

message_definitions/v1.0/common.xml
```diff
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+        If the mission is paused, the mission may need to be restarted before continuing to the new item.
```
hamishwillee, line 5187:
```suggestion
        If the mission is paused, it may need to be restarted before continuing.
```

message_definitions/v1.0/common.xml
```diff
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+        If the mission is paused, it may need to be restarted before continuing.
```
julianoes, line 5187:
```suggestion
        If the mission is paused, it may need to be resumed before continuing.
```

:thinking:

julianoes (approved):

message_definitions/v1.0/common.xml
```diff
@@ -5181,7 +5181,11 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <description>
+        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+        The system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+        If the mission is paused, it may need to be resumed before continuing.
+      </description>
```
hamishwillee, line 5188: Revert MISSION_SET_CURRENT change.

```suggestion
      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
```

hamishwillee:
@julianoes I have reverted the change to `SET_MISSION_CURRENT` and will move that to be part of the discussion with the command - they should all be agreed together.

That leaves this with just defining what Autocontinue should do. That's crystal clear, in particular since ArduPilot don't implement it.

Merging :-)

PR #1869, opened 2022-08-03
common: extend MISSION_CURRENT

julianoes:
The message MISSION_CURRENT is a bit limited as it sends the current mission item sequence but no context.

My proposition would be to include the total as well as the status of an ongoing mission.

Related to https://github.com/mavlink/mavlink/pull/1868.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4851,28 +4851,28 @@
         <description>NAK: NAK response</description>
       </entry>
     </enum>
-    <enum name="MISSION_FSM_STATE">
+    <enum name="MISSION_STATE">
       <description>
-        States of the mission finite state machine.
+        States of the mission state machine.
         Note that these states are independent of whether the mission is in a mode that can execute mission items or not (is suspended).
         They may not all be relevant on all vehicles.
       </description>
-      <entry value="0" name="MISSION_FSM_STATE_UNKNOWN">
+      <entry value="0" name="MISSION_STATE_UNKNOWN">
         <description>The mission status reporting is not supported.</description>
       </entry>
-      <entry value="1" name="MISSION_FSM_STATE_NO_MISSION">
+      <entry value="1" name="MISSION_STATE_NO_MISSION">
         <description>No mission on the vehicle.</description>
       </entry>
-      <entry value="2" name="MISSION_FSM_STATE_NOT_STARTED">
+      <entry value="2" name="MISSION_STATE_NOT_STARTED">
         <description>Mission has not started. This is the case after a mission has uploaded but not yet started executing.</description>
       </entry>
-      <entry value="3" name="MISSION_FSM_STATE_ACTIVE">
+      <entry value="3" name="MISSION_STATE_ACTIVE">
         <description>Mission is active, and will execute mission items when in auto mode.</description>
       </entry>
-      <entry value="4" name="MISSION_FSM_STATE_PAUSED">
+      <entry value="4" name="MISSION_STATE_PAUSED">
         <description>Mission is paused when in auto mode.</description>
       </entry>
-      <entry value="5" name="MISSION_FSM_STATE_COMPLETE">
+      <entry value="5" name="MISSION_STATE_COMPLETE">
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
@@ -5220,8 +5220,8 @@
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
-      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE" invalid="0">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
-      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
+      <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="mission_mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
     </message>
     <message id="43" name="MISSION_REQUEST_LIST">
       <description>Request the overall list of mission items from the system/component.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4862,13 +4862,16 @@
       <entry value="2" name="MISSION_STATUS_ACTIVE">
         <description>The mission is running and active.</description>
       </entry>
-      <entry value="3" name="MISSION_STATUS_ACTIVE_NOT_IN_MISSION_MODE">
-        <description>The mission would be running but we're not in mission mode.</description>
+      <entry value="3" name="MISSION_STATUS_PAUSED_MANUAL">
+        <description>The mission has been paused by the a user command or by changing the flight mode.</description>
       </entry>
-      <entry value="4" name="MISSION_STATUS_PAUSED">
-        <description>The mission has been paused with autocontinue=false.</description>
+      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
+        <description>The mission has been automatically paused due to completion of an item with autocontinue=false.</description>
       </entry>
       <entry value="5" name="MISSION_STATUS_FINISHED">
+        <description>The Mission has been finished.</description>
+      </entry>
+      <entry value="6" name="MISSION_STATUS_NO_MISSION_UPLOADED">
         <description>The Mission has been finished.</description>
       </entry>
     </enum>
@@ -5208,10 +5211,14 @@
       <field type="uint16_t" name="seq">Sequence</field>
     </message>
     <message id="42" name="MISSION_CURRENT">
-      <description>Message that announces the sequence number of the current active mission item. The MAV will fly towards this mission item.</description>
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
-      <field type="uint16_t" name="total">Total number of mission items</field>
+      <field type="uint16_t" name="total">Total number of mission items. 0 if no mission is present on the vehicle.</field>
       <field type="uint8_t" name="status" enum="MISSION_STATUS">Mission status</field>
     </message>
     <message id="43" name="MISSION_REQUEST_LIST">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5220,7 +5220,7 @@
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
-      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE" invalid="0">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
       <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
     </message>
     <message id="43" name="MISSION_REQUEST_LIST">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4862,11 +4862,11 @@
       <entry value="2" name="MISSION_STATUS_ACTIVE">
         <description>The mission is running and active.</description>
       </entry>
-      <entry value="3" name="MISSION_STATUS_PAUSED_MANUAL">
-        <description>The mission has been paused by the a user command or by changing the flight mode.</description>
+      <entry value="3" name="MISSION_STATUS_PAUSED_MODE_CHANGE">
+        <description>The mission has been paused by changing the flight mode.</description>
       </entry>
       <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
-        <description>The mission has been automatically paused due to completion of an item with autocontinue=false.</description>
+        <description>The mission has been paused in mission mode. For example, automatically paused due to completion of an item with autocontinue=false.</description>
       </entry>
       <entry value="5" name="MISSION_STATUS_FINISHED">
         <description>The Mission has been finished.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4851,28 +4851,29 @@
         <description>NAK: NAK response</description>
       </entry>
     </enum>
-    <enum name="MISSION_STATUS">
-      <description>Status of a mission.</description>
-      <entry value="0" name="MISSION_STATUS_UNKNOWN">
-        <description>The mission status is not supported.</description>
+    <enum name="MISSION_FSM_STATE">
+      <description>
+        States of the mission finite state machine.
+        Note that these states are independent of whether the mission is in a mode that can execute mission items or not (is suspended).
+        They may not all be relevant on all vehicles.
+      </description>
+      <entry value="0" name="MISSION_FSM_STATE_UNKNOWN">
+        <description>The mission status reporting is not supported.</description>
       </entry>
-      <entry value="1" name="MISSION_STATUS_NOT_STARTED">
-        <description>The mission has not been started yet. This is the case after upload when there is no active mission.</description>
+      <entry value="1" name="MISSION_FSM_STATE_NO_MISSION">
+        <description>No mission on the vehicle.</description>
       </entry>
-      <entry value="2" name="MISSION_STATUS_ACTIVE">
-        <description>The mission is running and active.</description>
+      <entry value="2" name="MISSION_FSM_STATE_NOT_STARTED">
+        <description>Mission has not started. This is the case after a mission has uploaded but not yet started executing.</description>
       </entry>
-      <entry value="3" name="MISSION_STATUS_PAUSED_MODE_CHANGE">
-        <description>The mission has been paused by changing the flight mode.</description>
+      <entry value="3" name="MISSION_FSM_STATE_ACTIVE">
+        <description>Mission is active, and will execute mission items when in auto mode.</description>
       </entry>
-      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
-        <description>The mission has been paused in mission mode. For example, automatically paused due to completion of an item with autocontinue=false.</description>
+      <entry value="4" name="MISSION_FSM_STATE_PAUSED">
+        <description>Mission is paused when in auto mode.</description>
       </entry>
-      <entry value="5" name="MISSION_STATUS_FINISHED">
-        <description>The Mission has been finished.</description>
-      </entry>
-      <entry value="6" name="MISSION_STATUS_NO_MISSION_UPLOADED">
-        <description>There is no mission on the vehicle.</description>
+      <entry value="5" name="MISSION_FSM_STATE_COMPLETE">
+        <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
   </enums>
@@ -5218,8 +5219,9 @@
       </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
-      <field type="uint16_t" name="total">Total number of mission items. 0 if no mission is present on the vehicle.</field>
-      <field type="uint8_t" name="status" enum="MISSION_STATUS">Mission status</field>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
     </message>
     <message id="43" name="MISSION_REQUEST_LIST">
       <description>Request the overall list of mission items from the system/component.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4851,6 +4851,27 @@
         <description>NAK: NAK response</description>
       </entry>
     </enum>
+    <enum name="MISSION_STATUS">
+      <description>Status of a mission.</description>
+      <entry value="0" name="MISSION_STATUS_UNKNOWN">
+        <description>The mission status is not supported.</description>
+      </entry>
+      <entry value="1" name="MISSION_STATUS_NOT_STARTED">
+        <description>The mission has not been started yet. This is the case after upload when there is no active mission.</description>
+      </entry>
+      <entry value="2" name="MISSION_STATUS_ACTIVE">
+        <description>The mission is running and active.</description>
+      </entry>
+      <entry value="3" name="MISSION_STATUS_ACTIVE_NOT_IN_MISSION_MODE">
+        <description>The mission would be running but we're not in mission mode.</description>
+      </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,21 @@
         <description>NAK: NAK response</description>
       </entry>
     </enum>
+    <enum name="MISSION_STATUS">
+      <description>Status of a mission.</description>
+      <entry value="0" name="MISSION_STATUS_NOT_STARTED">
```
hamishwillee, line 4856: I think you need to start from `MISSION_STATUS_UNKNOWN`=0 to indicate systems that don't support mission status.

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,21 @@
...
+        <description>The mission has not been started yet. This is the case after upload.</description>
```
hamishwillee, line 4857: Probably technically "This is the case after upload when there is no active mission". I mean if you upload a mission when there is an active mission you will be in the running state.

message_definitions/v1.0/common.xml
julianoes: Good point!

message_definitions/v1.0/common.xml
julianoes: Right.

message_definitions/v1.0/common.xml
```diff
@@ -5189,6 +5207,9 @@
     <message id="42" name="MISSION_CURRENT">
       <description>Message that announces the sequence number of the current active mission item. The MAV will fly towards this mission item.</description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total">Total number of mission items</field>
+      <field type="uint8_t" name="status" enum="MISSION_STATUS">Total number of mission items</field>
```
hamishwillee, line 5212:
```suggestion
      <field type="uint8_t" name="status" enum="MISSION_STATUS">Mission state.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,24 @@
...
+        <description>The mission status is not supported.</description>
+      </entry>
+      <entry value="1" name="MISSION_STATUS_NOT_STARTED">
+        <description>The mission has not been started yet. This is the case after upload when there is no active mission.</description>
+      </entry>
+      <entry value="2" name="MISSION_STATUS_RUNNING">
+        <description>The mission is running and active.</description>
+      </entry>
+      <entry value="3" name="MISSION_STATUS_PAUSED">
+        <description>The mission has been paused with autocontinue=false.</description>
```
hamishwillee, line 4866: Do we need `with autocontinue=false` clarification at the end here?

To put is another way, if I manually hold would this be the state that is set? If the mission auto pauses is this what I would set.

julianoes: I need to create a second "Paused" for that, I think.

message_definitions/v1.0/common.xml
julianoes: Thanks, I'll fix that.

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,27 @@
...
+      </entry>
+      <entry value="2" name="MISSION_STATUS_ACTIVE">
+        <description>The mission is running and active.</description>
+      </entry>
+      <entry value="3" name="MISSION_STATUS_ACTIVE_NOT_IN_MISSION_MODE">
+        <description>The mission would be running but we're not in mission mode.</description>
+      </entry>
+      <entry value="4" name="MISSION_STATUS_PAUSED">
+        <description>The mission has been paused with autocontinue=false.</description>
+      </entry>
```
hamishwillee, line 4870: This is accurate for PX4 but leaks "implementation details" and also means there's a state (paused in mission by user) that isn't captured for ArduPilot. I think we can do this without the implementation details, and also capture that you might have changed flight mode - since changing a flight mode will pause an active mission. How about:

```suggestion
      <entry value="3" name="MISSION_STATUS_PAUSED_MANUAL">
        <description>The mission has been paused by the a user command or by changing the flight mode.</description>
      </entry>
      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
        <description>The mission has been automatically paused due to completion of an item with autocontinue=false.</description>
      </entry>
```

message_definitions/v1.0/common.xml
```diff
@@ -5189,6 +5210,9 @@
     <message id="42" name="MISSION_CURRENT">
       <description>Message that announces the sequence number of the current active mission item. The MAV will fly towards this mission item.</description>
```
hamishwillee, line 5211: 1. It's worth saying this is streamed. 
   - Should this be streamed when the mission is paused/completed or only when the mission is active and running? 
choose to stream in this mode but not when in a normal loiter.
     - My thinking is that this should only be streamed if there a mission uploaded, but otherwise it can be streamed all the time.
   - Is there a rate at which this is streamed?
2. It is worth saying this is emitted after setting mission_current

Is this OK?

```suggestion
      <description>Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running). This should be streamed and emitted following MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.</description>
```

hamishwillee:
@peterbarker Would appreciate your comment on this.

At the moment you can infer some of the states of the mission from the mode (particularly in ArduPilot where pause is a sub mode of auto). This is not so easy in PX4 where "pause = hold mode" (except after an "autocontinue" pause). I don't think it is easy in either case to know if a mission has actually completed.

So this adds state information to the MISSION_CURRENT, along with the total number of mission items, so a user doesn't need to separately query the mission for that.

message_definitions/v1.0/common.xml
julianoes: > Should this be streamed when the mission is paused/completed or only when the mission is active and running?

It should always be streamed, so that you know what the next waypoint is even before you switch to mission mode.

The rate should be something like 1 Hz but that's always configurable. And as an answer to SET_MISSION_CURRENT as you suggest.

hamishwillee: Thanks @julianoes . I updated the suggestion above. 

But w.r.t. "always streaming" what if there is no mission uploaded? How about in this case a new status and values for the current and total set to 0 ?

hamishwillee: And note that this might have to change a bit depending on what ArduPilot do.

julianoes: I think sending it with 0 would be ok yes.

message_definitions/v1.0/common.xml
```diff
@@ -5189,6 +5210,9 @@
...
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total">Total number of mission items</field>
```
hamishwillee, line 5214:
```suggestion
      <field type="uint16_t" name="total">Total number of mission items. 0 if there is no mission.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,27 @@
...
+      <entry value="3" name="MISSION_STATUS_ACTIVE_NOT_IN_MISSION_MODE">
+        <description>The mission would be running but we're not in mission mode.</description>
+      </entry>
+      <entry value="4" name="MISSION_STATUS_PAUSED">
+        <description>The mission has been paused with autocontinue=false.</description>
+      </entry>
+      <entry value="5" name="MISSION_STATUS_FINISHED">
+        <description>The Mission has been finished.</description>
+      </entry>
+    </enum>
```
hamishwillee, line 4874:
```suggestion
      </entry>
      <entry value="6" name="MISSION_STATUS_NO_MISSION_UPLOADED">
        <description>The Mission has been finished.</description>
      </entry>
    </enum>
```

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,30 @@
...
+      </entry>
+      <entry value="2" name="MISSION_STATUS_ACTIVE">
+        <description>The mission is running and active.</description>
+      </entry>
+      <entry value="3" name="MISSION_STATUS_PAUSED_MANUAL">
+        <description>The mission has been paused by the a user command or by changing the flight mode.</description>
+      </entry>
+      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
+        <description>The mission has been automatically paused due to completion of an item with autocontinue=false.</description>
+      </entry>
```
hamishwillee, line 4870: @julianoes Refining these a bit. The cases are "paused out of auto" and "paused in auto".

```suggestion
      <entry value="3" name="MISSION_STATUS_PAUSED_MODE_CHANGE">
        <description>The mission has been paused by changing the flight mode.</description>
      </entry>
      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
        <description>The mission has been paused in mission mode. For example, automatically paused due to completion of an item with autocontinue=false.</description>
      </entry>
```

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,30 @@
...
+        <description>The mission has been paused by changing the flight mode.</description>
+      </entry>
+      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
+        <description>The mission has been paused in mission mode. For example, automatically paused due to completion of an item with autocontinue=false.</description>
+      </entry>
+      <entry value="5" name="MISSION_STATUS_FINISHED">
+        <description>The Mission has been finished.</description>
+      </entry>
+      <entry value="6" name="MISSION_STATUS_NO_MISSION_UPLOADED">
+        <description>The Mission has been finished.</description>
```
hamishwillee, line 4875:
```suggestion
        <description>There is no mission on the vehicle.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,30 @@
         <description>NAK: NAK response</description>
       </entry>
     </enum>
+    <enum name="MISSION_STATUS">
```
hamishwillee, line 4854: We might need more work on these states in light of what ArduPlane does. For example when it is complete, it goes straight into RTL. So at this point the mission is complete but the mode is something else.

hamishwillee:
Mission states are a little complex and we may need to rethink this to make sure that we're sending the right information. What that information depends on who/how this is consumed. In summary though

Summary, IMO
- States reflect mission engine when it is running: unknown, no-mission, not-started, active, paused, finished.
- If you're in another mode the mission engine is "suspended" - this is a useful distinction or you need to have multiple states to also cover the suspended case. You might infer this from the current mode, but I think we should expose this as a field to avoid the system having to know anything about autopilot modes. It is effectively "not mission mode or submode".

This is a change from having an option  "Paused not in auto". There is a difference between saying "Pausing a mission mission moves it into hold mode" and "PX4 implements pause as hold". I THINK we are the former - certainly the mode is advertised as Hold in QGC.

Make sense? There is some useful info about what the systems do below.

----

TLDR;

The "reality" is that we have a mission state machine that might have states "like[+]":
- mission not uploaded
- mission not yet started - mission uploaded and all jump counters[++] reset. But has not yet been started.
- mission running - mission is executing mission items
- mission paused - mission is pausing within the context of the mission mode - ie in a submode.
- mission complete - stopped

[+]: I say "like" because some implementations use the same "state" _mission stopped_ for mission _not yet started_ and _complete_. Further, plane does not support paused, even in theory.
[++] Jump counters are separate from state.

Then on top of this we have the case where a mission may be _suspended_ - i.e. you're in another mode like "Hold". In this case. The mission is not running at all but the state still exists and might be re-entered.

ArduCopter matches those states _precisely_ (as sub modes).
- so if you are paused you're still in mission mode. If moving you'll move to a graceful stop.
- If you leave the mission to another mode it is suspended. Say you move to hold, you'd immediately hold rather than gracefully stop.
- FYI you can't start mission from the initial state by calling the set mission current
- ArduPilot does not implement autocontinue flag.

ArduPlane does not implement the  mission paused state; you cannot pause mission, you can only suspend it by changing mode.
- The pause/continue command is not implemented because it doesn't make sense with this paradigm. Though continue might, if we assumed that this also requires a switch-to-mode.
- If you run out of items in the mission (complete) the internal state changes to mission-stop and the mode is forced to RTL. Sending message to set the current mission item sets the state back to active - you then have to change the mode back to resume.
- The plane never wants to say it is paused.
- The plane state is almost completely inferrable - you're either in mission mode and executing or stopped or you are in some other mode. 
- FYI you CAN start mission from the initial state by calling the set mission current

PX4 does a mix of these two, implementing both pause sub modes and emulating pause-as-switch to mode:
- It has mission sub mode for the pause state when `autocontinue=false`
- It has mission sub mode for the complete state (right?).
- Pausing a mission is implemented as moving to hold mode. But this is not done using a pause, but using a do_reposition.
- Restarting a mission is implemented as moving back to mission mode.
- Not sure/ do we know how PX4 sees an explicit mode change vs pause when running a mission? i.e. does it think of itself as in hold mode, or does it somehow know that actually it is in a paused mission mode?

So:
- Arduplane might want to output states: unknown, no-mission, not-started, active, complete (internally, both "not started" and "complete" are "stopped").
- ArduCopter and PX4 might want to output states: unknown, no-mission, not-started, active, paused, finished.

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5211,15 @@
...
     <message id="42" name="MISSION_CURRENT">
-      <description>Message that announces the sequence number of the current active mission item. The MAV will fly towards this mission item.</description>
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total">Total number of mission items. 0 if no mission is present on the vehicle.</field>
```
tridge, line 5221: zero can't be used as would break with implementations that don't have this extension, we need zero to mean unknown

tridge (changes requested):

message_definitions/v1.0/common.xml
julianoes: That's true. So we need to set it to UINT16_MAX for "no mission"/"empty mission", 0 for unknown, and > 0 else for a valid mission.

message_definitions/v1.0/common.xml
```diff
@@ -4851,6 +4851,30 @@
...
+      <entry value="4" name="MISSION_STATUS_PAUSED_AUTO">
+        <description>The mission has been paused in mission mode. For example, automatically paused due to completion of an item with autocontinue=false.</description>
+      </entry>
+      <entry value="5" name="MISSION_STATUS_FINISHED">
+        <description>The Mission has been finished.</description>
+      </entry>
+      <entry value="6" name="MISSION_STATUS_NO_MISSION_UPLOADED">
+        <description>There is no mission on the vehicle.</description>
+      </entry>
+    </enum>
```
hamishwillee, line 4877: Change MISSION_STATUS to reflect the "ideal" states of the mission finite state machine, as identified in https://github.com/mavlink/mavlink/pull/1869#issuecomment-1212621078. 

Because this is actually the state machine and not the mission, I have renamed to MISSION_FSM_STATE. That way people are less likely to think "but I paused my mission by moving to Hold mode, so I should emit paused". 

I will add a value for "suspended".


```suggestion
    <enum name="MISSION_FSM_STATE">
      <description>
        States of the mission finite state machine.
        Note that these states are independent of whether the mission is in a mode that can execute mission items or not (is suspended).
      </description>
      <entry value="0" name="MISSION_FSM_STATE_UNKNOWN">
        <description>The mission status reporting is not supported.</description>
      </entry>
      <entry value="1" name="MISSION_FSM_STATE_NO_MISSION">
        <description>No mission on the vehicle.</description>
      </entry>
      <entry value="2" name="MISSION_FSM_STATE_NOT_STARTED">
        <description>Mission has not started. This is the case after a mission has uploaded but not yet started executing.</description>
      </entry>
      <entry value="3" name="MISSION_FSM_STATE_ACTIVE">
        <description>Mission is executing mission items (when in auto mode).</description>
      </entry>
      <entry value="4" name="MISSION_FSM_STATE_PAUSED">
        <description>Mission is paused.</description>
      </entry>
      <entry value="5" name="MISSION_FSM_STATE_COMPLETE">
        <description>Mission has executed all mission items.</description>
      </entry>
    </enum>
```

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5211,15 @@
...
-      <description>Message that announces the sequence number of the current active mission item. The MAV will fly towards this mission item.</description>
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total">Total number of mission items. 0 if no mission is present on the vehicle.</field>
+      <field type="uint8_t" name="status" enum="MISSION_STATUS">Mission status</field>
```
hamishwillee, line 5222: Update MISSION_STATUS to MISSION_FSM_STATE in message

```suggestion
      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE">Mission finite state machine state.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5211,15 @@
...
     </message>
```
hamishwillee, line 5223: Add value to indicate whether mission is in a mode where it can be executed. The actual state can always be inferred from this and the one above.

```suggestion
      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
    </message>
```

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5212,16 @@
...
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
```
hamishwillee, line 5224:
```suggestion
      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5212,16 @@
...
-      <description>Message that announces the sequence number of the current active mission item. The MAV will fly towards this mission item.</description>
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
```
hamishwillee, line 5223:
```suggestion
      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE" invalid="0">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5212,16 @@
...
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE" invalid="0">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
```
hamishwillee, line 5224: @julianoes Do you think this needs to be an enum?

hamishwillee:
@julianoes @tridge I updated so all fields are 0 for the default "not supported" case.

In addition, have modified the reporting of state in line with the [notes above](https://github.com/mavlink/mavlink/pull/1869#issuecomment-1212621078). It now reports state using two fields. The first is the state of the mission state machine. The second is whether or not the mission is suspended (worded as whether it is in mission mode or not). That allows a recipient to absolutely know the what the mission state is.

Is it worth considering also reporting on whether jump counters are reset? If so, would making the last byte into an set of flags that includes suspended and jump counters be a better implementation?

tridge:
@hamishwillee thanks for the changes, looks much better

message_definitions/v1.0/common.xml
```diff
@@ -5220,8 +5220,8 @@
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
-      <field type="uint8_t" name="status" enum="MISSION_FSM_STATE" invalid="0">Mission finite state machine state. MISSION_FSM_STATE_UNKNOWN if state reporting not supported.</field>
-      <field type="uint8_t" name="mission mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items (not suspended). 0: Unknown, 1: Mission mode, 2: Not in mission mode (suspended).</field>
+      <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="mission_mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
```
hamishwillee, line 5224: An alternative. Do we like this more? I think not as much if we don't have an enum, and more if we do.

```suggestion
      <field type="uint8_t" name="suspended" minValue="0" maxValue="2" increment="1" invalid="0">Mission is suspended (is in a mode that cannot execute mission items). 0: Unknown, 1: Suspended, 2: In a mission mode (not suspended).</field>
```

hamishwillee: I guess not!

message_definitions/v1.0/common.xml
```diff
@@ -5187,8 +5212,16 @@
...
+      <description>
+        Message that announces the sequence number of the current target mission item (that the system will fly towards/execute when the mission is running).
+        This message should be streamed all the time (nominally at 1Hz).
+        This message should be emitted following a call to MAV_CMD_DO_SET_MISSION_CURRENT or SET_MISSION_CURRENT.
+      </description>
       <field type="uint16_t" name="seq">Sequence</field>
+      <extensions/>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
+      <field type="uint8_t" name="mission_mode" minValue="0" maxValue="2" increment="1" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
```
hamishwillee, line 5224: minValue="0" maxValue="2" increment="1" are only for params.

```suggestion
      <field type="uint8_t" name="mission_mode" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
```

hamishwillee:
As discussed in dev call, this has been reviewed by ArduPilot and PX4 and can be merged.

Note that this:

- adds extension fields to the previous mission in a way such that there is no change for systems that have not updated to emit the new information.
- The previous definition said "emit on changed waypoint" while the new one says "stream". This reflects what both PX4 and ardupilot actually do. It is not inconceivable some other system might have implemented this exactly per the old spec.

Thanks for this @julianoes , and @tridge for helping me understand how ArduPlane works with missions.

hamishwillee:
FYI @TSC21

PR #1870, opened 2022-08-04
MAV_CMD_DO_SET_MISSION_CURRENT - allow mission resetting

hamishwillee:
This changes one of the unused params in `MAV_CMD_DO_SET_MISSION_CURRENT` so that it can be used to reset a mission "fully".

Prior to this the command would just update the current mission item.
- if the mission was completed it would stay complete, with no way to restart other than re-uploading the mission or landing and taking off again (which triggers mission reset).
- if the mission was active you could go back to the start point but the mission would not reset jump counters. 

With this change you can optionally specify that setting the current mission item also clears any "completion status" and makes the mission active again, and that the mission counters are reset.

This effectively means that  `MAV_CMD_DO_SET_MISSION_CURRENT` can be used as a hypothetical "reset mission" command by setting the mission item to 1 and the new flag.

@auturgy @julianoes Make sense?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1756,9 +1756,17 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="224" name="MAV_CMD_DO_SET_MISSION_CURRENT" hasLocation="false" isDestination="false">
-        <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+        <description>
+          Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+          If the mission is running, the system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+          If the mission is paused it will remain paused until restarted.
+          If the mission is complete it remain complete unless param2 is set to true, in which case it will become paused.
+          If param2 is true the mission will be reset.
+	  The command will ACK with MAV_RESULT_FAILED if the sequence number is out of range (including if there is no mission item).
+	      
+        </description>
         <param index="1" label="Number" minValue="0" increment="1">Mission sequence value to set</param>
-        <param index="2" label="Reset Mission" minValue="0" maxValue="0" increment="1">Resets mission. 1: true, 0: false. Resetting unsets mission completion (a completed mission is now active/paused) and resets jump counters to initial values. Resetting does not affect pause/active state.</param>
+        <param index="2" label="Reset Mission" minValue="0" maxValue="0" increment="1">Resets mission. 1: true, 0: false. Resetting unsets mission completion (a completed mission is now active/paused) and resets jump counters to initial values. Paused missions will remain paused. Completed missions will become paused missions.</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
         <param index="5">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1758,16 +1758,21 @@
       <entry value="224" name="MAV_CMD_DO_SET_MISSION_CURRENT" hasLocation="false" isDestination="false">
         <description>
           Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
-          If the mission is running, the system will continue to this mission item on the shortest path, skipping any intermediate mission items.
-          If the mission is paused it may remain paused until resumed.
-          If the mission is complete it remains complete unless param2 is set.
+          If a mission is currently being executed, the system will continue to this new mission item on the shortest path, skipping any intermediate mission items.
+	  Note that mission jump repeat counters are not reset unless param2 is set (see MAV_CMD_DO_JUMP param2).
+
+          This command may trigger a mission state-machine change on some systems: for example from MISSION_STATE_NOT_STARTED or MISSION_STATE_PAUSED to MISSION_STATE_ACTIVE.
+          If the system is in mission mode, on those systems this command might therefore start, restart or resume the mission.
+          If the system is not in mission mode this command must not trigger a switch to mission mode.
+
           The mission may be "reset" using param2.
-          Resetting sets jump counters to initial values and unsets the "completed" mission state (if set), moving it to the "paused" state; it does not resume a paused mission.
-          To reset counters without changing the current mission item you should set the current item to `-1` .
+          Resetting sets jump counters to initial values (to reset counters without changing the current mission item set the param1 to `-1`).
+          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, allowing it to resume.
+
 	  The command will ACK with MAV_RESULT_FAILED if the sequence number is out of range (including if there is no mission item).
         </description>
         <param index="1" label="Number" minValue="-1" increment="1">Mission sequence value to set. -1 for the current mission item (use to reset mission without changing current mission item).</param>
-        <param index="2" label="Reset Mission" minValue="0" maxValue="1" increment="1">Resets mission. 1: true, 0: false. Resets jump counters to initial values and moves competed missions to the paused state.</param>
+        <param index="2" label="Reset Mission" minValue="0" maxValue="1" increment="1">Resets mission. 1: true, 0: false. Resets jump counters to initial values and changes mission state "completed" to be "active" or "paused".</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
         <param index="5">Empty</param>
@@ -5190,7 +5195,16 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
     <message id="41" name="MISSION_SET_CURRENT">
-      <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+      <deprecated since="2022-08" replaced_by="MAV_CMD_DO_SET_MISSION_CURRENT"/>
+      <description>     
+        Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+        If a mission is currently being executed, the system will continue to this new mission item on the shortest path, skipping any intermediate mission items.
+        Note that mission jump repeat counters are not reset (see MAV_CMD_DO_JUMP param2).
+
+        This message may trigger a mission state-machine change on some systems: for example from MISSION_STATE_NOT_STARTED or MISSION_STATE_PAUSED to MISSION_STATE_ACTIVE.
+        If the system is in mission mode, on those systems this command might therefore start, restart or resume the mission.
+        If the system is not in mission mode this message must not trigger a switch to mission mode.
+      </description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint16_t" name="seq">Sequence</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1767,7 +1767,7 @@
 
           The mission may be "reset" using param2.
           Resetting sets jump counters to initial values (to reset counters without changing the current mission item set the param1 to `-1`).
-          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, allowing it to resume when it is (next) in a mission mode.
+          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, potentially allowing it to resume when it is (next) in a mission mode.
 
 	  The command will ACK with MAV_RESULT_FAILED if the sequence number is out of range (including if there is no mission item).
         </description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1758,7 +1758,7 @@
       <entry value="224" name="MAV_CMD_DO_SET_MISSION_CURRENT" hasLocation="false" isDestination="false">
         <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
         <param index="1" label="Number" minValue="0" increment="1">Mission sequence value to set</param>
-        <param index="2">Empty</param>
+        <param index="2" label="Reset Mission" minValue="0" maxValue="0" increment="1">Resets mission. 1: true, 0: false. Resetting unsets mission completion (a completed mission is now active/paused) and resets jump counters to initial values. Resetting not affect pause/active state.</param>
```
julianoes, line 1761:
```suggestion
        <param index="2" label="Reset Mission" minValue="0" maxValue="0" increment="1">Resets mission. 1: true, 0: false. Resetting unsets mission completion (a completed mission is now active/paused) and resets jump counters to initial values. Resetting does not affect pause/active state.</param>
```

hamishwillee:
@julianoes And of course PX4 does not support this :-). 

Worth adding to [MISSION_SET_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_SET_CURRENT) as well or do I fix up the command and deprecate this version (on assumption commands are better for this kind of thing?)

hamishwillee:
@peterbarker Would appreciate your review/advice on this.

The current implementation of `MAV_CMD_DO_SET_MISSION_CURRENT` and `SET_MISSION_CURRENT` just resets the sequence number. If you have a completed mission the mission remains completed. If you have an active mission with jump counters it will change the mission number, but it doesn't reset the counters. 

What this change does is:
- Allow you to really reset a mission - you specify param2 to reset the counters and unset the completion status. I have stated that doing this when complete forces the mission to paused. This allows a full reset of the mission, as if you had uploaded a new mission or done an arm/disarm cycle.
- Clearly specifies that if the mission is paused or complete calling this does not continue the mission - it just updates the mission number. 
- Clearly specifies what happens if you set a sequence value out of range `MAV_RESULT_FAILED`

I have looked at the ArduPilot implementation of this you did in https://github.com/ArduPilot/ardupilot/pull/16687.
The definition here slightly disagrees with that in two respects.
1. If there is no mission APM ACKs with MAV_RESULT_UNSUPPORTED. I think it should return  `MAV_RESULT_FAILED` because this is a) a case of "out of range sequence" and b) because "unsupported" is defined as "the command is not supported", while "failed" means "this command is valid but you can't call it at this time and there is some environmental problem you will have to fix before you can call it - i.e you will have to change the mission.
2. It looks like the general implementation just sets the mission number but does not continue, while the plane will autocontinue if it gets this: 
   ```
       // if you change this you must change handle_mission_set_current
    plane.auto_state.next_wp_crosstrack = false;
    if (plane.control_mode == &plane.mode_auto && plane.mission.state() == AP_Mission::MISSION_STOPPED) {
        plane.mission.resume();
    }
    ```
    Is that correct? If so, then I might need to amend the text that this does not force autocontinue.

Point 2 is problematic, and the reason I am "hoping to do" this update. To give you some context, PX4 implements a normal pause just by changing the mode to hold mode (which is a separate mode, not a sub-mode of auto like in ArduPilot). But it implements a pause on autocontinue=false as a kind of sub mode. When `SET_MISSION_CURRENT` is called when you're in hold mode the call just updates the current mission number, but when you're in the "autocontinue=false" mode it will automatically restart the mission as well as updating the mission number.

I think this is very inconsistent. A mission should either always continue when you set the current mission item or always remain paused. But it shouldn't do something different because of the type of pause, or because you happen to be flying in a plane. That's just confusing for users.

So what I want to do is make `MAV_CMD_DO_SET_MISSION_CURRENT` reflect a consistent interface, and then move over time to deprecate `SET_MISSION_CURRENT`.

Note that the discussion about autocontinue is going on in https://github.com/mavlink/mavlink/pull/1868. I have paused that in the hope I can resolve it here :-)

hamishwillee:
PS Arguably we might further clarify this by making whether the mission should continue (unpause) explicit too. So by default calling the mission will remain paused, but if you set the `continue` param it will unpause. Old versions of plane would do what they like, ignoring this. New versions would update.

Either way, everything would be explicit.

message_definitions/v1.0/common.xml
```diff
@@ -1756,9 +1756,16 @@
...
       <entry value="224" name="MAV_CMD_DO_SET_MISSION_CURRENT" hasLocation="false" isDestination="false">
-        <description>Set the mission item with sequence number seq as current item. This means that the MAV will continue to this mission item on the shortest path (not following the mission items in-between).</description>
+        <description>
+          Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
+          If the mission is running, the system will continue to this mission item on the shortest path, skipping any intermediate mission items.
+          If the mission is paused it will remain paused until restarted.
+          If the mission is complete it remain complete unless param2 is set to true, in which case it will become paused.
+          If param2 is true the mission will be reset.
+	  The command will ACK with MAV_RESULT_FAILED if the sequence number is out of range (including if there is no mission item).
+        </description>
```
hamishwillee, line 1766: Clarify how to just reset the counters and not change the mission item. Setting the current mission in this case would be a race condition - causing problems if the mission has moved to the next step.

```suggestion
        <description>
          Set the mission item with sequence number seq as the current item and emit MISSION_CURRENT (whether or not the mission number changed).
          If the mission is running, the system will continue to this mission item on the shortest path, skipping any intermediate mission items.
          If the mission is paused it may remain paused until resumed.
          If the mission is complete it remains complete unless param2 is set.
          The mission may be "reset" using param2.
          Resetting sets jump counters to initial values and unsets the "completed" mission state (if set), moving it to the "paused" state; it does not resume a paused mission.
          To reset counters without changing the current mission item you should set the current item to `-1` .
	  The command will ACK with MAV_RESULT_FAILED if the sequence number is out of range (including if there is no mission item).
        </description>
```

message_definitions/v1.0/common.xml
```diff
@@ -1756,9 +1756,16 @@
...
         <param index="1" label="Number" minValue="0" increment="1">Mission sequence value to set</param>
```
hamishwillee, line 1767:
```suggestion
        <param index="1" label="Number" minValue="-1" increment="1">Mission sequence value to set. -1 for the current mission item (use to reset mission without changing current mission item).</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1756,9 +1756,16 @@
...
-        <param index="2">Empty</param>
+        <param index="2" label="Reset Mission" minValue="0" maxValue="0" increment="1">Resets mission. 1: true, 0: false. Resetting unsets mission completion (a completed mission is now active/paused) and resets jump counters to initial values. Paused missions will remain paused. Completed missions will become paused missions.</param>
```
hamishwillee, line 1768: Simplify description and set maxValue to 1.
```suggestion
        <param index="2" label="Reset Mission" minValue="0" maxValue="1" increment="1">Resets mission. 1: true, 0: false. Resets jump counters to initial values and moves competed missions to the paused state.</param>
```

hamishwillee:
@julianoes FYI I now know a bit more about how ArduPilot does "pause" and  works with this  `MAV_CMD_DO_SET_MISSION_CURRENT`

- Neither PX4 or ArduPilot allows `MAV_CMD_DO_SET_MISSION_CURRENT` to change the main mode. So if you're in hold and call it, you won't change mode and resume automatically.
- ArduCopter runs all functionality in sub modes of auto. So if you pause you are pausing in auto. 
  - `MAV_CMD_DO_SET_MISSION_CURRENT` does not change the mode from paused. The only mode like this in PX4 is start from autocontinue=false pause - I would say PX4 is incompatible since autocontinue is a late spec addition and their implementation predates it.
  - does not autostart if you call this after boot when landed. I think this differs from PX4 which will autostart from landed - right?
- ArduPlane does not support pause "semantically".
  - If you want to pause you just change to the mode you want. They don't implement a pause instruction or call it pause.
  -  If it runs out of items it performs an RTL. If you call  `MAV_CMD_DO_SET_MISSION_CURRENT` it will reset the mission sequence and unset the "stopped" status. However it will not resume until you manually change back to auto (i.e. not auto change from calling this).
  - if you call  `MAV_CMD_DO_SET_MISSION_CURRENT` from the ground when it is in auto it will take off. THis is more like PX4 copter.
  
I'm still trying to work out whether we can align. I think the best thing would be if PX4 implements  "not resuming following autocontinue=false" I can tell a good story. The other behaviours I can live with the inconsitency

hamishwillee:
@tridge @rmackay9 @peterbarker @julianoes Can I please have a new review of this MAV_CMD_DO_SET_MISSION_CURRENT/MISSION_SET_CURRENT update.

This aligns them with how ArduCopter, PX4 and ArduPlane actually work using the terminology we used for the MISSION_CURRENT update. The main purpose is the jump counter addition, which has not changed since your last review.

Specifically:

- It explicitly states that by default jump counters are not reset (that was previously inferred)
- It states that this command may change the mission-state on some systems, and that in mission mode this might resume a mission. 
- it states that this command must not change the flight mode - so if you're loitering this command will change the state machine, but not resume the mission.

What it doesn't do is explain which systems allow the commands to change the state, and why - for example, ArduPlane uses a model where MAV_CMD_DO_SET_MISSION_CURRENT is used to set mission-state active when the mission is suspended, and then move the mission to auto to take off. While ArduCopter would might sit in AUTO and then use MISSION_START to start a mission. That would have to be in flight stack docs, though I hope to capture some of it in the mavlink devguide once all this is settled.

@peterbarker I think the ideas are now baked enough to implement.

message_definitions/v1.0/common.xml
```diff
@@ -1756,9 +1756,23 @@
...
+          If a mission is currently being executed, the system will continue to this new mission item on the shortest path, skipping any intermediate mission items.
+	  Note that mission jump repeat counters are not reset unless param2 is set (see MAV_CMD_DO_JUMP param2).
+
+          This command may trigger a mission state-machine change on some systems: for example from MISSION_STATE_NOT_STARTED or MISSION_STATE_PAUSED to MISSION_STATE_ACTIVE.
+          If the system is in mission mode, on those systems this command might therefore start, restart or resume the mission.
+          If the system is not in mission mode this command must not trigger a switch to mission mode.
+
+          The mission may be "reset" using param2.
+          Resetting sets jump counters to initial values (to reset counters without changing the current mission item set the param1 to `-1`).
+          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, allowing it to resume.
```
hamishwillee, line 1770:
```suggestion
          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, allowing it to resume when it is (next) in a mission mode.
```

message_definitions/v1.0/common.xml
```diff
@@ -1756,9 +1756,23 @@
...
+          If a mission is currently being executed, the system will continue to this new mission item on the shortest path, skipping any intermediate mission items.
+	  Note that mission jump repeat counters are not reset unless param2 is set (see MAV_CMD_DO_JUMP param2).
+
+          This command may trigger a mission state-machine change on some systems: for example from MISSION_STATE_NOT_STARTED or MISSION_STATE_PAUSED to MISSION_STATE_ACTIVE.
+          If the system is in mission mode, on those systems this command might therefore start, restart or resume the mission.
+          If the system is not in mission mode this command must not trigger a switch to mission mode.
+
+          The mission may be "reset" using param2.
+          Resetting sets jump counters to initial values (to reset counters without changing the current mission item set the param1 to `-1`).
+          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, allowing it to resume when it is (next) in a mission mode.
```
hamishwillee, line 1770:
```suggestion
          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, potentially allowing it to resume when it is (next) in a mission mode.
```

julianoes (approved):
Thanks, I think this is pretty clear now!

hamishwillee:
Thanks @julianoes . Just waiting on ArduPilot review.

message_definitions/v1.0/common.xml
```diff
@@ -1756,9 +1756,23 @@
...
+          If the system is not in mission mode this command must not trigger a switch to mission mode.
+
+          The mission may be "reset" using param2.
+          Resetting sets jump counters to initial values (to reset counters without changing the current mission item set the param1 to `-1`).
+          Resetting also explicitly changes a mission state of MISSION_STATE_COMPLETE to MISSION_STATE_PAUSED or MISSION_STATE_ACTIVE, potentially allowing it to resume when it is (next) in a mission mode.
+
+	  The command will ACK with MAV_RESULT_FAILED if the sequence number is out of range (including if there is no mission item).
+        </description>
+        <param index="1" label="Number" minValue="-1" increment="1">Mission sequence value to set. -1 for the current mission item (use to reset mission without changing current mission item).</param>
+        <param index="2" label="Reset Mission" minValue="0" maxValue="1" increment="1">Resets mission. 1: true, 0: false. Resets jump counters to initial values and changes mission state "completed" to be "active" or "paused".</param>
```
tridge, line 1775: I like the counter reset idea.

tridge (approved):

hamishwillee:
Thanks for reviews! Merging.

PR #1872, opened 2022-08-06
Huiyulhy add vehicle local position message

huiyulhy:
Added a mavlink message under the dialect "development" which contains full information from the uORB topic vehicle_local_position in PX4. This is useful for users who are interested in the vehicle position in a global reference frame with the xyz coordinate system.


hamishwillee:
Thanks, but we won't be able to do this.

The thing is that development.xml is a place for work-in-progress messages that are on the path to being part of the MAVLink standard - eventually moving to common.xml or standard.xml. The only allowed messages are those that are intended to be shared by multiple flight stacks/systems. However `vehicle_local_position` expose the internals of PX4 - so it is very specific to PX4.

If you need this information then you would normally look at logs.

PR #1865, opened 2022-07-13
RemoteID v2 and adding additional remote ID definitions

BluemarkInnovations:
At the moment I'm working both on [integrating Remote ID MAVLink support in ArduPilot](https://github.com/ArduPilot/ardupilot/pull/21075) and releasing a Remote ID MAVLink capable [transponder](https://dronescout.co/dronebeacon-mavlink-remote-id-transponder/).

An outcome of this process is that -in my opinion - the current set of MAVLink message definitions are incomplete. Next, I have created an [issue](https://github.com/opendroneid/opendroneid-core-c/issues/55) at the OpenDrone ID project to discuss new message definitions with the maintainers. That is now complete and in this PR I'm submitting two additional message definitions:

- CMD message to control the radio transmission of a Remote ID transponder. For instance, it can be useful to switch off radio transmission when the UAV is on the ground.
- ACK message to signal incompatibility issues. For instance if the Remote ID transponder does not support a particular MAVLink Remote ID message, it can signal this using the ACK message.

PR #1873, opened 2022-08-07
common: added OPEN_DRONE_ID_ARM_STATUS

tridge:
this is required for the pre-arm check that the transmitter is ready
and passing tests

message ID chosen so as not to conflict with PR #1865

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4354,6 +4354,14 @@
         <description>CAA (Civil Aviation Authority) registered operator ID.</description>
       </entry>
     </enum>
+    <enum name="MAV_ODID_ARM_STATUS">
+      <entry value="0" name="MAV_ODID_GOOD_TO_ARM">
+        <description>Passing arming checks.</description>
+      </entry>
+      <entry value="1" name="MAV_ODID_PRE_ARM_FAIL_GENERIC">
+        <description>Generic arming failure, see error string for details.</description>
+      </entry>
+    </enum>
     <enum name="TUNE_FORMAT">
       <description>Tune formats (used for vehicle buzzer/tone generation).</description>
       <entry value="1" name="TUNE_FORMAT_QBASIC1_1">
@@ -7462,6 +7470,11 @@
       <field type="uint8_t" name="single_message_size" units="bytes">This field must currently always be equal to 25 (bytes), since all encoded OpenDroneID messages are specified to have this length.</field>
       <field type="uint8_t" name="msg_pack_size">Number of encoded messages in the pack (not the number of bytes). Allowed range is 1 - 9.</field>
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
+    </message>
+    <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
+      <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
+      <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
+      <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
     </message>
     <message id="12920" name="HYGROMETER_SENSOR">
       <description>Temperature and humidity from hygrometer.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7472,6 +7472,8 @@
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
     <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
       <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
       <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7472,8 +7472,8 @@
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
     <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
       <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
       <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>


hamishwillee:
Hi @tridge ,

Am I correct in thinking that this is information sent by the transmitter to the autopilot telling it that the transponder itself is ready to start transmitting or receiving. IN other words its something that the autopilot needs to know that the transponder is good to go? [my confusion is that it reads a bit like the transponder might be getting the arm state from some remote system].

So you might name this TRANSPONDER_READY_TO_ENABLE or similar for the same semantics?

If so:
1.  What is the trigger to sent this? Is this streamed by the transmitter, or requested by the autopilot using [MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE)? 
2. More generally, this is remote state. Is there any other state we might have in the transponder that needs to be sent as well. If so, can that be captured in a general message?

hamishwillee:
FYI @friissoren

message_definitions/v1.0/common.xml
```diff
@@ -7451,6 +7459,11 @@
       <field type="uint8_t" name="operator_id_type" enum="MAV_ODID_OPERATOR_ID_TYPE">Indicates the type of the operator_id field.</field>
       <field type="char[20]" name="operator_id">Text description or numeric value expressed as ASCII characters. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
+    <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
```
friissoren, line 7462: The order is now wrong. This will make 12918 come before 12915.

BluemarkInnovations:
My objections are mentioned in this comment: https://github.com/mavlink/mavlink/pull/1865#issuecomment-1206465074 Or more precise it, explains why remote ID v2 already provides the **full state of the transponder** to the autopilot. Hence, sending/asking a OPEN_DRONE_ID_ARM_STATUS is not needed. So I don't see any benefits in having this extra message. 

@tridge Can you give me examples/scenarios that are not covered in remote ID v2, where this message type could be useful?

tridge:
@BluemarkInnovations The autopilot does not know the full state of the transmitter, as it doesn't know the state of the radio/wifi etc. The transmitter is the only one that knows for sure if it is able to transmit and is happy with the state of the messages it is being sent. All the autopilot knows is what it is sending to the transmitter. It doesn't know if those messages are being received. It doesn't know if the transmitter has some hw issue that prevents transmit.
@hamishwillee yes, this is for transmitting from the transmitter to the flight controller, to tell the flight controller that everything looks good and we're ready to go. The transmitter looks at all the data it has and decides if that meets the requirements for arming to be OK.
There is an equivalent dronecan.remoteid.ArmStatus message in DroneCAN

tridge:
> 1. What is the trigger to sent this? Is this streamed by the transmitter,

it is streamed by the transmitter as both MAVLink and (if supported) DroneCAN. The ArduPilot implementation will not allow arming until it gets this ARM_STATUS message from the transmitter

> 2\. More generally, this is remote state. Is there any other state we might have in the transponder that needs to be sent as well

not that I came across. What we need is a definite "the transmitter has everything it needs to send compliant BT/WiFi messages". It is deliberately a very simple message

BluemarkInnovations:
> @BluemarkInnovations The autopilot does not know the full state of the transmitter, as it doesn't know the state of the radio/wifi etc. The transmitter is the only one that knows for sure if it is able to transmit and is happy with the state of the messages it is being sent. All the autopilot knows is what it is sending to the transmitter. It doesn't know if those messages are being received. It doesn't know if the transmitter has some hw issue that prevents transmit. 

I think there are two things to address: one is Remote ID v1 vs Remote ID v2 protocol. Message flow v2: https://github.com/mavlink/mavlink/pull/1865#issuecomment-1207977539 With Remote ID v2 the autopilot does know the state of radio/wifi and it makes sure the transponder receives the messages from the autopilot for sure. On the other hand I do see compliance documents requiring such a message. I basically try to say Remote ID v2 is like TCP/IP. In TCP/IP you also don't want to define an ACK packet as this is already handled in the TCP/IP stack. So the `OPEN_DRONE_ID_ARM_STATUS` feels a bit like that. So, I don't want start an academic debate; from regulation point of view it makes sense for such a packet.

tridge:
a couple of notes that may be relevant.
Based on the discussions I've had with US vendors I expect that the majority of deployments for flight controller <-> transmitter comms will be DroneCAN, not mavlink. GCS <-> flight controller will be MAVLink. The DroneCAN part is worth thinking about to ensure any MAVLink changes fit in a DroneCAN world (I made the DroneCAN messages mirror the fields in the MAVLink messages exactly to ease implementation of dual-protocol transmitters).
2nd main thing if we're working on an incredibly compressed timeline here. The means of compliance from ASTM for FAA RemoteID only just came out, but we're needing to get this into production vehicles by Sep 16th 2022, including testing and certification. So we're already in the mode of looking for minimum necessary changes. We released the first beta of ArduPilot with OpenDroneID support this morning, and we expect a stable release in the next 2 weeks. That means WIP is no longer really appropriate.
The ARM_STATUS message was added to the ArduPilot XML because of the PFST requirements in the ASTM MOC. I think it really is required.

friissoren (changes requested):
@tridge : Please change the order so that this new 12918 message comes after the 12915 message.

message_definitions/v1.0/common.xml
```diff
@@ -7451,6 +7459,11 @@
...
+      <description>Status from the transmitter telling the flight controller if the remote ID system is ready for arming.</description>
```
hamishwillee, line 7463: @tridge I think this is more clear to me because I think it is PX4 that is ready to arm based on this message, not the remote ID system ready to arm (as implied):

```suggestion
      <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
```

hamishwillee:
Thanks @tridge @BluemarkInnovations @friissoren . Matching up with DroneCAN messages is usually a "good thing". I haven't got back to the other PR today sorry.

1. This is a broadcast message. Would it make sense to make it targeted?
2. @tridge please move the message so the ids are not mixed up. Also mark it as wip. Those are "the rules" until there is a complete implementation and agreement from multiple stacks.
3. @friissoren Are you broadly OK with this? If so I will merge once the questions are resolve. 
4. @auturgy @julianoes Any last minute comments?

tridge:
> This is a broadcast message. Would it make sense to make it targeted?

no, I don't think so, multiple components will want to see the ARM_STATUS I think.

> please move the message so the ids are not mixed up. 

done

> also mark it as wip. Those are "the rules" until there is a complete implementation and agreement from multiple stacks.

err, really?!? this is in beta release as of today for ArduPilot, will be in general stable release in 2 weeks and widely used. Marking it as "hey, some else else may change this" is absolutely insane.
we need to remove WIP from all the OPEN_DRONE_ID msgs right now

friissoren (approved):
Ok for me.

I don't really have any comments for or against, with regards to the discussion on whether this should be a stand-alone message or merged with some of the other messages being defined in PR #1865.

It would be nice to finally get rid of the WIP markers. For the messages that are already merged, the standardization efforts are completed, so I don't expect any changes from that perspective. 

Whether there is some bug in one of the messages that would require changes is maybe something that still needs some more thorough testing with the Ardupilot implementation? I don't know what the normal policy is.

tridge:
@hamishwillee in case you missed it, ArduPilot does have a complete implementation, I assumed you knew this, but I could have been wrong:
https://docs.google.com/presentation/d/1JgnqcIDn7rGCc8nl46AutO0p0ABODdCgBHFfxmKuUw4/edit?usp=sharing

hamishwillee:
@tridge @friissoren Thanks. The TLDR; is:
- I'm going to add WIP to this and merge. 
- We'll talk about this in the MAV call next week and see if I can get agreement to remove the WIP across all of this.

FYI @auturgy 
----

The MAVLink process is that to become part of the "the standard" definitions get PRd into development.xml. The PR gets accepted if there is intent to prototype from one major flight stack and broad agreement to also adopt by at least one other. When there is a working implementation and sufficient documentation for someone else to unambiguously reproduce the implementation in the opinion of the mav call, another PR is created to push that into common.xml as "common".
The "wip" process predates all this, but is supposed to work the same way. 

The idea is that this will reduce the chance of messages being added speculatively with no implementation, or that will never really be standard (IMO we should add to this a further process to move messages into standard.xml when there really are two compatible implementations, but that's a story for another day.)

The OpenID messages bypassed some of this. They got PRed very early into common with WIP , documented, but at that point the addition was speculative - there was no one implementing. These have not had proper technical review from a second flight stack "to my knowledge" - my sanity check comments don't count. The MAVLink process has not been informed of the implementation other than by hearsay. This might all seem like pointless bureaucracy , but it is intended to reduce friction caused by historical misuse of common/standard.xml.

We'll have a mav call to confirm a path forward. My expectation is that we'll get confirmation from PX4 that this API will be adopted and closer engagement going forward to make sure everyone is happy.

hamishwillee:
@friissoren PS 
- @ThomasDebrunner from Auterion is going to be engaging on OpenID going forward. It is important all interested parties talk. So feel free to ping him on PRs.
- I also got this comment 
  > this is outdated already: https://mavlink.io/en/services/opendroneid.html. Transmission via internet is not an option anymore.

message_definitions/v1.0/common.xml
```diff
@@ -7463,6 +7471,11 @@
       <field type="uint8_t" name="msg_pack_size">Number of encoded messages in the pack (not the number of bytes). Allowed range is 1 - 9.</field>
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
+    <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
+      <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
```
hamishwillee, line 7475:
```suggestion
    <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
        <wip/>
        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
      <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -7463,6 +7471,13 @@
       <field type="uint8_t" name="msg_pack_size">Number of encoded messages in the pack (not the number of bytes). Allowed range is 1 - 9.</field>
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
+    <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
+      <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
+      <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
+    </message>
```
hamishwillee, line 7480: Fix formatting, I hope
```suggestion
    <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
      <wip/>
      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
      <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
      <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
      <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
    </message>
```

friissoren:
> this is outdated already: https://mavlink.io/en/services/opendroneid.html. Transmission via internet is not an option anymore.

@ThomasDebrunner: Neither US nor EU rules require you to use Network Remote ID but it is allowed to use it in addition to Broadcast remote ID. I am quite certain that the specification and rule requirements for Network remote ID will continue to be worked upon by the various authorities in context of UTM and USpace.

We should not exclude the usage of Network remote ID in the general system architecture.

PR #1876, opened 2022-08-11
minimal: try to make component ID less confusing

julianoes:
This is my attempt to avoid some confusion around MAV_COMP_ID_SYSTEM_CONTROL. Essentially, it shouldn't be used anymore.

Replacing it with MAV_COMP_ID_ALL only makes sense as a target though, and not as a source.

@hamishwillee let me know what you think. I had someone confused about this the other day but I can't remember the specifics. What I do remember is making a todo to make it less confusing, and this is my attempt to do so.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -690,7 +690,7 @@
         <description>Component handling TUNNEL messages (e.g. vendor specific GUI of a component).</description>
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
-        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</deprecated>
+        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, system commands should be sent with target_component=MAV_COMP_ID_ALL allowing the target component to use any appropriate component id.</deprecated>
         <description>Deprecated, don't use. Previously indicated component for handling system messages (e.g. to ARM, takeoff, etc.). Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -691,7 +691,7 @@
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
         <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID.</deprecated>
-        <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
+        <description>Deprecated, don't use. Previously indicated component for handling system messages (e.g. to ARM, takeoff, etc.). Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</description>
       </entry>
     </enum>
   </enums>
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -691,7 +691,7 @@
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
         <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, system commands should be sent with target_component=MAV_COMP_ID_ALL allowing the target component to use any appropriate component id.</deprecated>
-        <description>Deprecated, don't use. Previously indicated component for handling system messages (e.g. to ARM, takeoff, etc.). Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</description>
+        <description>Deprecated, don't use. Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
       </entry>
     </enum>
   </enums>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/minimal.xml
```diff
@@ -691,7 +691,7 @@
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
         <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID.</deprecated>
-        <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
+        <description>Deprecated, don't use. Previously used as component for handling system messages (e.g. to arm, takeoff, etc.). Instead, use MAV_COMP_ID_ALL to send a message or command to any component.</description>
```
hamishwillee, line 694: Really feels like they should have got this from the deprecation warning. Maybe we can reduce this a bit. How about:

```suggestion
        <description>Deprecated, don't use. Components that handle system messages (e.g. to arm, takeoff, etc.) must now process commands/messages addressed to MAV_COMP_ID_ALL.</description>
```

julianoes: That doesn't sound wrong but I don't know if that was once implemented that way.

hamishwillee: Hmmm

IDs are just addresses and no longer indicate type. So this is all a bit messy anyway. 

But typically if you set MAV_COMP_ID_SYSTEM_CONTROL  you're saying that the message must go to something with id MAV_COMP_ID_SYSTEM_CONTROL. Any component with another ID it should ignore it. 

In practice the autopilot can only set one ID so it uses [MAV_COMP_ID_AUTOPILOT1](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_AUTOPILOT1). Like all components it also processes messages addressed to MAV_COMP_ID_ALL. It can't use MAV_COMP_ID_SYSTEM_CONTROL "as well" because we don't expose those parts of the system as a separate component of the autopilot.

So actually that is indeed wrong. How about

```suggestion
        <description>Deprecated, don't use. Previously indicated component for handling system messages (e.g. to ARM, takeoff, etc.). Instead, such commands must be sent with target MAV_COMP_ID_ALL and the target may use any appropriate component id.</description>
```

This is not quite correct. It may use "any" component id. But I don't want to suggest that.

hamishwillee:
Works for me! Undraft and let's rubber stamp in the dev call.

message_definitions/v1.0/minimal.xml
```diff
@@ -691,7 +691,7 @@
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
         <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID.</deprecated>
```
hamishwillee, line 693:
```suggestion
        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</deprecated>
```

message_definitions/v1.0/minimal.xml
```diff
@@ -691,7 +691,7 @@
...
-        <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
+        <description>Deprecated, don't use. Previously indicated component for handling system messages (e.g. to ARM, takeoff, etc.). Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</description>
```
hamishwillee, line 694:
```suggestion
        <description>Deprecated (don't use). Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
```

Modified this to push the deprecation info into the deprecation tag. That's what it  is for.

message_definitions/v1.0/minimal.xml
```diff
@@ -690,8 +690,8 @@
         <description>Component handling TUNNEL messages (e.g. vendor specific GUI of a component).</description>
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
-        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID.</deprecated>
-        <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
+        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</deprecated>
```
hamishwillee, line 693:
```suggestion
        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, system commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</deprecated>
```

hamishwillee:
Merging, as discussed in the dev call. FYI @julianoes I modified this a little to move the specific deprecation information inside that tag.

message_definitions/v1.0/minimal.xml
```diff
@@ -690,8 +690,8 @@
         <description>Component handling TUNNEL messages (e.g. vendor specific GUI of a component).</description>
       </entry>
       <entry value="250" name="MAV_COMP_ID_SYSTEM_CONTROL">
-        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID.</deprecated>
-        <description>Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
+        <deprecated since="2018-11" replaced_by="MAV_COMP_ID_ALL">System control does not require a separate component ID. Instead, system commands should be sent with target_component=MAV_COMP_ID_ALL allowing the target component to use any appropriate component id.</deprecated>
+        <description>Deprecated, don't use. Previously indicated component for handling system messages (e.g. to ARM, takeoff, etc.). Instead, such commands must be sent with target_component=MAV_COMP_ID_ALL and the target component may use any appropriate component id.</description>
```
hamishwillee, line 694:
```suggestion
        <description>Deprecated, don't use. Component for handling system messages (e.g. to ARM, takeoff, etc.).</description>
```

PR #1877, opened 2022-08-11
Move MISSION_CHANGED to common

julianoes:
This moves the MISSION_CHANGED message from development.xml to common.xml as it has been successfully implemented end to end.

https://github.com/PX4/PX4-Autopilot/pull/20035
https://github.com/mavlink/MAVSDK/pull/1880

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5269,7 +5269,15 @@
       <extensions/>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
-    <!-- Note, id 52 reserved for MISSION_CHANGED message (development.xml) -->
+    <message id="52" name="MISSION_CHANGED">
+      <description>A broadcast message to notify any ground station or SDK if a mission, geofence or safe points have changed on the vehicle.</description>
+      <field type="int16_t" name="start_index">Start index for partial mission change (-1 for all items).</field>
+      <field type="int16_t" name="end_index">End index of a partial mission change. -1 is a synonym for the last mission item (i.e. selects all items from start_index). Ignore field if start_index=-1.</field>
+      <field type="uint8_t" name="origin_sysid">System ID of the author of the new mission.</field>
+      <field type="uint8_t" name="origin_compid" enum="MAV_COMPONENT">Component ID of the author of the new mission.</field>
+      <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
+    </message>
+    <!-- Note, id 53 reserved for MISSION_CHECKSUM message (development.xml) -->
     <message id="54" name="SAFETY_SET_ALLOWED_AREA">
       <description>Set a safety zone (volume), which is defined by two corners of a cube. This message can be used to tell the MAV which setpoints/waypoints to accept and which to reject. Safety areas are often enforced by national or competition regulations.</description>
       <field type="uint8_t" name="target_system">System ID</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -253,14 +253,6 @@
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
     </message>
     <!-- mission protocol enhancements -->
-    <message id="52" name="MISSION_CHANGED">
-      <description>A broadcast message to notify any ground station or SDK if a mission, geofence or safe points have changed on the vehicle.</description>
-      <field type="int16_t" name="start_index">Start index for partial mission change (-1 for all items).</field>
-      <field type="int16_t" name="end_index">End index of a partial mission change. -1 is a synonym for the last mission item (i.e. selects all items from start_index). Ignore field if start_index=-1.</field>
-      <field type="uint8_t" name="origin_sysid">System ID of the author of the new mission.</field>
-      <field type="uint8_t" name="origin_compid" enum="MAV_COMPONENT">Component ID of the author of the new mission.</field>
-      <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
-    </message>
     <message id="53" name="MISSION_CHECKSUM">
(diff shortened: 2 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5269,7 +5269,15 @@
       <extensions/>
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
     </message>
-    <!-- Note, id 52 reserved for MISSION_CHANGED message (development.xml) -->
+    <message id="52" name="MISSION_CHANGED">
+      <description>A broadcast message to notify any ground station or SDK if a mission, geofence or safe points have changed on the vehicle.</description>
+      <field type="int16_t" name="start_index">Start index for partial mission change (-1 for all items).</field>
+      <field type="int16_t" name="end_index">End index of a partial mission change. -1 is a synonym for the last mission item (i.e. selects all items from start_index). Ignore field if start_index=-1.</field>
+      <field type="uint8_t" name="origin_sysid">System ID of the author of the new mission.</field>
+      <field type="uint8_t" name="origin_compid" enum="MAV_COMPONENT">Component ID of the author of the new mission.</field>
```
hamishwillee, line 5277: Hmmm, not sure about this. There are far more ids than are defined in the enum

hamishwillee:
@julianoes Thank you. It is so long ago since this was proposed I was about to propose we removed it from development.xml until we recovered a stakeholder.

My biggest concern with this is that I am not sure whether this predates our governance model, and hence whether there was broad agreement with ArduPilot to also implement this. 

Further I no longer am confident in this message. I think it is flawed and [MISSION_CHECKSUM](https://mavlink.io/en/messages/development.html#MISSION_CHECKSUM) does what everyone needs much better.
- both messages are supposed to be emitted on mission/plan change, and you can request `MISSION_CHECKSUM` as well. 
- PX4 does not support partial updates, so you are always setting these to -1
- You actually don't know you got every change using mission changed, so you will have to update the whole mission anyway. Whereas if you have the checksum you know whether you match or not.
- Checksum is a more useful way of saying "this is my current mission" or "this is not my current mission".

Only downside is you have to calculate the checksum. I know @peterbarker was talking about implementing the  [MISSION_CHECKSUM](https://mavlink.io/en/messages/development.html#MISSION_CHECKSUM) too. I would actually push to accept that even if he is not, because it is a good idea. 

Upshot, can I change your direction? Sorry to crap on this work, but I think it will save us hassle  in the long run.

message_definitions/v1.0/common.xml
julianoes: I'm confused. I only moved this, and didn't change it.

And what do you mean there are more ids than defined in the enum? The compid is always 0..255, and the MAV_COMPONENT enum, so I don't see what is different now.

julianoes:
> PX4 does not support partial updates, so you are always setting these to -1

That's specced and fine though.

> You actually don't know you got every change using mission changed

Correct.

> I know @peterbarker was talking about implementing the [MISSION_CHECKSUM](https://mavlink.io/en/messages/development.html#MISSION_CHECKSUM) too

How do you know whether the mission changed and you need to request the checksum (again)? Wouldn't you still require MISSION_CHANGED for still?

hamishwillee:
>> I know @peterbarker was talking about implementing the [MISSION_CHECKSUM](https://mavlink.io/en/messages/development.html#MISSION_CHECKSUM) too

> How do you know whether the mission changed and you need to request the checksum (again)? Wouldn't you still require MISSION_CHANGED for still?

No, you should emit [MISSION_CHECKSUM](https://mavlink.io/en/messages/development.html#MISSION_CHECKSUM) on any change to a particular part of the plan or on request.

MISSION_CHANGED tells you that the mission changed, the changer id, and the range of items that changed. 
The range is useless IMO since you can't get this reliably and will almost certainly have to upload the lot. So those should be removed.

Then essentially we are comparing two messages, both of which we could send on change or on request. The main difference is that the checksum allows you to detect a change in any part of the plan, not just the mission, and uses a checksum to identify change, while the mission changed only detects mission changes and uses the ids of the changer to indicate "this change was not made by me". The cost of the checksum is higher, but then you might discover you don't need to update anyway based on the checksum.

So based on the above I think they allow the same use case, but the checksum does it in a more reliable way, at greater cost in computation. Given I know that ArduPilot are interested that feels an easier sell.

Yes, the implementation you have done is to spec and is absolutely fine. I just think that if we were starting from scratch, and we are, we would not have both messages, and we would have picked the checksum. BUT I'm not an implementer. So we should discuss in the mav call if you aren't convinced.

julianoes:
Ok you're right, should have thought of that before doing the implementation, as my todo said I should :joy:.

peterbarker:
On Thu, 11 Aug 2022, Hamish Willee wrote:

> Only downside is you have to calculate the checksum. I know @peterbarker was talking about implementing the MISSION_CHECKSUM too. I would actually push to accept that even if he is not, because it is a good
> idea.

https://github.com/ArduPilot/ardupilot/pull/20834

Note query from tridge on cost of computing checksum.  That might not be 
trivial, and given you have to do it on any change may need to be done on 
a thread.  I haven't measured yet, but I have a suspicion it's going to be 
Not Good.

hamishwillee:
THanks. @julianoes  I guess we might have to resurrect it if the checksum doesn't work out. 

@peterbarker There was also some discussion originally about making the gcs supply the checksum as a mission item. But that has not been fully thought through.

julianoes:
This is still unsolved for me. It just came up in https://github.com/mavlink/MAVSDK-Python/issues/527.

hamishwillee:
@julianoes , so @peterbarker is prototyping the "mission checksum" - which is only incidentally a mission checksum.

The plan is for an opaque checksum to be returned in the MISSION_ACK by the drone. The drone will then include the checksum in mission_current, cycling the value for each plan whenever it is emitted (ie. through mission, geofence, rally points). Make sense.

Perhaps chat to him directly about progress, or I can?

julianoes:
Ok, sounds like we have a plan forward then.

PR #1878, opened 2022-08-11
common.xml:TIMESYNC docs and extensions

hamishwillee:
This adds documentation to the TIMESYNC message including information about timestamp units (ns) and format, and the message sequence.

As discussed in https://github.com/mavlink/mavlink-devguide/pull/433 the original message was broadcast. That means that,  in the case that there might be multiple components requesting timesync, it is not possible to tell if a response is to a particular request. 
ArduPilot encode the system id in the request timestamp (ts1), which is perfectly "safe". However this is not 100% reliable because it is more common to sync multiple _components_ within a single system, so you still can't be sure if a response is intended for a particular component.

Therefore this adds extension fields for `target_system` and `target_component`, which allows robust determination of the intended recipient, and timesyncing to specific targets (eliminating unwanted respones).

@peterbarker @auturgy @julianoes FYI - TBD.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5828,9 +5828,20 @@
       <field type="uint8_t[251]" name="payload">Variable length payload. The length is defined by the remaining message length when subtracting the header and other fields. The content/format of this block is defined in https://mavlink.io/en/services/ftp.html.</field>
     </message>
     <message id="111" name="TIMESYNC">
-      <description>Time synchronization message.</description>
-      <field type="int64_t" name="tc1">Time sync timestamp 1</field>
-      <field type="int64_t" name="ts1">Time sync timestamp 2</field>
+      <description>
+        Time synchronization message.
+        The message is used for both timesync requests and responses.
+        The request is sent with `ts1=requesting component timestamp` and `tc1=0`, and may be broadcast or targeted to a specific system/component.
+        The response is sent with `ts1=requesting component timestamp` (mirror back unchanged), and `tc1=responding component timestamp`, with the `target_system` and `target_component` set to ids of the original request.
+        Systems can determine if they are receiving a request or response based on the value of `tc`.
+        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync.
+        Timestamps are UNIX Epoch time or time since system boot in nanoseconds (the timestamp format can be inferred by checking for the magnitude of the number; generally it doesn't matter as only the offset are used).
+      </description>
+      <field type="int64_t" name="tc1" units="ns">Time sync timestamp 1. Request: 0. Response: Timestamp of remote system.</field>
+      <field type="int64_t" name="ts1" units="ns">Time sync timestamp 2. Request: Timestamp of request. Response: Timestamp of request (mirrored).</field>
+      <extensions/>
+      <field type="uint8_t" name="target_system">Target system id. Request: 0 (broadcast) or id of specific system. Response must contain system id of the requesting component.</field>
+      <field type="uint8_t" name="target_component">Target component id. Request: 0 (broadcast) or id of specific component. Response must contain component id of the requesting component.</field>
     </message>
     <message id="112" name="CAMERA_TRIGGER">
       <description>Camera-IMU triggering and synchronisation message.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5834,7 +5834,7 @@
         The request is sent with `ts1=requesting component timestamp` and `tc1=0`, and may be broadcast or targeted to a specific system/component.
         The response is sent with `ts1=requesting component timestamp` (mirror back unchanged), and `tc1=responding component timestamp`, with the `target_system` and `target_component` set to ids of the original request.
         Systems can determine if they are receiving a request or response based on the value of `tc`.
-        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync.
+        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync; the requestor may report an error.
         Timestamps are UNIX Epoch time or time since system boot in nanoseconds (the timestamp format can be inferred by checking for the magnitude of the number; generally it doesn't matter as only the offset are used).
       </description>
       <field type="int64_t" name="tc1" units="ns">Time sync timestamp 1. Request: 0. Response: Timestamp of remote system.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5831,14 +5831,15 @@
       <description>
         Time synchronization message.
         The message is used for both timesync requests and responses.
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5828,9 +5828,20 @@
...
-      <description>Time synchronization message.</description>
-      <field type="int64_t" name="tc1">Time sync timestamp 1</field>
-      <field type="int64_t" name="ts1">Time sync timestamp 2</field>
+      <description>
+        Time synchronization message.
+        The message is used for both timesync requests and responses.
+        The request is sent with `ts1=requesting component timestamp` and `tc1=0`, and may be broadcast or targeted to a specific system/component.
+        The response is sent with `ts1=requesting component timestamp` (mirror back unchanged), and `tc1=responding component timestamp`, with the `target_system` and `target_component` set to ids of the original request.
+        Systems can determine if they are receiving a request or response based on the value of `tc`.
+        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync.
```
hamishwillee, line 5837: If an updated system was to sync against ardupilot or PX4 as they are now it would know that it wasn't dealing with a valid response. What should it do? I'd say report an error. THoughts?

julianoes: Depends if they require that time sync. If it's a nice to have, then ignore it, otherwise report an error.

julianoes (approved):
I think this is good, and much better than before.

message_definitions/v1.0/common.xml
hamishwillee: Thanks. I'm going to suggest report an error. There really aren't any reasonable fallbacks or ways to interwork the two versions. Given the original is broken I don't think we should try.

```suggestion
        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync; the requestor may report an error.
```

peterbarker:
On Thu, 11 Aug 2022, Hamish Willee wrote:

> Therefore this adds extension fields for target_system and target_component, which allows robust determination of the intended recipient, and timesyncing to specific targets (eliminating unwanted respones).

What does a system do when it receives a timesync with a target 
sydid/compid of 0/0?

hamishwillee:
?> What does a system do when it receives a timesync with a target
sydid/compid of 0/0?

@peterbarker 

> If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync; the requestor may report an error.

What else can it do? It knows that it is working with an old-style timesync. It could in theory still do the timesync thingy with the data it has but the whole point of this is to move to a reliable protocol.

hamishwillee:
>>> What does a system do when it receives a timesync with a target
sydid/compid of 0/0?

>> If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync; the requestor may report an error.

> What else can it do? It knows that it is working with an old-style timesync. It could in theory still do the timesync thingy with the data it has but the whole point of this is to move to a reliable protocol.

We discussed this in the mavlink 20220818 dev call. It should always report an error, but there are several things it might do:
- stop syncing (the current proposal)
- only sync whether there is low probability of clash. For example, monitor for all messages (whether addressed to you or not). If you get any with `tc1=0` there is a possibilty of clash. Wait X seconds before time syncing. 
- treat all messages as sync responses and throw away all significant outliers.

The second option is more work than just updating your receiving systems and is still susceptible to small race errors.

The third option might work particularly if combined with the ardupilot approach of encoding the system id in the ts1 timestamp. The whole thing is statistical anyway - if you have two significantly different source timestamps  you'd easily eliminate the wild outliers. The thing is whether it would work "well enough" for most purposes. Again, I think finding out and verifying this would be more effort than just making the changes in the systems. Even after doing that, you'll still have some greater level of doubt, and you'll still have more traffic from every single system that responded that did not have to.

@peterbarker I think @auturgy was inclined to accept this. Do you have any other suggestions?

tridge (approved):

julianoes:
CI fails until https://github.com/ArduPilot/pymavlink/pull/721 is merged.

message_definitions/v1.0/common.xml
```diff
@@ -5828,9 +5828,20 @@
...
+        The response is sent with `ts1=requesting component timestamp` (mirror back unchanged), and `tc1=responding component timestamp`, with the `target_system` and `target_component` set to ids of the original request.
+        Systems can determine if they are receiving a request or response based on the value of `tc`.
+        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync; the requestor may report an error.
+        Timestamps are UNIX Epoch time or time since system boot in nanoseconds (the timestamp format can be inferred by checking for the magnitude of the number; generally it doesn't matter as only the offset are used).
+      </description>
+      <field type="int64_t" name="tc1" units="ns">Time sync timestamp 1. Request: 0. Response: Timestamp of remote system.</field>
+      <field type="int64_t" name="ts1" units="ns">Time sync timestamp 2. Request: Timestamp of request. Response: Timestamp of request (mirrored).</field>
```
hamishwillee, line 5841:
```suggestion
      <field type="int64_t" name="tc1" units="ns">Time sync timestamp 1. Syncing: 0. Responding: Timestamp of responding component.</field>
      <field type="int64_t" name="ts1" units="ns">Time sync timestamp 2. Timestamp of syncing component (mirrored in response).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5828,9 +5828,20 @@
...
     </message>
     <message id="111" name="TIMESYNC">
-      <description>Time synchronization message.</description>
-      <field type="int64_t" name="tc1">Time sync timestamp 1</field>
-      <field type="int64_t" name="ts1">Time sync timestamp 2</field>
+      <description>
+        Time synchronization message.
+        The message is used for both timesync requests and responses.
+        The request is sent with `ts1=requesting component timestamp` and `tc1=0`, and may be broadcast or targeted to a specific system/component.
+        The response is sent with `ts1=requesting component timestamp` (mirror back unchanged), and `tc1=responding component timestamp`, with the `target_system` and `target_component` set to ids of the original request.
```
hamishwillee, line 5835:
```suggestion
        The request is sent with `ts1=syncing component timestamp` and `tc1=0`, and may be broadcast or targeted to a specific system/component.
        The response is sent with `ts1=syncing component timestamp` (mirror back unchanged), and `tc1=responding component timestamp`, with the `target_system` and `target_component` set to ids of the original request.
```

message_definitions/v1.0/common.xml
```diff
@@ -5828,9 +5828,20 @@
...
+        Systems can determine if they are receiving a request or response based on the value of `tc`.
+        If the response has `target_system==target_component==0` the remote system has not been updated to use the component IDs and cannot reliably timesync; the requestor may report an error.
+        Timestamps are UNIX Epoch time or time since system boot in nanoseconds (the timestamp format can be inferred by checking for the magnitude of the number; generally it doesn't matter as only the offset are used).
```
hamishwillee, line 5838:
```suggestion
        Timestamps are UNIX Epoch time or time since system boot in nanoseconds (the timestamp format can be inferred by checking for the magnitude of the number; generally it doesn't matter as only the offset is used).
```

PR #1879, opened 2022-08-12
development: remove MISSION_CHANGED

julianoes:
This proposed message is no longer required as the plan is to implement MISSION_CHECKSUM instead.


hamishwillee:
Thanks! Lets just "rubber stamp" this in the dev call.

hamishwillee:
This was agreed in mavlink devcall 20220818

Thanks @julianoes

PR #1880, opened 2022-08-14
common: added OPEN_DRONE_ID_SYSTEM_UPDATE message

tridge:
this is a bandwidth efficient way to update the OPEN_DRONE_ID_SYSTEM message data when there is limited uplink bandwidth. Testing on real vehicles shows that with RFD900x radios at an air data rate of 125kbit/s with OPEN_DRONE_ID messages with 1Hz update (as required by FAA RemoteID standard) that there is significant impact on the ability of the GCS to give commands to the flight controller. For example, I got a high degree of packet loss in downloading parameter pre-flight, and many/most in-flight commands failed from the GCS.
By using this message we can use the minimum required bandwidth for updating operator location while remaining FAA RemoteID standard compliant

MAVProxy implementation here: https://github.com/ArduPilot/MAVProxy/pull/1098
ArduPilot MAVLink PR: https://github.com/ArduPilot/mavlink/pull/282
ArduPilot PR: https://github.com/ArduPilot/ardupilot/pull/21435

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7478,6 +7478,15 @@
       <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
       <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
     </message>
+    <message id="12919" name="OPEN_DRONE_ID_SYSTEM_UPDATE">
+      <description>Update the data in the OPEN_DRONE_ID_SYSTEM message with new location information. This can be sent to update the location information for the operator when no other information in the SYSTEM message has changed. This message allows for efficient operation on radio links which have limited uplink bandwidth while meeting requirements for update frequency of the operator location.</description>
+      <field type="uint8_t" name="target_system">System ID (0 for broadcast).</field>
+      <field type="uint8_t" name="target_component">Component ID (0 for broadcast).</field>
+      <field type="int32_t" name="operator_latitude" units="degE7" invalid="0">Latitude of the operator. If unknown: 0 (both Lat/Lon).</field>
+      <field type="int32_t" name="operator_longitude" units="degE7" invalid="0">Longitude of the operator. If unknown: 0 (both Lat/Lon).</field>
+      <field type="float" name="operator_altitude_geo" units="m" invalid="-1000">Geodetic altitude of the operator relative to WGS84. If unknown: -1000 m.</field>
+      <field type="uint32_t" name="timestamp" units="s">32 bit Unix Timestamp in seconds since 00:00:00 01/01/2019.</field>
+    </message>
     <message id="12920" name="HYGROMETER_SENSOR">
       <description>Temperature and humidity from hygrometer.</description>
       <field type="uint8_t" name="id" instance="true">Hygrometer ID</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7479,6 +7479,8 @@
       <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
     </message>
     <message id="12919" name="OPEN_DRONE_ID_SYSTEM_UPDATE">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Update the data in the OPEN_DRONE_ID_SYSTEM message with new location information. This can be sent to update the location information for the operator when no other information in the SYSTEM message has changed. This message allows for efficient operation on radio links which have limited uplink bandwidth while meeting requirements for update frequency of the operator location.</description>
       <field type="uint8_t" name="target_system">System ID (0 for broadcast).</field>
       <field type="uint8_t" name="target_component">Component ID (0 for broadcast).</field>


hamishwillee:
@tridge This makes sense for the "v1" implementation - no point streaming a huge message for a small change.  Would have been even better if we'd split this up originally into high/low rate messages

My concern is that we're currently dealing with a proposed v2 implementation  https://github.com/mavlink/mavlink/pull/1865. IN that the system message was assumed static and is requested on first connection by the transponder. If this is now considered dynamic we need to think it through this message (and the other supposedly static ones).

On alternative would be to have the autopilot actually sent this information in a command. Or to send a command to the transponder to say "request this message again now, things have updated.
Can you advise on the best way to make this work in both v1/v2? Not streaming this at all would be better.

Any advice?

tridge:
@hamishwillee I think we should not move forward with v2 for now. 
I am concentrating on getting a minimal viable system working ready for the deadline next month. I think v2 is a step backwards in that, as it adds too much complexity. Instead we should refine v1 but only in ways that really matter. My flights today show that the bandwidth of the SYSTEM message update to meet the ASTM MOC (which requires 1Hz operator position update) was too high in practice on commonly used radios (RFD900x). So we fix that one issue and test again. We do not have the time for a rewrite.
I am also very wary of implementing pkts like the proposed RADIO_CONTROL packet, as it widens the scope of compliance testing. It means the flight controller, instead of the transmitter module, starts controlling radio transmission methods. This means we need more careful compliance testing of autopilot firmware as the radio tx methods are a critical part of the compliance method.

hamishwillee:
@tridge So just FMI, would you stream both system messages in v1, with the new one  at higher rate - or some other process?

@friissoren @ThomasDebrunner This is an update to the v1 protocol to allow system information to be streamed without choking the link. Given timelines I think it makes sense for v1. Any comments?

tridge:
@hamishwillee yes, both need to be streamed. SYSTEM_UPDATE needs to be 1Hz for standard remote ID under FAA/ASTM. The SYSTEM msg can be low rate.

hamishwillee:
@tridge If 1KHz is a condition of compliance, then shouldn't that be in the description?

tridge:
@hamishwillee 1Hz is required for compliance for FAA/ASTM MOC compliance in the USA for "Standard RemoteID". It is not needed for FAA/ASTM "broadcast" RemoteID in the USA, and each other jurisdiction will be different. We can't put the jurisdiction requirements of every country (and vehicle type in each country) in the xml. It would end up being hundreds of lines long

message_definitions/v1.0/common.xml
```diff
@@ -7478,6 +7478,15 @@
...
       <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
     </message>
+    <message id="12919" name="OPEN_DRONE_ID_SYSTEM_UPDATE">
+      <description>Update the data in the OPEN_DRONE_ID_SYSTEM message with new location information. This can be sent to update the location information for the operator when no other information in the SYSTEM message has changed. This message allows for efficient operation on radio links which have limited uplink bandwidth while meeting requirements for update frequency of the operator location.</description>
+      <field type="uint8_t" name="target_system">System ID (0 for broadcast).</field>
+      <field type="uint8_t" name="target_component">Component ID (0 for broadcast).</field>
+      <field type="int32_t" name="operator_latitude" units="degE7" invalid="0">Latitude of the operator. If unknown: 0 (both Lat/Lon).</field>
+      <field type="int32_t" name="operator_longitude" units="degE7" invalid="0">Longitude of the operator. If unknown: 0 (both Lat/Lon).</field>
```
hamishwillee, line 7486: 0 is a valid value for lat/lon. What if they really are at zero?

hamishwillee: Normally we'd use NaN. Unless you can zero truncate until the end
```suggestion
      <field type="int32_t" name="operator_latitude" units="degE7" invalid="NaN">Latitude of the operator. If unknown: NaN (both Lat/Lon).</field>
      <field type="int32_t" name="operator_longitude" units="degE7" invalid="NaN">Longitude of the operator. If unknown: NaN (both Lat/Lon).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7478,6 +7478,15 @@
...
+      <field type="float" name="operator_altitude_geo" units="m" invalid="-1000">Geodetic altitude of the operator relative to WGS84. If unknown: -1000 m.</field>
```
hamishwillee, line 7487: Can we use NAN because it is clearly and invalid case:

```suggestion
      <field type="float" name="operator_altitude_geo" units="m" invalid="-NaN">Geodetic altitude of the operator relative to WGS84. If unknown: NaN.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7478,6 +7478,15 @@
...
+      <field type="uint32_t" name="timestamp" units="s">32 bit Unix Timestamp in seconds since 00:00:00 01/01/2019.</field>
```
hamishwillee, line 7488: This is fine. FWIW though, should we call it a Unix timestamp if it is not since the unix Epoch? It's just a second timestamp.

```suggestion
      <field type="uint32_t" name="timestamp" units="s">32 bit timestamp in seconds since 00:00:00 01/01/2019.</field>
```

ThomasDebrunner:
This looks ok to me, given the OPEN_DRONE_ID_SYSTEM is pretty big. On more powerful links, this shouldn't really matter though

tridge:
@hamishwillee valid/invalid etc values are defined in openDroneID C library. We need to follow that library exactly. MAVLink is just a transport for filling in the structrures in opendroneid.h
https://github.com/opendroneid/opendroneid-core-c/blob/master/libopendroneid/opendroneid.h
to be useful it must follow that API exactly

hamishwillee:
@friissoren Any last minute comments? If not I plan to merge this tomorrow on the basis that the key autopilot representatives tridge and Thomas have approved this.

message_definitions/v1.0/common.xml
friissoren: Remember that this is given with 7 decimal precision. That roughly corresponds to 1.5 cm precision. I.e. the exact [0.0000000, 0.0000000] coordinate is a few square centimeters [in the middle of the ocean outside the coast of Africa](https://www.google.com/maps/place/0%C2%B000'00.0%22N+0%C2%B000'00.0%22E/@-1.3669564,-0.0594568,4.97z/data=!4m2!3m1!1s0x0:0xbc6dbea4e5fc3562). It is highly unlikely that anyone will be hovering right at that spot.

The remote ID standards use [zero, zero] as the unknown value. It would be easiest to keep the same in the messages as well. The existing ODID messages are already using that convention.

message_definitions/v1.0/common.xml
friissoren: Possible, but again, the standards are defining the invalid value as -1000 and this convention is already used for all altitudes in the existing ODID messages.

message_definitions/v1.0/common.xml
friissoren: The standard define this field as:
"Time of applicability of Location Message expressed as a 32 bit Unix Timestamp (UTC) in seconds since (epoch) 00:00:00 01/01/2019 (to relate back to standard Unix timestamp, add 1546300800 to the common epoch of 00:00:00 01/01/1970)"

Some reference to UTC, Unix or similar at least is needed in order to indicate the timezone reference that this is based on.

message_definitions/v1.0/common.xml
```diff
@@ -7478,6 +7478,15 @@
       <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
       <field type="char[50]" name="error">Text error message, should be empty if status is good to arm. Fill with nulls in unused portion.</field>
     </message>
+    <message id="12919" name="OPEN_DRONE_ID_SYSTEM_UPDATE">
```
friissoren, line 7481: Would this need the WIP marker or not?

hamishwillee: Yes, I'll add one. Though likely we'll remove all of these markers later in the week. Then I plan to merge. Thanks.

message_definitions/v1.0/common.xml
```diff
@@ -7478,6 +7478,15 @@
...
+      <description>Update the data in the OPEN_DRONE_ID_SYSTEM message with new location information. This can be sent to update the location information for the operator when no other information in the SYSTEM message has changed. This message allows for efficient operation on radio links which have limited uplink bandwidth while meeting requirements for update frequency of the operator location.</description>
```
hamishwillee, line 7482:
```suggestion
    <message id="12919" name="OPEN_DRONE_ID_SYSTEM_UPDATE">
      <wip/>
      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
      <description>Update the data in the OPEN_DRONE_ID_SYSTEM message with new location information. This can be sent to update the location information for the operator when no other information in the SYSTEM message has changed. This message allows for efficient operation on radio links which have limited uplink bandwidth while meeting requirements for update frequency of the operator location.</description>
```

friissoren (approved):
Ok for me.

@hamishwillee: I don't know how familiar you are with the drone ID standards. Some quick overview of what fields needs to be populated in what regions [can be found here](https://github.com/opendroneid/opendroneid-core-c#comparison). There are a lot more details to it than that though.

hamishwillee:
@friissoren Thanks - I don't understand it anywhere near as much as I would normally at this point.

But ultimately it isn't my job to assess this. I'm just here to make sure that the key stakeholders have done due diligence and have broad consensus before things merged. We have that. Thanks very much for your advice - are you planning to merge this information into the v1 docs?

friissoren:
Certainly an update of the documentation will be needed, but let's now wait until the implementation work settles down and then document where it ended up.

timtuxworth:
> @hamishwillee 1Hz is required for compliance for FAA/ASTM MOC compliance in the USA for "Standard RemoteID". It is not needed for FAA/ASTM "broadcast" RemoteID in the USA, and each other jurisdiction will be different. We can't put the jurisdiction requirements of every country (and vehicle type in each country) in the xml. It would end up being hundreds of lines long

Why is that a consideration? A few hundred lines of xml is a very small xml document in my experience.

hamishwillee:
@timtuxworth 
1. This XML is small; if we add a few hundred lines to it they will swamp the  relevant content for every other user.
2. It is not a good use of the MAVLink project time to track and maintain different government standards.

PR #1882, opened 2022-08-18
common: OpenDroneID messages no longer WIP

hamishwillee:
This removes wip tagging from the current set of OpenDroneID messages

Note that there was an open question in the dev call about whether `OPEN_DRONE_ID_ARM_STATUS` should/could be represented by a [HEARTBEAT.system_status](https://mavlink.io/en/messages/common.html#HEARTBEAT) of [MAV_STATE_STANDBY](https://mavlink.io/en/messages/common.html#MAV_STATE_STANDBY) (and maybe also [MAV_STATE_ACTIVE](https://mavlink.io/en/messages/common.html#MAV_STATE_ACTIVE)).
The argument being that the semantics are correct and this would (if possible) save unnecessary steaming of a whole additional message.
- James pointed out benefit of having the whole interface together and obvious, not hidden away in heartbeat
- Tridge argues the reason field in  `OPEN_DRONE_ID_ARM_STATUS` is needed.
- Those arguments have merit but you could, for example, emit a status using STATUSTEXT or an EVENT as we do everywhere else. 
- Is the message worth the cost? IMO no, but I'm not the one who has to implement this against very tight external constraints.
- EDIT: Tridge notes that mostly it is the tight constraints. This is not how we want to work "normally".

Anyway, the 20220818 mavlink call agreed this change provided the above was discussed. If there are no objections or follow on comments I plan to merge early next week. 

FYI @friissoren @auturgy @julianoes @tridge

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7490,6 +7490,7 @@
       <field type="uint8_t" name="msg_pack_size">Number of encoded messages in the pack (not the number of bytes). Allowed range is 1 - 9.</field>
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
+    <!-- The message ids 12916 - 12917 are reserved for OpenDroneID. -->
     <message id="12918" name="OPEN_DRONE_ID_ARM_STATUS">
       <description>Transmitter (remote ID system) is enabled and ready to start sending location and other required information. This is streamed by transmitter. A flight controller uses it as a condition to arm.</description>
       <field type="uint8_t" name="status" enum="MAV_ODID_ARM_STATUS">Status level indicating if arming is allowed.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7404,8 +7404,6 @@
       <field type="uint32_t" name="status" display="bitmask" enum="MAV_WINCH_STATUS_FLAG">Status flags</field>
     </message>
     <message id="12900" name="OPEN_DRONE_ID_BASIC_ID">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change, and should NOT be used in stable production environments -->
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -7505,15 +7491,11 @@
       <field type="uint8_t[225]" name="messages">Concatenation of encoded OpenDroneID messages. Shall be filled with nulls in the unused portion of the field.</field>
     </message>
```
friissoren, line 7492:
``` suggestion
    </message>
    <!-- The message ids 12916 - 12917 are reserved for OpenDroneID. -->
```

friissoren (approved):
Ok for me.

I added one small suggested addition, if we don't plan to merge PR #1865 any time soon.

@BluemarkInnovations: FYI

friissoren:
@hamishwillee, @tridge: I just noticed that the MAV_ODID_ARM_STATUS enum doesn't seem to follow the convention that the full enum name is included in the enum value names.
https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L4357

It contains `MAV_ODID_GOOD_TO_ARM` and `MAV_ODID_PRE_ARM_FAIL_GENERIC`.

Shouldn't those be `MAV_ODID_ARM_STATUS_GOOD_TO_ARM` and `MAV_ODID_ARM_STATUS_PRE_ARM_FAIL_GENERIC`?

hamishwillee:
@friissoren Yes they should. I have fixed in https://github.com/mavlink/mavlink/pull/1883 . SHould be fine to change them. - that requires update at compile time, but doesn't break a running system.

I'm holding of merging this to see what happens with that. Also because no one other than you has confirmed they are happy/unhappy with this on the PR. Probably merge anyway next week - not a huge urgency in my mind.

hamishwillee:
ODID messages merged. No objections and already agreed in last dev call.

@friissoren Can you make the necessary update to docs https://mavlink.io/en/services/opendroneid.html ?

What I'm hoping for is to make sure the set of entities are up to date, remove WIP, and capture anything that came out of actually implementing this

For example, discussion of the way things are sent in the first version. We might mention that to meet the US compliance the required streaming rates are xxxx (and note that other domains might have different requirements).

friissoren:
> @friissoren Can you make the necessary update to docs https://mavlink.io/en/services/opendroneid.html ?

https://github.com/mavlink/mavlink-devguide/pull/477

hamishwillee:
Thanks @friissoren - I will attempt to look at it next week!!!

PR #1883, opened 2022-08-24
MAV_ODID_ARM_STATUS entries update to naming convention

hamishwillee:
The MAV_ODID_ARM_STATUS entries did not following naming convention.
This fixes them. 

This is a breaking compile-time change, but does not break existing systems. So "OK".

@tridge @ThomasDebrunner


hamishwillee:
I am merging this. Best to do early and has no runtime impacts. Tridge says OK.

Issue #1860, opened 2022-06-28
Yet another effort towards unified gimbal device messages


olliw42:
Hey Folks, especially Julian (@julianoes)

quite some time ago we had extensive discussions on gimbal protocol v2, and as a result the gimbal device messages resulted. Unfortunately, we could not completely agree on them, and as a result I felt forced to introduce STorM32 variants of them, which can be found in the storm32.xml dialect.

_**Note**_: I am talking in this post exclusively about the gimbal device messages, not the gimbal manager ones.

In fact, the differences in the gimbal device messages as they can be found in common.xml and in storm32.xml are quite small, and I hence felt since ever that it is kind of really sad that a common agreement could not be reached.

I now, with this post, want to again try to come to a common set. This is very largely motivated by the recent efforts of Randy Mackay from the ArduPilot project (https://discuss.ardupilot.org/t/ardupilot-4-3-gremsy-gimbal-improvements-testers-wanted/86479), to implement a driver for the Gremsy gimbals. Importantly, in order to help with that, it appears that Gremsy is very willing to also work on their code, at least they offered some firmware updates related to Randy's efforts. Thus, I think, this is a good moment in time to try again to get a unified set of gimbal device messages, and it might be actually the last chance.

The differences are of two kinds
1. Handling of yaw absolute and relative
2. Handling of RC signals

Justification for these differences:

Both extensions I found to be required in order to deal with the STorM32 gimbal controller properly, and account for real-world applicvations which are still in widespread use.

In gimbal protocol v2 the gimbal is required to understand the notion of relative and absolute yaw, e.g., the ability to set an axis into LOCK mode requires the attitude information to be in earth frame. This however in turn requires (for STorM32, but also Gremsy) that the autopilot sends some data to the gimbal, which not all STorM32 users do. Yet, even though this info to determine absolute yaw is not available to STorM32, it allows an axis to be set to LOCK. And this feature was in fact demonstrated by users in real world to be highly useful for video taking. With gimbal protocol v2 this capability is simply lost completely.

In order to account for that, the STorM32 dialect has few more gimbal device flags, as well as an additional yaw value in the GIMBAL_DEVICE_STATUS message.

[...]

PR #1885, opened 2022-09-05
gimbal: more flexible attitude messages

julianoes:
This is an attempt to improve the attitude setpoint and attitude status messages for gimbals, as suggested by @olliw42 in #1860.

The changes break the coupling between "locking" an angle so it stays the same relative to Earth and the setpoint used and reported to do so.

Previously, these setpoints were possible:

- Pan: gimbal orientation relative to body (in body frame)
- Lock: gimbal orientation relative to earth (in earth frame)

The changes add one more setpoint:

- Lock: gimbal orientation relative to earth (in body frame)

Concerning the attitude status output from the gimbal, it means that the a gimbal can provide the angle either body or earth frame (depending on what it has available), and provide the yaw offset between the two, if available.

This commit tries to add this without breaking existing implementations by adding two flags, and an extension.
The alternative would be to change/replace the existing messages which could lead to a difficult transition but would make the explanations and descriptions easier.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -525,6 +525,15 @@
       <entry value="16" name="GIMBAL_DEVICE_FLAGS_YAW_LOCK">
         <description>Lock yaw angle to absolute angle relative to North (not relative to drone). If this flag is set, the quaternion is in the Earth frame with the x-axis pointing North (yaw absolute). If this flag is not set, the quaternion frame is in the Earth frame rotated so that the x-axis is pointing forward (yaw relative to vehicle).</description>
       </entry>
+      <entry value="16" name="GIMBAL_DEVICE_FLAGS_YAW_LOCK">
+        <description>Lock yaw angle to absolute angle relative to North (not relative to drone). If this flag is set, the quaternion is in the Earth frame with the x-axis pointing North (yaw absolute). If this flag is not set, the quaternion frame is in the Earth frame rotated so that the x-axis is pointing forward (yaw relative to vehicle).</description>
+      </entry>
+      <entry value="32" name="GIMBAL_DEVICE_FLAGS_YAW_IS_BODY">
+          <description>Signals that the yaw of the quaternion and angular rates is either yaw aligned with the vehicle (YAW_IS_BODY) or absolute (YAW_IS_EARTH).</description>
+      </entry>
+      <entry value="64" name="GIMBAL_DEVICE_FLAGS_YAW_IS_EARTH">
+          <description>Signals that the yaw of the quaternion and angular rates is either yaw aligned with the vehicle (YAW_IS_BODY) or absolute (YAW_IS_EARTH).</description>
+      </entry>
     </enum>
     <enum name="GIMBAL_MANAGER_FLAGS" bitmask="true">
       <description>Flags for high level gimbal manager operation The first 16 bits are identical to the GIMBAL_DEVICE_FLAGS.</description>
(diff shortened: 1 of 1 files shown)


olliw42:
@julianoes 
many thx, it is indeed sometimes usefull to just see it how it would look in reality :)

I am fundamentally ok with this choice, but I think there are some points whihc need more unambigious clarification. I will first comment on the current suggestion, then briefly suggest a "simplification" for you to think on and choose :). Lastl, I'd like to list the drawbacks of the approach (again), so that we are just clear about it.

I will call the q & rates in the original part of the message now the "main q&r". 

Comments on current suggestion:
------

1. YAW_IS_BODY & YAW_IS_EARTH dosent' make sense to me, as it's something about the frame. FRAME_IS_BODY or FRAME_IS_EARTH comes to mind, but see also disadvantages below.

2. From the description I infer that your preference is to make the main q&r "toggle" depending on the flags. That's fine to me.

3. We do have now three flags in the game, which give us 8 conditions, and I think we need to be crystal clear for each of the 8 conditions.

YAW_LOCK / IS_BODY / IS_EARTH
0/0/0:        this indicates fall back to the "old" behavior, main q&r is in BODY, and the extensions give the relation to EARTH
1/0/0:        this indicates fall back to the "old" behavior, main q&r is in EARTH, and the extensions give the relation to BODY
0/1/0:        main q&r is in BODY, and the extensions give the relation to EARTH
0/0/1:        main q&r is in EARTH, and the extensions give the relation to BODY

1/1/0:       ?????  we could equal that with 0/1/0, or make it forbidden
1/0/1:       ?????  we could equal that with 0/0/1, or make it forbidden
0/1/1:       ?????  doesn't make any sense and should be forbidden (= message shall be ignored?)
1/1/1:       ?????  doesn't make any sense and should be forbidden (= message shall be ignored?)

4. We need to precisely define the meaning of the values in the extension, in order to be able to calculate correctly. For the quaternion it could be
- q_earth = q_yaw_extension * q_body
- q_earth = q_body * q_yaw_extension
- q_body = q_yaw_extension * q_earth
- q_body = q_earth * q_yaw_extension

There is a "natural" choice for then the yaw axis is the outer axis, like in the most common 3 axis gimbals or pan-tilt gimbals. I just don't remembre from top of my head. Anyway, we need to specify.

Somewhat similarily for rates we have
- w_earth = w_yaw_extension + w_body
- w_body = w_yaw_extension + w_earth

Simplified suggestion
------
From the flags list we can see that we actually do have only 4 conditions to treat, and not 8, and that hence two flags should be fundamentally sufficient. One choice would be to add a new flag which indicates that we have the new extended message, that is to have the two flags

YAW_LOCK / IS_EXTENDED_MSG
0/0:        this indicates fall back to the "old" behavior, main q&r is in BODY, and the extensions give the relation to EARTH
1/0:        this indicates fall back to the "old" behavior, main q&r is in EARTH, and the extensions give the relation to BODY
0/1:        main q&r is in BODY, and the extensions give the relation to EARTH
1/1:        main q&r is in EARTH, and the extensions give the relation to BODY


Drawbacks of method
-----------

1. A fundamental restriction is that only the yaw angle relating earth and body frame is given. This implies that this in fact always can be done, but technically this doesn't always have to be so. I must admit I have difficulties from scratch to come up with a situation there this deficiency would bit, but mathematically it's obvious since also the q which relates earth and body is parameterized in general by 3 parameters.

I have to run now unfortunately, so some more drawbacks might be added in the next time (some there also listed in the discussion issue)

olliw42:
@julianoes 
I guess you can close this
all gimbal device related changes have been merged and everyone seems to be fully happy :)

julianoes:
You're right @olliw42, thanks!

PR #1886, opened 2022-09-07
HOME_POSITION - clarify local frame means NED

hamishwillee:
Removes ambiguity on direction of HOME_POSITION frame. 

@tridge @julianoes OK?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6507,9 +6507,9 @@
       <field type="int32_t" name="latitude" units="degE7">Latitude (WGS84)</field>
       <field type="int32_t" name="longitude" units="degE7">Longitude (WGS84)</field>
       <field type="int32_t" name="altitude" units="mm">Altitude (MSL). Positive for up.</field>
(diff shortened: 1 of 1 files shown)


julianoes (approved):
This describes the status quo for PX4, so I would say that's correct from that side.

julianoes:
It looks like ArduPilot just sets it to all 0: https://github.com/ArduPilot/ardupilot/blob/0c73da5e79e7c8fe91535d95e4f38dab5120a3c3/libraries/GCS_MAVLink/GCS_Common.cpp#L2584

hamishwillee:
Yeah, I think they are implementing it now, and want to make sure that it matches.

mustafa-gokce:
@hamishwillee, @julianoes with https://github.com/ArduPilot/ardupilot/pull/21642, we will also send local position in NED frame in HOME_POSITION message.
Thanks.

tridge (approved):

hamishwillee:
Excellent! Thanks @mustafa-gokce

PR #1887, opened 2022-09-07
pymavlink: update submodule to latest master

julianoes:
This should fix CI failing about the non-existent ns unit.


julianoes:
You were too quick, it was failing CI. Second PR coming.

PR #1888, opened 2022-09-07
Fix CI

julianoes:
Follow up to #1887.

### DIFF
--- a/scripts/format_xml.sh
+++ b/scripts/format_xml.sh
@@ -6,7 +6,7 @@ OPTIND=1         # Reset in case getopts has been used previously in the shell.
 
 # Initialize variables
 mode="format"
-xml_dir="."
+xml_dir="message_definitions"
 keep_old=0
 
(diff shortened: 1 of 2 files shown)


scripts/format_xml.sh
```diff
@@ -6,7 +6,7 @@ OPTIND=1         # Reset in case getopts has been used previously in the shell.
 
 # Initialize variables
 mode="format"
-xml_dir="."
+xml_dir="message_definitions"
```
hamishwillee, line 9: So was this not working at all before?

julianoes: It just tried to fix files inside pymavlink: https://github.com/mavlink/mavlink/runs/8219253471?check_suite_focus=true#step:4:36

julianoes:
Ok, this can be merged, I'd say.

hamishwillee:
OK, on you :-)

PR #1889, opened 2022-09-12
HIL_SENSOR updated to include windsensor measurement

henrykotze:
Updated HIL_SENSOR message to include new [PR](https://github.com/PX4/PX4-SITL_gazebo/pull/909) sensor added in sitl_gazebo.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5821,7 +5821,7 @@
       <field type="float" name="distance" units="m">Distance to the center of the flow field. Positive value (including zero): distance known. Negative value: Unknown distance.</field>
     </message>
     <message id="107" name="HIL_SENSOR">
-      <description>The IMU readings in SI units in NED body frame</description>
+      <description>The mininal set of sensors readings in SI units for various vehicles autonomous flight</description>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
       <field type="float" name="xacc" units="m/s/s">X acceleration</field>
       <field type="float" name="yacc" units="m/s/s">Y acceleration</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5836,6 +5836,8 @@
       <field type="float" name="diff_pressure" units="hPa">Differential pressure (airspeed)</field>
       <field type="float" name="pressure_alt">Altitude calculated from pressure</field>
       <field type="float" name="temperature" units="degC">Temperature</field>
+      <field type="float" name="wind_speed" units="m/s">Wind Speed</field>
+      <field type="float" name="wind_direction" units="rad">Wind Direction</field>
```
Jaeyoung-Lim, line 5840: I don't think we should put this as part of the `HIL_SENSOR` message, but define a new `HIL_*` message since it is not really part of a general sensor setup. Also adding this as an additional field in the mavlink message will break all other setups.

Jaeyoung-Lim (changes requested):

message_definitions/v1.0/common.xml
henrykotze: Cool, I will add another HIL_* message and update it accordingly in sitl_gazebo

hamishwillee: > Adding this as an additional field in the mavlink message will break all other setups that are suing this message definition.

@Jaeyoung-Lim Unless simulators use MAVLink 1 (?) you could add these fields as MAVLInk 2 extensions. 

I have no particular opinion whether it should more generally be the same or a different message. But if we're defining another message we should make sure that all the fields that might be useful are considered for addition.

@henrykotze When these are added 
1. You will need to specify the reference yaw for wind direction - is it NED or fixed the vehicle attitude or something else?
2. If it is possible that this information will not be sent by a  simulator then specify a value to indicate "we're sending invalid data". For a float this is usually NaN.

hamishwillee:
Closing, as it sounds like the proposal is to have a new message.

PR #1890, opened 2022-09-14
Support a return to default speed for DO_CHANGE_SPEED

WickedShell:
DO_CHANGE_SPEED is useful for temporarily overriding flight speeds, but when you are done with an override the only way I can see to either know the default tuned speed (which means reading vehicle parameters/have specific flight controller knowledge), or to change flight modes and then change back, and hope that your flight controller chooses to reset the target speed (which might just be an ArduPilot undocumented behavior). These are obviously both terrible solutions, because one requires specific knowledge about the flight stack, the other requires undocumented behaviour, requires more commands to be sent, and can result in changing more then desired.

This just adds a special case of -2 to be resume the default tuned value for the speed, allowing a very easy end to any speed override. An ArduPilot proposed implementation is available here: https://github.com/ArduPilot/ardupilot/pull/20541

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1353,7 +1353,7 @@
       <entry value="178" name="MAV_CMD_DO_CHANGE_SPEED" hasLocation="false" isDestination="false">
         <description>Change speed and/or throttle set points.</description>
         <param index="1" label="Speed Type" minValue="0" maxValue="3" increment="1">Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)</param>
-        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change)</param>
+        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change, -2 indicates return to default vehicle speed)</param>
         <param index="3" label="Throttle" units="%" minValue="-1">Throttle (-1 indicates no change)</param>
         <param index="4" reserved="true" default="0"/>
         <param index="5" reserved="true" default="0"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1353,8 +1353,8 @@
       <entry value="178" name="MAV_CMD_DO_CHANGE_SPEED" hasLocation="false" isDestination="false">
         <description>Change speed and/or throttle set points.</description>
         <param index="1" label="Speed Type" minValue="0" maxValue="3" increment="1">Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)</param>
-        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change, -2 indicates return to default vehicle speed)</param>
-        <param index="3" label="Throttle" units="%" minValue="-1">Throttle (-1 indicates no change)</param>
+        <param index="2" label="Speed" units="m/s" minValue="-2">Speed (-1 indicates no change, -2 indicates return to default vehicle speed)</param>
+        <param index="3" label="Throttle" units="%" minValue="-2">Throttle (-1 indicates no change, -2 indicates return to default vehicle throttle value)</param>
         <param index="4" reserved="true" default="0"/>
         <param index="5" reserved="true" default="0"/>
         <param index="6" reserved="true" default="0"/>


hamishwillee:
This sounds good to me.

However there is one issue wrinkle that came up recently about its use in mission vs command that I would like to resolve at the same time. The case was that someone was setting the value in a mission and the value was then still being used for goto and orbit commands after the mission completed. IMO that is broken - behaviour in a mission should not propagate.

My expectation is that if used as a command (outside of a mission) this will set the default flight stack cruise speed. 
- That speed will be used by default for goto, orbit, and to set the initial speed in a mission. 
- you could the override it in a mission for the duration of the mission, reverting to the default afterwards.
- If you were to set `-1` or `-2` in this "as a command" case it would have no effect.

In a mission, the behaviour would be as you described. You can override the temporary speed or you can revert to the system default. 

Does that make sense @WickedShell @julianoes and match what ArduPilot/PX4 actually do.

If so, I will spec it that way.

julianoes (approved):

WickedShell:
> My expectation is that if used as a command (outside of a mission) this will set the default flight stack cruise speed.
> 
>     * That speed will be used by default for goto, orbit, and to set the initial speed in a mission.
> 
>     * you could the override it in a mission for the duration of the mission, reverting to the default afterwards.
> 
>     * If you were to set `-1` or `-2` in this "as a command" case it would have no effect.

I'm afraid this doesn't match my expectation (or ArduPilot's behaviour). Speaking just for Plane at the moment (I think the other targets are very similar, but all my comments now are couched around how Plane works). Plane considers the command to be an override for what speed it should be flying at and will fly at the speed in the current mode (assuming it's any mode that uses automatic throttle). The moment the flight mode changes it's automatically reset to the tuned default value. IE the command sets the current speed, not the default speed. And I think that's definitely the correct behavior.

hamishwillee:
@WickedShell Thanks - that sounds very aligned with what @sfuhrer is saying in https://github.com/PX4/PX4-Autopilot/issues/20159#issuecomment-1240619736 .

This looks good to me. I'll see if anyone comments by tomorrow and if not merge then. Thanks for your patience.

I also update the behaviour w.r.t. mode changes separately after I confirm other vehicle types do the same.

sfuhrer (commented):
Not a huge fan of adding more magic values (-1, -2..), but agree that it's a useful extension of the command to have the option to reset to vehicle default speed. 
Shouldn't it the throttle field handles consistently with the speed field? So also there reset to default throttle if value is -2?

WickedShell:
> Shouldn't it the throttle field handles consistently with the speed field? So also there reset to default throttle if value is -2?

Probably, I targeted speed just because that's the one I'm having to deal with at the moment, but a release value for all the fields would make sense to me.

hamishwillee:
@WickedShell Can you put it in now - just to make it clear how this would look and that what we are doing now will not block that in future.

message_definitions/v1.0/common.xml
```diff
@@ -1353,7 +1353,7 @@
       <entry value="178" name="MAV_CMD_DO_CHANGE_SPEED" hasLocation="false" isDestination="false">
         <description>Change speed and/or throttle set points.</description>
         <param index="1" label="Speed Type" minValue="0" maxValue="3" increment="1">Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)</param>
-        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change)</param>
+        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change, -2 indicates return to default vehicle speed)</param>
         <param index="3" label="Throttle" units="%" minValue="-1">Throttle (-1 indicates no change)</param>
```
hamishwillee, line 1357:
```suggestion
        <param index="3" label="Throttle" units="%" minValue="-2">Throttle (-1 indicates no change, -2 indicates return to default vehicle throttle value)</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1353,7 +1353,7 @@
       <entry value="178" name="MAV_CMD_DO_CHANGE_SPEED" hasLocation="false" isDestination="false">
         <description>Change speed and/or throttle set points.</description>
         <param index="1" label="Speed Type" minValue="0" maxValue="3" increment="1">Speed type (0=Airspeed, 1=Ground Speed, 2=Climb Speed, 3=Descent Speed)</param>
-        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change)</param>
+        <param index="2" label="Speed" units="m/s" minValue="-1">Speed (-1 indicates no change, -2 indicates return to default vehicle speed)</param>
```
hamishwillee, line 1356:
```suggestion
        <param index="2" label="Speed" units="m/s" minValue="-2">Speed (-1 indicates no change, -2 indicates return to default vehicle speed)</param>
```

hamishwillee:
OK. I added the throttle reset too. 
- Can you set both the speed and throttle, and if you do, what happens if they imply a different value for the speed? 
-Any objections to this?

hamishwillee:
Devcall agreed this was good. Merged.

PR #1891, opened 2022-09-15
Clarify swarm fields

friissoren:


hamishwillee:
Makes sense. Thanks

hamishwillee:
BTW, I probably should add a linter to auto fix this stuff. One day.

PR #1894, opened 2022-09-22
MAV_FRAME_BODY_FRD - match to implementations and sanity

hamishwillee:
MAV_FRAME_BODY_FRD is actually implemented in ArduPilot and PX4 as a body-aligned local frame, and this is what users would expect.

@tridge @julianoes @dakejahl Are we OK with this wording?


julianoes (approved):

hamishwillee:
This will do then. Thanks all.

chobitsfan:
Hi @hamishwillee Thank you very much for helping. It makes definition more clear.

hamishwillee:
Thanks @chobitsfan - thanks to you and Tridge for keeping chasing this up. I'm very happy to make sensible changes when I can bet consensus and confirmation nothing will be broken.

PR #1895, opened 2022-09-27
MOTOR_INFO message

dakejahl:
Reviving https://github.com/mavlink/mavlink/pull/1695 in with a slightly different take. Any other ideas for useful fields? This would be a really low rate message since these values shouldn't change much. I'm also unsure if including failure flags here makes sense, because if you know that information you're probably getting it from the ESC, in which case ESC_INFO.ESC_FAILURE_FLAGS should tell you what you need to know.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4908,27 +4908,6 @@
         <description>T-Motor</description>
       </entry>
     </enum>
-    <enum name="MOTOR_FAILURE_FLAGS" bitmask="true">
-      <description>Flags to report Motor failures.</description>
-      <entry value="0" name="MOTOR_FAILURE_NONE">
-        <description>No Motor failures.</description>
-      </entry>
-      <entry value="1" name="MOTOR_WARNING_TEMPERATURE_FAULT">
-        <description>Motor has reached a dangerous temperature.</description>
-      </entry>
-      <entry value="2" name="MOTOR_CRITICAL_TEMPERATURE_FAULT">
-        <description>Motor has reached the critical temperature which can lead to compoment breakdown.</description>
-      </entry>
-      <entry value="4" name="MOTOR_OVER_RPM_FAULT">
-        <description>Over RPM failure.</description>
-      </entry>
-      <entry value="8" name="MOTOR_STALL_FAULT">
-        <description>Load applied on the motor is exceeding the breakdown torque.</description>
-      </entry>
-      <entry value="16" name="MOTOR_GENERIC_FAULT">
-        <description>Generic Motor failure.</description>
-      </entry>
-    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
@@ -7034,10 +7013,8 @@
       <description> Contains information about a motor. </description>
       <field type="uint8_t" name="number"> The motor number; index 1, 0 if unknown</field>
       <field type="uint8_t" name="type"> The MOTOR_TYPE of motor </field>
-      <field type="int16_t" name="temperature" units="cdegC"> Temperature of the motor</field>
       <field type="uint32_t" name="accumulated_time" units="s"> Total accumulated usage time. UINT32_MAX if unknown</field>
       <field type="uint32_t" name="serial_number"> Manufacturer serial number. 0 if unknown</field>
-      <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS" display="bitmask"> Bitmask of motor failure flags.</field>
     </message>
     <!-- id 295 reserved for AIRSPEED message (development.xml). -->
     <message id="299" name="WIFI_CONFIG_AP">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7031,13 +7031,13 @@
       <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
     <message id="292" name="MOTOR_INFO">
-        <description> Contains information about a motor. </description>
-        <field type="uint8_t" name="number"> The motor number; index 1, 0 if unknown</field>
-        <field type="uint8_t" name="type"> The MOTOR_TYPE of motor </field>
-        <field type="int16_t" name="temperature" units="cdegC"> Temperature of the motor</field>
-        <field type="uint32_t" name="accumulated_time" units="seconds"> Total accumulated usage time. UINT32_MAX if unknown</field>
-        <field type="uint32_t" name="serial_number"> Manufacturer serial number. 0 if unknown</field>
-        <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS" display="bitmask"> Bitmask of motor failure flags.</field>
+      <description> Contains information about a motor. </description>
+      <field type="uint8_t" name="number"> The motor number; index 1, 0 if unknown</field>
+      <field type="uint8_t" name="type"> The MOTOR_TYPE of motor </field>
+      <field type="int16_t" name="temperature" units="cdegC"> Temperature of the motor</field>
+      <field type="uint32_t" name="accumulated_time" units="seconds"> Total accumulated usage time. UINT32_MAX if unknown</field>
+      <field type="uint32_t" name="serial_number"> Manufacturer serial number. 0 if unknown</field>
+      <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS" display="bitmask"> Bitmask of motor failure flags.</field>
     </message>
     <!-- id 295 reserved for AIRSPEED message (development.xml). -->
     <message id="299" name="WIFI_CONFIG_AP">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4896,6 +4896,39 @@
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
+    <enum name="MOTOR_TYPE">
+      <description>Indicates the motor type.</description>
+      <entry value="0" name="MOTOR_TYPE_UNKNOWN">
+        <description>Motor type not specified.</description>
+      </entry>
+      <entry value="1" name="MOTOR_TYPE_TMOTOR_U8II_KV100">
+        <description>T-Motor</description>
+      </entry>
+      <entry value="2" name="MOTOR_TYPE_TMOTOR_U8II_KV85">
+        <description>T-Motor</description>
+      </entry>
+    </enum>
+    <enum name="MOTOR_FAILURE_FLAGS" bitmask="true">
+      <description>Flags to report Motor failures.</description>
+      <entry value="0" name="MOTOR_FAILURE_NONE">
+        <description>No Motor failures.</description>
+      </entry>
+      <entry value="1" name="MOTOR_WARNING_TEMPERATURE_FAULT">
+        <description>Motor has reached a dangerous temperature.</description>
+      </entry>
+      <entry value="2" name="MOTOR_CRITICAL_TEMPERATURE_FAULT">
+        <description>Motor has reached the critical temperature which can lead to compoment breakdown.</description>
+      </entry>
+      <entry value="4" name="MOTOR_OVER_RPM_FAULT">
+        <description>Over RPM failure.</description>
+      </entry>
+      <entry value="8" name="MOTOR_STALL_FAULT">
+        <description>Load applied on the motor is exceeding the breakdown torque.</description>
+      </entry>
+      <entry value="16" name="MOTOR_GENERIC_FAULT">
+        <description>Generic Motor failure.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6997,6 +7030,15 @@
       <field type="float[4]" name="voltage" units="V">Voltage measured from each ESC.</field>
       <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
+    <message id="292" name="MOTOR_INFO">
+      <description> Contains information about a motor. </description>
+      <field type="uint8_t" name="number"> The motor number; index 1, 0 if unknown</field>
+      <field type="uint8_t" name="type"> The MOTOR_TYPE of motor </field>
```
hamishwillee, line 7036:
```suggestion
      <field type="uint8_t" name="type" enum="MOTOR_TYPE">Motor type</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -6997,6 +7030,15 @@
...
       <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
+    <message id="292" name="MOTOR_INFO">
+      <description> Contains information about a motor. </description>
+      <field type="uint8_t" name="number"> The motor number; index 1, 0 if unknown</field>
+      <field type="uint8_t" name="type" enum="MOTOR_TYPE">Motor type</field>
+      <field type="int16_t" name="temperature" units="cdegC"> Temperature of the motor</field>
+      <field type="uint32_t" name="accumulated_time" units="s"> Total accumulated usage time. UINT32_MAX if unknown</field>
+      <field type="uint32_t" name="serial_number"> Manufacturer serial number. 0 if unknown</field>
+      <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS" display="bitmask"> Bitmask of motor failure flags.</field>
```
hamishwillee, line 7040: bitmask is a property of the enum now (though we haven't purged them).

```suggestion
      <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS">Bitmask of motor failure flags.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -6997,6 +7030,15 @@
...
     </message>
+    <message id="292" name="MOTOR_INFO">
+      <description> Contains information about a motor. </description>
+      <field type="uint8_t" name="number"> The motor number; index 1, 0 if unknown</field>
+      <field type="uint8_t" name="type" enum="MOTOR_TYPE">Motor type</field>
+      <field type="int16_t" name="temperature" units="cdegC"> Temperature of the motor</field>
+      <field type="uint32_t" name="accumulated_time" units="s"> Total accumulated usage time. UINT32_MAX if unknown</field>
+      <field type="uint32_t" name="serial_number"> Manufacturer serial number. 0 if unknown</field>
+      <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS">Bitmask of motor failure flags.</field>
+    </message>
```
hamishwillee, line 7041:
```suggestion
    <message id="292" name="MOTOR_INFO">
      <description> Contains information about a motor.</description>
      <field type="uint8_t" name="number">The motor number, indexed from 1. 0: unknown</field>
      <field type="uint8_t" name="type" enum="MOTOR_TYPE">Motor type</field>
      <field type="int16_t" name="temperature" units="cdegC">Temperature of the motor</field>
      <field type="uint32_t" name="accumulated_time" units="s">Total accumulated usage time. UINT32_MAX if unknown</field>
      <field type="uint32_t" name="serial_number">Manufacturer serial number. 0 if unknown</field>
      <field type="uint16_t" name="failure_flags" enum="MOTOR_FAILURE_FLAGS">Bitmask of motor failure flags.</field>
    </message>
```

message_definitions/v1.0/common.xml
```diff
@@ -6997,6 +7030,15 @@
       <field type="float[4]" name="voltage" units="V">Voltage measured from each ESC.</field>
       <field type="float[4]" name="current" units="A">Current measured from each ESC.</field>
     </message>
+    <message id="292" name="MOTOR_INFO">
+      <description> Contains information about a motor.</description>
+      <field type="uint8_t" name="number">The motor number, indexed from 1. 0: unknown</field>
```
hamishwillee, line 7035: - How can you ever "not know" your motor number?
- Are these directly mapped to esc numbers?

hamishwillee:
1. This would not be merged in common.xml. If you want this to merge it will need to be in development.xml until prototyped.
2. Given we have the ESC info, what are the use cases in which this is useful? i.e. do you genuinely need this, and where would it help you.
3. I imagine the motor failures should reflect actual motor failures that aren't ESC failures. Are you assuming a motor "type" in this? What if it is an ICE motor?

message_definitions/v1.0/common.xml
dakejahl: >How can you ever "not know" your motor number?

Good question... IDK. I just think motor numbers should be 1 indexed which leaves 0 unused :smile: 

>Are these directly mapped to esc numbers?

Another good question... I would say yes they probably should be if you're using ESC mavlink messages

dakejahl:
> 1. This would not be merged in common.xml. If you want this to merge it will need to be in development.xml until prototyped.

Got it. I'll move it.

> 2. Given we have the ESC info, what are the use cases in which this is useful? i.e. do you genuinely need this, and where would it help you.

We have eeproms in each of our motor assemblies. The motor assembly is essentially just a removable arm that contains the motor, the ESC, and this eeprom. The eeprom stores data such as: motor type, serial number, date of assembly, total flight time, etc. We use this information to supplement the airframe file configuration scheme in PX4 such that different motor sets can have different tuning parameters. 

> 3. I imagine the motor failures should reflect actual motor failures that aren't ESC failures. Are you assuming a motor "type" in this? What if it is an ICE motor?

I guess we should maybe use less ambiguous terms. The term "motor" encompasses too many things... I need to think about this.

hamishwillee:
>> Given we have the ESC info, what are the use cases in which this is useful? i.e. do you genuinely need this, and where would it help you.

> We have eeproms in each of our motor assemblies. The motor assembly is essentially just a removable arm that contains the motor, the ESC, and this eeprom. The eeprom stores data such as: motor type, serial number, date of assembly, total flight time, etc. We use this information to supplement the airframe file configuration scheme in PX4 such that different motor sets can have different tuning parameters.

I am not sure about "the right way" here. If this affects tuning it really feels like in PX4 some of this should actually be part of the control allocation: http://docs.px4.io/main/en/concept/control_allocation.html
It might not be obvious from the description there but the actuator setup information is controlled by a configuration file that is set up using [component metadata](https://mavlink.io/en/services/component_information.html).
Perhaps chat to @bkueng about those aspects?

W.r.t. to general metadata most of it sounds like you're actually more interested in asset tracking. Again, not sure that a motor specific message is the right way to do that.

I think we need more thinking about the use cases. Can you outline more clearly how you use each part of the data and how you want to use it in future - e.g.  flight time, temperature? Sorry if I'm being a pain but I know that this will come up again and again if I am not.

dakejahl:
>I am not sure about "the right way" here. If this affects tuning it really feels like in PX4 some of this should actually be part of the control allocation

Actually this information may not be necessary in the mavlink message. A better implementation would use this eeprom data for MOTOR_TYPE to choose the correct airframe file, and looking at SYS_AUTOSTART would tell you which configuration is attached. It can be somewhat useful if 3 out of the 4 motor assemblies are the same, but the fourth is different. But this could also just come down as a STATUSTEXT message to warn the user that the motor assemblies do not match. 

>most of it sounds like you're actually more interested in asset tracking

Yes... this is exactly what I am interested in... except how can we achieve a generic way to track assets that have unique characteristics? A JSON payload would certainly work. Maybe you can tell me if this has ever been discussed before? JSON over mavlink? 


### What the data is used for
`number` : the motor number (see [PX4 airframes](https://docs.px4.io/main/en/airframes/airframe_reference.html))
`type` : the motor type, this would be an enum that maps to something like MOTOR_TYPE_TMOTOR_U8II_KV100. Useful for asset tracking, or the user at a glance knows what propulsion system their using.
`accumulated_time` : total flight time on motor. So you can gauge wear and tear and end of life for the motor
`serial_number` : useful for tracking in a wider system (think database etc)

hamishwillee:
@dakejahl Yes, all that information under "What the data is used for" is for asset tracking. It is of zero interest to a pilot. What a pilot or GCS cares about is whether their vehicle is safe to fly.

Information like temperature and faults may be of interest to a pilot if they affect readiness.

No, there is no "one true way" for this, though it has been chatted about.
I know that smart battery information is used for this kind of purpose. What I want to avoid is adding a message for every single asset that someone decides is needed. Or at least not without collecting together all the assets we might need and seeing if we can handle that in a consistent way. 

One way we might handle this is using https://mavlink.io/en/services/component_information.html with peripheral metadata - or some new "asset tracking metadata".

There is nothing to stop you using generic metadata to map to flight stack parameters - that allows this to be independent of a particular flight stack. So for example, we all agree that the metadata key "motor_runtime" maps to a the flight stack specific parameter that contains the relevant information. So you could then query the vehicle, get the asset listing, get the key that holds the parameter, and then query the parameter value for the current value.

Of course ArduPilot does not yet support component information, so this might not be the right approach. Something to think about though.

I'll see if we can discuss this in tonights mavlink call.

hamishwillee:
>  But this could also just come down as a STATUSTEXT message to warn the user that the motor assemblies do not match.

Or in PX4, and EVENT.

dakejahl:
I'll take a look at the Component Metadata Protocol and see if it can work. Do you know if there are any examples of it in use? 

Seems like this just changes the model from mavlink messages to mavlink ftp -- which I think totally makes sense for single shot unstructured data like this. The caveat is that my "components" (eeprom drivers) would need to write to the file system to update the json file.

dakejahl:
Hmm I don't see any runtime code for modifying the **component_general.json** to append to the `peripherals`. Also if this file is compressed... not sure how we'd write to it. 

Looks like all of the files are generated at build time right now. I like the idea but I'm not sure it works for this use case. If there was a run time mechanism to update the **component_general.json** to add a new uri for a new "peripheral", as well as some helper code to easily generate a json doc (think jsoncpp) from within a driver/module... then it would be useful.

I do wonder if this is outside of the scope of the autopilot... There's got to be others who have their own "drivers" for their own components running on PX4, and they may have some unique data they want to pass around via mavlink. The "custom mavlink messages" problem is so frustrating... I will keep thinking on this

hamishwillee:
@dakejahl So the component metadata thingy is in its infancy. The current implementation is entirely static and done with files known at runtime. We have a clear plan how it could support dynamic addition/removal but that would require @bkueng input.

Chatted about this on the dev call. Some thoughts:
- This could be done with messages. They would be low rate or more likely sent on request. 
- The main argument against is "what messages". We don't want to drip feed a whole bunch of messages - we'd rather make this as generic as possible and know how we'd extend this in future. More messages = more clutter.
- Things we think it might apply to: cameras, gimbals, batteries, motors, propellers. 
- Info needed stuff like identifiers (serial numbers, type etc.)., last maintenance date, time in air since last maintenance. What else for all of these types of things?
- It might not be necessary to standardize for now. 
  - If you've got it on the eeprom you can suck it out and log the information. It would be good to provide an easy way to get the information though because I'm sure the use case will be wanted by others.
  - We assume the main audience might actually not be a GCS but a cloud server.
  
 In summary, more thinking about what information is needed, then we can work out a way to get it where has to go. Our leaning is component metadata though, because that's quite flexible.

dakejahl:
>We don't want to drip feed a whole bunch of messages - we'd rather make this as generic as possible and know how we'd extend this in future

Yeah I totally agree, it needs to be generic.

>Info needed stuff like identifiers (serial numbers, type etc.)., last maintenance date, time in air since last maintenance. What else for all of these types of things?

If we were truly generic, we would not enforce any schema on the payload. It would be some JSON data with perhaps a file size limit.

>We assume the main audience might actually not be a GCS but a cloud server

Yes for something like this it's more likely to be a server than a GCS. 

>Our leaning is component metadata though, because that's quite flexible.

>The current implementation is entirely static and done with files known at runtime. We have a clear plan how it could support dynamic addition/removal but that would require @bkueng input

I think this is the best option if the plan is clear! I can definitely see this being useful in other applications. For example a driver for some thingy that has a bunch of configuration settings and you want to get that block of data in a programmatic way. 

**tangent:** I have actually been using python scripts that run nuttx shell commands over mavlink to query driver information that can't otherwise be easily accessed... having a "standard" (and not totally hacky) way of doing this would be great.

hamishwillee:
I'm not sure whether a json payload and complete flexibility is a good idea. I guess if you argue that asset tracking is always "custom" then sure; but I think a lot of this stuff is common across systems, and if you perhaps want to be able to shift from one asset tracking system to another, having all the information in a common parsable format would be convenient. Might be overthinking it.

> tangent: I have actually been using python scripts that run nuttx shell commands over mavlink to query driver information that can't otherwise be easily accessed... having a "standard" (and not totally hacky) way of doing this would be great.

Wouldn't it just. ArduPilot support Lua scripting for that kind of thing which is pretty magic.

> I think this is the best option if the plan is clear! I can definitely see this being useful in other applications. For example a driver for some thingy that has a bunch of configuration settings and you want to get that block of data in a programmatic way.

Yes, so we thought this through for the dynamic cases like "lua script added parameters", or whatever. There is some discussion of it here: https://github.com/mavlink/mavlink/pull/1806#issuecomment-1068659434
IN essence we'd stream a message with CRCs of component_metadata_json files that can potentially change, allowing a consumer to tell when they need to be re-read.

dakejahl:
>I'm not sure whether a json payload and complete flexibility is a good idea. I guess if you argue that asset tracking is always "custom" then sure; but I think a lot of this stuff is common across systems, and if you perhaps want to be able to shift from one asset tracking system to another, having all the information in a common parsable format would be convenient. Might be overthinking it.

I don't disagree, but using the same logic I would argue a mavlink message makes the most sense then. But I also think mavlink only gets us halfway there. Sure we can add a new ASSET_INFO message that contains some common fields like `serial`, `usage` and `version`. But for any asset that you may want more information about, you'd have to use the COMPONENT_METADATA method. So IMO we may as well use COMPONENT_METADATA from the get go.

>I guess if you argue that asset tracking is always "custom" then sure

I wouldn't argue that it's _always_ custom, but I would argue that in most cases you will want more or less information depending on the asset. I would suggest we impose a certain amount of structure, such as requiring fields like `name[]` and `serial`, but there could also be a nested payload object which is asset dependent, and only meaninful to something that knows what to do with that payload for that given asset.

hamishwillee:
Fair enough. This kind of question is "bigger than me". Dev call would be better, and quick working group even more so.

PR #1896, opened 2022-09-28
MAV_CMD_COMPONENT_ARM_DISARM - support for prearm

hamishwillee:
@ThomasRigi proposed this.

It allows pre-arm via the [MAV_CMD_COMPONENT_ARM_DISARM](https://mavlink.io/en/messages/common.html#MAV_CMD_COMPONENT_ARM_DISARM) command.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1886,7 +1886,7 @@
       </entry>
       <entry value="400" name="MAV_CMD_COMPONENT_ARM_DISARM" hasLocation="false" isDestination="false">
         <description>Arms / Disarms a component</description>
-        <param index="1" label="Arm" minValue="0" maxValue="1" increment="1">0: disarm, 1: arm</param>
+        <param index="1" label="Arm" minValue="0" maxValue="2" increment="1">0: disarm, 1: arm, 2:prearm (non-dangerous actuators such as flight surfaces are allowed to move)</param>
         <param index="2" label="Force" minValue="0" maxValue="21196" increment="21196">0: arm-disarm unless prevented by safety checks (i.e. when landed), 21196: force arming/disarming (e.g. allow arming to override preflight checks and disarming in flight)</param>
       </entry>
       <entry value="401" name="MAV_CMD_RUN_PREARM_CHECKS" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1886,7 +1886,7 @@
       </entry>
       <entry value="400" name="MAV_CMD_COMPONENT_ARM_DISARM" hasLocation="false" isDestination="false">
         <description>Arms / Disarms a component</description>
-        <param index="1" label="Arm" minValue="0" maxValue="2" increment="1">0: disarm, 1: arm, 2:prearm (non-dangerous actuators such as flight surfaces are allowed to move)</param>
+        <param index="1" label="Arm" minValue="0" maxValue="2" increment="1">0: disarm, 1: arm, 2:prearm (arming of non-dangerous actuators such as flight surfaces. Motors remain unpowered).</param>
         <param index="2" label="Force" minValue="0" maxValue="21196" increment="21196">0: arm-disarm unless prevented by safety checks (i.e. when landed), 21196: force arming/disarming (e.g. allow arming to override preflight checks and disarming in flight)</param>
       </entry>
       <entry value="401" name="MAV_CMD_RUN_PREARM_CHECKS" hasLocation="false" isDestination="false">


message_definitions/v1.0/common.xml
```diff
@@ -1886,7 +1886,7 @@
       </entry>
       <entry value="400" name="MAV_CMD_COMPONENT_ARM_DISARM" hasLocation="false" isDestination="false">
         <description>Arms / Disarms a component</description>
-        <param index="1" label="Arm" minValue="0" maxValue="1" increment="1">0: disarm, 1: arm</param>
+        <param index="1" label="Arm" minValue="0" maxValue="2" increment="1">0: disarm, 1: arm, 2:prearm (non-dangerous actuators such as flight surfaces are allowed to move)</param>
```
hamishwillee, line 1889:
```suggestion
        <param index="1" label="Arm" minValue="0" maxValue="2" increment="1">0: disarm, 1: arm, 2:prearm (arming of non-dangerous actuators such as flight surfaces. Motors remain unpowered).</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1886,7 +1886,7 @@
       </entry>
       <entry value="400" name="MAV_CMD_COMPONENT_ARM_DISARM" hasLocation="false" isDestination="false">
         <description>Arms / Disarms a component</description>
-        <param index="1" label="Arm" minValue="0" maxValue="1" increment="1">0: disarm, 1: arm</param>
+        <param index="1" label="Arm" minValue="0" maxValue="2" increment="1">0: disarm, 1: arm, 2:prearm (arming of non-dangerous actuators such as flight surfaces. Motors remain unpowered).</param>
```
hamishwillee, line 1889: Note - MAVLink doesn't define "prearm" so I have added that as a comment. Is that sufficiently tight definition?

I assume we don't need to create an enum here.

ThomasRigi:
Thanks @hamishwillee for opening this PR!

To summarize the conversation we've had so far in [slack](https://px4.slack.com/archives/C52DXKYLV/p1663864313909959):
I would like to be able to (un-)prearm the drone via MAVLink from the onboard computer rather than having to approach the drone to press the safety button.

@auturgy (I hope I tagged you correctly James Pattison) mentioned that this behaviour is already possible with existing mavlink messages via [MAV_CMD_DO_SET_MODE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_MODE) and using [MAV_MODE_FLAG_DECODE_POSITION_SAFETY](https://mavlink.io/en/messages/common.html#MAV_MODE_FLAG_DECODE_POSITION_SAFETY). In fact, MAVProxy uses it this way for ArduPilot.

I think that this existing possible way is too convoluted and un-intuitive, so I'm still in favour of having a more straightforward definition in the ARM_DISARM message as presented in this PR. But I can also understand @auturgy's point on not having duplicate methods.

I'd like to add that I still think that a field 3: un-prearm makes sense for some use cases, and it doesn't hurt IMO to add it now as we're already reworking it. I know that in PX4 you can already un-prearm automatically after every disarming with the parameter [COM_FORCE_SAFETY](https://docs.px4.io/main/en/advanced_config/parameter_reference.html#COM_FORCE_SAFETY), but there might be scenarios where where you might only sometimes want to un-prearm after disarming. E.g. different behaviour after automatic flights vs. manual flights. For example, in our use case we'd have the onboard computer responsible for handling automatic flights, but if we want to do some quick manual flights we don't want to have to approach the drone again after every landing to press the safety button again. Granted, I could just change the value of COM_FORCE_SAFETY, so it's not strictly required, but just more convenient.

hamishwillee:
@ThomasRigi We discussed this in the dev call and have decided not to proceed this.

- Arming is very safety critical changing this message is considered risky.
- There is a concern that the param is sent as an integer encoded as a float. It is possible that an implementation already uses "1 or greater" for ARM. This won't affect PX4 (we checked) but it might affect other stacks.
- The understanding and implementation of "prearm" is different on PX4 and ArduPilot. I am not completely clear on the concern here, but I _think_ they see "not prearmed" as a safety button state.
- This could be implemented by directly setting the modes in the same way as ardupilot (apparently) https://github.com/ardupilot/MAVProxy/blob/master/MAVProxy/modules/mavproxy_arm.py#L158 - so that could be the "standard way" to set prearm.

IF you still want to change the command you could attend the meeting to better understand the implications and what might have to happen for this to be accepted.

hamishwillee:
@ThomasRigi Sorry, I updated this issue when it had not updated and did not see your post.

I agree that the current approach is convoluted, and I argued for this command. The problem is that I don't really understand what the concerns that ArduPilot sees with the prearm state and these interactions - so I can't argue this effectively. I suggest you set up a conversation with @auturgy. Make sense?

ThomasRigi:
I saw the dev call too late, didn't work well with my schedule unfortunately.

Ultimately I'm ok with going the convoluted way if this is deemed better / safer because it clearly does not affect existing mavlink implementations as there is no mavlink change required. No duplicate methods is a point that I can accept. So I'll go ahead and implement the logic in PX4 based on the existing MAVProxy approach, and then we can talk PX4 specifics in a PR there.

PR #1897, opened 2022-09-29
MAV_CMD_BATTERY_RESET move to common

hamishwillee:
This is a proposal to move the `MAV_CMD_BATTERY_RESET` command in the ArduPilot dialect to common.xml.
It is dependent on someone in PX4 agreeing this is useful and they would like to implement it.

----

Power modules are powered on boot and start measuring the consumed power. The assumption is normally that the battery is full on boot, and the remaining power will be the full battery level minus the consumed power.

This works (provided the battery is full on boot) for a single battery. However when you hotswap a battery with a full battery the  values are all immediately wrong because nothing has been consumed of the new battery.

The  `MAV_CMD_BATTERY_RESET` command in the ArduPilot dialect allows a GCS to set the remaining percentage of the battery. So you could immediately set the value back to 100%.
The command should be ignored by a smart battery, which knows its battery level and reports it accurately.

> NOTE: I am not sure how the autopilot would actually manage this. It might just immediately set the original capacity to be doubled (or whatever value is appropriate for the setting). Or it might assume a battery swap, and store the current consumed current as an offset.

The command is also used to reset fuel-based power systems, which are often set to less than 100%.

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -262,11 +262,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="42651" name="MAV_CMD_BATTERY_RESET" hasLocation="false" isDestination="false">
-        <description>Reset battery capacity for batteries that accumulate consumed battery via integration.</description>
-        <param index="1" label="battery mask">Bitmask of batteries to reset. Least significant bit is for the first battery.</param>
-        <param index="2" label="percentage" minValue="0" maxValue="100" increment="1">Battery percentage remaining to set.</param>
-      </entry>
+      <!-- 42651 MAV_CMD_BATTERY_RESET moved to common -->
       <entry value="42700" name="MAV_CMD_DEBUG_TRAP" hasLocation="false" isDestination="false">
         <description>Issue a trap signal to the autopilot process, presumably to enter the debugger.</description>
         <param index="1">Magic number - set to 32451 to actually trap.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2542,6 +2542,11 @@
         <param index="6">Empty.</param>
         <param index="7">Empty.</param>
       </entry>
+      <entry value="42651" name="MAV_CMD_BATTERY_RESET" hasLocation="false" isDestination="false">
+        <description>Reset battery capacity for batteries that accumulate consumed battery via integration.</description>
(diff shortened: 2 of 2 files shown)


hamishwillee:
This fixes a hole with power module based battery level tracking. This is a "discussion" PR. My thoughts:

- "Battery" is an overly specific name for something that can be used for any kind of fuel based system.
- This allows up to 7 batteries to be reset - will we ever need more? I guess if so we could add an extra bitmask.
- Not sure PX4 will be interested. Power modules are supported, but the main audience are moving more towards smart batteries. 
- We should make it clear how it behaves - i.e. does calling this assume a battery swap and treat power module values as an offset.

hamishwillee:
@dakejahl @dagar Is this something that might be useful for PX4 too?

dakejahl:
I think it would be useful for supporting hot swapping. But to really unlock it we'd need to add a button in QGC

hendjoshsr71:
Following....

 Like you said the bitmask approach is limiting. AP already allows 9 battery instances.  (And of course custom code can include far more.) 
 
 For the battery reset button on QGC, I do know someone who has done that but I can't promise it will get released soon to the greater community.

hendjoshsr71:
`MAV_CMD_FUEL_LEVEL` seems fine by me.

If it was adopted I can easily add it to AP along side the current implementation of MAV_CMD_BATTERY_RESET until such time `MAV_CMD_BATTERY_RESET` could be removed from AP.

dakejahl:
>If it was adopted I can easily add it to AP along side the current implementation of MAV_CMD_BATTERY_RESET until such time MAV_CMD_BATTERY_RESET could be removed from AP.

I could easily add it to QGC as well, I'd be happy to do this for the sake of supporting it. I could probably add it to PX4 as well but don't really want to commit to that since it might be more involved to access the battery module directly and change its state. Idk, I'd have to look to know for sure.

hamishwillee:
> I guess there are fewer folks on PX4 using fuel systems? A search in the docs for "fuel" didn't come up with more than 3 hits.

Yes. It seems to be an "area of interest". I just haven't had anything concrete to add to PX4 docs yet.

I'll ask in the dev call. I know that there are a few messages around fuel systems but really it would be good to have started from "what do we need" rather than "how do we extend our battery messages". Might have had same result, but I think we'd have captured more stuff up front.

hamishwillee:
We discussed in the dev call. The PX4 position was that this works for ArduPilot so would probably be fine for PX4. However there is no particular need to merge it right now. Since there is no commitment to integrate in PX4 we should not merge at this time.

I'm going to close this. We can reopen and merge when it is needed. 

Thanks for the discussion. Should simplify things later on.

PR #1898, opened 2022-10-02
ci: add minimum GitHub token permissions for workflow

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using [secure-workflows](https://github.com/step-security/secure-workflows).

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes. Here is an example of the permissions in one of the workflow runs:
https://github.com/mavlink/mavlink/actions/runs/3148604498/jobs/5119313621#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflow:

- test_and_deploy.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- [GitHub recommends](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token) defining minimum GITHUB_TOKEN permissions.
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -12,6 +12,9 @@ on:
 # - "*"
 # - "!README.md" <-- don't rebuild on doc change
 
+permissions:
+  contents: read
+
 jobs:
   format:
     name: Formatting check


hamishwillee:
@julianoes Is this something we want to do? I don't understand whether this will stop deployment phase, and the risks of not accepting this.

julianoes:
@hamishwillee I think this makes sense. I believe our deploy step will still work but we'll find out.

Thanks @boahc077 for the PR!

julianoes (approved):

julianoes:
Push to https://github.com/mavlink/c_library_v2/ still works.

hamishwillee:
Thank you @boahc077 and @julianoes

PR #1899, opened 2022-10-04
Winch extended support

dakejahl:
- Extended status (state) flags 
- Extended actions to support Eject Package and Load Package
- Added Component ID for Winch

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -614,6 +614,9 @@
       <entry value="8" name="WINCH_ABANDON_LINE">
         <description>Spool out the entire length of the line. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
+      <entry value="9" name="WINCH_LOAD_PAYLOAD">
+        <description>Spools out just enough to present the hook to the user to load the payload. Only action and instance command parameters are used, others are ignored</description>
+      </entry>
     </enum>
     <!-- UAVCAN node health enumeration -->
     <enum name="UAVCAN_NODE_HEALTH">
@@ -4652,6 +4655,15 @@
       </entry>
       <entry value="1024" name="MAV_WINCH_STATUS_ABANDON_LINE">
         <description>Winch is abandoning the line and possibly payload. Winch unspools the entire calculated line length. This is a failover state from REDELIVER if the number of attempts exceeds a threshold.</description>
+      </entry>
+      <entry value="2048" name="MAV_WINCH_STATUS_LOCKING">
+        <description>Winch is engaging the locking mechanism.</description>
+      </entry>
+      <entry value="4096" name="MAV_WINCH_STATUS_LOAD_LINE">
+        <description>Winch is spooling on line.</description>
+      </entry>
+      <entry value="8192" name="MAV_WINCH_STATUS_LOAD_PAYLOAD">
+        <description>Winch is loading a payload.</description>
       </entry>
     </enum>
     <enum name="MAG_CAL_STATUS">
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -605,6 +605,9 @@
       <entry value="161" name="MAV_COMP_ID_PARACHUTE">
         <description>Parachute component.</description>
       </entry>
+      <entry value="169" name="MAV_COMP_ID_WINCH">
+        <description>Winch component.</description>
+      </entry>
       <entry value="171" name="MAV_COMP_ID_GIMBAL2">
         <description>Gimbal #2.</description>
       </entry>


message_definitions/v1.0/common.xml
```diff
@@ -614,6 +614,12 @@
       <entry value="8" name="WINCH_ABANDON_LINE">
         <description>Spool out the entire length of the line. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
+      <entry value="9" name="WINCH_LOAD_PACKAGE">
+        <description>Spools out just enough to present the hook to the user to load the package. Only action and instance command parameters are used, others are ignored</description>
```
hamishwillee, line 618: So  what would you do without this? Is it that normally you'd set WINCH_RELAXED and manually drag out the line and attach. So this is just a quick way to spool out a foot or so? Presumably the winch knows how much it should spool out, and you send twice if you need more?

message_definitions/v1.0/common.xml
```diff
@@ -614,6 +614,12 @@
...
+      </entry>
+      <entry value="10" name="WINCH_EJECT_PACKAGE">
+        <description>Ejects the loaded package. Only action and instance command parameters are used, others are ignored</description>
```
hamishwillee, line 621: What does ejecting mean? I'd guess it means that your package can be automatically released from the hook and you can command that from the GCS?

message_definitions/v1.0/common.xml
```diff
@@ -4653,6 +4659,15 @@
       <entry value="1024" name="MAV_WINCH_STATUS_ABANDON_LINE">
         <description>Winch is abandoning the line and possibly payload. Winch unspools the entire calculated line length. This is a failover state from REDELIVER if the number of attempts exceeds a threshold.</description>
       </entry>
+      <entry value="2048" name="MAV_WINCH_STATUS_LOCKING">
+        <description>Winch is engaging the locking mechanism.</description>
```
hamishwillee, line 4663: Why is it useful to know when you are locking?

hamishwillee:
@dakejahl A couple of questions there. There doesn't sound like anything that someone would have to implement so I'm good with it.

@rmackay9 Your comments appreciated.

message_definitions/v1.0/common.xml
dakejahl: >So what would you do without this? Is it that normally you'd set WINCH_RELAXED and manually drag out the line and attach

Yeah basically. Or wind it in and then send the "hold" command just at the right moment so that the hook is dangling but can bear weight because the motor is on.

> Presumably the winch knows how much it should spool out, and you send twice if you need more?

Yeah all these commands assumes the winch knows some things, we expose a ton of these configurable settings via parameters over mavlink. Sending the command twice does nothing on our winch. To get it lock you'd send the LOCK command.

The way it works with our winch is you send the LOAD_PACKAGE command (or press the button on the winch) and that causes the winch to unlock itself, reel down a half foot or so, and then holds that position. When the package is placed on the hook the winch detects the increased load and measures the motor torque which can be converted to weight. You then send the LOCK command (or push that same button again) and the winch reels in and then locks itself.

message_definitions/v1.0/common.xml
dakejahl: Yeah on our winch there is slot where you can place a package to instead do an ejection delivery. The standard winch delivery and ejection delivery are mutually exclusive on our winch but if you were a really savvy designer you could do both on a single mechanism... we tried and it's mechanically very difficult
Selection_098

message_definitions/v1.0/common.xml
dakejahl: IMO it's useful to know what the winch is doing at all times. The winch software is at it's core a state machine and each one of these status flags is a state in our software.

hamishwillee: Sounds reasonable.

message_definitions/v1.0/common.xml
hamishwillee: This makes sense, but I think that others will wonder about it too. So perhaps a comment like "This is used on winches that have automated release mechanisms (or similar)."

This all makes sense to me. Just important that we get broad experience in here, so let's see if @rmackay9 has any comments. If not, remind me next week and I'll chase up offline.

hamishwillee, line 621: Should we generalize this to auto-release? 
```suggestion
      <entry value="10" name="WINCH_AUTO_RELEASE">
        <description>Auto release the winch from its payload (if supported). Only action and instance command parameters are used, others are ignored</description>
```

hamishwillee: Thinking a bit more, proposal is: https://github.com/mavlink/mavlink/pull/1899/files#r992906705

dakejahl: I think "auto" doesn't make a lot of sense because that implies it's automatic, but it's not automatic because you're sending the command to do it. So if anything I'd say release, but I think eject is a more intuitive descriptor than release... but I call chips french fries so what do I know

hamishwillee: @dakejahl I was using "auto" to indicate that it was something you could use to trigger a release rather than manually  unhooking. 

We discussed in the dev call and it seems to me that I don't really understand the hardware capabilities.

When I use a winch on a boat it basically has freewheel, lock, wind. To release a boat, I move the winch to freewheel, manually drag out some line, unhook, then push boat out. To pull boat in I freewheel out the line, attach to boat, engage winch, wind in, lock. 

On the current winch, what I am envisaging is that we have the above but the winch is motorized and we perhaps also have motorised release of the hooking mechanism. So if you were to use WINCH_DELIVER (say), when the winch gets to the bottom it would automatically release the payload [I presume that it cannot do so if it does not have some kind of automated release?]

For this message you're now talking about package release. What I am trying to understand is whether this is a separate mechanism than (say) the mechanised release I refer to above. 

If it is separate, then it might be that this should not be part of the winch, but instead perhaps a separate command like https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GRIPPER

I hope the query about the architecture makes sense? If not, we could jump on a call, or you could come to the mavlink call in a couple of weeks.

dakejahl: >perhaps also have motorised release of the hooking mechanism

It does not have this, it's just a regular hook and the packages are typically gable boxes. If it did have a motorized hook I would agree the DO_GRIPPER command should be used

The winch delivery (which lowers the package on a hook) is different from the eject delivery -- take a look at the image above, the package is located in two different spots. I guess you could argue that this design is both a winch and a gripper. Where you load the package determines which command you use.

Here's a link to the product page where there are some more images, if it still doesn't make sense let's have a quick chat on discord.
https://wattsinnovations.com/pages/reel

dakejahl: I could use MAV_CMD_DO_GRIPPER for the ejection if that makes more sense to you. Since that's really all that slot on the side is, it's just a servo

dakejahl: Although I can still see a use case where the winch design does allow for the package to be lifted/pushed from the hook and released from the winch altogether. Like I mentioned we initially tried to design it this way but mechanically it was very difficult to get right

hamishwillee: For this particular case I think the eject is a gripper. My reasoning is that it has nothing to do with the winch other than the fact that you've mounted it on the same physical hardware.

If this was an electronically controlled hook that had to be used in conjunction with the rest of the winch then you could argue that it might be _either_ a gripper or an inbuilt part of the winch functionality. I can see either approach working.

You'd consider having it as part of the winch (IMO) if you wanted to report on the state in WINCH_STATUS - it is a flaw of our grippers that they cannot report status. It also does make it easier for anyone who has a controllable hooking/release mechanism to "know" that this is how it is released/triggered, rather than understanding that this "other thing" is needed to do that.
I don't think there is anything in the current design though that mandates this - for example [WINCH_DELIVER](https://mavlink.io/en/messages/common.html#WINCH_DELIVER) would automate the case of release for a winch that supported detaching in this way. 

The argument against is the same as for separate camera and gimbal. The payload might potentially do something a bit different to latch/unlatch - having it separate means that you  can extend the behaviour without having to think too much about the winch.

Personally, I think that some kind of auto release/engage as part of the winch mechanism would be a good addition - it is logically part of the winch and all that most would need. 
But I would do your case as a gripper because IMO it is nothing to do with winching.

Let's see what the dev call thinks.

dakejahl: > For this particular case I think the eject is a gripper. My reasoning is that it has nothing to do with the winch other than the fact that you've mounted it on the same physical hardware.

I can't really disagree with your logic here

> Personally, I think that some kind of auto release/engage as part of the winch mechanism would be a good addition - it is logically part of the winch and all that most would need.

Yes I agree. If we had designed it such that the package is loaded onto the winch hook in only one location, and you can choose between a normal winch delivery or an ejection, then this command would be appropriate for that use case. 

> But I would do your case as a gripper because IMO it is nothing to do with winching.

It is a gripper because as you stated above, it is a separate physical thing that happens to be built into the winch hardware, but it is not logically part of the winching mechanism. There is no state so you don't need feedback, all it does is open or close.

All of that being said, I'll just use the DO_GRIPPER command for our use case, but I think the eject/release (whatever we decide to call it) command is still useful in the proper context and should be included.

hamishwillee: @dakejahl So we discussed this in the dev call, and there was broad agreement what I said, and also with your comments here.

> but I think the eject/release (whatever we decide to call it) command is still useful in the proper context and should be included.

We discussed this explicitly.
- We broadly agree  that it would be convenient - in particular if the instruction was very simple like a basic gripper with just "open" and "close" commands. This would not preclude using a gripper or some other more advanced functionality in parallel or instead if you wanted something more than that.
- But we don't really want to explore this now until someone has a winch where this is part of the hooking mechanism so that the different use cases that someone might want to meet can be properly prototyped and tested.

Hope that makes sense?

We can still use "load package" from this PR, but perhaps should be generalized to something like "Load Payload", and you can use that if you need to? [this wasn't discussed in call - is just "IMO"]

dakejahl: Yup that makes sense. Agreed, I renamed to "payload"

message_definitions/v1.0/common.xml
```diff
@@ -614,6 +614,9 @@
       <entry value="8" name="WINCH_ABANDON_LINE">
         <description>Spool out the entire length of the line. Only action and instance command parameters are used, others are ignored.</description>
       </entry>
+      <entry value="9" name="WINCH_LOAD_PAYLOAD">
+        <description>Spools out just enough to present the hook to the user to load the payload. Only action and instance command parameters are used, others are ignored</description>
```
hamishwillee, line 618: I think this is good. If we end up adding mechanised hooks we might also add text that indicates this also makes the hook "open".

hamishwillee (approved):
@julianoes @auturgy I am approving this as the controversial bits are now removed - i.e. the stuff about grippers.

Essentially this now just adds a command to spool out a little line to load a payload, and some additional states. None of this has to be implemented by anyone.

I will merge in a few days if there is no response about approval.

julianoes (approved):

hamishwillee:
Merging. This is non harmful for existing implementations. Thanks for your patience @dakejahl

PR #1900, opened 2022-10-05
Airspeed message: update

IamPete1:
This moves the Airspeed message it to common and re-works it in preparation for a ArduPilot implementation (https://github.com/ArduPilot/ardupilot/pull/21729#issuecomment-1263644965).

This drops several fields to save bandwidth. Changes two pressures to a single raw pressure. Removes error and type in favor of a flags field. Finally it changes to always report indicated airspeed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ include/
 *.pyc
 include/mavlink/v1.0
 .idea
+/.vscode
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -26,22 +26,13 @@
         <description>WPA3.</description>
       </entry>
     </enum>
-    <enum name="AIRSPEED_SENSOR_TYPE">
-      <description>Types of airspeed sensor/data. May be be used in AIRSPEED message to estimate accuracy of indicated speed.</description>
-      <entry value="0" name="AIRSPEED_SENSOR_TYPE_UNKNOWN">
-        <description>Airspeed sensor type unknown/not supplied.</description>
+    <enum name="AIRSPEED_SENSOR_FLAGS" bitmask="true">
+      <description>Airspeed sensor flags</description>
+      <entry value="0" name="AIRSPEED_SENSOR_UNHEALTHY">
+        <description>Airspeed sensor is unhealthy</description>
       </entry>
-      <entry value="1" name="AIRSPEED_SENSOR_TYPE_DIFFERENTIAL">
-        <description>Differential airspeed sensor</description>
-      </entry>
-      <entry value="2" name="AIRSPEED_SENSOR_TYPE_MASS_FLOW">
-        <description>Mass-flow airspeed sensor.</description>
-      </entry>
-      <entry value="3" name="AIRSPEED_SENSOR_TYPE_WINDVANE">
-        <description>Windvane airspeed sensor.</description>
-      </entry>
-      <entry value="4" name="AIRSPEED_SENSOR_TYPE_SYNTHETIC">
-        <description>Synthetic/calculated airspeed.</description>
+      <entry value="1" name="AIRSPEED_SENSOR_USING">
+        <description>True if the data from this sensor is being actively used by the flight controller for guidance, navigation or control.</description>
       </entry>
     </enum>
     <!-- Transactions for parameter protocol -->
@@ -376,12 +367,10 @@
     <message id="295" name="AIRSPEED">
       <description>Airspeed information from a sensor.</description>
       <field type="uint8_t" name="id" instance="true">Sensor ID.</field>
-      <field type="float" name="airspeed" units="m/s">Calibrated airspeed (CAS) if available, otherwise indicated airspeed (IAS).</field>
+      <field type="float" name="airspeed" units="m/s">Calibrated airspeed (CAS).</field>
       <field type="int16_t" name="temperature" units="cdegC">Temperature. INT16_MAX for value unknown/not supplied.</field>
-      <field type="float" name="press_diff" units="hPa">Differential pressure. NaN for value unknown/not supplied.</field>
-      <field type="float" name="press_static" units="hPa">Static pressure. NaN for value unknown/not supplied.</field>
-      <field type="float" name="error" units="m/s">Error/accuracy. NaN for value unknown/not supplied.</field>
-      <field type="uint8_t" name="type" enum="AIRSPEED_SENSOR_TYPE">Airspeed sensor type. NaN for value unknown/not supplied. Used to estimate accuracy (i.e. as an alternative to using the error field).</field>
+      <field type="float" name="raw_press" units="hPa">Raw differential pressure. NaN for value unknown/not supplied.</field>
+      <field type="uint8_t" name="flags" enum="AIRSPEED_SENSOR_FLAGS">Airspeed sensor flags.</field>
     </message>
     <message id="298" name="WIFI_NETWORK_INFO">
       <description>Detected WiFi network status information. This message is sent per each WiFi network detected in range with known SSID and general status parameters.</description>


hamishwillee:
@IamPete1 Thanks. As I understand it the model is that it is modified and prototyped in development.xml, then moves to common.xml when you have an implementation. Can you move the changes to development?

Also:
1. This was introduced in https://github.com/mavlink/mavlink/issues/1584  by @Davidsastresas - David, do you have any comment.
2. I have this ultrasonic airspeed sensor coming along in HIL, which will presumably need a "real" sensor. This has direction - do you think we could add relevant info from that here too? https://github.com/mavlink/mavlink/pull/1893/files - FYI @henrykotze
3. @sfuhrer FYI, new airspeed message. Been in development.xml for a while. Doesnt appear to be used by PX4. Does it "do what you need"? Note, this is slightly modified version of what is in dev as per comments

IamPete1:
I can split into a rework PR and a move to common PR, the intent is do the AP implementation as soon as this gets the OK.

re - 2. I don't think we should add fields to this costing bandwidth for everyone, such rarely used sensors would be better with a dedicated message. Or this message could be used for the forward complement and a angle of attack / side slip message added to allow the 3D vector to be re-constructed. BTW, we do support such sensors in ArduRover already, there we just re-use the wind message.

re - 4. Yes we have our sights set on killing scaled pressure (and scaled pressure 2 and 3), to complete that we would need a dedicated baro message in a similar style to this one. id, altitude, pressure, temperature and flags.

hamishwillee:
Thanks @IamPete1 
- Re "2", if you mean [WIND_COV](https://mavlink.io/en/messages/common.html#WIND_COV), that might be a better place for this anyway. I have no opinion - I just want related parts discussed together so we don't miss anything obvious. 
- Awesome re scaled pressure. One message that fits the purpose a good thing.

I would just rework this message in development to start with. Once you've prototyped then we can look at the PR to move it into common. I'm happy enough to discuss in the PR as it is, but I wouldn't press the merge button until there was consensus.

Davidsastresas:
Thanks for letting me know @hamishwillee, and I am sorry I wasn't aware of you latest message in the issue thread, I missed it. I am usually available in Discord if at some point it is needed to reach me quickly.

We can adapt QGC real quick to whatever changes are introduced when this gets stable in Ardupilot. We will prepare a PR for QGC when this happens.

Also thank you Peter for working on this.

message_definitions/v1.0/development.xml
```diff
@@ -270,12 +261,10 @@
     <message id="295" name="AIRSPEED">
       <description>Airspeed information from a sensor.</description>
       <field type="uint8_t" name="id" instance="true">Sensor ID.</field>
-      <field type="float" name="airspeed" units="m/s">Calibrated airspeed (CAS) if available, otherwise indicated airspeed (IAS).</field>
+      <field type="float" name="airspeed" units="m/s">Indicated airspeed (IAS).</field>
```
hamishwillee, line 264: Why this change? Doesn't it make sense we're just saying it's airspeed that might be CAS or indicated?

message_definitions/v1.0/development.xml
```diff
@@ -270,12 +261,10 @@
...
       <field type="int16_t" name="temperature" units="cdegC">Temperature. INT16_MAX for value unknown/not supplied.</field>
-      <field type="float" name="press_diff" units="hPa">Differential pressure. NaN for value unknown/not supplied.</field>
-      <field type="float" name="press_static" units="hPa">Static pressure. NaN for value unknown/not supplied.</field>
-      <field type="float" name="error" units="m/s">Error/accuracy. NaN for value unknown/not supplied.</field>
-      <field type="uint8_t" name="type" enum="AIRSPEED_SENSOR_TYPE">Airspeed sensor type. NaN for value unknown/not supplied. Used to estimate accuracy (i.e. as an alternative to using the error field).</field>
```
hamishwillee, line 278: I don't know enough to have an opinion on the removals here. Why are they not useful?

message_definitions/v1.0/development.xml
```diff
@@ -26,22 +26,13 @@
...
-        <description>Differential airspeed sensor</description>
-      </entry>
-      <entry value="2" name="AIRSPEED_SENSOR_TYPE_MASS_FLOW">
-        <description>Mass-flow airspeed sensor.</description>
-      </entry>
-      <entry value="3" name="AIRSPEED_SENSOR_TYPE_WINDVANE">
-        <description>Windvane airspeed sensor.</description>
-      </entry>
-      <entry value="4" name="AIRSPEED_SENSOR_TYPE_SYNTHETIC">
-        <description>Synthetic/calculated airspeed.</description>
```
hamishwillee, line 44: Is it not useful to know the sensor type? Again, I don't "care" but some thought obviously went into this message originally, so it would be good to understand why the thinking has changed.

IamPete1: It could be useful, but we don't need to stream it at multiple Hz, were not expecting the type to change in flight. In the case of AP one could pull from params.

message_definitions/v1.0/development.xml
IamPete1: I have never come across a sensor in the wild that actually gives a accuracy, so in most every case we would just be sending NaN. Of course we could derive an accuracy based on sensor fission of the EKF, but we have other messages to give the EKF accuracy. 

For type see https://github.com/mavlink/mavlink/pull/1900#discussion_r989512623.

message_definitions/v1.0/development.xml
IamPete1: My thinking was that this is the direct sensor reading, so would be indicated. As the original message was structured there was no way to tell if the value was CAS or IAS, so very hard for the user to tell what is going on (i.e. is IAS wrong or is it CAS). We do now have a flags field, so I guess we could use one of our flag bits to show that were sending CAS, but until anyone actually has a need for it I would be tempted to leave it. A GCS should be able to calculate it using the other data it has in any case.

message_definitions/v1.0/development.xml
```diff
@@ -26,22 +26,13 @@
         <description>WPA3.</description>
       </entry>
     </enum>
-    <enum name="AIRSPEED_SENSOR_TYPE">
-      <description>Types of airspeed sensor/data. May be be used in AIRSPEED message to estimate accuracy of indicated speed.</description>
-      <entry value="0" name="AIRSPEED_SENSOR_TYPE_UNKNOWN">
-        <description>Airspeed sensor type unknown/not supplied.</description>
+    <enum name="AIRSPEED_SENSOR_FLAGS" bitmask="true">
+      <description>Airspeed sensor flags</description>
+      <entry value="0" name="AIRSPEED_SENSOR_HEATHY">
```
IamPete1, line 31:
```suggestion
      <entry value="0" name="AIRSPEED_SENSOR_HEALTHY">
```

message_definitions/v1.0/development.xml
IamPete1: Same thinking for static pressure, very few sensors give a separate static reading, vast majority are just differential, so we would be sending NaN most of the time.

sfuhrer (approved):
I like the new AIRSPEED_SENSOR_FLAGS flags, it's useful to know the heath state of all airspeed sensors and which one is currently used. 

Also agree that the "CAS is available, IAS otherwise" makes it quite ambiguous, and am okay to keep it to IAS. Adding another field (CAS) would be fairly straight forward, and if anybody in PX4 ever feels the need to implement this message then I will maybe propose to add the CAS to a separate field (as in PX4 every sensor's IAS gets auto-scaled to a CAS by the FC).

message_definitions/v1.0/development.xml
```diff
@@ -26,22 +26,13 @@
...
       </entry>
     </enum>
-    <enum name="AIRSPEED_SENSOR_TYPE">
-      <description>Types of airspeed sensor/data. May be be used in AIRSPEED message to estimate accuracy of indicated speed.</description>
-      <entry value="0" name="AIRSPEED_SENSOR_TYPE_UNKNOWN">
-        <description>Airspeed sensor type unknown/not supplied.</description>
+    <enum name="AIRSPEED_SENSOR_FLAGS" bitmask="true">
+      <description>Airspeed sensor flags</description>
+      <entry value="0" name="AIRSPEED_SENSOR_HEALTHY">
+        <description>Airspeed sensor is healthy</description>
```
tridge, line 32: it would save us one byte on the wire if we made it an UNHEALTHY flag instead of HEALTHY, as flags would normally be zero and would be truncated

IamPete1: Done, Thanks.

IamPete1:
Change healthy flag to unhealthy per @tridge's suggestion.

henrykotze:
Thanks.

Just sharing for more context : 
Re: "2", our mavlink message is intended to be used in HIL setups and as such sends the "raw" measurement from the sensor itself.  The sensor is used to estimate the wind speed and wind direction, but its measurement will be the combined airflow of the vehicle (due to it moving) and the wind, so there needs to be some processing to remove the vehicles movements to the measurement to get the true wind speed and direction.

.gitignore
```diff
@@ -28,3 +28,4 @@ include/
 *.pyc
 include/mavlink/v1.0
 .idea
+/.vscode
```
hamishwillee, line 31: Not keen on unrelated changes in PRs.

message_definitions/v1.0/development.xml
```diff
@@ -26,22 +26,13 @@
...
-        <description>Mass-flow airspeed sensor.</description>
-      </entry>
-      <entry value="3" name="AIRSPEED_SENSOR_TYPE_WINDVANE">
-        <description>Windvane airspeed sensor.</description>
-      </entry>
-      <entry value="4" name="AIRSPEED_SENSOR_TYPE_SYNTHETIC">
-        <description>Synthetic/calculated airspeed.</description>
+      <entry value="1" name="AIRSPEED_SENSOR_USING">
+        <description>Vehicle is using this sensor for navigation</description>
       </entry>
```
hamishwillee, line 36: My understanding here is that a flight controller might have multiple airspeed sensors, only some of which are in "active use" by an estimator - you want to show which. This is not, for example, to show that we're in MC mode on a VTOL (say) and that the sensor is not being used anymore?

If so, you could argue the same thing as for healthy/unhealthy - if you have only one sensor, won't you most commonly be using it for navigation? So this would always be true. 

It seems a bit weird to me, but perhaps:

```suggestion
      <entry value="1" name="AIRSPEED_SENSOR_INACTIVE">
        <description>This sensor not currently being used for navigation</description>
      </entry>
```

If this is just generally to indicate the status of the sensor system, should we instead/as well have an addition to [](https://mavlink.io/en/messages/common.html#SYS_STATUS).

hamishwillee:
Other than comment to https://github.com/mavlink/mavlink/pull/1900/files#r992874417 looks good to me.

@sfuhrer Any objections to merging this?

message_definitions/v1.0/development.xml
sfuhrer: I think two separate flags are better - a sensor can be "healthy" and "not using", and this distinction may be useful to know. From the "invalid" flag its not possible to know if the sensor is not used because it's "unhealthy" or just not "used", so it's less informative.

Also I would not link it with flight controller states (e.g. VTOL mode), as it is hard for the instance publishing this "inactive" flag to know whether the airspeed data is used somewhere (even in Hover mode it can totally be used).

message_definitions/v1.0/development.xml
```diff
@@ -26,22 +26,13 @@
...
-      <entry value="2" name="AIRSPEED_SENSOR_TYPE_MASS_FLOW">
-        <description>Mass-flow airspeed sensor.</description>
-      </entry>
-      <entry value="3" name="AIRSPEED_SENSOR_TYPE_WINDVANE">
-        <description>Windvane airspeed sensor.</description>
-      </entry>
-      <entry value="4" name="AIRSPEED_SENSOR_TYPE_SYNTHETIC">
-        <description>Synthetic/calculated airspeed.</description>
+      <entry value="1" name="AIRSPEED_SENSOR_USING">
+        <description>Vehicle is using this sensor for navigation</description>
```
sfuhrer, line 35: Why does it need to be constrained to navigation? What about guidance and control?

IamPete1: I think this is just poor wording. I was was trying to say if it was being used for anything, I.E. "not reporting only". I guess we could just change the wording to exactly that. As you say "guidance and control" is better than "navigation".

message_definitions/v1.0/development.xml
IamPete1: The intent was not to link in VTOL modes. Just as a way to distinguish which out of a number of sensors is the one that were currently using / would be using. Its common to have two healthy sensors, one of which is the primary and is used preferentially.

message_definitions/v1.0/development.xml
sfuhrer:
```suggestion
        <description>True if the data from his sensor is currently being actively used by the flight controller for guidance, navigation or control</description>
```

IamPete1: updated.

IamPete1:
Updated the `AIRSPEED_SENSOR_USING` description.

message_definitions/v1.0/development.xml
hamishwillee: @auturgy Are you OK with this?

hamishwillee:
We chatted in the dev call, and @auturgy wanted to think about this a little more. 

My two bits (not opinion of the dev call) are 
- the flags are appropriate - there is no other way to get this information; even if we were to push the airspeed selection into EKF it would be unlikely that EKF would output additional detail of the inputs that were not selected.
- Knowing whether you have CAS or IAS seems to be important for users. I think that if all sensors can supply CAS then that is a better value to send (even if it can be derived from IAS).

IamPete1:
> Knowing whether you have CAS or IAS seems to be important for users. I think that if all sensors can supply CAS then that is a better value to send (even if it can be derived from IAS).

Knowing which you have is important, in as it currently stands you know will always have IAS. I do not know of any sensors that actually provide CAS. It is typically calculated using information from the barometer. The intent here is for a airspeed sensor message, I'm not sure it makes sense to report a number derived using values from other sensors. However I can see that some may want it. We have six bits of our flags field left. I don't see any problem with adding a flag bit to indicate that the reported airspeed is CAS, that could be added in the future. 

I'm not a fan of adding CAS as a extension. Typically a single scale factor would be used to convert all IAS readings to CAS. Better to report that scale factor directly than use it to calculate CAS for x sensors and report them all. 

IMHO CAS is better reported as a single value from the primary sensor in `VFR_HUD`. E.G. per vehicle not per sensor.

This message is still in `developlent.xml`, for the time being in any case. I'm keen to get this in so we can move forward with the ArduPilot implementation.

hamishwillee:
@IamPete1 From my perspective it can go in, pending you having a discussion with @auturgy .

hamishwillee (approved):

.gitignore
hamishwillee: Fine in this case :-)

hamishwillee:
@IamPete1 Sorry for the delay. @auturgy tells me he got your ping and is still trying to find time to evaluate this.

auturgy:
The clarification I need is whether this is a raw data message from a sensor (ie IAS), or reporting of processed data from an autopilot (could/should be CAS), or either - in which case we need a bit flag to indicate.
I suspect that the latter is the way to go.

IamPete1:
@auturgy As the message is written it is clear that is it IAS.

```<field type="float" name="airspeed" units="m/s">Indicated airspeed (IAS).</field>```

Possibly you want clarification on what type of airspeed people actually want rather than what is currently in the message?

I think IAS is correct, see https://github.com/mavlink/mavlink/pull/1900#issuecomment-1276829540

**Comment by hamishwillee** [comment]
I think that @auturgy is asking about the intended consumer of this data - who is the intended recipient and what will they do with the data, and what is supplying the data.

If the theory is that this is just raw airspeed from a sensor then we probably should call the message `AIRSPEED_RAW` - then there is no question of the creator, and that it must be IAS. If this is data synthesized for display on a GCS then the preference would be CAS. If it might be either then you probably need to make it be either and indicate the type.

**Comment by IamPete1** [comment]
Rebased and renamed to airspeed raw. 

I really think we should be focusing more on keeping minimum bandwidth for the things users want now than trying to foresee every combination of things users may want in the future. If we overlook something we can add it in the future.

**Comment by hamishwillee** [comment]
> I really think we should be focusing more on keeping minimum bandwidth for the things users want now than trying to foresee every combination of things users may want in the future.

The counter argument is that the need for a CAS based indicator is forseeable, "cheaper" than having a separate message, and cleaner than adding an extension. I don't think it matters much for this case, but as a general statement that is arguable.

Anyway, @auturgy now that it has been made clear that this is purely sensor output for consumption by a GCS however it likes, are you OK with it?

**Comment by auturgy** [comment]
Just to confirm with @IamPete1 : you intend to only be sending raw_airspeed?  Why?  What use is that to a GCS operator by itself? The changes made in this PR don't make sense when referenced against the ArduPilot PR.
ArduPilot calculates both airspeed (basically CAS) and raw_airspeed (basically IAS).  Both are logged, but raw_airspeed isn't sent via mavlink because it isn't useful to do so (apart from logging, raw_airspeed isn't used at all, not even for derivation of airspeed...)

**Review by tridge — APPROVED** [review]
looks good, thanks @IamPete1 !

**Comment by tridge** [comment]
@auturgy we really only have IAS and TAS in ArduPilot. IAS includes compensation for the sensor offset, but no other factors. TAS is just IAS*EAS2TAS
we do have an experimental branch that adds CAS, which matters when compressibility matters at high speed. 
@IamPete1 I'd be fine with just calling this AIRSPEED. If we do end up merging the compressibility compensation then we likely would want to send the result in this msg

**Comment by tridge** [comment]
@IamPete1 actually, we do always include the airspeed ratio, which means what we will be sending is a CAS, not an IAS

**Comment by IamPete1** [comment]
Changed back to `AIRSPEED` and from IAS to CAS.

I can see the argument that the internal calibration we do moves from IAS to sort of CAS. However we still do not to compressibility correction, so its not really CAS. 

My hesitation in moving away from IAS is that we start to include reading from other sensors (baro), so it becomes a derived number. 

> you intend to only be sending raw_airspeed? Why? What use is that to a GCS operator by itself?

I don't think we need to restrict ourselves to only sending values that are directly useful to the pilot. We should send the minimum amount of data that lets a GCS calculate the useful values. To take the argument to the extreme, we send airspeed in meters per second. I'm sure many pilots would find knots more useful. Does this mean we should send airspeed in knots? Or that we should send the value in knots and meters per second? Of course not! We send the value once and rely on the GCS to present the data in the way that best suits the pilot.

**Comment by tridge** [comment]
> However we still do not to compressibility correction, so its not really CAS.

compressibility is EAS (equivalent airspeed), not CAS

File: message_definitions/v1.0/development.xml [hunk 9]
```diff
@@ -270,12 +261,10 @@
     <message id="295" name="AIRSPEED">
       <description>Airspeed information from a sensor.</description>
       <field type="uint8_t" name="id" instance="true">Sensor ID.</field>
-      <field type="float" name="airspeed" units="m/s">Calibrated airspeed (CAS) if available, otherwise indicated airspeed (IAS).</field>
+      <field type="float" name="airspeed" units="m/s">Cailbrate airspeed (CAS).</field>
```
hamishwillee, line 264:
```suggestion
      <field type="float" name="airspeed" units="m/s">Calibrated airspeed (CAS).</field>
```

hamishwillee:
@auturgy @IamPete1 @tridge Modulo the typo fix in https://github.com/mavlink/mavlink/pull/1900#discussion_r1005040531 are we OK with this?

FWIW I am OK with it provided the assumption is true that all senders can provide CAS and that CAS is the desirable format for a user of this particular message. If those aren't true, then we could add a flag to indicate IAS being used - but it sounds like that is not the case?

hamishwillee:
@auturgy Is still not sure about this and plans to discuss with Pete. I believe the fundamental problem is that it is still not clear what this is really for - i.e. is it just for display on a GCS, or is it for something else. I'm absenting myself from the discussion until he posts.

message_definitions/v1.0/development.xml
IamPete1: Yep, thanks.

IamPete1:
Fixed the typo. 

@auturgy We don't currently have any way to directly report airspeed readings from sensors. We do report a vehicle level airspeed in `VFR_HUD`. So this is just to allow reporting of airspeed readings from sensors. It is also a step towards retiring the three scaled pressure messages. Currently in AP were using the named value float message to report the second airspeed sensor reading because there is nothing else. 

Just a reminder that this is still in `development.xml`, so we can change it in the future, it is a shame there was not this level of review when the message was first added.

IamPete1:
Rebased, should get past CI now.

IamPete1:
@hamishwillee @auturgy Any updates?

hamishwillee:
From my perspective, this is waiting on @auturgy and you to talk (still). Last time we spoke he sounded most interested in "how this is going to be used", "who is it for" - as those are the questions that define what data and data format you need to carry

I'll ask him again at the call this evening.

IamPete1:
@auturgy Any update? Do you have any questions?

auturgy:
I'm happy. Apologies for the delay.

auturgy:
I'm going to merge this: I was the only hold up.

PR #1902, opened 2022-10-10
Add radiation sensor message

marcirsch:


hamishwillee:
Will review this when it is out of draft. FMI is there any context? I.e. would be good to know if there is an associated project, which flight stacks you expect to integrate this in first, what sort of timeframes you're looking at.

marcirsch:
@hamishwillee Thanks for jumping on it so quickly! This is coming from the Auterion dialect, more context is also coming

hamishwillee:
Great! I'll keep a lookout then, but ping me if you mark public and I don't respond in a timely manner.

PR #1903, opened 2022-10-13
MAV_CMD_SET_MESSAGE_INTERVAL - default interval may be 0

hamishwillee:
MAV_CMD_SET_MESSAGE_INTERVAL can be used to set a default interval. This update makes it clear that the default interval might be 0 (and probably will be if the message isn't streaming when this is first sent).

This is to reduce the minor risk that someone will call the method expecting default to mean "some sensible non-zero value"; if the value is not streaming then the sensible default value is zero.

As discussed with @peterbarker .


peterbarker (approved):
Thanks!

julianoes (approved):

PR #1908, opened 2022-11-01
Component Metadata: Add Frames

junwoo091400:
## About
This is a MAVLink part for the Airframe selection improvement project I have been working on for the last 2 weeks.

Frame metadata is meant to serve as a definite reference that can be dynamically fetched by the GCS (as done with all other component metadata), to show the frames that are actually supported.

Additionally, the frame JSON schema is defined in a way where we can structure the frames within another frame in a cascaded fashion, which enables a User-Interface workflow that wasn't possible before, which is a JSON- defined Frame selection workflow.

This is best demonstrated by the GIF below.

### Notes
For now, the JSON Schema & Examples were fetched from [this commit](https://github.com/junwoo091400/PX4-Airframe-Metadata-support/commit/75d54c40dd234959ab284604fc6c3942c909ef09) in my Qt UI repository.

This is what will eventually happen to [this](https://github.com/mavlink/qgroundcontrol/pull/10343) too I hope, as having MAVLink defined schema would make it easier for different Autopilots to agree on the format used by the QGC / Mission Planner.

### Frames metadata example
As included as part of the PR, the Frames JSON file would look something like this:

```json
{
    "version": 1,

    "settings": {
        "frame_id_parameter": "SYS_AUTOSTART"
    },

    "frames_v1": [
        {
            "name": "multicopter",
            "type": "group",
            "description": "Rotary vehicles with multiple rotors to lift it's body",
            "image": "Multirotor",
            "subgroups": [
                {
                    "name": "generic multicopter",
                    "type": "frame",
                    "description": "Generic multicopter with customizable number of rotors",
                    "image": "",
                    "frame_id": 1300
                },
                {
                    "name": "quadcopter",
                    "type": "group",
                    "description": "Multicopters with 4 rotors",
                    "image": "Quadrotor",
                    "subgroups": [
                        {
                            "name": "Holybro X500 v2",
                            "type": "frame",
                            "description": "Holybro's X500 V2 frame kit",
                            "manufacturer": "Holybro",
                            "image-custom": "holybro_x500_v2",
                            "url": "http://www.holybro.com/product/x500-v2-kit/",
                            "frame_id": 1313
                        }
                    ]
                },
                {
                    "name": "hexacopter",
                    "type": "group",
                    "description": "Multicopters with 6 rotors",
                    "image": "Hexarotor",
                    "subgroups": [
                        {
                            "name": "Yuneec H520E Hexacopter",
                            "type": "frame",
                            "description": "Yuneec H520E Hexacopter with a camera & landing gear",
                            "manufacturer": "Yuneec",
                            "image-custom": "yuneec_h520e",
                            "url": "https://www.conrad.ch/de/p/yuneec-h520e-speccombo-hexacopter-rtf-profi-kameraflug-mit-waermebild-orange-schwarz-2450997.html",
                            "frame_id": 1254
                        },
                        {
                            "name": "Vector Hexacopter 2.2",
                            "type": "frame",
                            "description": "Yuneec H520E Hexacopter with a camera & landing gear",
                            "manufacturer": "Vision Aerial",
                            "image-custom": "vision_vector_hexacopter",
                            "url": "https://visionaerial.com/product/vector-hexacopter/",
                            "frame_id": 1532
                        }
                    ]
                }
            ]
        }
    ]
}
```

### Final GCS outcome demo

Demo

This is from the [Qt application repository](https://github.com/junwoo091400/PX4-Airframe-Metadata-support#px4-frame-metadata-support-test-repository) linked below.

### Resources
- You can check out the progress / details of this project in the document here: https://docs.google.com/document/d/1wpGDQyJLv3AHAqMH-vxWuZ1HYksX7oq_vldEXCbxmPI/edit#
- This is a standalone Qt application that implements the Visualization / parsing of the Frames metadata, which will get ported over to QGC: https://github.com/junwoo091400/PX4-Airframe-Metadata-support

## Contents
- Added JSON schema and example file for the Frames componenet metadata

## Description
- This adds a new metadata type 'Frames' for supporting dynamically loading the supported frames from the Autopilot to the GCS
- Frame is a generic term for a previously used term `Airframe`, to inclusively bundle ground vehicle / boat / etc, not just aerial vehicles
- The Frame JSON structure is recursive, and any frame can contain many other frames in it's sub-group, which also defines the GCS user selection workflow (Starting from high level, down to low level)
- Schema also needs to be updated whenever there's a new type of a vehicle that we need to add an image for, in order to have a valid `image` Enumerator string tag
- The frames are in a list, to make sure we preserve the order, in order to deliver a consistent GCS usage experience

## TODOs
- [x] Create QGC Implementation for supporting Frames metadata: https://github.com/mavlink/qgroundcontrol/pull/10472

### DIFF
--- a/component_metadata/frames.example.json
+++ b/component_metadata/frames.example.json
@@ -1,0 +1,212 @@
+{
+    "version": 1,
+    
+    "settings": {
+        "frame_id_parameter": "SYS_AUTOSTART"
+    },
+
+    "frames_v1": [
+        {
+            "name": "multicopter",
+            "type": "group",
+            "description": "Rotary vehicles with multiple rotors to lift it's body",
+            "image": "Multirotor",
+            "subgroups": [
+                {
+                    "name": "generic multicopter",
+                    "type": "frame",
+                    "description": "Generic multicopter with customizable number of rotors",
+                    "image": "",
+                    "frame_id": 1300
+                },
+                {
+                    "name": "quadcopter",
+                    "type": "group",
+                    "description": "Multicopters with 4 rotors",
+                    "image": "Quadrotor",
+                    "subgroups": [
+                        {
+                            "name": "Holybro X500 v2",
+                            "type": "frame",
+                            "description": "Holybro's X500 V2 frame kit",
+                            "manufacturer": "Holybro",
+                            "image-custom": "holybro_x500_v2",
+                            "url": "http://www.holybro.com/product/x500-v2-kit/",
+                            "frame_id": 1313
+                        }
+                    ]
+                },
+                {
+                    "name": "hexacopter",
+                    "type": "group",
+                    "description": "Multicopters with 6 rotors",
+                    "image": "Hexarotor",
+                    "subgroups": [
+                        {
+                            "name": "Yuneec H520E Hexacopter",
+                            "type": "frame",
+                            "description": "Yuneec H520E Hexacopter with a camera & landing gear",
+                            "manufacturer": "Yuneec",
+                            "image-custom": "yuneec_h520e",
+                            "url": "https://www.conrad.ch/de/p/yuneec-h520e-speccombo-hexacopter-rtf-profi-kameraflug-mit-waermebild-orange-schwarz-2450997.html",
+                            "frame_id": 1254
+                        },
+                        {
+                            "name": "Vector Hexacopter 2.2",
+                            "type": "frame",
+                            "description": "Yuneec H520E Hexacopter with a camera & landing gear",
+                            "manufacturer": "Vision Aerial",
+                            "image-custom": "vision_vector_hexacopter",
+                            "url": "https://visionaerial.com/product/vector-hexacopter/",
+                            "frame_id": 1532
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "fixed wing",
+            "type": "group",
+            "description": "Aerial vehicles with fixed wing mounted to generate lift",
+            "image": "FixedWing",
+            "subgroups": [
+                {
+                    "name": "generic fixed wing",
+                    "type": "frame",
+                    "description": "Generic fixed wing with customizable set of actuators",
+                    "image": "",
+                    "frame_id": 1592
+                },
+                {
+                    "name": "plane a tail",
+                    "type": "group",
+                    "description": "Fixed wing with 2 ailerons, 2 flaps, a-tail, 2 a-tail actuators, throttle, wheel",
+                    "image": "FixedWingATail",
+                    "subgroups": [
+                        {
+                            "name": "Applied Aeronautics Albatross",
+                            "type": "frame",
+                            "description": "Applied Aeronautics Albatross Fixed Wing platform",
+                            "manufacturer": "Applied Aeronautics",
+                            "image-custom": "applied_aeronautics_albatross",
+                            "url": "https://www.appliedaeronautics.com/albatross-uav",
+                            "frame_id": 1212
+                        }
+                    ]
+                },
+                {
+                    "name": "simple flying wing",
+                    "type": "group",
+                    "description": "Fixed-wing with 2 aileron and 1 throttle",
+                    "image": "FixedWingFlyingWing",
+                    "subgroups": [
+                        {
+                            "name": "Generic flying wing",
+                            "type": "frame",
+                            "description": "Generic flying wing with customizable set of actuators",
+                            "manufacturer": "",
+                                                        "image-custom": "FixedWingFlyingWing",
+                            "url": "",
+                            "frame_id": 1426
+                        },
+                        {
+                            "name": "Wing wing (Z-84)",
+                            "type": "frame",
+                            "description": "Wing wing model",
+                                                        "manufacturer": "E-flite",
+                                                        "image-custom": "eflite_wing_wing",
+                            "url": "https://hobbyking.com/en_us/wing-wing-z-84-epo-845mm-kit.html?___store=en_us",
+                            "frame_id": 1426
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "vtol",
+            "type": "group",
+            "description": "Aerial vehicles with fixed wing and multicopter configuration combined",
+            "image": "VTOL",
+            "subgroups": [
+                {
+                    "name": "tiltrotor vtol",
+                    "type": "group",
+                    "description": "VTOLs with tilting rotors",
+                    "image": "VTOLTiltrotor",
+                    "subgroups": [
+                        {
+                            "name": "Convergence VTOL",
+                            "type": "frame",
+                            "description": "Famous Convergence VTOL!",
+                            "manufacturer": "E-flite",
+                            "image-custom": "convergence-vtol",
+                            "url": "https://www.horizonhobby.com/product/convergence-vtol-bnf-basic-650mm/EFL11050.html",
+                            "frame_id": 6435
+                        }
+                    ]
+                },
+                {
+                    "name": "quad tail-sitter vtol",
+                    "type": "group",
+                    "description": "VTOLs with quadcopter rotor configuration that sits on the tail when launching",
+                    "image": "VTOLQuadrotorTailsitter",
+                    "subgroups": [
+                        {
+                            "name": "Generic Quadrotor X Tailsitter",
+                            "type": "frame",
+                            "description": "Generic Quadrotor tailsitter with rotors positioned in X configuration",
+                            "manufacturer": "",
+                            "image": "VTOLQuadrotorTailsitter",
+                            "url": "",
+                            "frame_id": 5123
+                        }
+                    ]
+                }
+            ]
+        },
+        {
+            "name": "helicopter",
+            "type": "group",
+            "description": "Aerial vehicles with big rotor lifting the body",
+            "image": "Helicopter",
+            "subgroups": [
+                {
+                    "name": "generic helicopter",
+                    "type": "frame",
+                    "description": "Generic helicopter with customizable set of actuators",
+                    "image": "Helicopter",
+                    "frame_id": 4212
+                },
+                {
+                    "name": "coaxial helicopter",
+                    "type": "frame",
+                    "description": "Helicopter with coaxial rotor spinning in different directions",
+                    "image": "HelicopterCoaxial",
+                    "frame_id": 9374
+                }
+            ]
+        },
+        {
+            "name": "simulation",
+            "type": "group",
+            "description": "SIH vehicle frames for running Simulation In Hardware",
+            "image": "Simulation",
+            "subgroups": [
+                {
+                    "name": "sih quadcopter",
+                    "type": "frame",
+                    "description": "SIH quadcopter",
+                    "image": "Quadrotor",
+                    "frame_id": 1523
+                },
+                {
+                    "name": "sih fixed wing",
+                    "type": "frame",
+                    "description": "SIH fixed wing",
+                    "image": "FixedWing",
+                    "frame_id": 1523
+                }
+            ]
+        }
+    ]
+}
--- a/component_metadata/frames.schema.json
+++ b/component_metadata/frames.schema.json
@@ -1,0 +1,115 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "title": "Frames",
+    "description": "List and structure of a Frames supported by the Autopilot",
+    "type": "object",
+
+    "$defs": {
+        "imageTypes": {
+            "description": "Enum representing the image to show for the frame. Convention is to have capital letter on each new phrase",
+            "default": "Undefined",
+            "enum": [
+                "",
+                "Undefined",
+                "Simulation",
+                "Helicopter",
+                    "HelicopterCoaxial",
+                "Multirotor",
+                    "Quadrotor",
+                        "QuadrotorX", "QuadrotorH", "QuadrotorWide", 
+                    "Hexarotor",
+                "FixedWing",
+                    "FixedWingATail", "FixedWingFlyingWing",
+                "VTOL",
+                    "VTOLTiltrotor", "VTOLQuadrotorTailsitter",
+                "Boat",
+                "Rover",
+                "UUV"
+            ]
+        },
+        "frame": {
+            "type": "object",
+            "description": "A group (could be a single frame) of frames",
+
+            "properties": {
+                "type": {
+                    "enum": ["group", "frame"],
+                    "description": "Type of the frame object. If it's a group, it will include other group/frames in `subgroups` property"
+                },
+                "name": {
+                    "type": "string"
+                },
+                "description": {
+                    "type": "string"
+                },
+                "image": {
+                    "$ref": "#/$defs/imageTypes",
+                    "description": "Generic enum value (non GCS specific) representing the frame image. GCS needs to map the value to an appropriate iamge resource. If empty, GCS should use the `image` property in the parent node. If no valid image can be found in parent node, it should show `Undefined` frame image"
+                },
+                "image-custom": {
+                    "type": "string",
+                    "description": "Name of the custom image to load. GCS should have a mechanism to map the string to the appropriate image resource. So compatability with GCS is not guaranteed. It overrides the `image` property if valid"
+                },
+                "manufacturer": {
+                    "type": "string",
+                    "description": "Name of the manufacturer"
+                },
+                "url": {
+                    "type": "string",
+                    "description": "Link to the relevant material online about the frame"
+                },
+                "frame_id": {
+                    "type": "integer",
+                    "description": "ID of the frame"
+                },
+                "subgroups": {
+                    "type": "array",
+                    "description": "Sub-group of other frame / frame-groups",
+                    "items": {
+                        "$ref": "#/$defs/frame"
+                    }
+                }
+            },
+
+            "if": {
+                "properties": { "type": { "const": "group" } }
+            },
+            "then": {
+                "required": ["subgroups"]
+            },
+            "else": {
+                "description": "For frame types, frame_id must be provided",
+                "required": ["frame_id"]
+            },
+
+            "required": ["type", "name"],
+            "additionalProperties": false
+        }
+    },
+
+    "properties": {
+        "version": {
+            "type": "integer",
+            "description": "Version of the Frames JSON Schema"
+        },
+
+        "settings": {
+            "type": "object",
+            "description": "Common settings",
+            "properties": {
+                "frame_id_parameter": {
+                    "description": "Parameter name matching the Frame ID",
+                    "type": "string"
+                }
+            }
+        },
+
+        "frames_v1": {
+            "type": "array",
+            "description": "List of frames as defined in v1 spec",
+            "items": {
+                "$ref": "#/$defs/frame"
+            }
+        }
+    }
+}
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -852,6 +852,9 @@
       <entry value="5" name="COMP_METADATA_TYPE_ACTUATORS">
         <description>Meta data for actuator configuration (motors, servos and vehicle geometry) and testing.</description>
       </entry>
+      <entry value="6" name="COMP_METADATA_TYPE_FRAMES">
+        <description>Meta data for frames configuration (definition of supported frame types) and testing.</description>
+      </entry>
     </enum>
     <enum name="ACTUATOR_CONFIGURATION">
       <description>Actuator configuration, used to change a setting on an actuator. Component information metadata can be used to know which outputs support which commands.</description>


component_metadata/frames.example.json
```diff
@@ -0,0 +1,212 @@
+{
+    "version": 1,
+    
+    "settings": {
+        "frame_id_parameter": "SYS_AUTOSTART"
+    },
+
+    "frames_v1": [
```
hamishwillee, line 8: do you need `_v1` if the version is flagged above?

component_metadata/frames.example.json
```diff
@@ -0,0 +1,212 @@
...
+        {
+            "name": "multicopter",
+            "type": "group",
+            "description": "Rotary vehicles with multiple rotors to lift it's body",
+            "image": "Multirotor",
+            "subgroups": [
```
hamishwillee, line 14:
```suggestion
            "frames": [
```

component_metadata/frames.example.json
```diff
@@ -0,0 +1,212 @@
...
+                    "type": "frame",
+                    "description": "Generic multicopter with customizable number of rotors",
+                    "image": "",
+                    "frame_id": 1300
+                },
+                {
+                    "name": "quadcopter",
+                    "type": "group",
+                    "description": "Multicopters with 4 rotors",
+                    "image": "Quadrotor",
```
hamishwillee, line 26: Where is this image coming from? Should it/could it be a URL?

component_metadata/frames.example.json
```diff
@@ -0,0 +1,212 @@
...
+            "subgroups": [
+                {
+                    "name": "generic multicopter",
+                    "type": "frame",
+                    "description": "Generic multicopter with customizable number of rotors",
+                    "image": "",
+                    "frame_id": 1300
+                },
+                {
+                    "name": "quadcopter",
```
hamishwillee, line 23: There are lots of geometries of quadcopter. Is there only one level of indirection? i.e. would you allow quadcopter to contain frame quad + and quadX or would those be separate frames.

component_metadata/frames.example.json
```diff
@@ -0,0 +1,212 @@
...
+                    "image": "Quadrotor",
+                    "subgroups": [
+                        {
+                            "name": "Holybro X500 v2",
+                            "type": "frame",
+                            "description": "Holybro's X500 V2 frame kit",
+                            "manufacturer": "Holybro",
+                            "image-custom": "holybro_x500_v2",
+                            "url": "http://www.holybro.com/product/x500-v2-kit/",
+                            "frame_id": 1313
```
hamishwillee, line 35: On PX4 the frame_id maps to a particular airframe file, which is essentially a shell script for loading parameters (and perhaps starting a few applications). This is PX4 specific. We need to make sure the approach is generic.

hamishwillee:
@junwoo091400 Cool! I very much like the idea of a standardized way of updating a GCS about available frames. Note however  that we need to make sure that this works for other flight stacks and Ground stations that might choose to use different presentation, or need other metadata.

Have you looked at metadata QGC gets from ArduPilot? Also what Mission Planner displays? Can this information map it?
The current structure looks like this from what I can tell:
```
group (multicopter) - which has list of frames:
   frame 1
   frame 2.
```
That implies a flat structure. I.e you can have all multicopters grouped under one icon (say) or you can split out the groups so you have "quacopters", "hexacopter", "dodecathingycopters". 
But you can't have
```
multicopters
   quadcopters
      X
      +
    hexacopters
boats
```
I think you probably want to allow deeper nesting and allow the GCS to compose the UI as it wishes.    


@bkueng Can you sanity check this from a "metadata design" point of view. For example, will this worth with cases like:
- translations
- dynamic addition of frames (?).
- Is any more data needed by PX4 for the GCS.
- Is there any other data that PX4 could/should provide that might be useful?
- All the other things I haven't thought of :-)

hamishwillee:
@meee1 @WickedShell  FYI - we will need your feedback on what metadata would be needed for airframe information in ArduPilot. Whether or not ArduPilot chooses to use this mechanism it is important design consideration that you would be able to if you wanted.

component_metadata/frames.example.json
junwoo091400: As defined in the schema, the images need to have a mapping mechanism to find an appropriate image corresponding to the image Enum / or a custom image.

https://github.com/mavlink/mavlink/blob/48d3de374e63130467012d9be5955843c530371c/component_metadata/frames.schema.json#L45-L52

In the Qt project for Frames selection GUI, this was done via searching for an image asset that has the same string (e.g. `Quadrotor`) as the file name (case-insensitive).

https://github.com/junwoo091400/PX4-Airframe-Metadata-support/blob/75d54c40dd234959ab284604fc6c3942c909ef09/Frames_UI/frames.cpp#L114-L130

component_metadata/frames.example.json
junwoo091400: We can add Quad+, QuadX, and so on under the quadcopter frame of course!

I didn't do that yet (not all the frames were ported to the image Enums definition yet:

https://github.com/mavlink/mavlink/blob/48d3de374e63130467012d9be5955843c530371c/component_metadata/frames.schema.json#L8-L27

component_metadata/frames.example.json
junwoo091400: Currently, this PX4-specific relationship between parameter SYS_AUTOSTART and the frame ID is captured in this settings string:

https://github.com/mavlink/mavlink/blob/48d3de374e63130467012d9be5955843c530371c/component_metadata/frames.schema.json#L100-L103

However, yes I also agree that not all Autopilots would like to use the 'frame ID' integer as an identifier.

I wonder how we can make it truly generic (e.g. Enable having identifier string, instead of an integer, or even an enum, etc)

Thoughts? @bkueng

junwoo091400:
> Have you looked at metadata QGC gets from ArduPilot? Also what Mission Planner displays? Can this information map it?

Nope. But I was aware that the Actuators metadata PR only added support for PX4, and if not supported (e.g. Ardupilot, as of now), it can simply fallback to the previous implementation (For this Frames case, using the cached Airframes.xml file).

Therefore I didn't check exactly what metadata gets fetched, etc. But I will check it out!

Actuators QGC PR for reference: https://github.com/mavlink/qgroundcontrol/pull/9952

> I think you probably want to allow deeper nesting and allow the GCS to compose the UI as it wishes.

This is totally supported by the current JSON schema! For example, notice how the `Wing Wing` frame is nested under: "Fixed Wing > Simple Flying Wing".

We can have any arbitrary number of frames sub-nested, and any depth we want 😉 

https://github.com/mavlink/mavlink/blob/48d3de374e63130467012d9be5955843c530371c/component_metadata/frames.example.json#L112-L120

junwoo091400:
> I talked to the main ArduPilot GCS vendors and @auturgy . They don't see any value in this, I think primarily because none of them need to support multiple flight stacks. They also made the point that airframes do not change as much as parameters do or particularly get out of sync - so not worth carrying along with the vehicle (flash space).

I just checked how Ardupilot handles different frames parameter sets.

https://github.com/ArduPilot/ardupilot/tree/master/Tools/Frame_params

Also referring to how the Parrot Disco frame's startup script index & parameter gets configured:

https://ardupilot.org/plane/docs/airframe-disco.html#installation

It seems like these files are not included in the flash from the beginning & the concept of 'airframe selection' doesn't exist, as it does for PX4.

So I get why this definitely wouldn't make sense.

And I guess that's what you mean by "Ardupilot doesn't support different flight stacks", as a single Ardupilot firmware will only support a single vehicle type & a single airframe (unless modified manually by the user).

Am I correct about this? @hamishwillee

junwoo091400:
Note: I started a conversation in Ardupilot discord & during EU Dev call today on this matter.

And the conclusion seems to be that Ardupilot wouldn't benefit from this feature (Frames metadata). Reason being:
1. Ardupilot doesn't really utilize 'choosing' the airframe via the parameter like PX4 does, but rather often bakes in a single frame / parameter set to a single firmware, and this is how a lot of vendors are using it for their product too.
2. There are a lot of custom airframes grouped under the Quadrotor Vtol frame class, and there's a question on how to show an image for those 'weird' configurations
3. TBD

Find the discord message here: https://discordapp.com/channels/674039678562861068/728017546313466047/1037273772408971294

@bkueng I think we would need to find a different way to make this work. Any thoughts on this? (Note: Check the discord as well, there's a nice discussion!)

auturgy:
Baking a json describing every possible frame type into the firmware doesn't make sense.
I suggest revisiting the problem definition, so that it is clear what the issues are, then look at the available options to solve them.
Suggest closing this PR.

junwoo091400:
> Baking a json describing every possible frame type into the firmware doesn't make sense. I suggest revisiting the problem definition, so that it is clear what the issues are, then look at the available options to solve them. Suggest closing this PR.

Maybe it doesn't make sense for Ardupilot but for PX4 this is a viable option. Maybe we can have a reserved component metadata types for PX4 specific? So that Ardupilot won't be affected?

I believe your conclusion is coming from the discrepancy on how Ardupilot considers an Airframe vs how PX4 considers it. I will try out Ardupilot to check how this fits in the view & to understand your questions better :wink:

auturgy:
I really don’t think you’ve thought about the problem enough: you’ve simply leaped to a solution, put a significant amount of effort into implementing it, and are now defending it. I get it.
If PX4 want this, great. Close this PR, fork MAVLink, then do whatever you like.

junwoo091400:
I found that Ardupilot does have a manual mapping of the Frames in the python script edited as well:

https://github.com/ArduPilot/ardupilot/blob/58cdf8154adbb0390abee310a4e3878e5ab9853d/Tools/autotest/pysim/vehicleinfo.py#L10-L187

junwoo091400:
> I really don’t think you’ve thought about the problem enough: you’ve simply leaped to a solution, put a significant amount of effort into implementing it, and are now defending it. I get it.

The only reason I am defending it is because you are not giving concrete / exact reason why you think this is flawed. You just said "it doesn't make sense", but why exactly? It is hard to read your intentions if you don't state them verbally :smile: 

Also, that's why i suggested creating a reserved component metadata type for PX4, for that we don't need to fork MAVLink.

I have no intention of making this so complicated / opinionated, I just need a good reasons / valid criticisms, which I'm sure you can do but for me your comments are too implicit and not verbose enough.

By the way, do you have experience in PX4? Maybe you may also be lacking understanding on how the Airframes are dealt with in PX4. Let's try to understand each other and figure out a viable solution :wink:

junwoo091400:
## Trying out ArduPilot
To understand Ardupilot's perspective on the Frame selection, I have been experimenting with ArduPilot SITL.

image

Findings:
1. The frames images changes suddenly when the 'Frame Type' (e.g. X, V, Plus) is changed, even within the same Frame class (e.g. Quad / Hexa). This can definitely be improved by having a better UI
2. A lot of the images are UFO (Undefined Airframes). We can fix that.

However also the question is:
"How many Ardupilot users use QGC?".

I will be trying out Mission Planner & MAVProxy / APM Planner in the meantime :+1:

hamishwillee:
@junwoo091400 FYI @auturgy probably has a better understanding of more of PX4 than I do. 

Component metadata is a useful way of delivering firmware-specific information that allows you to be sure that the information is always synchronised with what the firmware can do. It has the benefit that you don't have to keep a GCS up to date with changes to that metadata, and in particular that multiple GCS which you have no control over can get the information even if they don't know the capabilities of your particular drone.
It has a cost in that either you have to carry data in firmware OR have it hosted somewhere and hence require a network connection to get the data (at least once, after which it is cached).

Unfortunately while I understand that this makes no sense for ArduPilot for airframe data, I don't understand the objection to it as a delivery mechanism for other  flight stacks if they want to pay the cost. That said, not do I see it as _necessary_ to deliver the data this way for PX4. The data updates so rarely. 

I'm not sure where this leaves us.

hamishwillee:
PS I still like the idea of being able to create a "standard" or generic UI such that any GCS can say "if I get data in this format I can display an airframe selection UI" (irrespective of how it is delivered). 

You _might_ find some traction for such a UI even if it isn't delivered by component metadata (at least by ArduPilot) - provided it makes updating and integrating with QGC easier for them. That would require decoupling the startup mechanism from having a hard coded id for the frame, which the ground station needs to understand represents a parameter file.

junwoo091400:
# Discussion with James
So I had a lengthy discussion with @auturgy this morning, and here are some things we discussed.

## Problem definition
1. **QGC sometimes displays airframes that are not supported**, as it has no clue which airframes are supported in the board that is connected (It uses cached Airframes XML file, read more about it [here](https://beginnerjunwoo.wordpress.com/2020/08/23/px4-deepdive-2-all-about-px4s-airframes-xml-parsing-and-where-its-used-qgc-px4/)). So the user may select a particular airframe and reboot, but the airframe isn't selected, this is confusing and undesirable.
2. **QGC Airframe selection UI is scattered**: As described well in the [airframe UI improvement doc](https://docs.google.com/document/d/1wpGDQyJLv3AHAqMH-vxWuZ1HYksX7oq_vldEXCbxmPI/edit?pli=1#heading=h.5vutt232yxo3), current selection UI is quite scattered & can beneift from having a cascaded frame selection workflow

## Solutions
1. **QGC displaying unsupported airframes**
   - This is a PX4 Problem.
      - It is already fixable via having a mechanism in GCS to check which airframe IDs are supported in the PX4 connected.
      - 'Ideal' way of doing that would be via having component metadata as this PR suggested, but pitfalls will be discussed below.
      - One way is having some sort of mechanism in PX4 to detect if the autostart script with the specific airframe ID as a prefix
      - Or, we can show an error message if an unsupported airframe is selected, as the airframe ID will be reset, GCS can simply display "Warning! The airframe is not set up properly. This may be due to you choosing an unsupported airframe."
   - ArduPilot: I suspect it's also an issue there as well, but apparently doesn't happen as often, as ArduPilot users usually build a custom firmware that supports a specific vehicle, so quite often airframe selection is not even necessary
2. **QGC Airframe UI problem**
   - The UI is not ideal, and that's what everyone can agree on. Improving the UI itself isn't an issue.
   - Especially, as ArduPilot shows all the possible frame class configurations in their UI whenever a user selects a specific type (e.g. X, +, Y, etc), this can definitely be improved.

## Need for a frame metadata
- PX4
   - Since a single PX4 binary *usually supports all the airframes (except for some flash-constrained boards like FMUv2, where some airframes are omitted), there are a lot of airframes (Rover / Plane / Multirotor / etc)
   - Since it doesn't strictly classify airframes into certain class/type enum combinations, but rather assigns a specific number (and a certain type of aircraft can have multiple airframe files: Holybro X500 V2, Quadcopter Q250, Mantis Edu, etc), there's a lot of airframes (again!)
   - **So PX4 needs a more elegant way to control/display all these various types of supported airframes**
- ArduPilot
   - The build binaries are separated into different vehicle designs (Rover / Copter / Plane / etc), from the beginning
   - On a single stable release for the specific vehicle design (e.g. Copter), there is only a confined set of Frame [Class](https://ardupilot.org/plane/docs/parameters.html#q-frame-class)(e.g. Quad / Hexa / etc) & [Type](https://ardupilot.org/plane/docs/parameters.html#q-frame-type-frame-type-x-or-v) (+ / x / y / Y6F / etc) that a user can select
   - Therefore, there is no concept of a specific 'airframe' (e.g. Holybro X500 V2). Everything is a combination of: "Vehicle design", "Frame Class" and "Frame Enum", which is a confined & pre-deterministic set.
   - **So ArduPilot doesn't need such a flexible way of defining frame variations, as they are confined & the notion of custom airframes doesn't exist in regular build**
   - When there's a need for a custom airframe, [users build a custom binary](https://custom.ardupilot.org/) with that airframe's configuration only, so there is even no need to select a Frame, as users know exactly which frame they will fly in, and it gets hard-coded into the firmware.

## Pros and Cons of Frame metadata
- Pros
   - Would improve the Frame selection UI workflow
   - Would always have the supported frames in sync between the GCS & Autopilot
   - Unified way of defining the dynamically generated UI (like Actuators), which reduces maintenance effort/customization needed in GCS
- Cons
   - We probably only use this once in an Autopilot's lifetime, so it should not consume too much memory/resources (Usually people don't have to select a different airframe)
      - Considering this, we probably need to have the Frames metadata in the cloud as well. But then, how do we know which metadata version corresponds to which firmware?
   - We need to manage all the different Vehicle 'types', like done [here](https://github.com/junwoo091400/PX4-Airframe-Metadata-support/blob/75d54c40dd234959ab284604fc6c3942c909ef09/Frames_UI/frames.schema.json#L8-L28). How can we properly name all these different frame types? Who maintains them?
   - How can we ensure that we have the correct frame metadata? What happens when a user has a custom build that omits support for certain vehicle types?

## Questions
1. PX4 Airframes: @bkueng, regarding the Pros/Cons - Cons point 3, right now we don't really regulate the Airframe inclusion based on what the target actually supports (e.g. If we build a build variant only for a rover, [like on FMUv2](https://github.com/PX4/PX4-Autopilot/blob/main/boards/px4/fmu-v2/rover.px4board), we probably don't filter airframes based on which frame type it is).
   - **How would we reliably determine which frames are *actually supported?**
2. Frame type maintenance: @auturgy, this Frames schema doesn't confine you in terms of cascaded structure. So having a custom Quadplane with variable geometry, etc can definitely be placed under the Quadplane category for example.
   - Actually not having a *standardized way of calling different vehicle types is not a problem. That's the point of the JSON schema, you can define whatever kind of frame you want. You aren't forced to update the schema every time there is a new airframe (although that would be ideal)
   - Where the problem arises (and the part you were worried about, I think) is determining which image to show for the frame. This, unfortunately, have to be standardized to make it GCS agnostic (We want to give a GCS a glimpse on what kinds of frames are expected, to have the images).
   - This was done [here](https://github.com/junwoo091400/PX4-Airframe-Metadata-support/blob/75d54c40dd234959ab284604fc6c3942c909ef09/Frames_UI/frames.schema.json#L8-L28) for example.
   - But, as I said, this isn't required. The GCS will automatically fallback into the parent frame's image (which should give a clue on what the frame is like, although not exactly true)
   - **Was your concern about the image classifications as I described here?**
3. PX4 Custom metadata: @hamishwillee, @auturgy said that having a custom Enum range to handle PX4-specific metadata types would be a bad choice in the long run, and I agree, as MAVLink shouldn't be tied to a specific flightstack.
   - You indicated above that having it as metadata that can be used by one flight stack, but not so useful for the other would be ok. Has there been a case like this in MAVLink's past? Just curious.

## Moving Forward
Here's what I propose we should do next:
1. Agree on whether Frames metadata is something we could support (by addressing 2 Questions above)
2. If we can support this, agree on the structure/definition of the schema
3. Improve both Ardupilot / PX4's Frame selection UI 🤔 

### Future of Airframes
On a side note, as discussed in the [PX4 dev call yesterday](https://discuss.px4.io/t/px4-dev-call-november-02-2022/29527#p1-frames-airframe-component-metadata-support-junwoo0914-9), it does seem like the PX4's concept of Airframes as a script with a prefix of a frame ID & specific set of parameters, can indeed be consumed by the Control allocation (Actuators metadata), and the actual configuration of the specific param values, etc should rather be kept in the GCS side, and not the Firmware side. So in theory GCS can just detect which types of vehicles are supported & can show some custom presets that a user can select (e.g. Holybro X500 V2). FYI @dagar

auturgy:
Remove that comment and run it through me for correction.  It does not reflect our discussion.

junwoo091400:
> Remove that comment and run it through me for correction

Why not post your response here? Why can't you formulate a valid reasoning?



> It does not reflect our discussion.

Well I think it does, and the whole point of me writing this down is to let others know of what we discussed (at least from my point of view), and if you think it is inaccurate, you should say exactly why, instead of just leaving a comment like this (which you have done so couple of times, in discord and in previous comments) and closing the issue.

@auturgy, Please comprehend that not everyone thinks the same way as you do, and that you can't control everything in a way that you want it to be. If you don't like something, explain why 😁, instead of shutting off the conversation and refusing to leave your reasoning here.

Please re-open the issue 😊

junwoo091400:
> Remove that comment and run it through me for correction. It does not reflect our discussion.

Ok I was being too dramatic in the previous comment, I apologize. I have said we should keep things civil, but found myself worsening the tension. I shouldn't have done that to you, and I am sorry.

What I really want to ask is, what is not accurate in my comment regarding what we have discussed? Please let me know!

junwoo091400:
Note: one of the reason / use-case is described in this Issue: https://github.com/mavlink/qgroundcontrol/issues/10204

hamishwillee:
@junwoo091400 We discussed this in the MAVLink call last night. The use case currently makes no sense "as part of the standard" because it can't be generic/standard.  

Have a chat to @bkueng for more detail. 
This might be something that PX4 chooses to pursue as a dialect. If there are further steps to take, they should be added to a new PR  - for example, if the decision was made to do a PX4-specific solution. At that point the mavlink engagement would be around defining a mechanism for dialect-specific metadata.

Issue #1669, opened 2021-08-12
Common.xml - WIP message review


hamishwillee:
The following messages, commands, enums, enum values have been tagged as "Work in Progress" in common.xml/minimal.xml. WIP tagging is not intended to last forever, and under the current governance model messages that are "work in progress" should not be included in common.xml at all (they should go into development.xml until a prototype has been created, and they are approved for inclusion).

The proposal is to review all of these and either remove WIP or determine a timeframe for removing WIP. If no implementation exists or is being developed imminently, then these might alternatively be moved to development.xml.

----

Mixed bag propose go in as I believe their are implementations:

- [TUNNEL](https://mavlink.io/en/messages/common.html#TUNNEL)
- [ONBOARD_COMPUTER_STATUS](https://mavlink.io/en/messages/common.html#ONBOARD_COMPUTER_STATUS)
- [COMPONENT_INFORMATION](https://mavlink.io/en/messages/common.html#COMPONENT_INFORMATION)
- [SUPPORTED_TUNES](https://mavlink.io/en/messages/common.html#SUPPORTED_TUNES)
- [WINCH_STATUS](https://mavlink.io/en/messages/common.html#WINCH_STATUS) - I am 99% sure this is live in ArduPilot
- [MISSION_CHANGED](https://mavlink.io/en/messages/common.html#MISSION_CHANGED)
- [RAW_RPM](https://mavlink.io/en/messages/common.html#RAW_RPM)
- [MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN](https://mavlink.io/en/messages/common.html#MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN) - params 3, 4, 7 - camera, gimbal, camera ID shutdown.
- [MAV_CMD_NAV_ROI](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_ROI) - added 2018
- [MAV_CMD_DO_UPGRADE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_UPGRADE)
- [MAV_CMD_OBLIQUE_SURVEY](https://mavlink.io/en/messages/common.html#MAV_CMD_OBLIQUE_SURVEY)
- [MAV_CMD_INJECT_FAILURE](https://mavlink.io/en/messages/common.html#MAV_CMD_INJECT_FAILURE)
- [PROTOCOL_VERSION](https://mavlink.io/en/messages/common.html#PROTOCOL_VERSION)

[...]

PR #1909, opened 2022-11-02
Camera Protocol v2 - Remove WIP tracking

hamishwillee:
This removes wip tagging from: MAV_CMD_CAMERA_TRACK_POINT, MAV_CMD_CAMERA_TRACK_RECTANGLE, MAV_CMD_CAMERA_STOP_TRACKING, CAMERA_FOV_STATUS, CAMERA_TRACKING_IMAGE_STATUS, CAMERA_TRACKING_GEO_STATUS

This is (I understand) in a number of mavlink camera vendors and has also been integrated into QGC and MAVSDK. I am not sure about PX4 "integration". I do know that both ArduPilot and PX4 allow passthrough of these messages, so will work with cameras supporting the API.

Follows up plan to remove WIP where possible in #1669

FYI @auturgy @julianoes 

- [ ] Once WIP moved, update camera docs with these

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2103,16 +2103,12 @@
         <param index="3" label="Pause" minValue="-1" maxValue="1" increment="2">1 to pause triggering, but without switching the camera off or retracting it. -1 to ignore</param>
       </entry>
       <entry value="2004" name="MAV_CMD_CAMERA_TRACK_POINT" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>If the camera supports point visual tracking (CAMERA_CAP_FLAGS_HAS_TRACKING_POINT is set), this command allows to initiate the tracking.</description>
         <param index="1" label="Point x" minValue="0" maxValue="1">Point to track x value (normalized 0..1, 0 is left, 1 is right).</param>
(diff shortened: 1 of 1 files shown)


auturgy:
Can we reference the implementations (preferably get feedback so we know whether this should actually be considered stable or if there’s rub points that need addressing).

julianoes:
I'll forward that ask for references to @ThomasDebrunner.

hamishwillee:
Thanks - good point @auturgy  - I had assumed these were accidental omissions from the docs, not late additions that might not actually be implemented.

1. My assumption is that since the status might not be sent and the mav commands would be rejected  if not supported these can be considered as part of camera protocol v2 rather than forcing the existence of camera protocol v3?
2. I'll also ping Thomas offline.

As @julianoes pointed out there is a bit of a chicken and egg situation for this API, since the implementation is usually outside of the flight stack code. I "think" a good way to address this is to complete the MAVSDK camera server implementation and provide a USB camera library backend. That way at least all the interactions can be verified with some real code on a platform that is supported.

hamishwillee:
@auturgy These are implemented and deployed in the (closed source) Auterion camera server for multiple cameras. While I'd be happier with an open source implementation such as MAVSDK camera server, this has been validated as fit for purpose by real users - which was your main concern.

Note, @damurashov is also interested in providing an implementation, but was concerned about these being WIP - I don't have more detail on that.

OK to merge?

auturgy:
If they're fit for purpose, no issue.

hamishwillee:
OK, so they are deployed, and I understand "fit for purpose". @julianoes do you have any reservations about these? I was about to merge then saw no one has approved it.

julianoes (approved):

PR #1910, opened 2022-11-07
DO_LAND_START to consider altitude (3D distance)

magicrub:
Change DO_LAND_START from 2D distance (lat, lng only) to 3D (also consider altitude).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1473,7 +1473,7 @@
         <param index="4">Empty</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
-        <param index="7">Empty</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="190" name="MAV_CMD_DO_RALLY_LAND" hasLocation="false" isDestination="false">
         <description>Mission command to perform a landing from a rally point.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1466,7 +1466,10 @@
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
       <entry value="189" name="MAV_CMD_DO_LAND_START" hasLocation="true" isDestination="false">
-        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts. It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used. The Latitude/Longitude is optional, and may be set to 0 if not needed. If specified then it will be used to help find the closest landing sequence.</description>
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts.
+	  It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used.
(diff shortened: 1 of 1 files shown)


magicrub:
@hamishwillee do you have any ideas about this CI failure? It's not failing on the ardupilot PR https://github.com/ArduPilot/mavlink/pull/293

magicrub:
> 1. How will the altitude modify the above description
> 2. What's the value? You have two landing sequences at same location but different altitude and want the system to select the best one?

As described in my implementation PR https://github.com/ArduPilot/ardupilot/pull/22147 
```This is useful for having two DO_LAND_START's near each other for things like main go-home at cruise-ish altitude on RTL event and and a low-alt one for more tailored aborted landings```

**Comment by magicrub** [comment]
@hamishwillee what changes to the description do you suggest?

File: message_definitions/v1.0/common.xml [hunk 1]
```diff
@@ -1466,14 +1466,17 @@
...
       </entry>
       <entry value="189" name="MAV_CMD_DO_LAND_START" hasLocation="true" isDestination="false">
-        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts. It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used. The Latitude/Longitude is optional, and may be set to 0 if not needed. If specified then it will be used to help find the closest landing sequence.</description>
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts.
+	  It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used.
+	  The Latitude/Longitude/Altitude is optional, and may be set to 0 if not needed. If specified then it will be used to help find the closest landing sequence.
```
hamishwillee, line 1471: The only addition here is "/Altitude".

hamishwillee:
@sfuhrer This change adds altitude to the DO_START_LAND in addition to lat/lon. This information, if specified, is used to select the closest mission landing to use. Having altitude as well allows selection between low and high altitude landing patterns as per Tom's comment above. Any objections?

Thanks for clarifying @magicrub - this makes sense to me. We can merge if Silvan is also OK with it.

hamishwillee (approved):

sfuhrer (approved):
Makes sense for me to - the only commend I'd have is to make it more explicit that Altitude is meant to be in AMSL (I guess).

magicrub:
@sfuhrer you know, I didn't even consider the altitude frame because in Ardupilot this is only used as an AUTO mission item and in that scope ArduPilot contains the knowledge of the altitude frame type for each individual mission item. Specifiying it here in the doc would conflict with how a mission would be parsed. Sooooo I looked how [MAV_CMD_NAV_WAYPOINT](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L1039-L1047) does it and it also doesn't mention the altitude frame.

message_definitions/v1.0/common.xml
magicrub: approved!

hamishwillee:
@magicrub @sfuhrer 
1. Frames are in theory defined in the [MISSION_ITEM_INT.frame](https://mavlink.io/en/messages/common.html#MISSION_ITEM_INT) or in the COMMAND_INT, so not specifying AMSL here is OK. 
   - FWIW I say in theory because all the assumptions around mission design are that it is a global frame MAV_FRAME_GLOBAL
2. I don't think it matters because all of the items in the plan are likely to be the same frame relative to each other.

Could be wrong. @sfuhrer So you're otherwise OK with this?

hamishwillee:
Actually, I'm merging. I'm fairly confident of my comment above, and if I am wrong, it's an easy addition.

PR #1911, opened 2022-11-08
Gimbal Device: Some spelling

olliw42:
This is the first of a sequence of four PRs, which are intended to finally get the looming changes with regards to the gimbal device protocol merged. These are of two kinds: handling of yaw absolute and relative, and handling of RC signals.

The four PRs are:
1. https://github.com/mavlink/mavlink/pull/1911
2. https://github.com/mavlink/mavlink/pull/1912
3. https://github.com/mavlink/mavlink/pull/1914
4. https://github.com/mavlink/mavlink/pull/1913

The additions to the gimbal device flags and messages have been discussed and agreed to before in
- https://github.com/mavlink/mavlink/issues/1860 "Yet another effort towards unified gimbal device messages"
- https://github.com/mavlink/mavlink/pull/1885 "gimbal: more flexible attitude messages"

see especially https://github.com/mavlink/mavlink/issues/1860#issuecomment-1189939693, https://github.com/mavlink/mavlink/issues/1860#issuecomment-1204714314

These PRs are intended to weed out the little kinks which were remaining.

The four PRs can technically each be merged for itself, but depend on each other, i.e., should not merged at all but in the proper sequence, and they will require rebasing to do so. However, I have done them spearately so you can better see what changes are there. Hopefully that's an appropriate approach.

### This PR: Some spelling

Some little mistakes have been corrected, and the language has been prepared to the upcoming language concerning earth frame and vehicle frame. Except of taste the changes should not be controversial.

This PR should be good as is and could be merged. Some of the later PRs will override soem text, but that's in the nature of it.

Olli :)

PR #1912, opened 2022-11-08
Gimbal Device: Adding more invalid and bitmask attributes

olliw42:
This is the second of a sequence of four PRs, which are intended to finally get the looming changes with regards to the gimbal device protocol merged. These are of two kinds: handling of yaw absolute and relative, and handling of RC signals.

The four PRs are:
1. https://github.com/mavlink/mavlink/pull/1911
2. https://github.com/mavlink/mavlink/pull/1912
3. https://github.com/mavlink/mavlink/pull/1914
4. https://github.com/mavlink/mavlink/pull/1913

The additions to the gimbal device flags and messages have been discussed and agreed to before in
- https://github.com/mavlink/mavlink/issues/1860 "Yet another effort towards unified gimbal device messages"
- https://github.com/mavlink/mavlink/pull/1885 "gimbal: more flexible attitude messages"

see especially https://github.com/mavlink/mavlink/issues/1860#issuecomment-1189939693, https://github.com/mavlink/mavlink/issues/1860#issuecomment-1204714314

These PRs are intended to weed out the little kinks which were remaining.

The four PRs can technically each be merged for itself, but depend on each other, i.e., should not merged at all but in the proper sequence, and they will require rebasing to do so. However, I have done them spearately so you can better see what changes are there. Hopefully that's an appropriate approach.

### This PR: Adding more invalid and bitmask attributes

Invalid attributes are added to
- the GIMBAL_DEVICE_INFORMATION's min max fields. This allows to account for situations where a gimbal device doesn't provide such a feature (STorM32, any other "cheap" gimbal)
- the GIMBAL_DEVICE_SET_ATTITUDE's flags field: that's in fact an important one: Situations can occur where one does NOT want to set the flags but only some of the other fields. This makes this possible.
- the AUTOPILOT_STATE_FOR_GIMBAL_DEVICE's fields: some of the info may not be available, this is to indicate it

Some few bitmasks are added

Olli :)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6927,12 +6927,12 @@
       <field type="uint64_t" name="uid" invalid="0">UID of gimbal hardware (0 if unknown).</field>
       <field type="uint16_t" name="cap_flags" enum="GIMBAL_DEVICE_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
       <field type="uint16_t" name="custom_cap_flags" display="bitmask">Bitmap for use for gimbal-specific capability flags.</field>
-      <field type="float" name="roll_min" units="rad">Minimum hardware roll angle (positive: rolling to the right, negative: rolling to the left)</field>
-      <field type="float" name="roll_max" units="rad">Maximum hardware roll angle (positive: rolling to the right, negative: rolling to the left)</field>
-      <field type="float" name="pitch_min" units="rad">Minimum hardware pitch angle (positive: up, negative: down)</field>
-      <field type="float" name="pitch_max" units="rad">Maximum hardware pitch angle (positive: up, negative: down)</field>
-      <field type="float" name="yaw_min" units="rad">Minimum hardware yaw angle (positive: to the right, negative: to the left)</field>
-      <field type="float" name="yaw_max" units="rad">Maximum hardware yaw angle (positive: to the right, negative: to the left)</field>
+      <field type="float" name="roll_min" units="rad" invalid="NaN">Minimum hardware roll angle (positive: rolling to the right, negative: rolling to the left). NAN if unknown.</field>
+      <field type="float" name="roll_max" units="rad" invalid="NaN">Maximum hardware roll angle (positive: rolling to the right, negative: rolling to the left). NAN if unknown.</field>
+      <field type="float" name="pitch_min" units="rad" invalid="NaN">Minimum hardware pitch angle (positive: up, negative: down). NAN if unknown.</field>
+      <field type="float" name="pitch_max" units="rad" invalid="NaN">Maximum hardware pitch angle (positive: up, negative: down). NAN if unknown.</field>
+      <field type="float" name="yaw_min" units="rad" invalid="NaN">Minimum hardware yaw angle (positive: to the right, negative: to the left). NAN if unknown.</field>
+      <field type="float" name="yaw_max" units="rad" invalid="NaN">Maximum hardware yaw angle (positive: to the right, negative: to the left). NAN if unknown.</field>
     </message>
     <message id="284" name="GIMBAL_DEVICE_SET_ATTITUDE">
       <wip/>
@@ -6978,12 +6978,12 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS">Current gimbal flags set.</field>
+      <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS" display="bitmask">Current gimbal flags set.</field>
       <field type="float[4]" name="q">Quaternion components, w, x, y, z (1 0 0 0 is the null-rotation). The frame is described in the message description.</field>
       <field type="float" name="angular_velocity_x" units="rad/s" invalid="NaN">X component of angular velocity (positive: rolling to the right). The frame is described in the message description. NaN if unknown.</field>
       <field type="float" name="angular_velocity_y" units="rad/s" invalid="NaN">Y component of angular velocity (positive: pitching up). The frame is described in the message description. NaN if unknown.</field>
       <field type="float" name="angular_velocity_z" units="rad/s" invalid="NaN">Z component of angular velocity (positive: yawing to the right). The frame is described in the message description. NaN if unknown.</field>
-      <field type="uint32_t" name="failure_flags" display="bitmask" enum="GIMBAL_DEVICE_ERROR_FLAGS">Failure flags (0 for no failure)</field>
+      <field type="uint32_t" name="failure_flags" enum="GIMBAL_DEVICE_ERROR_FLAGS" display="bitmask">Failure flags (0 for no failure)</field>
       <extensions/>
       <field type="float" name="delta_yaw" units="rad" invalid="NAN">Yaw angle relating the quaternions in earth and body frames (see message description). NaN if unknown.</field>
       <field type="float" name="delta_yaw_velocity" units="rad/s" invalid="NAN">Yaw angular velocity relating the angular velocities in earth and body frames (see message description). NaN if unknown.</field>
@@ -6996,11 +6996,11 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint64_t" name="time_boot_us" units="us">Timestamp (time since system boot).</field>
       <field type="float[4]" name="q">Quaternion components of autopilot attitude: w, x, y, z (1 0 0 0 is the null-rotation, Hamilton convention).</field>
-      <field type="uint32_t" name="q_estimated_delay_us" units="us">Estimated delay of the attitude data.</field>
-      <field type="float" name="vx" units="m/s">X Speed in NED (North, East, Down).</field>
-      <field type="float" name="vy" units="m/s">Y Speed in NED (North, East, Down).</field>
-      <field type="float" name="vz" units="m/s">Z Speed in NED (North, East, Down).</field>
-      <field type="uint32_t" name="v_estimated_delay_us" units="us">Estimated delay of the speed data.</field>
+      <field type="uint32_t" name="q_estimated_delay_us" units="us" invalid="0">Estimated delay of the attitude data. 0 if unknown.</field>
+      <field type="float" name="vx" units="m/s" invalid="NaN">X Speed in NED (North, East, Down). NAN if unknown.</field>
+      <field type="float" name="vy" units="m/s" invalid="NaN">Y Speed in NED (North, East, Down). NAN if unknown.</field>
+      <field type="float" name="vz" units="m/s" invalid="NaN">Z Speed in NED (North, East, Down). NAN if unknown.</field>
+      <field type="uint32_t" name="v_estimated_delay_us" units="us" invalid="0">Estimated delay of the speed data. 0 if unknown.</field>
       <field type="float" name="feed_forward_angular_velocity_z" units="rad/s" invalid="NaN">Feed forward Z component of angular velocity (positive: yawing to the right). NaN to be ignored. This is to indicate if the autopilot is actively yawing.</field>
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE" invalid="MAV_LANDED_STATE_UNDEFINED">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>


message_definitions/v1.0/common.xml
```diff
@@ -6950,11 +6950,11 @@
...
-      <field type="uint32_t" name="q_estimated_delay_us" units="us">Estimated delay of the attitude data.</field>
-      <field type="float" name="vx" units="m/s">X Speed in NED (North, East, Down).</field>
-      <field type="float" name="vy" units="m/s">Y Speed in NED (North, East, Down).</field>
-      <field type="float" name="vz" units="m/s">Z Speed in NED (North, East, Down).</field>
-      <field type="uint32_t" name="v_estimated_delay_us" units="us">Estimated delay of the speed data.</field>
+      <field type="uint32_t" name="q_estimated_delay_us" units="us" invalid="0">Estimated delay of the attitude data. 0 if unknown.</field>
+      <field type="float" name="vx" units="m/s" invalid="NaN">X Speed in NED (North, East, Down). NAN if unknown.</field>
+      <field type="float" name="vy" units="m/s" invalid="NaN">Y Speed in NED (North, East, Down). NAN if unknown.</field>
+      <field type="float" name="vz" units="m/s" invalid="NaN">Z Speed in NED (North, East, Down). NAN if unknown.</field>
+      <field type="uint32_t" name="v_estimated_delay_us" units="us" invalid="0">Estimated delay of the speed data. 0 if unknown.</field>
```
julianoes, line 6957: Or `UINT32_MAX` alternatively?

message_definitions/v1.0/common.xml
```diff
@@ -6950,11 +6950,11 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint64_t" name="time_boot_us" units="us">Timestamp (time since system boot).</field>
       <field type="float[4]" name="q">Quaternion components of autopilot attitude: w, x, y, z (1 0 0 0 is the null-rotation, Hamilton convention).</field>
-      <field type="uint32_t" name="q_estimated_delay_us" units="us">Estimated delay of the attitude data.</field>
-      <field type="float" name="vx" units="m/s">X Speed in NED (North, East, Down).</field>
-      <field type="float" name="vy" units="m/s">Y Speed in NED (North, East, Down).</field>
-      <field type="float" name="vz" units="m/s">Z Speed in NED (North, East, Down).</field>
-      <field type="uint32_t" name="v_estimated_delay_us" units="us">Estimated delay of the speed data.</field>
+      <field type="uint32_t" name="q_estimated_delay_us" units="us" invalid="0">Estimated delay of the attitude data. 0 if unknown.</field>
```
julianoes, line 6953: Or `UINT32_MAX` alternatively?

julianoes (approved):
Nice!

message_definitions/v1.0/common.xml
olliw42: I didn't since I got a bit tired of all the Uxxx_MAX and 0 should work equally ... no deep thoughts on my side
pl decide :)

julianoes: @hamishwillee help :shrug: 

What's generally `invalid` for extensions? 0 or UINT32 max?

olliw42: the two fields under discussion here are not extensions

message_definitions/v1.0/common.xml
hamishwillee: As long as 0 is actually never likely to be a real value then it's OK. If potentially someone might use 0 and mean it to be used in some way then better off using UINT32_MAX

message_definitions/v1.0/common.xml
```diff
@@ -6909,20 +6909,20 @@
...
+      <field type="float" name="yaw_max" units="rad" invalid="NaN">Maximum hardware yaw angle (positive: to the right, negative: to the left). NAN if unknown.</field>
     </message>
     <message id="284" name="GIMBAL_DEVICE_SET_ATTITUDE">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Low level message to control a gimbal device's attitude. This message is to be sent from the gimbal manager to the gimbal device component. Angles and rates can be set to NaN according to use case.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS">Low level gimbal flags.</field>
+      <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS" invalid="UINT16_MAX" display="bitmask">Low level gimbal flags. UINT16_MAX to be ignored.</field>
```
hamishwillee, line 6925: By contrast, flags do not have an invalid value. Generally this does not matter. If you really need to know they are supported you might use one flag for this, forced true if enabled.

message_definitions/v1.0/common.xml
```diff
@@ -6935,12 +6935,12 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS">Current gimbal flags set.</field>
+      <field type="uint16_t" name="flags" enum="GIMBAL_DEVICE_FLAGS" display="bitmask">Current gimbal flags set.</field>
```
hamishwillee, line 6938: usually we put the display= on the enum itself.

message_definitions/v1.0/common.xml
olliw42: that's true, I made the implizit assumption that it is just unrealistic that any real dealy(s) could be 0 microseconds. 
I mean, 1 us is really short on any relevant time scale, and if it is so small that it is 0 us, and we thus specify it is 0 us, what is the code supposed to do which is different to not doing anything? I just can't imagine any situation where you could do anything reasonable/usefull with knowing, hey, it is 0 us.

hamishwillee: If you both agree that, then 0 is fine. It seems logical to me that no system should have a 0 delay.

message_definitions/v1.0/common.xml
olliw42: ah ... oh ...

that's however an important value here (IMHO), since one exactly wants (needs) that possibility to force the recipient to ignore the flags ... !!
so, it's not about beeing supported or not, here it's indeed a function
(note that "invalid" does not always only mean "not supported/not available" but indeed can have the meaning "ignore" like in "do not pay attention")(usually this is then made clear in the field description by words like ignore, unkown, etc)

I noted that this might be kind of the first or one of the very few occurences of such a situation. In the FLAGS enum you will see that I have added a flag for UINT16_MAX. This may also maybe be seen as somewhat unusual. It all feels however consistent, more consistent than having e.g. a flag (1<<21) to indicate "ignore" (I guess implementers would easily stumble)

If this piece is offensive, I would not cry out if we would drop the invalid attribute here. I mean, the invalids are more "guides to the eyes" and don't have massive consequences if they are there or not. 
But the function must be there, we need some way to say "don't listen to this flag".

message_definitions/v1.0/common.xml
olliw42: ah, that's an interesting comment ... I was actually wondering about this (obvious) duplicity in info

if that's the rule (and how the automatic docs work), then I will be happy to change. 
Pl confirm.

I added them since I have seen them in other messages, so concluded that's what one wants. If not, even better.

message_definitions/v1.0/common.xml
hamishwillee: Lets see what @julianoes says. I'm in Friday evening mode :-)

olliw42:
to summarize the discussion points, to help make it easier to advance

* `invalid="0"` for fields `q_estimated_delay_us` and `v_estimated_delay_us`: It seems that this is ok for Hamish and me.

* `display="bitmask"` in messages: The current status is that it is often found to be used in messages, but Hamish argues that a display= should usually be put only on the enum itself. That's a contradiction. So, you need to decide how you want it (so basically if the PR should be "adding more bitmask" or be "removing bitmask").

* `invalid="UINT16_MAX"` in field `flags` which is a `GIMBAL_DEVICE_FLAGS ` enum: Function-wise, there needs to be a way to indicate that this flag should be ignored. Questions are by which flags value this should be represented (0 doesn't work) and if a flag can have an invalid attribute.

message_definitions/v1.0/common.xml
julianoes: Ok, so the question is whether the delay could ever be estimated to be 0. I guess the answer is no, so we might as well use 0 for invalid.

message_definitions/v1.0/common.xml
julianoes: Yer, let's leave it as 0.

message_definitions/v1.0/common.xml
julianoes: I don't think we can use `UINT16_MAX` which is `0xffff` so it means that all bits are 1, so all flags are set!

If you really need a way to say ignore the flags, then we need an "IGNORE_FLAGS` flag.

Why we needs this? I don't understand that.

message_definitions/v1.0/common.xml
julianoes: @hamishwillee I don't think that is correct. It's always used for the message fields as @olliw42 stated.

message_definitions/v1.0/common.xml
olliw42: > If you really need a way to say ignore the flags, then we need an "IGNORE_FLAGS` flag.

I find that not so nice, but if that is the way to go, I should be happy with it 

> Why we needs this? I don't understand that.

that's more interesting. First off, one could argue that's teh same for the quaternion/rates. 
Truth is however that I must admit that from top of my head I cannot reconstruct the cases there I found this to be relevant. It was about edge cases, and all I recall is that I ended up adding this to achieve defined states in certain transitions of state/conditions, in which moments of ambiguity could occur, yielding gimbal twitching.
I always wondered how you could deal with these situations with standard v2, and came to think - assume - that in your implementation you ignore it and simply hope that the next message is coming soon, so that gimbal twitches won't get too strong.
Unfortunately that's all I can say momentarily, not very compelling, I agree. Fact is, I just took it over from my STorM32 dialect ... (assuming that I got it right at the time when I did this LOL)

julianoes: Hehe, let's try without it I'd say.

olliw42: hmhmhm ... I scanned my current codes, and the problem is that I didn't use any gimbal manager stuff outside of STorM32 in the last two years or so (i.e. I use the STorM32 as gimbal manager and not as gimbal device) ... I note that I do have such a flag NONE for the STorM32 gimbal manager protocol too, and it is used, and makes the code/coding enourmously easier ... let me think about it a bit more, pl

message_definitions/v1.0/common.xml
hamishwillee: @olliw42 Please leave it as it is - sorry about that.

@julianoes Bugger We're in an in-between state.

The theory is that a ground station can use knowledge that a field is a bitmask to display it as checkboxes. This was applied as `display="bitmask"` to the field. Then we realised (quite some time ago) that enums are often used in multiple places, so it makes more sense to use `bitmask="true"` as a property on the enum and remove the value from the field.

However this last step has not been rolled out. Let me first find out if QGC "really" uses the display field.

message_definitions/v1.0/common.xml
hamishwillee: Please use a flag for none. It might not be elegant, but setting all the flags to indicate "no flags" is even less so. 

My concern here is that people know how flags are used in mavlink. We shouldn't create a special case if at all possible.

hamishwillee:
Only this one is blocking merging now:

> invalid="UINT16_MAX" in field flags which is a GIMBAL_DEVICE_FLAGS enum: Function-wise, there needs to be a way to indicate that this flag should be ignored. Questions are by which flags value this should be represented (0 doesn't work) and if a flag can have an invalid attribute.

Please use a flag, as discussed in https://github.com/mavlink/mavlink/pull/1912#discussion_r1029984424.

message_definitions/v1.0/common.xml
julianoes: I only see it in message fields. So I don't understand what you mean with in-between. Do you have an example where it's used in an enum? Maybe I'm blind or my grepping sucks.

hamishwillee: By "in between", I mean "we started to change from indicating bitmask on fields to indicating it on the enum, but we didn't finish the job".

You're not searching on `"bitmask="true"`": https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L25

Sorry I was unclear!

message_definitions/v1.0/common.xml
olliw42: I agree that 0xFFFF is not good and - if at all - should be a flag

I went through my things a bit deeper, and while for the manager the analog NONE is used and is important it seems that for the giombal device I only use it to manage initialization, but I can see that I don't have to do it that way.

So, my suggestion is to indeed follow Julian, i.e., to remove this in this PR. 

I mean, I will have quite some work to do to deprecate "my" gimbal device messages and go with mavlink's, and if in that process I see I do need this none flag, I can always ask again for it :) But at this point in time I do in fact have no real argument to justify it's need.

message_definitions/v1.0/common.xml
olliw42: well, I don't have an opinion in the sense that I do not really care. 
All I would suggest is to have a plan, and work towards having it consistent in the xml.

by my nature I would go with only having it in the enum definition, and not also in the message fields, as it's clearly duplicate info. Also: I very much like that bitmasks seem to be consistently named FLAGS/flag since a while. So the info is now actually tripple reduandant. I find that "naming convention" much more helpfull, since it is visible in the code and helps reminding what is a flag and what not while coding. So, if it would be me, my plan would be (1) naming convention FLAGS/flag and (2) bitmask attribute in the enum definition.

olliw42:
my plan now was to wait for the final vote on the bitmask thing, and then do the changes and rebase

message_definitions/v1.0/common.xml
julianoes: Makes sense to me!

message_definitions/v1.0/common.xml
julianoes: I'll leave it up to @hamishwillee. I don't have context on this.

hamishwillee: @olliw42 That ^^^ is pretty much my intent. But it's going to take some time to work though; what you have done now is consistent so this should not block merging THIS PR. I.e. conversation ended - I'll take this part forward separately.

message_definitions/v1.0/common.xml
hamishwillee: Great, if you remove this, and add the none flage (if needed), we can merge.

hamishwillee:
I think next steps agreed. Leave the bitmap thing alone for this PR, remove the invalid value for the flag field/replace it with a specific flag value.

olliw42:
fantastic :)

I rebased but seem to have screwed it up. Hence did a new PR https://github.com/mavlink/mavlink/pull/1927, and closing this here.

Sorry, sir.

PR #1913, opened 2022-11-08
Gimbal Device: Adding RC handling

olliw42:
This is the fourth of a sequence of four PRs, which are intended to finally get the looming changes with regards to the gimbal device protocol merged. These are of two kinds: handling of yaw absolute and relative, and handling of RC signals.

The four PRs are:
1. https://github.com/mavlink/mavlink/pull/1911
2. https://github.com/mavlink/mavlink/pull/1912
3. https://github.com/mavlink/mavlink/pull/1914
4. https://github.com/mavlink/mavlink/pull/1913

The additions to the gimbal device flags and messages have been discussed and agreed to before in
- https://github.com/mavlink/mavlink/issues/1860 "Yet another effort towards unified gimbal device messages"
- https://github.com/mavlink/mavlink/pull/1885 "gimbal: more flexible attitude messages"

see especially https://github.com/mavlink/mavlink/issues/1860#issuecomment-1189939693, https://github.com/mavlink/mavlink/issues/1860#issuecomment-1204714314

These PRs are intended to weed out the little kinks which were remaining.

The four PRs can technically each be merged for itself, but depend on each other, i.e., should not merged at all but in the proper sequence, and they will require rebasing to do so. However, I have done them spearately so you can better see what changes are there. Hopefully that's an appropriate approach.

### This PR: Adding RC handling

This PR adds two flags to allow RC input handling. This was discussed at length in https://github.com/mavlink/mavlink/issues/1860, and agreed to, see https://github.com/mavlink/mavlink/issues/1860#issuecomment-1189939693. This PR should hence be non-controversial, except of naming and language maybe.

_**Note**_: There are gaps in the enums, these areas are for the other yaw-related gimbal device flags from the 3rd PR. So, this PR should not be merged before the 3rd is merged, or needs to be modified.

Olli :)

PR #1914, opened 2022-11-08
Gimbal Device: Adding handling of absolute/relative yaw

olliw42:
This is the third of a sequence of four PRs, which are intended to finally get the looming changes with regards to the gimbal device protocol merged. These are of two kinds: handling of yaw absolute and relative, and handling of RC signals.

The four PRs are:
1. https://github.com/mavlink/mavlink/pull/1911
2. https://github.com/mavlink/mavlink/pull/1912
3. https://github.com/mavlink/mavlink/pull/1914
4. https://github.com/mavlink/mavlink/pull/1913

The additions to the gimbal device flags and messages have been discussed and agreed to before in
- https://github.com/mavlink/mavlink/issues/1860 "Yet another effort towards unified gimbal device messages"
- https://github.com/mavlink/mavlink/pull/1885 "gimbal: more flexible attitude messages"

see especially https://github.com/mavlink/mavlink/issues/1860#issuecomment-1189939693, https://github.com/mavlink/mavlink/issues/1860#issuecomment-1204714314

These PRs are intended to weed out the little kinks which were remaining.

The four PRs can technically each be merged for itself, but depend on each other, i.e., should not merged at all but in the proper sequence, and they will require rebasing to do so. However, I have done them spearately so you can better see what changes are there. Hopefully that's an appropriate approach.

### This PR: Adding handling of absolute/relative yaw

This is propbably the most "difficult" PR among the four, given that the abs/real yaw topic produced the most discussion and that it's also where things got stuck again.

The PR here is essentially @julianoes attempt https://github.com/mavlink/mavlink/pull/1885, with the kinks worked out, some additions, a hopefully complete definiton of the things, and maybe some language changes.

As part of the discussion in https://github.com/mavlink/mavlink/issues/1860 it was concluded that the revision 
- should be achieved via extensions to the messages
- should be backwards compatible, in all directions, old-old, new-old, old-new, and new-new.

These two requirements essentially rule out all the other suggestions, which may be simpler and cleaner, and leaves us with the approach to add two flags to indicate the frame for yaw (which is exactly what @julianoes' PR was also doing).

With this given, let's go through the various changes made in this PR:

#### 1. It adds the two flags GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME and GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME. 

These are to indicate which frame is used for yaw in the messages GIMBAL_DEVICE_SET_ATTITUDE and GIMBAL_DEVICE_ATTITUDE_STATUS. One needs them in order to break with the current apporach there the frame is linked to the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK, which however was identified to be too restrictive. One may think one can do it with only one flag, but this would violate backwards compatibility. Hence two flags are needed. The descriptions hopefully explain their usage.

Just as a memo, there is actually a migration path towards there these flags can be deprecated in some future:
- Today, in order to achieve full back wards compatibility, one should set GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME if the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is not set, and set GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME if the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK  is set.
- In a next step, i.e. a next version round of the implementations, the implementations could go towards always setting  GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME, but never setting GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME. This would much simplify the usage of the messages, compared to the slight hoops necessary for now.
- In a last step, i.e. yet another next version round, bith flags would become obsolete. 

#### 2. The extensions delta_yaw and delta_yaw_velocity are added to the message GIMBAL_DEVICE_ATTITUDE_STATUS. 

These are to allow the conversion from one frame to the other with the data given in the main body of the message. The message descriptions hopefully precisely define the relation ships. In order to ensure back wards compatibility, some definition is required to distinguish and handle the case delta_yaw=delta_yaw_velocity=0, which means no extension provided, from other cases.

#### 3. The cap flag GIMBAL_DEVICE_CAP_FLAGS_SUPPORTS_YAW_IN_EARTH_FRAME is added. 
 
It might be that one can find a set of definitions such that this flag is not needed, it however seems to be an obvious good idea, to just tell straight forwardly. So far this would have been achieved by setting the flag GIMBAL_DEVICE_CAP_FLAGS_HAS_YAW_LOCK, i.e., if this flag was set one would have concluded that the gimbal must also be able to handle absolute yaw. Since we broke with this correlation however, we want to have a new respective cap flag. Not in @julianeous attempt, but should be an obvious addition.

#### 4. The flag GIMBAL_DEVICE_FLAGS_ACCEPTS_YAW_IN_EARTH_FRAME is added. 

This is a new one. It allows to achieve two things:
- first, it resolves a potential race condition at startup, which goes as follows: The gimbal first needs to receive AUTOPILOT_STATE_FOR_GIMBAL_DEVICE before it can decide if it can support absolute yaw, however the capabilities would be usually requested very early on. The only way out currently would be for the gimbal device to NOT send the capabilities or to nack such requests until it eventually has gotten the AUTOPILOT_STATE_FOR_GIMBAL_DEVICE messages, with the estimator flags set as needed. In short, the gimbal device discovery would have to be postponed until the autopilot has achieved position lock (or similar). With the suggested flags, the gimbal device could immediately respond to capability requests, but could indicate to a gimbal manager that it can't yet use yaw in earth frame features.
- second, it well can happen that during operation e.g. position lock is lost, and that the gimbal cannot do the yaw earth frame things for the period of time there this error condition exists.

One could instead of adding this flag add a further condition to the ERROR_FLAGS enum. This would bode well with the second situation, I however didn't considered this because I always ened up with something which didn't look natural to me. E.g., if a gimbal just would not support yaw earth frame, there would be always a flag raised in the error flags, although nothing went wrong. I.e., erroflags == 0 would not mean no error anymore.

#### 5. The extension "angular_velocity_z" was added to the AUTOPILOT_STATE_FOR_GIMBAL_DEVICE message. 

This is also a new one, and in fact corrects for an omission which should not have occured. Without that data, the gimbal cannot translate the angular velocities from earth to vehicle and vice versa. This means that it e.g. cannot achieve a steady rotation around z when the control is in earth frame, i.e., the camera would wiggle then the vehicle would wiggle.

Olli :)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -462,6 +462,9 @@
       <entry value="2048" name="GIMBAL_DEVICE_CAP_FLAGS_SUPPORTS_INFINITE_YAW">
         <description>Gimbal device supports yawing/panning infinetely (e.g. using slip disk).</description>
       </entry>
+      <entry value="4096" name="GIMBAL_DEVICE_CAP_FLAGS_SUPPORTS_YAW_IN_EARTH_FRAME">
+        <description>Gimbal device supports yaw angles and angular velocities relative to North (earth frame). This usually requires support by an autopilot via AUTOPILOT_STATE_FOR_GIMBAL_DEVICE. Support can go on and off during runtime, which is reported by the flag GIMBAL_DEVICE_FLAGS_CAN_ACCEPT_YAW_IN_EARTH_FRAME.</description>
+      </entry>
       <entry value="8192" name="GIMBAL_DEVICE_CAP_FLAGS_HAS_RC_INPUTS">
         <description>Gimbal device supports radio control inputs as an alternative input for controlling the gimbal orientation.</description>
       </entry>
@@ -527,6 +530,15 @@
       </entry>
       <entry value="16" name="GIMBAL_DEVICE_FLAGS_YAW_LOCK">
         <description>Lock yaw angle to absolute angle relative to North (not relative to vehicle). If this flag is set, the yaw angle and z component of angular velocity are relative to North (earth frame, x-axis pointing North), else they are relative to the vehicle heading (vehicle frame, earth frame rotated so that the x-axis is pointing forward).</description>
+      </entry>
+      <entry value="32" name="GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME">
+        <description>Yaw angle and z component of angular velocity are relative to the vehicle heading (vehicle frame, earth frame rotated such that the x-axis is pointing forward).</description>
+      </entry>
+      <entry value="64" name="GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME">
+        <description>Yaw angle and z component of angular velocity are relative to North (earth frame, x-axis is pointing North).</description>
+      </entry>
+      <entry value="128" name="GIMBAL_DEVICE_FLAGS_ACCEPTS_YAW_IN_EARTH_FRAME">
+        <description>Gimbal device can accept yaw angle inputs relative to North (earth frame). This flag is only for reporting (attempts to set this flag are ignored).</description>
       </entry>
       <entry value="256" name="GIMBAL_DEVICE_FLAGS_RC_EXCLUSIVE">
         <description>The gimbal orientation is set exclusively by the RC signals feed to the gimbal's radio control inputs. MAVLink messages for setting the gimbal orientation (GIMBAL_DEVICE_SET_ATTITUDE) are ignored.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6919,28 +6931,31 @@
...
...EARTH_FRAME is set, then they are relative to absolute North (earth frame). If neither of these flags are set, then it holds: If the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set, then they are relative to absolute North (earth frame), else they are relative to the vehicle heading (vehicle frame). Other conditions of the flags are not allowed. If both delta_yaw and delta_yaw_velocity are not zero (NaN counts as non-zero), then the quaternion and angular velocities in the other frame can be calculated from delta_yaw and delta_yaw_velocity as q_earth = q_delta_yaw * q_vehicle and w_earth = w_delta_yaw_velocity + w_vehicle. In the unlikely event that delta_yaw and delta_yaw_velocity should be provided but are both exactly zero, then delta_yaw should be set to a very small value, e.g. 1.0E-20. If both delta_yaw and delta_yaw_velocity are zero, then these data are not provided in the message and these fields are to be ignored. These rules are to ensure backwards compatibility. New implementations should always set either GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME or GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME, and always should set delta_yaw_velocity either to the proper value or NaN.</description>
```
julianoes, line 6946:
```suggestion
      <description>Message reporting the status of a gimbal device. This message should be broadcasted by a gimbal device component at a low regular rate (e.g. 10 Hz). For the angles encoded in the quaternion and the angular velocities holds: If the flag GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME is set, then they are relative to the vehicle heading (vehicle frame). If the flag GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME is set, then they are relative to absolute North (earth frame). If neither of these flags are set, then (for backwards compatibility) it holds: If the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set, then they are relative to absolute North (earth frame), else they are relative to the vehicle heading (vehicle frame). Other conditions of the flags are not allowed. If both delta_yaw and delta_yaw_velocity are not zero (NaN counts as non-zero), then the quaternion and angular velocities in the other frame can be calculated from delta_yaw and delta_yaw_velocity as q_earth = q_delta_yaw * q_vehicle and w_earth = w_delta_yaw_velocity + w_vehicle. In the unlikely event that delta_yaw and delta_yaw_velocity should be provided but are both exactly zero, then delta_yaw should be set to a very small value, e.g. 1.0E-20. If both delta_yaw and delta_yaw_velocity are zero, then these data are not provided in the message and these fields are to be ignored. These rules are to ensure backwards compatibility. New implementations should always set either GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME or GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME, and always should set delta_yaw_velocity either to the proper value or NaN.</description>
```

I don't like the note about having to set a value to 1E-20 but I don't have a better suggestion :thinking:.

message_definitions/v1.0/common.xml
```diff
@@ -6919,28 +6931,31 @@
...
+      <field type="float" name="angular_velocity_x" units="rad/s" invalid="NaN">X component of angular velocity (positive: rolling to the right). The frame is described in the message description. NaN if unknown.</field>
+      <field type="float" name="angular_velocity_y" units="rad/s" invalid="NaN">Y component of angular velocity (positive: pitching up). The frame is described in the message description. NaN if unknown.</field>
+      <field type="float" name="angular_velocity_z" units="rad/s" invalid="NaN">Z component of angular velocity (positive: yawing to the right). The frame is described in the message description. NaN if unknown.</field>
       <field type="uint32_t" name="failure_flags" display="bitmask" enum="GIMBAL_DEVICE_ERROR_FLAGS">Failure flags (0 for no failure)</field>
+      <extensions/>
+      <field type="float" name="delta_yaw" units="rad">Yaw angle relating the quaternions in earth and body frames (see message description).</field>
+      <field type="float" name="delta_yaw_velocity" units="rad/s" invalid="NAN">Yaw angular velocity relating the angular velocities in earth and body frames (see message description). NaN if unknown.</field>
```
julianoes, line 6958: Given these are extensions I wonder if unknown should be 0 or NAN. If you set it 0, then it can get trimmed, so is more efficient. However, when you get a 0 you don't know if it's unknown or not populated yet.

message_definitions/v1.0/common.xml
```diff
@@ -6958,6 +6973,8 @@
       <field type="float" name="feed_forward_angular_velocity_z" units="rad/s" invalid="NaN">Feed forward Z component of angular velocity, positive is yawing to the right, NaN to be ignored. This is to indicate if the autopilot is actively yawing.</field>
       <field type="uint16_t" name="estimator_status" enum="ESTIMATOR_STATUS_FLAGS" display="bitmask">Bitmap indicating which estimator outputs are valid.</field>
       <field type="uint8_t" name="landed_state" enum="MAV_LANDED_STATE" invalid="MAV_LANDED_STATE_UNDEFINED">The landed state. Is set to MAV_LANDED_STATE_UNDEFINED if landed state is unknown.</field>
+      <extensions/>
+      <field type="float" name="angular_velocity_z" units="rad/s" invalid="NaN">Z component of angular velocity in NED (North, East, Down). NaN if unknown.</field>
```
julianoes, line 6977: Same question here: NaN or 0. I'm not sure.

message_definitions/v1.0/common.xml
olliw42: yeah, that's the least appealing piece of the kit
I'll give it another thought

olliw42: note to myself: broadcast & 5 Hz

message_definitions/v1.0/common.xml
olliw42: the idea is that the extension so to say becomes a historical fluke and everyone will follow the recommendation to set them ... so, IMHO we shouldn't think and decide from the perspective that this is an extension. floats should be NAN for unknown (and all are, except few flukes here and there).

message_definitions/v1.0/common.xml
hamishwillee: 0 by default. 0 will always be recieved if this is sent by a non-updated component, so the recipient needs to know that the data is invalid in this case.

If 0 is a valid value for the range, we either note that you should use some small range around 0 that makes no difference to represent "0" (e.g. less than 0.01 - whatever is appropriate for radians), or you have another field that defines a bitmask for the enabled extensions. If it isn't set because the sender hasn't been updated you know that the extension has garbage.

message_definitions/v1.0/common.xml
hamishwillee: See https://github.com/mavlink/mavlink/pull/1914#discussion_r1026117640

We don't normally believe in flukes, but design not to break things. However I can't assess how screwed up a GCS or other system would be if it was connected to the current set of gremsy gimbals that do not understand these changes .

If you use zero or an extension bit, you don't need to worry.

message_definitions/v1.0/common.xml
hamishwillee: Yes, use "broadcast". I'm going to suggest using whitespace to make this readable. It makes the generated code render rubbish, but makes our job easier here. Based on Julian's change above, but with broadcast

```suggestion
      <description>Message reporting the status of a gimbal device. 
        This message should be broadcasted by a gimbal device component at a low regular rate (e.g. 10 Hz).
        
        For the angles encoded in the quaternion and the angular velocities holds: 
        - If the flag GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME is set, then they are relative to the vehicle heading (vehicle frame). 
        - If the flag GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME is set, then they are relative to absolute North (earth frame).
        - If neither of these flags are set, then (for backwards compatibility) it holds: If the flag GIMBAL_DEVICE_FLAGS_YAW_LOCK is set, then they are relative to absolute North (earth frame), else they are relative to the vehicle heading (vehicle frame).
        - Other conditions of the flags are not allowed. 
        
        If both delta_yaw and delta_yaw_velocity are not zero (NaN counts as non-zero), then the quaternion and angular velocities in the other frame can be calculated from delta_yaw and delta_yaw_velocity as q_earth = q_delta_yaw * q_vehicle and w_earth = w_delta_yaw_velocity + w_vehicle.
        In the unlikely event that delta_yaw and delta_yaw_velocity should be provided but are both exactly zero, then delta_yaw should be set to a very small value, e.g. 1.0E-20.
        If both delta_yaw and delta_yaw_velocity are zero, then these data are not provided in the message and these fields are to be ignored. These rules are to ensure backwards compatibility.
        
        New implementations should always set either GIMBAL_DEVICE_FLAGS_YAW_IN_VEHICLE_FRAME or GIMBAL_DEVICE_FLAGS_YAW_IN_EARTH_FRAME, and always should set delta_yaw_velocity either to the proper value or NaN.</description>
```

olliw42: broadcast & 5Hz ;) this second sentence simply should become identical to what was just merged in the "spelling" PR (I will ensure it is)

olliw42:
so, guys, I did two things, thinking through again backwards compatibility, and going through common.xml and check how extensions and invalids are handled elsewhere.


## backwards compatibility

I conclude we were trying to "solve" an issue which is not existing. Let's look at the three cases "old -> old", "old -> new", "new -> old" for each message GIMBAL_DEVICE_ATTITUDE_STATUS and AUTOPILOT_STATE_FOR_GIMBAL_DEVICE separately (the case "new -> new" obviously doesn't incurr backwards issues).

#### GIMBAL_DEVICE_ATTITUDE_STATUS


"old -> old": 

The gimbal sends the status message with no GIMBAL_DEVICE_FLAGS_xxx_FRAME flags set, and without extension. The GCS does receive it with no FRAME flags set and without extension. That is, things work as today. The GCS can NOT produce the other frame, as it only gets the data for the one or the other.

"old -> new": 

The gimbal sends the status message with no GIMBAL_DEVICE_FLAGS_xxx_FRAME flags set, and without extension. The GCS does receive it with no FRAME flags set and with extensions set to 0. Since no FRAME flags are set, the GCS knows that the extension values are invalid information. That is, the GCS must NOT produce the other frame, as it only gets the data for the one or the other.

"new -> old": 

The gimbal sends the status message with at least one GIMBAL_DEVICE_FLAGS_xxx_FRAME flag set, and with extension. The GCS does receive it with the FRAME flags set but does NOT know about them, and also does not know about the extension. That is, the GCS can NOT produce the other frame, as it only gets the date for the one or the other.

Thus, in all three cases we end up with exactly the same result, namely that the GCS wouldn't produce the other frame, which implies that whatever data are sent in the extension is totally irrelevant. Ergo, we don't need to spend any effort for trying to sepcify if there is an extension or not.

In fact, it is my fault that I have forgotten that the reason for why we came up with TWO flags for specifying the frame was exactly to resolve the ambiguity with respect to the extension! 

The description text needs to change, and it in fact should become simpler now.

#### AUTOPILOT_STATE_FOR_GIMBAL_DEVICE

"old -> old": 

The autopilot sends the AS4GD message with no angular_velocity_z extension. The gimbal device receives it with no extension. That is, things work as today. The gimbal can NOT account for angular_velocity_z, i.e., would produce the wrong behavior as of today, where it assumes that the vehicle is not yawing.

"old -> new": 

The autopilot sends the AS4GD message with no angular_velocity_z extension. The gimbal device receive it with extension value 0. That is, the gimbal would produce a behavior where it assumes that the vehicle is not yawing.

"new -> old": 

The autopilot sends the AS4GD message with angular_velocity_z extension. The gimbal can NOT account for angular_velocity_z, i.e., would produce the wrong behavior as of today, where it assumes that the vehicle is not yawing.

Thus, in all three cases we end up with exactly the same result, namely that the gimbal device will not account for that the vehicle is yawing. Ergo, we don't need to do any extra effort for sepcifing if there is an extension or not.

olliw42:
I've tried to rebase and push changes, not sure I managed to do this correctly ... looks not so wrong, but ...

julianoes:
@olliw42 I think your argument is valid, so I'm happy with that.

olliw42:
@julianoes @hamishwillee 

rebased it and changed the descriptions to refelect the latest insight. Should getting closer now.

julianoes (approved):

julianoes (approved):

olliw42:
JUCHUUUU :)

many thx, guys

julianoes:
Thank you @olliw42 for tackling all these, very much appreciated. It had stalled on my side.

PR #1915, opened 2022-11-08
For discussion: changes to standard flight modes

bkueng:
Standard modes were added in https://github.com/mavlink/mavlink/pull/1750. I'm currently going through the implementation of that and have a few things I'd like to bring up for discussion.

- removes 'base_mode', as this contains dynamic bits, like `MAV_MODE_FLAG_SAFETY_ARMED`. Maybe I just don't understand why this was added in the first place?
- adds properties to modes:
  - whether a mode should be user-selectable. In QGC this is currently hard-coded and includes takeoff, orbit or land.
  - whether a mode is flagged as 'advanced', so a UI can hide it
- adds a user intended mode field. This is generally equal to custom_mode, but might be different in a failsafe condition
- adds a change notification for available modes. This allows to dynamically add and remove modes to a system


hamishwillee:
> removes 'base_mode', as this contains dynamic bits, like MAV_MODE_FLAG_SAFETY_ARMED. Maybe I just don't understand why this was added in the first place?

The theory is that this `AVAILABLE_MODES` contains all the custom and standard mode information so you can get information about either/both, and also map your standard modes to custom modes and visa versa.

@peterbarker Has ArduPilot started implemented standard modes yet?

hamishwillee:
> adds properties to modes:
> - whether a mode should be user-selectable. In QGC this is currently hard-coded and includes takeoff, orbit or land.
> - whether a mode is flagged as 'advanced', so a UI can hide it

I'm not certain I understand the intend for both cases, or whether these are needed. I think "user selectable" means you'll allow the mode to be directly set by a user via a button or whatever, and that for not user-selectable you'll allow it to be displayed but not selected in the UI - so for example an acro mode would be not user selectable and a position mode would be. The acro mode would then be displayed if activated by a switch, but UI wouldn't expose it?

For "advanced" you're suggesting that this would be a property of the mode that a UI wouldn't show it even if selected? WOuldn't that be confusing?

I don't understand why these are needed for standard modes; the modes are well defined so surely a GCS can hard code how it handles these. It isn't up to the flight stack to say "hey, my standard mode is X, Y, Z".

Where these might be useful is actually on custom modes - so a GCS that knows nothing about the mode can still have a hint on how it is displayed.

hamishwillee:
> adds a user intended mode field. This is generally equal to custom_mode, but might be different in a failsafe condition

I don't understand how this is used.

> adds a change notification for available modes. This allows to dynamically add and remove modes to a system

This is a good idea if we have the case where a new mode can be added after boot. Is this real or just forward looking?

message_definitions/v1.0/development.xml
```diff
@@ -466,8 +478,14 @@
         It may be requested using MAV_CMD_REQUEST_MESSAGE.
       </description>
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
+      <field type="uint32_t" name="custom_mode_user_intention">custom_mode of the current user intended mode. This is generally equal to custom_mode, but might be different in a failsafe condition.</field>
+    </message>
+    <message id="437" name="AVAILABLE_MODES_CHANGE_NOTIFICATION">
+      <description>Notification message for updates to AVAILABLE_MODES. A receiver must re-request all avaible modes whenever the sequence number changes.
```
hamishwillee, line 485:
```suggestion
      <description>Notification message for updates to AVAILABLE_MODES. A receiver must re-request all available modes whenever the sequence number changes.
```

bkueng: Applied

message_definitions/v1.0/development.xml
```diff
@@ -456,18 +464,24 @@
...
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
+      <field type="uint32_t" name="custom_mode_user_intention">custom_mode of the current user intended mode. This is generally equal to custom_mode, but might be different in a failsafe condition.</field>
+    </message>
+    <message id="437" name="AVAILABLE_MODES_CHANGE_NOTIFICATION">
+      <description>Notification message for updates to AVAILABLE_MODES. A receiver must re-request all available modes whenever the sequence number changes.
+        This is only emitted after the first change and should then be broadcast at low rate (nominally 0.3 Hz) or on change.
+      </description>
+      <field type="uint8_t" name="seq">Sequence number. This is increased whenever AVAILABLE_MODES changes</field>
     </message>
```
hamishwillee, line 485: Simplification:
```suggestion
    <message id="437" name="AVAILABLE_MODES_CHANGED">
      <description>Notifies that the set of AVAILABLE_MODES has changed.
        A receiver must re-request all available modes whenever the sequence number changes.
        This is only emitted after the first change and should then be broadcast at low rate (nominally 0.3 Hz) and on change.
      </description>
      <field type="uint8_t" name="seq">Sequence number. This iterates sequentially whenever AVAILABLE_MODES changes. Note that the value itself is not useful, only the fact that it has changed.</field>
    </message>
```

message_definitions/v1.0/development.xml
```diff
@@ -456,18 +464,24 @@
...
+      <field type="uint32_t" name="properties" enum="MAV_STANDARD_MODE_PROPERTY">Mode Properties.</field>
+      <field type="char[25]" name="mode_name">Name of custom mode, with null termination character. Should be omitted for standard modes.</field>
     </message>
     <message id="436" name="CURRENT_MODE">
       <description>Get the current mode.
         This should be emitted on any mode change, and broadcast at low rate (nominally 0.5 Hz).
         It may be requested using MAV_CMD_REQUEST_MESSAGE.
       </description>
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
```
hamishwillee, line 469: Should this also go from here? 

We agreed removing this from `AVAILABLE_MODES` because the information isn't needed for infering the custom/standard mode relationship, and because this is not streamed it cannot be a mechanism for getting the dynamic information in this field.

However the CURRENT_MODE is streamed, so this is potentially useful if you are interested in the base mode. Yes this information is in the HEARTBEAT so it "could" go - but it is convenient to get all relevant information in one place.

@auturgy Did you have an opinion?

message_definitions/v1.0/development.xml
```diff
@@ -151,6 +151,14 @@
         </description>
       </entry>
     </enum>
+    <enum name="MAV_STANDARD_MODE_PROPERTY">
```
hamishwillee, line 154: Is this going to be a flag? If so:

```suggestion
    <enum name="MAV_STANDARD_MODE_PROPERTY" bitmask="true">
```
Do we really  need a uint32_t?

message_definitions/v1.0/development.xml
bkueng: Fine with me I just thought a name like that indicates it's changed whenever the message is sent out whereas in fact it's only when `seq` changes.

message_definitions/v1.0/development.xml
bkueng: Yes that is what I was looking for but didn't see it.
We can use 16 bits if you prefer. I don't think it matters here going for 32 bits though.

message_definitions/v1.0/development.xml
hamishwillee: @bkueng You make a good point, but isn't that an possible interpretation for either name? I've modified it a little to make the description "A change to the sequence number indicates that the set of AVAILABLE_MODES has changed."

hamishwillee: On reflection, yours is slightly better - but you can still infer getting the message is enough. Perhaps AVAILABLE_MODES_CHANGE_MONITOR or similar?

peterbarker:
On Tue, 8 Nov 2022, Hamish Willee wrote:

> @peterbarker Has ArduPilot started implemented standard modes yet?

I don't believe I've seen anybody do any work towards it.

message_definitions/v1.0/development.xml
bkueng: Yes, let's go with MONITOR

hamishwillee: OK. I've modified suggestion above - look OK?

bkueng:
I came across another use-case for the user-selectable bit (`MAV_STANDARD_MODE_PROPERTY_CAN_BE_SET`): A mode replacing another mode, where the replacing mode is selectable, but not the replaced. And the system might still switch into the replaced mode if the other one becomes unavailable.
Any better ideas to handle that?

hamishwillee:
> I came across another use-case for the user-selectable bit (MAV_STANDARD_MODE_PROPERTY_CAN_BE_SET): A mode replacing another mode, where the replacing mode is selectable, but not the replaced. And the system might still switch into the replaced mode if the other one becomes unavailable.
Any better ideas to handle that?

@bkueng When you say "replacing" do you mean the case where you set a mode and have a failsafe, so you go to fallback mode and put the desired mode in `custom_mode_user_intention`?

So by "imagined" example, I think you're saying that we are in mission mode, we set position mode (not advanced), but we don't have a position estimate so we fall back to (say) Acro mode (advanced). I can set position, and I'm not trying to set ACRO. Why do I need MAV_STANDARD_MODE_PROPERTY_CAN_BE_SET here?

bkueng:
> @bkueng When you say "replacing" do you mean the case where you set a mode and have a failsafe, so you go to fallback mode and put the desired mode in custom_mode_user_intention?

No, replacing a specific autopilot internal mode with an external implementation. For example the autopilot has an RTL mode, and a companion computer replacing that with a more elaborate version. Triggering RTL (from a GCS or failsafe) will then use the companion mode, and only fallback to the internal if it's not available. So the user would only ever select RTL (it's transparent for the most part, but it could have a different name).

hamishwillee:
>> @bkueng When you say "replacing" do you mean the case where you set a mode and have a failsafe, so you go to fallback mode and put the desired mode in custom_mode_user_intention?

> No, replacing a specific autopilot internal mode with an external implementation. For example the autopilot has an RTL mode, and a companion computer replacing that with a more elaborate version. Triggering RTL (from a GCS or failsafe) will then use the companion mode, and only fallback to the internal if it's not available. So the user would only ever select RTL (it's transparent for the most part, but it could have a different name).

Thanks. This is taking things way beyond my original thinking. In particular I'm not sure how you'd handle the fallback stuff - which is out of scope.

Scenario: 
- The autopilot initially outputs a bunch AVAILABLE_MODES including the standard mode "return mode" (which probably maps to the custom return mode in PX4's case).
-  The companion computer also outputs some modes - which for sake of argument also includes a standard mode "return mode".

1. What I think you're saying is that you want the GCS to aggregate all the available flight modes from vehicle, irrespective of the source. In this case you have duplicates, and you'd like the ability to hide the one from the autopilot (for setting) in the UI. 
  - Is that right? 
2. Further, the companion might fail the return mode and you want the Flight stack return to take over - but this should not be obvious in terms of reported mode to the GCS
   - My assumption here is that both autopilot and companion continue to output the same available modes. The companion might fail when you set the mode or while running the mode, but it will silently do whatever is needed to trigger the autopilot fallback? 
3. Modes from multiple sources is completely valid use case that has not been addressed (and must be).  My initial thinking is that 
    - a GCS should accept all modes from all components on a system, and will need to track the modes vs component.
    - all unique modes should be displayed according to their hidden flag and GCS preferences.
    - all duplicate custom modes should be displayed
    - I don't think you want to ever display duplicate standard modes on one vehicle. So yes, the flag is a valid use case.
    
Would it be a pain to have a flag "can be set"? By default an autopilot would have return mode "can be set" so you'd need a parameter or whatever to turn this off. 

Are there alternatives? 
- Could we use a rule such as use duplicate mode from companion computer as preference, other component as fallback, and autopilot as last resort?
- Or would a prefer_this flag be better - i.e. a companion could set this flag and it would be the default version used?

bkueng:
> Is that right?

Yes

> but this should not be obvious in terms of reported mode to the GCS 

There's both options, and I expect some people want to have a different label (e.g. `RTL (external)`) in order to know what's going on. Hence it's exposed to the GCS.

> The companion might fail when you set the mode or while running the mode, but it will silently do whatever is needed to trigger the autopilot fallback?

Yes, but not silent, it will report any errors.

> Modes from multiple sources is completely valid use case that has not been addressed (and must be). My initial thinking is that 

This is not how it will be implemented. From a GCS point of view, all modes come from the autopilot component.

> By default an autopilot would have return mode "can be set" so you'd need a parameter or whatever to turn this off.

No, as above, the autopilot knows which modes are replaced.

hamishwillee:
THanks @bkueng - I added question on discord related to the implementation which I think will help me understand this better.

> This is not how it will be implemented. From a GCS point of view, all modes come from the autopilot component.

This bit confuses me - if the companion supports "Hamish Mode", how can the autopilot emit that?

More importantly though, I think we do need to consider the architectures that others might adopt .

message_definitions/v1.0/development.xml
hamishwillee: FYI, @auturgy Was OK with this.

message_definitions/v1.0/development.xml
```diff
@@ -151,6 +151,21 @@
...
+      </description>
+      <entry value="1" name="MAV_STANDARD_MODE_PROPERTY_ADVANCED">
+        <description>If set, this mode is an advanced mode.
+          For example a rate-controlled manual mode might be advanced, whereas a position-controlled manual mode is not.
+          A GCS can optionally use this flag to configure the UI for its intended users. 
+        </description>
+      </entry>
+      <entry value="2" name="MAV_STANDARD_MODE_PROPERTY_NOT_USER_SELECTABLE">
+        <description>If set, this mode should not be added to the list of selectable modes.
+		  The mode might still be selected by the FC directly (for example as part of a failsafe).
```
hamishwillee, line 165:
```suggestion
          The mode might still be selected by the FC directly (for example as part of a failsafe).
```

message_definitions/v1.0/development.xml
```diff
@@ -457,18 +472,25 @@
...
     </message>
     <message id="436" name="CURRENT_MODE">
       <description>Get the current mode.
         This should be emitted on any mode change, and broadcast at low rate (nominally 0.5 Hz).
         It may be requested using MAV_CMD_REQUEST_MESSAGE.
       </description>
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
+      <field type="uint32_t" name="custom_mode_user_intention">custom_mode of the current user intended mode. This is generally equal to custom_mode, but might be different in a failsafe condition.</field>
```
hamishwillee, line 486: @bkueng Just trying to make this a little more clear. 

Note, to be useful with standard modes this means that every standard mode must have an associated custom mode. Otherwise if the intended mode was a standard mode the value will be unknown. I don't see that as onerous, since likely everyone will create one with associated custom mode. But anyhow, I have set 0 as the invalid value - OK?


```suggestion
      <field type="uint32_t" name="intended_custom_mode" invalid="0">The custom_mode of the mode that was last commanded by the user (for example, with MAV_CMD_DO_SET_STANDARD_MODE or MAV_CMD_DO_SET_MODE). This should usually be the same as custom_mode. It will be different if the vehicle is unable to enter the intended mode, or has left that mode due to a failsafe condition. 0 indicates the intended custom mode is unknown/not supplied</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -457,18 +472,25 @@
       <field type="uint8_t" name="number_modes">The total number of available modes for the current vehicle type.</field>
       <field type="uint8_t" name="mode_index">The current mode index within number_modes, indexed from 1.</field>
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
-      <field type="char[50]" name="mode_name">Name of custom mode, with null termination character. Should be omitted for standard modes.</field>
+      <field type="uint32_t" name="properties" enum="MAV_STANDARD_MODE_PROPERTY">Mode Properties.</field>
+      <field type="char[25]" name="mode_name">Name of custom mode, with null termination character. Should be omitted for standard modes.</field>
```
hamishwillee, line 477: - Custom modes that are standard modes may or may not have knowns strings for the mode name in the recipient. 
  - This field allows the name to be supplied for the mode that is unknown to the recipient.
  - My original intent was that a GCS could use its own information for a known GCS, including translations, if it had such information. 
- Standard modes should generally have known strings in any consumer of this service (including translations), so most of the time you wouldn't want or need to send a string for the mode name.

We have a new use case - what do you do if you need to modify the string for a standard mode?

For example, say you have a return mode implemented offboard (standard mode), with fallback to an onboard return mode (custom). Both modes are emitted by the flight stack. The internal mode is custom, so it can send whatever string it likes "Return mode (internal)". The standard mode would normally get the standard text "Return mode" (say) but we want it to instead say something like "Return Mode (external). 

@bkueng proposal here is that if specified the mode name string overrides whatever is stored in the GCS if sent. 

```suggestion
      <field type="char[25]" name="mode_name">Name of custom mode, with null termination character. Should normally be omitted for standard modes. If supplied it should be used instead of any mode name information stored by a recipient.</field>
```

Is this OK?
- It does mean that if you send the name then you can't choose to use translated strings for a custom mode that the GCS might already have.
- We could make this only override standard mode text?
- Is there a better way?

message_definitions/v1.0/development.xml
```diff
@@ -151,6 +151,21 @@
...
+    <enum name="MAV_STANDARD_MODE_PROPERTY" bitmask="true">
+      <description>Mode properties.
+      </description>
+      <entry value="1" name="MAV_STANDARD_MODE_PROPERTY_ADVANCED">
+        <description>If set, this mode is an advanced mode.
+          For example a rate-controlled manual mode might be advanced, whereas a position-controlled manual mode is not.
+          A GCS can optionally use this flag to configure the UI for its intended users. 
+        </description>
+      </entry>
+      <entry value="2" name="MAV_STANDARD_MODE_PROPERTY_NOT_USER_SELECTABLE">
```
hamishwillee, line 163: This differs from advanced - which a GCS might choose to offer for users to select or not, or might choose to enable for some users and not others. 
This flag indicates a mode that the should never directly trigger, and so it should not be offered in selection lists or for setting via a switch.

hamishwillee:
Note that Beat is implementing this to support an architecture that uses standard modes, and also supports a companion computer.

His architecture has all modes emitted/managed by the flight stack. However the actual setpoints for some modes might come from an offboard controller. There will be fallback to a default mode if the companion control fails. 

In the particular case of return mode he has a "Standard mode" with setpoint supplied by the companion, and a fallback (custom) mode with setpoint controlled by the flight stack.

Note,
- there is just one standard mode in this architecture. 
- the service theoretically allows for multiple standard modes provided:
  - the modes all comply with the definition of the standard mode. For many cases this would mean there was no point having multiples, because they would have to be "the same mode" (so why would you bother)
  - The standard modes are emitted by different components - e.g. a companion and a flight stack.
    - If they were on the same component you would have no way to address which standard mode version to select.
    - We have not addressed this case - i.e. what should the GCS do if it detects modes emitted by multiple systems.
      - IMO it should aggregate and display all modes, using the flags to indicate the target. Ignore for now?

auturgy:
Not much about this feels like we’re working towards a “standard” at all anymore, and the description doubles down on that feeling. It very much sounds like you’re trying to call a very custom implementation “standard”, and introducing a huge amount of complexity in order to do so.

hamishwillee:
> Not much about this feels like we’re working towards a “standard” at all anymore, and the description doubles down on that feeling. It very much sounds like you’re trying to call a very custom implementation “standard”, and introducing a huge amount of complexity in order to do so.

I agree with the "vibe", but almost all of it makes sense even if you take Beat's implementation out of the picture:
- commanding a mode may fail or you might be in a mode and go into a failsafe. In this case you agreed last meeting that it was reasonable to have in the current mode information about what the mode the user wanted or actually selected.
- Being able to tag a mode as advanced was agreed several meetings ago - it is reasonable for a GCS to filter it's view on this information.
- The user settable is new, and we previously saw no value over just having advanced. I now see that it is reasonable that you might have a mode that should not be set by a GCS, as opposed to one that you would let it decide. That might be a particular fallback that only the current system might know is reasonable and possible.

I am not at all sure about whether the description should override a GCS string for a custom mode. Mostly because the mode name is not translatable. My intent was that this should be for systems that don't know the autopilot - and that they would use the flight stack if they had it.

It is reasonable to want to do modes on a companion computer and that even a standard mode might be controlled by setting setpoints from a companion computer. The only problem is that I assumed in this case the companion computer would be the source of the mode, and therefore we would be able to indicate the source was the companion.

message_definitions/v1.0/development.xml
bkueng: Fine with me.

message_definitions/v1.0/development.xml
bkueng: Yes that should be ok. PX4 doesn't use 0.

hamishwillee:
I discussed whether a value for the mode name  should over-ride the value of stored information for the modes with @ThomasDebrunner in the last call. We both felt that this probably wasn't the right way forward - the main reason being that it precludes use of translation information even if said information exists.

There are several options - for example you could have translation information for your modes for your particular vehicle as component metadata - this API does not preclude that - it just assumes that the GCS is able to recognise a standard mode and will use it's stored strings/translations for that mode. As long as the GCS is able to connect to some arbitrary flight stack and still show the correct text for a mode this is still fine.

I'd _really_ rather not, but we could also consider adding a flag to indicate that a mode is provided "offboard". 

@bkueng Thomas planned to have a chat to you about options here.

message_definitions/v1.0/development.xml
```diff
@@ -151,6 +151,21 @@
         </description>
       </entry>
     </enum>
+    <enum name="MAV_STANDARD_MODE_PROPERTY" bitmask="true">
```
hamishwillee, line 154: These aren't actually "standard mode properties". The properties of a standard mode are specified in the standard mode definition, and are otherwise up to the GCS to decide.
These are custom mode properties.

message_definitions/v1.0/development.xml
```diff
@@ -457,18 +472,25 @@
       <field type="uint8_t" name="number_modes">The total number of available modes for the current vehicle type.</field>
       <field type="uint8_t" name="mode_index">The current mode index within number_modes, indexed from 1.</field>
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
-      <field type="char[50]" name="mode_name">Name of custom mode, with null termination character. Should be omitted for standard modes.</field>
+      <field type="uint32_t" name="properties" enum="MAV_STANDARD_MODE_PROPERTY">Mode Properties.</field>
```
hamishwillee, line 476: As above, we should probably make these `MAV_CUSTOM_MODE_PROPERTY`

hamishwillee:
```suggestion
      <field type="uint32_t" name="properties" enum="MAV_CUSTOM_MODE_PROPERTY">Properties for custom modes.</field>
```

hamishwillee:
@bkueng @ThomasDebrunner We had a big discussion in [20230104-Dev-Meeting](https://github.com/mavlink/mavlink/wiki/20230104-Dev-Meeting) about this.

The original RFC had quite a simple rationale: "Define standard flight modes such that a GCS can be used with an unknown flight stack". We expect in implementation that there might be minor changes to an initial design to meet the goal. However these changes reflect a whole other set of requirements that the various stakeholders have not agreed to and might not have wanted. 

We're not rejecting this outright but we're not ready to merge this. 

I'm trying to work out offline which bits we are comfortable with and which bits we're not. Hopefully after that I can offer you clearer direction.

hamishwillee:
@bkueng , had a bit more discussion with @auturgy 

We think that these additions should be made as suggestions to the RFC with justifications/use cases that make sense in the context of the original requirement. That way we can get broad acceptance from key stakeholders - no one is surprised. 

We want to keep the complexity as low as possible. It is fine to have new stuff like "dynamic addition of flight modes" but it should be clearly separated into its own part of the proposal, and something that is clearly optional for a flight stack to implement.
I can help with that when I get back if you want.

Our last concern is testing. The work you are doing with the companion computer is exercising this in ways that are way outside our original intent. We're want to make sure that this gets tested as it will be used without a companion computer.  So probably MAVSDK integration and able to work with PX4 that has no companion.

DonLakeFlyer:
I don't quite get all of it. But as far as I can tell the key things which are important for QGC are there:
* Know that a flight mode should not be user selectable
* The concept of advanced flight modes. This allows QGC to initially hide these unless the user adds them back in. This is basically what I am doing manually in the new QGC flight mode list editing thingy. There is a list to not show initially which is bolted into the code. Onboard for example is initially hidden. I won't need to hard code that any more.

message_definitions/v1.0/development.xml
hamishwillee: Thinking more about this, the advanced properties _should_ be known for standard modes - i.e. if advanced or not. But I'm not sure future properties we might choose to support will be known in advance. So you could argue that actually we have is MAV_MODE_PROPERTIES.

bkueng: Fine with me, I can change it.

bkueng:
I rebased and removed the dynamic update for now (going through rfc)

message_definitions/v1.0/development.xml
hamishwillee: Its a mode property, applying to any modes. MAV_MODE_PROPERTY? (updated above)?

bkueng: Applied

message_definitions/v1.0/development.xml
```diff
@@ -480,18 +495,18 @@
       <field type="uint8_t" name="number_modes">The total number of available modes for the current vehicle type.</field>
       <field type="uint8_t" name="mode_index">The current mode index within number_modes, indexed from 1.</field>
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
-      <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
```
hamishwillee, line 483: @bkueng My understanding is that this is not needed to identify a mode and is independent of the custom mode, so on that basis we do not need it, because it provides no identity.

However if you have a custom mode, does it tell you anything useful you might otherwise have to query for in some other way?

For example, might you use https://mavlink.io/en/messages/common.html#MAV_MODE_FLAG to know that the mode is automatic, and would it be useful to know that about an unknown custom mode?

If we stripped it out of the heartbeat, would we want it in the CURRENT_MODE? i.e. does having it here give us a path to one day having a simpler heartbeat?

My gut feeling is no but I wanted to confirm if you see any reason to leave it.

bkueng: > For example, might you use https://mavlink.io/en/messages/common.html#MAV_MODE_FLAG to know that the mode is automatic, and would it be useful to know that about an unknown custom mode?

I did not think it's useful.

> If we stripped it out of the heartbeat, would we want it in the CURRENT_MODE? i.e. does having it here give us a path to one day having a simpler heartbeat?

That's a possibility. Given the heartbeat is such a central message, I don't consider changing it is an option, but that's not up to me.

hamishwillee: I don't think it is either. FYI, I created a PR to update RFC assuming this PR goes in https://github.com/mavlink/rfcs/pull/24

Chatting to James about it tonight hopefully.

hamishwillee:
@bkueng We discussed in the dev call. Still not happy that the simple concept of standard flight modes feels a bit hijacked. But we also can't see a better way to do what you want. So merging. I will also merge the PRs on the RFC.

message_definitions/v1.0/development.xml
```diff
@@ -480,18 +495,18 @@
...
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
-      <field type="char[50]" name="mode_name">Name of custom mode, with null termination character. Should be omitted for standard modes.</field>
+      <field type="uint32_t" name="properties" enum="MAV_MODE_PROPERTY">Mode properties.</field>
```
hamishwillee, line 499: @bkueng FMI, and sorry if I already asked, but is 32 bits too many, or just right?

hamishwillee: I guess it doesn't matter - these are only emitted on request. Ignore.

Issue #979, opened 2018-08-21
Lat lon field for sim_state should be int32_t


khancyr:
see https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L3851

llschloesser:
@khancyr, why was this closed? We would like to see this fixed. Switching to int32_t DegE7 would increase accuracy 100x, right? From roughly 1.1m to 0.01m.

khancyr:
No idea...
If you are using ArduPilot then we have 
Simstate message that got the position as uint32.

Otherwise this message will need an extension to add the uint lat lon version

hamishwillee:
@llschloesser What flight stack are you using?

Just to clarify, I think you are asking to change the lat, lon fields on [SIM_STATE](https://mavlink.io/en/messages/common.html#SIM_STATE) in common.xml? What @khancyr is saying is that the ArduPilot dialect uses (emits?) the [SIMSTATE](https://mavlink.io/en/messages/ardupilotmega.html#SIMSTATE) message which has the fields as `int32`,

The fields should be in `int32` in the "common" message (`SIM_STATE`), but we can't just change them without breaking existing deployments. Options:
- Add lat/lon values as `int32` extensions.
- Port the ArduPilot message across to common (if it is an acceptable replacement) and deprecate the current version.
- Port the ArduPilot message across to common, extending it if needed,
- completely new message
- Live with it.

One of the first three is most likely, all have challenges. Whether we do this depends on what flight stack you're working on @llschloesser - because there is no point making changes unless someone is interested in implementing them.

jasonir129:
> @llschloesser What flight stack are you using?
> 
> Just to clarify, I think you are asking to change the lat, lon fields on [SIM_STATE](https://mavlink.io/en/messages/common.html#SIM_STATE) in common.xml? What @khancyr is saying is that the ArduPilot dialect uses (emits?) the [SIMSTATE](https://mavlink.io/en/messages/ardupilotmega.html#SIMSTATE) message which has the fields as `int32`,
> 
> The fields should be in `int32` in the "common" message (`SIM_STATE`), but we can't just change them without breaking existing deployments. Options:
> 
> * Add lat/lon values as `int32` extensions.
> * Port the ArduPilot message across to common (if it is an acceptable replacement) and deprecate the current version.
> * Port the ArduPilot message across to common, extending it if needed,
> * completely new message
> * Live with it.
> 
[...]

PR #1916, opened 2022-11-10
SIM_STATE - add extension fields for latitude/longitude

hamishwillee:
This adds `int32_t/degE7` extension fields for "high precision" latitude and longitude to the `SIM_STATE` message in common. As noted in the fields, both the old and new fields should be set, but the new fields is what should be used (unless it is 0).
Note that these will not be usable on MAVLink 1 systems.

Are there any other additions that should be made to this message "while we're here"?

Fixes https://github.com/mavlink/mavlink/issues/979

FYI @peterbarker @auturgy @julianoes


hamishwillee:
Note/PS: We couldn't see any compelling reason to choose either this approach, or moving SIMSTATE from ArduPilot across and deprecating SIM_STATE. In the end we chose this way as it won't break any existing system and means that we only have one message in common with this name (reduces confusion).

message_definitions/v1.0/common.xml
```diff
@@ -5861,14 +5861,17 @@
...
+      <field type="float" name="lon" units="deg">Longitude  (lower precision). Both this and the longitude field should be set.</field>
       <field type="float" name="alt" units="m">Altitude</field>
       <field type="float" name="std_dev_horz">Horizontal position standard deviation</field>
       <field type="float" name="std_dev_vert">Vertical position standard deviation</field>
       <field type="float" name="vn" units="m/s">True velocity in north direction in earth-fixed NED frame</field>
       <field type="float" name="ve" units="m/s">True velocity in east direction in earth-fixed NED frame</field>
       <field type="float" name="vd" units="m/s">True velocity in down direction in earth-fixed NED frame</field>
+      <extensions/>
+      <field type="int32_t" name="latitude" units="degE7">Latitude (higher precision). If 0, lon field value should be used (otherwise this field is preferred)</field>
+      <field type="int32_t" name="longitude" units="degE7">Longitude (higher precision). If 0, lon field value should be used (otherwise this field is preferred).</field>
```
peterbarker, line 5874: Do we have precedent for `lat_int`, for example.  Might help to draw the difference between the fields.  I'm not really fussed, 'though,

peterbarker (approved):

message_definitions/v1.0/common.xml
hamishwillee: we don't have a precedent, probably because the type is explicit. I'm not fussed either.

message_definitions/v1.0/common.xml
```diff
@@ -5861,14 +5861,17 @@
       <field type="float" name="xgyro" units="rad/s">Angular speed around X axis</field>
       <field type="float" name="ygyro" units="rad/s">Angular speed around Y axis</field>
       <field type="float" name="zgyro" units="rad/s">Angular speed around Z axis</field>
-      <field type="float" name="lat" units="deg">Latitude</field>
-      <field type="float" name="lon" units="deg">Longitude</field>
+      <field type="float" name="lat" units="deg">Latitude (lower precision). Both this and the latitude field should be set.</field>
+      <field type="float" name="lon" units="deg">Longitude  (lower precision). Both this and the longitude field should be set.</field>
```
hamishwillee, line 5865:
```suggestion
      <field type="float" name="lon" units="deg">Longitude (lower precision). Both this and the longitude field should be set.</field>
```

hamishwillee:
@julianoes YOu OK with this? If you are, I'll merge.

khancyr (approved):
LGTM

message_definitions/v1.0/common.xml
```diff
@@ -5861,14 +5861,17 @@
...
+      <field type="float" name="lat" units="deg">Latitude (lower precision). Both this and the latitude field should be set.</field>
+      <field type="float" name="lon" units="deg">Longitude (lower precision). Both this and the longitude field should be set.</field>
       <field type="float" name="alt" units="m">Altitude</field>
       <field type="float" name="std_dev_horz">Horizontal position standard deviation</field>
       <field type="float" name="std_dev_vert">Vertical position standard deviation</field>
       <field type="float" name="vn" units="m/s">True velocity in north direction in earth-fixed NED frame</field>
       <field type="float" name="ve" units="m/s">True velocity in east direction in earth-fixed NED frame</field>
       <field type="float" name="vd" units="m/s">True velocity in down direction in earth-fixed NED frame</field>
+      <extensions/>
+      <field type="int32_t" name="latitude" units="degE7">Latitude (higher precision). If 0, lon field value should be used (otherwise this field is preferred)</field>
```
julianoes, line 5873:
```suggestion
      <field type="int32_t" name="latitude" units="degE7">Latitude (higher precision). If 0, lan field value should be used (otherwise this field is preferred)</field>
```

message_definitions/v1.0/common.xml
julianoes: In this case `lat_int` might make it clearer.

hamishwillee: OK, I will change!

message_definitions/v1.0/common.xml
```diff
@@ -5861,14 +5861,17 @@
       <field type="float" name="xgyro" units="rad/s">Angular speed around X axis</field>
       <field type="float" name="ygyro" units="rad/s">Angular speed around Y axis</field>
       <field type="float" name="zgyro" units="rad/s">Angular speed around Z axis</field>
-      <field type="float" name="lat" units="deg">Latitude</field>
-      <field type="float" name="lon" units="deg">Longitude</field>
+      <field type="float" name="lat" units="deg">Latitude (lower precision). Both this and the latitude field should be set.</field>
+      <field type="float" name="lon" units="deg">Longitude (lower precision). Both this and the longitude field should be set.</field>
```
hamishwillee, line 5865:
```suggestion
      <field type="float" name="lat" units="deg">Latitude (lower precision). Both this and the lat_int field should be set.</field>
      <field type="float" name="lon" units="deg">Longitude (lower precision). Both this and the lon_int field should be set.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5861,14 +5861,17 @@
...
       <field type="float" name="alt" units="m">Altitude</field>
       <field type="float" name="std_dev_horz">Horizontal position standard deviation</field>
       <field type="float" name="std_dev_vert">Vertical position standard deviation</field>
       <field type="float" name="vn" units="m/s">True velocity in north direction in earth-fixed NED frame</field>
       <field type="float" name="ve" units="m/s">True velocity in east direction in earth-fixed NED frame</field>
       <field type="float" name="vd" units="m/s">True velocity in down direction in earth-fixed NED frame</field>
+      <extensions/>
+      <field type="int32_t" name="latitude" units="degE7" invalid="0">Latitude (higher precision). If 0, recipients should use the lat field value (otherwise this field is preferred).</field>
+      <field type="int32_t" name="longitude" units="degE7">Longitude (higher precision). If 0, lon field value should be used (otherwise this field is preferred).</field>
```
hamishwillee, line 5874:
```suggestion
      <field type="int32_t" name="lat_int" units="degE7" invalid="0">Latitude (higher precision). If 0, recipients should use the lat field value (otherwise this field is preferred).</field>
      <field type="int32_t" name="lon_int" units="degE7" invalid="0">Longitude (higher precision). If 0, recipients should use the lon field value (otherwise this field is preferred).</field>
```

hamishwillee:
@peterbarker @khancyr FYI, @julianoes agreed with @peterbarker that using `lat_int`/`lon_int` would be less confusing so I have made that change. I also did a minor tweak to the wording (does not change meaning).

Self merging now.

PR #1917, opened 2022-11-15
Clarifying Euler angle axis order

arikrupnik:
Following discussion with Peter Barker and AndrewTridgell, a small change to the description of the ATTITUDE message to clarify Euler rotation order is ZYX.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5091,7 +5091,7 @@
       <field type="int16_t" name="temperature_press_diff" units="cdegC" invalid="0">Differential pressure temperature (0, if not available). Report values of 0 (or 1) as 1 cdegC.</field>
     </message>
     <message id="30" name="ATTITUDE">
-      <description>The attitude in the aeronautical frame (right-handed, Z-down, Y-right, X-front, ZYX order).</description>
+      <description>The attitude in the aeronautical frame (right-handed, Z-down, Y-right, X-front, ZYX, intrinsic).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="float" name="roll" units="rad">Roll angle (-pi..+pi)</field>
       <field type="float" name="pitch" units="rad">Pitch angle (-pi..+pi)</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5091,7 +5091,7 @@
       <field type="int16_t" name="temperature_press_diff" units="cdegC" invalid="0">Differential pressure temperature (0, if not available). Report values of 0 (or 1) as 1 cdegC.</field>
     </message>
     <message id="30" name="ATTITUDE">
-      <description>The attitude in the aeronautical frame (right-handed, Z-down, X-front, Y-right).</description>
+      <description>The attitude in the aeronautical frame (right-handed, Z-down, Y-right, X-front, ZYX order).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="float" name="roll" units="rad">Roll angle (-pi..+pi)</field>
       <field type="float" name="pitch" units="rad">Pitch angle (-pi..+pi)</field>


olliw42:
this is, unfortunately, nonsense. The axes of the frame have nothing to do with the order of the Euler angles, and you can list the axes in whatever order you want, doesn't matter, the axes are unambigiously defined by the additional attributes. If one would follow wikipedia they would be actually Tait-Brain angles. And whether they are intrinsic or extrinisic (in the language of wikipedia)(or passive or active in some other language)(or...). 
If I were to make a suggestion when "aeronautical" isn't considered sufficient I guess it would be "Euler angles in Z(yaw)Y(pitch)X(roll) convention (intrinsic rotations)".

message_definitions/v1.0/common.xml
```diff
@@ -5091,7 +5091,7 @@
       <field type="int16_t" name="temperature_press_diff" units="cdegC" invalid="0">Differential pressure temperature (0, if not available). Report values of 0 (or 1) as 1 cdegC.</field>
     </message>
     <message id="30" name="ATTITUDE">
-      <description>The attitude in the aeronautical frame (right-handed, Z-down, X-front, Y-right).</description>
+      <description>The attitude in the aeronautical frame (right-handed, Z-down, Y-right, X-front, ZYX order).</description>
```
hamishwillee, line 5094: I think we're talking this frame right? 

image

I don't see the current text as particular confusing if you know what the "aeronautical frame" is. You don't need to know that it is right handled, or that there is an order.

> Euler angles in Z(yaw)Y(pitch)X(roll) convention (intrinsic rotations)

I think @olliw42 suggestion is technically correct, but it isn't as easy to understand as a reader - you have to unpack what is meant by intrinsic etc.

How about we just get rid of the "right hand rule" so the order is irrelevent, and add extra axis info to make it more intuitive what we mean by the aeronautical frame?

```suggestion
      <description>The attitude in the aeronautical frame: Z-down (yaw axis), Y-right (pitch axis), X-front (roll-axis).</description>
```

@dakejahl @peterbarker Do you have an opinion?

olliw42: the problem is that this does not specify the order of the rotations/angles, it just associates them to some axes

so it could be Rz(yaw)Ry(pitch)Rx(roll) or Rz(yaw)Rx(roll)Ry(pitch) or Rx(roll)Ry(pitch)Rz(yaw) or ...

so, both the assignment of anlges to axes AND the order of the sequence matter, but the assignment to the axes does NOT imply the order of the sequence of rotations! (and it also matters in which "direction" the transformation shall be applied). That's just the mathematical situation.

so, as sad as it is, your suggestion does nothing as regards clarifying the order of angles

One can try to be "clearer" in a sense some may think is clearer, but when someone else will come by and stumple across another inclarity because the mathematical details are not fully speficied.

one really would have hoped that specifing "aeronautical" should have been fully sufficient. Probably it could be better to just strip of all except "aeronautical" and have a link to an authorative reference.

hamishwillee: @peterbarker I don't want to have this discussion - I don't understand (or care to understand) the implications of the change, or why the order of rotations matters.

Can you work with @olliw42 and @arikrupnik to come up with something mutally satisfying and accurate? If not then I will reject this PR as incorrect and not worth my time to resolve.

arikrupnik:
I added "intrinsic" as a further clarification.

arikrupnik:
Also, is it the expected result for "Node 12" and "Node 14" tests to fail?

message_definitions/v1.0/common.xml
peterbarker: > @dakejahl @peterbarker Do you have an opinion?

I think this change is a good one.  If we're going to list a set of axes - even if they're not in the same "units" as OlliW they should be in the order in which they're applied.  It would have saved one person some pain, and I can't see it causing anyone else pain.

FWIW, I hadn't heard of the term "aeronautical frame".  We tend to use "Tait-Bryan" or "3-2-1" in ArduPilot.

I think "intrinsic" I think should address @olliw42 's concern?

I think this should go in.

olliw42: I think you don't yet fully comprehended the difference between angles and axis relation ship and sequence of angles. 
frankly, "(right-handed, Z-down, Y-right, X-front, ZYX, intrinsic)" is as cryptic as it possibly can get ... it's a mystery to me how this possibly can help anyone who doesn't already know what these cryptic symbols exactly mean.
anyway, it's just my one cent, if you feel it's an improvement, I'm happy to hear that. :)

arikrupnik: @olliw42 one way this can help is by telling users what arguments to pass into third-party libraries. There are many libraries that work with Euler/Tait/Bryan angles. They need to caller to pass in rotation order and whether the rotations are intrinsic or extrinsic. If MAVLink documentation tells the user which of the 24 possible combinations are correct for the data in this message, that user can pass in the correct arguments into a math library even if the user doesn't understand the intricacies of the nomenclature.

hamishwillee:
> Also, is it the expected result for "Node 12" and "Node 14" tests to fail?

Yes. I don't know why, but it has been happening ever since the node tests were added. I'll look into it at some point.

message_definitions/v1.0/common.xml
```diff
@@ -5091,7 +5091,7 @@
       <field type="int16_t" name="temperature_press_diff" units="cdegC" invalid="0">Differential pressure temperature (0, if not available). Report values of 0 (or 1) as 1 cdegC.</field>
     </message>
     <message id="30" name="ATTITUDE">
-      <description>The attitude in the aeronautical frame (right-handed, Z-down, X-front, Y-right).</description>
+      <description>The attitude in the aeronautical frame (right-handed, Z-down, Y-right, X-front, ZYX, intrinsic).</description>
```
hamishwillee, line 5094: Does this make it less confusing?

```suggestion
      <description>The attitude in the Tait-Bryan frame (right-handed, Z-down, Y-right, X-front, ZYX, intrinsic).</description>
```

arikrupnik: @hamishwillee this makes it more confusing. I don't think anyone had a problem with "aeronautical." "Tait-Bryan" refers to angles rather than frames, and makes you look like a nit-pickier. Most people treat Tait-Bryan angles as a sub-category of Euler angles, even if they aren't proper Euler angles. "Aeronautical" means that right bank is positive and left bank is negative, etc. "right-handed, X-front" means the same thing. This is orthogonal (no pun intended) to the order in which you apply rotations, which is ZYX or (same thing) 3-2-1. You cold have an aeronautical frame and a different rotation order. One common example: 3-axis camera gimbals, many of which use the same frame but ZXY order.

hamishwillee: Thanks @arikrupnik - fair enough. This was an attempt to provide a link you could actually search on, since I could not find a reference to "aeronautical frame" on googling.

hamishwillee:
This was discussed this in the MAV call last night. It isn't perfect, but we're accepting on the basis that this is no worse (no more confusing) than the previous text, and may well help some people.

The best thing to do would be to replace this with a link to an authoratitive reference. We'll happily update to that when someone either finds one or writes one. 

This is a compromise (i.e. no one will be completely happy, but hopefully no one things we've made the world a worse place).

hamishwillee:
PS, thanks @arikrupnik and @olliw42 for engaging on this.

message_definitions/v1.0/common.xml
arikrupnik: @hamishwillee I can totally see where you're coming from. Would it that the number of Google results were an indicator of usefulness :=)

PR #1919, opened 2022-11-16
RADIO_RC_CHANNELS message

olliw42:
Suggestion for a new RADIO_RC_CHANNELS message.

The backgorund is given in the issue https://github.com/mavlink/mavlink/issues/1921.

If generally approved, the details of it may be discused here in this PR.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -400,6 +400,15 @@
         <description>Other sensor frame for target observations neither in local NED nor in body FRD.</description>
       </entry>
     </enum>
+    <enum name="RADIO_RC_CHANNELS_FLAGS" bitmask="true">
+      <description>RADIO_RC_CHANNELS flags (bitmask).</description>
+      <entry value="1" name="RADIO_RC_CHANNELS_FLAGS_FAILSAFE">
+        <description>Failsafe is active. The content of the RC channels data in the RADIO_RC_CHANNELS message is implementation dependent.</description>
+      </entry>
+      <entry value="2" name="RADIO_RC_CHANNELS_FLAGS_OUTDATED">
+        <description>Channel data may be out of date. This is set when the receiver is unable to validate incoming data from the transmitter and has therefore resent the last valid data it received.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <!-- Transactions for parameter protocol -->
@@ -505,6 +514,30 @@
       <field type="uint32_t" name="mission_checksum">CRC32 checksum of current plan for MAV_MISSION_TYPE_ALL. As defined in MISSION_CHECKSUM message.</field>
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot).
         The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+    </message>
+    <message id="420" name="RADIO_RC_CHANNELS">
+      <description>RC channel outputs from a MAVLink RC receiver for input to a flight controller or other components (allows an RC receiver to connect via MAVLink instead of some other protocol such as PPM-Sum or S.BUS).
+        Note that this is not intended to be an over-the-air format, and does not replace RC_CHANNELS and similar messages reported by the flight controller.
+        The target_system field should normally be set to the system id of the system to control, typically the flight controller.
+        The target_component field can normally be set to 0, so that all components of the system can receive the message.
+        The channels array field can publish up to 32 channels; the number of channel items used in the array is specified in the count field.
+        The time_last_update_ms field contains the timestamp of the last received valid channels data in the receiver's time domain.
+        The count field indicates the first index of the channel array that is not used for channel data (this and later indexes are zero-filled).
+        The RADIO_RC_CHANNELS_FLAGS_OUTDATED flag is set by the receiver if the channels data is not up-to-date (for example, if new data from the transmitter could not be validated so the last valid data is resent).
+        The RADIO_RC_CHANNELS_FLAGS_FAILSAFE failsafe flag is set by the receiver if the receiver's failsafe condition is met (implementation dependent, e.g., connection to the RC radio is lost).
+        In this case time_last_update_ms still contains the timestamp of the last valid channels data, but the content of the channels data is not defined by the protocol (it is up to the implementation of the receiver).
+        For instance, the channels data could contain failsafe values configured in the receiver; the default is to carry the last valid data.
+        Note: The RC channels fields are extensions to ensure that they are located at the end of the serialized payload and subject to MAVLink's trailing-zero trimming.
+      </description>
+      <field type="uint8_t" name="target_system">System ID (ID of target system, normally flight controller).</field>
+      <field type="uint8_t" name="target_component">Component ID (normally 0 for broadcast).</field>
+      <field type="uint32_t" name="time_last_update_ms" units="ms">Time when the data in the channels field were last updated (time since boot in the receiver's time domain).</field>
+      <field type="uint16_t" name="flags" enum="RADIO_RC_CHANNELS_FLAGS" display="bitmask">Radio RC channels status flags.</field>
+      <field type="uint8_t" name="count">Total number of RC channels being received. This can be larger than 32, indicating that more channels are available but not given in this message.</field>
+      <extensions/>
+      <field type="int16_t[32]" name="channels" minValue="-4096" maxValue="4096">RC channels.
+        Channel values are in centered 13 bit format. Range is -4096 to 4096, center is 0. Conversion to PWM is x * 5/32 + 1500.
+        Channels with indexes equal or above count should be set to 0, to benefit from MAVLink's trailing-zero trimming.</field>
     </message>
     <message id="435" name="AVAILABLE_MODES">
       <description>Get information about a particular flight modes.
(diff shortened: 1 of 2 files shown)


olliw42:
list of points which occured in discussions
- **message id**: The message is currently located in storm32.xml with an id in the storm32.xml's range. It doesn't smell correct to keep that id when moved to development. So I changed it. One could use any free id in the mavlink2 range as onenromally would do. In this PR I however suggest to use a gap in the mavlink1 area, which could host a number of releated messages. Personally I totally do not care, so fill happily follow any suggestion. It just needs to be decided.
- **target fields**: The message has no target fields, and it is proposed that the desired targeting is achieved by special code in the routers, similar to how the existing RADIO_STATUS is handled. This can be much debated, with pros and cons, but a decission needs to be made.
- **source system id**: The receivers sending out this message are supposed to have a source system id different from the vehicle/system they are connected to. Typically 51. This is similar to SiK radios and alike.
- The receivers are normally not supposed to send HEARTBEAT messages (unless they want to support microservices, like e.g. parameters)
 
The message originally was sugested to carry 24 channels max. I have extended to 32 channels, since due to zero-padding it doesn't cost anything but may make it super future proof.

olliw42:
@hamishwillee maybe I can direct your attention towards this PR for a moment. Last week there had been some encouraging activity on the ArduPilot side with respect of implementing the proposed RADIO_RC_CHANNELS message, and the ArduPilot devs seem to have generally agreed to the concept: https://github.com/ArduPilot/ardupilot/pull/24577#pullrequestreview-1751596027, https://github.com/ArduPilot/ardupilot/pull/24577#pullrequestreview-1751592965. While activity has lowered a bit, I think it could be the appropriate time to discuss and settle the topic of which _**message id**_ to choose.

Since besides RADIO_RC_CHANNELS there would be few more messages in the "RADIO series", it could make sense to look for slots of 5 messages or so (I currently do see 4 messages).  

The PR currently suggests a message id in the MAVLink v1 range, namely 95, as there happens to be a free slot from 93 - 99.

The canonical approach would be to occupy the "next" slot in the MAVLink v2 area, which seems to be 420 currently.

All I would want you to decide is if 95 is ok to choose, with the understanding that it essentially would reserve the slot 95-99, or if I should change the PR to use 420?

Personally I have no real opinion here, It just needs to be decided. :)

THX. Olli

hamishwillee:
IMO the range should change to something above 256. It doesn't have to be 420 - but if that looks like a good free block then sure.

olliw42:
ok, I'll change. Many thx!!

message_definitions/v1.0/development.xml
```diff
@@ -481,6 +490,16 @@
...
+    <message id="420" name="RADIO_RC_CHANNELS">
+      <description>Radio channels. Supports up to 32 channels. Channel values are in centered 13 bit format. Range is [-4096,4096], center is 0. Conversion to PWM is x * 5/32 + 1500. 
+	    Should be emitted only by components with component id MAV_COMP_ID_TELEMETRY_RADIO.
+		The message should be handled, e.g. by routers, as if it had target fields with target_system = system id of system the radio receiver is connected to and target_component = 0. 
+		</description>
+      <field type="uint8_t" name="count">Total number of RC channels being received. This can be larger than 24, indicating that more channels are available but not given in this message.</field>
+      <field type="uint16_t" name="flags" enum="RADIO_RC_CHANNELS_FLAGS" display="bitmask">Radio rc channels status flags.</field>
+      <extensions/>
+      <field type="int16_t[32]" name="channels">RC channels. Channels above count should be set to 0, to benefit from MAVLink's zero padding.</field>
+    </message>
```
peterbarker, line 502: No RSSI or LQ?  People really like having this!

peterbarker (changes requested):

message_definitions/v1.0/development.xml
olliw42: NO!!

I'm a bit inclined to direct you to the fuller discussion https://github.com/mavlink/mavlink/issues/1921, but you sure deserve a briefer response. 

It's IMHO a strength of the suggested concept to separate the rc data from the link stats data, for a number of reasons. The link stats data is carried in the proposed RADIO_LINK_STATS message, the upstream PR is https://github.com/mavlink/mavlink/pull/1920. 

The intention is clearly that once the RADIO_RC_CHANNELS thing should have been accepted, to also raise the PR to bring in the RADIO_LINK_STATS. The code is available since long. The proposed RADIO_LINK_STATS message has seen btw already substantial discussion, since it was the template for a similar DroneCAN message.
So, in short, the link stats PR will follow.

peterbarker: OK, had a quick read of the relevant section in the linked issue - that's a fair argument, thanks!

olliw42:
@hamishwillee 
FYI, the topic of this PR has made progress: 
The RADIO_RC_CHANNELS message and potential implementation in ArduPilot has been dicussed intensively over at ArduPilot, and there is now a PR in the ardupilot mavlink repository which from "our" point of view might be short of being accepted.

So, it would be great to also get your input. 

I kindly want to ask you to look over it and comment in the ArduPilot repo: https://github.com/ArduPilot/mavlink/pull/343
It would safe me from substantial double-burden - once the message has been merged to the ArduPilot mavlink repo I will update this PR here to be in sync, so that the two repos don't stray away. Hope you find this ok.

The main changes to the message proposed here had been
- adding target fields
- adding a time stamp for the when the rc data was valid the last time
- more comments in the description

THX

message_definitions/v1.0/development.xml
```diff
@@ -481,6 +490,16 @@
...
         The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
     </message>
+    <message id="420" name="RADIO_RC_CHANNELS">
+      <description>Radio channels. Supports up to 32 channels. Channel values are in centered 13 bit format. Range is [-4096,4096], center is 0. Conversion to PWM is x * 5/32 + 1500. 
+	    Should be emitted only by components with component id MAV_COMP_ID_TELEMETRY_RADIO.
+		The message should be handled, e.g. by routers, as if it had target fields with target_system = system id of system the radio receiver is connected to and target_component = 0. 
+		</description>
+      <field type="uint8_t" name="count">Total number of RC channels being received. This can be larger than 24, indicating that more channels are available but not given in this message.</field>
+      <field type="uint16_t" name="flags" enum="RADIO_RC_CHANNELS_FLAGS" display="bitmask">Radio rc channels status flags.</field>
+      <extensions/>
```
hamishwillee, line 500: Given the id is changing anyway, the message is breaking. So we don't need extensions.

hamishwillee:
```suggestion
```

message_definitions/v1.0/development.xml
```diff
@@ -481,6 +490,16 @@
       <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot).
         The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
     </message>
+    <message id="420" name="RADIO_RC_CHANNELS">
+      <description>Radio channels. Supports up to 32 channels. Channel values are in centered 13 bit format. Range is [-4096,4096], center is 0. Conversion to PWM is x * 5/32 + 1500. 
+	    Should be emitted only by components with component id MAV_COMP_ID_TELEMETRY_RADIO.
+		The message should be handled, e.g. by routers, as if it had target fields with target_system = system id of system the radio receiver is connected to and target_component = 0. 
+		</description>
+      <field type="uint8_t" name="count">Total number of RC channels being received. This can be larger than 24, indicating that more channels are available but not given in this message.</field>
```
hamishwillee, line 498: ?
```suggestion
      <field type="uint8_t" name="count">Total number of RC channels being received. This can be larger than 32, indicating that more channels are available but not given in this message.</field>
```

hamishwillee: Argg. I'll look on the one on ArduPilot.

hamishwillee:
@olliw42 Re https://github.com/mavlink/mavlink/pull/1919#issuecomment-1914268242 I have put in some comments, which hopefully make sense. The little picture is
- Extensibility - how many channels
- How does the timestamp and RC-lost behaviour work

The big picture is "can this be a single path for analog manual control input" with Joystick. Essentially addressing James comments in https://github.com/mavlink/mavlink/issues/1921#issuecomment-1331303555

Other than that, we probably need review from someone like @ThomasDebrunner and @julianoes - but let's see if my questions make any sense to you first :-)

olliw42:
MANY thx for looking at it
I'll comment in the ArduPilot PR https://github.com/ArduPilot/mavlink/pull/343#issuecomment-1918417134, if you allow.

hamishwillee:
Of course. Looks like broader interest :-)

olliw42:
@hamishwillee brings in the changes to RADIO_RC_CHANNELS discussed in https://github.com/ArduPilot/mavlink/pull/343

also removes all other related enums/messages from the storm32.xml, as it seems it's all going to come :)

olliw42:
ups, error ... it seems you had added minValue/maxValue fields to the message which are in ardupilot's repo but not yet in the schema here ... what to do? I guess you need to update the schema ...

olliw42:
@hamishwillee removed the offending minValue,maxValue attributes, so that this message can be merged as soon as possible, I think the minValue,maxValue can be reinserted once the schema is done.

hamishwillee:
The schema should have come back in today. If you can add back I'll merge tomorrow.

olliw42:
done :)

olliw42:
THX!

PR #1920, opened 2022-11-16
RADIO_LINK_STATS message

olliw42:
Suggestion for a new RADIO_LINK_STATS message.

The backgorund is given in the issue xxx.

If generally approved, the details of it may be discused here in this PR.


David-OConnor:
Status?

hamishwillee:
I can't remember the precise discussion on this PR. One thing MAVLink does need is a stakeholder - someone who will take the time to implement. We get a lot of great ideas but what we need is prototyping and validation.

> It sure would be great if both the MAVLink and the DroneCAN versions would be similar in that they both could cover the same range of applications and cases. I think it's more like learning from each other than actual code implementation benefits. I mean, most of the work will be specific to the respective protocols, and it's not so difficult anyway.

Yes. The fact that Dronecan are working on this is a good reason to coordinate. We did so with the battery API v2 with some success (though still no implementations).

I'll add this back to list to discuss in next week's dev call.

olliw42:
@hamishwillee 

> One thing MAVLink does need is a stakeholder - someone who will take the time to implement  ...  prototyping and validation.

with all due respect, the first part and second part does not match. There is a very substantial difference between "someone" and a "stakeholder". The "someone" who is doing or has done the implementation, prototyping and validation is often actually available - and this is so especially in this case here! 

Moreover, in this case, the case is so obviously well founded in terms of that it is so much desired by a substantial (albeit obviously largely irrelevant) user base, but it's the stakeholder(s) which seem not so much interested.  

So, pl, don't blurr the situation. Describing it as "we get a lot of great ideas" is totally inappropriate. I really would wish that the facts of a situation would more often been better reflected. 

As a side note, since such comments really can make one wonder if some basic principles are understood by the stakeholders and their surrounding. This habit and attitude can in fact lead exactly to that these "great ideas" do never materialize or take ages to materialize. I can provide you with more than one example for this phenomenon.

message_definitions/v1.0/development.xml
```diff
@@ -373,6 +379,24 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
       <field type="uint32_t" name="checksum">CRC32 checksum of current plan for specified type.</field>
     </message>
+    <message id="96" name="RADIO_LINK_STATS">
+      <description>Radio link statistics. Should be emitted only by components with component id MAV_COMP_ID_TELEMETRY_RADIO. Per default, rssi values are in MAVLink units: 0 represents weakest signal, 254 represents maximum signal, but can be changed to dBm with the flag RADIO_LINK_STATS_FLAGS_RSSI_DBM.</description>
```
hamishwillee, line 383: Is it possible/desirable to settle on one -RSSI or dBm? My assumption is that links might emit either or both, but not guaranteed and that it isn't possible to convert them and just emit one type?

Either way, we should be clear here that if dBm, the values are negative up to 0.

message_definitions/v1.0/development.xml
```diff
@@ -373,6 +379,24 @@
...
+      <field type="uint8_t" name="flags" enum="RADIO_LINK_STATS_FLAGS" display="bitmask">Radio link statistics flags.</field>
+      <field type="uint8_t" name="rx_LQ" units="c%" invalid="UINT8_MAX">Values: 0..100. UINT8_MAX: invalid/unknown.</field>
```
hamishwillee, line 385: How is this defined? I.e. is it RSSI squished from 0 to 100 or is it some measure of packet loss, or what?

message_definitions/v1.0/development.xml
```diff
@@ -373,6 +379,24 @@
...
+      <field type="uint8_t" name="rx_rssi1" invalid="UINT8_MAX">Rssi of antenna1. UINT8_MAX: invalid/unknown.</field>
```
hamishwillee, line 386: The setup I _think_ I see below is a possible setup where
- you have a radio unit defined as rx_ or tx_, which I presume are either end of the communications channel?
- each radio unit might have two antenna, and if they have two then one is used for Tx and one is for Rx.

If so:
- Which unit sends this - rx_ or tx_or both
- How do they know they are tx_ or rx_?
- If the one component sends stats for both units, how does it get info from remote unit?

message_definitions/v1.0/development.xml
olliw42: I choose to have the two options "mavlink 0...254" and "dBm" with dbm positive for these reasons
- not every system is able to deliver realistic or good dBm values. In that cases it is really better to have a range instead of having to constantly mock up some fake dbm rssi values (one just has to go through some codes to see what mockery has been done here in the past)
- "mavlink 0...254" is what is currently used in many places, and thus provides a good level of backwards compatibility and maybe more relevant ease of migration
- it is really just a matter of "taste" and definition if to use negative-to-0 or positive-to-0 for representing dbm. TBS e.g. and all EdgeTx/OpenTx radios which are used with crsf protocol (with rc systems like Crossfire,Tracer,ELRs,Ghost) will get rssi in positive values! (i.e. a rssi of -123 dBm will be displayed as 123 in the telemetry) So, systems are widespread where your requirement "_if dBm, the values are negative up to 0_" is not respected anyway.
- with using positive-to-0 for dBm, both "mavlink 0...254" and "dBm" can be represented in the same type of integer, uint8.
- with using positive-to-0 for dBm, both "mavlink 0...254" and "dBm" can be indicated to be invalid by the same value, 255. Otherwise, it would have to be 0 for "dBm", and still 255 for "mavlink 0...254"

message_definitions/v1.0/development.xml
olliw42: LQ has become extremly popular since TBS/CrossFire introduced it, and in the communities I'm around LQ is IMHO essentially nearly always looked at.

I'm not sure it really needs a definition, except that it should reflect the integrity of the link in percentage. And I'm not sure there is a universaly agreed on defeinition.

However, it appears that the majority would probably define it as "percantage of number of valid packets received over a period of time"

To just illustrated the formula, e.g. with a frame rate of 100 Hz, receiving 68 valid packets in a time period of 1 sec would translate to a LQ = 68%. So, it's a statment on packet loss. But it's not necessary it's 1 sec. There is a bit of controversy in the "community" if it should be determined by a moving average (i.e. always look back the given time period from the current time) or for time slices. But all these details IMHO do not need to, or should not, be specified. The concept of a LQ as indicator of percantage of lost packets is  somewhat clear.

I fully agree though that one should leave some hint in the description, like "percantage of number of valid packets received"

message_definitions/v1.0/development.xml
olliw42: The strategy here diverts from the old RADIO_STATUS's strategy, where the meaning of the fields swap depending on which side of the link one is. 

Here the idea is that there is a receiver and a transmitter. It can be uniquely identified by the flow of the rc data: The unit which send the rc data is the transmitter and the unit which is receiving the rc data is the receiver.

With that:
- all rx_xxx fields give info on the receiver side, 
- all tx_xxx fields give info on the transmitter side.

If the naming of the fields is inconvennient, I'm of course very open to suggestions for renaming. But I think this strategy is much less confusing than the old RADIO_STATU's strategy. I'm however also open to do it as for RADIO_STATUS, it would need just slight change in the naming. But as said, I think as is it removes a good chunk of confusion.

> Which unit sends this - rx_ or tx_or both

the unit sending rc data is "tx" and teh unit receiving rc data is "rx"

> How do they know they are tx_ or rx_?

They know if they are sending rc data or recieving rc data

> If the one component sends stats for both units, how does it get info from remote unit?

that's a question of how that link works and a technical design question. Today the majority of links have some sort of bidirectional communication, and be it only very low data rate in one direction, and use that capability to feed back some link stats to the orther side. The details are the "secret" of the link manufacturer.

olliw42: > each radio unit might have two antenna, and if they have two then one is used for Tx and one is for Rx.

a unit which has two antenna may or may not split them into one is for Tx and the other one for Rx. 
Usually, at least in nearly all cases I'm aware of, they both would be used to e.g. receive, which is when called receive diversity. Note that many bidirectional units are half-duplex, i.e. work as transmit-receive-transmit-receive-.... so that both antenna can be used for receiving one frame at the same time.
I guess the system you have in mind is a system there two over-the-air channels/frequencies are used, in a full-duplex sense, there one channel/frequency is for transmission in one direction and the other for transmission in the other direction. As said, essentially all systems I'm aware of are of the half-duplex nature.
Anyway, also the full-duplex case can be represented by the suggested fields.

message_definitions/v1.0/development.xml
hamishwillee: Thanks very much!

message_definitions/v1.0/development.xml
hamishwillee: Sounds good. I have mapped this in proposal to Dronecan:

> Link quality, as a percentage of valid packets received over some time period [1..100]. 256 means data is unavailable.

OK?

message_definitions/v1.0/development.xml
hamishwillee: Thanks @olliw42 

So you're using the old convention from RC transmitters of  the old days, where the ground unit was always a pure transmitter and the vehicle was a dumb receiver. I've always found it a bit confusing for new users to continue to name the comms modules on the direction of RC signal, when they might be able to transmit video etc bidirectionally.

I completely agree that this is better than the old status message, and it is a valid convention. However I think it is extra confusing because you end up with names like `rx_transmit` and so on.

Dronecan are proposing `uplink` and `downlink` instead of `tx`, `rx`, which to my mind results in easier to parse names.
However we should also consider `ground/air` or `base/rover` or `base/remote` or similar. Thoughts?

FYI dronecan use the unit "received signal strength, in decibel-milliwatts (RSSI dBm) for antennas 1 and 2 respectively.  Ranges from # -255 to 0. (An integer value of 100 means RSSI is -100dBm)"
- as I understand it you're using 0 to 255 in both cases with max value for invalid.

With respect to the antenna configuration, I'm just trying to understand what is possible. 

My confusion based on your answer is how you would indicate, say, that the both antenna on Rx are active and receiving?

olliw42: well, naming is always an topic of discussion

I really totally "hate" uplink, downlink ... it is used by TBS and it seem that not only me gets confused but I think mixups are in some prominent codes. If it's bidirectional, what the heck is then up and down? I just can't see how that's a useful notation. But that's my personal limitation obviously. I don't like the other suggestions a lot. 

If I look around, it's actually nearly always obvious what is "tx" and "rx" from the hardware side. The one notable exception are the old SiK radios, which are totally symmetric, hardware wise but in fact also firmware wise. Systems like DragonLink, RFD900, Herelink, and all rc systems, for all it's obvious what tx is (and many of them use language like e.g. "tx module"). So, yes, the rx/tx language is not totally appropriate for all cases, but this would be so too for a uplink/downlink etc. pp notation

> FYI dronecan use the unit "received signal strength, in decibel-milliwatts (RSSI dBm) for antennas 1 and 2 respectively. Ranges from # -255 to 0. (An integer value of 100 means RSSI is -100dBm)"

the text confuses me a bit. It is saying that dronecan also does represent rssi by its negative value?

>  I understand it you're using 0 to 255 in both cases with max value for invalid.

yes. To be accurate rssi is using 0..254 with 255 for invalid

> My confusion based on your answer is how you would indicate, say, that the both antenna on Rx are active and receiving?

if only one antenna is active, then all antenna2 fields would be set all to invalid. So, you can determine if two antenna are active by the fact that at least one of the antenna2 fields carries valid data. IMHO this makes sense.

Alternatively, one could try to represent the hardware by some flags, like HAS_RECEIVE_DIVERSITY, HAS_TRANSMIT_DIVERSITY.

olliw42:
@hamishwillee 

many thanks for putting efforts into this. I feel I should make one point clear however, to avoid potential misunderstandings:
I am totally NOT interested in all this and this message here unless (i) also the RADIO_RC_CHANNELS message would be merged and (ii) the probability that these two messages are supported in ArduPilot in a reasonably short time frame approaches 100%. I don't have any indication these would be met, so pl don't be surprised that I wouldn't get excited if this PR here would make progress. AFAICS we are having this discussion here only to help making the dronecan message "better" :)

message_definitions/v1.0/development.xml
olliw42: maybe, maybe not. Just peeked into the dronecan thread and found that the start with 1 is to use 0 for invalid. If it starts with 1 it should go [1...101], shouldn't it? if it starts with 0, it shoudl be [0...100]. 256 can be represented in droncan with a  uint8?
but that's dronecan now LOL

hamishwillee: Yeah, I'll work to make them MAVlinky to match you. To this point, I've just been trying to understand why you have designed it this way

hamishwillee:
@olliw42 Understood, and we're both aware that what you need is unlikely to happen.

It is to some extent because of that that I am so grateful for your advice. I'm personally not hung up on getting this in either. I just want to make sure that what happens with Dronecan does not preclude this work going in.

PR #464, opened 2015-11-27
More detail for MANUAL_CONTROL.z forward and reverse thrust values

DonLakeFlyer:

PR #1922, opened 2022-11-19
MANUAL_CONTROL: Redefine the use of `z` (throttle) to be independent to vehicle's thrust direction

junwoo091400:
## Description
Previously, the `z` component definition relied on having a knowledge of which direction the thrust would result in, depending on its value, which is tied to the vehicle's setup (e.g. Rover, which can take in reverse-throttle commands vs. conventional quadcopter with only positive throttle).

This is a bad practice because:
1. With this definition, we need to tailor the `MANUAL_CONTROL` message for each different type of vehicle to use the 'correct' range of `z`. So having a GCS with a Joystick input will have to know which range of throttle would result in positive/negative thrust for each connected vehicle, which doesn't make sense.
2. The concept of 'Manual Control' should be as simple as possible (just a Joystick command). And just how Joysticks all have min/max ranges, it would be simpler to not have the 'thrust' condition, but to just map the ranges simply to [-1, 1]

With this new definition, the MANUAL_CONTROL message will not be vehicle-specific and truly represent the Joystick / RC Transmitter's throttle position.

## Alternatives
We could keep the convention as is 🤔 

## Discussion Points
1. How do we make sure previous users who were sticking to this convention don't get affected? (Or is this inevitable?)

## Context
This issue was discussed in a PX4 PR trying to standardize the range for throttle internally: https://github.com/PX4/PX4-Autopilot/pull/15949#discussion_r1025243016


message_definitions/v1.0/common.xml
```diff
@@ -5448,7 +5448,7 @@
...
       <field type="int16_t" name="y" invalid="INT16_MAX">Y-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to left(-1000)-right(1000) movement on a joystick and the roll of a vehicle.</field>
-      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to a separate slider movement with maximum being 1000 and minimum being -1000 on a joystick and the thrust of a vehicle. Positive values are positive thrust, negative values are negative thrust.</field>
+      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally corresponds to throttle high (1000) - throttle low (-1000) on a joystick. Centered throttle should result in value of 0 (center).</field>
```
MaEtUgR, line 5451:
```suggestion
      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Joystick or slider fully down corresponds to -1000, fully up to 1000, the center of the range to 0.</field>
```

MaEtUgR (approved):
Most importantly remove:
> Positive values are positive thrust, negative values are negative thrust.

That part causes ambiguity and confusion because what happens if the vehicle does not have any negative thrust? In practice, this message was designed and is broadly used to report joystick positions over MAVLink. But with this definition, it instead results in the MAVLink-enabled remotes needing to understand the thrust situation of the vehicle just to be able to report the throttle stick or slider position.

It's sad that the original easier to understand definition was abandoned in https://github.com/mavlink/mavlink/pull/64/files
but
https://github.com/mavlink/mavlink/pull/464/files
really started the confusion in my eyes.

I'd vote for a very clear definition like this:
https://github.com/PX4/PX4-Autopilot/blob/7977ba3217422b042a6848ef529d9ea61ba0b686/msg/ManualControlSetpoint.msg#L24-L27
but for this message, I think the ship has sailed and a new one would need to be added for backwards compatibility.

message_definitions/v1.0/common.xml
hamishwillee: I prefer this ^^^

hamishwillee: Not sure about this. Before we were saying its another slider, now we're enforcing that the alignment of that slider is "up/down". I think the desired change is just to remove the emphasis on the positive/negative thrust. Something like.

```suggestion
      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Generally the axis is mapped to vehicle thrust. Vehicles that support only positive thrust map the range from maximum thrust (1000) to mininimum (-1000). Vehicles that support negative thrusts map the range from maximum positive thrust (1000) to maximum negative thrust (-1000), with 0 (centred) implying no thrust).</field>
```

hamishwillee:
As the wording is now "Positive values are positive thrust, negative values are negative thrust.". That implies that on a vehicle with only positive thrust you have to have a deadzone from the low position of the axis range to the centre. I bet that no one actually does that.

This is conceptually a breaking change from a MAVLink perspective but I think it makes sense, if we can get the wording right. I will take to the dev call this evening.

@MaEtUgR 

1. I don't like the "[simpler definition"](https://github.com/PX4/PX4-Autopilot/blob/7977ba3217422b042a6848ef529d9ea61ba0b686/msg/ManualControlSetpoint.msg#L24-L27). Refering to pitch, roll, yaw, throttle thrust is annoying unless that is how it maps to all vehicles. For example, pitch makes little sense in MC position mode right? That has always made our discussions of stick movements confusing. But, I'm not the boss of MAVLink and you know more than me about the practicalities of interpreting joysticks. 
2. But if you want to explore this, we are already thinking of how to [iterate this message](https://mavlink.io/en/services/manual_control.html#future-extensions). We haven't really thought about the axis. The main concern is that currently we have a message that provides button mappings and the flight stack has to map those button mappings to functions  - that's a lot of code in the flight stack, and requires that the GCS knows quite a bit about the flight stack it is taking to. What we're thinking about is removing the button array and having the ground station just send mavlink commands on button presses.

peterbarker:
> 2\. But if you want to explore this, we are already thinking of how to [iterate this message](https://mavlink.io/en/services/manual_control.html#future-extensions). We haven't really thought about the axis. The main concern is that currently we have a message that provides button mappings and the flight stack has to map those button mappings to functions  - that's a lot of code in the flight stack, and requires that the GCS knows quite a bit about the flight stack it is taking to. What we're thinking about is removing the button array and having the ground station just send mavlink commands on button presses.  

Oddly, I'm looking at doing the opposite in (non-Sub) ArduPilot (https://github.com/ArduPilot/ardupilot/pull/22259) - actually taking action based on the buttons field in `MANUAL_CONTROL`.  ArduPilot has a set of pre-canned "auxiliary functions" which I'll be using to provide the button functions - we use that list for physical buttons present on the vehicle, RC switches, mission items, lua script and direct mavlink calls, so reusing that list is making sense so far.

Turns out that MissionPlanner actually already does what you're suggesting here - makes mavlink calls rather than passing through the button bits (a small impediment to my little project).

hamishwillee:
> Turns out that MissionPlanner actually already does what you're suggesting here - makes mavlink calls rather than passing through the button bits (a small impediment to my little project).

Sounds daft unless the functions you want don't have corresponding mavlink messages, or can't do so for some reason. I understand that in Canberra they call you "Mad Peter" :-)

Currently 
- ArduPilot has parameters that map what buttons are, and GCS has to know what those parameters are
- PX4 hard codes specific buttons in the message to specific functions, so the GCS needs to know what functions are used for each case and send that set of values in the message.

The send-a-mavlink-message approach seems much smarter and more flexible. 
- The GCS does not need to know anything about the flight stack than it already knows - what commands are supported. (and even those can be in many cases checked without side effects). It doesn't need to know what parameters the flight stack uses to map button presses or what functions are called. 
- The flight stack doesn't have to have special code paths to process instructions it is already processing through mavlink.
- The manual control message doesn't have to cope with the possiblity of joysticks with more buttons than we expose.

message_definitions/v1.0/common.xml
MaEtUgR: > Before we were saying its another slider

Conceptually it should represent an analog input axis be it slider, stick axis or whatever else someone would like to map. In most cases, it will be a stick axis which might be worth mentioning but certainly not enforcing.

> now we're enforcing that the alignment of that slider is "up/down"

My wording is not general enough. I do not want to enforce that. The axis just needs to have a minimum and maximum if that's up, down, sideways or diagonal is up to the user. Again in most cases, I'm pretty sure it's up and down.

> `Generally the axis is mapped to vehicle thrust.`

That is exactly the part that is in my eyes only half the story. Manual up, forward, gas is the first thing that comes to mind but often it's mapped to steer vertical or even horizontal speed. I'd try to avoid mentioning an exact use because even though it's true in certain scenarios it can be very confusing in others.

> `Vehicles that support only positive thrust map the range from maximum thrust (1000) to mininimum (-1000). Vehicles that support negative thrusts map the range from maximum positive thrust (1000) to maximum negative thrust (-1000), with 0 (centred) implying no thrust).`

Very good wording. Whatever the thrust, speed or other desired quantity range is, should be mapped to the full available input range -1000 to 1000.

peterbarker:
On Tue, 29 Nov 2022, Hamish Willee wrote:

> The send-a-mavlink-message approach seems much smarter and more flexible.
>
>  *  The GCS does not need to know anything about the flight stack than it already knows - what commands are supported. (and even those can be in many cases checked without side effects). It doesn't need to
>     know what parameters the flight stack uses to map button presses or what functions are called.
>  *  The flight stack doesn't have to have special code paths to process instructions it is already processing through mavlink.
>  *  The manual control message doesn't have to cope with the possiblity of joysticks with more buttons than we expose.

That's all true, and we did note the first two when we discussed.

In this case one of our partners wanted to have fence enable/disable on a 
button.  MissionPlanner didn't permit binding that to a button, and would 
need patching and updating to support the functionality.  Having the 
function instantly available was what I was after - as soon as it was 
available in ArduPilot it could be configured.

I would *like* MissionPlanner to use our DO_AUX_FUNC list (I was mistaken 
in my initial assesment there; it has specialised code for each function 
it offers).  Addition of an arbitrary-number-combo-box interface into the 
relevant configuration page might fix things.

Problem with that last is that now *every* GCS has to have this interface. 
The list of valid functions will be different autopilot-to-autopilot.... 
and while ArduPilot has a convenient list of functions that can be 
activated by MAVLink (I might actually generate a aux_func.xml at some 
stage), other autopilots may not...

There are advantages both ways for storing the configuration in the GCS 
and on the vehicle; one allows you to switch laptops and know the config 
will be the same.  The other allows the interface to change between 
laptops based on input devices....

hamishwillee:
@peterbarker There was an orthogonal discussion here https://github.com/mavlink/mavlink/pull/1848 which looks at the use case of a MAVLink joystick. This is a joystick that does not need to be connected to a GCS when used. 

The idea was that a joystick would have a component metadata mechanism to configure its buttons a lot like the mavlink camera API - i.e. you'd query the joystick for its axes and buttons and get metadata back on what is supported. The buttons would have set options that map to parameter values - i.e. button 1 might have param_button_1 on the joystick with options "enable fence", "kill switch". 
So you connect to a GCS once, use the config API to select the things you want on each button based on the options it supplies. The joystick then sends its messages direct to the autopilot node on the network (or other component ID) and when a button is pressed sends the appropriate MAVLink command.

My point is that the sending mavlink command mechanism is more flexible here too. Component information helps with the problem of knowing what the joystick can do, and means that the autopilot doesn't need to have any extra config to work.

It's just an idea at this time.

hamishwillee:
Anyway, appreciate that you're probably making a pragmatic decision based on what mission planner does now (and PX4). Nothing is going to change in any sort of short timeframe so you can do what you like. But if you go with sending mavlink messages that will work whatever is done in future with the MANUAL_CONTROL message.

message_definitions/v1.0/common.xml
hamishwillee: > Whatever the thrust, speed or other desired quantity range is, should be mapped to the full available input range -1000 to 1000.

@MaEtUgR No, that's good wording ^^^ :-)

Maybe?

```suggestion
      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. The value may be mapped to thrust, speed, throttle, or any other property.
      The available output range of the property should be mapped to the full range of the axis (-1000 to 1000).
      Properties that can only be positive are still normalized from minimum=-1000 to maximum=1000 across the whole range of the axis.
      Properties that can be both negative and positive map the range more naturally: from negative minimum (-1000) to 0 (centred), to positive maximum (1000).</field>
```

MaEtUgR: Coming back to this in 2025 after https://github.com/PX4/PX4-Autopilot/pull/24576, which is a product of my suggested consistent use of the throttle/z range and the existing inconsistent use. I might slowly lean towards a completely new message defining the input in a specific consistent way because as much as I'd like to correct `MANUAL_CONTROL` I'm not sure how many use cases break if we clarify the definition and use it consistently 😬
If there's another easy way to ensure the transition is smooth, I'm eager to learn.

hamishwillee: @MaEtUgR There's some appetite for an improve message here. The best way forward would be to create a PR that outlines the messages you'd like to see. I can then add to the dev call.

hamishwillee:
Discussed this in the [20250410](https://github.com/mavlink/mavlink/wiki/20250410-Dev-Meeting) dev call. This was explained to me as follows (capturing so I don't have to ask this all again!)

1. [`MANUAL CONTROL`](https://mavlink.io/en/messages/common.html#MANUAL_CONTROL) is defined as a mapping of x, y, z, axes (FRD) with a statement that these generally map to pitch, roll, thrust.
2. The intent of the design was that these actually map to an acceleration in each of the axis. The x, y, z are mapped as intended (to pitch, roll, thrust) on a multicopter vehicle.
3. However this mapping is not observed for other vehicle types - the z axis is always mapped to thrust, which on a fixed wing vehicle would be forwards. So there is an inconsistency across vehicles.
4. PX4 and ArduPilot both do this "wrong thing". PX4 maps directly to the thrust setpoint, and ArduPilot maps this through an RC_CHANNEL_OVERRIDE (Used by ArduSub).
5. This "works" because QGC is the only thing that sends this, so it is consistent. 

Upshot I am not sure how to change the messages but if we change anything we break everything.

I'm not sure how we'd fix this - presumably a similar message where the definitions are as intended by `MANUAL_CONTROL` and that is then implemented "properly".

@julianoes Is that about right? Feel free to edit this if I misunderstood!

julianoes:
> (FRD) with a statement that these generally map to pitch, roll, thrust.

That's debatable. The naming would indicate this but the way it is described and implemented does not.

> Upshot I am not sure how to change the messages but if we change anything we break everything.

They way how I read this is that @junwoo091400 does not actually want to change/break anything but merely document how it is actually used and implemented in QGC.

message_definitions/v1.0/common.xml
hamishwillee: > I'm not sure how many use cases break if we clarify the definition and use it consistently

@MaEtUgR It looks like QGC is the only supplier, and both PX4 and ArduPilot map this "to the same effect". What this means is that we can really only clarify the definition to "what QGC does". We should do that and nothing else here.

peterbarker:
Ping @Williangalvani

junwoo091400:
So

QGC
* [Sends](https://github.com/mavlink/qgroundcontrol/blob/12efa39090385df4cec9ece7269b4d6b2a8bb811/src/Joystick/Joystick.cc#L654): [0, 1000] or [-1000, 1000] depending on vehicle type ([PX4](https://github.com/mavlink/qgroundcontrol/blob/12efa39090385df4cec9ece7269b4d6b2a8bb811/src/FirmwarePlugin/PX4/PX4FirmwarePlugin.cc#L716): Rover + Submarine, [Ardupilot](https://github.com/mavlink/qgroundcontrol/blob/12efa39090385df4cec9ece7269b4d6b2a8bb811/src/FirmwarePlugin/APM/ArduRoverFirmwarePlugin.h#L48): Rover)
* Receives: [x] (Only shows it in MAVLink Inspector)

PX4
* [Sends](https://github.com/PX4/PX4-Autopilot/blob/937998b739b9ffa0b2dc1aba97d20096a06b45af/src/modules/mavlink/streams/MANUAL_CONTROL.hpp#L75): [0, 1000]
* [Receives](https://github.com/PX4/PX4-Autopilot/blob/937998b739b9ffa0b2dc1aba97d20096a06b45af/src/modules/mavlink/mavlink_receiver.cpp#L2099): [0, 1000]

Ardupilot
* Sends [x] (Not in [streamed messages list])(https://github.com/ArduPilot/ardupilot/blob/b0a961ce179772ae343159d2d494ea81ceda0c4f/libraries/GCS_MAVLink/ap_message.h#L75))
* Receives [-1000, 1000] ([Rover](https://github.com/ArduPilot/ardupilot/blob/b0a961ce179772ae343159d2d494ea81ceda0c4f/Rover/GCS_MAVLink_Rover.cpp#L644)) or [0, 1000] ([Plane](https://github.com/ArduPilot/ardupilot/blob/b0a961ce179772ae343159d2d494ea81ceda0c4f/ArduPlane/GCS_MAVLink_Plane.cpp#L1044), [Submarine](https://github.com/ArduPilot/ardupilot/blob/b0a961ce179772ae343159d2d494ea81ceda0c4f/ArduSub/joystick.cpp#L125), Copter: Will even [reject all messages with MANUAL_CONTROL.z with negative value](https://github.com/ArduPilot/ardupilot/blob/b0a961ce179772ae343159d2d494ea81ceda0c4f/ArduCopter/GCS_MAVLink_Copter.cpp#L899)!) depending on the vehicle

Thus, it seems fairly customized per vehicle type, and the unification of the definition won't be trivial (expected range already differs per vehicle in Firmwares), and will lead breaking changes (e.g. [-1000, 1000] range will be rejected for Arducopter, PX4, etc)

So I think there isn't really a solution here, and the best we could do is just to clarify this difference of MANUAL_CONTROL.z per vehile type in the MAVLink message definition, and support this legacy implementation.

Thoughts?
---
p.s. It's interesting that QGC originally allowed users to choose whether RC transmitter's throttle channel is spring loaded or not, but now that logic isn's used AFAIK (assumes not spring loaded): https://github.com/mavlink/qgroundcontrol/pull/1827

message_definitions/v1.0/common.xml
junwoo091400, line 5451:
```suggestion
      <field type="int16_t" name="z" invalid="INT16_MAX">Z-axis, normalized to the range [-1000,1000]. A value of INT16_MAX indicates that this axis is invalid. Due to legacy reasons, [0, 1000] is used for PX4, ArduCopter, ArduPlane and ArduSub, whereas [-1000, 1000] is used for ArduRover. However originally, it generally corresponds to throttle high (1000) - throttle low (-1000) on a joystick. Centered throttle should result in value of 0 (center).</field>
```

Williangalvani:
I stand with @junwoo091400's latest suggestion.

Many times I thought about ways of remapping ArduSub to [-1000,1000], but I never got to a solution that wouldn't brake things.
Given that most vehicle are using the [0,1000] range, I vote we just document it properly.

julianoes:
Without having read everything, so how do you go backwards?

hamishwillee:
This is never going to happen right? Can we close it?

I'd take a PR for a new set of MANUAL_CONTROL  messages that address all our concers.

PR #1923, opened 2022-11-22
Gimbal Device: Add error flag no gimbal manager

olliw42:
This adds a flag to the gimbal device error flags which indicates that the gimbal device is not assigned to a gimbal manger.

This is useful to check/see if the system is healthy (operational). It is not strictly an error, like calibration is not strictly an error, but like calibration it is a mal condition. Note that gimbal devices must not be used except by the gimbal manager, which implies that there must be such a gimbal manager and that the gimbal device must have discovered its gimbal manager.

Such as flag is also in the storm32.xml protocol, and was found to be useful.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -571,7 +571,7 @@
         <description>There is an error with the gimbal power source.</description>
       </entry>
       <entry value="32" name="GIMBAL_DEVICE_ERROR_FLAGS_MOTOR_ERROR">
-        <description>There is an error with the gimbal motor's.</description>
+        <description>There is an error with the gimbal motors.</description>
       </entry>
       <entry value="64" name="GIMBAL_DEVICE_ERROR_FLAGS_SOFTWARE_ERROR">
         <description>There is an error with the gimbal's software.</description>
@@ -580,7 +580,10 @@
         <description>There is an error with the gimbal's communication.</description>
       </entry>
       <entry value="256" name="GIMBAL_DEVICE_ERROR_FLAGS_CALIBRATION_RUNNING">
-        <description>Gimbal is currently calibrating.</description>
+        <description>Gimbal device is currently calibrating.</description>
+      </entry>
+      <entry value="512" name="GIMBAL_DEVICE_ERROR_FLAGS_NO_MANAGER">
+        <description>Gimbal device is not assigned to a gimbal manager.</description>
       </entry>
     </enum>
     <!-- gripper action enum -->


hamishwillee (approved):
@julianoes This makes a lot of sense/helpful for debugging. I'm approving but up to you to decide if there are any issues.

julianoes (commented):
Does a gimbal device know that it doesn't have a gimbal manager? I don't think so as it doesn't keep track of that state. It just does what a manager says, I think :thinking:

olliw42:
it cannot know if a gimbal device message it receives is comming from its gimbal manager, but it can
- detect if it gets GD messages at all
- detect if it gets GD messages from just one sys/compid source or more
- detect if GD messages dissapear again
etc pp

STorM32 uses this determine if it shall accept a gimbal device message or not, and the state of that error flag

(technically there hence could be also an errors state TWO_OR_MORE_MANAGER, but if it detects two or more in ca 1 sec, it behaves as if NO MANAGER, i.e., rejects the messages)

julianoes:
Alright, I guess it can be worthwhile to have then.

hamishwillee (approved):
The justification is good, and it makes sense the gimbal should take input from just one manager. I think that requirement should be added to the docs somewhere, if it is not already - Ollie can you take that task?

Note, I did consider suggesting this was inverted to not being an error - "gimbal device has manager". But that doesn't add any real value.

olliw42:
> Ollie can you take that task?

you mean, I should add to this page https://mavlink.io/en/services/gimbal_v2.html, right?

technically, it is already specified by 
- "a Gimbal Manager has a 1:1 relationship with a particular Gimbal Device"
- "The Gimbal Device must act only upon messages that come from the associated Gimbal Manager!"

But maybe that's not clear enough. I will have a look at it.

> Note, I did consider suggesting this was inverted to not being an error - "gimbal device has manager". But that doesn't add any real value.

but would have had one side effect I would not have been so fond off, namely that "error = 0" would not have menat all ok anymore ...

hamishwillee:
> > Ollie can you take that task?
> 
> you mean, I should add to this page https://mavlink.io/en/services/gimbal_v2.html, right?
> 
> technically, it is already specified by
> 
> * "a Gimbal Manager has a 1:1 relationship with a particular Gimbal Device"
> * "The Gimbal Device must act only upon messages that come from the associated Gimbal Manager!"
> 
> But maybe that's not clear enough. I will have a look at it.
> 

@olliw42 I don't think it is sufficient. Yes it says the requirement but we could easily hint at _how_ you can make this happen.

If you don't want to do it, let me know and I'll create an issue to update this when I find the energy!

olliw42:
ok
I will be happy to give it a try :) If it will make things better will have to be seen  LOL

PR #1930, opened 2022-12-07
COMPONENT_INFORMATION_BASIC- software/hardware version as int

hamishwillee:
This proposes additional version fields in COMPONENT_INFORMATION_BASIC for the hardware and software version, and fields for custom model name.

Proposed in  https://github.com/mavlink/mavlink/pull/1788

#### Versions

Currently COMPONENT_INFORMATION_BASIC has strings for software and hardware version. The strings can have any format, but semver is recommended. The string is intended to be displayed by a GCS.

The proposal is 2 new uint32 that encodes major.minor.build.patch. This is efficient and allows the version to be used for more than just display - i.e. it can be used for comparing specific software versions and performing comparative operations.

#### Custom "user defined" names

See https://github.com/mavlink/mavlink/pull/1788#discussion_r1030475717

The idea is that if you have four gimbals (say) a user can give each one of them a name of their choice. This could be displayed in the UI of a GCS to make identification more user friendly.


message_definitions/v1.0/development.xml
```diff
@@ -431,6 +431,9 @@
...
       <field type="char[24]" name="software_version">Software version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
       <field type="char[24]" name="hardware_version">Hardware version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
+      <field type="char[32]" name="custom_name">Custom name of the component given to it by the user. Needs to be zero terminated. The field is optional and can be empty/all zeros.</field>      <field type="uint32_t" name="software_version_int">Component software version. Encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). 0 if unused.</field>
+      <field type="uint32_t" name="software_version_int">Component software version. Encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). 0 if unused.</field>
```
hamishwillee, line 435: What is dev? i.e. our string version just has major.minor.patch ?

hamishwillee: Is this better than using a `uint8_t[4]`

olliw42:
MANY thx, sir !!

message_definitions/v1.0/development.xml
olliw42: it's the same format as used in few other places already
the idea here is certainly also to not introduce yet another variant of version specification ... !!
hence u32 and x.x.x.x, i.e., taken over from the other places

(FYI, some idea of what dev could be: https://en.wikipedia.org/wiki/Software_versioning#Designating_development_stage)

message_definitions/v1.0/development.xml
```diff
@@ -431,6 +431,9 @@
       <field type="char[32]" name="model_name">Name of the component model. Needs to be zero terminated. The field is optional and can be empty/all zeros.</field>
       <field type="char[24]" name="software_version">Software version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
       <field type="char[24]" name="hardware_version">Hardware version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
+      <field type="char[32]" name="custom_name">Custom name of the component given to it by the user. Needs to be zero terminated. The field is optional and can be empty/all zeros.</field>
```
hamishwillee, line 434: Discussed this in dev call. General feeling was that for custom name we are OK with it if you are willing to add support for it to both QGC and Mission Planner upstream. Otherwise no.

message_definitions/v1.0/development.xml
```diff
@@ -431,6 +431,9 @@
       <field type="char[32]" name="model_name">Name of the component model. Needs to be zero terminated. The field is optional and can be empty/all zeros.</field>
       <field type="char[24]" name="software_version">Software version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
       <field type="char[24]" name="hardware_version">Hardware version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
```
hamishwillee, line 433: @olliw42 Dev call discussion:
- very much like the idea of predictable version values that you can actually rely on to be some format.
- free text field is needed to hold things like git hashes. But currently not so useful for tooling, because no way to know that it is a git hash.
- prefer just one field for the data. 
- Don't want to add fields for every format.

The compromise is to add a field that indicates the formatting (in both fields). The default is that it is just free text as now, but if you want to indicate that it is really semver, you can set a value of 1, or git hash you can set a value of 2. This is extensible.

Arguably we might make semver the default "0" and free text an option. Not sure.


```suggestion
      <field type="uint8_t" name="version_format">Format of software_version and hardware_version fields. 0: Free text format (for display only). 1: SEMVER 'major.minor.patch' (dots are required, even if values are not specified. 2: Git hash. </field>
      <field type="char[24]" name="software_version">Software version. Format is defined by `version_format` field. Default is arbitrary format, which may be displayed. SEMVER (1) is recommended. The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
      <field type="char[24]" name="hardware_version">Hardware version. Format is defined by `version_format` field. Default is arbitrary format, which may be displayed. SEMVER (1) is recommended. The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
```

olliw42:
many thx for considering and discussing this, but I think this should be closed. 
I think I rather do not want to support/use this message and service.
many thx again, much appreciated :)

message_definitions/v1.0/development.xml
hamishwillee: The version_format tells you what is being sent in the other fields. So if that has value 1, then you would parse this as a string `major.minor.patch`. If it does not match that string format then you can discard it.

olliw42: I read this in the context of the suggestion of the u32 fields. 
So, when I read "_very much like the idea of predictable version values that you can actually rely on to be some format_" I read this as "_very much like the idea of predictable version values that you can actually rely on to be some format as it is realized by the u32 fields_". The new suggestion cannot offer this same "can actually rely on". All what the new suggestion does is help parsing the string at the time of when you got it, but it does NOT help you to know what you are going to get. In contrast with u32, I can tell you already now what format it is going to be for you :). The 1st and 3rd bullets are logically exclusive, and the suggestion is thus not a compromise, it's a different thing.
I am not saying that it is therefore bad, or one shouldn't do that, only that it's not a substitute and doesn't nearly accomplish the same thing. All I am saying. :)

(and I'm implicitely saying that I'm not very interested in the discussion of the string ... my string version would be going to be in my format anyways LOL)

olliw42: just as food for thought, I guess if I were to do such a flag I would consider if allowing sending multiple versions might be worth a thought. E.g., it could be bitmask, indicating which formats are contained in the string, and in the string they could be separated by \n or \t or ';' or whatever, i.e., the string could be made an array of strings. The first raised bit would be then associated with the first array element, and so on. One could then send semver, git, and plain all at once.
(I think reading a string as array of strings does occur in some other message already, can't recall which from top of my head, mftp?)
One also could then actually drop the flag at all, i.e., specify that the string must be an array of fixed length, and that empty entries are just empty in this string, e.g. ",,githash" or "vXYZ,,githash" or ",1.2.3.4," etc
so, at the end one also has multiple versions and fields, just it's stored more compactly in the message. (and it is easily extensible with more versions formats)
just as food for thought :)

hamishwillee: The dev call was more keen on specifying the content of the string than adding an int. If we add the int to get a reliable semver then we would still need to think about how we tell consumers that the string happens to be a githash.

I don't know whether supporting multiple version formats at the same time is desirable. I kind of think it might be. Or to be more precise I think it likely you would want to be able to say the hardware format is a semver and the software format is a git hash - and that can be done using a formatted string but not with a single enum.

olliw42: concerning the str, another approach could be to allow several parts, separated e.g. with a ';', and define shorts to prepend them if of a certain type, like "semver:2.3.10;git:54fgh6;uid:123456;argh my own version variant v2.63j"
char24 might be a bit short then

message_definitions/v1.0/development.xml
hamishwillee: OK, removing it
```suggestion
```

message_definitions/v1.0/development.xml
hamishwillee: Dev call still thinks that using enums is the most flexible way to indicate the content of the fields.

```suggestion
      <field type="uint8_t" name="software_version_format">Format of text in software_version field. 0: Free text format (for display in GCS), 1: SEMVER 'major.minor.patch' (dots are required, even if values are not specified, 2: Git hash. 3. Free format text (not suitable for display).</field>
      <field type="uint8_t" name="hardware_version_format">Format of text in hardware_version field. 0: Free text format (Displayable), 1: SEMVER 'major.minor.patch' (dots are required, even if values are not specified, 2: Git hash. 3. Free format text (not suitable for display).</field>
      <field type="char[24]" name="software_version">Software version. Format is defined by `software_version_format` field. Default is arbitrary format, which may be displayed. SEMVER (1) is recommended. The field must be null terminated if it has a value of less than 24 chars. If not used the field should be zero-filled.</field>
      <field type="char[24]" name="hardware_version">Hardware version. Format is defined by `hardware_version_format` field. Default is arbitrary format, which may be displayed. SEMVER (1) is recommended. The field must be null terminated if it has a value of less than 24 chars. If not used the field should be zero-filled.</field>
```

julianoes: I guess that would be ok.

julianoes: I have no strong opinion here, you can suggest what you want.

I came up with a string because I expected people to come up with whatever version they want. Some want 3 numbers, some want 5 numbers, some want a date, and some want a number with a git hash, what do I know :man_shrugging:.

hamishwillee: @olliw42 @julianoes 

The real question is what are the requirements here? The original was:
- to allow any desired format to be sent

The enum meets the additional requirement:
- allow the format of the string to be known so consumer can decide whether to display it or do something else.

Now it looks like there might be more requirements
- Allow multiple formats to be sent at same time - mentioned cases are semver, githash, date
- Allow separate formats for display and "use"

Are those use cases we want to meet? 

If so, then I see two options. Either we encode multiple formats in the string as suggested by @olliw42 or we decide our set of supported versions and provide fields for just those - so:
- semver in an int with 5 values (or whatever most people use for server - optional ones get dumped)
- date - uint32 in YYYYMMDD format
- githash - as a 10 char array

olliw42: as much as I see from the dicsussion (also taking into account where it came from), it seems to me that
- one may want a version thing which allows some automated proecssing, like a firmware update. If I got it right the git hash was suggested for that purpose
- one may want a version thing which one can display to a user, so that the user knows what is up (which implies that it is in a format which the user is more used to than a githash), and which probably would also be used in any support ticket, i.e. customer-vendor interaction
- some version formats are already in use in other messages, specifically the unit32_t semver of format x.x.x.x, and in this sense it's already the version which is in use

just to have droped that thought, instead of the encoding the multiple formats in teh string with the labels, one also could use the enum as bit mask, and have a simpler ',' separated format. Makes it easier to process by code.

hamishwillee:
After all that we decided to do nothing. The ids are strings that can be used by component sender/receivers any way they like.  
Thanks for the discussion @olliw42

PR #1931, opened 2022-12-07
[message_definitions:v1] Bug | `WIFI_CONFIG_AP` field types are inconsistent with semantics.

damurashov:
`password` field description stipulates that MD5 hash must be returned as a response, which is, practically, a non human-readable byte sequence. Therefore, `uint8_t` type seems to be more appropriate.


hamishwillee:
We wouldn't change this because it would break any users who currently rely on the message.

PR #1928, opened 2022-11-30
Gimbal manager: Add gimbal device flags to low word of gimbal manager flags

olliw42:
The lower 16 bits of the gimbal manager flags and cap flags are supposed to coincide the gimbal device flags and cap flags. This PR aligns the manager flags & cap flags with the recent additions to the gimbal device flags & cap flags. I have also added .'s in the descriptions of the gimbal manager flags, as it's so in the descriptions of the gimbal manager cap flags.

PR #1932, opened 2022-12-07
Gimbal manager cap flags: resolve duplication mistake

olliw42:
omg ... my fault, I'm so sorry :(
https://github.com/mavlink/mavlink/pull/1928#issuecomment-1341373366


dagar:
Thanks for the quick fix.

PR #1933, opened 2022-12-12
Add the MAV_TYPE_VTOL_TILTWING aircraft type

maaaat:
Quoting the Wikipedia article on [Tiltwing](https://en.wikipedia.org/wiki/Tiltwing):

> A tiltwing aircraft features a wing that is horizontal for conventional forward flight and rotates up for vertical takeoff and landing. It is similar to the tiltrotor design where only the propeller and engine rotate. Tiltwing aircraft are typically fully capable of VTOL operations.

I took the liberty of re-purposing `MAV_TYPE_VTOL_RESERVED4` for this aircraft type. Of course I would be totally fine with adding a new type instead.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -142,10 +142,10 @@
       <entry value="23" name="MAV_TYPE_VTOL_TAILSITTER">
         <description>Tailsitter VTOL. Fuselage and wings orientation changes depending on flight phase: vertical for hover, horizontal for cruise. Use more specific VTOL MAV_TYPE_VTOL_DUOROTOR or MAV_TYPE_VTOL_QUADROTOR if appropriate.</description>
       </entry>
-      <!-- Entries 24/25 reserved for other VTOL airframe -->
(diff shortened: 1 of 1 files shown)


hamishwillee:
@maaaat Sounds reasonable. Do you actually have a vehicle using MAVLink that needs this?

@auturgy @julianoes Any objections?

maaaat:
@hamishwillee Yeah I am currently working on such an aircraft. It has two rotating wings, front and rear, with two propellers each, so it can convert between quad-rotor and fixed-wing configurations.

auturgy:
No objections. This configuration is supported by ArduPilot (one of the setup options in https://ardupilot.org/plane/docs/guide-tilt-rotor.html).

julianoes:
I'll pass my review to @sfuhrer.

hamishwillee:
@maaaat Awesome. I don't see any reason why we shouldn't get this in. Just waiting on @sfuhrer since we try make sure major stakeholders all get a chance to comment.

hamishwillee (approved):

sfuhrer (approved):

PR #1941, opened 2023-01-16
Update storm32.xml

olliw42:
This updates the storm32.xml dialect file. The changes are three-fold

1. Remove STorM32 gimbal device messages. This set of messages/commands/enums is completely removed. The functionality is completely endorsed by the gimbal device messages of the MAVLink standard gimbal protocol v2, which shall be and is used now instead. The STorM32 gimbal manager messages/commands/enums have been adapted were required to reflect that change, which doesn't break compatibility.
2. Some more changes were made in the STorM32 gimbal manager section (a cmd and two messages have been removed), and some cosmetic changes were made.
3. Add mLRS project related messages/enums. (mLRS is a lora based rc & mavlink radio link, https://github.com/olliw42/mLRS).
 
While (1) and (2) should not be controversial, and (1) should be actually welcome, point (3) is the part where I wasn't sure how I should approach this. On the one hand, since these are messages related to another project but not strictly STorM32, the canonical approach would be to introduce a respective dialect (e.g. mLRS.xml). However, this felt quite hilarious to me given that there is already a dialect of mine in here - and I felt more so since the external dialcets are not separated out in an extra folder anymore but clutter the main folder. Another approach would have been to substitute/rename storm32.xml to e.g. olliw.xml, which looked even more hilarious to me. So I puit them also into the existing storm32.xml file. I hope you find this acceptable. If not, I would be happy to hear your suggestion.

Cheers, Olli

### DIFF
--- a/message_definitions/v1.0/storm32.xml
+++ b/message_definitions/v1.0/storm32.xml
@@ -7,9 +7,10 @@ Contact:
 Range of IDs:
   messages: 60000 - 60049
   commands: 60000 - 60049
-Documentation:  
+Documentation:
   STORM32 and QSHOT additions
-  6. Okt. 2021
+  with mLRS additions merged
+  16. Jan. 2023
   All messages are technically WIP, but some are quite stable now.
   Quite stable means that it is in practical use, but may see extension.
   A more detailed description of the concept underlying the STORM32 and QSHOT messages can be found here:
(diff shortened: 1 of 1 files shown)


olliw42:
@hamishwillee 
any life out there?
(no devcall?)
:)

hamishwillee (approved):
@olliw42 Thanks - just back from several weeks holiday off the grid. 

Yes, 1, 2, welcome. The existence of any dialects pains us greatly, but you're quite right that there is no more obvious or reasonable place to put them. I don't want to talk about it :-(.

My take on this is that any change in a dialect file that is made with the owner of that dialect is not something I can block, provided it does not use any clashing ids, or any ids outside of the dialect-granted range. 

Of course I may later have a tantrum if we need to create our own thing named "RADIO_RC_CHANNELS"

@julianoes Any objections?

olliw42:
I hope you had a great vaccation

if some of the messages in this dialect ever would want to exist in the standard I can ensure you that I will be constructive, and would not hesitate to move to standard equivalents etc. pp. I am all for a standard too, but on the other side I am also all for making progress. Sometimes this puts me in the situation to change the dialect.

many thx as always

julianoes (approved):

hamishwillee:
For sure! Merged.

olliw42:
THANKS folks!

PR #1943, opened 2023-01-19
New landing target messages TARGET_ABSOLUTE/RELATIVE

JonasPerolini:
This pull request extends the [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) message to include position and orientation uncertainties in order to properly fuse the landing target estimations in a Kalman filter https://github.com/PX4/PX4-Autopilot/pull/20653. It also adds the sensor's orientation at the moment of the observation, which is necessary if there is processing time (e.g. detect fiducial marker) between the observation and the message being sent. Finally, it extends the [MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME) message to allow for generic sensor frames. This allows to properly handle the [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) message: 

- if MAV_FRAME == MAV_FRAME_LOCAL_NED, the target is directly sent to the precision landing algorithm.
- if MAV_FRAME == MAV_FRAME_SENSOR, the target is sent to the Kalman filter which will rotate it into vehicle-carried NED using the sensor's orientation `q_sensor_att`. The estimation is then sent to the precision landing algorithm.
- Otherwise: the target is sent to the Kalman filter which will rotate it into vehicle-carried NED using: 
    - `q_sensor_att` if filled 
    - the current attitude of the drone (properly transformed into MAV_FRAME). 
   
Changes: 

- Added position and orientation uncertainties to the [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) message
- Added sensor's orientation at the moment of the observation to the [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) message
- Extended the [MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME) message to allow for generic sensor frames

Please let me know if there are any further modifications needed.

Update: 

We currently have two main tasks using targets: 
1. following a target with the message "FOLLOW_TARGET" 
2. landing on a target with the message "LANDING_TARGET." 

However, the main difference between these messages is not the task, but rather the physical location of the sensor used. [FOLLOW_TARGET](https://mavlink.io/en/messages/common.html#FOLLOW_TARGET) reports observations from a **sensor on the target** (e.g. GPS or IMUs), while [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) reports observations from an **onboard sensor** (e.g. vision-based fiducial marker detection or irlock beacon detection). One might want to follow a target using a fiducial marker (LANDING_TARGET msg). Inversely, one might want to land on a target with a GPS attached (FOLLOW_TARGET msg). Each message should thus be usable independently of the task. Additionally, the new target estimator allows to fuse several measurements such as GPS, fiducial markers and irlock beacons in a Kalman Filter to estimate the target's state (for static and moving targets) https://github.com/PX4/PX4-Autopilot/pull/20653. The same filter class can then be used for precision landing and the target following tasks, reinforcing the idea that both messages should be used in parallel.

Proposed changes: 
1. "LANDING_TARGET" becomes "RELATIVE_TARGET" (for onboard sensors such as fiducial markers, irlock beacons, etc.)
2. "FOLLOW_TARGET" becomes "ABSOLUTE_TARGET" (for sensors on the target such as GPS, IMUs, etc.). 
3. Both "RELATIVE_TARGET" and "ABSOLUTE_TARGET" have a "TARGET_TASK" field, which can either be "TARGET_TASK_FOLLOW" or "TARGET_TASK_LAND," depending on the task at hand.

Additionally, adding a [MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME) field to the ABSOLUTE_TARGET msg allows for measurements to be entered in the target's frame and use the `quaternion` field to rotate it to the navigation NED frame, making the message less confusing.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -376,8 +376,8 @@
       <field type="float[4]" name="q_target" invalid="[0]">Quaternion of the target's orientation from its body frame to the vehicle's NED frame.</field>
       <field type="float[3]" name="rates" units="rad/s" invalid="[0]">Target's roll, pitch and yaw rates</field>
       <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
-      <field type="float[3]" name="vel_var" units="[m/s]^2">Variance of the target's velocity in its body frame</field>
-      <field type="float[3]" name="acc_var" units="[m/s/s]^2">Variance of the target's acceleration in its body frame</field>
+      <field type="float[3]" name="vel_std" units="[m/s]">Standard deviation of the target's velocity in its body frame</field>
+      <field type="float[3]" name="acc_std" units="[m/s/s]">Standard deviation of the target's acceleration in its body frame</field>
     </message>
     <!-- Target info measured by MAV's onboard sensors -->
     <message id="511" name="TARGET_RELATIVE">
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -302,8 +302,8 @@
       <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time)</field>
       <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
-      <field type="float" name="x" units="m">X Position of the target in MAV_FRAME</field>
-      <field type="float" name="y" units="m">Y Position of the target in MAV_FRAME</field>
+      <field type="float" name="x" units="m">X Position of the target in MAV_FRAME (For IR-Lock: x-axis distance from the center of image to the center of the target in units of tan(theta))</field>
+      <field type="float" name="y" units="m">Y Position of the target in MAV_FRAME (For IR-Lock: y-axis distance from the center of image to the center of the target in units of tan(theta))</field>
       <field type="float" name="z" units="m">Z Position of the target in MAV_FRAME</field>
       <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in MAV_FRAME (var_x, var_y, var_z)</field>
       <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in MAV_FRAME</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -396,8 +396,8 @@
       <field type="float" name="x" units="m">X Position of the target in TARGET_OBS_FRAME</field>
       <field type="float" name="y" units="m">Y Position of the target in TARGET_OBS_FRAME</field>
       <field type="float" name="z" units="m">Z Position of the target in TARGET_OBS_FRAME</field>
-      <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in TARGET_OBS_FRAME (var_x, var_y, var_z)</field>
-      <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in TARGET_OBS_FRAME</field>
+      <field type="float[3]" name="pos_std" units="m">Standard deviation of the target's position in TARGET_OBS_FRAME</field>
+      <field type="float" name="yaw_std" units="rad">Standard deviation of the target's orientation in TARGET_OBS_FRAME</field>
       <field type="float[4]" name="q_target">Quaternion of the target's orientation from the target's frame to the TARGET_OBS_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -391,7 +391,7 @@
       <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in TARGET_OBS_FRAME (var_x, var_y, var_z)</field>
       <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in TARGET_OBS_FRAME</field>
       <field type="float[4]" name="q_target">Quaternion of the target's orientation from the target's frame to the TARGET_OBS_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
-      <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if TARGET_OBS_FRAME is in LOCAL_NED) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
+      <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
   </messages>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -248,7 +248,10 @@
       <entry value="1" name="TARGET_OBS_FRAME_BODY_FRD">
         <description>FRD local frame aligned to the vehicle's attitude (x: Forward, y: Right, z: Down) with an origin that travels with vehicle.</description>
       </entry>
-      <entry value="2" name="TARGET_OBS_FRAME_OTHER">
+      <entry value="2" name="TARGET_OBS_FRAME_LOCAL_OFFSET_NED">
+        <description>NED local tangent frame (x: North, y: East, z: Down) with an origin that travels with vehicle.</description>
+      </entry>
+      <entry value="3" name="TARGET_OBS_FRAME_OTHER">
         <description>Other sensor frame for target observations neither in local NED nor in body FRD.</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -396,8 +396,8 @@
       <field type="float" name="x" units="m">X Position of the target in TARGET_OBS_FRAME</field>
       <field type="float" name="y" units="m">Y Position of the target in TARGET_OBS_FRAME</field>
       <field type="float" name="z" units="m">Z Position of the target in TARGET_OBS_FRAME</field>
-      <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in TARGET_OBS_FRAME</field>
-      <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in TARGET_OBS_FRAME</field>
+      <field type="float[3]" name="pos_std" units="m">Standard deviation of the target's position in TARGET_OBS_FRAME</field>
+      <field type="float" name="yaw_std" units="rad">Standard deviation of the target's orientation in TARGET_OBS_FRAME</field>
       <field type="float[4]" name="q_target">Quaternion of the target's orientation from the target's frame to the TARGET_OBS_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -240,6 +240,14 @@
         <param index="7" reserved="true" default="NaN"/>
       </entry>
     </enum>
+    <enum name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAGS" bitmask="true">
+      <description>These flags indicate the sensor reporting capabilities for TARGET_ABSOLUTE.</description>
+      <entry value="1" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_POSITION"/>
+      <entry value="2" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_VELOCITY"/>
+      <entry value="4" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_ACCELERATION"/>
+      <entry value="8" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_ATTITUDE"/>
+      <entry value="16" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_RATES"/>
+    </enum>
     <enum name="TARGET_OBS_FRAME">
       <description>The frame of a target observation from an onboard sensor.</description>
       <entry value="0" name="TARGET_OBS_FRAME_LOCAL_NED">
@@ -367,12 +375,12 @@
       <description>Current motion information from sensors on a target</description>
       <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time).</field>
       <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
-      <field type="uint8_t" name="sensor_capabilities">bit positions for the sensor's reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT = 3, RATES = 4)</field>
+      <field type="uint8_t" name="sensor_capabilities" enum="TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAGS" display="bitmask">Bitmap to indicate the sensor's reporting capabilities</field>
       <field type="int32_t" name="lat" units="degE7">Target's latitude (WGS84)</field>
       <field type="int32_t" name="lon" units="degE7">Target's longitude (WGS84)</field>
       <field type="float" name="alt" units="m">Target's altitude (AMSL)</field>
-      <field type="float[3]" name="vel" units="m/s" invalid="[0]">target's velocity in its body frame</field>
-      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">linear target's acceleration in its body frame</field>
+      <field type="float[3]" name="vel" units="m/s" invalid="[0]">Target's velocity in its body frame</field>
+      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">Linear target's acceleration in its body frame</field>
       <field type="float[4]" name="q_target" invalid="[0]">Quaternion of the target's orientation from its body frame to the vehicle's NED frame.</field>
       <field type="float[3]" name="rates" units="rad/s" invalid="[0]">Target's roll, pitch and yaw rates</field>
       <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -240,6 +240,15 @@
         <param index="7" reserved="true" default="NaN"/>
       </entry>
     </enum>
+    <enum name="TARGET_TASK">
+      <description>Type of task associated with the target measurement</description>
+      <entry value="0" name="TARGET_TASK_FOLLOW">
+        <description>Follow the target from a given distance</description>
+      </entry>
+      <entry value="1" name="TARGET_TASK_LAND">
+        <description>Land on the target</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <!-- Transactions for parameter protocol -->
@@ -251,6 +260,42 @@
       <field type="float" name="param_value">Parameter value (new value if PARAM_ACCEPTED, current value otherwise)</field>
       <field type="uint8_t" name="param_type" enum="MAV_PARAM_TYPE">Parameter type.</field>
       <field type="uint8_t" name="param_result" enum="PARAM_ACK">Result code.</field>
+    </message>
+    <!-- Target info from a sensor on the target -->
+    <message id="145" name="ABSOLUTE_TARGET">
+      <description>Current motion information from sensors on a target</description>
+      <field type="uint64_t" name="timestamp" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="est_capabilities">bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT + RATES = 3)</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL)</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float[3]" name="vel" units="m/s" invalid="[0]">target velocity in MAV_FRAME (0,0,0) for unknown</field>
+      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">linear target acceleration in MAV_FRAME (0,0,0) for unknown</field>
+      <field type="float[4]" name="attitude_q" invalid="[0]">Quaternion of the target's orientation from MAV_FRAME to vehicle-carried NED. (Not used if MAV_FRAME is MAV_FRAME_LOCAL_NED). (0 0 0 0 for unknown)</field>
+      <field type="float[3]" name="rates" invalid="[0]">(0 0 0 for unknown)</field>
+      <field type="float[2]" name="position_cov">eph epv</field>
+      <field type="float[3]" name="vel_var" units="[m/s]^2">Variance of the target's velocity in MAV_FRAME</field>
+      <field type="float[3]" name="acc_var" units="[m/s/s]^2">Variance of the target's acceleration in MAV_FRAME</field>
+      <field type="uint64_t" name="custom_state">button states or switches of a tracker device</field>
+      <field type="uint8_t" name="task" enum="TARGET_TASK">Type of task associated with the target measurement (e.g. follow or land)</field>
+    </message>
+    <!-- Target info from an onboard sensor -->
+    <message id="150" name="RELATIVE_TARGET">
+      <description>The location of a target from an onboard sensor. </description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (time since system boot)</field>
+      <field type="uint8_t" name="target_num">The ID of the target if multiple targets are present</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x" units="m">X Position of the landing target in MAV_FRAME</field>
+      <field type="float" name="y" units="m">Y Position of the landing target in MAV_FRAME</field>
+      <field type="float" name="z" units="m">Z Position of the landing target in MAV_FRAME</field>
+      <field type="float[3]" name="pos_var" units="m^2">Variance of landing target position in MAV_FRAME (var_x, var_y, var_z)</field>
+      <field type="float" name="target_yaw_unc" units="rad^2">Variance of landing target's orientation in MAV_FRAME</field>
+      <field type="float[4]" name="q">Quaternion of landing target orientation (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
+      <field type="float[4]" name="q_sensor_att">Quaternion of the sensor's orientation from MAV_FRAME to vehicle-carried NED. (Not used if MAV_FRAME is already MAV_FRAME_LOCAL_NED) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
+      <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of landing target</field>
+      <field type="uint8_t" name="task" enum="TARGET_TASK">Type of task associated with the target measurement (e.g. follow or land)</field>
+      <field type="uint8_t" name="position_valid" invalid="0">Boolean indicating whether the position fields (x, y, z, q, type) contain valid target position information (valid: 1, invalid: 0). Default is 0 (invalid).</field>
     </message>
     <!-- mission protocol enhancements -->
     <message id="53" name="MISSION_CHECKSUM">
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -276,42 +276,6 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
       <field type="uint32_t" name="checksum">CRC32 checksum of current plan for specified type.</field>
     </message>
-    <!-- Target info from a sensor on the target -->
-    <message id="145" name="ABSOLUTE_TARGET">
-      <description>Current motion information from sensors on a target</description>
-      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time).</field>
-      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
-      <field type="uint8_t" name="est_capabilities">bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT + RATES = 3)</field>
-      <field type="int32_t" name="lat" units="degE7">Target's latitude (WGS84)</field>
-      <field type="int32_t" name="lon" units="degE7">Target's longitude (WGS84)</field>
-      <field type="float" name="alt" units="m">Target's altitude (AMSL)</field>
-      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
-      <field type="float[3]" name="vel" units="m/s" invalid="[0]">target's velocity in MAV_FRAME (0,0,0) for unknown</field>
-      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">linear target's acceleration in MAV_FRAME (0,0,0) for unknown</field>
-      <field type="float[4]" name="q_att" invalid="[0]">Quaternion of the target's orientation from MAV_FRAME to vehicle-carried NED. (Not used if MAV_FRAME is MAV_FRAME_LOCAL_NED). (0 0 0 0 for unknown)</field>
-      <field type="float[3]" name="rates" invalid="[0]">(0 0 0 for unknown)</field>
-      <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
-      <field type="float[3]" name="vel_var" units="[m/s]^2">Variance of the target's velocity in MAV_FRAME</field>
-      <field type="float[3]" name="acc_var" units="[m/s/s]^2">Variance of the target's acceleration in MAV_FRAME</field>
-      <field type="uint64_t" name="custom_state">button states or switches of a tracker device</field>
-      <field type="uint8_t" name="task" enum="TARGET_TASK">Type of task associated with the target observation (e.g. FOLLOW or LAND)</field>
-    </message>
-    <!-- Target info from an onboard sensor -->
-    <message id="150" name="RELATIVE_TARGET">
-      <description>The location of a target from an onboard sensor. </description>
-      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time)</field>
-      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
-      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
-      <field type="float" name="x" units="m">X Position of the target in MAV_FRAME (For IR-Lock: x-axis distance from the center of image to the center of the target in units of tan(theta))</field>
-      <field type="float" name="y" units="m">Y Position of the target in MAV_FRAME (For IR-Lock: y-axis distance from the center of image to the center of the target in units of tan(theta))</field>
-      <field type="float" name="z" units="m">Z Position of the target in MAV_FRAME</field>
-      <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in MAV_FRAME (var_x, var_y, var_z)</field>
-      <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in MAV_FRAME</field>
-      <field type="float[4]" name="q_att">Quaternion of the target's orientation from the target's frame to the MAV_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
-      <field type="float[4]" name="q_sensor_att">Quaternion of the sensor's orientation from MAV_FRAME to vehicle-carried NED. (Not used if MAV_FRAME is already MAV_FRAME_LOCAL_NED) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
-      <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
-      <field type="uint8_t" name="task" enum="TARGET_TASK">Type of task associated with the target observation (e.g. FOLLOW or LAND)</field>
-    </message>
     <message id="295" name="AIRSPEED">
       <description>Airspeed information from a sensor.</description>
       <field type="uint8_t" name="id" instance="true">Sensor ID.</field>
@@ -391,6 +355,42 @@
       <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
       <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
+    </message>
+    <!-- Target info from a sensor on the target -->
+    <message id="510" name="ABSOLUTE_TARGET">
+      <description>Current motion information from sensors on a target</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time).</field>
+      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
+      <field type="uint8_t" name="est_capabilities">bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT + RATES = 3)</field>
+      <field type="int32_t" name="lat" units="degE7">Target's latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Target's longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Target's altitude (AMSL)</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float[3]" name="vel" units="m/s" invalid="[0]">target's velocity in MAV_FRAME (0,0,0) for unknown</field>
+      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">linear target's acceleration in MAV_FRAME (0,0,0) for unknown</field>
+      <field type="float[4]" name="q_att" invalid="[0]">Quaternion of the target's orientation from MAV_FRAME to vehicle-carried NED. (Not used if MAV_FRAME is MAV_FRAME_LOCAL_NED). (0 0 0 0 for unknown)</field>
+      <field type="float[3]" name="rates" invalid="[0]">(0 0 0 for unknown)</field>
+      <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
+      <field type="float[3]" name="vel_var" units="[m/s]^2">Variance of the target's velocity in MAV_FRAME</field>
+      <field type="float[3]" name="acc_var" units="[m/s/s]^2">Variance of the target's acceleration in MAV_FRAME</field>
+      <field type="uint64_t" name="custom_state">button states or switches of a tracker device</field>
+      <field type="uint8_t" name="task" enum="TARGET_TASK">Type of task associated with the target observation (e.g. FOLLOW or LAND)</field>
+    </message>
+    <!-- Target info from an onboard sensor -->
+    <message id="511" name="RELATIVE_TARGET">
+      <description>The location of a target from an onboard sensor. </description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time)</field>
+      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x" units="m">X Position of the target in MAV_FRAME (For IR-Lock: x-axis distance from the center of image to the center of the target in units of tan(theta))</field>
+      <field type="float" name="y" units="m">Y Position of the target in MAV_FRAME (For IR-Lock: y-axis distance from the center of image to the center of the target in units of tan(theta))</field>
+      <field type="float" name="z" units="m">Z Position of the target in MAV_FRAME</field>
+      <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in MAV_FRAME (var_x, var_y, var_z)</field>
+      <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in MAV_FRAME</field>
+      <field type="float[4]" name="q_att">Quaternion of the target's orientation from the target's frame to the MAV_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
+      <field type="float[4]" name="q_sensor_att">Quaternion of the sensor's orientation from MAV_FRAME to vehicle-carried NED. (Not used if MAV_FRAME is already MAV_FRAME_LOCAL_NED) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
+      <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
+      <field type="uint8_t" name="task" enum="TARGET_TASK">Type of task associated with the target observation (e.g. FOLLOW or LAND)</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -280,6 +280,7 @@
     <message id="145" name="ABSOLUTE_TARGET">
       <description>Current motion information from sensors on a target</description>
       <field type="uint64_t" name="timestamp" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="target_id">The ID of the target if multiple targets are present</field>
       <field type="uint8_t" name="est_capabilities">bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT + RATES = 3)</field>
       <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
       <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
@@ -298,8 +299,8 @@
     <!-- Target info from an onboard sensor -->
     <message id="150" name="RELATIVE_TARGET">
       <description>The location of a target from an onboard sensor. </description>
-      <field type="uint64_t" name="time_usec" units="us">Timestamp (time since system boot)</field>
-      <field type="uint8_t" name="target_num">The ID of the target if multiple targets are present</field>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (time since system boot)</field>
+      <field type="uint8_t" name="target_id">The ID of the target if multiple targets are present</field>
       <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
       <field type="float" name="x" units="m">X Position of the landing target in MAV_FRAME</field>
       <field type="float" name="y" units="m">Y Position of the landing target in MAV_FRAME</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -240,15 +240,6 @@
         <param index="7" reserved="true" default="NaN"/>
       </entry>
     </enum>
-    <enum name="TARGET_TASK">
-      <description>Type of task associated with the target measurement</description>
-      <entry value="0" name="TARGET_TASK_FOLLOW">
-        <description>Follow the target from a given distance</description>
-      </entry>
-      <entry value="1" name="TARGET_TASK_LAND">
-        <description>Land on the target</description>
-      </entry>
-    </enum>
     <enum name="TARGET_OBS_FRAME">
       <description>The frame of a target observation from an onboard sensor.</description>
       <entry value="0" name="LOCAL_NED">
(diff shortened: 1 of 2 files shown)


auturgy:
This breaks every existing implementation. In my opinion this needs to be new messages, as a pr against development.xml

JonasPerolini:
Hi, thank you for your response @auturgy. 

I do see your point, users will have to change the message name, and we will have to slightly change [MAVROS](https://github.com/mavlink/mavros). However, having new messages just delays the problem in my opinion because: 

1. It will make the whole target situation confusing i.e. should I use the new messages or the old ones.
2. The proposed changes keep the existing fields, therefore, the user only needs to change a name which seems reasonable when considering the benefits (i.e. new Kalman Filter for target tracking, please have a look at https://github.com/PX4/PX4-Autopilot/pull/20653).
3. Eventually, we will keep either the new messages or the old ones, and "every existing implementation will break".
4. New messages means new handling in the [mavlink_receiver](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/mavlink_receiver.cpp) and the [SimulatorMavlink](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/simulation/simulator_mavlink/SimulatorMavlink.cpp)
5. The current [handling of the landing target](https://github.com/PX4/PX4-Autopilot/blob/bd9d09663f2ccee14996e31df00933c1c4a5c202/src/modules/mavlink/mavlink_receiver.cpp#L2501-L2529) is extremely limited: 
     - Only Irlock reports are being forwarded to the Kalman filter. However, I would assume that most users use the [irlock plugin](https://github.com/PX4/PX4-Autopilot/blob/main/src/drivers/irlock/irlock.cpp) which is directly sent to the Kalman filter and does not require the MAVLINK msg. 
     - The `landing_target.type` field is not even used. 
     - All other observations must be handled directly by the user and be somehow converted to local NED.
     
     Therefore, the main (if not only) used case is to send a target in local NED through MAVLINK which is directly sent to the precision landing algorithm. With the new messages, this will not break, we can clearly state in the documentation that handled observation must be sent in local NED in the RELATIVE_TARGET msg (note: the ABSOLUTE_TARGET does not have a position field, only lat/lon so I don't see how someone would confuse the two).

Now there is another option: we keep the same message names (LANDING_TARGET and FOLLOW_TARGET), we add the proposed fields which drastically improve the overall behavior (see first post) and we clearly explain in the documentation that the names will eventually change to RELATIVE_TARGET and ABSOLUTE_TARGET (or any other msg name which does not contain "land" or "follow"). With this solution, no current implementation breaks and we keep all the benefits. 

Would that work for you @auturgy ?

auturgy:
This changes the crc: if someone updates their flight controller to a version that includes mavlink with your changes, but doesn't update a fully functional offboard implementation of follow/prec land (common scenario), the messages will be rejected by the parser. Needs to be new messages against development.xml

JonasPerolini:
I've created the new messages in development.xml. To properly handle the sensor's measurement frame transformation, I need to include the frame: `MAV_FRAME_SENSOR`. Is this an issue @auturgy ?

JonasPerolini:
Summary and use: 

New messages in `development.xml`: 
- `ABSOLUTE_TARGET` for sensors on the target e.g. GPS
- `RELATIVE_TARGET` for onboard sensors detecting the target e.g. fiducial marker, irlock, ... 
- Each message contains:  
     - `TARGET_TASK` which can be `FOLLOW` or `LAND`. This will allow to use the same Kalman Filter for precision landing and target following in the future. 
     -  `q_sensor_att`, the attitude of the sensor at the time of the measurement. The main purpose is to avoid overshoots due to target detection processing time e.g. if we have a complex target detection algorithm that must process the frame for 1 second, we absolutely need the attitude of the sensor when the camera frame was grabbed, not when the msg was sent. Note: it also allows to only send a diagonal covariance matrix (which will become full once we rotate it into the KF frame vc-NED `covRotated = R * covDiag * R^T)`
     - Variance of the observations, for proper Kalman Filtering 

Logic: 
- depending on the `TARGET_TASK`, the observation is forwarded to the `landing_target` estimator or to the `follow_target` estimator.
- The Kalman Filter fuses data in `vehicle-carried NED` frame to avoid being dependent on a local position estimate. Therefore, If the target is sent in `MAV_FRAME == MAV_FRAME_LOCAL_NED` the measurement is either: 
     - sent directly to the precision landing algorithm if the KF is disabled (same as current implementation).
     - ignored if the KF is enabled. (We assume that there are other observations)

use case example: Land on a target which is equipped with a GPS and a fiducial marker. 
- GPS data will be sent via `ABSOLUTE_TARGET`
- Fiducial marker data will be sent via `RELATIVE_TARGET`

Important note: with the new modifications, the current implementations using the messages [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) and [FOLLOW_TARGET](https://mavlink.io/en/messages/common.html#FOLLOW_TARGET) do not break: 
- irlock observation from [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) goes into the KF (as done right now)
- fiducial marker observation from [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) goes straight to the precision landing algorithm, if the Kalman Filter is disabled (default state)

The goal is to eventually replace [LANDING_TARGET](https://mavlink.io/en/messages/common.html#LANDING_TARGET) and [FOLLOW_TARGET](https://mavlink.io/en/messages/common.html#FOLLOW_TARGET) by `RELATIVE_TARGET` and `ABSOLUTE_TARGET`

message_definitions/v1.0/development.xml
```diff
@@ -267,6 +276,42 @@
       <field type="uint8_t" name="mission_type" enum="MAV_MISSION_TYPE">Mission type.</field>
       <field type="uint32_t" name="checksum">CRC32 checksum of current plan for specified type.</field>
     </message>
+    <!-- Target info from a sensor on the target -->
+    <message id="145" name="ABSOLUTE_TARGET">
```
hamishwillee, line 280: Message ids must not be in this range (Up to 256 reserved for reasons I don't want to go into). Try some values around 510 if they are unused in common.xml. In common.xml add a comment noting which ids are used at the location they would be in the file if they were in common.xml.

hamishwillee:
@dakejahl You've done a landing target implementation - does this make sense to you? Would this help or harm your implementation? Is there anyone else you can suggest as a reviewer?

Note that you kind of need to read the top description to get the full picture, even though the final "implementation" is in https://github.com/mavlink/mavlink/pull/1943#issuecomment-1400948146

I'm not sure about the names - I didn't read "ABSOLUTE" and think "onboard".
- GPS data will be sent via ABSOLUTE_TARGET - 
- Fiducial marker data will be sent via RELATIVE_TARGET

potaito:
@hamishwillee 
Onboard and Offboard are often two confusing terms, which is why we were looking for new names. Both terms can mean anything really. Particularly in this case you can have an algorithm run offboard or onboard, but produce an absolute measurement in both cases. 

- Offboard and Onboard often refer to where code is executed
- Absolute and Relative tries to tell that it does not matter where the code is running, but what kind of reference frame it is

Suggestions more than welcome of course :smile:

message_definitions/v1.0/development.xml
JonasPerolini: Ok, message ids updated in https://github.com/mavlink/mavlink/pull/1943/commits/b79e2f2a70bd8797d413b7d23028c1da57fc8f0a

hamishwillee:
@potaito I get that; the problem is that "absolute" might be a fixed frame, but the actual target can be moving. In other words, correct but not intuitive to me.

Thanks for updating the ids. I'm waiting on @dakejahl for a sanity check - I don't have the background to assess this.

hamishwillee:
@JonasPerolini What I'd prefer is feedback from others who are doing implementations and are at the right expertise level; I'm not competent in the domain to know what is useful information and what is intuitive to someone working in the area. So I won't offer an opinion at this time.

As the messages are quite different I'd prefer them separate as you have done; no point sending lots of unused information. FYI though, the packet length is about 255 bytes, so this would be doable if it was desired: https://mavlink.io/en/guide/serialization.html

dakejahl:
- `type` How is this data useful? Is this data actionable?
- `task` I don't think it makes sense for the target to know what its task is. If you want to use a target for a specific task this should be defined within the autopilot. For example use a parameter to associate the tag ID with the specific task. 
- `id` isn't very useful on its own. If you were using aruco tags you'd want to know both the dictionary and the tag ID. One simple option would be to use an enum, this would have the benefit of defining "ID"s for things other than aruco tags which may have a different convention (for example IR beacons). Although would require an update every time someone wants to add a new definition. I'm open to suggestions but IMO just a `uint8_t id` is not enough information. Another option could be `id` and `dict` with different definitions depending on the `type` (so we'd need the `type` field in this case), but this increases the message size.
```
ARUCO_DICT_4X4_50_ID_1
ARUCO_DICT_4X4_50_ID_2
ARUCO_DICT_4X4_100_ID_1
ARUCO_DICT_4X4_250_ID_1
ARUCO_DICT_7X7_50_ID_1
IR_BEACON_ID_1
IR_BEACON_ID_2
etc ...
```
- `custom_state` is this really necessary? This seems too device-specific
- `est_capabilities` is this used to determine which fields of the message are populated? e.g. lat/lon if bit 0 is set?
- `rates` RPY? How is this useful?
- `frame` are frames other than MAV_FRAME_LOCAL_NED actually useful?

Maybe rename to **TARGET_ABSOLUTE** and **TARGET_RELATIVE**
Maybe rename `yaw_var` to `q_var`
Maybe rename `q_att` to `q_target`
Maybe rename `q_sensor_att` to `q_sensor`

You will want to have both `q_att` and `q_sensor_att` regardless of the frame. As you stated above, you need to know the sensor pose and target pose at the time of measurement. 

Keep in mind you probably don't want your drone to follow or land on targets that are not yours or that are not part of the currently active operation. A drone should be able to dynamically change which target is associated with which task. Hence the need for a unique identifier as well as the task being dictated by the autopilot rather than the target.

JonasPerolini:
Thanks for the detailed review @dakejahl !

Renaming: 

- ok to rename `q_att` to `q_target` 
- ok to rename `q_sensor_att` to `q_sensor`
- ok to rename `ABSOLUTE_TARGET` and `RELATIVE_TARGET` to `TARGET_ABSOLUTE` and `TARGET_RELATIVE` respectively
- I prefer `yaw_var` as it is the uncertainty of the `yaw`, not the uncertainty of the quaternion

Field changes: 
- `custom_state`, `est_capabilities`, `rates` to be discussed with @potaito 
- `type`, `frame`, `id`, `task` should not change in my opinion, see detail below. Please let me know what you think.

> `type` How is this data useful? Is this data actionable?

Depending on the target `type`, you have to forward it to different messages e.g. irlock_report, fiducial_marker_report, ... which do not necessarily require the same processing.

> `id` isn't very useful on its own. If you were using aruco tags you'd want to know both the dictionary and the tag ID.

I agree, we cannot have the `id` be the ID of the marker for the specific detection algorithm (e.g. aruco marker 5 in dict 4x4) as every detection algorithm (e.g. Aruco, AprilTag) will be different. Detecting one specific tag but not another one, must be handled either off-board or in a driver. It has no place in a message in my opinion. The purpose of the `id` field is to allow the user to define an identifier  for a given target. In the future, we could run multiple KF instance in parallel i.e. track 2 different targets, and differentiate which one goes into which KF using the `id`. 

> `task` I don't think it makes sense for the target to know what its task is. 

For me the `task` is the **simplest** way to associate a target to an estimator (e.g. KF FOLLOW , KF LAND). It also allows for complex and unique off-board behaviors e.g. dynamically changing the task. Note that the `task` is only here to forward the observation to the target estimator, it does not dictate the behavior of the drone. One can track a landing target and a follow target at the same time.

- > If you want to use a target for a specific task this should be defined within the autopilot. For example use a parameter to associate the tag ID with the specific task.

- What if you want to follow a target and then land on it? You would have to associate this specific tag ID to both tasks somehow. Also the target estimator (or MAVLINK receiver) would have to check the list of given tag ids and compare them. It is  much simpler to change the `task` off-board in my opinion.

- > Keep in mind you probably don't want your drone to follow or land on targets that are not yours or that are not part of the currently active operation.

- Totally agree, but this should be done off-board or in a driver. It makes no sense to send a target that is not part of the active operation via MAVLINK, and rely on the target tracker to ignore it. In my opinion, the drone's job given a target is to track it (of course the drone must check that the values in the fields make sense, but not the target id).

- > A drone should be able to dynamically change which target is associated with which task. Hence the need for a unique identifier as well as the task being dictated by the autopilot rather than the target.

- Yes, it should be able to dynamically change which target is associated with which task, for example following a target and then landing on it. Why can't this work using the `task` field?

> `frame` are frames other than MAV_FRAME_LOCAL_NED actually useful?

Let's say you have an extremely fast detection algorithm in body frame which requires no processing time. Then you only need to fill the `MAV_FRAME` , and since you don't specify `q_sensor_att`, the current attitude will be used. Otherwise, the user somehow must get the attitude of the sensor off-board which is overkill as there is negligible processing time.


> `custom_state` is this really necessary? This seems too device-specific
`est_capabilities` is this used to determine which fields of the message are populated? e.g. lat/lon if bit 0 is set?
`rates` RPY? How is this useful?

I've added those to keep compatibility with the `FOLLOW_TARGET` msg. From my limited understanding, `rates` is used when following cars to include the steering wheel input. @potaito can provide a better explanation.

hamishwillee:
W.r.t. names we currently have:
- GPS data will be sent via TARGET_ABSOLUTE
- Fiducial marker data will be sent via TARGET_RELATIVE

So TARGET_ABSOLUTE is for targets that know their position in an absolute frame - such as GPS? TARGET_RELATIVE is where the target is in some frame relative to the vehicle, and where the location is generated from data/sensors on the vehicle?  That makes sense.

@dakejahl Can you respond to the response?

Would/will you use them? IF you were designing the messages would they look like this? Are there any missing use cases.  Are they awesome improvement.

I have nothing useful to add other than w.r.t. "I've added those to keep compatibility with the `FOLLOW_TARGET`" -  IMO we need to find out if anyone uses them or has a use case that requires them. Then look and see whether the use case is met in this case without them. If not, I'd suggest we prototype without them.**

auturgy:
@rishabsingh3003 can you have a look at this please

hamishwillee:
Thanks very much @JonasPerolini - and for taking the time to come to the MAV call and discuss. My summary of the discussion in [20230201-Dev-Meeting](https://github.com/mavlink/mavlink/wiki/20230201-Dev-Meeting)

- Need to change the addition to MAV_FRAME - it is inconsistent and that enum is a pain point. - which I can see you have already done.
- We need more review from domain experts - requested.
- We don't want you to wait on the MAVLink process - it will delay your project.
  - Please create your own fork and start development
  - Continue discussion on PR to get best possible messages.
  - Thanks for contributing!

dakejahl:
I think @JonasPerolini justifications for id/type/task make sense in the context of how he is using the targets and the estimator(s). One question I have is why do you need two separate estimators for a moving and a non-moving target? 

I haven't looked at the PX4 PR in detail (it's really quite large and full of math) but the new TARGET_RELATIVE message is definitely a big improvement over LANDING_TARGET and would be immediately beneficial. In my fork where I have improved precision landing I basically have to misuse the landing target message anyway. I can't give a much better review than that since I am not actively developing any kind of analogous system. I think the most important thing is that we document the implementation assumptions that go along with these messages. 

I think getting that PX4 PR merged is going to be a lot bigger challenge due to its complexity.

JonasPerolini:
Yess, @dakejahl it will definitely be a big challenge to merge the PX4 PR!

> why do you need two separate estimators for a moving and a non-moving target?

I'm using a Kalman filter which is composed of a prediction step (predict the evolution of the target) and an update step (correct the prediction using the observations). The prediction step is performed using the dynamical model of the target which depends on whether the target is moving or not. If your question was about the follow vs land tasks, eventually the goal will be to use the KF equations for both (and replace the current follow target filter). Even if they use the same equations, we would still have two estimators to be able to have a different tuning and different specifics.

dakejahl:
w.r.t PX4 PR, some food for thought ... something to consider is that simple usually trumps better (perfect is the enemy of good?). I don't know the specifics of your use case but if I were you I would try to reduce the software complexity rather than strive for optimal performance. Maybe an EKF can do both good enough. I have no idea what your performance objectives might be but in general simpler solutions are much easier to maintain and much more likely to get merged. 

Anyway back on topic, I think the messages look fine and are an overall improvement over the existing ones.

potaito:
@dakejahl 
> `type` How is this data useful? Is this data actionable?

It's a field similar to the `type` field of [DISTANCE_SENSOR](https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR) for example. Whether and how it's actionable is up to the receiver. You might want to know the target type in order to distinguish IR beacons from ComputerVision markers in cases where both are used. Particularly now since we determined that a generic `id` is not going to work.

Certain (static) target types might be used for improving the onboard estimator of the drone for example. For that I would want to know the type of target and whether it's moving or not.

@dakejahl 
> `task` I don't think it makes sense for the target to know what its task is. If you want to use a target for a specific task this should be defined within the autopilot. For example use a parameter to associate the tag ID with the specific task.

I share your opinion. As @JonasPerolini mentioned, this is how he can tell PX4 which estimator to use for this message. But given that MAVLink is more general than PX4, and that the PX4 implementation could be adopted, we probably need to find a better solution for the `task` field.

@dakejahl 
>  `id` isn't very useful on its own. If you were using aruco tags you'd want to know both the dictionary and the tag ID. One simple option would be to use an enum, this would have the benefit of defining "ID"s for things other than aruco tags which may have a different convention (for example IR beacons). Although would require an update every time someone wants to add a new definition. I'm open to suggestions but IMO just a `uint8_t id` is not enough information. Another option could be `id` and `dict` with different definitions depending on the `type` (so we'd need the `type` field in this case), but this increases the message size.

@JonasPerolini 
> I agree, we cannot have the id be the ID of the marker for the specific detection algorithm (e.g. aruco marker 5 in dict 4x4) as every detection algorithm (e.g. Aruco, AprilTag) will be different. Detecting one specific tag but not another one, must be handled either off-board or in a driver. It has no place in a message in my opinion

I disagree and think that an `id` field can be useful.

The idea would be that the sender of this message would assign a different ID to different targets. What exactly the underlying ID is, in your example the Aruco names, would not matter. As a receiver of this message I'd only want to know if the algorithm is seeing two different targets for example.

It's up to the application that publishes this message to decide whether it wants to use the `id` field and how it should be set. I don't think we can have a universal unique ID across all types of markers. But within one application, it should be possible, don't you think?

The `id` would not be directly mapped to Aruco tag IDs or April tag IDs. I.e. knowing `id` would not allow me to infer the unique ID of the April tag. But I'd like to be able to detect, when my camera is suddenly picking up a different marker than it did just a moment ago.

In fact I think that `id` could replace how @JonasPerolini is using `task` right now. I.e. to determine which of the two filters in his PX4 code is using which of the target information. 

@JonasPerolini 
> I'm using a Kalman filter which is composed of a prediction step (predict the evolution of the target) and an update step (correct the prediction using the observations). The prediction step is performed using the dynamical model of the target which depends on whether the target is moving or not. If your question was about the follow vs land tasks, eventually the goal will be to use the KF equations for both (and replace the current follow target filter). 

This implementation makes perfect sense. But then distinguishing the job for the two estimators is an implementation challenge, not a protocol problem.

@JonasPerolini 
> Even if they use the same equations, we would still have two estimators to be able to have a different tuning and different specifics.

Yes, but in the end the same could be achieved with only the `id` field, no? Let's put PX4 aside and assume we just have a flying computer that can run estimators. It's then up to us and our implementation whether we want to run two estimators in total, or ten. Perhaps I'd want to track multiple targets of the same type and purpose simultaneously. That's why I think the `task` field as part of the MAVLink message is too specific and makes assumptions based on the current PX4 implementation. 



@dakejahl 
>  `custom_state` is this really necessary? This seems too device-specific

Yeah, this can probably be a separate message if necessary.

@dakejahl 
>  `est_capabilities` is this used to determine which fields of the message are populated? e.g. lat/lon if bit 0 is set?

The existing [FOLLOW_TARGET](https://mavlink.io/en/messages/common.html#FOLLOW_TARGET) message has fields for the velocity for example, and specified that a velocity of 0.0 means "unknown/undefined". Same for acceleration, where (0,0,0) means "unknown". This works in theory, assuming that an estimator will never output 0.0 for velocity or acceleration due to noise. This turned out to not be the case in reality though. I found that mobile devices such as iPhone and Android tend to set the velocity and acceleration to 0.0 when the phone's estimator is certain enough, that the phone is not moving significantly. And then it's impossible to tell whether the phone is estimating these fields and the device is actually standing still, or whether the estimator does not support these fields. It's ambiguous. If we keep the field `est_capabilities`, we need to remove `(0 0 0 for unknown)` from the message descriptions in this PR.

@dakejahl 
>  `rates` RPY? How is this useful?

Rates can help to do a better job estimating the motion of the target. There are devices such as cars where you can actually estimate the turning rate quite nicely. And if you know that the target is a car, and you know the turning rate, you can model the motion. Without rate you still can, but you'll have to take the derivative of the heading.

Android and iOS on the other hand "only" estimate the heading, not the rates.


@dakejahl 
> Keep in mind you probably don't want your drone to follow or land on targets that are not yours or that are not part of the currently active operation. A drone should be able to dynamically change which target is associated with which task. Hence the need for a unique identifier as well as the task being dictated by the autopilot rather than the target.

But that's not a problem that these messages want to solve. These messages should convey the information about targets that the camera (or other sensor) detects.

JonasPerolini:
If I summarize, @potaito you would want to: 

1. remove `custom_state`. ok for me

1. remove `(0 0 0 for unknown)` ok for me

1. remove the `task` and only use the `id` instead, arguing that the `task` is too specific, and `id` allows to run several estimators with several functions in parallel e.g. follow, land, ... . I understand your point, but why do `id` and `task` must be mutually exclusive? Can't we have both? I'm not against having the target `id` as I said above, it makes sense if you want to run multiple estimators in parallel, (or as you mention above, if you get a new target, you might want to start a new estimator). However, the identification of a target is different than its intended purpose. I feel like you're trying to fuse two different pieces of information in one message. For me a target has one `id` which is not related to its purpose as each target can have several purposes e.g. follow and then land. 
Lets say you only have an `id` field, how would you relate each id to a task? Haven't thought about it too much, but I don't see a simple solution (If you have a practical one, please let me know): 

    - Option 1: a param to relate one `id` to a task e.g. `id 1 == LAND`, `id 2 == FOLLOW`, `id 3 == DO NOTHING`. Then you cannot have several KFs in parallel for the same task because the id becomes the task.

    - Option 2: relate several ids to a task e.g.  `id 1,2,3,4 == LAND` but then you would have to compare each id to the corresponding list and the user would have to make sure to enter the proper list. Additionally, if you want to change the task of a marker, you would have to change the param list on the fly. Why wouldn't all of this be done by the user off-board or in a driver? There are not so many possible tasks for a target, surely we can group them and list them in an enum no?

I do agree that the `task` field is very specific to my problem. What if we add: `task = LOCALIZATION` , `task = NONE` and `task = OTHER` ?

To summarize, 
1. Why not keep both `id` and `task`?

1. If we keep `task` and make it more general would that work for you?

potaito:
>  I understand your point, but why do `id` and `task` must be mutually exclusive? Can't we have both?

They don't have to be. I was just pointing out, that the problem specifically in PX4 could be solved this way. I.e. you have two Kalman Filters in PX4. If they are configurable via parameters, one might be set to track only `id=1` and the other to track only `id=2`. And, if the user wishes so, he could swap the two filters by swapping the parameters. In this case the publisher of this mavlink message should not need to know anything about it.

I'm on the fence whether a target should publish "I'm a landing target" or "I'm a follow-target". Theoretically this would solve the problem, and it would just be a rename of `task`.  But my points here are, that
1. a target can be both, a landing target and a follow-target. And perhaps in the future they might also be localization helper targets for indoor applications, all at the same time.
2. the drone should decide what it wants to do with the targets, not the target

In my opinion we are looking at a message similar to a sensor measurement. As such it should say what it is, and not how it's supposed to be used.

That's why I thought, that in your specific case with the thesis you could use the `id` in whatever way you want and accomplish the same results. But maybe I'm missing something. Let's see what other opinions have to say.

JonasPerolini:
Thanks for the precision @potaito ! 

> In my opinion we are looking at a message similar to a sensor measurement. As such it should say what it is, and not how it's supposed to be used.

From this perspective, it makes a lot of sense, I totally agree. Ultimately it is a sensor message and should be treated as such. I'm ok with removing the `task` field. It also allows to re-direct the observation towards a given task once it has been processed and estimated. (e.g. if you want to follow a target and then land on it, the KF estimation should not change throughout the procedure, just that at some point you switch from follow to land). Note that the implementation will be more complex and the `id` field might be somewhat confusing to the user if not properly documented. 
 
The way I see it In PX4, the mavlink receiver just forwards the current observation to the corresponding Uorb report. Each estimator will track a given `id` and report its estimation. Then we somehow (e.g. param) link one estimator to a given task. For this implementation, the `id` "becomes" the `task` as each `id` is associated to one estimator which is associated to one `task`.
 The following argument still holds but I don't see why you would run several estimators in parallel for a given task.
 > Option 1: a param to relate one id to a task e.g. id 1 == LAND, id 2 == FOLLOW, id 3 == DO NOTHING. Then you cannot have several KFs in parallel for the same task because the id becomes the task.
 
 Overall, I'm ok with removing the `task` following the sensor observation argument. As @dakejahl  was also in favor to use the `id` , I will make the change.

JonasPerolini:
Changes made in https://github.com/mavlink/mavlink/pull/1943/commits/6770c21b2a015b2e7101f5842d3dce529f986f0f . I believe that we have resolved every conversation. @potaito and @dakejahl could you please confirm that you agree with the current implementation? 

We're just missing the review from @rishabsingh3003

hamishwillee:
@dakejahl @potaito Thanks for your responses! This has changed (a little) again. Are you both still essentially happy with them?

Yuu mentioned that we need to capture the reasoning behind the fields - do they do that sufficiently. I'm not going to be able to do that as a post process.

auturgy:
From @rishabsingh3003 "My views are that as long as its a separate message and they aren't replacing the existing message, its okay. I do understand the requirement for the additional fields to improve LANDING_TARGET, especially the kalman filter that runs behind it. So a new message is probably justified"

hamishwillee:
Discussed in the MAVLink call last night. Seems to be a general consensus that the existing LANDING_TARGET is flawed, and that these messages, in particular TARGET_RELATIVE, are a big improvement.

So waiting on final confirmation from @dakejahl @potaito and then can merge into development.

Note that this does not mean they will become part of the standard. They don't move to common until there is a prototype and enough testing we're happy this does what it needs to do.

potaito:
@rishabsingh3003 I think this part of Jonas' PR description describes the gist of the problem:

> However, the main difference between these messages is not the task, but rather the physical location of the sensor used. [FOLLOW_TARGET](https://mavlink.io/en/messages/common.html#FOLLOW_TARGET) reports observations from a sensor on the target (e.g. GPS or IMUs), while [LANDING_TARGET](https://mavlink.io/en/services/landing_target.html) reports observations from an onboard sensor (e.g. vision-based fiducial marker detection or irlock beacon detection). One might want to follow a target using a fiducial marker (LANDING_TARGET msg). Inversely, one might want to land on a target with a GPS attached (FOLLOW_TARGET msg). Each message should thus be usable independently of the task. Additionally, the new target estimator allows to fuse several measurements such as GPS, fiducial markers and irlock beacons in a Kalman Filter to estimate the target's state (for static and moving targets) https://github.com/PX4/PX4-Autopilot/pull/20653. The same filter class can then be used for precision landing and the target following tasks, reinforcing the idea that both messages should be used in parallel.

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +359,37 @@
       <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
     </message>
+    <!-- Target info from a sensor on the target -->
+    <message id="510" name="TARGET_ABSOLUTE">
+      <description>Current motion information from sensors on a target</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time).</field>
+      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
```
potaito, line 366: @hamishwillee throughout this PR we established, that the ID of a target or tag is purely specified by the sender of the message, and that it's not a universally unique ID. Would it make sense to call this the instance ID like in other messages that have multiple instances?

Very strictly speaking I think it would be wrong, because a separate instance of the TARGET_ABSOLUTE message would indicate that it's sent from a different sensor. Or does that not matter?

I'm asking so that we re-use as many concepts as possible without inventing new things that are not 100% well defined.

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +359,37 @@
...
+      <field type="uint8_t" name="est_capabilities">bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT + RATES = 3)</field>
```
potaito, line 367: @JonasPerolini should the estimator capabilities also include a flag for the orientation quaternion `q_target`? The way I interpret it now is that `ATT + RATES` refers to `q_target` and `rates` at the same time.

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +359,37 @@
...
+      <field type="int32_t" name="lat" units="degE7">Target's latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Target's longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Target's altitude (AMSL)</field>
+      <field type="float[3]" name="vel" units="m/s" invalid="[0]">target's velocity in its body frame</field>
+      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">linear target's acceleration in its body frame</field>
+      <field type="float[4]" name="q_target" invalid="[0]">Quaternion of the target's orientation from its body frame to NED.</field>
```
potaito, line 373:
```suggestion
      <field type="float[4]" name="q_target" invalid="[0]">Quaternion of the target's orientation from its body frame to the vehicle's NED frame.</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +359,37 @@
...
+      <field type="float[3]" name="rates" units="rad/s" invalid="[0]">Target's roll, pitch and yaw rates</field>
+      <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
+      <field type="float[3]" name="vel_var" units="[m/s]^2">Variance of the target's velocity in its body frame</field>
+      <field type="float[3]" name="acc_var" units="[m/s/s]^2">Variance of the target's acceleration in its body frame</field>
+    </message>
+    <!-- Target info from an onboard sensor -->
+    <message id="511" name="TARGET_RELATIVE">
+      <description>The location of a target from an onboard sensor. </description>
```
potaito, line 381:
```suggestion
      <description>The location of a target measured by MAV's onboard sensors. </description>
```

message_definitions/v1.0/development.xml
```diff
@@ -240,6 +240,18 @@
...
     </enum>
+    <enum name="TARGET_OBS_FRAME">
+      <description>The frame of a target observation from an onboard sensor.</description>
+      <entry value="0" name="LOCAL_NED">
+        <description>NED local tangent frame (x: North, y: East, z: Down) with origin fixed relative to earth.</description>
+      </entry>
+      <entry value="1" name="BODY_FRD">
+        <description>FRD local frame aligned to the vehicle's attitude (x: Forward, y: Right, z: Down) with an origin that travels with vehicle.</description>
+      </entry>
+      <entry value="2" name="GENERIC">
```
potaito, line 251: @JonasPerolini how does the GENERIC frame work?

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +359,37 @@
...
+      <field type="float" name="x" units="m">X Position of the target in TARGET_OBS_FRAME (For IR-Lock: x-axis distance from the center of image to the center of the target in units of tan(theta))</field>
+      <field type="float" name="y" units="m">Y Position of the target in TARGET_OBS_FRAME (For IR-Lock: y-axis distance from the center of image to the center of the target in units of tan(theta))</field>
+      <field type="float" name="z" units="m">Z Position of the target in TARGET_OBS_FRAME</field>
+      <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in TARGET_OBS_FRAME (var_x, var_y, var_z)</field>
+      <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in TARGET_OBS_FRAME</field>
+      <field type="float[4]" name="q_target">Quaternion of the target's orientation from the target's frame to the TARGET_OBS_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
+      <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Not used if TARGET_OBS_FRAME is in LOCAL_NED) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
```
potaito, line 391:
```suggestion
      <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if TARGET_OBS_FRAME is in LOCAL_NED) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +359,37 @@
...
+      <field type="float[3]" name="vel_var" units="[m/s]^2">Variance of the target's velocity in its body frame</field>
+      <field type="float[3]" name="acc_var" units="[m/s/s]^2">Variance of the target's acceleration in its body frame</field>
+    </message>
+    <!-- Target info from an onboard sensor -->
+    <message id="511" name="TARGET_RELATIVE">
+      <description>The location of a target from an onboard sensor. </description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time)</field>
+      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
+      <field type="uint8_t" name="frame" enum="TARGET_OBS_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x" units="m">X Position of the target in TARGET_OBS_FRAME (For IR-Lock: x-axis distance from the center of image to the center of the target in units of tan(theta))</field>
```
potaito, line 385: I don't want to keep treating IRLocks differently than other measurement / sensor types :smile: 
Do we really have to?

I think this is a convenience problem since the Arducams used for Precision Landing using IR Beacons only output the coordinates of the beacon in the sensor frame. The important question is: Are we expecting a camera to send this new mavlink message? Or are we expecting the camera to be connected to a 2nd system that has a driver for the camera and then sends the mavlink message to the Autopilot, if it's not the same system.

For example, if the Arducam is connected to the Autopilot directly, the mavlink message is not necessary since the Autopilot will have a I2C or whatever driver to get the coordinates from the camera.

potaito: Another way to look at this is what the difference between fiducial markers and IR beacons are. Both are seen by a camera and then the XY(Z) position is estimated. Why does IR beacon need special treatment?

potaito (commented):
Generally I think this is it. My comments are mostly about linguistics. I'm also still trying to find a better name for `TARGET_ABSOLUTE` and `TARGET_RELATIVE`, but can't think of anything better.

What about `TARGET_GLOBAL` and `TARGET_LOCAL`?

message_definitions/v1.0/development.xml
JonasPerolini: It's any other sensor frame. Do you want me to change it to "OTHER" ? @potaito

message_definitions/v1.0/development.xml
JonasPerolini: Yes, makes sense. I'm ok with removing the parenthesis. Does this work for you? @potaito

message_definitions/v1.0/development.xml
JonasPerolini: Yes, I've modified to: `bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT = 3, ATT + RATES = 4)`

message_definitions/v1.0/development.xml
potaito: Yes, if you confirm that I didn't miss something and didn't write nonsense :smile:

message_definitions/v1.0/development.xml
potaito: I wonder how I would use this information as a receiver of the message. Is it an undefined frame?

message_definitions/v1.0/development.xml
potaito: Nice, thanks :+1:

message_definitions/v1.0/development.xml
potaito: Ah, now I see the mention in the meta information of the quaternion field, and that it's only used in `TARGET_OBS_FRAME` :+1: 

Thanks!

message_definitions/v1.0/development.xml
JonasPerolini: Wait, since it's the bit positions, it should be: `bit positions for tracker reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT = 3, RATES = 4` right @potaito ? 

E.g. to have pos and vel you set it to 3 (bit 0 and bit 1)

JonasPerolini:
Here's how I would use the Mavlink messages for the Kalman filter. (The `id` field is ignored, but it can be added once we have several estimators)

```
void
MavlinkReceiver::handle_message_target_absolute(mavlink_message_t *msg)
{
	mavlink_target_absolute_t target_absolute;
	mavlink_msg_target_absolute_decode(msg, &target_absolute);

	landing_target_gnss_s target_GNSS_report{};
	target_GNSS_report.timestamp = target_absolute.timestamp;

	bool updated = false;

	// Position: bit 0
	if (!(target_absolute.est_capabilities & (1 << 0))) {
		target_GNSS_report.abs_pos_updated = false;

	} else {
		target_GNSS_report.abs_pos_updated = true;
		target_GNSS_report.lat = target_absolute.lat;
		target_GNSS_report.lon = target_absolute.lon;
		target_GNSS_report.alt = target_absolute.alt * 1000.f; //Convert to [mm]
		target_GNSS_report.eph = target_absolute.position_std[0];
		target_GNSS_report.epv = target_absolute.position_std[1];
		updated = true;
	}

	// Velocity: bit 1
	if (!(target_absolute.est_capabilities & (1 << 1))) {
		target_GNSS_report.vel_ned_updated = false;

	} else {
		target_GNSS_report.vel_ned_updated = true;
		target_GNSS_report.vel_n_m_s = target_absolute.vel[0];
		target_GNSS_report.vel_e_m_s = target_absolute.vel[1];
		target_GNSS_report.vel_d_m_s = target_absolute.vel[2];
		target_GNSS_report.s_variance_m_s = target_absolute.vel_std[0];
		updated = true;
	}

	if (updated) { _landing_target_gnss_pub.publish(target_GNSS_report); }
}
```

```
void
MavlinkReceiver::handle_message_target_relative(mavlink_message_t *msg)
{
	mavlink_target_relative_t target_relative;
	mavlink_msg_target_relative_decode(msg, &target_relative);

	vehicle_attitude_s	vehicle_attitude;
	vehicle_local_position_s vehicle_local_position;

	bool publish_target = false;
	matrix::Quatf q_sensor(target_relative.q_sensor);

	// Unsupported sensor frame
	if (target_relative.frame != TARGET_OBS_FRAME_LOCAL_OFFSET_NED && target_relative.frame != TARGET_OBS_FRAME_LOCAL_NED &&
	    target_relative.frame != TARGET_OBS_FRAME_BODY_FRD && target_relative.frame != TARGET_OBS_FRAME_OTHER) {
		mavlink_log_critical(&_mavlink_log_pub, "landing target: coordinate frame %" PRIu8 " unsupported.\t",
				     target_relative.frame);
		events::send<uint8_t>(events::ID("mavlink_rcv_target_rel_unsup_coord_frame"), events::Log::Error,
				      "Target relative: unsupported coordinate frame {1}", target_relative.frame);

	} else {
		/* Check if the sensor's attitude field is filled. */
		if ((abs(q_sensor(0)) + abs(q_sensor(1)) + abs(q_sensor(2)) + abs(q_sensor(3))) > (float)1e-6) {
			publish_target = true;

		} else {
			/* Fill q_sensor depending on the frame*/
			if (target_relative.frame == TARGET_OBS_FRAME_LOCAL_OFFSET_NED || target_relative.frame == TARGET_OBS_FRAME_LOCAL_NED) {
				// Observation already in NED, set quaternion to identity (1,0,0,0)
				q_sensor.setIdentity();
				publish_target = true;

			} else if (target_relative.frame == TARGET_OBS_FRAME_BODY_FRD && _vehicle_attitude_sub.update(&vehicle_attitude)) {
				// Set the quaternion to the current attitude
				const matrix::Quaternionf quat_att(&vehicle_attitude.q[0]);
				q_sensor = quat_att;
				publish_target = true;

			} else if (target_relative.frame == TARGET_OBS_FRAME_OTHER) {
				// Target sensor frame: OTHER and no quaternion is provided
				mavlink_log_critical(&_mavlink_log_pub,
						     "landing target: coordinate frame %" PRIu8 " unsupported when the q_sensor field is not filled.\t",
						     target_relative.frame);
				events::send<uint8_t>(events::ID("mavlink_rcv_target_rel_q_sensor_not_filled"), events::Log::Error,
						      "Target relative: unsupported coordinate frame {1} when the q_sensor field is not filled", target_relative.frame);
			}
		}
	}

	if (publish_target) {
		int32_t ltest_enabled;
		param_get(param_find("LTEST_EN"), &ltest_enabled);

		if (!ltest_enabled) {

			// If the landing target estimator is disabled, send the target directly to the precision landing algorithm in local NED frame
			landing_target_pose_s landing_target_pose{};

			// Measurement already in local NED
			if (target_relative.frame == TARGET_OBS_FRAME_LOCAL_NED) {

				landing_target_pose.timestamp = _mavlink_timesync.sync_stamp(target_relative.timestamp);
				landing_target_pose.x_abs = target_relative.x;
				landing_target_pose.y_abs = target_relative.y;
				landing_target_pose.z_abs = target_relative.z;
				landing_target_pose.abs_pos_valid = true;

				_landing_target_pose_pub.publish(landing_target_pose);

			} else if (_vehicle_local_position_sub.update(&vehicle_local_position) && vehicle_local_position.xy_valid) {

				// If the local position is available, convert from sensor frame to local NED
				matrix::Vector3f target_relative_frame(target_relative.x, target_relative.y, target_relative.z);

				if (target_relative.frame == TARGET_OBS_FRAME_BODY_FRD || target_relative.frame == TARGET_OBS_FRAME_OTHER) {
					// Rotate measurement into vehicle-carried NED
					target_relative_frame = q_sensor.rotateVector(target_relative_frame);
				}

				// Convert from vehicle-carried NED to local NED
				landing_target_pose.x_abs = target_relative_frame(0) + vehicle_local_position.x;
				landing_target_pose.y_abs = target_relative_frame(1) + vehicle_local_position.y;
				landing_target_pose.z_abs = target_relative_frame(2) + vehicle_local_position.z;
				landing_target_pose.abs_pos_valid = true;

				_landing_target_pose_pub.publish(landing_target_pose);
			}

		} else {

			// Send target to the landing target estimator
			if (target_relative.type == LANDING_TARGET_TYPE_LIGHT_BEACON) {

				irlock_report_s irlock_report{};

				irlock_report.timestamp = hrt_absolute_time();
				irlock_report.pos_x = target_relative.x;
				irlock_report.pos_y = target_relative.y;

				q_sensor.copyTo(irlock_report.q);
				_irlock_report_pub.publish(irlock_report);

			} else if (target_relative.type == LANDING_TARGET_TYPE_VISION_FIDUCIAL) {

				// Position report
				fiducial_marker_pos_report_s fiducial_marker_pos_report{};

				fiducial_marker_pos_report.timestamp = _mavlink_timesync.sync_stamp(target_relative.timestamp);
				fiducial_marker_pos_report.x_rel_body = target_relative.x;
				fiducial_marker_pos_report.y_rel_body = target_relative.y;
				fiducial_marker_pos_report.z_rel_body = target_relative.z;

				fiducial_marker_pos_report.cov_x_rel_body = target_relative.pos_var[0];
				fiducial_marker_pos_report.cov_y_rel_body = target_relative.pos_var[1];
				fiducial_marker_pos_report.cov_z_rel_body = target_relative.pos_var[2];

				q_sensor.copyTo(fiducial_marker_pos_report.q);
				_fiducial_marker_pos_report_pub.publish(fiducial_marker_pos_report);

				// Yaw report
				fiducial_marker_yaw_report_s fiducial_marker_yaw_report{};
				fiducial_marker_yaw_report.timestamp = _mavlink_timesync.sync_stamp(target_relative.timestamp);

				// Transform quaternion from the target's frame to the TARGET_OBS_FRAME to the yaw relative to NED
				const matrix::Quatf q_target(target_relative.q_target);
				const matrix::Quatf q_in_ned = q_sensor * q_target;
				const float target_yaw_ned = matrix::wrap_pi(matrix::Eulerf(q_in_ned).psi());

				fiducial_marker_yaw_report.yaw_ned = target_yaw_ned;
				fiducial_marker_yaw_report.yaw_var_ned = target_relative.yaw_var;
				_fiducial_marker_yaw_report_pub.publish(fiducial_marker_yaw_report);

			}
		}
	}
}
```

message_definitions/v1.0/development.xml
JonasPerolini: I've modified the name to `sensor_capabilities` to avoid confusions because this message is for the sensor on the target i.e. the estimator may estimate the accel. but the sensor does not observe the accel. `      <field type="uint8_t" name="sensor_capabilities">bit positions for the sensor's reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT = 3, RATES = 4)</field>` ok for you @potaito ?

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +362,37 @@
       <field type="uint8_t" name="base_mode" enum="MAV_MODE_FLAG" display="bitmask">System mode bitmap.</field>
       <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
     </message>
+    <!-- Target info from a sensor on the target -->
+    <message id="510" name="TARGET_ABSOLUTE">
+      <description>Current motion information from sensors on a target</description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time).</field>
+      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
+      <field type="uint8_t" name="sensor_capabilities">bit positions for the sensor's reporting capabilities (POS = 0, VEL = 1, ACCEL = 2, ATT = 3, RATES = 4)</field>
```
potaito, line 370:
```suggestion
      <field type="uint8_t" name="sensor_capabilities" enum "TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAG">bit positions for the sensor's reporting capabilities</field>
```

And then add the enum on top of development.xml.
```
<enum name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAG">
  <description>Sensor reporting capabilities for TARGET_ABSOLUTE</description>
  <entry value="1" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_POSITION">
    <description>...</description>
  </entry>
  <entry value="2" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_VELOCITY">
    <description>...</description>
  </entry>
  <entry value="4" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_ACCELERATION">
    <description>...</description>
  </entry>
  <entry value="8" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_ATTITUDE">
    <description>...</description>
  </entry>
  <entry value="16" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_RATES">
    <description>...</description>
  </entry>
</enum>
```

@hamishwillee please confirm :smile:

potaito, line 370:
```suggestion
      <field type="uint8_t" name="sensor_capabilities" enum "TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAG">bit positions for the sensor's reporting capabilities</field>
```

And then add the enum on top of development.xml.
```
<enum name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAG">
  <description>Sensor reporting capabilities for TARGET_ABSOLUTE</description>
  <entry value="1" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_POSITION">
    <description>...</description>
  </entry>
  <entry value="2" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_VELOCITY">
    <description>...</description>
  </entry>
  <entry value="4" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_ACCELERATION">
    <description>...</description>
  </entry>
  <entry value="8" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_ATTITUDE">
    <description>...</description>
  </entry>
  <entry value="16" name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_RATES">
    <description>...</description>
  </entry>
</enum>
```

@hamishwillee please confirm :smile:

JonasPerolini: I've implemented the flag enum following the structure from `ADSB_FLAGS` @potaito

JonasPerolini:
Status: every conversation is resolved, there is an [example of use case](https://github.com/mavlink/mavlink/pull/1943#issuecomment-1436114560).

For me (and @potaito I believe) it's ready to be merged in dev once @hamishwillee confirms the following changes: [id name](https://github.com/mavlink/mavlink/pull/1943#discussion_r1108405603)  and[ sensor capabilities flags](https://github.com/mavlink/mavlink/pull/1943#discussion_r1111616059).

potaito:
> Status: every conversation is resolved, there is an [example of use case](https://github.com/mavlink/mavlink/pull/1943#issuecomment-1436114560).
> 
> For me (and @potaito I believe) it's ready to be merged in dev once @hamishwillee confirms the following changes: [id name](https://github.com/mavlink/mavlink/pull/1943#discussion_r1108405603) and[ sensor capabilities flags](https://github.com/mavlink/mavlink/pull/1943#discussion_r1111616059).

yep :blush:

JonasPerolini:
Minor change from variance to standard deviation because as in EKF2, most applications will have a lower range for the standard deviation of the measurement.

message_definitions/v1.0/development.xml
hamishwillee: @potaito Many apologies - missed this. Feel free to email me if you don't get a response on anything in a week - that means I probably missed it. Email is reliable.

My feeling on this is that `TARGET_ABSOLUTE` is not an instance but the ID on TARGET_RELATIVE IS an instance. I don't mind whether or not you refer to it as such, but on `TARGET_RELATIVE.id` we should add this attribute `instance="true"`.

Why? Well the instance ID is used to differentiate multiples of what appear to be the same thing _from a MAVLink perspective_. Each of the TARGET_ABSOLUTE messages has a unique sender system/component ID so you a GCS can display a plot of the values separately. However TARGET_RELATIVE all originate from the same system; to differentiate these a GCS will need to know that the same message comes from different "things", which is what that attribute is supposed to do.

On `TARGET_ABSOLUTE` I think calling it an ID is better in any case - it isn't a part of something else. A bigger question is whether we will ever need more than 255 targets?

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +370,37 @@
...
+      <field type="float[3]" name="rates" units="rad/s" invalid="[0]">Target's roll, pitch and yaw rates</field>
+      <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
+      <field type="float[3]" name="vel_std" units="[m/s]">Standard deviation of the target's velocity in its body frame</field>
+      <field type="float[3]" name="acc_std" units="[m/s/s]">Standard deviation of the target's acceleration in its body frame</field>
+    </message>
+    <!-- Target info measured by MAV's onboard sensors -->
+    <message id="511" name="TARGET_RELATIVE">
+      <description>The location of a target measured by MAV's onboard sensors. </description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time)</field>
+      <field type="uint8_t" name="id">The ID of the target if multiple targets are present</field>
```
hamishwillee, line 394:
```suggestion
      <field type="uint8_t" name="id" instance="true">The ID of the target if multiple targets are present</field>
```

message_definitions/v1.0/development.xml
hamishwillee: Excellent. I added suggestion for adding instance to target relative. Anything else outstanding on this "for now"?

potaito: I don't think so. I'm now prototyping the interface in MAVSDK, to make sure I understand how to use the new messages

message_definitions/v1.0/development.xml
hamishwillee: Yes, this was good.

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +370,37 @@
...
+      <field type="int32_t" name="lat" units="degE7">Target's latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Target's longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Target's altitude (AMSL)</field>
+      <field type="float[3]" name="vel" units="m/s" invalid="[0]">Target's velocity in its body frame</field>
+      <field type="float[3]" name="acc" units="m/s/s" invalid="[0]">Linear target's acceleration in its body frame</field>
+      <field type="float[4]" name="q_target" invalid="[0]">Quaternion of the target's orientation from its body frame to the vehicle's NED frame.</field>
+      <field type="float[3]" name="rates" units="rad/s" invalid="[0]">Target's roll, pitch and yaw rates</field>
+      <field type="float[2]" name="position_std" units="[m]">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
+      <field type="float[3]" name="vel_std" units="[m/s]">Standard deviation of the target's velocity in its body frame</field>
+      <field type="float[3]" name="acc_std" units="[m/s/s]">Standard deviation of the target's acceleration in its body frame</field>
```
hamishwillee, line 388: Failing tests. Units must match schema https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd

```suggestion
      <field type="float[2]" name="position_std" units="m">Standard deviation of horizontal (eph) and vertical (epv) position errors</field>
      <field type="float[3]" name="vel_std" units="m/s">Standard deviation of the target's velocity in its body frame</field>
      <field type="float[3]" name="acc_std" units="m/s/s">Standard deviation of the target's acceleration in its body frame</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -347,5 +370,37 @@
...
+    <message id="511" name="TARGET_RELATIVE">
+      <description>The location of a target measured by MAV's onboard sensors. </description>
+      <field type="uint64_t" name="timestamp" units="us">Timestamp (UNIX epoch time)</field>
+      <field type="uint8_t" name="id" instance="true">The ID of the target if multiple targets are present</field>
+      <field type="uint8_t" name="frame" enum="TARGET_OBS_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x" units="m">X Position of the target in TARGET_OBS_FRAME</field>
+      <field type="float" name="y" units="m">Y Position of the target in TARGET_OBS_FRAME</field>
+      <field type="float" name="z" units="m">Z Position of the target in TARGET_OBS_FRAME</field>
+      <field type="float[3]" name="pos_var" units="m^2">Variance of the target's position in TARGET_OBS_FRAME</field>
+      <field type="float" name="yaw_var" units="rad^2">Variance of the target's orientation in TARGET_OBS_FRAME</field>
```
hamishwillee, line 400: Units must match schema https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L92-L122

Neither m^2 or rad^2 are supported (cm^2 is supported but might not be sufficient for the m case). Can you submit a PR to the schema doc linked to add the values you need?

hamishwillee:
We agreed in the mavlink call last night that we are happy for this to be merged for prototyping. I'm happy to do this, but we'll need to get the schema updated to allow some of the new units - see https://github.com/mavlink/mavlink/pull/1943/files#r1137815803

Once that is done, please ping me/cross link and I'll see if I can expedite that.

kripper:
Hi,

I'm the current maintainer of [RossetaDrone](https://github.com/The1only/rosettadrone): a MAVLink 2 implementation for DJI drones and a "flightcontroller layer" built on top of DJI's SDK.

I just released [Vision Landing 2](https://github.com/kripper/vision-landing-2) and was now implementing the code for landing on the target when I felt that the MAVLink message definition was a bit too limited given the many different options and ideas that could/should be applied to accomplish a fast and smooth landing on the target:
https://github.com/The1only/rosettadrone/issues/132

We would like to contribute to this topic and implement something standard and well defined from the beginning (good practices).
I believe the only difference in our use case is that we additionaly have a closed-source positioning system provided by DJI that is **out of our control** and that we may or may not use for landing (I have a hunch that visual computing may be more accurate or reliable).
Besides that, our use case should be the same, so we would like to work on and help to maintain a shared code base/model. I'm 100% dedicated to this topic, so please give me your feedback or let me know how I can contribute.

I'm now reading more in detail all your previous comments and progress.

potaito:
Hey @kripper 
The best thing you can do is read the proposed message definitions and check
- whether you understand them / they make sense to you
- whether they would support your use case without limitations

hamishwillee:
@kripper Your contribution would be very welcome - we'd like to get this right for as many use cases as possible. As @potaito says, the best way to start this is to review the messages, check that you understand them, and that they meet your use case.

If they do not meet your requirements, you can then either add comments to this PR, or if it is closed, add a follow on PR or issues that expand on the problems you see.

Note that they are likely to be broader than your use case because we're looking at all sorts of targets, not just vision. 

I had a very brief look at https://github.com/The1only/rosettadrone/issues/132 . I don't have the expertise to add useful comments sorry.

kripper:
Target IDs: we use multiple markers for 1 landing point (target). It makes sense that the target id is defined by the user.

A landing yaw is important (auto charging).

Do we have a draft defining good practices for the precision landing?

hamishwillee:
> Do we have a draft defining good practices for the precision landing?

Not within MAVLink. The only documentation I know of that might touch on this is the flight stack user guides (which are very much "how do you get it working":
- https://docs.px4.io/main/en/advanced_features/precland.html
- https://ardupilot.org/copter/docs/precision-landing-with-irlock.html

As part of accepting these messages into common we will need good docs explaining the use of the interfaces.

potaito:
@kripper 

> A landing yaw is important (auto charging).

That would be part of a precision landing command message. I.e. the messages in this PR communicate the position of the marker / tag. And another message that tells the drone to perform precision landing would ideally convey this yaw angle.

> Do we have a draft defining good practices for the precision landing?

Not yet. This PR is half of the equation and aims to get the messages for the measurements / observations right this time. In the next step, we could look at whether a new command message needs to be added.

message_definitions/v1.0/development.xml
JonasPerolini: Hi @hamishwillee, I've updated the messages to use the standard deviation ([m],[rad]) instead of the variance.

JonasPerolini:
Hi @hamishwillee, I've updated the messages to use the standard deviation ([m],[rad]) instead of the variance. I see that you've already pushed the other changes, thanks!

Anything else needed?

hamishwillee:
@JonasPerolini Thanks, no this is good - merged.

@potaito @JonasPerolini Keep me informed please. Very pleased if we can verify these do what we need with hard data and get documentation / tests in place that allow us to make sure that it is possible to create compatible implementations.

potaito:
Thanks, that's great!

PR #1944, opened 2023-01-22
minimal: add `MAV_TYPE`  `MAV_TYPE_GENERIC_MULTIROTOR`

IamPete1:
GCS's typically use `MAV_TYPE` to tailor the interface to a specific vehicle type, if the vehicle has not yet been configured, or has a unusual type that does not fall in to the provided types the existing `TYPE_GENERIC` is not specific enough to give the GCS a clue about what sort of vehicle it is. 

https://github.com/ArduPilot/ardupilot/pull/22658, https://github.com/ArduPilot/ardupilot/issues/22657

On AP this can happen during initial setup when the frame class is not setup. Or if a frame type is setup that falls outside of the given types, one could use scripting to setup a vehicle with 5 rotors for example.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -200,6 +200,9 @@
       <entry value="42" name="MAV_TYPE_WINCH">
         <description>Winch</description>
       </entry>
+      <entry value="43" name="MAV_TYPE_GENERIC_MULTIROTOR">
+        <description>Generic multirotor that does not fit into a specific type or whose type is unknown</description>
+      </entry>
     </enum>
     <enum name="MAV_MODE_FLAG" bitmask="true">
       <description>These flags encode the MAV mode.</description>


message_definitions/v1.0/minimal.xml
```diff
@@ -200,6 +200,9 @@
       <entry value="42" name="MAV_TYPE_WINCH">
         <description>Winch</description>
       </entry>
+      <entry value="43" name="MAV_TYPE_GENERIC_MULTIROTOR">
```
hamishwillee, line 203: @IamPete1 Makes sense to me. @julianoes ?

Given that we've just been through the process of giving our VTOL vehicles consistent names, perhaps we should also do that for MC? E.g. MAV_TYPE_TYPE_ROTORCOUNT_VARIANT? So would it make sense to name this MAV_TYPE_MC_GENERIC?

I'm not arguing we have to, but it might be useful for readers. So:
- MAV_TYPE_MC_QUAD with variations MAV_TYPE_MC_QUAD_PLUS and MAV_TYPE_MC_QUAD_X
- MAV_TYPE_OCTOROTOR > MAV_TYPE_MC_OCTO or MAV_TYPE_MC_OCTOROTOR
- MAV_TYPE_TRICOPTER > MAV_TYPE_MC_TRICOPTER or MAV_TYPE_MC_TRICOPTER

auturgy: I don't think it's worth it: inserting "MC" doesn't aid understanding or clarity in my view.  I think this is good to go in.

auturgy (approved):

julianoes (approved):

hamishwillee:
I don't see how this can be responsible for the node failures, so merging. Thanks @auturgy, @julianoes @IamPete1

PR #1945, opened 2023-01-23
Extend cellular_status msg and add new cellular_modem_info msg for static fields

asimopunov:
- expand fields and allow multiple instances and add new msg for static cellular_info

## PRs
 * Proto: [Add cellular status and cellular_modem_info #309](https://github.com/mavlink/MAVSDK-Proto/pull/309)
 * MAVSDK: [Add cellular status and cellular_modem_info #1960](https://github.com/mavlink/MAVSDK/pull/1960)
 * MAVSDK-Python: [Add cellular status and cellular_modem_info #546](https://github.com/mavlink/MAVSDK-Python/pull/546)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7612,7 +7612,6 @@
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
       <extensions/>
-      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. NULL terminated if ID string is less than 9 chars.</field>
       <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
       <field type="float" name="band_frequency" units="MHz" invalid="0">LTE radio frequency.</field>
       <field type="uint16_t" name="arfcn" invalid="0">Absolute radio-frequency channel number.</field>
@@ -7625,6 +7624,7 @@
       <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
       <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
       <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. Zero-fill the unused characters at the end of the array if the ID string is less than 9 chars.</field>
     </message>
     <message id="335" name="ISBD_LINK_STATUS">
       <description>Status of the Iridium SBD link.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7621,7 +7621,7 @@
       <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
       <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
       <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
-      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
+      <field type="uint8_t" name="id" instance="true"  invalid="0">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
       <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. This must be NULL terminated if the length is less than 9 human-readable chars, and without the null termination (NULL) byte if the length is exactly 9 chars.</field>
     </message>
     <message id="335" name="ISBD_LINK_STATUS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7614,12 +7614,10 @@
       <extensions/>
       <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
       <field type="float" name="band_frequency" units="MHz" invalid="0">LTE radio frequency.</field>
-      <field type="uint16_t" name="arfcn" invalid="0">Absolute radio-frequency channel number.</field>
-      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
+      <field type="uint16_t" name="channel_number" invalid="0">The channel number (CN). Absolute radio-frequency (ARFCN) / E-UTRA (EARFCN) / UTRA (UARFCN) / New radio (NR_CH).</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0">On 3G is Received Signal Code Power (RSCP). On LTE Reference Signal Received Power (RSRP). On 5G  is New Radio Reference Signal Received Power (NR_RSRQ).</field>
       <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
-      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
-      <field type="float" name="rsrp" units="dBm" invalid="0">Reference Signal Received Power</field>
-      <field type="float" name="rsrq" units="dBm" invalid="0">Reference Signal Received Quality</field>
+      <field type="float" name="rx_quality" units="dBm" invalid="0">On 3G is  Receiver quality (RxQual). On LTE is Reference Signal Received Quality (RSRQ). On 5G is New radio Reference Signal Received Quality (NR_RSRQ).</field>
       <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
       <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
       <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7622,7 +7622,7 @@
       <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
       <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
       <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
-      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. Zero-fill the unused characters at the end of the array if the ID string is less than 9 chars.</field>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. This must be NULL terminated if the length is less than 9 human-readable chars, and without the null termination (NULL) byte if the length is exactly 9 chars.</field>
     </message>
     <message id="335" name="ISBD_LINK_STATUS">
       <description>Status of the Iridium SBD link.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7611,6 +7611,20 @@
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. NULL terminated if ID string is less than 9 chars.</field>
+      <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
+      <field type="float" name="band_frequency" units="MHz" invalid="0">LTE radio frequency.</field>
+      <field type="uint16_t" name="arfcn" invalid="0">Absolute radio-frequency channel number.</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
+      <field type="float" name="rsrp" units="dBm" invalid="0">Reference Signal Received Power</field>
+      <field type="float" name="rsrq" units="dBm" invalid="0">Reference Signal Received Quality</field>
+      <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
+      <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
+      <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
     </message>
     <message id="335" name="ISBD_LINK_STATUS">
       <description>Status of the Iridium SBD link.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7621,7 +7621,7 @@
       <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
       <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
       <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
-      <field type="uint8_t" name="id" instance="true"  invalid="0">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
+      <field type="uint8_t" name="id" instance="true" invalid="0">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
       <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. This must be NULL terminated if the length is less than 9 human-readable chars, and without the null termination (NULL) byte if the length is exactly 9 chars.</field>
     </message>
     <message id="335" name="ISBD_LINK_STATUS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7607,7 +7607,7 @@
       <field type="uint8_t" name="status" enum="CELLULAR_STATUS_FLAG">Cellular modem status</field>
       <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in in CELLULAR_STATUS_FLAG_FAILED</field>
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
-      <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. If unknown, set to UINT8_MAX</field>
+      <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. May be used for Received Signal Strength Indicator (RSSI). If unknown, set to UINT8_MAX</field>
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
@@ -7617,7 +7617,7 @@
       <field type="uint16_t" name="channel_number" invalid="0">The channel number (CN). Absolute radio-frequency (ARFCN) / E-UTRA (EARFCN) / UTRA (UARFCN) / New radio (NR_CH).</field>
       <field type="float" name="rx_level" units="dBm" invalid="0">On 3G is Received Signal Code Power (RSCP). On LTE Reference Signal Received Power (RSRP). On 5G  is New Radio Reference Signal Received Power (NR_RSRQ).</field>
       <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
-      <field type="float" name="rx_quality" units="dBm" invalid="0">On 3G is  Receiver quality (RxQual). On LTE is Reference Signal Received Quality (RSRQ). On 5G is New radio Reference Signal Received Quality (NR_RSRQ).</field>
+      <field type="float" name="rx_quality" units="dBm" invalid="0">On 3G is Receiver Quality (RxQual). On LTE is Reference Signal Received Quality (RSRQ). On 5G is New radio Reference Signal Received Quality (NR_RSRQ).</field>
       <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
       <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
       <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -415,10 +415,10 @@
       <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
       <field type="uint64_t" name="imei" invalid="NULL">Unique Modem International Mobile Equipment Identity Number.</field>
       <field type="uint64_t" name="imsi" invalid="NULL">Current SIM International mobile subscriber identity.</field>
-      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Zero-fill unused characters at the end of the array if the string is less than 10 chars.</field>
-      <field type="char[20]" name="iccid" invalid="[0]">Integrated Circuit Card Identification Number of SIM Card.</field>
-      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Zero-fill unused characters at the end of the array if the string is less than 24 chars. The format is not intended for display.</field>
-      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Zero-fill unused characters at the end of the array if the string is less than 50 chars.</field>
+      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. This must be NULL terminated if the length is less than 10 human-readable chars, and without the null termination (NULL) byte if the length is exactly 10 chars.</field>
+      <field type="char[20]" name="iccid" invalid="[0]">Integrated Circuit Card Identification Number of SIM Card.  This must be NULL terminated if the length is less than 20 human-readable chars, and without the null termination (NULL) byte if the length is exactly 20 chars.</field>
+      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. This must be NULL terminated if the length is less than 24 human-readable chars, and without the null termination (NULL) byte if the length is exactly 24 chars. The format is not intended for display.</field>
+      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name.  This must be NULL terminated if the length is less than 50 human-readable chars, and without the null termination (NULL) byte if the length is exactly 50 chars.</field>
     </message>
     <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
       <wip/>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -416,9 +416,9 @@
       <field type="uint64_t" name="imei" invalid="NULL">Unique Modem International Mobile Equipment Identity Number.</field>
       <field type="uint64_t" name="iccid" invalid="NULL">Integrated Circuit Card Identification Number of SIM Card.</field>
       <field type="uint64_t" name="imsi" invalid="NULL">Current SIM International mobile subscriber identity.</field>
-      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Field is NULL terminated if string is less than 10 chars.</field>
-      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Field is NULL terminated if string is less than 24 chars. The format is not intended for display.</field>
-      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Field is NULL terminated if string is less than 50 chars.</field>
+      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Zero-fill unused characters at the end of the array if the string is less than 10 chars.</field>
+      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Zero-fill unused characters at the end of the array if the string is less than 24 chars. The format is not intended for display.</field>
+      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Zero-fill unused characters at the end of the array if the string is less than 50 chars.</field>
     </message>
     <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
       <wip/>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -410,6 +410,16 @@
       <field type="float" name="vertical_speed_limit" default="NaN" units="m/s">Limit for vertical movement in MAV_FRAME_LOCAL_NED. NaN: No limit applied</field>
       <field type="float" name="yaw_rate_limit" default="NaN" units="rad/s">Limit for vehicle turn rate around its yaw axis. NaN: No limit applied</field>
     </message>
+    <message id="337" name="CELLULAR_MODEM_INFORMATION">
+      <description>Cellular modem device information. This contains static information about a modem and should be sent at low rate. Note that the dynamic information about the connection is provided in an associated CELLULAR_STATUS message.</description>
+      <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
+      <field type="uint64_t" name="imei" invalid="NULL">Unique Modem International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="iccid" invalid="NULL">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="uint64_t" name="imsi" invalid="NULL">Current SIM International mobile subscriber identity.</field>
+      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Field is NULL terminated if string is less than 10 chars.</field>
+      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Field is NULL terminated if string is less than 24 chars. The format is not intended for display.</field>
+      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Field is NULL terminated if string is less than 50 chars.</field>
+    </message>
     <message id="361" name="FIGURE_EIGHT_EXECUTION_STATUS">
       <wip/>
       <!-- This message is work-in-progress it can therefore change, and should NOT be used in stable production environments -->
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -414,9 +414,9 @@
       <description>Cellular modem device information. This contains static information about a modem and should be sent at low rate. Note that the dynamic information about the connection is provided in an associated CELLULAR_STATUS message.</description>
       <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
       <field type="uint64_t" name="imei" invalid="NULL">Unique Modem International Mobile Equipment Identity Number.</field>
-      <field type="uint64_t" name="iccid" invalid="NULL">Integrated Circuit Card Identification Number of SIM Card.</field>
       <field type="uint64_t" name="imsi" invalid="NULL">Current SIM International mobile subscriber identity.</field>
       <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Zero-fill unused characters at the end of the array if the string is less than 10 chars.</field>
+      <field type="char[20]" name="iccid" invalid="[0]">Integrated Circuit Card Identification Number of SIM Card.</field>
       <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Zero-fill unused characters at the end of the array if the string is less than 24 chars. The format is not intended for display.</field>
       <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Zero-fill unused characters at the end of the array if the string is less than 50 chars.</field>
     </message>


potaito:
@julianoes thanks for the link. Luckily the case here is a bit more straightforward I hope (it's probably not in fact :joy: ), because we want to extend the information that can be conveyed regarding LTE modems. I write straightforward because I would assume, that an LTE modem is a better defined concept than a "radio".

potaito:
Let me comment on the new proposed fields in `CELLULAR_STATUS`:
- `id` it's becoming a trend to include more than one LTE modem on a drone. In the simplest case it would be the LTE modem included in the flight controller, in Auterion's case the Skynode, and a separate external modem.  In extreme cases there are manufacturers such as Elsight, who have 4-1 LTE modem devices for link aggregation. We need to be able to distinguish these modem instances.
- `download_rate`: This is the rate of a specific LTE modem. There is another mavlink message called [ONBOARD_COMPUTER_STATUS](https://mavlink.io/en/messages/common.html#ONBOARD_COMPUTER_STATUS) that has fields for the link tx and rx rate. In case of a router with a single LTE modem these fields would be identical. However, looking at routers that perform link aggregation over multiple LTE modems, the fields are not redundant anymore. The proposed `download_rate` would refer to a specific modem instance's rate, whereas `link_tx_rate` from ONBOARD_COMPUTER_STATUS would refer to the aggregated rate of all LTE modems connected to the router.
- `upload_rate`: same as above
- `bit rate error`: Nice to have, don't have a strong opinion here
- `rx_level`: [dBm] absolute RX transmission power
- `tx_level`: [dBm] absolute TX transmission power
- `signal_to_noise`: Signal to noise ratio is useful when analysing interferences.
- `band_number`:  LTE frequency band number: https://en.wikipedia.org/wiki/LTE_frequency_bands. Specifies the frequencies for the up and the downlink, among other attributes.
- `band_frequency`: LTE frequency [MHz]. Assuming the modem uses the correct frequency for the selected band number, `band_frequency` and `band_number` are _somewhat redundant_. Though from what I have seen, the actual frequency can deviate from the selected band's frequency by a few MHz. On top of that, there are bands with overlapping frequencies. It's thus not possible to infer the band number from a frequency alone, and I think both fields are relevant. 
- `arfcn`: "Absolute radio-frequency channel number absolute radio-frequency channel number (ARFCN) is a code that specifies a pair of physical radio carriers used for transmission and reception" from [Wikipedia](https://en.wikipedia.org/wiki/Absolute_radio-frequency_channel_number). **ARFCN** is from the GSM days, and for 3G/4G this was replaced by **EARFCN**, in UMTS with **UARFCN**. I'm not sure this field is really necessary, if we already send the band number and the band frequency.

Note that some of these are LTE specific, while the CELLULAR_STATUS message is not

message_definitions/v1.0/common.xml
```diff
@@ -7261,6 +7273,16 @@
       <field type="uint8_t" name="roaming">Enable/disable roaming. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
       <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
+    <message id="337" name="NIC_INFO">
+      <description>This message contains NIC device specific information.</description>
+      <field type="uint8_t" name="id" instance="true">NIC instance number.</field>
+      <field type="uint8_t" name="nic_id" invalid="0">Unique id for NIC (Vlan).</field>
```
MaEtUgR, line 7279:
```suggestion
      <field type="uint8_t" name="nic_id" invalid="0">Unique id for NIC (VLAN).</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7261,6 +7273,16 @@
...
+      <field type="char[50]" name="nic_model" invalid="[0]">NIC model.</field>
+      <field type="uint64_t" name="imei" invalid="0">Unique NIC International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="iccid" invalid="0">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="uint64_t" name="imsi" invalid="0"> Current SIM International mobile subscriber identity.</field>
```
MaEtUgR, line 7283:
```suggestion
      <field type="uint64_t" name="imsi" invalid="0">Current SIM International mobile subscriber identity.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number</field>
+      <field type="uint64_t" name="download_rate" invalid="0">download rate in kbits/s</field>
+      <field type="uint64_t" name="upload_rate" invalid="0">upload rate in kbits/s</field>
+      <field type="uint64_t" name="ber" invalid="0">bit rate error measurement</field>
```
MaEtUgR, line 7243:
```suggestion
      <field type="uint64_t" name="ber" invalid="0">bit error rate measurement</field>
```
No?

Also I'd leave away the point at the end of the descriptions but for sure keep it consistent.

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="float" name="rx_level" invalid="0">rx level.</field>
```
MaEtUgR, line 7244:
```suggestion
      <field type="float" name="rx_level" invalid="0">Reception signal strength.</field>
```
?

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="float" name="tx_level" invalid="0">rx level.</field>
```
MaEtUgR, line 7245:
```suggestion
      <field type="float" name="tx_level" invalid="0">Transmit signal strength.</field>
```
? Or transmit power? Repeating the filed name in the description generally doesn't help.

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="float" name="signal_to_noise" invalid="0">signal to noise.</field>
```
MaEtUgR, line 7246:
```suggestion
      <field type="float" name="signal_to_noise" invalid="0">Signal to noise ratio in range 0-1 of the received signal.</field>
```
?

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="char[9]" name="cell_tower_id" invalid="0">signal to noise.</field>
```
MaEtUgR, line 7247:
```suggestion
      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="uint8_t" name="band_number" invalid="0">band number.</field>
```
MaEtUgR, line 7248:
```suggestion
      <field type="uint8_t" name="band_number" invalid="0">Number describing the frequency ban currently in use.</field>
```
?

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="float" name="band_frequency" invalid="0">band number. </field>
```
MaEtUgR, line 7249:
```suggestion
      <field type="float" name="band_frequency" invalid="0">Radio frequency currently in use.</field>
```
?

MaEtUgR (commented):
I read through without much context out of curiosity and realized I wouldn't understand what these fields contain from their description and would need to consult the code assigning to them. So I made some example suggestions on what could be more helpful for the developer reading this but don't take them as the right description because I'm missing the detailed context.

message_definitions/v1.0/common.xml
akkawimo: that's not the range I am getting from the modem. datasheet doesn't have units but my guess is db

message_definitions/v1.0/common.xml
akkawimo: that's correct, thanks

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number</field>
+      <field type="uint64_t" name="download_rate" invalid="0">download rate in kbits/s</field>
+      <field type="uint64_t" name="upload_rate" invalid="0">upload rate in kbits/s</field>
+      <field type="uint64_t" name="ber" invalid="0">bit error rate measurement</field>
+      <field type="float" name="rx_level" invalid="0">rx level.</field>
+      <field type="float" name="tx_level" invalid="0">rx level.</field>
```
hamishwillee, line 7245: What are the units and allowed values? What is the required resolution?

message_definitions/v1.0/common.xml
hamishwillee: You need to add units. If there is an allowed range, that should be specified too.

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
      <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
```

hamishwillee: @MaEtUgR Completely agree these need better descriptions. They don't have to be as "complete" as you have put, but they do have to be inferable. I think LTE frequency band number. is good because it is a searchable term and unambigous.

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
      <field type="float" name="band_frequency" invalid="0">LTE radio frequency.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
...
+      <field type="float" name="signal_to_noise" invalid="0">signal to noise.</field>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower.</field>
+      <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
+      <field type="float" name="band_frequency" invalid="0">LTE radio frequency.</field>
```
hamishwillee, line 7249: Units?

message_definitions/v1.0/common.xml
hamishwillee: What are the units?

message_definitions/v1.0/common.xml
hamishwillee: What are the units

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number</field>
```
hamishwillee, line 7240: This makes sense to me.

hamishwillee:
```suggestion
      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7236,6 +7236,18 @@
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" invalid="0">download rate in kbits/s</field>
```
hamishwillee, line 7241: Units can be specified in `units` attribute. Not sure kbit/s allowed - the ones you can use now are https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd#L81 - more can be added with justification.

hamishwillee: Make all the descriptions in this capitalized first letter senences o .e.g. Download, not download ...

message_definitions/v1.0/common.xml
```diff
@@ -7261,6 +7273,16 @@
...
     </message>
+    <message id="337" name="NIC_INFO">
+      <description>This message contains NIC device specific information.</description>
+      <field type="uint8_t" name="id" instance="true">NIC instance number.</field>
+      <field type="uint8_t" name="nic_id" invalid="0">Unique id for NIC (VLAN).</field>
+      <field type="char[50]" name="nic_model" invalid="[0]">NIC model.</field>
+      <field type="uint64_t" name="imei" invalid="0">Unique NIC International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="iccid" invalid="0">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="uint64_t" name="imsi" invalid="0">Current SIM International mobile subscriber identity.</field>
+      <field type="float" name="firmware_version" invalid="NAN">The firmware version installed on the modem.</field>
```
hamishwillee, line 7284: YOu might get this from https://mavlink.io/en/messages/development.html#COMPONENT_INFORMATION_BASIC

What is its format, what is it going to be used for? We have same question for `COMPONENT_INFORMATION_BASIC` - it would be more useful in some ways to force semver or to have a separate field to indicate the format.

message_definitions/v1.0/common.xml
```diff
@@ -7261,6 +7273,16 @@
...
       <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
+    <message id="337" name="NIC_INFO">
+      <description>This message contains NIC device specific information.</description>
+      <field type="uint8_t" name="id" instance="true">NIC instance number.</field>
+      <field type="uint8_t" name="nic_id" invalid="0">Unique id for NIC (VLAN).</field>
+      <field type="char[50]" name="nic_model" invalid="[0]">NIC model.</field>
+      <field type="uint64_t" name="imei" invalid="0">Unique NIC International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="iccid" invalid="0">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="uint64_t" name="imsi" invalid="0">Current SIM International mobile subscriber identity.</field>
```
hamishwillee, line 7283: Why does a nic have a sim card?

message_definitions/v1.0/common.xml
```diff
@@ -7261,6 +7273,16 @@
       <field type="uint8_t" name="roaming">Enable/disable roaming. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
       <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
+    <message id="337" name="NIC_INFO">
+      <description>This message contains NIC device specific information.</description>
```
hamishwillee, line 7277: Why do you need this? What will you do with the data?

hamishwillee:
@akkawimo  For the specific messages we need units and descriptions that are unambiguous - @potaito table above https://github.com/mavlink/mavlink/pull/1945#issuecomment-1410373500 is pretty clear - I updated a few, but you will need to think about the others.

I'll raise with the mavlink call tonight.

message_definitions/v1.0/common.xml
akkawimo: I'm going to go by this:
https://documentation.meraki.com/MR/Wi-Fi_Basics_and_Best_Practices/Signal-to-Noise_Ratio_(SNR)_and_Wireless_Signal_Strength?dtid=osscdc000283

message_definitions/v1.0/common.xml
```diff
@@ -7230,12 +7230,24 @@
...
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">download rate in kbits/s</field>
+      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">upload rate in kbits/s</field>
+      <field type="uint64_t" name="ber" units="kbits/s" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received.</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0"> The Rx level is the power of the received signal</field>
+      <field type="float" name="tx_level" units="dBm" invalid="0">The Tx level is the power at which a modem transmits its signal</field>
+      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower.</field>
```
hamishwillee, line 7247:
```suggestion
      <field type="char[9]" name="cell_tower_id" invalid="0">IDs of the currently connected cell towers.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7230,12 +7230,24 @@
...
       <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. If unknown, set to UINT8_MAX</field>
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">download rate in kbits/s</field>
+      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">upload rate in kbits/s</field>
+      <field type="uint64_t" name="ber" units="kbits/s" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received.</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0"> The Rx level is the power of the received signal</field>
```
hamishwillee, line 7244:
```suggestion
      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7230,12 +7230,24 @@
...
+      <field type="float" name="tx_level" units="dBm" invalid="0">The Tx level is the power at which a modem transmits its signal</field>
```
hamishwillee, line 7245:
```suggestion
      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7230,12 +7230,24 @@
...
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">download rate in kbits/s</field>
+      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">upload rate in kbits/s</field>
+      <field type="uint64_t" name="ber" units="kbits/s" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received.</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
+      <field type="char[9]" name="cell_tower_id" invalid="[0]">IDs of the currently connected cell towers.</field>
```
hamishwillee, line 7247: Why a char?

message_definitions/v1.0/common.xml
```diff
@@ -7230,12 +7230,24 @@
...
+      <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in CELLULAR_STATUS_FLAG_FAILED</field>
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
       <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. If unknown, set to UINT8_MAX</field>
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">download rate in kbits/s</field>
+      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">upload rate in kbits/s</field>
```
hamishwillee, line 7242:
```suggestion
      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">Download rate.</field>
      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">Upload rate.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7261,6 +7273,16 @@
       <field type="uint8_t" name="roaming">Enable/disable roaming. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
       <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
+    <message id="337" name="NIC_INFO">
+      <description>This message contains NIC device specific information.</description>
+      <field type="uint8_t" name="id" instance="true">NIC instance number.</field>
+      <field type="uint8_t" name="nic_id" invalid="0">Unique id for NIC (VLAN).</field>
+      <field type="char[50]" name="nic_model" invalid="[0]">NIC model name.</field>
```
hamishwillee, line 7280:
```suggestion
      <field type="char[50]" name="nic_model" invalid="[0]">NIC model name. Field is NULL terminated if string is less than 50 chars.</field>
```

message_definitions/v1.0/common.xml
akkawimo: I only get one string ex. A60B08
not sure if that means more than one tower

message_definitions/v1.0/common.xml
akkawimo: maybe nic should be changed to modem

message_definitions/v1.0/common.xml
```diff
@@ -7230,12 +7230,24 @@
...
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">Download rate.</field>
+      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">Upload rate.</field>
+      <field type="uint64_t" name="ber" units="kbits/s" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received.</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
+      <field type="char[9]" name="cell_tower_id" invalid="[0]">IDs of the currently connected cell towers.</field>
```
hamishwillee, line 7247:
```suggestion
      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. NULL terminated if ID string is less than 9 chars.</field>
```

message_definitions/v1.0/common.xml
akkawimo: This msg could be triggered with a mav_cmd just once to store the data. Might be required for tracking and correlating of lte 'signal quailty', uav pose and hardware onboard.

message_definitions/v1.0/common.xml
akkawimo: I went with this:
https://www.cisco.com/c/en/us/support/docs/dial-access/asynchronous-connections/15380-trans-rec-15380.html

message_definitions/v1.0/common.xml
akkawimo: I went with this:
https://www.cisco.com/c/en/us/support/docs/dial-access/asynchronous-connections/15380-trans-rec-15380.html

message_definitions/v1.0/common.xml
akkawimo: added kbit/s, will make a pr on pymavlink

message_definitions/v1.0/common.xml
akkawimo: added

akkawimo:
I guess we can have one field for ARFCN, UARFCN and EARFCN?

message_definitions/v1.0/common.xml
akkawimo: or just cellular_info

message_definitions/v1.0/common.xml
akkawimo: at the moment, this is the example I have from the api docs "0." but I didn't get anything in return from calling the real hardware.  
Anyway, I think we can just put it in component_information_basic. Not sure though if it matters that it's development.xml regarding compatibility. with mavlink-router, I guess it just lets it pass if it's on common.xml?
And px4 would have to include COMPONENT_INFORMATION_BASIC?

akkawimo:
@hamishwillee 
I also missed RSRP, and RSRQ since they weren't in the docs but can see from what the api returns. Those could be added along RSSI
https://5gstore.com/blog/2021/04/08/understanding-rssi-rsrp-and-rsrq/

message_definitions/v1.0/common.xml
potaito: This second message is supposed to contain the static information of the modem and the SIM card. This would be transmitted at a very low rate or per request only as @akkawimo wrote.

message_definitions/v1.0/common.xml
```diff
@@ -7231,12 +7231,26 @@
...
-      <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in in CELLULAR_STATUS_FLAG_FAILED</field>
+      <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in CELLULAR_STATUS_FLAG_FAILED</field>
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
       <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. If unknown, set to UINT8_MAX</field>
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">Download rate.</field>
```
potaito, line 7242: The ONBOARD_COMPUTER_STATUS message has two fields called `link_tx_rate` and `link_rx_rate`. Should we use the same names for these fields here?

message_definitions/v1.0/common.xml
```diff
@@ -7262,6 +7276,16 @@
       <field type="uint8_t" name="roaming">Enable/disable roaming. 0: setting unchanged, 1: disabled, 2: enabled. Current setting when sent back as a response.</field>
       <field type="uint8_t" name="response" enum="CELLULAR_CONFIG_RESPONSE">Message acceptance response (sent back to GS).</field>
     </message>
+    <message id="337" name="NIC_INFO">
```
potaito, line 7279:
```suggestion
    <message id="337" name="CELLULAR_MODEM_INFORMATION">
```

message_definitions/v1.0/common.xml
akkawimo: Sure

message_definitions/v1.0/common.xml
```diff
@@ -7262,6 +7276,16 @@
...
     </message>
+    <message id="337" name="CELLULAR_MODEM_INFORMATION">
+      <description>This message contains NIC device specific information.</description>
+      <field type="uint8_t" name="id" instance="true">NIC instance number.</field>
+      <field type="uint8_t" name="nic_id" invalid="0">Unique id for NIC (VLAN).</field>
+      <field type="char[50]" name="nic_model" invalid="[0]">NIC model name. Field is NULL terminated if string is less than 50 chars.</field>
+      <field type="uint64_t" name="imei" invalid="0">Unique NIC International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="iccid" invalid="0">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="uint64_t" name="imsi" invalid="0">Current SIM International mobile subscriber identity.</field>
+      <field type="float" name="firmware_version" invalid="NAN">The firmware version installed on the modem.</field>
```
potaito, line 7287:
```suggestion
```

As per the discussion below, there is already a message in the works to convey the firmware version of a device.

message_definitions/v1.0/common.xml
```diff
@@ -7231,12 +7231,26 @@
...
+      <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in CELLULAR_STATUS_FLAG_FAILED</field>
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
       <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. If unknown, set to UINT8_MAX</field>
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
+      <field type="uint64_t" name="download_rate" units="kbits/s" invalid="0">Download rate.</field>
+      <field type="uint64_t" name="upload_rate" units="kbits/s" invalid="0">Upload rate.</field>
```
potaito, line 7243:
```suggestion
      <field type="uint64_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
      <field type="uint64_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
```

Re-using field name and unit from the ONBOARD_COMPUTER_STATUS message.

message_definitions/v1.0/common.xml
akkawimo: Cool, I guess it's fine that it's in development.xml

potaito: Well I hope it gets out of development soon, because I've seen the need for it a couple of times already :sweat_smile:

akkawimo: Ah okay I'm not familiar with the road to common.xml

message_definitions/v1.0/common.xml
```diff
@@ -7231,12 +7231,26 @@
...
+      <field type="uint64_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
+      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. NULL terminated if ID string is less than 9 chars.</field>
+      <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
+      <field type="float" name="band_frequency" units="MHz" invalid="0">LTE radio frequency.</field>
+      <field type="float" name="arfcn" invalid="0">Absolute radio-frequency channel number.</field>
```
hamishwillee, line 7251: Integer?

message_definitions/v1.0/common.xml
```diff
@@ -7231,12 +7231,26 @@
...
       <field type="uint8_t" name="status" enum="CELLULAR_STATUS_FLAG">Cellular modem status</field>
-      <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in in CELLULAR_STATUS_FLAG_FAILED</field>
+      <field type="uint8_t" name="failure_reason" enum="CELLULAR_NETWORK_FAILED_REASON">Failure reason when status in CELLULAR_STATUS_FLAG_FAILED</field>
       <field type="uint8_t" name="type" enum="CELLULAR_NETWORK_RADIO_TYPE">Cellular network radio type: gsm, cdma, lte...</field>
       <field type="uint8_t" name="quality" invalid="UINT8_MAX">Signal quality in percent. If unknown, set to UINT8_MAX</field>
       <field type="uint16_t" name="mcc" invalid="UINT16_MAX">Mobile country code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="mnc" invalid="UINT16_MAX">Mobile network code. If unknown, set to UINT16_MAX</field>
       <field type="uint16_t" name="lac" invalid="0">Location area code. If unknown, set to 0</field>
+      <extensions/>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 0.</field>
```
hamishwillee, line 7241: Make relationship between instance numbers and messages clear - descriptive names, descriptions at message level too.

message_definitions/v1.0/common.xml
```diff
@@ -7231,12 +7231,26 @@
...
+      <field type="uint64_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
```
hamishwillee, line 7242: Is this size appropriate?

Which are most likely to be populated most?

message_definitions/v1.0/common.xml
potaito: Yes, `arfcn` is Integer. I would have to check the range though.

hamishwillee:
Firstly, thank you @akkawimo for coming along to the mav call last night. 2.30am ish is over-and-above. My summary of the discussion is:

  - Order items to allow better [zero byte truncation](https://mavlink.io/en/guide/serialization.html#payload_truncation)
    - _Extension_ fields are put at the end of the message, ordered by the declaration order in XML. So fields that are more likely to be zero should should be at the end of the the extensions, to increase the chance they will be truncated.
    - For normal fields (modem message), big data types are ordered before small ones but otherwise the order used is the declaration order - so you "try" to declare small data types that are likely to be zero last. 
      - I don't think this is worth worrying about much unless you have arrays of bytes - then those definitely should be declared last.
  - Select appropriate data types for the size of data - _check all the fields_. Smallest practical size for your data, but consider whether the technology might reasonably change in the next 5 years to allow rates (say) that would overflow.
  - Make it clear that the nic msg is the low rate info associated with cellular status. Use similar names for them. Make sure the descriptions cross reference each message. Make sure that it is clear that the index in both references the same item.
  
We can't take extensions untested which means this PR can be approved, but we can't merge it until we're sure it is "good". IN practice that means a working prototype for example in MAVSDK. You'll have to create a branch for testing.

message_definitions/v1.0/common.xml
hamishwillee: Yes, we need to check size and order of all these fields as per https://github.com/mavlink/mavlink/pull/1945#issuecomment-1412784579

Consider reasonable future proofing  - so if rates could likely double in 5 years, we need to make sure the field is big enough.

message_definitions/v1.0/common.xml
hamishwillee: @potaito We discussed in dev call.
- We think this _could_ stay in this message as a convenience to users. Might argue otherwise if the component info was common.
- Needs to be null terminated strings so it can zero truncate.
- If you can know the format of the field this should be stated. If not, it might be worth stating whether the field is intended for display or what? This has been the problem in the component info - generally the assumption is that fields are for display, which makes it hard for them to be used reliably for other purposes.

message_definitions/v1.0/common.xml
hamishwillee: FWIW Good. The important thing here is to draw a clear connection between the messages - the modem information is the low rate static cellular information. Upshot, you should cross reference them in the descriptions and also make it clear that the indexes must match.

message_definitions/v1.0/common.xml
hamishwillee, line 7241: Thinking about this, I suggest modifying to indexed from 1; 0 means not supplied. That way the recipient can know that if it gets 0 the sender has not been updated. This comes at the cost of one byte per message.

@auturgy You OK with this?

```suggestion
      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
```

If accepted also needs to be updated in the other message.

akkawimo: makes sense to me

message_definitions/v1.0/common.xml
akkawimo: nevermind, KiB/s already there

message_definitions/v1.0/common.xml
akkawimo: will keep firmware field as discussed in mav call

message_definitions/v1.0/development.xml
```diff
@@ -380,6 +380,16 @@
       <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.</field>
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
     </message>
+    <message id="337" name="CELLULAR_MODEM_INFORMATION">
+      <description>This message contains modem device specific information.</description>
+      <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message. This message provides static information of the modem. </field>
```
hamishwillee, line 385: Perhaps?

```suggestion
      <description>Cellular modem device information. This contains static information about a modem and should be sent at low rate. Note that the dynamic information about the connection is provided in an associated CELLULAR_STATUS message.</description>
      <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
```

akkawimo:
@hamishwillee thanks for all the tips. Did I miss anything?

message_definitions/v1.0/common.xml
```diff
@@ -7237,6 +7237,20 @@
...
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
+      <field type="float" name="rsrp" units="dBm" invalid="0">Reference Signal Received Power</field>
+      <field type="float" name="rsrq" units="dBm" invalid="0">Reference Signal Received Quality</field>
+      <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
+      <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
+      <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
```
hamishwillee, line 7253: The char[9] should be at the end. In addition, there was some confusion about the wording NULL terminated in https://github.com/mavlink/mavlink/issues/1946

Essentially that just means that the string will end at the first array value = 0 - but actually we want the unused values to be zero-filled. So updating here:

```suggestion
      <field type="uint8_t" name="band_number" invalid="0">LTE frequency band number.</field>
      <field type="float" name="band_frequency" units="MHz" invalid="0">LTE radio frequency.</field>
      <field type="uint16_t" name="arfcn" invalid="0">Absolute radio-frequency channel number.</field>
      <field type="float" name="rx_level" units="dBm" invalid="0">Received signal absolute power level.</field>
      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
      <field type="float" name="signal_to_noise" units="dBm" invalid="0">The Signal-to-Noise Ratio (SNR) is the difference between the received wireless signal and the noise floor i.e. erroneous background transmissions</field>
      <field type="float" name="rsrp" units="dBm" invalid="0">Reference Signal Received Power</field>
      <field type="float" name="rsrq" units="dBm" invalid="0">Reference Signal Received Quality</field>
      <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
      <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
      <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. Zero-fill the unused characters at the end of the array if the ID string is less than 9 chars.</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -380,6 +380,16 @@
...
+      <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
+      <field type="uint64_t" name="imei" invalid="NULL">Unique Modem International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="iccid" invalid="NULL">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="uint64_t" name="imsi" invalid="NULL">Current SIM International mobile subscriber identity.</field>
+      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Field is NULL terminated if string is less than 10 chars.</field>
+      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Field is NULL terminated if string is less than 24 chars. The format is not intended for display.</field>
+      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Field is NULL terminated if string is less than 50 chars.</field>
```
hamishwillee, line 391: Make it clear we mean zero-fill and not "null terminate".

```suggestion
      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Zero-fill unused characters at the end of the array if the string is less than 10 chars.</field>
      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Zero-fill unused characters at the end of the array if the string is less than 24 chars. The format is not intended for display.</field>
      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Zero-fill unused characters at the end of the array if the string is less than 50 chars.</field>
```

hamishwillee:
@akkawimo It's looking pretty good modulo the suggestions I just made. @julianoes @auturgy Any comment.

Note that as discussed in last call, we can't merge until there is an implementation. But once everyone thinks it makes sense we can give "conditional approval".

akkawimo:
@hamishwillee I spoke with the OEM that I am working with and got the following info for how these fields are used within their api:

rx_level - is RSCP (on 3G) / RSRP (on LTE) / NR_RSRP (on 5G)
signal_to_noise - is RxQual (on 3G) / RSRQ (on LTE) / NR_RSRQ (on 5G)
arfcn - is the channel number (ARFCN / EARFCN / UARFCN / NR_CH)

Updated to remove redundancies.

akkawimo:
fyi reference:
https://www.rangeful.com/what-is-rssi-sinr-rsrp-rsrq-how-does-this-affect-signal-quality/

message_definitions/v1.0/development.xml
```diff
@@ -380,6 +380,16 @@
...
+      <field type="uint8_t" name="id" instance="true">Modem instance number. Indexed from 1. Matches index in corresponding CELLULAR_STATUS message.</field>
+      <field type="uint64_t" name="imei" invalid="NULL">Unique Modem International Mobile Equipment Identity Number.</field>
+      <field type="uint64_t" name="imsi" invalid="NULL">Current SIM International mobile subscriber identity.</field>
+      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. Zero-fill unused characters at the end of the array if the string is less than 10 chars.</field>
+      <field type="char[20]" name="iccid" invalid="[0]">Integrated Circuit Card Identification Number of SIM Card.</field>
+      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. Zero-fill unused characters at the end of the array if the string is less than 24 chars. The format is not intended for display.</field>
+      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name. Zero-fill unused characters at the end of the array if the string is less than 50 chars.</field>
```
hamishwillee, line 391: Reverting my original change to zero-fill following feedback

```suggestion
      <field type="char[10]" name="modem_id" invalid="[0]">Unique id for modem. This must be NULL terminated if the length is less than 10 human-readable chars, and without the null termination (NULL) byte if the length is exactly 10 chars.</field>
      <field type="char[20]" name="iccid" invalid="[0]">Integrated Circuit Card Identification Number of SIM Card.  This must be NULL terminated if the length is less than 20 human-readable chars, and without the null termination (NULL) byte if the length is exactly 20 chars.</field>
      <field type="char[24]" name="firmware" invalid="[0]">The firmware version installed on the modem. This must be NULL terminated if the length is less than 24 human-readable chars, and without the null termination (NULL) byte if the length is exactly 24 chars. The format is not intended for display.</field>
      <field type="char[50]" name="modem_model" invalid="[0]">Modem model name.  This must be NULL terminated if the length is less than 50 human-readable chars, and without the null termination (NULL) byte if the length is exactly 50 chars.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7233,10 +7233,22 @@
...
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="rx_quality" units="dBm" invalid="0">On 3G is Receiver Quality (RxQual). On LTE is Reference Signal Received Quality (RSRQ). On 5G is New radio Reference Signal Received Quality (NR_RSRQ).</field>
+      <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
+      <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
+      <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
+      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. Zero-fill the unused characters at the end of the array if the ID string is less than 9 chars.</field>
```
hamishwillee, line 7251: Revert 
```suggestion
      <field type="char[9]" name="cell_tower_id" invalid="0">ID of the currently connected cell tower. This must be NULL terminated if the length is less than 9 human-readable chars, and without the null termination (NULL) byte if the length is exactly 9 chars.</field>
```

hamishwillee:
@akkawimo Looks good to me. I reverted the changes for "zero filling"; they affect existing implementations.

@auturgy @julianoes Any review comments to add. I think all suggestions have been addressed.

message_definitions/v1.0/common.xml
```diff
@@ -7233,10 +7233,22 @@
...
+      <field type="float" name="rx_level" units="dBm" invalid="0">On 3G is Received Signal Code Power (RSCP). On LTE Reference Signal Received Power (RSRP). On 5G  is New Radio Reference Signal Received Power (NR_RSRQ).</field>
+      <field type="float" name="tx_level" units="dBm" invalid="0">Transmitter (modem) signal absolute power level.</field>
+      <field type="float" name="rx_quality" units="dBm" invalid="0">On 3G is Receiver Quality (RxQual). On LTE is Reference Signal Received Quality (RSRQ). On 5G is New radio Reference Signal Received Quality (NR_RSRQ).</field>
+      <field type="uint32_t" name="link_tx_rate" units="KiB/s" invalid="0">Download rate.</field>
+      <field type="uint32_t" name="link_rx_rate" units="KiB/s" invalid="0">Upload rate.</field>
+      <field type="uint16_t" name="ber" invalid="0">The bit error rate (BER) is determined by comparing the erroneous bits received with the total number of bits received. It is a ratio.</field>
+      <field type="uint8_t" name="id" instance="true">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
```
auturgy, line 7250: invalid="0"?

hamishwillee:
```suggestion
      <field type="uint8_t" name="id" instance="true"  invalid="0">Cellular instance number. Indexed from 1. Matches index in corresponding CELLULAR_MODEM_INFORMATION message.</field>
```

hamishwillee:
@akkawimo We discussed in the MAV call last night. General feeling is that it is pretty good, but as discussed in the previous meeting can't merge into common.xml until tested in implementation.

So "conditional approval" granted. That's pretty much the same as you'd get for merging into development. It means that we thing the idea is sane and worth testing, and we won't spuriously block it going into common once it is implemented. But we're also still open to more feedback and we would allow changes if needed.

potaito:
> @akkawimo We discussed in the MAV call last night. General feeling is that it is pretty good, but as discussed in the previous meeting can't merge into common.xml until tested in implementation.
> 
> So "conditional approval" granted. That's pretty much the same as you'd get for merging into development. It means that we thing the idea is sane and worth testing, and we won't spuriously block it going into common once it is implemented. But we're also still open to more feedback and we would allow changes if needed.

Just as an update: We are currently working on a proof of concept, including MAVSDK and real hardware.

akkawimo:
@hamishwillee @potaito what do you think about squeezing in a failure bitmask field for CELLULAR_MODEM_INFORMATION? right now, cellular status has a failure field that's uint8. I'm trying to track more than 8 reasons. 64?

julianoes:
Should this still go in? I'm just looking at the MAVSDK part of it:
https://github.com/mavlink/MAVSDK/pull/1960

hamishwillee:
@julianoes The last comments of interest were that the mav call were generally supportive but we need proof of concept, and @potaito responded that this was being worked on in https://github.com/mavlink/mavlink/pull/1945#issuecomment-1446220518

From my point of view, we're at the point where @asimopunov and @potaito need to confirm their interest/implementation. Otherwise, no, we probably should just close this. Let's see how they respond.

sfuhrer:
Why was this closed? Is there no more interest of getting it in?

hamishwillee:
The author closed it. From a MAVLink point of view it has no stakeholder, so there is no prototype or proof of fitness.

PR #1947, opened 2023-02-03
remove invalid enum reference from storm32.xml

aler9:
#1941 contains an invalid enum reference that breaks generation of external libraries.

I searched for the enum definition but i couldn't find it, therefore it should be removed unless @olliw42 provides it.

### DIFF
--- a/message_definitions/v1.0/storm32.xml
+++ b/message_definitions/v1.0/storm32.xml
@@ -10,7 +10,7 @@ Range of IDs:
 Documentation:
   STORM32 and QSHOT additions
   with mLRS additions merged
-  16. Jan. 2023
+  3. Feb. 2023
   All messages are technically WIP, but some are quite stable now.
   Quite stable means that it is in practical use, but may see extension.
   A more detailed description of the concept underlying the STORM32 and QSHOT messages can be found here:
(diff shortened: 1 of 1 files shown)


olliw42:
omg, indeed ... MANY thx!

could you please also change the date to 3. Feb. 2023 (it seems one can't do a suggestion for non modified lines)

olliw42 (changes requested):

aler9:
@olliw42  all done.

olliw42:
perfect, many thx sir :)

it's interesting that some generators do test for such things...

aler9:
@olliw42 in some strongly typed languages, an enum corresponds to a dedicated type. If the enum doesn't exist, the field has an invalid type.

While in C, that is still a strongly typed language, an enum-based field may be converted into

```c
#define ENUM_MYENUM_VAL1 1
#define ENUM_MYENUM_VAL2 2

typedef struct {
  uint16_t my_field;
} mymsg;
```

In Typescript, it is converted into

```typescript
enum myenum {
  myval1 = 1,
  myval2 = 2,
}

type mymsg = {
  my_field: myenum
};
```

In Golang, it is converted into


```go
type myenum int

const (
  myval1 myenum = 1
  myval2 myenum = 2
)

type mymsg struct {
  my_field myenum
}
```

Therefore myenum must exist.

olliw42:
ah, it's not a generator thing but a language thing
thx for the insight :)

hamishwillee (approved):
@olliw42 I'm taking your comment as approval and merging :-)

PR #1948, opened 2023-02-07
development: add command to configure SiK radio

julianoes:
This is a command that allows to set AT parameters of SiK radios connected to a drone.

For example, this can be used to set the ID of a radio allowing it to be paired to another ground radio. This can be used for configuration at initial setup but also during a mission to specifically swap to a new ground radio.

More details on the possible configuration options can be found in:
https://ardupilot.org/copter/docs/common-3dr-radio-advanced-configuration-and-technical-information.html#using-the-at-command-set
https://docs.px4.io/main/en/data_links/sik_radio.html

For PX4, this would replace the existing ID configuration using a parameter which does not have proper feedback.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -336,6 +336,17 @@
         <param index="6" reserved="true" default="0"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
+      <entry value="550" name="MAV_CMD_SET_AT_S_PARAM" hasLocation="false" isDestination="false">
+        <description>Allows setting an AT S command of an SiK radio.
+        </description>
+        <param index="1" label="Radio instance">The radio instance, one-based, 0 for all.</param>
+        <param index="2" label="Index">The Sx index, e.g. 3 for S3 which is NETID.</param>
+        <param index="3" label="Value">The value to set it to, e.g. default 25 for NETID</param>
+        <param index="4" reserved="true"/>
+        <param index="5" reserved="true"/>
+        <param index="6" reserved="true"/>
+        <param index="7" reserved="true"/>
+      </entry>
     </enum>
   </enums>
   <messages>


message_definitions/v1.0/development.xml
```diff
@@ -336,6 +336,17 @@
         <param index="6" reserved="true" default="0"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
+      <entry value="550" name="MAV_CMD_SET_AT_PARAM" hasLocation="false" isDestination="false">
```
hamishwillee, line 339: I know what an AT command is, but I'm not sure the name is all that clear. Some suggestions:
- MAV_CMD_SET_MODEM_AT_PARAM
- MAV_CMD_SET_MODEM_PARAM
- MAV_CMD_SET_MODEM_CONFIG
- MAV_CMD_SET_AT15_PARAM

hamishwillee:
@julianoes We can probably tweak the message name and description a little to make it more clear. What I am not sure on is whether this is likely a good "common" message. 

1. Can/should this be more generalized? I.e. I suspect you need this as an easy way to set the network ID, but most modems, not just SiK might need this feature.
   - Should it be tied to AT
   - Should it be tied to Sik?
   - If we can abstract it, what set of parameters would we want to set.
2. If it is tied to AT commands, should we make this a general AT command sender -..e not just the user params?

W.r.t. the use case, I get it. However I'm wary of commands in missions that tie the mission to very specific sets of hardware.

@auturgy Any thoughts on this one?

message_definitions/v1.0/development.xml
julianoes: Maybe `MAV_CMD_SET_AT_S_PARAM`, as that's what it is.

julianoes:
@hamishwillee thanks for the review.

1. It can be generalized but I don't know how general this is, I have only ever seen it with SiK radios.
2. I don't think we can easily generalize it to all AT commands because some (like ATI5) have quite a bit of text feedback which we can't translate via the command ACK.

If this isn't common enough it will have to live in private, however, it means I can't upstream the feature for PX4 and PX4 will be stuck with a param without proper feedback which is odd and buggy.

auturgy:
Re SiK, @tridge is the expert.

hamishwillee:
FYI, asked @tridge for comment here https://discord.com/channels/899843920987562014/902774723925147649/1073002491404365986 when he has time.

tridge:
seems reasonable for current SiK. My plan is to replace SiK with a radio firmware that is fully mavlink capable and exposes normal mavlink parameter listing/setting as a component, but that is a way off

hamishwillee:
Awesome!

PR #1949, opened 2023-02-07
Update submodule and check CI

julianoes:
Let's look at why the node test is failing.

Turns out we need to do a `npm install` to get mocha.


hamishwillee:
Thanks for sorting this out!

PR #1950, opened 2023-02-08
param_id char[] description

### LINKED ISSUE #1946: Convinience function(s?) blindly copy 16bytes, - even when less are provided... resulting in somehow corrupt packets.
mavlink_msg_param_value_pack 
uses:
mav_put_char_array(buf, 8, param_id, 16);
mav_array_memcpy(packet.param_id, param_id, sizeof(char)*16);
- this is causing UB / Undefined Behaviour

The compiler does not like this, and it compiles ... but produced odd packets when the parameter name comes from a struct. - oddly enough this does not happen if the parameter is just inside a const char.

When the incoming array with parameter name is for example 5 bytes ... the rest of the copied data is random.
Actually -- this produces an ok first param packet, but the consecutive packets are no good. confusing MAVProxy.

I have an Arduino application that can reproduce this. - the discussion is here: https://forum.arduino.cc/t/when-a-stuct-char-is-not-a-char-anymore/1083746/29

hamishwillee:
Param id char[16] array must be zero filled, not just "null terminated". Null termination implies that the rest of the array does not matter, but in fact the parser will pick up any non-zero after the first null and use it. This might be considered a parser bug, but given that zero filling allows zero-byte truncation, better to fix the definition.

Fixes #1946

@peterbarker @julianoes OK?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5031,7 +5031,7 @@
       <description>Request to read the onboard parameter with the param_id string id. Onboard parameters are stored as key[const char*] -&gt; value[float]. This allows to send a parameter to any other component (such as the GCS) without the need of previous knowledge of possible parameter names. Thus the same GCS can store different parameters for different autopilots. See also https://mavlink.io/en/services/parameter.html for a full documentation of QGroundControl and IMU code.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="char[16]" name="param_id">Onboard parameter id, terminated by NULL if the length is less than 16 human-readable chars and WITHOUT null termination (NULL) byte if the length is exactly 16 chars - applications have to provide 16+1 bytes storage if the ID is stored as string</field>
+      <field type="char[16]" name="param_id">Onboard parameter id. If the parameter ID is less than 16 human-readable chars, zero-fill the remaining chars in the array (terminate with NULL bytes). Applications have to provide 16+1 bytes storage if the ID is stored as a string.</field>
       <field type="int16_t" name="param_index" invalid="-1">Parameter index. Send -1 to use the param ID field as identifier (else the param id will be ignored)</field>
     </message>
     <message id="21" name="PARAM_REQUEST_LIST">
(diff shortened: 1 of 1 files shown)


hamishwillee:
PS I also used similar wording in https://github.com/mavlink/mavlink/pull/1945 " Zero-fill unused characters at the end of the array if the string is less than Xxxx chars."

OK for me to change this everywhere in common - we always mean zero-fill.

peterbarker:
> PS I also used similar wording in #1945 " Zero-fill unused characters at the end of the array if the string is less than Xxxx chars."
> 
> OK for me to change this everywhere in common - we always mean zero-fill.

Sorry, I don't think I like that wording.  It's a significant change to the traditional, "null-terminate your string if there is room" thing we've had in mavlink for a long time.

The problem is that we're making previous implementations non-compliant *and* forcing those previous implementations to be less efficient (at least in terms of CPU clock cycles).  ArduPilot is one of those implementations - this is our method for copying these strings in: https://github.com/ardupilot/ardupilot/blob/master/libraries/AP_Common/AP_Common.cpp#L77 - you can see it copies one more byte if there's room for the null.  This would get rather more expensive if we needed to zero-out the remainder.

The source error is that the input buffer is too short - AFAICS it's unrelated to the content of the buffer.

hamishwillee:
Thanks @peterbarker - OK, we definitely can't do this then.

I think what you're generally saying is that if we specify a data and/or buffer size in the interface we expect data of that size to be used - right? So even though this particular case the data might be stuffed in a smaller buffer the data size we set must be used? 

What I'm trying to get at is whether we need a general comment like that above, because there are lots of cases where the interface takes null terminated strings. Note, closing this as it is now, to prevent accidental merging.

peterbarker:
> Thanks @peterbarker - OK, we definitely can't do this then.
> 
> I think what you're generally saying is that if we specify a data and/or buffer size in the interface we expect data of that size to be used - right? So even though this particular case the data might be stuffed in a smaller buffer the data size we set must be used?
> 
> What I'm trying to get at is whether we need a general comment like that above, because there are lots of cases where the interface takes null terminated strings. Note, closing this as it is now, to prevent accidental merging.

Yes, that's all correct.

You can see ArduPilot doing exactly this (supplying a right-sized-buffer) here: https://github.com/ardupilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L2692

PR #1951, opened 2023-02-13
minimal: Add MAV_STATE_REBOOT_REQUIRED in MAV_STATE

patrickelectric:
The idea is for the vehicle to inform the GCS that a reboot is required, right now there are palliative methods like parameter metadata, but that does not work when the GCS is restarted or a second computer connects to the vehicle. The vehicle should be able to inform the GCS that a reboot is required.

### DIFF
--- a/message_definitions/v1.0/minimal.xml
+++ b/message_definitions/v1.0/minimal.xml
@@ -283,6 +283,9 @@
       <entry value="8" name="MAV_STATE_FLIGHT_TERMINATION">
         <description>System is terminating itself.</description>
       </entry>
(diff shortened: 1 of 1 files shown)


hamishwillee:
A few questions:
- Do flight stacks themselves "know" what parameters require a reboot? I mean is this information stored onboard?
- Would anything other than a parameter update trigger this?
- I like the concept that an unknown flight stack might be able to signal a GCS/user to reboot that a parameter update requires a reboot. But if they have the metadata to know how to safely set a parameter, don't they also know that the reboot is required?

@auturgy @julianoes Thoughts?

hamishwillee:
MAV_STATE is not a bitmask, so you can't have this and another state.

PR #1956, opened 2023-02-28
Positional NAV command - test wording of scaling recommendation

hamishwillee:
Generally if you're sending lat/lon in a MAV_CMD then using COMMAND_INT is better, as it offers greater precision.
I propose to make this clear in every command where this applies. 
This PR tests the wording in one case before rolling it out more widely.

@peterbarker OK?

Following on from discussion in https://github.com/ArduPilot/MAVProxy/pull/1149

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1087,8 +1087,8 @@
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
         <param index="3" label="Pass Radius" units="m">0 to pass through the WP, if &gt; 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
         <param index="4" label="Yaw" units="deg">Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
-        <param index="5" label="Latitude">Latitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended).</param>
-        <param index="6" label="Longitude">Longitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended).</param>
+        <param index="5" label="Latitude">Latitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended for greater precision).</param>
+        <param index="6" label="Longitude">Longitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended for greater precision).</param>
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="17" name="MAV_CMD_NAV_LOITER_UNLIM" hasLocation="true" isDestination="true">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1087,8 +1087,8 @@
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
         <param index="3" label="Pass Radius" units="m">0 to pass through the WP, if &gt; 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
         <param index="4" label="Yaw" units="deg">Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
-        <param index="5" label="Latitude">Latitude</param>
-        <param index="6" label="Longitude">Longitude</param>
+        <param index="5" label="Latitude">Latitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended).</param>
+        <param index="6" label="Longitude">Longitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended).</param>
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1087,8 +1087,8 @@
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
         <param index="3" label="Pass Radius" units="m">0 to pass through the WP, if &gt; 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.</param>
         <param index="4" label="Yaw" units="deg">Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).</param>
-        <param index="5" label="Latitude">Latitude</param>
-        <param index="6" label="Longitude">Longitude</param>
+        <param index="5" label="Latitude">Latitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended).</param>
+        <param index="6" label="Longitude">Longitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended).</param>
```
hamishwillee, line 1091:
```suggestion
        <param index="5" label="Latitude">Latitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended for greater precision).</param>
        <param index="6" label="Longitude">Longitude. Multiply degree value by 1E7 if sent in COMMAND_INT (recommended for greater precision).</param>
```

julianoes (commented):
I'm confused because a MAV_CMD_NAV_WAYPOINT is generally not sent as a COMMAND_INT but as a MISSION_ITEM_INT. The comment makes sense in general and for some commands like [MAV_CMD_DO_REPOSITION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_REPOSITION) but not for this one, I'd say.

hamishwillee:
Curses, you're right. Wording OK in principle for commands where it makes sense?

We probably should mark up things we don't expect to be supported as commands/mission items.

peterbarker (approved):
Wording looks good.

hamishwillee:
@peterbarker @auturgy Might we plan to get away from all this (in some far future) with yet another command and deprecating the others. Essentially make param6, 7 into a `double` so any current MAV_CMD can just be shoved into the one command? We could even use this in Missions.
Something like:

Field Name | Type | Values | Description
-- | -- | -- | --
target_system | uint8_t |   | System ID
target_component | uint8_t |   | Component ID
frame | uint8_t | MAV_FRAME | The coordinate system of the COMMAND.
opts | uint8_t | COMMAND_OPTS | Test disable
command | uint16_t | MAV_CMD | The scheduled action for the mission item.
current | uint8_t |   | ?
param1 | float |   | PARAM1, see MAV_CMD enum
param2 | float |   | PARAM2, see MAV_CMD enum
param3 | float |   | PARAM3, see MAV_CMD enum
param4 | float |   | PARAM4, see MAV_CMD enum
param6  | double |   | PARAM5/ local: x position in meters, global: latitude in degrees
param7   | double |   | PARAM6 / local: y position in meters, global: longitude in degrees
param8   | float |   | PARAM7 / z position: global: altitude in meters (relative or absolute, depending on frame).


- This includes `frame` - another reason not to use `COMMAND_LONG` for positional commands because you shouldn't be guessing about the intended frame. 
- @peterbarker What frame does ardupilot use for COMMAND_LONG or does this vary by MAV_CMD? If we all use the same frame then we should specify what that is in the message. 
- does anyone use `current` or `confirmation`? They appear to be for sequence counting, but are different in the two existing commands.
- The opts field is a separate discussion. Essentially I was hoping to be able to indicate to a flight stack "this is a test command" - tell me if the params are valid but don't do other side effects". 

I know we keep on going around these kinds of things. That's because we never resolve them in a satisfying way.

julianoes (commented):
That's ok but I would call it accuracy or resolution rather than precision.

auturgy:
Why a double? Int32 should be more than enough

hamishwillee:
> Why a double? Int32 should be more than enough

Because then we wouldn't be able to pack MAV_CMD that have param5,6 as floats.
This is a way to be able to pack all commands into one message, without having to think about scaling. 

I'm aware that this is yet another message, and you still have the complication that MAV_CMDS positions are set as ints in mission_int. I'm interested in it to resolve things like 
- "what is the frame of a command long if someone uses this command",
- "how do I know if a particular command is supported without having the side effects - hence the new opt flag
- Removing the autocontinue field which is an accidental copy from mission_int

So I'm after some kind of cost/benefit analysis. Are their alternatives that achieve the same kinds of things.

peterbarker:
> > Why a double? Int32 should be more than enough

64-bit is *way* too much.  It allows for sub-micron navigation across our solar system, which is going a little far.

ArduPilot has seen suggestions to add another 4 bits of precision to our lat/lng - that takes your positional accuracy on Earth from ~1cm to ~1mm.

Has this PR gotten a little rerailed?  It was just supposed to be a nudge to move people to using `COMMAND_INT` - which we are now looking at forcing at a technical level anyway....

hamishwillee:
> Has this PR gotten a little rerailed? It was just supposed to be a nudge to move people to using `COMMAND_INT` - which we are now looking at forcing at a technical level anyway....

@peterbarker Thanks for commenting. Yes, the discussion got derailed from "can we use these words to reduce confusion about scaling and use of the `_INT` messages to "can we have a single message for sending commands so there is no confusion". 

The original PR is more about "should we add a comment to every mav_cmd with lat/lon hinting how it can be used, so people scale it properly. The idea being that https://mavlink.io/en/messages/common.html#COMMAND_INT is not all that clear.

I'm going to close this.

What I think I should do is dynamically update the description of each message with `hasLocation` to indicate that it should be sent in COMMAND_INT by preference and scaled.  - see https://github.com/mavlink/mavlink/pull/2603

PR #1959, opened 2023-03-08
Fix enum value for payload battery

hamishwillee:
Enum values are supposed to be prefixed by their enum name. This changes `MAV_BATTERY_TYPE_PAYLOAD` to `MAV_BATTERY_FUNCTION_PAYLOAD`. This requires a recompile for users that specifically access this item, but is non-breaking over the wire.


peterbarker (approved):

hamishwillee:
FYI @julianoes @auturgy merging - this makes sense.

PR #1967, opened 2023-03-29
COMMAND_ACK progress/result_param2 clarification

hamishwillee:
There is no value in having `progress` being "either progress or result_param1" other than potentially it allows two separate errors to be reported at a time (which you could do anyway by providing a bitmask as your enum). However it does make the command harder to understand and implement. 

So what this does is clarify that `progress` is for progress (only) and `result_param2` is for additional result information - and that this is a separate command specific enum, with 0 value meaning "unknown", that must be documented in the associated command. 

This matches the known implementations that use `COMMAND_ACK`. It is possible that some use progress for this feedback, and this change will make those "non compliant" - but we think fairly unlikely.

This follows on from #1965 and discussion in MAV call.

@julianoes @auturgy Good with this?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5592,7 +5592,7 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID (of acknowledged command).</field>
       <field type="uint8_t" name="result" enum="MAV_RESULT">Result of command.</field>
       <extensions/>
-      <field type="uint8_t" name="progress" invalid="UINT8_MAX" units="%">The progress percentage when result is MAV_RESULT_IN_PROGRESS (UINT8_MAX if the progress is unknown). Values: [0-100].</field>
+      <field type="uint8_t" name="progress" invalid="UINT8_MAX" units="%">The progress percentage when result is MAV_RESULT_IN_PROGRESS. Values: [0-100], or UINT8_MAX if the progress is unknown.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5592,8 +5592,8 @@
       <field type="uint16_t" name="command" enum="MAV_CMD">Command ID (of acknowledged command).</field>
       <field type="uint8_t" name="result" enum="MAV_RESULT">Result of command.</field>
       <extensions/>
-      <field type="uint8_t" name="progress" invalid="UINT8_MAX">Also used as result_param1, it can be set with an enum containing the errors reasons of why the command was denied, or the progress percentage when result is MAV_RESULT_IN_PROGRESS (UINT8_MAX if the progress is unknown).</field>
-      <field type="int32_t" name="result_param2">Additional parameter of the result, example: which parameter of MAV_CMD_NAV_WAYPOINT caused it to be denied.</field>
+      <field type="uint8_t" name="progress" invalid="UINT8_MAX" units="%">The progress percentage when result is MAV_RESULT_IN_PROGRESS (UINT8_MAX if the progress is unknown). Values: [0-100].</field>
```
hamishwillee, line 5595:
```suggestion
      <field type="uint8_t" name="progress" invalid="UINT8_MAX" units="%">The progress percentage when result is MAV_RESULT_IN_PROGRESS. Values: [0-100], or UINT8_MAX if the progress is unknown.</field>
```

julianoes (approved):

PR #1971, opened 2023-04-03
storm32: Use 'GIMBAL_MANAGER_CAP_FLAGS' over 'GIMBAL_DEVICE_CAP_FLAGS'

### LINKED ISSUE #1969: `storm32` and `common` has inconsistent usage for `GIMBAL_DEVICE_CAP_FLAGS`
From: common
https://github.com/mavlink/mavlink/blob/d6d86d3f0c90a67c275d6833931f7c712b858dba/message_definitions/v1.0/common.xml#L6952

And storm32
https://github.com/mavlink/mavlink/blob/d6d86d3f0c90a67c275d6833931f7c712b858dba/message_definitions/v1.0/storm32.xml#L318

GIMBAL_DEVICE_CAP_FLAGS is being used in 32 and 16 bits, which one should be used ?

patrickelectric:
in `STORM32_GIMBAL_MANAGER_INFORMATION`

May fix #1969


hamishwillee:
We can't do this. As I understand it the device/manager flags have some semantic differences, and in any case we want them to be able to diverge in future if needed.

@julianoes can you confirm

julianoes (commented):
I don't think that's right but it would be up to @olliw42 to decide.

PR #1972, opened 2023-04-05
common.xml: remove MAV_CMD_ACK

peterbarker:
This starts off looking very much like MAV_RESULT but then starts to differ

Based off the name, remove this entirely so implementers are forced to move to MAV_RESULT.

Some discussion between @hamishwillee and myself here: https://github.com/ArduPilot/mavlink/pull/307

tldr is that this is broken as an enumeration to use in an extension field as a value of zero is always "unknown".

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2644,36 +2644,6 @@
         <description>Point toward of given id.</description>
       </entry>
     </enum>
-    <enum name="MAV_CMD_ACK">
-      <description>ACK / NACK / ERROR values as a result of MAV_CMDs and for mission item transmission.</description>
-      <entry value="0" name="MAV_CMD_ACK_OK">
-        <description>Command / mission item is ok.</description>
-      </entry>
-      <entry value="1" name="MAV_CMD_ACK_ERR_FAIL">
-        <description>Generic error message if none of the other reasons fails or if no detailed error reporting is implemented.</description>
-      </entry>
-      <entry value="2" name="MAV_CMD_ACK_ERR_ACCESS_DENIED">
-        <description>The system is refusing to accept this command from this source / communication partner.</description>
-      </entry>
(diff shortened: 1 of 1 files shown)


hamishwillee:
This is an example of an enum that a particular MAV_CMD might specify could be returned as a value in [`COMMAND_ACK.result_param2`](https://mavlink.io/en/messages/common.html#COMMAND_ACK). 

For example, you might document in [MAV_CMD_NAV_WAYPOINT](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_WAYPOINT) that it can be sent as a command, but if rejected a [`MAV_CMD_ACK`](https://mavlink.io/en/messages/common.html#MAV_CMD_ACK) value could be set in `COMMAND_ACK.result_param` to tell you why - such as [MAV_CMD_ACK_ERR_X_LAT_OUT_OF_RANGE](https://mavlink.io/en/messages/common.html#MAV_CMD_ACK_ERR_X_LAT_OUT_OF_RANGE).

I think the concept is really good - wouldn't it be useful to know more detail on why some commands failed? 

But this enum is not used anywhere and IMO is flawed:
- Extensions should have a 0 value meaning "unknown", otherwise you can't tell if the field is used or not.
- You could have multiple things being invalid and we can't specify more than one unless we define as a bitmask.

Upshot, I'm kind of interested in the idea, but I think we should think about this in the context of some specific messages. As it is now this feels like "a good idea that was never tested out". IMO we should delete it, and perhaps have a quick discussion of whether we can reinvent it more usefully (if we find a stakeholder).

@auturgy @julianoes Got an opinion?

hamishwillee:
PS Normally I'd try drag in other opinions, but the current format of this message is simply not fit for purpose.

julianoes (approved):

hamishwillee:
Discussed in dev call. Not used by anyone and not fit for purpose. Removing. Thanks @peterbarker

PR #1976, opened 2023-04-19
common.xml: exclude any stored home location from MISSION_CURRENT.total

peterbarker:
ArduPilot stores its home location in slot-0.

Before this change, in the case you have that home location and a single waypoint uploaded ArduPilot would report a total of 2 - but when running the mission towards this single waypoint would report going to waypoint 1.  This is contrary to expectations of the user, so exclude slot 0 from the count.

@hamishwillee please discuss but don't merge just yet?  See https://github.com/ArduPilot/ardupilot/pull/23230

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5274,7 +5274,7 @@
       </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
-      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle.  If the autopilot stores its home location as part of the mission this should be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle (on last item, sequence == total). If the autopilot stores its home location as part of the mission this will be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
       <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -5274,7 +5274,7 @@
       </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
-      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
+      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle.  If the autopilot stores its home location as part of the mission this should be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
```
hamishwillee, line 5277: @peterbarker My recollection was that the `total` was intended to indicate the maximum sequence value, such that when `sequence == total` you know that you're on the last item. This was badly worded (by me) because I don't think of the home location as a mission item. Your change corrects the intent. 

We should make this change, or possibly the following:

```suggestion
      <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle (on last item, sequence ==total).  If the autopilot stores its home location as part of the mission this will be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
```

@julianoes Thoughts?

hamishwillee:
@peterbarker We should absolutely make this change. I would like to merge it.

peterbarker:
On Wed, 19 Apr 2023, Hamish Willee wrote:

> @peterbarker We should absolutely make this change. I would like to merge it.

Great!

I'm good with your rewording - it includes the exact problem I was trying 
to solve (but without the whitespace discrepancies around "==" :-) )

hamishwillee:
Awesome. I merged my "updated" suggestion. @peterbarker Just to be clear, are you OK if I merge _now_? Julian how about you?

peterbarker:
> Awesome. I merged my "updated" suggestion. @peterbarker Just to be clear, are you OK if I merge _now_? Julian how about you?

Yes, thanks.  Discussed at DevCall and I updated the PR to conform to the semantics here.

We'll be merging that PR soon-ish.

Issue #1952, opened 2023-02-26
Installation instructions


pjalocha:
Would be very good to add simple installation instructions, as it is not trivial and would greatly speedup practical usage.
For example on Linux like Raspian/Ubuntu I had to run the following, before I could use #include <mavlink.h> in my code.

sudo apt-get install python3-pip
pip3 install pymavlink
git clone https://github.com/mavlink/mavlink.git
cd mavlink/
git submodule update --init --recursive
mkdir build
cd build
cmake ..
make
sudo make install

And then, for g++, the following is needed:

-I/usr/local/include/mavlink/v2.0/common/

which decides which MAVlink version one uses.

Once this is known/done all works as expected, but it takes time to find out, thus why not add such to README ?

hamishwillee:
These look pretty simple

https://mavlink.io/en/getting_started/installation.html
https://mavlink.io/en/getting_started/generate_libraries.html

Using libraries then depends on which language you're using. So for example, https://mavlink.io/en/mavgen_c/

I guess arguably you might state "C" is the most important language so simple instructions for that setup in the readme wouldn't go astray.

hamishwillee:
@pjalocha Looking at your instructions in more detail:
1. You do not need to do `pip3 install pymavlink`. That installs a prebuilt Python library that uses a slightly modified version of common.xml for ArduPilot, and not the official common.xml.
2. I have added a quickstart to the readme which summarizes how you would set this up on Ubuntu 20.04: https://github.com/mavlink/mavlink#quickstart.
   Note that this builds using `pymavlink.tools.mavgen`, thereby fetching the particular submodule of Pymavlink associated with the cloned repo.

Looking at the rest of your instructions what I think you are doing is using cmake to "build" the library and install it system wide. What I think that does is build all the C libraries for common and dialects and install them  to `/usr/local/include/mavlink/v2.0/common/`.
Is that right?
```
mkdir build
cd build
cmake ..
make
sudo make install
```

So then you're saying if you wan to use these in your code, when g++, the following is needed:
   ```
    -I/usr/local/include/mavlink/v2.0/common/
   ```
Is that all correct.

IF so, then I need to think on this a little more. The docs are written with a focus on writing your own dialect and are general to C, Python and so on. What I think your focus might be is using the existing C libraries in your own project. 
So for you the important thing is having a system wide install and a predictable way to access the libraries. 

Does that make sense?

pjalocha:
Hello, and thank you.

[...]

PR #1977, opened 2023-04-24
New cmakelists

JonasVautherin:
It seems to me that the current CMakeLists in mavlink/mavlink is very outdated, possibly not used at all. The proposition here is to replace it with a new one that allows installing MAVLink.

It can be used like this:

```sh

# Install the library with CMake
cmake -DCMAKE_INSTALL_PREFIX=install -Bbuild -S. -DMAVLINK_DIALECT=common -DMAVLINK_VERSION=2.0
cmake --build build --target install
```

The above commands result in MAVLink being installed in `install`, together with a `MAVLinkConfig.cmake` file that enables the use of MAVLink in CMake projects like this:

```cmake
find_package(MAVLink REQUIRED)

[...]

target_link_libraries(<target>
    MAVLink::mavlink
)
```

Notes @JonasVautherin:

* I removed the `pc.in` file, but I would be happy to bring back package-config if useful
* I originally made that [for MAVSDK a year ago](https://github.com/mavlink/MAVSDK/pull/1848), but as pointed out by @julianoes, it would make more sense if MAVLink did that upstream (i.e. here) instead of MAVSDK patching it.

Notes @julianoes:

* I've included the generation inside the cmake build step. Because it's an INTERFACE library, I needed to add the ALL dependency, so it gets generated every time. I haven't found a better way to do that yet.
* I've replaced the old example and based it on cmake, and added it to CI.
* I've updated the README, however, the docs will also need some updates.


julianoes:
Closes #1952

hamishwillee:
@auturgy Any comments? This looks good to me modulo the TODO on docs.

auturgy:
This silently removes Python 2 support: which is fine/great, but we should probably make it clear that we're doing so. Also I'd prefer that the default dialect is all.xml

CMakeLists.txt
```diff
@@ -1,248 +1,74 @@
...
+        --wire-protocol=${MAVLINK_VERSION}
+        --output ${CMAKE_CURRENT_BINARY_DIR}/include/mavlink/v${MAVLINK_VERSION}/
+        message_definitions/v1.0/${MAVLINK_DIALECT}.xml
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    DEPENDS message_definitions/v1.0/${MAVLINK_DIALECT}.xml
+    COMMENT "Generating C headers")
+
+# Unfortunately, the dependencies don't work for INTERFACE libraries.
+# The only way I could make it work is to add ALL which means it
+# will do the file generation every time even when nothing has changed.
```
JonasVautherin, line 32: Which makes sense, right? Why would you run the build if nothing has changed? :thinking: :sweat_smile:

README.md
```diff
@@ -32,14 +32,46 @@ You can then build the MAVLink2 C-library for `message_definitions/v1.0/common.x
...
+
+add_executable(my_program my_program.c)
+
+target_link_libraries(my_program PRIVATE MAVLink::mavlink)
+```
+
+And pass the local install directory to cmake (adapt to your directory structure):
+```
+cd ../my_program
+cmake -Bbuild -H. -DCMAKE_PREFIX_PATH=../mavlink/install
```
JonasVautherin, line 57: Hmm I think `CMAKE_PREFIX_PATH` has to be absolute. At least I always do `$(pwd)/`, I think that's why :grin:

JonasVautherin (commented):
Nice, thanks for the changes @julianoes!

CMakeLists.txt
julianoes: No, it never runs, even if everything changed.

README.md
julianoes: Good point

julianoes:
@auturgy all.xml as the default is good but it means users will get development.xml by default, right? Would they now get warnings if they used something from development.xml?

Regarding Python2, I guess I could allow it to fallback to Python 2 if it doesn't find 3. Or maybe I can omit specifying that. I myself don't have much compassion with users still on Python 2 but if it's easily possible I don't mind keeping support.

julianoes:
@auturgy I changed it to Python, so it should take 2 or 3 whatever your environment/system prefers.

Also, I'd rather not change it to all.xml because when I use something from development.xml I don't get a warning that it's not stable.

README.md
julianoes: Actually works for me like that.

CMakeLists.txt
julianoes: Without `ALL` it doesn't get generated on install but if you have a target that depends on it, it does, so that's the tricky bit.

JonasVautherin (commented):
Good for me, but I can't approve because I opened the PR :see_no_evil:

examples/c/README.md
```diff
@@ -0,0 +1,21 @@
+# Simple C example
+
+Simple example receiving and sending MAVLink v2 over UDP based on POSIX APIs (e.g. Linux, BSD, macOS).
+
+## Install MAVLink
+
+In top level directory, build and install the MAVLink headers locally into the install folder:
```
hamishwillee, line 7:
```suggestion
In top level directory, build and install the MAVLink headers locally into the install folder:

```

README.md
```diff
@@ -9,21 +9,21 @@ MAVLink is a very lightweight, header-only message library for communication bet
...
 sudo apt install python3-pip
-sudo apt install python3-lxml libxml2-utils
 
 # Clone mavlink into the directory of your choice
 git clone https://github.com/mavlink/mavlink.git --recursive
 cd mavlink
 
-# Set the PYTHONPATH environment variable to the path of the root of the cloned mavlink repository
-PYTHONPATH=$PWD
+python3 -m pip install pymavlink/requirements.txt
```
hamishwillee, line 26: Good idea. I'm a little concerned because I think there was a reason I had libxml2-utils, which is not in requirements.txt. I presume you tested this?

README.md
```diff
@@ -32,14 +32,46 @@ You can then build the MAVLink2 C-library for `message_definitions/v1.0/common.x
...
+
+```
+find_package(MAVLink REQUIRED)
+
+add_executable(my_program my_program.c)
+
+target_link_libraries(my_program PRIVATE MAVLink::mavlink)
+```
+
+And pass the local install directory to cmake (adapt to your directory structure):
```
hamishwillee, line 54:
```suggestion
And pass the local install directory to cmake (adapt to your directory structure):

```

README.md
```diff
@@ -32,14 +32,47 @@ You can then build the MAVLink2 C-library for `message_definitions/v1.0/common.x
 python3 -m pymavlink.tools.mavgen --lang=C --wire-protocol=2.0 --output=generated/include/mavlink/v2.0 message_definitions/v1.0/common.xml
 ```
 
+### Use from cmake
+
+To include the headers in cmake, install them locally, e.g. into the directory `install`:
```
hamishwillee, line 37: so install is the name of any subdirectory relative to the current location you are calling cmake? Do you need to create it first?

README.md
```diff
@@ -32,14 +32,47 @@ You can then build the MAVLink2 C-library for `message_definitions/v1.0/common.x
...
+```
+
+Then use `find_package` to get the dependency in `CMakeLists.txt`:
+
+```
+find_package(MAVLink REQUIRED)
+
+add_executable(my_program my_program.c)
+
+target_link_libraries(my_program PRIVATE MAVLink::mavlink)
```
hamishwillee, line 51: What `PRIVATE` mean here? Or more precisely, is this sufficient for anyone to get started? (it might be, just checking!)

README.md
hamishwillee: I think that you're saying you run this inside the root of the mavlink repo? That creates mavlink/install and builds the headers into it for C. You then reference that from your project as shown further down using the prefixpath?

README.md
```diff
@@ -32,14 +32,47 @@ You can then build the MAVLink2 C-library for `message_definitions/v1.0/common.x
...
+cmake -Bbuild -H. -DCMAKE_PREFIX_PATH=../mavlink/install
+```
+
+For a full example, check [examples/c](examples/c).
+
+*Note: even though we use `target_link_libraries` in cmake, it doesn't actually "link" to MAVLink as it's just a header-only library.*
+
+### Other instructions
+
 Instructions for using the C libraries are then covered in [Using C MAVLink Libraries (mavgen)](https://mavlink.io/en/mavgen_c/).
```
hamishwillee, line 67: Linked doc will need to be updated

hamishwillee:
Thanks @JonasVautherin @julianoes This looks really good, though I have not tried it, and I know little about cmake.

As discussed in the call, very happy for this to be merged, but would be good if we could first create a draft PR that updates
https://mavlink.io/en/mavgen_c/ and https://mavlink.io/en/mavgen_c/examples.html (below that).

README.md
JonasVautherin: `-DCMAKE_INSTALL_PREFIX` tells CMake where to install the project when the "install" target is run (`cmake --build build --target install`). By default it will be on your system (something like `/usr/lib/`). But for development it is good practice to install it locally somewhere. You could use an absolute path `-DCMAKE_INSTALL_PREFIX=/tmp/my-local-install`, or a relative path like here. I personally like to install in `build/install`, so that it gets cleaned when I `rm -rf build`. But that's really just a preference.

So really what this will do is install MAVLink in `<path/to/this/repo>/install/`, to be used lated with `-DCMAKE_PREFIX_PATH=/path/to/this/repo/install`. No need to create `mavlink/install`, CMake will do it.

> That creates mavlink/install and builds the headers into it for C. You then reference that from your project as shown further down using the prefixpath?

Precisely.

README.md
JonasVautherin: I think it should be fine, that's a standard CMake thing (documented [here](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents)). IMO people will either already understand it (or know how to find the docs), or they will happily copy-paste it.

In any case I wouldn't spend too much effort explaining how CMake targets work here :innocent:.

CMakeLists.txt
```diff
@@ -1,248 +1,74 @@
...
+message(STATUS "MAVLink version: ${MAVLINK_VERSION}")
+
+set(EXAMPLE_HEADER ${CMAKE_CURRENT_BINARY_DIR}/include/mavlink/${MAVLINK_DIALECT}/mavlink.h)
+
+add_custom_command(OUTPUT ${EXAMPLE_HEADER}
+    COMMAND Python::Interpreter
+        -m pymavlink.tools.mavgen
+        --lang=C
+        --wire-protocol=${MAVLINK_VERSION}
+        --output ${CMAKE_CURRENT_BINARY_DIR}/include/mavlink/
```
JonasVautherin, line 24: @julianoes: I am just realizing now the extent of changing this: it will break many includes in MAVSDK and in the code using mavlink-passthrough. I guess essentially all the code already using MAVLink a bit everywhere.

Would this work instead? I have kind of forgotten why we didn't do it in the first place, though I think I remember that you told me :see_no_evil:.

```suggestion
        --output ${CMAKE_CURRENT_BINARY_DIR}/include/mavlink/v2.0
```

julianoes: Hm, I thought it only requires the change here: https://github.com/mavlink/MAVSDK/blob/main/src/mavsdk/core/include/mavsdk/mavlink_include.h.in#L3

JonasVautherin: Oh, you're right, I had missed that :+1:. Sorry :see_no_evil:

hamishwillee:
THanks very guys. This looks awesome. Can you post back here to confirm integration into MAVSDK worked OK?

JonasVautherin:
Seems like it's working, though for the CI there I needed the following patch (more info [here](https://github.com/actions/setup-python/issues/121#issuecomment-777748504) and [here](https://cmake.org/cmake/help/v3.15/module/FindPython.html#module:FindPython):

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 92ed5676..bb05beda 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,20 @@ cmake_minimum_required(VERSION 3.13)
 
 project(mavlink)
 
+if (POLICY CMP0094)
+    cmake_policy(SET CMP0094 NEW)  # FindPython should return the first matching Python
+endif ()
+
+# Needed to find the right Python on macOS
+if (NOT DEFINED Python_FIND_FRAMEWORK)
+    set(Python_FIND_FRAMEWORK "LAST")
+endif()
+
+# Needed to find the right Python on Windows
+if (NOT DEFINED Python_FIND_REGISTRY)
+    set(Python_FIND_REGISTRY "LAST")
+endif()
+
 find_package(Python COMPONENTS Interpreter REQUIRED)
 
 if (NOT MAVLINK_DIALECT)
```

However I did not need the patch on my personal Windows, so I don't know if that should be changed here or not. Maybe let's keep this in mind in case someone has the issue later?

hamishwillee:
Thanks! I'm not the right person to ask. Is this change likely to be something "everyone" will need to do? In that case perhaps some docs/guidance on CI integration. I'm hoping you and @julianoes will at least convince yourselfs of the right path.

julianoes:
The fixes are presumably CI specific, so I wouldn't want to push them here.

JonasVautherin:
Ok, let's keep the patch in MAVSDK for now, and see how it evolves. At some point maybe someone using macOS/Windows will let us know what's really needed there :+1:.

PR #1979, opened 2023-04-26
common.xml: Add ZOOM_TYPE_HORIZONTAL_FOV to CAMERA_ZOOM_TYPE enum

sypaq-nexton:
Allow the user to specify the camera zoom level as a horizontal field-of-view (HFOV) value. The primary use case for this is for systems with two cameras (with different lens/sensor combinations) where you want the two captured images to be the same.


auturgy:
@hamishwillee this one probably doesn't need to go to a call, and can just go in I think.

rmackay9:
This makes sense to me.  It could be more convenient for some callers and allows them to avoid a calculation.

hamishwillee:
Thanks!

PR #1973, opened 2023-04-05
common - un-WIP gimbal messages

hamishwillee:
With ArduPilot implementing these too, and some manufacturers, presumably time to remove the WIP tagging.

Any issues I should be aware of? I.e. something that no one has implemented that should remain WIP?

@julianoes Service docs indicate `MAV_CMD_DO_GIMBAL_MANAGER_TRACK_RECTANGLE` but I think you guys decided that was camera API `MAV_CMD_CAMERA_TRACK_RECTANGLE` right? 

@julianoes @olliw42 @rmackay9 @auturgy FYI Suggestions, feedback, welcome.

PR #1980, opened 2023-04-27
Gimbal manager messages - remove WIP tagging

hamishwillee:
This follows on from discussion in https://github.com/mavlink/mavlink/pull/1973

The messages are already deployed and are therefore frozen. 

@Davidsastresas is working on a QGC implementation. According to @julianoes this PR is currently not compatible enough to merge, but is a fairly good indication that the existing manager set is OK. Unless there is a compelling argument that we need to change one of these messages I think this should merge.

FYI @rmackay9

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2100,8 +2100,6 @@
       </entry>
       <!-- entry value 900 reserved for MAV_CMD_PARAM_TRANSACTION (development.xml) -->
       <entry value="1000" name="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>High level setpoint to be sent to a gimbal manager to set a gimbal attitude. It is possible to set combinations of the values below. E.g. an angle as well as a desired angular rate can be used to get to this angle at a certain angular rate, or an angular rate only will result in continuous turning. NaN is to be used to signal unset. Note: a gimbal is never to react to this command but only the gimbal manager.</description>
         <param index="1" label="Pitch angle" units="deg" minValue="-180" maxValue="180">Pitch angle (positive to pitch up, relative to vehicle for FOLLOW mode, relative to world horizon for LOCK mode).</param>
         <param index="2" label="Yaw angle" units="deg" minValue="-180" maxValue="180">Yaw angle (positive to yaw to the right, relative to vehicle for FOLLOW mode, absolute to North for LOCK mode).</param>
@@ -2111,8 +2109,6 @@
         <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</param>
       </entry>
       <entry value="1001" name="MAV_CMD_DO_GIMBAL_MANAGER_CONFIGURE" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Gimbal configuration to set which sysid/compid is in primary and secondary control.</description>
         <param index="1" label="sysid primary control">Sysid for primary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
         <param index="2" label="compid primary control">Compid for primary control (0: no one in control, -1: leave unchanged, -2: set itself in control (for missions where the own sysid is still unknown), -3: remove control if currently in control).</param>
@@ -6870,8 +6866,6 @@
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
     <message id="280" name="GIMBAL_MANAGER_INFORMATION">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Information about a high level gimbal manager. This message should be requested by a ground station using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="cap_flags" enum="GIMBAL_MANAGER_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
@@ -6884,8 +6878,6 @@
       <field type="float" name="yaw_max" units="rad">Maximum yaw angle (positive: to the right, negative: to the left)</field>
     </message>
     <message id="281" name="GIMBAL_MANAGER_STATUS">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Current status about a high level gimbal manager. This message should be broadcast at a low regular rate (e.g. 5Hz).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags currently applied.</field>
@@ -6896,8 +6888,6 @@
       <field type="uint8_t" name="secondary_control_compid">Component ID of MAVLink component with secondary control, 0 for none.</field>
     </message>
     <message id="282" name="GIMBAL_MANAGER_SET_ATTITUDE">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>High level message to control a gimbal's attitude. This message is to be sent to the gimbal manager (e.g. from a ground station). Angles and rates can be set to NaN according to use case.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
@@ -6994,8 +6984,6 @@
       <field type="float" name="angular_velocity_z" units="rad/s" invalid="NaN">Z component of angular velocity in NED (North, East, Down). NaN if unknown.</field>
     </message>
     <message id="287" name="GIMBAL_MANAGER_SET_PITCHYAW">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>High level message to control a gimbal's pitch and yaw angles. This message is to be sent to the gimbal manager (e.g. from a ground station). Angles and rates can be set to NaN according to use case.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
@@ -7007,8 +6995,6 @@
       <field type="float" name="yaw_rate" units="rad/s" invalid="NaN">Yaw angular rate (positive: to the right, negative: to the left, NaN to be ignored).</field>
     </message>
     <message id="288" name="GIMBAL_MANAGER_SET_MANUAL_CONTROL">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>High level message to control a gimbal manually. The angles or angular rates are unitless; the actual rates will depend on internal gimbal manager settings/configuration (e.g. set by parameters). This message is to be sent to the gimbal manager (e.g. from a ground station). Angles and rates can be set to NaN according to use case.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2100,7 +2100,7 @@
       </entry>
       <!-- entry value 900 reserved for MAV_CMD_PARAM_TRANSACTION (development.xml) -->
       <entry value="1000" name="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW" hasLocation="false" isDestination="false">
-        <description>High level setpoint to be sent to a gimbal manager to set a gimbal attitude. It is possible to set combinations of the values below. E.g. an angle as well as a desired angular rate can be used to get to this angle at a certain angular rate, or an angular rate only will result in continuous turning. NaN is to be used to signal unset. Note: a gimbal is never to react to this command but only the gimbal manager.</description>
+        <description>Set gimbal manager pitch/yaw setpoints (low rate command). It is possible to set combinations of the values below. E.g. an angle as well as a desired angular rate can be used to get to this angle at a certain angular rate, or an angular rate only will result in continuous turning. NaN is to be used to signal unset. Note: only the gimbal manager will react to this command - it will be ignored by a gimbal device. Use GIMBAL_MANAGER_SET_PITCHYAW if you need to stream pitch/yaw setpoints at higher rate. </description>
         <param index="1" label="Pitch angle" units="deg" minValue="-180" maxValue="180">Pitch angle (positive to pitch up, relative to vehicle for FOLLOW mode, relative to world horizon for LOCK mode).</param>
         <param index="2" label="Yaw angle" units="deg" minValue="-180" maxValue="180">Yaw angle (positive to yaw to the right, relative to vehicle for FOLLOW mode, absolute to North for LOCK mode).</param>
         <param index="3" label="Pitch rate" units="deg/s">Pitch rate (positive to pitch up).</param>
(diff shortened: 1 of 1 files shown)


rmackay9:
Thanks for this.  I'm happy with the WIP being removed for these messages

- MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
- GIMBAL_MANAGER_INFORMATION
- GIMBAL_MANAGER_SET_ATTITUDE
- GIMBAL_MANAGER_SET_MANUAL_CONTROL <-- AP doesn't use this yet but it could

I think we should remove the GIMBAL_MANAGER_SET_PITCHYAW because we already have MAV_CMD_DO_GIMBAL_MANAGER_SET_PITCHYAW.  Surely we don't need both do we?

I think the WIP should stay in place for the GIMBAL_MANAGER_STATUS message though.

hamishwillee:
@julianoes Can you please comment on Randy's response

-  I think we should remove the GIMBAL_MANAGER_SET_PITCHYAW because we already have MAV_CMD_DO_GIMBAL_MANAGER_SET_PITCHYAW. Surely we don't need both do we?
- I think the WIP should stay in place for the GIMBAL_MANAGER_STATUS message though.

hamishwillee:
@julianoes I agree with you re both, but obviously it is a source of confusion that there is both a command and a message that appear to do the same thing. Can we fix this in docs - i.e. add something that says when it is reasonable to use each?

I'll look at this probably next week.

julianoes:
> Can we fix this in docs - i.e. add something that says when it is reasonable to use each?

That would be good :).

rmackay9:
I'm sure that @julianoes knows what he's talking about but I don't understand a couple of things:

- why would the GCS be sending mav-cmd-do-gimbalmanager-pitchyaw commands at 20hz or 50hz?  I can guess that this would happen if the GCS was implementing point-at-location or maybe image tracking itself instead of doing it on the vehicle.
- if the GCS can handle sending commands at 20 to 50hz why wouldn't it be able to handle acks coming back at the same rate?

It's of course not hard for AP to support both, it's just a bit of extra code space.  We've actually got the code ready we just haven't merged it because it seems silly to have both.  Maybe it's not silly..

julianoes:
There are various reasons why a GCS would want to send setpoints at 20 Hz, for instance for a joystick or on screen controls to smoothly command it, similar to RC but not RC (because I'd rather send a typed value in degrees than PWM us). Or, it could be an application on the companion sending gimbal attitude at a high rate for smooth control.

Now, sending 20 Hz as a command is possible in theory, but the command/ack Mavlink protocol is fundamentally not designed for that. For instance, if the command method is used in QGC it would automatically queue those commands, retry them several times, and that leads to delays due to buffering best case and complete congestion of commands worst case. (This is not my theoretical invention, I have seen exactly that with the old CONTROL command.)

Now you can argue that you can just treat these commands differently and not wait for ack, etc. In that case you need special code and documentation to make sure they are treated differently. You create basically an exception to "treat a command like a streamed message" at which point why not just use a streamed message instead?

Also, let's envision a future where we have mavlink 3 and it's based on a transport which e.g. makes sure commands have quality of service to arrive exactly once (instead of hand implementing that ourselves like now) and streamed messages are best effort but low latency. Now suddenly the command messages would be a problem again.

If however you press the button in a ground station to go to 90 deg down, or 0 forward, then you absolutely do want a command, as you don't want to have to send it 3 times to hope that it arrives at least once.

So there are use cases for both and that's why they both exist.

Davidsastresas:
I agree with preserving the command and the message as @julianoes argumented, it is handy to have a message meant to be sent at high rate for this matter, for all this onscreen controls, and for possible expansions in the future. Also a companion computer might use them for commanding continuous tracking based on image feedback, etc.

But I also agree with @rmackay9 about holding just for a bit on the status message. Right now Ardupilot has some internal mount status reference based on the old mavlink 1 gimbal states that are not considered in GIMBAL_MANAGER_STATUS:

-     MAV_MOUNT_MODE_GPS_POINT
-      MAV_MOUNT_MODE_SYSID_TARGET
-     MAV_MOUNT_MODE_HOME_LOCATION

So I would like to solve this, although maybe only adaptation on Ardupilot side is needed. Randy and I talked the other day about this, and we think the above status could probably be a subset of mavlink targeting, considered  in GIMBAL_MANAGER_STATUS, so I kindly ask for your opinion on this front, what do you think should be done in Ardupilot and or mavlink about it? Because those 3 modes are actually useful actions, but maybe they should be considered "mavlink targeting status", but in this case, how could we trigger them using gimbal protocol v2? 

One could argue they could be managed by the autopilot under "mavlink targeting" gimbal manager state, so the mount_set_mode messages could be directed to the autopilot in order to engage them? If so, how would this look like from the GCS? Do Px4 has support for something like this? 

Thanks!

julianoes:
> - MAV_MOUNT_MODE_GPS_POINT

So the way I thought about this was that is not a "mode" as such but instead you send it specific message to do that task: [MAV_CMD_DO_SET_ROI_LOCATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION).

What is probably missing is a status flag, right? Should we add a `GIMBAL_MANAGER_FLAGS_ROI_LOCATION_TRACKING`?

> - MAV_MOUNT_MODE_SYSID_TARGET

Here, I would use [MAV_CMD_DO_SET_ROI_SYSID](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_SYSID).

Again, if you think that a status bit is required, what about adding `GIMBAL_MANAGER_FLAGS_ROI_SYSID_TRACKING`.

> - MAV_MOUNT_MODE_HOME_LOCATION

I don't find that one. Should we add a command `MAV_CMD_DO_SET_ROI_HOME` similar to the commands above? Or should we add an enum to [MAV_CMD_DO_SET_ROI_LOCATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION) specifying `HOME`?

Regarding the status flags, I wonder if we should have a different flag for each of these cases, or just have one ROI flag, but then a `ROI_STATUS` message containing all information of what is actually being tracked?

Davidsastresas:
Thanks for the answer!

I would be more inclined about a "global" ROI flag, and then somewhere else specify which kind of ROI. Because I think of a lot of extra ROI cases we could have use for, and it would not make sense to add all of them to the flags. Makes sense?

rmackay9:
@julianoes,

Txs for the explanation on the ACK handling.  It seems like QGC should have some internal state to say whether it wants to guaranteed delivery or not to avoid the issue bleeding out into the mavlink message space.  Anyway, I'm not fussed enough to press the point further.

Re the modes, I agree with @Davidsastresas that within AP we should move GPS_POINT, SYSID_TARGET and HOME_LOCATION "modes" to be subsets of the MAVLINK_TARGETING mode and I agree they can be triggered with the MAV_CMDs.

I guess the purpose of reporting the submode to the GCS is so that it can show buttons properly (e.g. maybe show the currently active mode as a depressed button) and maybe draw some lines on the map showing where the gimbal is pointed (either at an angle or at a location)?

By the way, how should the GIMBAL_STATUS message's secondary_control_sysid be populated?  Should the gimbal manager automatically shift the sysid/compid from primary to secondary when the primary is changed?

julianoes:
> Txs for the explanation on the ACK handling. It seems like QGC should have some internal state to say whether it wants to guaranteed delivery or not to avoid the issue bleeding out into the mavlink message space. Anyway, I'm not fussed enough to press the point further.

I would not say that the issue is bleeding out of QGC but rather that the command protocol is a fundamental MAVLink protocol (see https://mavlink.io/en/services/command.html#sequences) and QGC is implementing it.

So then you agree to have one bit signalling ROI or how would it look like?

> By the way, how should the GIMBAL_STATUS message's secondary_control_sysid be populated? Should the gimbal manager automatically shift the sysid/compid from primary to secondary when the primary is changed?

That hadn't occurred to me. The secondary control is up to an implementation. It would be one possibility to treat it as a fallback. I had thought of it more as a primary and nudge control. For instance if you have primary control setting a ROI on an object, you could have secondary control being able to nudge it by a couple of degrees. Once secondary control is let go, it snaps back to the initial ROI. In PX4 it's not implemented at the moment.

message_definitions/v1.0/common.xml
```diff
@@ -2100,8 +2100,6 @@
       </entry>
       <!-- entry value 900 reserved for MAV_CMD_PARAM_TRANSACTION (development.xml) -->
       <entry value="1000" name="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>High level setpoint to be sent to a gimbal manager to set a gimbal attitude. It is possible to set combinations of the values below. E.g. an angle as well as a desired angular rate can be used to get to this angle at a certain angular rate, or an angular rate only will result in continuous turning. NaN is to be used to signal unset. Note: a gimbal is never to react to this command but only the gimbal manager.</description>
```
hamishwillee, line 2103: @julianoes Does this work for you? I'll add something similar for the other cases if this one is OK.

```suggestion
        <description>Set gimbal pitch/yaw attitude setpoint on gimbal manager. It is possible to set combinations of the values below. E.g. an angle as well as a desired angular rate can be used to get to this angle at a certain angular rate, or an angular rate only will result in continuous turning. NaN is to be used to signal unset. Note: only the gimbal manager will react to this command - it will be ignored by a gimbal device. Use GIMBAL_MANAGER_SET_PITCHYAW if you need to stream pitch/yaw setpoints at high rate. </description>
```

message_definitions/v1.0/common.xml
```diff
@@ -6994,8 +6984,6 @@
       <field type="float" name="angular_velocity_z" units="rad/s" invalid="NaN">Z component of angular velocity in NED (North, East, Down). NaN if unknown.</field>
     </message>
     <message id="287" name="GIMBAL_MANAGER_SET_PITCHYAW">
-      <wip/>
-      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>High level message to control a gimbal's pitch and yaw angles. This message is to be sent to the gimbal manager (e.g. from a ground station). Angles and rates can be set to NaN according to use case.</description>
```
hamishwillee, line 6987:
```suggestion
      <description>Set gimbal manager pitch and yaw angles (high rate message). This message is to be sent to the gimbal manager (e.g. from a ground station) and will be ignored by gimbal devices. Angles and rates can be set to NaN according to use case. Use MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW for low-rate adjustments that require confirmation.</description>
```

hamishwillee:
@julianoes I've proposed some wording to explain when you use the different command/message - essentially just high rate/low rate. I'll let users work out what they think is meant by those terms. If OK, please accept those suggestions.

The rest of the discussion looks like this removal is OK, but that there are other things you want to resolve.
1. OK then to merge this?
2. Are there other things in this discussion that need to be captured, and if so, can one of you take care of that?

message_definitions/v1.0/common.xml
julianoes: Yep that's good.

message_definitions/v1.0/common.xml
julianoes: Yep, thanks!

julianoes:
@rmackay9 Are you ok with stabilizing the control/deconfliction GIMBAL_MANAGER_STATUS logic as it is now or do you have changes that you would like to make? If not, then we can merge this PR with the docstring improvements and WIP removals.

julianoes:
Awesome thanks for the note @rmackay9!

rmackay9:
Ah, one thing that I forgot to mention was that I noticed that GIMBAL_MANAGER_SET_PITCHYAW (the message) uses radians (or rad/sec) while the MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW (the command) uses degrees (or deg/sec).

I suspect this was done on purpose, perhaps because the command can appear in a mission list and users are more comfortable using degrees rather than radians... still, it's a bit odd to have such similar messages/commands using different units.  It's a bit of a gotcha.

hamishwillee:
@rmackay9 That sux a bit. @julianoes Was this intentional? If not, possible to change? Either way, any hints at how we might not the gotcha better?

julianoes:
> I suspect this was done on purpose, perhaps because the command can appear in a mission list and users are more comfortable using degrees rather than radians... still, it's a bit odd to have such similar messages/commands using different units. It's a bit of a gotcha.

Good catch. Not sure if this was an oversight or on purpose.
A bit odd indeed. It is implemented correctly in MAVSDK but actually wrong on the PX4 side: The input for angular rate is assumed to be radians for the command when it is degrees.

Also, when looking at this I discovered that we have two streamed APIs `GIMBAL_MANAGER_SET_ATTITUDE` and `GIMBAL_MANAGER_SET_PITCHYAW` so that's indeed a bit overlapping. PX4 only implements the `GIMBAL_MANAGER_SET_ATTITUDE` (and `GIMBAL_MANAGER_SET_MANUAL_CONTROL`).
Apologies, I hadn't realized that earlier. Is that why you consider `GIMBAL_MANAGER_SET_PITCHYAW` to be somewhat redundant? I can see why... And I'd consider removing/deprecating it then.

rmackay9:
I was considering MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW and GIMBAL_MANAGER_SET_PITCHYAW as overlapping (which we've already discussed).  I had assumed that the GIMBAL_MANAGER_SET_ATTITUDE was there for the rare case where someone wanted to set the roll as well.

AP currently implements:

- MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW
- GIMBAL_MANAGER_SET_ATTITUDE

.. but doesn't implement these two:

- GIMBAL_MANAGER_SET_PITCHYAW <-- we have a PR for this
- GIMBAL_MANAGER_SET_MANUAL_CONTROL <-- no PR yet but planning to add support eventually

For me, I don't mind whether GIMBAL_MANAGER_SET_PITCHYAW stays or goes.

Davidsastresas:
I actually find in some rare cases setting roll could be useful from a companion computer. For instance, see this feature on some DJI drones to orientate the sensor vertically or horizontally, for using the whole sensor in portrait or landscape mode. It could be useful to have something like that available for a companion computer at some point.

julianoes:
I agree that we should have GIMBAL_MANAGER_SET_ATTITUDE. The question is whether we also need GIMBAL_MANAGER_SET_PITCHYAW or if that's just a bit redundant, give all we have to do is just convert to quaternion and use GIMBAL_MANAGER_SET_ATTITUDE.

rmackay9:
@julianoes,

Not sure if this is the best place to discuss or not but I noticed that the [GIMBAL_MANAGER_INFORMATION](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_INFORMATION) and [GIMBAL_MANAGER_STATUS](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_STATUS) messages doesn't specify whether the **gimbal_device_id** field is 0-based or 1-based.

My guess is it should be 1-based because other messages (including [GIMBAL_MANAGER_SET_ATTITUDE](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_SET_ATTITUDE)) with **gimbal_device_id** fields are 1-based with "0" meaning "all gimbals".  Surely we don't want some messages 0-based and others 1-based.

From a quick peek at your team's code though it looked like you might be using 0-based in at least some cases.

julianoes:
@rmackay9 good point. I would say it should match [GIMBAL_MANAGER_SET_ATTITUDE](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_SET_ATTITUDE), so 1 based:

> Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).

hamishwillee:
> I agree that we should have GIMBAL_MANAGER_SET_ATTITUDE. The question is whether we also need GIMBAL_MANAGER_SET_PITCHYAW or if that's just a bit redundant, given all we have to do is just convert to quaternion and use GIMBAL_MANAGER_SET_ATTITUDE.

@julianoes My thinking on this
-  Who would be break if we removed GIMBAL_MANAGER_SET_PITCHYAW?
- I tend to think that reducing the message set makes it easier for both implementers and adopters. 
- I don't understand quaternions. My assumption is that we can reasonably expect anyone who is likely to need this feature should be required to do so? 

Upshot, what's the cost of removal?

@rmackay9 Let me know if you want me to do the PR for https://github.com/mavlink/mavlink/pull/1980#issuecomment-1550594716 - I'd rather you do it, but most of all I want to make sure that it doesn't fall off the list :-)

rmackay9:
Hi @hamishwillee,

Ah, I'd love it if you fixed up the comments to save me the trouble.  Txs very much!

hamishwillee:
@rmackay9 @julianoes Done - please sanity check because I am not convinced that using mavlink component IDs was ever intended for the gimbal manager id: https://github.com/mavlink/mavlink/pull/1988

@julianoes Note the comment from Randy above: "From a quick peek at your team's code though it looked like you might be using 0-based in at least some cases." Worth checking?

julianoes:
> @julianoes Note the comment from Randy above: "From a quick peek at your team's code though it looked like you might be using 0-based in at least some cases." Worth checking?

Can you point me where that would be? It's a direct mapping to MAVLink component IDs which are also 1 based, and 0 there means "all" as well.

rmackay9:
@julianoes,

OK, txs for checking.  We're not too familiar with your team's code of course so it was probably a misunderstanding.

julianoes:
Well, I'm not saying it's all correct. It usually more likely isn't :smile:. This part I only just added recently:

Issue #1981, opened 2023-04-28
Implicit MAV_FRAME for MAV_CMDs sent as COMMAND_LONG


sypaq-nexton:
There are a number of `MAV_CMD_*` messages (such as [`MAV_CMD_DO_SET_ROI_LOCATION`](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI_LOCATION)) that include a `(lat, lon, alt)` location. 
* When sent as a [`COMMAND_INT`](https://mavlink.io/en/messages/common.html#COMMAND_INT), the [`MAV_FRAME`](https://mavlink.io/en/messages/common.html#MAV_FRAME) is included, and the location can be correctly decoded.
* When sent as a [`COMMAND_LONG`](https://mavlink.io/en/messages/common.html#COMMAND_LONG), the `MAV_FRAME` is implicit, and can lead to errors when the sending and receiving systems use a different reference frame (e.g. see https://github.com/ArduPilot/ardupilot/pull/23266).

There are also a number of `MAV_CMD_*` messages (such as [`MAV_CMD_DO_GO_AROUND`](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GO_AROUND)) that include an altitude, but not altitude reference (either explicitly as is done in [`MAV_CMD_DO_CHANGE_ALTITUDE`](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CHANGE_ALTITUDE), or implicitly in the description of the altitude field). 

The definitions of the `MAV_CMD_*` messages that include a location or altitude should be updated to explicitly define the `MAV_FRAME` to use when using these messages in a `COMMAND_LONG`.

sypaq-nexton:
The other option is to update the [Command Protocol](https://mavlink.io/en/services/command.html) to explicitly state that `MAV_CMD_*` messages that include a location must only be accepted as a `COMMAND_INT`, and must be NACKed (possibly with a new error code in [MAV_RESULT](https://mavlink.io/en/messages/common.html#MAV_RESULT)) when received as a `COMMAND_LONG`.

hamishwillee:
@sypaq-nexton Yes, this is a known issue.

The simple fact is that no one should be using `COMMAND_LONG` for positional commands. Firstly because of the loss of precision, but also because, as you noticed, there is no way to set the frame. Worse, there is no way to know the frame that a particular flight stack might be using.

[...]

PR #1982, opened 2023-05-01
common.xml: Add MAV_RESULT_COMMAND_LONG_ONLY and MAV_RESULT_COMMAND_INT_ONLY to MAV_RESULT

sypaq-nexton:
Per the descriptions on `COMMAND_INT` and `COMMAND_LONG`, certain `MAV_CMD`s should only be sent as one or the other. However, there is no mechanism for the receiving system to indicate this. The closest we can get at the moment is by using `MAV_RESULT_DENIED`, but that is ambiguous as to whether the issue is the `COMMAND_*` message or the actual parameters.

This PR adds two new entries to the `MAV_RESULT` enum, `MAV_RESULT_COMMAND_LONG_ONLY` and `MAV_RESULT_COMMAND_INT_ONLY`, to address this.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2736,6 +2736,9 @@
       <entry value="6" name="MAV_RESULT_CANCELLED">
         <description>Command has been cancelled (as a result of receiving a COMMAND_CANCEL message).</description>
       </entry>
+      <entry value="7" name="MAV_RESULT_COMMAND_LONG_ONLY">
+        <description>Command is valid, but it is only accepted when sent as a COMMAND_LONG (as it has float values for params 5 and 6).</description>
+      </entry>
     </enum>
     <enum name="MAV_MISSION_RESULT">
       <description>Result of mission operation (in a MISSION_ACK message).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2739,6 +2739,9 @@
       <entry value="7" name="MAV_RESULT_COMMAND_LONG_ONLY">
         <description>Command is valid, but it is only accepted when sent as a COMMAND_LONG (as it has float values for params 5 and 6).</description>
       </entry>
+      <entry value="8" name="MAV_RESULT_COMMAND_INT_ONLY">
+        <description>Command is valid, but it is only accepted when sent as a COMMAND_INT (as it encodes a location in params 5, 6 and 7, and hence requires a reference MAV_FRAME).</description>
+      </entry>
     </enum>
     <enum name="MAV_MISSION_RESULT">
       <description>Result of mission operation (in a MISSION_ACK message).</description>


tridge:
we've discussed on ardupilot dev call and we think we will auto automatic mapping for all commands so can be sent as either, but precision of location commands will be poor (approx 1.5m) if sent as a COMMAND_LONG

hamishwillee:
@tridge It's not just about resolution. The COMMAND_LONG doesn't include frame information, so unless we standardize that for all commands across flight stacks a bunch of commands are ambiguous, in particular with respect to altitude.

Does ArduPilot always use the same frame for all commands sent using COMMAND_LONG, and is that going to be fixed forever?

I'm also a bit concerned that it is bad for users - now they can send the message in both `COMMAND_LONG` and `COMMAND_INT`, but if they are sending in `COMMAND_LONG` when command int is supported they are probably doing it unintentionally.

Wouldn't this approach of rejecting the command if sent in the wrong type allow progressive addition of COMMAND_INT support?

peterbarker:
> Wouldn't this approach of rejecting the command if sent in the wrong type allow progressive addition of COMMAND_INT support?

I think our path in ArduPilot might be to start supporting all location-containing commands via `COMMAND_INT`.

Are you suggesting that we split the list of commands in the standard - so going forward the stacks only support a command via one and only one of `COMMAND_LONG` and `COMMAND_INT`?

The marginal cost of supporting e.g. `MAV_COMPONENT_ARM_DISARM` via either `COMMAND_LONG` or `COMMAND_INT` is very low for ArduPilot.  Right up 'til someone wants to use parameter 6 to store a float!

If we were to make a split, would the descriptions for every command have to include which mechanisms they can be used?

sypaq-nexton:
@peterbarker : This PR came from issue #1981 (and from https://github.com/ArduPilot/ardupilot/pull/23266) -- the issue is ambiguous altitude reference frames in `COMMAND_LONG`s. 

While I _personally_ think the solution is to update each command with its expected altitude frame, that's not really feasible. This PR is intended to provide autopilots with a mechanism to remove this ambiguity on a case-by-case basis. 

> Are you suggesting that we split the list of commands in the standard - so going forward the stacks only support a command via one and only one of COMMAND_LONG and COMMAND_INT?

No:
* Some commands can be sent as either as either `COMMAND_LONG` and `COMMAND_INT`
* Some commands should only be sent as a `COMMAND_INT`, as they need a `MAV_FRAME`
* Some commands should only be sent as a `COMMAND_LONG`, as params 5 and 6 need to be floats

> If we were to make a split, would the descriptions for every command have to include which mechanisms they can be used?

Not necessarily: there is already guidance on which type should be used in the descriptions of the [`COMMAND_INT`](https://mavlink.io/en/messages/common.html#COMMAND_INT) and [`COMMAND_LONG`](https://mavlink.io/en/messages/common.html#COMMAND_LONG) messages.

peterbarker:
> > If we were to make a split, would the descriptions for every command have to include which mechanisms they can be used?
> 
> Not necessarily: there is already guidance on which type should be used in the descriptions of the [`COMMAND_INT`](https://mavlink.io/en/messages/common.html#COMMAND_INT) and [`COMMAND_LONG`](https://mavlink.io/en/messages/common.html#COMMAND_LONG) messages.

Should we actually deprecate the use of `COMMAND_LONG` for locations, rather than just recommending it?  At some stage, then, we can remove the ambiguities (and LONG->INT conversion code :-) )

sypaq-nexton:
> Should we actually deprecate the use of `COMMAND_LONG` for locations, rather than just recommending it?

Not necessarily. To me, it comes down to a choice between:
1. Add the expected `MAV_FRAME` to the definition of each command that contains a location (i.e. is annotated with either `hasLocation=true` or `isDestination=true`); or
2. Deprecate `COMMAND_LONG` for all `MAV_CMD`s that contain a location.

The issue with option one is that it requires consensus between the different flight stacks. 

The issue with option two is that you still need a way for the autopilot to communicate that it would accept the `MAV_CMD` if sent as a `COMMAND_INT` (hence this PR).

hamishwillee:
> Should we actually deprecate the use of COMMAND_LONG for locations, rather than just recommending it? At some stage, then, we can remove the ambiguities (and LONG->INT conversion code :-) )

Yes. 

We should have done so all along. What this does is provide a mechanism to do it gracefully.

We can probably use `hasLocation=true` or `isDestination=true`) to annotate the Command_int version. However we'd need to think about how to annotate "requires command_long".

We might also use that `COMMAND_ACK.results_param2` thingy to tell people what frame/s are supported on a particular stack for that particular message - finally a good use for it :-)

The other options I see are like those that @sypaq-nexton points to, and unpalatable for those reasons:

1. Annotate every method with the recommendation of what commands it should be sent with (and you still don't get any runtime feedback as a user that the frame is ambiguous)
2. Add a frame field to COMMAND_LONG ... and you still don't get any feedback that the command long version has less precision. Plus the other issues re truncation etc.
3. Agree a blanket frame for any command sent using COMMAND_LONG unless explicitly specified in its description of the field.

This PR  feels to me like the first palatable suggestion we've had.

hamishwillee:
Are there any objections to this, suggestions for improvement? Please add any comments here. 

I plan to discuss/merge after MAV call next week unless there are strong objections or better suggestions before then.

Note that there would be follow on PRs to update the COMMAND_LONG/INT docs and protocol docs about this, along with instrumenting the XML.

julianoes:
I think this is a good idea.

hamishwillee:
This has been added to call for tonight. I expect it will probably merge

hamishwillee:
Agreed in dev call. Thanks @sypaq-nexton

PR #1986, opened 2023-05-16
development: add available modes monitor

bkueng:
This brings the message from https://github.com/mavlink/rfcs/pull/23.
Allows to dynamically update the list of modes.


hamishwillee (approved):
This was effectively approved by the MAVLink dev call on approval of the RFC.
As with everything, it can't move out of WIP without rugged testing and confirmation that this would be adopted by other flight stacks

hamishwillee:
@bkueng Sorry, doing cleanup and realized I forgot to merge this.

PR #1988, opened 2023-05-18
Gimbal manager/status - device ids

hamishwillee:
This change arises out of https://github.com/mavlink/mavlink/pull/1980#issuecomment-1552208805 "should the device id be 0 or 1 indexed"

@julianoes I have taken the text you suggested from [GIMBAL_MANAGER_SET_ATTITUDE](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_SET_ATTITUDE). However I'm not completely convinced it is correct to use gimbal component IDs.

The id absolutely does have to be indexed from one. However I thought the concept of the id as set in the manager was to abstract component IDs for the gimbal - so we'd always use 1-6 and not have to think about what device IDs were actually mapped to - the manager would sort that out.

Conceptually I find that much easier when working with a gimbal manager. I'm also a bit worried that you might have both mavlink and non-mavlink gimbals and then it isn't necessarily obvious whether you think you're talking to a component id 1 or a device that is mapped to one.

The only reason we couldn't do this is if something other than the gimbal manager needs to directly know the component IDs of the gimbals, and there is no way to infer the mapping from the gimbal manager.

What am I missing? (I could be completely wrong about the intent)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6890,7 +6890,7 @@
       <description>Current status about a high level gimbal manager. This message should be broadcast at a low regular rate (e.g. 5Hz).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags currently applied.</field>
-      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID (or 1-6 for a non-MAVLink gimbal).</field>
+      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID.</field>
       <field type="uint8_t" name="primary_control_sysid">System ID of MAVLink component with primary control, 0 for none.</field>
       <field type="uint8_t" name="primary_control_compid">Component ID of MAVLink component with primary control, 0 for none.</field>
       <field type="uint8_t" name="secondary_control_sysid">System ID of MAVLink component with secondary control, 0 for none.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6878,7 +6878,7 @@
       <description>Information about a high level gimbal manager. This message should be requested by a ground station using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="cap_flags" enum="GIMBAL_MANAGER_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
-      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for.</field>
+      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID (or 1-6 for a non-MAVLink gimbal).</field>
       <field type="float" name="roll_min" units="rad">Minimum hardware roll angle (positive: rolling to the right, negative: rolling to the left)</field>
       <field type="float" name="roll_max" units="rad">Maximum hardware roll angle (positive: rolling to the right, negative: rolling to the left)</field>
       <field type="float" name="pitch_min" units="rad">Minimum pitch angle (positive: up, negative: down)</field>
@@ -6890,7 +6890,7 @@
       <description>Current status about a high level gimbal manager. This message should be broadcast at a low regular rate (e.g. 5Hz).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags currently applied.</field>
-      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for.</field>
+      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID (or 1-6 for a non-MAVLink gimbal).</field>
       <field type="uint8_t" name="primary_control_sysid">System ID of MAVLink component with primary control, 0 for none.</field>
       <field type="uint8_t" name="primary_control_compid">Component ID of MAVLink component with primary control, 0 for none.</field>
       <field type="uint8_t" name="secondary_control_sysid">System ID of MAVLink component with secondary control, 0 for none.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6878,7 +6878,7 @@
       <description>Information about a high level gimbal manager. This message should be requested by a ground station using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="cap_flags" enum="GIMBAL_MANAGER_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
-      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID (or 1-6 for a non-MAVLink gimbal).</field>
+      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID.</field>
       <field type="float" name="roll_min" units="rad">Minimum hardware roll angle (positive: rolling to the right, negative: rolling to the left)</field>
       <field type="float" name="roll_max" units="rad">Maximum hardware roll angle (positive: rolling to the right, negative: rolling to the left)</field>
       <field type="float" name="pitch_min" units="rad">Minimum pitch angle (positive: up, negative: down)</field>


rmackay9 (approved):
Thanks very much for this.  This is fine or alternatively we could make it completely consistent with the other gimbal_device_id field descriptions by making it, "Component ID of gimbal device to address or 1-6 for non-MAVLink gimbal".

hamishwillee:
@rmackay9 I don't want to make it completely consistent because that other field is a setter command. So the value can be set to zero. In these cases the value is the reported value of the device ID, and as discussed, must **not** be zero. Hope that makes sense.

message_definitions/v1.0/common.xml
```diff
@@ -6878,7 +6878,7 @@
       <description>Information about a high level gimbal manager. This message should be requested by a ground station using MAV_CMD_REQUEST_MESSAGE.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="cap_flags" enum="GIMBAL_MANAGER_CAP_FLAGS" display="bitmask">Bitmap of gimbal capability flags.</field>
-      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for.</field>
+      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID (or 1-6 for a non-MAVLink gimbal).</field>
```
julianoes, line 6881: MAVLink component ID is correct. I'm not sure about using 1-6 for non-mavlink gimbal because that can potentially conflict with real component IDs.

For now, I have used the same component ID of the manager, if the gimbal device is on the same device, e.g. an RC gimbal in PX4.

Or we could just say that the limitation is that it's only possible to have one gimbal with the same component ID as the manager.

rmackay9: I also wondered about the conflict but I think in general MAVLink gimbals will use the MAV_COMP_ID_GIMBAL, xx_GIMBAL2, etc component ids.

julianoes: In general yes, but we don't want to rely on it.

rmackay9: I think the only choice is to either overload (which is what we're doing now) or add another field.  .. and this applies to all the gimbal-manager messages of course not just this one.

julianoes: Right, or sorta have more than one component ID on a "thing" that has two gimbals, also not great.

rmackay9: I think it is OK to have multiple GIMBAL_MANAGER_INFORMATION messages being sent with the same sysid/compid because they're differentiated by the gimbal_device_id and the message is going to be streamed at some regular rate.

I'm not sure if you're perhaps worried about handling a gimbal-manager letting go of one gimbal and starting to use another but if this is the case (which would be a real edge case) then the GCS could just throw away the reference to the old gimbal after some timeout (5sec or whatever).  I'm probably misunderstanding the concern though.

julianoes: My concern is only for the case of "internal gimbal device" where you can use the numbers 1 to 6 which are in itself component IDs and so it's not clear to me if that's a good idea.

rmackay9: Indeed, it is a slight bit of ugliness.  I just think that perfection (e.g. adding another field or bit to differentiate them) comes at too high a cost so I'd vote for "fixing it in documentation" which means we add to the mavlink gimbalv2 spec that gimbal **devices** must never use compids from 0-6 (gimbal managers can use the lower component ids though).

julianoes: Ok that makes sense.

- 0 is to signal that it doesn't have a component.
- 1-5 for "same device/onboard"
- 6..255 usable component ID

hamishwillee: OK, except @julianoes you indicated 1-5, rather than 1-6 for the gimbal manager. We can update the docs next week because I want to make sure I get the sync of the spec document right too.

julianoes: Sorry, 1-6 is fine.

hamishwillee: @julianoes So the current docs for the device ID would be right, but we need to update the other docs for the device to make it clear that it can't used ID from range 0 - 6.

Are device messages like [`GIMBAL_DEVICE_ATTITUDE_STATUS`](https://mavlink.io/en/messages/common.html#GIMBAL_DEVICE_ATTITUDE_STATUS) ever emitted from an autopilot for hardware attached gimbals? My concern here is that if the gimbal is autopilot connected and this was emitted, a recipient would not be able to know what id to address it with.

hamishwillee: @julianoes ^^^^

dayjaby: @hamishwillee in gimbalv2 spec it says "Note that GIMBAL_DEVICE_ATTITUDE_STATUS is broadcast to anyone, so to the gimbal manager and also the ground station."

So it looks like "target_component" is a not necessary field in GIMBAL_DEVICE_ATTITUDE_STATUS.

> a recipient would not be able to know what id to address it with.

You mean the recipient does not know who sent the gimbal device attitude status? Isn't that contained in the compid of the mavlink message itself, or am I missing something?

julianoes: > Are device messages like [GIMBAL_DEVICE_ATTITUDE_STATUS](https://mavlink.io/en/messages/common.html#GIMBAL_DEVICE_ATTITUDE_STATUS) ever emitted from an autopilot for hardware attached gimbals?

I would say: yes for the case where the gimbal is not MAVLink, so e.g. just PWM. The component ID is then the same as the manager in that case, and the gimbal_device_id would be 1 for instance.

Does that make sense?

hamishwillee: > You mean the recipient does not know who sent the gimbal device attitude status? Isn't that contained in the compid of the mavlink message itself, or am I missing something?

@dayjaby  This all works for mavlink gimbals but I'm not sure for PWM-connected gimbals.

No, I mean that if [GIMBAL_DEVICE_ATTITUDE_STATUS](https://mavlink.io/en/messages/common.html#GIMBAL_DEVICE_ATTITUDE_STATUS) is emitted by a GIMBAL **MANAGER** that is implementing the gimbal device behaviour using PWM, there is no way to know which device it is associated with.

In this case you'd have the component ID of the flight stack, but the actual gimbal this is associated with could be the "virtual MAVLink gimbal" 1 to 6. For most messages it doesn't matter because the manager and device are separate, but for these ones that are emitted and supposed to be consumed by the GCS it might. Hence my question.

> So it looks like "target_component" is not a necessary field in GIMBAL_DEVICE_ATTITUDE_STATUS.

I would agree with that, though it is too late to remove.


> > Are device messages like [GIMBAL_DEVICE_ATTITUDE_STATUS](https://mavlink.io/en/messages/common.html#GIMBAL_DEVICE_ATTITUDE_STATUS) ever emitted from an autopilot for hardware attached gimbals?
> 
> I would say: yes for the case where the gimbal is not MAVLink, so e.g. just PWM. The component ID is then the same as the manager in that case, and the gimbal_device_id would be 1 for instance.
> 
> Does that make sense?

Not really. As above, what if you have 2 gimbals attached - which is it from?

dayjaby: In the case of a PWM gimbal, there should still be a "gimbal device" implementation. This can even be done on the same MCU as the gimbal manager, albeit having its own gimbal device component ID.

At least in PX4 the implementation would be straightforward and the gimbal manager itself doesn't have to handle any PWM stuff.

dayjaby: Note that such a custom gimbal device implementation can not only help with PWM gimbals, but with any MAVLink capable gimbal that does not adhere to the gimbalv2 standard.

hamishwillee: @dayjaby  I have been told that it is hard for flight stacks to implement components as separate entities with their own heartbeats, which is why no autopilot to my knowledge emits any component ID other than 1.
Instead flight stacks accept commands addressed to id 1 or 0 (and sometimes others) and route them internally to the component if there is only one. In some cases there are additional indexes or IDs that allow routing to "multiples of things" mapped by a flight stack. 

But perhaps this is not actually a problem. Generally you will have one gimbal with one PWM connected device, so it can have component id of one. You can have more autopilot-connected gimbals if you want, but then you have to give them a unique heartbeat etc. In practice they will do what has been done for cameras - use MAVLink if you need more than one.


@julianoes @rmackay9 How about that? The change would then be to remove the restriction of 1-6 reserved for PWM connected gimbals and just note that you can use id of that for a single PWM connected gimbal. If you want more you can use mavlink gimbals or properly implement a gimbal device component on the autopilot (which no one will do).

That is actually cleaner since we might need 6 gimbals, but for devices where that makes sense I suspect most would be connected via mavlink.

dayjaby: So your argument is we have to use comp id 1 just because we can't emit a HEARTBEAT with a different comp id? That doesn't make any sense to be honest. A HEARTBEAT not only has to have the correct comp id, but also the correct MAV_TYPE. What to set then for the type? MAV_TYPE_AUTOPILOT and/or MAV_TYPE_GIMBAL? Do we have to send two HEARTBEATs with the same component id but different types? With your suggestions it becomes really hard to determine whether a component is actually a gimbal device or not. All code snippets I can find right now test against MAV_TYPE_GIMBAL and MAV_COMP_ID_GIMBAL1-6

But looking at the mavlink code, it *is* possible to send HEARTBEATS with different IDs. Well, not with `_mav_finalize_message_chan_send` but with e.g. `mavlink_msg_to_send_buffer`.

> If you want more you can use mavlink gimbals or properly implement a gimbal device component on the autopilot (which no one will do).

1. It will take years until all MAVLink gimbals are *good* MAVLink gimbals and not some weird random abomination of the MAVLink standard. The latest gimbal I had the "pleasure" to work with has all its gimbal control via MAV_CMD_DO_DIGICAM_CONTROL.

2. Implementing gimbal device component on autopilot is the best thing you can do if you want to integrate these weird MAVLink typed gimbals, PWM gimbals or other weird Chinese gimbals having their own serial protocols.

julianoes: @dayjaby is right in that it is possible as long as a separate mavlink channel is selected which means the seq numbering from that separate compid is correct.

hamishwillee: Thanks @dayjaby Works for me. This right then?:

```suggestion
      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID.</field>
```

And we note that `GIMBAL_DEVICE_ATTITUDE_STATUS` must be implemented by a gimbal manager and emitted with the correct type/id .

dayjaby: Looking good to me. Thanks!

message_definitions/v1.0/common.xml
```diff
@@ -6890,7 +6890,7 @@
       <description>Current status about a high level gimbal manager. This message should be broadcast at a low regular rate (e.g. 5Hz).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags currently applied.</field>
-      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for.</field>
+      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID (or 1-6 for a non-MAVLink gimbal).</field>
```
hamishwillee, line 6893:
```suggestion
      <field type="uint8_t" name="gimbal_device_id" instance="true">Gimbal device ID that this gimbal manager is responsible for. This should be a MAVLink component ID.</field>
```

hamishwillee:
Great. I think we have agreement. Merging.

PR #1990, opened 2023-05-19
update instructions to install pymavlink

rotu:
update instructions to install pymavlink package instead of just its dependencies
note the instructions were broken anyway - you need the -r flag to install from a requirements file


hamishwillee (approved):
Thank you. Good catch.

PR #1994, opened 2023-05-25
isAltitude, hasLocation, isDestination discussion and application

hamishwillee:
This discussion evolved quite considerably. The resolution is that hasLocation and isDestination only apply to the param5,6,7 values, and should be applied that way. This now updates just a few MAV_CMD that don't meet this case.

----

[`MAV_CMD_DO_GO_AROUND`](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_GO_AROUND) has an altitude with units in metres but no specified frame.

While I suspect that most flight stacks would assume this to be AMSL because they should have a barometer, but some might have terrain info, rangefinder, or assume AGL from home. If sent in a COMMAND_LONG the frame is ambiguous.

For discussion, this adds a new attribute `hasAltitude` that should be applied to any item where `hasLocation` and `isDestination` are `false` but there is an altitude setting that does not have a fixed and specified frame.
This is needed so we can auto-add notes to indicate what command this should be sent in.

Initially I intended to change the attributes `hasLocation` and `isDestination` from false to true, which would allow me to auto-add a note that the command should be send in COMMAND_INT.

However @sypaq-nexton had a reasonable question - should `hasLocation` and `isDestination` only apply to things that have a position in the ground plane (like lat/lon).
The answer is yes. `isDestination` was invented so a GCS could identify waypoints for drawing lines through automatically. If we apply this to things without a location, the GCS might screw up rendering. 
I don't know for sure, but I suspect that `hasLocation` might have similar semantics.

1. Does adding this make sense.
2. Does adding it only if the other attributes make sense, or is it better to add it to everything that has an altitude, including those that have location. 
3. You probably only add it if the frame is fixed in the description. The only case you would want it would be if param 5/6 was used for altitude. We would avoid this, so I think we don't add it for those cases.

Falls out of discussion in https://github.com/mavlink/mavlink/pull/1992

@julianoes @auturgy Opinions? If you agree I will update XSD.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1645,7 +1645,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="191" name="MAV_CMD_DO_GO_AROUND" altitudeOnly="true">
+      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false">
         <description>Mission command to safely abort an autonomous landing.</description>
         <param index="1" label="Altitude" units="m">Altitude</param>
         <param index="2">Empty</param>
@@ -1715,7 +1715,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="198" name="MAV_CMD_DO_SET_ROI_SYSID">
+      <entry value="198" name="MAV_CMD_DO_SET_ROI_SYSID" hasLocation="false" isDestination="false">
         <description>Mount tracks system with specified system ID. Determination of target vehicle position may be done with GLOBAL_POSITION_INT or any other means. This command can be sent to a gimbal manager but not to a gimbal device. A gimbal device is not to react to this message.</description>
         <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">System ID</param>
         <param index="2" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</param>
@@ -2102,7 +2102,7 @@
         <param index="6">Reserved</param>
         <param index="7">Reserved</param>
       </entry>
-      <entry value="420" name="MAV_CMD_INJECT_FAILURE">
+      <entry value="420" name="MAV_CMD_INJECT_FAILURE" hasLocation="false" isDestination="false">
         <description>Inject artificial failure for testing purposes. Note that autopilots should implement an additional protection before accepting this command such as a specific param setting.</description>
         <param index="1" label="Failure unit" enum="FAILURE_UNIT">The unit which is affected by the failure.</param>
         <param index="2" label="Failure type" enum="FAILURE_TYPE">The type how the failure manifests itself.</param>
@@ -2606,7 +2606,7 @@
         <param index="7">Reserved</param>
       </entry>
       <!-- from ardupilotmega.xml (hence ID is in that range) -->
-      <entry value="42006" name="MAV_CMD_FIXED_MAG_CAL_YAW">
+      <entry value="42006" name="MAV_CMD_FIXED_MAG_CAL_YAW" hasLocation="false" isDestination="false">
         <description>Magnetometer calibration based on provided known yaw. This allows for fast calibration using WMM field tables in the vehicle, given only the known yaw of the vehicle. If Latitude and longitude are both zero then use the current vehicle location.</description>
         <param index="1" label="Yaw" units="deg">Yaw of vehicle in earth frame.</param>
         <param index="2" label="CompassMask">CompassMask, 0 for all.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1645,7 +1645,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false">
+      <entry value="191" name="MAV_CMD_DO_GO_AROUND" altitudeOnly="true">
         <description>Mission command to safely abort an autonomous landing.</description>
         <param index="1" label="Altitude" units="m">Altitude</param>
         <param index="2">Empty</param>


message_definitions/v1.0/common.xml
```diff
@@ -1533,7 +1533,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false">
+      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="true" isDestination="true">
```
hamishwillee, line 1536:
```suggestion
      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false" hasAltitude="true">
```

sypaq-nexton:
> While I suspect that most flight stacks would assume this to be AMSL because they should have a barometer, but some might have terrain info, rangefinder, or assume AGL from home.

* PX4 doesn't use the altitude: [FixedwingPositionControl.cpp#L227](https://github.com/PX4/PX4-Autopilot/blob/ee96d209d78510fe7bcc2c34e526648fa79c0135/src/modules/fw_pos_control/FixedwingPositionControl.cpp#L227)
* ArduPilot uses the altitude as height above home:
    * [GCS_Mavlink.cpp#L1038](https://github.com/ArduPilot/ardupilot/blob/2c675246c977c626a63aa325a186c78f9b1427f1/ArduPlane/GCS_Mavlink.cpp#L1038), which calls into [ArduPlane.cpp#L685](https://github.com/ArduPilot/ardupilot/blob/2c675246c977c626a63aa325a186c78f9b1427f1/ArduPlane/ArduPlane.cpp#L685)

sypaq-nexton:
I think adding `hasAltitude` makes sense. And unless there's already an established meaning for `hasLocation`, I think specifying that as lat/lon only makes sense. So most commands would get tagged with both `hasLocation` and `hasAltitude.

sypaq-nexton:
My other question still stands though:
* Does `hasLocation` imply that the latitude and longitude are in params 5 and 6, or just somewhere in the command?
* Ditto for the new `hasAltitude` -- in param 7, or anywhere?

Would something like the following be better:
* `latitudeParam="5", longitudeParam="6", altitudeParam="7"`

It would allow for better markup on the documentation and/or better code-gen.

hamishwillee:
> > While I suspect that most flight stacks would assume this to be AMSL because they should have a barometer, but some might have terrain info, rangefinder, or assume AGL from home.
> 
> * PX4 doesn't use the altitude: [FixedwingPositionControl.cpp#L227](https://github.com/PX4/PX4-Autopilot/blob/ee96d209d78510fe7bcc2c34e526648fa79c0135/src/modules/fw_pos_control/FixedwingPositionControl.cpp#L227)
> * ArduPilot uses the altitude as height above home:
>   
>   * [GCS_Mavlink.cpp#L1038](https://github.com/ArduPilot/ardupilot/blob/2c675246c977c626a63aa325a186c78f9b1427f1/ArduPlane/GCS_Mavlink.cpp#L1038), which calls into [ArduPlane.cpp#L685](https://github.com/ArduPilot/ardupilot/blob/2c675246c977c626a63aa325a186c78f9b1427f1/ArduPlane/ArduPlane.cpp#L685)

As you say - PX4 uses the greater of a parameter or the current height: https://docs.px4.io/main/en/flight_modes/mission.html#operator-abort
I guess we could specify as "AGL in whatever way calculated by the vehicle". Not sure we need to be more specific. 

It's fine by me that PX4 ignores the altitude - I'd hope that ArduPilot would below some threshold too. What isn't so fine is that there is no way for a recipient to know that. I wonder if we might accept the command (because if someone does DOAROUND they really want it to happen) but allow feedback on the particular params that were ignored in the command ack. This might be a genuine use for `COMMAND_ACK.result_param2`.

Perhaps also worth making an invalid value of 0 for the altitude to indicate "use whatever value you want".

hamishwillee:
> I think adding `hasAltitude` makes sense. And unless there's already an established meaning for `hasLocation`, I think specifying that as lat/lon only makes sense. So most commands would get tagged with both `hasLocation` and `hasAltitude`.

That is a good idea too, and if we were starting "from scratch" I'd say for sure. Whether it is viable depends on whether ground stations or anyone are using `hasLocation` for anything. 

> My other question still stands though:
> 
> * Does `hasLocation` imply that the latitude and longitude are in params 5 and 6, or just somewhere in the command?
> * Ditto for the new `hasAltitude` -- in param 7, or anywhere?

Its still an open question. IMO we should make that assumption (param 5, 6) and use the convention to enforce it. Mostly because of the desire for higher resolution data, and for the ground station to be able to just assume that these values are the ones it should use.

Not so sure for `hasAltitude`, mostly because there are cases with multiple altitudes. 

> Would something like the following be better:
> 
> * `latitudeParam="5", longitudeParam="6", altitudeParam="7"`
> 
> It would allow for better markup on the documentation and/or better code-gen.

Yes it would.

hamishwillee:
@DonLakeFlyer We're trying to better instrument MAVLink positional information so that we can auto-add documentation about whether commands should be sent in command_int. We've got some problems because there are cases where various values are not in the expected positions. 

Minimally we would like to
- be able to differentiate cases that have only an altitude.
- know what we can assume about the isDestination/hasLocation

What assumptions does QGC make? 
- Does it use `isDestination`? [I think it does, to draw lines between waypoints]
- Does it actually use `hasLocation`? [not sure]
- For both of these, do the commands assume that lat, lon, altitude, are in param 5, 6, 7?

We have some proposals on the table. We can't really assess them without knowing the implications. For example, adding `hasAltitude`, and making `hasLocation` just mean lat/lon.

DonLakeFlyer:
QGC has the json formation for mission items doc'ed here: https://github.com/mavlink/qgroundcontrol/blob/master/src/MissionManager/MissionCommandUIInfo.h#L88

The original proposal was based on that QGC code. I'm not exactly sure if there is one to one mapping or no with the mavlink spec. But it does tell you all the stuff that QGC needs to know about to creates it's ui for planning missions.

The generic mavlink version of QGC mavlink ui mission info are in MissionManager directory an all start with MavCmdInfo*.json. The firmware specific overrides are in for example: FirmwarePlugin/PX4/PX4-MavCmdInfo.json. You can use these as examples. I thought I had docmented all of this somewhere in the dev guide but can't seem to find it any more.

Can you take a look at that and see if you still have questions. 

From a standpoint of param 5/6/7 being lat/lon/alt, QGC does not use the MissionCommandUInfo json to know where those are. That is hardwired into the code and various UI bits.

DonLakeFlyer:
Also keep in mind that QGC still uses it's own json description files for all of this. It does not currently use the mavlink spec metadata directly. Although if there is a one to one correspondence it could be changed to use this with some coding and build system work.

DonLakeFlyer:
> Although if there is a one to one correspondence it could be changed to use this with some coding and build system work.

Sorry haven't been in this space for a while. That statement is incorrect. It is way more complicated than having a single definition of the mavlink messages. If anyone wants the gory details on why that it is it would easier to explain in a call.

hamishwillee:
Thanks @DonLakeFlyer

@sypaq-nexton 

What I'm taking away from this is:
1. The position co-ordinate metadata QGC uses is
   - `specifiesCoordinate`  - Command specifies a lat/lon/alt coordinate
   - `specifiesAltitudeOnly` - Command specifies an altitude only (no coordinate)
   - `standaloneCoordinate`  Vehicle does not fly through coordinate associated with command (exampl: ROI)
2.  `hasLocation =  specifiesCoordinate`, but then standaloneCoordinate is the opposite of `isDestination` - i.e. mavlink marks the locations that are specifically destinations, while QGC marks the locations that are specifically ROI-like. 
3. I suspect that QGC always deals in positions (lat, lon, alt) or altitudes (alt), so in that case it is easiest to have attributes that represent those cases.
4. We can change any of this without breaking QGC.

I'm ambivalent as to whether it is better to have (match QGC-ish):

- `hasLocation` - has lat/lon/alt
- `altitudeOnly` - has `alt`

or

- `hasLocation` - has lat/lon
- `hasAltitude` - has `alt`

OR latitudeParam="5", longitudeParam="6", altitudeParam="7"

The first is the closest to actual use, the last is the most flexible. I lean towards one or two. I slightly prefer 2

DonLakeFlyer:
The QGC definitions leans towards the common cases requiring less doc'ing and only the un-common cases require extra tagging. So since:
* Majority or lat/lon mission items include alt so specifiesCoordinate means lat/lon/alt so less flagging to do to indicate items which are only alt, or only lat/lon
* Majority of lat/lon items are also for flying through hence only a few items require standaloneCoordinate

Simpler, less tags needed everywhere

hamishwillee:
Thanks - I appreciate your reasoning. It makes sense for flying drones, and that is the case that MAVLink messages have been mostly evolved for. However it isn't quite so reasonable for ground systems/any command where you don't care about the altitude. I'm also not concerned about a few extra tags here. We can focus on air systems and how the command set looks now, or try make something general. Or we can decide we don't like the churn.

The good compromise general format would be:

- hasLocation - has lat/lon
- hasAltitude - has alt
- isStandaloneCoordinate - is an ROI and such. Otherwise assumed to be a location.

The low churn variant would be just to add `altitudeOnly`

Either case I think we assume that these only apply to cases where the values apply to param5, 6, 7.

I'll see what the mavlink call people think.

message_definitions/v1.0/common.xml
```diff
@@ -1533,7 +1533,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false">
+      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false" hasAltitude="true">
```
hamishwillee, line 1536:
```suggestion
      <entry value="191" name="MAV_CMD_DO_GO_AROUND" altitudeOnly="true">
```

hamishwillee:
We discussed in dev call. The general feeling was that this is pretty pointless unless someone is using them or indicates they want to use them and can be a stakeholder in the discussion. For now, the main point is to be clear about what the set do mean. Since we want to reduce churn, and keep it possible for QGC to use this data:

1. Add `altitudeOnly`, a synonym for QGC `specifiesAltitudeOnly`.
2. Clarify that the tagging ONLY applies altitude in param7 and lat/lon (x/y) in param5, 6. The tags would not be applied otherwise.

I'll add the attribute to XSD and we can proceed from there.

FYI @sypaq-nexton

sypaq-nexton:
> 1. Add altitudeOnly, a synonym for QGC specifiesAltitudeOnly.
> 2. Clarify that the tagging ONLY applies altitude in param7 and lat/lon (x/y) in param5, 6. The tags would not be applied otherwise.

Sounds good to me! Thanks for keeping me in the loop.

hamishwillee:
You're welcome. Good to have someone looking at this with fresh eyes.

message_definitions/v1.0/common.xml
hamishwillee: This is false because as discussed here, hasLocation requires things in the param5,6,7. Though this does contain location information, a GCS can't use it given the wrong params are defined.

```suggestion
<entry value="42006" name="MAV_CMD_FIXED_MAG_CAL_YAW" hasLocation="false" isDestination="false">
<description>Magnetometer calibration based on provided known yaw. This allows for fast calibration using WMM field tables in the vehicle, given only the known yaw of the vehicle. If Latitude and longitude are both zero then use the current vehicle location.</description>
<param index="1" label="Yaw" units="deg">Yaw of vehicle in earth frame.</param>
<param index="2" label="CompassMask">CompassMask, 0 for all.</param>
<param index="3" label="Latitude" units="deg">Latitude.</param>
<param index="4" label="Longitude" units="deg">Longitude.</param>
<param index="5">Empty.</param>
<param index="6">Empty.</param>
<param index="7">Empty.</param>
```

hamishwillee: Needed because the default values for these attributes is true.


```suggestion
      <entry value="198" name="MAV_CMD_DO_SET_ROI_SYSID"  hasLocation="false" isDestination="false">
```

hamishwillee: This is false because as discussed here, hasLocation requires things in the param5,6,7. Though this does contain location information, a GCS can't use it given the wrong params are defined.

```suggestion
      <entry value="42006" name="MAV_CMD_FIXED_MAG_CAL_YAW" hasLocation="false" isDestination="false">
```

```suggestion
      <entry value="420" name="MAV_CMD_INJECT_FAILURE" hasLocation="false" isDestination="false">
```

message_definitions/v1.0/common.xml
```diff
@@ -1533,7 +1533,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false">
+      <entry value="191" name="MAV_CMD_DO_GO_AROUND" altitudeOnly="true">
```
hamishwillee: Revert - won't use altitudeOnly

```suggestion
      <entry value="191" name="MAV_CMD_DO_GO_AROUND" hasLocation="false" isDestination="false">
```

message_definitions/v1.0/common.xml
```diff
@@ -1603,7 +1603,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="198" name="MAV_CMD_DO_SET_ROI_SYSID">
+      <entry value="198" name="MAV_CMD_DO_SET_ROI_SYSID"  hasLocation="false" isDestination="false">
```
```suggestion
      <entry value="198" name="MAV_CMD_DO_SET_ROI_SYSID" hasLocation="false" isDestination="false">
```

hamishwillee:
This stalled for a while, and I want to move it on or close it. The last update was https://github.com/mavlink/mavlink/pull/1994#issuecomment-1581629423 which resulted in agreement by dev call that
1. `hasLocation` and `isDestination` should only apply to lat/lon/alt in param 5,6,7. This makes sense because the attributes need to provide predictable behaviour for a consumer. If the lat/lon/alt are in different params this can't be relied on.
2. Add `altitudeOnly`, a synonym for QGC specifiesAltitudeOnly.

The PR now just fixes up the three remaining commands that don't do the first case properly.

I abandoned `altitudeOnly` (and closed the XSD change). It makes sense in theory for the same reason as hasLocation/isDestination. However in practice it is not useful, because there are only a few messages that are altitude only, and they don't use the same param.

peterbarker (approved):

julianoes (approved):

DonLakeFlyer:
Give me a day or so to stare at this again.

DonLakeFlyer:
So for dropping altitudeOnly, it depends on what you want to use this information. For QGC this information tells it to create mission UI editors which only show altitude setting UI which is stored in param7. In the current QGC base mission command tree json there are two cases for that:

* MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT (30) — [MavCmdInfoCommon.json:240](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), altitude in param7
* MAV_CMD_CONDITION_CHANGE_ALT (113) — [MavCmdInfoCommon.json:469](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html), altitude in param7

So if that was there the QGC code drop it's base command json and just keep the overrides part of the command tree.

hamishwillee:
> So for dropping altitudeOnly, it depends on what you want to use this information. For QGC this information tells it to create mission UI editors which only show altitude setting UI which is stored in param7. In the current QGC base mission command tree json there are two cases for that:
> 
> * MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT (30) — MavCmdInfoCommon.json:240, altitude in param7
> * MAV_CMD_CONDITION_CHANGE_ALT (113) — MavCmdInfoCommon.json:469, altitude in param7
> 
> So if that was there the QGC code drop it's base command json and just keep the overrides part of the command tree.

@DonLakeFlyer Yes, that is the case we were after. I missed the two cases that do use param7 and only use param7. 

I'm going to merge this to fix up the other cases and create a new PR to see if we can get the atltidude in.

DonLakeFlyer:
> I'm going to merge this to fix up the other cases and create a new PR to see if we can get the atltidude in.

Thanks @hamishwillee.

PR #1996, opened 2023-05-25
MAV_CMD_VIDEO_START_CAPTURE et al change param5/6 to 0 default

hamishwillee:
[`MAV_CMD_VIDEO_START_CAPTURE`](https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_START_CAPTURE), [`MAV_CMD_VIDEO_STOP_CAPTURE`](https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_STOP_CAPTURE), 
[`MAV_CMD_AIRFRAME_CONFIGURATION`](https://mavlink.io/en/messages/common.html#MAV_CMD_AIRFRAME_CONFIGURATION), 

All have param 5, 6 [reserved](https://mavlink.io/en/guide/define_xml_element.html#reserved) as `NaN`, indicating that senders should always set the values to `NaN` even if they are unused. This allows the reserved values to be resused later on for properties where "0" is an allowed value of the setting, because you would still sent `NaN` to mean "no change".

This PR changes the reserved params to have default values of 0, such that a GCS should send 0 by default for the command in param5, 6. This allows the command to be send in a COMMAND_INT by by default. We are trying to encourage that most commands should be sendable in COMMAND_INT so it becomes a kind of default, with only very few exceptions where COMMAND_LONG must be used.

I don't think there are any downsides to this:
- The commands still have reserved float params that can be used if the commands need to be extended with a float param in future.
- GCS should be sending NaN in these values now in a COMMAND long, and should change to sending 0 - both in the long and int variants. Right now though, the field is not being used. There is plenty of time, possibly unlimited time, for flight stacks to adapt. 
 - I also wouldn't be surprised if they are sending 0 instead by mistake. 
 
This fell out of discussion in https://github.com/mavlink/mavlink/pull/1992

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2122,8 +2122,8 @@
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Total Images" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
         <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1), otherwise set to 0. Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted.</param>
-        <param index="5" reserved="true" default="NaN"/>
-        <param index="6" reserved="true" default="NaN"/>
+        <param index="5" reserved="true"/>
+        <param index="6" reserved="true"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
@@ -2132,6 +2132,8 @@
         <param index="2" reserved="true" default="NaN"/>
         <param index="3" reserved="true" default="NaN"/>
         <param index="4" reserved="true" default="NaN"/>
+        <param index="5" reserved="true"/>
+        <param index="6" reserved="true"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>
       <entry value="2002" name="MAV_CMD_REQUEST_CAMERA_IMAGE_CAPTURE" hasLocation="false" isDestination="false">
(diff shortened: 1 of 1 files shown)


julianoes:
I understand where you're coming from but by accepting this I'm basically creating work for myself and I have to go into all places where I set NaN by default and remove it again.

hamishwillee:
> I would challenge whether INT should be the default, I would argue LONG as the default is status quo and ok.

I think it is worth having a default that "just works" with most commands because I think it increases the chance that users will get the behaviour they expect. If we could have just one message I'd advocate for that. As it is now you have to know too much to get it right.  I will accept an argument that no default makes sense and people just have to check each case, but I won't like it.

If you accept a default is "a good thing", then it is best to choose the default that gives expected behaviour when used with most commands. That is not COMMAND_LONG since there are far more commands with position than commands that use real values in param 5/6.

Perhaps you can explain why COMMAND_LONG would be a better default?

> I understand where you're coming from but by accepting this I'm basically creating work for myself and I have to go into all places where I set NaN by default and remove it again.

Would it force you to change things if we allows it to be set default as 0 if sent in a command_int. What if we allowed it to be sent as either? That would mean we might one day have to do contortions to allow it to be set as zero, but probably we'd just never use those values.

hamishwillee:
> Because it allows to use NaN to indicate not set 😂. The problem is that by saying INT is the default, it still isn't because that's not how things are generally implemented. And changing it all over won't happen quickly because commands always require both ends to change and so you either have to implement always retrying both in case INT is not supported or you just stick with what worked in the past which is usually LONG.

So the reason I had/ve this opinion is that Don tells me that QGC always defaults to zero for the values of used parameters. 

If COMMAND_LONG is used everywhere then that's irritating, but I see your point. I'd still argue that COMMAND_INT is better as a default though, because you can use it for the vast majority more commands.

julianoes:
> So the reason I had/ve this opinion is that Don tells me that QGC always defaults to zero for the values of used parameters.

Right, and so does ArduPilot. It's probably mostly me influenced by Lorenz pushing for NaN. I'm fine going to 0 instead, it seems like that's the path of least resistance.

hamishwillee:
@julianoes So it is OK to merge as-is? (Just to be very clear).

hamishwillee:
Thank you!

PR #1998, opened 2023-05-26
Add Payload and Autonomy Microservice Support from Marine Dialect

bazfp:
This has been in development for over a year and is currently operational on underwater vehicles (UUVs) like the popular [Remus vehicle from Hydroid.](https://en.wikipedia.org/wiki/REMUS_(AUV)) It's stable enough now we have decided to present the changes upstream.

This is version **v22.08.01**

There are also matching functional implementations in MAVSDK which are also in operation on both vehicle-side and autonomy-control side.

> UUVs tend to operate in very comms limited environments (no RF / limited acoustic comms). As with most robots they have an on-board "autopilot" or "front-seat" that controls low-level vehicle movement (Control Fins to set Heading, Control Motor to Set Speed etc). They also often run a generic Autonomy system on an additional "back-seat" computer, which accepts high-level commands and breaks them down for the front-seat vehicle to consume (Scan this point, Scan this area etc.)
> 
> Typically each vehicle manufacturer creates their own bespoke interface to allow external-control from an autonomy system, we are pushing to standardise this approach with an open standard.
> 
> We are using MAVLink now as our vehicle interface for this Autonomy <-> Frontseat link and have been working with vehicle manufacturers have MAVLink integrated as a "frontseat" control interface.

The things of note are:

- **Generic Vehicle Payload Registering, Status and Control**
  - Ability to add/swap/remove/enable/disable/pause generic external sensors, sonar, sidescan, FLS, etc.
- **Support for External Autonomy Systems** - Components, Health, State and Control
  - These autonomy systems run on a companion computer and issue MAVLINK commands to the vehicle front-seat controller.
- Water current, DVL (Doppler Velocity Logger)
  -  Domain specific environment monitoring
- Various assorted messages for information sharing between squads. 

# Payload Microservice

The Payload Microservice is used to discover, configure and control `Payloads<Payload>` (which could be sensors, actuators or other connected components) registered on a `MAVLink` system.

You can use this Microservice to:

>   - Retrieve a list of all Payloads on the system
>   - Retrieve the status of a single Payload
>   - Receive updates on newly registered Payloads
>   - Control Payload state

It is expected that the AutoPilot maintains a list of registered Payloads, containing for each:

>   - ID
>       - Any number in the range \(1 - UINT8\_MAX\)
>       - **Must be unique** in reguards to the other Payload IDs
>   - Type
>       - Must be one of the values defined in the `PAYLOAD_TYPE` enum
>   - States
>       - The states that the Payload can be set to
>       - 16 byte bitmask corresponding to the `PAYLOAD_STATE` enum
>   - Health
>       - The Payload health
>       - 16 byte bitmask corresponding to the `PAYLOAD_HEALTH` enum
>   - State
>       - The the current states of the vehicle
>       - 16 byte bitmask corresponding to the `PAYLOAD_STATE` enum
>   - Name
>       - *Optional*
>       - Maximum length of `16` characters

A system can use the Payload Microservice to request details about the Payloads on the MAVLink system and to configure said Payloads.

When an external system enables a Payload it will set the Payload state to:

  - PAYLOAD\_STATE\_POWERED = true
  - PAYLOAD\_STATE\_ARMED = true
  - PAYLOAD\_STATE\_ACTIVE = true

When a Payload is disabled, An external system will set the Payload state to:

  - PAYLOAD\_STATE\_POWERED = true
  - PAYLOAD\_STATE\_ARMED = false
  - PAYLOAD\_STATE\_ACTIVE = false

In order to request a `PAYLOAD_STATUS` from the AutoPilot, A system will
send a [MAV\_CMD\_REQUEST\_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE) with:

  - <span class="title-ref">Param 1</span> set to `44206` (the Message
    ID of `PAYLOAD_STATUS`)
  - <span class="title-ref">Param 2</span> set to the Payload ID of the
    Payload being enquired about

For a list of the enumerations used in the payload messages, please see `payload-enum-summary`.

The AutoPilot also needs to respect the implementation details layed out in `payload-microservice-operation-exceptions`.

##### Workflows

`payload_discovery_process` below shows the communication sequence external systems use to retrieve the list of registered Payloads from the `AutoPilot` (assuming all operations succeed).
payload_list_request
> Payload Discovery Operation

In more detail, the sequence of operations is:

1.  Autonomy sends a `PAYLOAD_REQUEST_LIST` to start the discovery and
    download process
      - Autonomy will wait for a specific length of time for a
        `PAYLOAD_COUNT` response from the AutoPilot. If no response is
        received during that time, Autonomy will re-send the
        <span class="title-ref">PAYLOAD\_REQUEST\_LIST</span> message.
2.  The AutoPilot receives the message and responds with a
    `PAYLOAD_COUNT` where the <span class="title-ref">count</span> field
    is the number of registered payloads
3.  Autonomy sends a `PAYLOAD_LIST_ITEM_REQUEST` for the first list item
    ( `list position = 0`)
      - Autonomy will wait for a specific length of time for a
        `PAYLOAD_LIST_ITEM_REQUEST` response from the AutoPilot. If no
        response is received during that time, Autonomy will re-send the
        `PAYLOAD_LIST_ITEM_REQUEST` message.
4.  The AutoPilot receives the `PAYLOAD_LIST_ITEM_REQUEST` message and
    responds with the requested Payload details via a
    `PAYLOAD_LIST_ITEM` message
5.  Autonomy and AutoPilot repeat steps 3 and 4 until Autonomy has the
    deatils of all Payloads
6.  After receiving the details of the last Payload, Autonomy responds
    with a single `PAYLOAD_LIST_ACK` with the
    <span class="title-ref">type</span> field set to
    `PAYLOAD_LIST_ACCEPTED` (see the `PAYLOAD_LIST_RESULT` enum) to
    indicate Payload discovery success
7.  The AutoPilot receives the `PAYLOAD_LIST_ACK` containing
    `PAYLOAD_LIST_ACCEPTED` and understands the operation to be
    complete.

**List request errors are considered unrecoverable see `payload-microservice-errors` for details on how to handle errors.**

`payload_status_update` below shows the communication sequence that autonomy follows to request the status of a registered Payload.

Autonomy sends a [MAV\_CMD\_REQUEST\_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE) message with the ID of the Payload it wants the status of (`IDX`) to the AutoPilot. The AutoPilot then responds with a `PAYLOAD_STATUS` message containing the Type, Health and Status of the Payload.

payload_status_update
> Payload Status Update Operation

The diagram below shows the communication sequence Autonomy uses to change the State of a registered Payload. The `MAV_CMD_PAYLOAD_SET_STATE` message is used to set the State of a Payload using a bitmask corresponding to the `PAYLOAD_STATE` enum, where the bits are set to `1` to indicate the state is `true` and `0` to indicate the state is `false`.

As part of the Payload discovery process, the valid States that can be set for each Payload are specified in the `PAYLOAD_LIST_ITEM` message. If trying to set a unsupported state on a payload (e.g. setting `Armed=true` on a Payload that only has <span class="title-ref">Powered</span> as a valid state), then the `MAV_CMD_PAYLOAD_SET_STATE` command should be rejected by the AutoPilot with a [COMMAND\_ACK](https://mavlink.io/en/messages/common.html#COMMAND_ACK) with a [MAV\_RESULT](https://mavlink.io/en/messages/common.html#MAV_RESULT) of `MAV_RESULT_DENIED`.

If the <span class="title-ref">id</span> field in the `MAV_CMD_PAYLOAD_SET_STATE` command sent by Autonomy is `0`, the AutoPilot must apply the state change to all registered Payloads for which it is a valid option. If a Payload ID that is not registered is specified in the command, it should be rejected by sending a
[COMMAND\_ACK](https://mavlink.io/en/messages/common.html#COMMAND_ACK) with a [MAV\_RESULT](https://mavlink.io/en/messages/common.html#MAV_RESULT) of `MAV_RESULT_DENIED`.

If AutoPilot successfully handles the command, a [COMMAND\_ACK](https://mavlink.io/en/messages/common.html#COMMAND_ACK) message should be sent back to Autonomy with a [MAV\_RESULT](https://mavlink.io/en/messages/common.html#MAV_RESULT) of `MAV_RESULT_ACCEPTED` followed by AutoPilot sending a `PAYLOAD_STATUS` message containing the updated Payload State.

payload_state_control

> Payload State Control Operation

When a Payload is added or removed to the AutoPilot's Payload list, a `PAYLOAD_CHANGE` message needs to be broadcast in order to inform all connected Companion Computers (including Autonomy) of the change, specifying the Payload ID that was affected and whether it has been added or removed.

If the update was about a new Payload, the Companion Computers may subsequently sent a
[MAV\_CMD\_REQUEST\_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE)
message to obtain details of the new Payload.

payload_list_change

# Autonomy Microservice

The Autonomy Microservice allows the `AutoPilot` to set Autonomy execution state, for control handover, and to monitor the state of Autonomy. The various `MAVLink Messages` used in the Autonomy Microservice are detailed in `Autonomy Message Table`.

| Message                   | ID      | Direction            | Description                      |
| ------------------------- | ------- | -------------------- | -------------------------------- |
| `MAV_CMD_AUTONOMY_DEMAND` | \#44000 | AutoPilot to Autonomy | Send Autonomy Demand to Autonomy. |
| `AUTONOMY_STATUS`         | \#44001 | Autonomy to AutoPilot | The state of Autonomy.            |

Messages used by the Autonomy Microservice

Central to the Autonomy Microservice is a state machine representing the autonomy system's execution state. numref:<span class="title-ref">autonomy\_state\_diagram</span> shows this state machine, where:

  - **Each state** is defined in the `AUTONOMY_STATE` enum (sent from
    Autonomy to the AutoPilot via the `AUTONOMY_STATUS` message).

  - **Each transition** is defined in the `AUTONOMY_DEMAND` enum (sent
    from the AutoPilot to Autonomy via the `MAV_CMD_AUTONOMY_DEMAND`
    message).

autonomy_state_diagram

The Autopilot **MUST** send a `Stop` or `Pause` command to Autonomy whenever it takes forcible control of the system.

In addition to providing an overall system state, the `AUTONOMY_STATUS` message also provides status bitmaps for a number of sub-components. Autonomy provides status for the following components:

  - `AUTONOMY_COMPONENT_CORE`
  - `AUTONOMY_COMPONENT_ATR`
  - `AUTONOMY_COMPONENT_MISSION`

Status bits for all other components are not used by Autonomy and set to zero.

##### Health Reporting

`Component Health Descriptions` describes in further detail what each
health flag reported by `AUTONOMY_STATUS` indicates.

| Autonomy Component           | Health Description                                                                                                                                                    |
| ---------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `AUTONOMY_COMPONENT_CORE`    | When health is reported as true for this component, it indicates that Autonomy has initialized successfully.                                                           |
| `AUTONOMY_COMPONENT_ATR`     | When health is reported as true for this component, it indicates that Autonomy's Embedded ATR component has initialized successfully and is ready for data processing. |
| `AUTONOMY_COMPONENT_MISSION` | When health is reported as true for this component, it indicates that Autonomy has a valid mission plan loaded and is ready for launch.                                |

Autonomy Component Health

##### Workflows

`autonomy_request_state` below shows the communication sequence
AutoPilot needs to follow to get the status of Autonomy.

Autonomy responds to the
[MAV\_CMD\_REQUEST\_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE)
command (see `Command Types
Table`) sent by the AutoPilot with a `AUTONOMY_STATUS` message
containing the presence, state and health of the components present on
the Autonomy system, and the current execution state of the system.

autonomy_request_status
> Request Autonomy State Process

`autonomy_send_autonomy_demand` below shows the communication sequence
when the AutoPilot sends a `MAV_CMD_AUTONOMY_DEMAND` containing an
`AUTONOMY_DEMAND` to Autonomy.

Autonomy responds with a
[COMMAND\_ACK](https://mavlink.io/en/messages/common.html#COMMAND_ACK)
and a `AUTONOMY_STATUS` message as stated in
`payload-microservice-errors`.

autonomy_send_autonomy_demand
> Send Autonomy Demand Process

#### Additional MAVLink Messages 

In addition to the messages already detailed in the various `MAVLink
Microservices`,

  - `MESSAGE_CONTACT_SNIPPET`
  - `MESSAGE_AUTONOMY_MISSION_STATUS`
  - `SQUAD_VEHICLE_STATE`

Also see the following enumerations:

  - `MAV_SYS_STATUS_SENSOR_EXTENDED`

#### External Communications

For integrations where all external communications (acoustic, RF etc.) for a vehicle are handled entirely by the `AutoPilot`:

  - Autonomy will forward all outgoing communication it wishes to send to
    the AutoPilot
  - The AutoPilot must send Autonomy incoming communications

| Message                           | ID    | Direction            | Description                                                   |
| --------------------------------- | ----- | -------------------- | ------------------------------------------------------------- |
| `MESSAGE_CONTACT_SNIPPET`         | 44100 | Autonomy to AutoPilot | Forward message containing details of an ATR Contact Snippet. |
| `MESSAGE_AUTONOMY_MISSION_STATUS` | 44101 | Bidirectional        | Forward message containing the status of the Autonomy Mission  |

Autonomy will provide the following metadata to define how the AutoPilot should handle the communication:

  - <span class="title-ref">destination\_address</span> - The ID of the
    external target of the communication, or `0` for broadcast
  - <span class="title-ref">message\_length</span> - The length of the
    communication to be sent (32 bytes by default)
  - <span class="title-ref">snippet\_confidence</span> - Only used by
    `MESSAGE_CONTACT_SNIPPET`, to be used by the AutoPilot to sort the
    order of the stored snippets

AutoPilot must provide the following metadata when sending a `MESSAGE_AUTONOMY_MISSION_STATUS` to Autonomy:

  - <span class="title-ref">source\_vehicle\_id</span> - The ID of the
    external source of the communication

**ALL** `MESSAGE_CONTACT_SNIPPET` messages received by the AutoPilot should be sent in the order they were received; while only the most recent `MESSAGE_AUTONOMY_MISSION_STATUS` should be sent.

It is expected that the AutoPilot will forward any incoming Mission Status communications to Autonomy using the `MESSAGE_AUTONOMY_MISSION_STATUS` message.

If sending outgoing communications over a limited medium, such as acoustic, it is expected that the AutoPilot uses the type of outgoing communication to determine the sending priority.

If the AutoPilot receives telemetry, vehicle state or fault updates from other vehicles in the squad via acoustic of other communications, these must be forwarded to Autonomy using the `SQUAD_VEHICLE_STATE` message.

# APPENDIX 1

#### AUTONOMY\_DEMAND

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) Demands
that can be made to an autonomy system.

| Value | Field Name                         | Description                                  |
| ----- | ---------------------------------- | -------------------------------------------- |
| 0     | AUTONOMY\_DEMAND\_DEFAULT          | Default value, ignored.                      |
| 1     | AUTONOMY\_DEMAND\_START            | Start Autonomy System.                       |
| 2     | AUTONOMY\_DEMAND\_STOP             | Stop Autonomy System.                        |
| 3     | AUTONOMY\_DEMAND\_GO\_TO\_RECOVERY | Abort Autonomy System to the Recovery Point. |
| 4     | AUTONOMY\_DEMAND\_PAUSE            | Pause Autonomy System.                       |
| 5     | AUTONOMY\_DEMAND\_SUSPEND          | Suspend Autonomy System Control Temporarily. |
| 6     | AUTONOMY\_DEMAND\_RESUME           | Resume Autonomy System.                      |

#### AUTONOMY\_COMPONENT

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) These
encode the components that can be present in an autonomy system.

| Value | Field Name                   | Description                       |
| ----- | ---------------------------- | --------------------------------- |
| 1     | AUTONOMY\_COMPONENT\_CORE    | 0x01 Autonomy System Core         |
| 2     | AUTONOMY\_COMPONENT\_ATR     | 0x02 Automated Target Recognition |
| 4     | AUTONOMY\_COMPONENT\_SAS     | 0x04 Synthetic Aperture Sonar     |
| 8     | AUTONOMY\_COMPONENT\_MBES    | 0x08 Multi Beam Echo Sounder      |
| 16    | AUTONOMY\_COMPONENT\_FLS     | 0x10 Forward Look Sonar           |
| 32    | AUTONOMY\_COMPONENT\_CAMERA  | 0x20 Camera                       |
| 64    | AUTONOMY\_COMPONENT\_ACOMMS  | 0x40 Acomms Modem                 |
| 128   | AUTONOMY\_COMPONENT\_MISSION | 0x80 Autonomy Mission             |

#### AUTONOMY\_STATE

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) Describes
the states an autonomy system can be in.

| Value | Field Name                  | Description                                       |
| ----- | --------------------------- | ------------------------------------------------- |
| 0     | AUTONOMY\_STATE\_DEFAULT    | Default state, ignored.                           |
| 1     | AUTONOMY\_STATE\_IDLE       | System has not been started.                      |
| 2     | AUTONOMY\_STATE\_EXECUTING  | Autonomy System Executing and in control.         |
| 3     | AUTONOMY\_STATE\_PAUSED     | Autonomy System execution paused.                 |
| 4     | AUTONOMY\_STATE\_SUSPENDED  | Autonomy System Control Suspended Temporarily.    |
| 5     | AUTONOMY\_STATE\_COMPLETED  | Autonomy System execution completed successfully. |
| 6     | AUTONOMY\_STATE\_STOPPED    | Autonomy System execution stopped early.          |
| 7     | AUTONOMY\_STATE\_RECOVERING | Autonomy System heading to recovery position.     |

#### MAV\_SYS\_STATUS\_SENSOR\_EXTENDED

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) Extends the
common
[MAV\_SYS\_STATUS\_SENSOR](https://mavlink.io/en/messages/common.html#MAV_SYS_STATUS_SENSOR)
enum.

| Value | Field Name                         | Description                                             |
| ----- | ---------------------------------- | ------------------------------------------------------- |
| 1     | MAV\_SYS\_STATUS\_RECOVERY\_SYSTEM | 0x01 Recovery system (parachute, balloon, retracts etc) |
| 2     | MAV\_SYS\_STATUS\_ACOUSTIC\_MODEM  | 0x02 Acoustic Modem                                     |
| 4     | MAV\_SYS\_STATUS\_SAS              | 0x04 Single Aperture Sonar                              |
| 8     | MAV\_SYS\_STATUS\_MBES             | 0x08 Multi Beam Echo Sounder                            |
| 16    | MAV\_SYS\_STATUS\_FLS              | 0x10 Forward Look Sonar                                 |

#### PAYLOAD\_TYPE

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) Defines the
various types of `Payloads<Payload>` that can be available on the
MAVLink system.

| Value | Field Name                     | Description                         |
| ----- | ------------------------------ | ----------------------------------- |
| 0     | PAYLOAD\_TYPE\_CAMERA          | Generic Camera Payload              |
| 1     | PAYLOAD\_TYPE\_LIDAR           | Light Detection and Ranging Payload |
| 2     | PAYLOAD\_TYPE\_WINCH           | Winch Payload                       |
| 3     | PAYLOAD\_TYPE\_ACOUSTIC\_MODEM | Acoustic Modem Payload              |
| 4     | PAYLOAD\_TYPE\_SAS             | Single Aperture Sonar Payload       |
| 5     | PAYLOAD\_TYPE\_MBES            | Multi Beam Echo Sounder Payload     |
| 6     | PAYLOAD\_TYPE\_FLS             | Forward Look Sonar Payload          |
| 101   | PAYLOAD\_TYPE\_GENERIC\_1      | Generic Payload 1                   |
| 102   | PAYLOAD\_TYPE\_GENERIC\_2      | Generic Payload 2                   |
| 103   | PAYLOAD\_TYPE\_GENERIC\_3      | Generic Payload 3                   |
| 104   | PAYLOAD\_TYPE\_GENERIC\_4      | Generic Payload 4                   |
| 105   | PAYLOAD\_TYPE\_GENERIC\_5      | Generic Payload 5                   |

#### PAYLOAD\_HEALTH

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) These
define the operational health a `Payload` can be in.

| Value | Field Name               | Description                   |
| ----- | ------------------------ | ----------------------------- |
| 1     | PAYLOAD\_HEALTH\_GENERIC | 0x01 Payload Health - Generic |

#### PAYLOAD\_STATE

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) These
define the states a `Payload` can be in.

| Value | Field Name              | Description                  |
| ----- | ----------------------- | ---------------------------- |
| 1     | PAYLOAD\_STATE\_POWERED | 0x01 Payload State - Powered |
| 2     | PAYLOAD\_STATE\_ARMED   | 0x02 Payload State - Armed   |
| 4     | PAYLOAD\_STATE\_ACTIVE  | 0x04 Payload State - Active  |

#### PAYLOAD\_LIST\_RESULT

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) These
define the results of a payload discovery operation.

| Value | Field Name                  | Description                                  |
| ----- | --------------------------- | -------------------------------------------- |
| 0     | PAYLOAD\_LIST\_ACCEPTED     | Payload List downloaded OK                   |
| 1     | PAYLOAD\_LIST\_ERROR        | Generic error with downloading Payload List  |
| 2     | PAYLOAD\_LIST\_REPEATED\_ID | Multiple entries for same Payload ID in List |

#### VEHICLE\_MODE

[\[Enum\]](https://mavlink.io/en/messages/common.html#enums) Vehicle
Mode.

| Value | Field Name                       | Description                                    |
| ----- | -------------------------------- | ---------------------------------------------- |
| 0     | VEHICLE\_MODE\_DEFAULT           | Default state, ignored.                        |
| 1     | VEHICLE\_MODE\_MISSION\_FAULT    | There is a fault with the current mission.     |
| 2     | VEHICLE\_MODE\_MISSION\_ABORT    | The current mission has been aborted.          |
| 3     | VEHICLE\_MODE\_BASELINE\_MISSION | The vehicle is executing its baseline mission. |
| 4     | VEHICLE\_MODE\_REMOTE\_MISSION   | THe vehicle is executing a remote mission.     |

### MAVLink Commands ([MAV\_CMD](https://mavlink.io/en/messages/common.html#mav_commands))

<div class="note">

<div class="admonition-title">

Note

</div>

MAVLink commands
([MAV\_CMD](https://mavlink.io/en/messages/common.html#mav_commands))
and messages are different\! These commands define the values of up to 7
parameters that are packaged INSIDE specific messages used in the
Mission Protocol and Command Protocol. Use commands for actions in
missions or if you need acknowledgment and/or retry logic from a
request. Otherwise use messages.

</div>

#### MAV\_CMD\_AUTONOMY\_DEMAND (\#44000)

[\[Command\]](https://mavlink.io/en/messages/common.html#mav_commands)
Send autonomy demands to a `Companion Computer`.

| Parameter (:Label) | Units | Values           | Description     |
| ------------------ | ----- | ---------------- | --------------- |
| 1: Autonomy Demand | N/A   | AUTONOMY\_DEMAND | Autonomy Demand |
| 2                  | N/A   | N/A              | Empty           |
| 3                  | N/A   | N/A              | Empty           |
| 4                  | N/A   | N/A              | Empty           |
| 5                  | N/A   | N/A              | Empty           |
| 6                  | N/A   | N/A              | Empty           |
| 7                  | N/A   | N/A              | Empty           |

#### MAV\_CMD\_PAYLOAD\_SET\_STATE (\#44002)

[\[Command\]](https://mavlink.io/en/messages/common.html#mav_commands)
Set the states for the payload with the specified ID.

Command will be rejected if trying to set a state not supported by the
payload.

<div class="note">

<div class="admonition-title">

Note

</div>

This command should always be sent as
<span class="title-ref">COMMAND\_INT</span> which has:

  - <span class="title-ref">Param 5/x</span> as int32\_t rather than
    float

</div>

| Parameter (:Label) | Units | Values                     | Description                                    |
| ------------------ | ----- | -------------------------- | ---------------------------------------------- |
| 1: Payload ID      | N/A   | min: 0 max:999 increment:1 | Payload ID, 0 to apply to all Payloads in List |
| 2                  | N/A   | N/A                        | Empty                                          |
| 3                  | N/A   | N/A                        | Empty                                          |
| 4                  | N/A   | N/A                        | Empty                                          |
| 5: components      | N/A   | PAYLOAD\_STATE             | Bitmask of enabled Payload states              |
| 6                  | N/A   | N/A                        | Empty                                          |
| 7                  | N/A   | N/A                        | Empty                                          |

#### AUTONOMY\_STATUS (\#44001)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages) Used
to specify the state of an autonomy system.

<table>
<caption>AUTONOMY_STATUS</caption>
<thead>
<tr class="header">
<th>Field Name</th>
<th>Default Value</th>
<th>Type</th>
<th>Units</th>
<th>Description</th>
</tr>
</thead>
<tbody>
<tr class="odd">
<td>target_system</td>
<td>N/A</td>
<td>uint8_t</td>
<td>N/A</td>
<td>System ID</td>
</tr>
<tr class="even">
<td>target_component</td>
<td>N/A</td>
<td>uint8_t</td>
<td>N/A</td>
<td>Component ID</td>
</tr>
<tr class="odd">
<td>autonomy_components_present</td>
<td>N/A</td>
<td>uint32_t</td>
<td>MAV_AUTONOMY_COMPONENT</td>
<td>A 32 bit bitmap corresponding to the <code class="interpreted-text" data-role="ref">AUTONOMY_COMPONENT&lt;AUTONOMY_COMPONENT&gt;</code> enum, showing which autonomy components are present on the <code class="interpreted-text" data-role="term">MAVLink</code> system:
<ul>
<li>Value of <code>0</code> if not present</li>
<li>Value of <code>1</code> if present</li>
</ul></td>
</tr>
<tr class="even">
<td>autonomy_components_enabled</td>
<td>N/A</td>
<td>uint32_t</td>
<td>MAV_AUTONOMY_COMPONENT</td>
<td>A 32 bit bitmap corresponding to the <code class="interpreted-text" data-role="ref">AUTONOMY_COMPONENT&lt;AUTONOMY_COMPONENT&gt;</code> enum, showing which autonomy components are enabled:
<ul>
<li>Value of <code>0</code> if not enabled</li>
<li>Value of <code>1</code> if enabled</li>
</ul></td>
</tr>
<tr class="odd">
<td>autonomy_components_health</td>
<td>N/A</td>
<td>uint32_t</td>
<td>MAV_AUTONOMY_COMPONENT</td>
<td>A 32 bit bitmap corresponding to the <code class="interpreted-text" data-role="ref">AUTONOMY_COMPONENT&lt;AUTONOMY_COMPONENT&gt;</code> enum, showing which autonomy components have an error (or are operational):
<ul>
<li>Value of <code>0</code> if it has an error</li>
<li>Value of <code>1</code> if it is healthy</li>
</ul></td>
</tr>
<tr class="even">
<td>system_state</td>
<td>N/A</td>
<td>uint8_t</td>
<td><code class="interpreted-text" data-role="ref">AUTONOMY_STATE&lt;AUTONOMY_COMPONENT&gt;</code></td>
<td>The current execution state of the Autonomy system.</td>
</tr>
</tbody>
</table>

#### MESSAGE\_CONTACT\_SNIPPET (\#44100)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Contact snippet message to be sent to specified destination with message
priority, as well as confidence of contact snippet.

| Field Name           | Default Value | Type            | Units | Description                                                                |
| -------------------- | ------------- | --------------- | ----- | -------------------------------------------------------------------------- |
| target\_system       | N/A           | uint8\_t        | N/A   | System ID                                                                  |
| target\_component    | N/A           | uint8\_t        | N/A   | Component ID                                                               |
| destination\_address | N/A           | uint8\_t        | N/A   | Acomms message destination, 0 for broadcast.                               |
| snippet\_confidence  | N/A           | uint8\_t        | %     | Contact snippet confidence (0 to 100).                                     |
| snippet\_id          | N/A           | uint16\_t       | N/A   | Contact snippet identifier. Unique for each snippet across all contacts    |
| source\_vehicle\_id  | N/A           | uint8\_t        | N/A   | Source Vehicle ID. UINT8\_MAX: field not provided.                         |
| message\_length      | N/A           | uint8\_t        | N/A   | Length of the data transported in message                                  |
| message              | N/A           | uint8\_t\[128\] | N/A   | Variable length message. The message length is defined by message\_length. |

#### MESSAGE\_AUTONOMY\_MISSION\_STATUS (\#44101)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Autonomy System Mission Status message to be sent to specified
destination.

| Field Name           | Default Value | Type            | Units | Description                                                                |
| -------------------- | ------------- | --------------- | ----- | -------------------------------------------------------------------------- |
| target\_system       | N/A           | uint8\_t        | N/A   | System ID                                                                  |
| target\_component    | N/A           | uint8\_t        | N/A   | Component ID                                                               |
| destination\_address | N/A           | uint8\_t        | N/A   | Acomms message destination, 0 for broadcast.                               |
| source\_vehicle\_id  | N/A           | uint8\_t        | N/A   | Source Vehicle ID. UINT8\_MAX: field not provided.                         |
| message\_length      | N/A           | uint8\_t        | N/A   | Length of the data transported in message                                  |
| message              | N/A           | uint8\_t\[128\] | N/A   | Variable length message. The message length is defined by message\_length. |

#### SQUAD\_VEHICLE\_STATE (\#44102)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Vehicle state data recieved from another vehicle in the squad.

| Field Name          | Default Value | Type     | Units                  | Description                                                                                         |
| ------------------- | ------------- | -------- | ---------------------- | --------------------------------------------------------------------------------------------------- |
| target\_system      | N/A           | uint8\_t | N/A                    | System ID                                                                                           |
| target\_component   | N/A           | uint8\_t | N/A                    | Component ID                                                                                        |
| source\_vehicle\_id | N/A           | uint8\_t | N/A                    | Source Vehicle ID.                                                                                  |
| vehicle\_mode       | N/A           | uint8\_t | N/A                    | Current vehicle mode. See `VEHICLE_MODE` for allowed values.                                        |
| lat                 | N/A           | int32\_t | Decimal Degrees \* 1e7 | Latitude, expressed                                                                                 |
| lon                 | N/A           | int32\_t | Decimal Degrees \* 1e7 | Longitude, expressed                                                                                |
| depth               | N/A           | float    | m                      | Current depth in meters positive down from the surface.                                             |
| altitude            | N/A           | float    | m                      | Current altitude from the terrain                                                                   |
| heading             | N/A           | float    | degrees                | Heading in degrees clockwise from North.                                                            |
| velocity            | N/A           | float    | m/s                    | Speed in meters per second.                                                                         |
| battery\_state      | N/A           | uint8\_t | %                      | Remaining battery energy. Values: \[0-100\], -1: AutoPilot does not estimate the remaining battery. |
| vehicle\_faults     | N/A           | int32\_t | N/A                    | Vehicle faults - defined specific to vehicle.                                                       |

#### PAYLOAD\_REQUEST\_LIST (\#44200)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages) Used
to request the list of registered Payloads from the AutoPilot.

| Field Name        | Default Value | Type     | Units | Description  |
| ----------------- | ------------- | -------- | ----- | ------------ |
| target\_system    | N/A           | uint8\_t | N/A   | System ID    |
| target\_component | N/A           | uint8\_t | N/A   | Component ID |

#### PAYLOAD\_COUNT (\#44201)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages) This
message is emitted as response to `PAYLOAD_REQUEST_LIST`. The `Companion
Computer` can then request the individual payload item based on the
knowledge of the total number of `Payloads<Payload>`.

| Field Name        | Default Value | Type      | Units | Description                   |
| ----------------- | ------------- | --------- | ----- | ----------------------------- |
| target\_system    | N/A           | uint8\_t  | N/A   | System ID                     |
| target\_component | N/A           | uint8\_t  | N/A   | Component ID                  |
| count             | N/A           | uint16\_t | N/A   | Number of registered payloads |

#### PAYLOAD\_LIST\_ITEM\_REQUEST (\#44202)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Request the status of the Payload with the specified Payload ID.

| Field Name              | Default Value | Type     | Units | Description           |
| ----------------------- | ------------- | -------- | ----- | --------------------- |
| target\_system          | N/A           | uint8\_t | N/A   | System ID             |
| target\_component       | N/A           | uint8\_t | N/A   | Component ID          |
| payload\_list\_position | N/A           | uint8\_t | N/A   | Payload List position |

#### PAYLOAD\_LIST\_ITEM (\#44203)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Description of a single `Payload`.

| Field Name        | Default Value | Type                       | Units | Description       |
| ----------------- | ------------- | -------------------------- | ----- | ----------------- |
| target\_system    | N/A           | uint8\_t                   | N/A   | System ID         |
| target\_component | N/A           | uint8\_t                   | N/A   | Component ID      |
| payload\_id       | N/A           | uint8\_t                   | N/A   | Payload ID, 1 - N |
| payload\_name     | N/A           | char\[16\]                 | N/A   | Payload Name      |
| payload\_type     | N/A           | PAYLOAD\_TYPE              | N/A   | Payload Type      |
| valid\_states     | N/A           | uint16\_t - PAYLOAD\_STATE | N/A   | Payload Type      |

#### PAYLOAD\_LIST\_ACK (\#44204)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Acknowledgment Message from a `Companion Computer` upon receiving all
<span class="title-ref">PAYLOAD\_LIST\_ITEM</span> Messages.

| Field Name        | Default Value | Type                  | Units | Description         |
| ----------------- | ------------- | --------------------- | ----- | ------------------- |
| target\_system    | N/A           | uint8\_t              | N/A   | System ID           |
| target\_component | N/A           | uint8\_t              | N/A   | Component ID        |
| result            | N/A           | PAYLOAD\_LIST\_RESULT | N/A   | Payload List result |

#### PAYLOAD\_STATUS (\#44206)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages) This
message is emitted as response to
<span class="title-ref">MAV\_CMD\_REQUEST\_MESSAGE</span> messages with
<span class="title-ref">Param 1</span> specifying the message ID
(`44206`) and <span class="title-ref">Param 2</span> specifying the
payload ID. This message contains the details (ID and Type) and health
and status (as bitmasks) of the specified payload.

| Field Name        | Default Value | Type          | Units           | Description              |
| ----------------- | ------------- | ------------- | --------------- | ------------------------ |
| target\_system    | N/A           | uint8\_t      | N/A             | System ID                |
| target\_component | N/A           | uint8\_t      | N/A             | Component ID             |
| payload\_id       | N/A           | uint8\_t      | N/A             | Payload ID, 1 - N        |
| payload\_type     | N/A           | PAYLOAD\_TYPE | N/A             | Payload Type             |
| payload\_health   | N/A           | uint16\_t     | PAYLOAD\_HEALTH | Payload Health (bitmask) |
| payload\_state    | N/A           | uint16\_t     | PAYLOAD\_STATE  | Payload State (bitmask)  |

#### PAYLOAD\_CHANGE (\#44207)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Notification that a `Payload` has been added or removed from the MAVLink
system.

| Field Name        | Default Value | Type     | Units                | Description       |
| ----------------- | ------------- | -------- | -------------------- | ----------------- |
| target\_system    | N/A           | uint8\_t | N/A                  | System ID         |
| target\_component | N/A           | uint8\_t | N/A                  | Component ID      |
| payload\_change   | N/A           | uint8\_t | 1: Added, 0: Removed | Payload ID change |
| payload\_id       | N/A           | uint8\_t | N/A                  | Payload ID (1-N)  |

#### WATER\_CURRENT (\#44300)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages) Water
current values.

<div class="note">

<div class="admonition-title">

Note

</div>

Calculated Water current = (Vehicle Velocity - Measured Current).

</div>

| Field Name             | Default Value | Type  | Units | Description                                     |
| ---------------------- | ------------- | ----- | ----- | ----------------------------------------------- |
| measured\_current\_x   | N/A           | float | m/s   | Raw Measured water current in X (NED) direction |
| measured\_current\_y   | N/A           | float | m/s   | Raw Measured water current in Y (NED) direction |
| measured\_current\_z   | N/A           | float | m/s   | Raw Measured water current in Z (NED) direction |
| calculated\_current\_x | N/A           | float | m/s   | Calculated Water current in X (NED)             |
| calculated\_current\_y | N/A           | float | m/s   | Calculated Water current in Y (NED)             |
| calculated\_current\_z | N/A           | float | m/s   | Calculated Water current in Z (NED)             |
| distance               | N/A           | float | m     | Measurement distance from the sensor            |

#### DVL (\#44301)

[\[Message\]](https://mavlink.io/en/messages/common.html#messages)
Doppler Velocity Log Sensor. Used to provide vehicle velocity.

| Field Name           | Default Value | Type        | Units | Description                                                                     |
| -------------------- | ------------- | ----------- | ----- | ------------------------------------------------------------------------------- |
| has\_lock            | N/A           | uint8\_t    | N/A   | Boolean indicating if the DVL has bottom lock to the floor (true: 1, false: 0). |
| altitude\_water      | N/A           | float\[16\] | m     | Vehicle altitude to the floor of body of water for each beam                    |
| vehicle\_velocity\_x | N/A           | float       | m/s   | Vehicle velocity in X (NED)                                                     |
| vehicle\_velocity\_y | N/A           | float       | m/s   | Vehicle velocity in Y (NED)                                                     |
| vehicle\_velocity\_z | N/A           | float       | m/s   | Vehicle velocity in Z (NED)                                                     |

# APPENDIX 2

## Autonomy Protocol

The autonomy microservice is used by `AutoPilots<AutoPilot>` to get
information about `Companion Computers<Companion Computer>` that provide
autonomy. It also allows AutoPilots to configure the state of autonomy
systems.

Full definitions can be found in
[marine.xml](https://github.com/mavlinkmarine/mavlink/blob/v22.08/message_definitions/v1.0/marine.xml).

## Message/Enum Summary

| Message                   | Description                       |
| ------------------------- | --------------------------------- |
| `MAV_CMD_AUTONOMY_DEMAND` | Send a demand to autonomy system. |
| `AUTONOMY_STATUS`         | The state of the autonomy system. |

Autonomy Protocol Message Summary

| Message              | Description                                                            |
| -------------------- | ---------------------------------------------------------------------- |
| `AUTONOMY_DEMAND`    | Demands to be made to the autonomy system.                             |
| `AUTONOMY_COMPONENT` | These encode the components that can be present in an autonomy system. |
| `AUTONOMY_STATE`     | Current state of the Autonomy System.                                  |

Autonomy Protocol Enum Summary

### Timeouts and Retires

A timeout should be set for all messages that require a response. If the
expected response is not received before the timeout then the message
must be resent. If no response is received after a number of retries
then the client must cancel the operation and return to a clean state as
it was before the initial message was sent.

The recommended timeout values before resending, and the number of
retries are:

  - Timeout (default): `1500 ms`
  - Retries (max): `5`

### Errors/Completion

On acceptance of a `MAV_CMD_AUTONOMY_DEMAND`, the autonomy component
must respond with a <span class="title-ref">COMMAND\_ACK</span> message
with a `MAV_RESULT_ACCEPTED` <span class="title-ref">MAV\_RESULT</span>
and then it must publish a `AUTONOMY_STATUS` message containing the
updated execution state.

If a command is received by the autonomy component when it is not in a
state to carry it out, the resulting
<span class="title-ref">COMMAND\_ACK</span> must contain the
`MAV_RESULT_TEMPORARILY_REJECTED`
<span class="title-ref">MAV\_RESULT</span>.

If a command is received by the autonomy component when would constitute
an invalid state transition (for example if the autonomy sytem does not
allow a `AUTONOMY_DEMAND_START` when the current execution mode is
`AUTONOMY_STATE_EXECUTING`), then the autonomy system should continue in
its current state, rejecting the new one, and send a resulting
<span class="title-ref">COMMAND\_ACK</span> containing the
`MAV_RESULT_DENIED` <span class="title-ref">MAV\_RESULT</span>.

# APPENDIX 3

## Payload Protocol

The payload microservice is used by `Companion Computers<Companion
Computer>` to get information about `Payloads<Payload>` from the
AutoPilot. It also allows Companion Computers to configure Payloads via
the AutoPilot.

Full definitions can be found in
[marine.xml](https://github.com/mavlinkmarine/mavlink/blob/v22.08/message_definitions/v1.0/marine.xml).

## Message/Enum Summary

<table>
<caption>Payload Protocol Message Summary</caption>
<thead>
<tr class="header">
<th>Message</th>
<th>Description</th>
</tr>
</thead>
<tbody>
<tr class="odd">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_REQUEST_LIST</code></td>
<td>Used to request the list of registered Payloads from the AutoPilot.</td>
</tr>
<tr class="even">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_COUNT</code></td>
<td>Emitted as response to <span class="title-ref">PAYLOAD_REQUEST_LIST</span></td>
</tr>
<tr class="odd">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_LIST_ITEM_REQUEST</code></td>
<td>Request the status of the Payload with the specified Payload ID.</td>
</tr>
<tr class="even">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_LIST_ITEM</code></td>
<td>This message is emitted as response to <code class="interpreted-text" data-role="ref">PAYLOAD_LIST_ITEM_REQUEST</code>.</td>
</tr>
<tr class="odd">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_LIST_ACK</code></td>
<td>Acknowledgment Message from a Companion Computer upon receiving all <code class="interpreted-text" data-role="ref">PAYLOAD_LIST_ITEM</code> Messages.</td>
</tr>
<tr class="even">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_STATUS</code></td>
<td><p>Emitted as response to <span class="title-ref">MAV_CMD_REQUEST_MESSAGE</span> with <span class="title-ref">Param 1</span> set to <code>44206</code>.</p>
<p>It contains the ID, Type, Health and State of the Payload specified in <span class="title-ref">Param 2</span> of the <span class="title-ref">MAV_CMD_REQUEST_MESSAGE</span> Message.</p></td>
</tr>
<tr class="odd">
<td><code class="interpreted-text" data-role="ref">PAYLOAD_CHANGE</code></td>
<td>Notification that a Payload has been added or removed from the MAVLink system.</td>
</tr>
<tr class="even">
<td><code class="interpreted-text" data-role="ref">MAV_CMD_PAYLOAD_SET_STATE</code></td>
<td>Put the Payload with the specified ID into the given state.</td>
</tr>
</tbody>
</table>

| Message               | Description                                              |
| --------------------- | -------------------------------------------------------- |
| `PAYLOAD_TYPE`        | List of supported payload types and generic placeholders |
| `PAYLOAD_HEALTH`      | Flags to report health of a registered payload           |
| `PAYLOAD_STATE`       | Flags to report state of a registered payload            |
| `PAYLOAD_LIST_RESULT` | Results of payload discovery operation                   |

Payload Protocol Enum Summary

### Timeouts and Retires

A timeout should be set for all messages that require a response. If the
expected response is not received before the timeout then the message
must be resent. If no response is received after a number of retries
then the client must cancel the operation and return to a clean state as
it was before the initial message was sent.

The recommended timeout values before resending, and the number of
retries are:

  - Timeout (default): `1500 ms`
  - Timeout (payload items): `250 ms`
  - Retries (max): `5`

### Errors/Completion

On successful completion of a payload list exchange the participant that
made the initial `PAYLOAD_REQUEST_LIST`, will send a `PAYLOAD_LIST_ACK`
messages with a <span class="title-ref">result</span> value of
`PAYLOAD_LIST_ACCEPTED` (see the `PAYLOAD_LIST_RESULT` enum).

Any other value in the <span class="title-ref">result</span> field of
the `PAYLOAD_LIST_ACK` message indicates an error. Both participants in
the exchange must reset themselves to the state they were in before the
initial `PAYLOAD_REQUEST_LIST` message, and the initiator must ignore
any information it received during the erronious list exchange.

### DIFF
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -52,6 +52,9 @@
       <entry value="64" name="AUTONOMY_COMPONENT_ACOMMS">
         <description>0x40 Acomms Modem</description>
       </entry>
+      <entry value="128" name="AUTONOMY_COMPONENT_MISSION">
+        <description>0x80 Autonomy Mission</description>
+      </entry>
     </enum>
     <enum name="AUTONOMY_STATE">
       <description>Current state of the Autonomy System.</description>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -192,7 +192,7 @@
         <param index="7">Empty.</param>
       </entry> -->
       <entry value="44002" name="MAV_CMD_PAYLOAD_SET_STATE" hasLocation="false" isDestination="false">
-        <description>Set the states for the payload with the specified ID. Command will be rejected if trying to set a state not supported by the payload.</description>
+        <description>Set the states for the payload with the specified ID. Command will be rejected if trying to set a state not supported by the payload. Note that this command should always be sent as COMMAND_INT which has param5/x as int32_t rather than float.</description>
         <param index="1" label="Payload ID" minValue="0" maxValue="999" increment="1">Payload ID, 0 to apply to all Payloads in List.</param>
         <param index="2">Empty.</param>
         <param index="3">Empty.</param>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -259,6 +259,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="destination_address">Acomms message destination, 0 for broadcast.</field>
+      <field type="uint8_t" name="source_vehicle_id">Source Vehicle ID.</field>
       <field type="uint8_t" name="snippet_confidence" units="%" >Contact snippet confidence (0 to 100).</field>
       <field type="uint16_t" name="snippet_id">Contact snippet identifier. Unique for each snippet.</field>
       <field type="uint8_t" name="message_length">Length of the data transported in message.</field>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -299,13 +299,14 @@
       <field type="uint8_t" name="result" enum="PAYLOAD_LIST_RESULT">Payload List result.</field>
     </message>
     <message id="44205" name="PAYLOAD_STATUS_REQUEST">
-      <description>Request the status of the payload item with the specified payload ID. The response of the system to this message should be a PAYLOAD_STATUS message.</description>
+      <deprecated since="2022-05" replaced_by="MAV_CMD_REQUEST_MESSAGE"/>
+      <description>Request the status of the payload item with the specified payload ID. The response of the system to this message should be a PAYLOAD_STATUS message. Note: this message has been deprecated in favor of using MAV_CMD_REQUEST_MESSAGE with param1 specifiying the message ID and param2 specifying the payload ID.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="payload_id">Payload ID.</field>
     </message>
     <message id="44206" name="PAYLOAD_STATUS">
-      <description>This message is emitted as response to PAYLOAD_STATUS_REQUEST, and contains the details (ID and Type) and health and status (as bitmasks) of the specified payload.</description>
+      <description>This message is emitted as response to MAV_CMD_REQUEST_MESSAGE with param1 specifying the message ID (44206) andparam2 specifying the payload ID, and contains the details (ID and Type) and health and status (as bitmasks) of the specified payload.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="payload_id">Payload ID, 1 - N.</field>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -76,6 +76,9 @@
       <entry value="6" name="AUTONOMY_STATE_STOPPED">
         <description>Autonomy System execution stopped early.</description>
       </entry>
+      <entry value="7" name="AUTONOMY_STATE_RECOVERING">
+        <description>Autonomy System heading to recovery position.</description>
+      </entry>
     </enum>
     <enum name="PAYLOAD_TYPE">
       <description>Supported Payload types.</description>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -83,6 +83,24 @@
         <description>Autonomy System heading to recovery position.</description>
       </entry>
     </enum>
+    <enum name="VEHICLE_MODE">
+      <description>Current mode of a squad vehicle.</description>
+      <entry value="0" name="VEHICLE_MODE_DEFAULT">
+        <description>Default state, ignored.</description>
+      </entry>
+      <entry value="1" name="VEHICLE_MODE_MISSION_FAULT">
+        <description>There is a fault with the current mission.</description>
+      </entry>
+      <entry value="2" name="VEHICLE_MODE_MISSION_ABORT">
+        <description>The current mission has been aborted.</description>
+      </entry>
+      <entry value="3" name="VEHICLE_MODE_BASELINE_MISSION">
+        <description>The vehicle is executing its baseline mission.</description>
+      </entry>
+      <entry value="4" name="VEHICLE_MODE_REMOTE_MISSION">
+        <description>The vehicle is executing a remote mission.</description>
+      </entry>
+    </enum>
     <enum name="PAYLOAD_TYPE">
       <description>Supported Payload types.</description>
       <entry value="0" name="PAYLOAD_TYPE_CAMERA">
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -71,7 +71,7 @@
         <description>Autonomy System Control Suspended Temporarily.</description>
       </entry>
       <entry value="5" name="AUTONOMY_STATE_COMPLETED">
-        <description>Autonomy System execution completed succesfully.</description>
+        <description>Autonomy System execution completed successfully.</description>
       </entry>
       <entry value="6" name="AUTONOMY_STATE_STOPPED">
         <description>Autonomy System execution stopped early.</description>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -100,19 +100,19 @@
       <entry value="6" name="PAYLOAD_TYPE_FLS">
         <description>Forward Look Sonar Payload.</description>
       </entry>
-      <entry value="1001" name="PAYLOAD_TYPE_GENERIC_1">
+      <entry value="101" name="PAYLOAD_TYPE_GENERIC_1">
         <description>Generic Payload 1.</description>
       </entry>
-      <entry value="1002" name="PAYLOAD_TYPE_GENERIC_2">
+      <entry value="102" name="PAYLOAD_TYPE_GENERIC_2">
         <description>Generic Payload 2.</description>
       </entry>
-      <entry value="1003" name="PAYLOAD_TYPE_GENERIC_3">
+      <entry value="103" name="PAYLOAD_TYPE_GENERIC_3">
         <description>Generic Payload 3.</description>
       </entry>
-      <entry value="1004" name="PAYLOAD_TYPE_GENERIC_4">
+      <entry value="104" name="PAYLOAD_TYPE_GENERIC_4">
         <description>Generic Payload 4.</description>
       </entry>
-      <entry value="1005" name="PAYLOAD_TYPE_GENERIC_5">
+      <entry value="105" name="PAYLOAD_TYPE_GENERIC_5">
         <description>Generic Payload 5.</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -325,7 +325,7 @@
       <field type="float" name="calculated_current_x" units="m/s">Calculated Water current in X (NED)</field>
       <field type="float" name="calculated_current_y" units="m/s">Calculated Water current in Y (NED)</field>
       <field type="float" name="calculated_current_z" units="m/s">Calculated Water current in Z (NED)</field>
-      <field type="float" name="distance" units="cm">Measurement distance from the sensor</field>
+      <field type="float" name="distance" units="m">Measurement distance from the sensor</field>
     </message>
     <message id="44301" name="DVL">
       <description>Doppler Velocity Log Sensor. Used to provide vehicle velocity.</description>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -277,7 +277,7 @@
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint8_t" name="source_vehicle_id">Source Vehicle ID.</field>
-      <field type="uint8_t" name="vehicle_mode">Vehicle Mode.</field>
+      <field type="uint8_t" name="vehicle_mode" enum="VEHICLE_MODE">Vehicle Mode.</field>
       <field type="int32_t" name="lat" units="degE7">Latitude, expressed</field>
       <field type="int32_t" name="lon" units="degE7">Longitude, expressed</field>
       <field type="float" name="depth" units="m">Current depth in meters positive down from the surface.</field>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -260,6 +260,7 @@
       <field type="uint8_t" name="target_component">Component ID</field>
       <field type="uint32_t" name="destination_address">Acomms message destination, 0 for broadcast.</field>
       <field type="uint8_t" name="snippet_confidence" units="%" >Contact snippet confidence (0 to 100).</field>
+      <field type="uint16_t" name="snippet_id">Contact snippet identifier. Unique for each snippet.</field>
       <field type="uint8_t" name="message_length">Length of the data transported in message.</field>
       <field type="uint8_t[128]" name="message">Variable length message. The message length is defined by message_length.</field>
     </message>
--- a/message_definitions/v1.0/marine.xml
+++ b/message_definitions/v1.0/marine.xml
@@ -1,0 +1,339 @@
+<?xml version="1.0"?>
+<mavlink>
+  <!-- XML file for the MAVLink dialect  -->
+  <include>standard.xml</include>
+  <version>0</version>
+  <dialect>0</dialect>
+  <enums>
+    <enum name="AUTONOMY_DEMAND">
+      <description>Demands to be made to the autonomy system.</description>
+      <entry value="0" name="AUTONOMY_DEMAND_DEFAULT">
+        <description>Default value, ignored.</description>
+      </entry>
+      <entry value="1" name="AUTONOMY_DEMAND_START">
+        <description>Start Autonomy System.</description>
+      </entry>
+      <entry value="2" name="AUTONOMY_DEMAND_STOP">
+        <description>Stop Autonomy System.</description>
+      </entry>
+      <entry value="3" name="AUTONOMY_DEMAND_GO_TO_RECOVERY">
+        <description>Abort Autonomy System to the Recovery Point.</description>
+      </entry>
+      <entry value="4" name="AUTONOMY_DEMAND_PAUSE">
+        <description>Pause Autonomy System.</description>
+      </entry>
+      <entry value="5" name="AUTONOMY_DEMAND_SUSPEND">
+        <description>Suspend Autonomy System Control Temporarily.</description>
+      </entry>
+      <entry value="6" name="AUTONOMY_DEMAND_RESUME">
+        <description>Resume Autonomy System.</description>
+      </entry>
+    </enum>
+    <enum name="AUTONOMY_COMPONENT" bitmask="true">
+      <description>These encode the components that can be present in an autonomy system.</description>
+      <entry value="1" name="AUTONOMY_COMPONENT_CORE">
+        <description>0x01 Autonomy System Core</description>
+      </entry>
+      <entry value="2" name="AUTONOMY_COMPONENT_ATR">
+        <description>0x02 Automated Target Recognition</description>
+      </entry>
+      <entry value="4" name="AUTONOMY_COMPONENT_SAS">
+        <description>0x04 Synthetic Aperture Sonar</description>
+      </entry>
+      <entry value="8" name="AUTONOMY_COMPONENT_MBES">
+        <description>0x08 Multi Beam Echo Sounder</description>
+      </entry>
+      <entry value="16" name="AUTONOMY_COMPONENT_FLS">
+        <description>0x10 Forward Look Sonar</description>
+      </entry>
+      <entry value="32" name="AUTONOMY_COMPONENT_CAMERA">
+        <description>0x20 Camera</description>
+      </entry>
+      <entry value="64" name="AUTONOMY_COMPONENT_ACOMMS">
+        <description>0x40 Acomms Modem</description>
+      </entry>
+    </enum>
+    <enum name="AUTONOMY_STATE">
+      <description>Current state of the Autonomy System.</description>
+      <entry value="0" name="AUTONOMY_STATE_DEFAULT">
+        <description>Default state, ignored.</description>
+      </entry>
+      <entry value="1" name="AUTONOMY_STATE_IDLE">
+        <description>System has not been started.</description>
+      </entry>
+      <entry value="2" name="AUTONOMY_STATE_EXECUTING">
+        <description>Autonomy System Executing and in control.</description>
+      </entry>
+      <entry value="3" name="AUTONOMY_STATE_PAUSED">
+        <description>Autonomy System Paused.</description>
+      </entry>
+      <entry value="4" name="AUTONOMY_STATE_SUSPENDED">
+        <description>Autonomy System Control Suspended Temporarily.</description>
+      </entry>
+      <entry value="5" name="AUTONOMY_STATE_COMPLETED">
+        <description>Autonomy System execution completed succesfully.</description>
+      </entry>
+      <entry value="6" name="AUTONOMY_STATE_STOPPED">
+        <description>Autonomy System execution stopped early.</description>
+      </entry>
+    </enum>
+    <enum name="PAYLOAD_TYPE">
+      <description>Supported Payload types.</description>
+      <entry value="0" name="PAYLOAD_TYPE_CAMERA">
+        <description>Generic Camera Payload.</description>
+      </entry>
+      <entry value="1" name="PAYLOAD_TYPE_LIDAR">
+        <description>Light Detection and Ranging Payload.</description>
+      </entry>
+      <entry value="2" name="PAYLOAD_TYPE_WINCH">
+        <description>Winch Payload.</description>
+      </entry>
+      <entry value="3" name="PAYLOAD_TYPE_ACOUSTIC_MODEM">
+        <description>Acoustic Modem Payload.</description>
+      </entry>
+      <entry value="4" name="PAYLOAD_TYPE_SAS">
+        <description>Single Aperture Sonar Payload.</description>
+      </entry>
+      <entry value="5" name="PAYLOAD_TYPE_MBES">
+        <description>Multi Beam Echo Sounder Payload.</description>
+      </entry>
+      <entry value="6" name="PAYLOAD_TYPE_FLS">
+        <description>Forward Look Sonar Payload.</description>
+      </entry>
+      <entry value="1001" name="PAYLOAD_TYPE_GENERIC_1">
+        <description>Generic Payload 1.</description>
+      </entry>
+      <entry value="1002" name="PAYLOAD_TYPE_GENERIC_2">
+        <description>Generic Payload 2.</description>
+      </entry>
+      <entry value="1003" name="PAYLOAD_TYPE_GENERIC_3">
+        <description>Generic Payload 3.</description>
+      </entry>
+      <entry value="1004" name="PAYLOAD_TYPE_GENERIC_4">
+        <description>Generic Payload 4.</description>
+      </entry>
+      <entry value="1005" name="PAYLOAD_TYPE_GENERIC_5">
+        <description>Generic Payload 5.</description>
+      </entry>
+    </enum>
+    <enum name="PAYLOAD_HEALTH">
+      <description>Flags to report health of a registered payload.</description>
+      <entry value="1" name="PAYLOAD_HEALTH_GENERIC">
+        <description>0x01 Payload Health - Generic.</description>
+      </entry>
+    </enum>
+    <enum name="PAYLOAD_STATE">
+      <description>Flags to report state of a registered payload.</description>
+      <entry value="1" name="PAYLOAD_STATE_POWERED">
+        <description>0x01 Payload State - Powered.</description>
+      </entry>
+      <entry value="2" name="PAYLOAD_STATE_ARMED">
+        <description>0x02 Payload State - Armed.</description>
+      </entry>
+      <entry value="4" name="PAYLOAD_STATE_ACTIVE">
+        <description>0x04 Payload State - Active.</description>
+      </entry>
+    </enum>
+    <enum name="PAYLOAD_LIST_RESULT">
+      <description>Results of payload discovery operation (in a PAYLOAD_LIST_ACK message).</description>
+      <entry value="0" name="PAYLOAD_LIST_ACCEPTED">
+        <description>Payload List downloaded OK</description>
+      </entry>
+      <entry value="1" name="PAYLOAD_LIST_ERROR">
+        <description>Generic error with downloading Payload List.</description>
+      </entry>
+      <entry value="2" name="PAYLOAD_LIST_REPEATED_ID">
+        <description>Multiple entries for same Payload ID in List.</description>
+      </entry>
+    </enum>
+    <!-- Additions to the MAV_SYS_STATUS_SENSOR_EXTENDED enum: -->
+    <enum name="MAV_SYS_STATUS_SENSOR_EXTENDED" bitmask="true">
+      <description>These encode the sensors whose status is sent as part of the SYS_STATUS message in the extended fields.</description>
+      <entry value="2" name="MAV_SYS_STATUS_ACOUSTIC_MODEM">
+        <description>0x02 Acoustic Modem</description>
+      </entry>
+      <entry value="4" name="MAV_SYS_STATUS_SAS">
+        <description>0x04 Single Aperture Sonar</description>
+      </entry>
+      <entry value="8" name="MAV_SYS_STATUS_MBES">
+        <description>0x08 Multi Beam Echo Sounder</description>
+      </entry>
+      <entry value="16" name="MAV_SYS_STATUS_FLS">
+        <description>0x10 Forward Look Sonar</description>
+      </entry>
+    </enum>
+    <!-- The MAV_CMD enum entries describe either: -->
+    <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
+    <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
+    <!-- ALL the entries in the MAV_CMD enum have a maximum of 7 parameters -->
+    <enum name="MAV_CMD">
+      <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data. NaN and INT32_MAX may be used in float/integer params (respectively) to indicate optional/default values (e.g. to use the component's current yaw or latitude rather than a specific value). See https://mavlink.io/en/guide/xml_schema.html#MAV_CMD for information about the structure of the MAV_CMD entries</description>
+      <entry value="44000" name="MAV_CMD_AUTONOMY_DEMAND" hasLocation="false" isDestination="false">
+        <description>Send Autonomy Demand to Autonomy System.</description>
+        <param index="1" label="Autonomy Demand"  enum="AUTONOMY_DEMAND">Autonomy Demand.</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4">Empty.</param>
+        <param index="5">Empty.</param>
+        <param index="6">Empty.</param>
+        <param index="7">Empty.</param>
+      </entry>
+      <!-- <entry value="44001" name="MAV_CMD_AUTONOMY_COMPONENT_ENABLE" hasLocation="false" isDestination="false">
+        <description>Enable the bitmask specified autonomy components (1 enable, 0 disable). Command will be rejected if trying to enable a component that is not present. </description>
+        <param index="1">Empty.</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4">Empty.</param>
+        <param index="5" label="components" enum="AUTONOMY_COMPONENT">Bitmask of enabled Autonomy Components.</param>
+        <param index="6">Empty.</param>
+        <param index="7">Empty.</param>
+      </entry> -->
+      <entry value="44002" name="MAV_CMD_PAYLOAD_SET_STATE" hasLocation="false" isDestination="false">
+        <description>Set the states for the payload with the specified ID. Command will be rejected if trying to set a state not supported by the payload.</description>
+        <param index="1" label="Payload ID" minValue="0" maxValue="999" increment="1">Payload ID, 0 to apply to all Payloads in List.</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4">Empty.</param>
+        <param index="5" label="components" enum="PAYLOAD_STATE">Bitmask of enabled Payload states.</param>
+        <param index="6">Empty.</param>
+        <param index="7">Empty.</param>
+      </entry>
+      <entry value="44010" name="MAV_CMD_NAV_TRACK_START" hasLocation="true" isDestination="true">
+        <description>Navigate between two points maintaining trackline between them</description>
+        <param index="1" label="Hold" units="s" minValue="0">Hold time</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4" label="Tolerance" units="m" minValue="0">Maximum distance from track centre-line before correction is required</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
+      </entry>
+      <entry value="44011" name="MAV_CMD_NAV_TRACK_END" hasLocation="true" isDestination="true">
+        <description>Navigate between two points maintaining trackline between them</description>
+        <param index="1" label="Hold" units="s" minValue="0">Hold time</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4">Empty.</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
+      </entry>
+    </enum>
+  </enums>
+  <messages>
+    <message id="44001" name="AUTONOMY_STATUS">
+      <description>The state of the autonomy system.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="autonomy_components_present" enum="AUTONOMY_COMPONENT" display="bitmask" print_format="0x%04x">Bitmap showing which autonomy components are present. Value of 0: not present. Value of 1: present.</field>
+      <field type="uint32_t" name="autonomy_components_enabled" enum="AUTONOMY_COMPONENT" display="bitmask" print_format="0x%04x">Bitmap showing which autonomy components are enabled:  Value of 0: not enabled. Value of 1: enabled.</field>
+      <field type="uint32_t" name="autonomy_components_health" enum="AUTONOMY_COMPONENT" display="bitmask" print_format="0x%04x">Bitmap showing which autonomy components have an error (or are operational). Value of 0: error. Value of 1: healthy.</field>
+      <field type="uint8_t" name="system_state" enum="AUTONOMY_STATE">The Execution state of the Autonomy System.</field>
+    </message>
+    <message id="44100" name="MESSAGE_CONTACT_SNIPPET">
+      <description>Contact snippet message to be sent to specified destination, as well as confidence of contact snippet.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint32_t" name="destination_address">Acomms message destination, 0 for broadcast.</field>
+      <field type="uint8_t" name="snippet_confidence" units="%" >Contact snippet confidence (0 to 100).</field>
+      <field type="uint8_t" name="message_length">Length of the data transported in message.</field>
+      <field type="uint8_t[128]" name="message">Variable length message. The message length is defined by message_length.</field>
+    </message>
+    <message id="44101" name="MESSAGE_AUTONOMY_MISSION_STATUS">
+      <description>Autonomy System Mission Status message to be sent to specified destination.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t" name="destination_address">Acomms message destination, 0 for broadcast.</field>
+      <field type="uint8_t" name="source_vehicle_id" invalid="UINT8_MAX">Source Vehicle ID. UINT8_MAX: field not provided.</field>
+      <field type="uint8_t" name="message_length">Length of the data transported in message.</field>
+      <field type="uint8_t[128]" name="message">Variable length message. The message length is defined by message_length.</field>
+    </message>
+    <message id="44102" name="SQUAD_VEHICLE_STATE">
+      <description>Vehicle state data recieved from another vehicle in the squad.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
+      <field type="uint8_t" name="source_vehicle_id">Source Vehicle ID.</field>
+      <field type="uint8_t" name="vehicle_mode">Vehicle Mode.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude, expressed</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude, expressed</field>
+      <field type="float" name="depth" units="m">Current depth in meters positive down from the surface.</field>
+      <field type="float" name="altitude" units="m">Current altitude from the terrain.</field>
+      <field type="float" name="heading" units="deg">Heading in degrees clockwise from North.</field>
+      <field type="float" name="velocity" units="m/s">Speed in meters per second.</field>
+      <field type="int8_t" name="battery_state" units="%" invalid="-1">Remaining battery energy. Values: [0-100], -1: autopilot does not estimate the remaining battery.</field>
+      <field type="uint32_t" name="vehicle_faults">Vehicle faults - defined specific to vehicle.</field>
+    </message>
+    <message id="44200" name="PAYLOAD_REQUEST_LIST">
+      <description>Request the list of registered payloads from the system/component.</description>
+      <field type="uint8_t" name="target_system">System ID</field>
+      <field type="uint8_t" name="target_component">Component ID</field>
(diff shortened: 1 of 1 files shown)


auturgy:
This is a pretty substantial pr: will likely take a little while to work through it.
@stephendade @Williangalvani

bazfp:
I'm happy to answer any questions about the implementation and the rationale behind this change

For a little bit more background:

UUVs tend to operate in very comms limited environments (no RF / limited acoustic comms). As with more robots they have an on-board "autopilot" or "front-seat" that controls low-level vehicle movement (Control Fins to set Heading, Control Motor to Set Speed etc). They also often run a generic Autonomy system on an additional "back-seat" computer, which accepts high-level commands and breaks them down for the front-seat vehicle to consume (Scan this point, Scan this area etc.)

Typically each vehicle manufacturer creates their own bespoke interface to allow external-control from an autonomy system, we are pushing to standardise this approach with an open standard.

We are using MAVLink now as our vehicle interface for this Autonomy <-> Frontseat link and have been working with vehicle manufacturers have MAVLink integrated as a "frontseat" control interface.

hamishwillee:
Brief brain dump of some thoughts/concerns (not expecting any particular response at this time, just capturing first impressions). You may well have answered these already in the docs.
- This is essentially a "protocol dialect" or "service dialect". I very much like the concept, but there is work for MAVLink team to make sure the role of this dialect and how it fits with "the standard" is clear.
- This isn't just marine centric and really shouldn't have that name. Probably should split into payload, autonomy, sensors, the stuff around squads etc. In particular this is true if this is to go into common.xml at some point. But either way, if it doesn't apply as something all marine systems use, it should have some more specific name for vendor or more standard name for microservice functions.
- Payload service:
  - Payload is an overloaded term. 
  - The interworking and overloading of payload with component id is not clear to me. How are the IDs allocated? Does a payload have a component ID as well?
  - Is the intent just to convey status and health? Or is there more generic value to this framework? Where I'm getting here is consider a mavlink camera-gimbal. What does this give you over just having a gimbal and the gimbal info
  - Does every payload have a separate MAVLink identity? Can we have a payload that is not a mavlink item - e.g. is this just your way of showing status for everything, including non mavlink stuff in a common way?
  - The status is emitted on request. Should it be streamed?

- Autonomy service. 
  - Not entirely clear how this interworks with existing autopilot modes.
  - Who demands what? I think you're saying that an autonomy system might set a mode such as offboard in which it has full control over the vehicle. It then does whatever control algorithm makes sense for the autonomy system (I'm going to call it companion computer). At any time the autopilot can say "no, I'm in control" and switch mode. But you expect it to send a demand for control at that point? Is that right?
    -  I kind of think that the autopilot always has control and chooses to take it. A remote system should suck it up and monitor for these kinds of changes. It certainly shouldn't rely on a message in a constrained comms environment.
  - One example is a mission on a companion computer. How do the ground station, autopilot, and companion computer interact here? I.e. Where does QGC send the mission. Does the autopilot sync it, and so on?
  - @bkueng This is worth you looking at. There is some overlap with the way you are doing mode overloads on a companion computer with failsafe fallback. 
 
- Other stuff. The sensors and squad stuff is really outlying. NOt sure how it would be split up, but not really part of any kind of "standard" as is.

hamishwillee:
As @auturgy says, this is a lot to digest. 

I very much like the idea of standardization of component information on the network, including non-mavlink components.
However I am not sure how this interworks with various autopilot services and "normal mavlink". My first thoughts, in order were:
- not another enum with ids and types for stuff like camera, etc
- can/could this API simplify our existing APIs or future designs. Specifically, how might we have designed the recent gimbal API if this approach existed
- How does this interwork with GCS for various use cases, such as missions

Generally we can take a dialect like this, but if this is to be in standard it would have to be split and probably go through RFC process, as per my comment above.

FYI @julianoes @bkueng @rmackay9

bazfp:
Thanks for your reply Hamish, really useful to get your views :)

Splitting of Autonomy + Payload + Sensor things sounds good to me. I would like the new services to be available to others for maximum compatibility. I am open to iteration to get them up to scratch. I never liked running as a separate MAVLink dialect and I aimed to keep features as generic as possible. Do you suggest I pull out the services into their own files?

So the concept of the payload service stems from us having the ability to turn on onboard power-intensive sensors/payloads that are not directly MAVLink aware, but the vehicle is. Examples include things like sidescan sonar, where you only want them enabled during segments of the mission. Payload ID is unique to payloads only. We also needed a way to configure payloads, which we achieved through a mapping between parameters naming and payload IDs (not included here, but effectively PayloadID 1 will have its params accessible via `P001_SOMENAME`). We also needed to be able to support the use case where the user can add and remove these payloads on the fly. We had a brief look at gimble control, but to be honest the naming confused manufacturers a lot and we couldn't identify a way to easily register and signal these kinds of payloads with MAVLink.

I like the idea of combining payloads with components. For our use case all the payloads were controlled internally by the vehicle. The payloads do not have the capacity to be MAVLink-aware themselves. We could have had the vehicle emulate a component for each payload (MAVSDK work here @julianoes?) as they were added or removed but i'm still unsure how controlling the payloads (on/off/arm) would have worked in this case. Perhaps this is something to look at. Our payloads are controlled entirely by the Autonomy at current, but we could embed them as the a mavlink mission item also.

Now some background to the Autonomy service, the Autonomy is unrated to AP modes. The AP is almost always in control of the actual vehicle via Auto/Mission. The Autonomy is monitoring the situation and packaging missions and sending them to the AP to execute and cancelling/re-issuing where appropriate. You can think of it as an autonomous GCS. The Autonomy has a higher level view of the objectives than an Autopilot. Imagine a mission with tens of surveys. This would be an unwieldy single MAVLink mission for the AP to run. The Autonomy is aware of the full-scale mission and is planning on the fly based on the vehicle situation. The MAVLink AP is receiving at most a "chunk" of the mission that is up-to-date with the vehicles current situation. As the vehicle completes the mission, the second one is issued from the Autonomy with the latest up to date information. Imagine while the AP completes a mission "chunk" some future surveys in the overall mission are now no longer needed, the autonomy is tracking this and can adjust the future mission "chunks" as they are issued. It can also add in new chunks and respond to information recovered during the survey.

The Autonomy demands are control messages to the Autonomy system only. They can come from a ground control system to start/stop the Autonomy. They can come from an AP if the AP has a reason to demand start/stop the Autonomy system (fault? Mission from elsewhere?). If the AP accepts a mission from QGC, it should probably stop the Autonomy (if it is running!) as it's unrelated to the Autonomy mission. Alternatively as you say, the AP should just ignore any requests if Autonomy is in control, until it gets stopped or booted out. It's worth detailing these scenarios.

hamishwillee:
Thanks @bazfp - I'm on other tasks until Wednesday so I can't give this the attention it deserves until then. But yes, I think that this would be better split into 2 or three files with a protocol-specific name (might or might not eventually end up inside common.xml, or linked from it). This is mostly so that they can be digested and accepted or rejected in parts. It is such a lot to take in otherwise.

I have added this to discussion in the dev call next week. Given how much there is, I expect that will be just a broad chat about our top level thinking. I am personally very interested in this but want to understand the problems that it solves/doesn't solve better.

hamishwillee:
Payload stuff:

- we already have "payload services" here https://mavlink.io/en/services/payload.html so we need to think how this works with that.
- The normal MAVLink way to this would be to create a MAVLink identity for each component you want to control. This has a command for doing stuff and a message for notifying on status. 
  - For example, consider the winch, which has [MAV_CMD_DO_WINCH](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_WINCH) and [WINCH_STATUS](https://mavlink.io/en/messages/common.html#WINCH_STATUS)
  - The "thing" does not need a specific mavlink identity - it can be accessed through the autopilot or a companion. 
  - This is "generally better" than a generic "command something with ID", "Get status from ID", because all components are different. 
    - A mission become becomes opaque - you can't interpret it without a custom UI that knows your payloads.
    - Missions are hard to write - consider a winch - it isn't instant, and you have no real feedback. So you need to think a lot to make it work properly and you can't make it easy to be generic (FYI PX4 just added new code to allow a gripper to be used a mission sensibly)
    - Generally UIs are hard - "do actuator" in the UI could mean anything - winch, payload, sidescan sonar, or whatever.
- We do have generic things like `MAV_CMD_DO_SET_ACTUATOR` because getting stuff into MAVLink can be slow, and this allows people to do something manually. So perhaps there is a place for this.
- It is useful to understand this is primarily for non-mavlink components.
- MAVLink does not support the concept of named and reserved parameters. 
  - Parameters are explicitly flight stack specific except for the these in the [camera definition](https://mavlink.io/en/services/camera_def.html#common-parameters) - that was required to allow UIs to choose the location of specific elements.
  - That is not set in stone.
  - In addition to ideology and having to manage these, we are a bit worried about param proliferation, and there are technical reasons why this approach sucks
    - cases where the data is naturally nested
    - cases where multiple parameter need to be changed together/atomically.
- Autopilots have trouble emitting multiple component ids/identities (an architectural limitation) but you might have a component for sonar (say) on a companion. You don't need it though as long as you can discover what is available - once you have an ID you can address a command and that command can have an index for "multiples" - as per my point on the winch API.
- Adding extra component mappings for non mavlink things does add extra traffic, and needs systems to be able to properly forward stuff.

Another tool that might be useful for this is currently only available on PX4 - [component metadata](https://mavlink.io/en/services/component_information.html). The component metadata concept allows you to define deeply nested structures mapping to parameters, and has been used to define UIs. See the PX4 actuators setup - and the same concept came first in the camera definition file.
This could allow you to document your generic payload in such a way that a GCS could interpret it. 

Anyway, may take on this is that dedicated APIs mediated through companion or autopilot are generally more flexible.

hamishwillee:
Re Autonomy - I get it. The model is pretty much
- use the AP to do basic stabilization, manual control, and safety failsafe
- implement all higher level control and advanced features on a companion computer.

Just to be aware of, a flight stack always has to be able to seize control, and you can't guarantee that it will send a "taking control" message, or that the companion will get it. Normally the companion detects this by monitoring mode - i.e. if control is via a dynamically updating mission, then it will monitor the mission mode and detect it has lost control by the mode changing.
If control is via offboard/guided mode setpoints then it will monitor for change out off offboard mode.

The last case to consider is control from a companion in another mode. The example I know of here is that PX in a mission can enable obstacle avoidance. The mode is mission and the AP emits a desired trajectory. The companion responds by mirroring the trajectory or providing an avoidance path. The AP will just use its own path if no points come back. 
The approach needs to work with this kind of model too.

hamishwillee:
@bazfp We discussed briefly in the dev call https://github.com/mavlink/mavlink/wiki/20230607-Dev-Meeting. Would love to discuss next steps in the dev call in two weeks.

Wr.t.. handover of control, we discussed in the dev call. A vehicle can always take control of itself and should already emit values to indicate this in the heartbeat state - e.g. https://mavlink.io/en/messages/common.html#MAV_STATE_CRITICAL, 
https://mavlink.io/en/messages/common.html#MAV_STATE_EMERGENCY and the flight termination case. A companion should monitor for that.

hamishwillee:
@bazfp Are you up for coming along to the MAVLink call in two weeks to discuss? Julian says he tried to skype you to no avail.

FYI We talked about it in last nights call briefly. Some comments were similar 
- needs to be split into separate PRs for easier digestion
- needs better name for the protocols than "marine.xml".
- If name is changed can be merged as just a dialect but MAVLink team need to somehow make it more clear that it is a dialect. That's on us to work out how. 

New questions
- Who allocated you the 44xxx range? WHat range was allocated? This needs to be added to https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/all.xml
- There was a general comment that this PR looks great, with lots of effort clearly put in
- MAVLink team needs to think about this from the Autonomy centric world you come from, where the flight stack is not the center of the universe, and make sure that either it works for both, or the separation is clear. There are other perspectives such as the "one holistic vehicle" viewpoint, where all messages go to a vehicle, and it manages internal routing etc.

bazfp:
Thanks Hamish, appreciate these discussions and feedback. I was on leave last week so not available.

We're keen to join the next discussion and discuss in detail the next actions.

> * Who allocated you the 44xxx range? WHat range was allocated? This needs to be added to https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/all.xml

We allocated the range based on usage, happy to change if needed. 

> * There was a general comment that this PR looks great, with lots of effort clearly put in

Great to here :) we tried to maintain genericness as much as we could.

> * MAVLink team needs to think about this from the Autonomy centric world you come from, where the flight stack is not the center of the universe, and make sure that either it works for both, or the separation is clear. There are other perspectives such as the "one holistic vehicle" viewpoint, where all messages go to a vehicle, and it manages internal routing etc.

Yes this autonomy control is a new paradigm for MAVLINK, for UUVs this frontseat <-> backseat control has been used for a long time. I am keen to make sure we slot this in nicely.

I will look to split up this PR in the next two weeks

MatthewMcGill-Ocius:
A couple of questions. Separated into the separate services.
### Payload Service
**NB**: This assumes your system is running some form of MAVLINK router allowing multiple streams to be combined/split.
To register with the Autopilot the payload needs to send a MAVLINK message which requires it to be on the MAVLINK network, in which case can it just send the PAYLOAD_STATUS message itself regularly as a heartbeat rather than using the Autopilot as an intermediary.  If using the of the vehicle and a system unique COMPID it can then receive targeted commands directly.
Targeting the PAYLOAD_REQUEST_LIST to (SYSID:X,COMPID:0) could then be sent to all components with NACKS from any non-payloads.
Is there any specific requirement for registering the payloads in the autopilot?  (ie. for referencing via LUA scripts)

### Autonomy Service
**NB**: Similar to the Payload service I'm not sure using the autopilot as an intermediary for status updates is necessary as the Autonomy component will need to be on the MAVLINK network anyway.
Would this service allow registering of more generic Autonomy components? (ie AUTONOMY_COMPONENT_PATH_PLANNING, AUTONOMY_COMPONENT_GEOFENCING)

hamishwillee:
@bazfp Mav call this evening.

>> Who allocated you the 44xxx range? WHat range was allocated? This needs to be added to https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/all.xml

> We allocated the range based on usage, happy to change if needed.

Let's discuss. I'd prefer to restrict the range - probably to 100 ids in first place at 52100 - 52199

hamishwillee:
Closing as stalled. 

Reopen or repost when/if you're ready to continue the discussion. I have lost all context :-0

PR #2001, opened 2023-06-07
Add MAV_RESULT_COMMAND_UNSUPPORTED_MAV_FRAME

hamishwillee:
If MAV_CMD has a location where the frame is not explicitly specified this is taken from the `COMMAND_INT.frame` field. There is no way to query what frames are supported for a particular command by a particular flight stack. Many flight stacks just ignore the frame, using whatever arbitrary frame they support.

This adds a `MAV_RESULT_COMMAND_UNSUPPORTED_MAV_FRAME` to tell a user that the command may be valid, but not with the supported frame. If implemented a user could use this to determine what frames are supported. 

Note that if autopilots continue to do the very bad thing of silently just accepting any frame this will make no difference. But this is still "the right design" by any measure.


hamishwillee:
Agreed in dev call: https://github.com/mavlink/mavlink/wiki/20230607-Dev-Meeting

PR #2005, opened 2023-06-08
Add missionOnly="true" attribute to fence cases

hamishwillee:
The fence mav_cmds only be make sense in a mission. This tags them as such. I don't know if there are others.

FYI I did find some cases for "command only" - specifically the ones where you fetch information - unless we support dynamically configurable missions, having a command to get autopilot version or camera information not useful in a mission. Not sure if worth adding though.

Once this goes in I'll update docs to autogenerate a note for those cases.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2302,7 +2302,7 @@
         <param index="6" label="Longitude">Longitude</param>
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
-      <entry value="5001" name="MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION" hasLocation="true" isDestination="false">
+      <entry value="5001" name="MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION" hasLocation="true" isDestination="false" missionOnly="true">
         <description>Fence vertex for an inclusion polygon (the polygon must not be self-intersecting). The vehicle must stay within this area. Minimum of 3 vertices required.
         </description>
         <param index="1" label="Vertex Count" minValue="3" increment="1">Polygon vertex count</param>
@@ -2313,7 +2313,7 @@
         <param index="6" label="Longitude">Longitude</param>
         <param index="7">Reserved</param>
       </entry>
-      <entry value="5002" name="MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION" hasLocation="true" isDestination="false">
+      <entry value="5002" name="MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION" hasLocation="true" isDestination="false" missionOnly="true">
         <description>Fence vertex for an exclusion polygon (the polygon must not be self-intersecting). The vehicle must stay outside this area. Minimum of 3 vertices required.
         </description>
         <param index="1" label="Vertex Count" minValue="3" increment="1">Polygon vertex count</param>
@@ -2324,7 +2324,7 @@
         <param index="6" label="Longitude">Longitude</param>
         <param index="7">Reserved</param>
       </entry>
-      <entry value="5003" name="MAV_CMD_NAV_FENCE_CIRCLE_INCLUSION" hasLocation="true" isDestination="false">
+      <entry value="5003" name="MAV_CMD_NAV_FENCE_CIRCLE_INCLUSION" hasLocation="true" isDestination="false" missionOnly="true">
         <description>Circular fence area. The vehicle must stay inside this area.
         </description>
         <param index="1" label="Radius" units="m">Radius.</param>
@@ -2335,7 +2335,7 @@
         <param index="6" label="Longitude">Longitude</param>
         <param index="7">Reserved</param>
       </entry>
-      <entry value="5004" name="MAV_CMD_NAV_FENCE_CIRCLE_EXCLUSION" hasLocation="true" isDestination="false">
+      <entry value="5004" name="MAV_CMD_NAV_FENCE_CIRCLE_EXCLUSION" hasLocation="true" isDestination="false" missionOnly="true">
         <description>Circular fence area. The vehicle must stay outside this area.
         </description>
         <param index="1" label="Radius" units="m">Radius.</param>


julianoes (approved):
Makes sense although it's a bit confusing because geofence is not actually a "mission", but rather it's uploaded via the mission protocol as the geofence type.

hamishwillee:
You make a good point. Though missionProtocolOnly is messier. My bigger worry is that people will just choose to assume this means that everything not with this can be used in a mission.

peterbarker (approved):
LGTM

hamishwillee:
@peterbarker Thanks for looking at this. Did you see my comment in https://github.com/mavlink/mavlink/pull/2005#issuecomment-1585537053 - my concern is that `missionOnly` implies it can only be used in a mission, wheras I am applying this to a fence. 

Originally really just wanted to differentiate things between command and  mission protocol to make it easier for users and autoconstruction of UIs. If this is more confusing then I don't think we should do it.

What's your thoughts on adding attributes `fenceOnly` and `commandOnly` to the XSD. We can then group properly using the breakdown below (From claude). We'd apply only to things that are structurally those types.


<html>
<details>
<body>
<html><head></head><body><p>Based on the current <code>common.xml</code> (171 MAV_CMDs). Note upfront: only the fence and rally sets are formally partitioned by the protocol (via <code>MAV_MISSION_TYPE</code>). Mission-vs-command is mostly convention — the XML carries an explicit <code>missionOnly="true"</code> attribute on just two entries.</p>
<h2>Fence only (<code>MAV_MISSION_TYPE_FENCE</code>)</h2>

Value | Command
-- | --
5000 | MAV_CMD_NAV_FENCE_RETURN_POINT
5001 | MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION
5002 | MAV_CMD_NAV_FENCE_POLYGON_VERTEX_EXCLUSION
5003 | MAV_CMD_NAV_FENCE_CIRCLE_INCLUSION
5004 | MAV_CMD_NAV_FENCE_CIRCLE_EXCLUSION


<p>That's the whole rally set — one command.</p>
<h2>Mission only</h2>
<p>Flagged <code>missionOnly="true"</code> in the XML:</p>
<ul>
<li>177 <code>MAV_CMD_DO_JUMP</code></li>
<li>601 <code>MAV_CMD_DO_JUMP_TAG</code></li>
</ul>
<p>Mission-only by construction (they reference list position, the mission state machine, or a neighbouring item):</p>
<ul>
<li>600 <code>MAV_CMD_JUMP_TAG</code> — a label; meaningless outside a list</li>
<li>112 <code>MAV_CMD_CONDITION_DELAY</code></li>
<li>113 <code>MAV_CMD_CONDITION_CHANGE_ALT</code></li>
<li>114 <code>MAV_CMD_CONDITION_DISTANCE</code> — "distance of next NAV point"</li>
<li>4501 <code>MAV_CMD_CONDITION_GATE</code></li>
<li>93 <code>MAV_CMD_NAV_DELAY</code> — "delay the next navigation command"</li>
<li>30 <code>MAV_CMD_NAV_CONTINUE_AND_CHANGE_ALT</code> — "continue to the next command"</li>
<li>16 <code>MAV_CMD_NAV_WAYPOINT</code> — description explicitly redirects guided use to <code>DO_REPOSITION</code></li>
<li>82 <code>MAV_CMD_NAV_SPLINE_WAYPOINT</code> — spline is defined by adjacent items</li>
<li>36 <code>MAV_CMD_NAV_ARC_WAYPOINT</code> — arc start comes from the preceding item</li>
<li>25 <code>MAV_CMD_NAV_FOLLOW</code></li>
<li>81 <code>MAV_CMD_NAV_PATHPLANNING</code></li>
<li>196 <code>MAV_CMD_DO_SET_ROI_WPNEXT_OFFSET</code> — points at "next waypoint"</li>
<li>190 <code>MAV_CMD_DO_RALLY_LAND</code> — "Mission command..."</li>
<li>191 <code>MAV_CMD_DO_GO_AROUND</code> — "Mission command..."</li>
<li>31000–31004 <code>MAV_CMD_WAYPOINT_USER_1..5</code> — "user defined waypoint item"</li>
<li>31005–31009 <code>MAV_CMD_SPATIAL_USER_1..5</code> — "user defined spatial item"</li>
</ul>
<p>Watch out for two near-misses: <code>DO_LAND_START</code> (189) and <code>DO_RETURN_PATH_START</code> (188) read like mission markers, but both descriptions explicitly define command-mode behaviour (trigger a mission landing, position params ignored), so they're dual-use.</p>
<h2>Command only</h2>
<p><strong>Request/query</strong> — inherently need a live ACK + response:
410 <code>GET_HOME_POSITION</code>, 510 <code>GET_MESSAGE_INTERVAL</code>, 511 <code>SET_MESSAGE_INTERVAL</code>, 512 <code>REQUEST_MESSAGE</code>, 519 <code>REQUEST_PROTOCOL_VERSION</code> (deprecated), 520 <code>REQUEST_AUTOPILOT_CAPABILITIES</code>, 521 <code>REQUEST_CAMERA_INFORMATION</code>, 522 <code>REQUEST_CAMERA_SETTINGS</code>, 525 <code>REQUEST_STORAGE_INFORMATION</code>, 527 <code>REQUEST_CAMERA_CAPTURE_STATUS</code>, 528 <code>REQUEST_FLIGHT_INFORMATION</code>, 2002 <code>REQUEST_CAMERA_IMAGE_CAPTURE</code>, 2504 <code>REQUEST_VIDEO_STREAM_INFORMATION</code>, 2505 <code>REQUEST_VIDEO_STREAM_STATUS</code>, 5200 <code>UAVCAN_GET_NODE_INFO</code></p>
<p><strong>Preflight / config / test</strong> — most are gated on being disarmed:
241 <code>PREFLIGHT_CALIBRATION</code>, 242 <code>PREFLIGHT_SET_SENSOR_OFFSETS</code>, 243 <code>PREFLIGHT_UAVCAN</code>, 245 <code>PREFLIGHT_STORAGE</code>, 246 <code>PREFLIGHT_REBOOT_SHUTDOWN</code>, 209 <code>DO_MOTOR_TEST</code>, 310 <code>ACTUATOR_TEST</code>, 311 <code>CONFIGURE_ACTUATOR</code>, 401 <code>RUN_PREARM_CHECKS</code>, 420 <code>INJECT_FAILURE</code>, 500 <code>START_RX_PAIR</code>, 526 <code>STORAGE_FORMAT</code>, 529 <code>RESET_CAMERA_SETTINGS</code>, 533 <code>SET_STORAGE_USAGE</code>, 534 <code>SET_CAMERA_SOURCE</code>, 2520 <code>AIRFRAME_CONFIGURATION</code>, 5300 <code>DO_SET_SAFETY_SWITCH_STATE</code>, 32000 <code>CAN_FORWARD</code>, 42006 <code>FIXED_MAG_CAL_YAW</code>, 611 <code>DO_SET_GLOBAL_ORIGIN</code>, 43003 <code>EXTERNAL_POSITION_ESTIMATE</code></p>
<p><strong>Arming and mode</strong>:
400 <code>COMPONENT_ARM_DISARM</code>, 3001 <code>ARM_AUTHORIZATION_REQUEST</code>, 176 <code>DO_SET_MODE</code>, 262 <code>DO_SET_STANDARD_MODE</code></p>
<p><strong>Mission control</strong> — operate <em>on</em> a mission, so can't live inside one:
300 <code>MISSION_START</code>, 224 <code>DO_SET_MISSION_CURRENT</code>, 252 <code>OVERRIDE_GOTO</code>, 193 <code>DO_PAUSE_CONTINUE</code></p>
<p><strong>Guided / live control</strong>:
192 <code>DO_REPOSITION</code> (description says so outright), 33 <code>DO_FOLLOW_REPOSITION</code>, 221 <code>DO_GUIDED_MASTER</code>, 4000 <code>SET_GUIDED_SUBMODE_STANDARD</code>, 4001 <code>SET_GUIDED_SUBMODE_CIRCLE</code>, 43000 <code>GUIDED_CHANGE_SPEED</code>, 43001 <code>GUIDED_CHANGE_ALTITUDE</code>, 43002 <code>GUIDED_CHANGE_HEADING</code></p>
<p><strong>Interactive payload / link</strong>:
1001 <code>DO_GIMBAL_MANAGER_CONFIGURE</code>, 2004 <code>CAMERA_TRACK_POINT</code>, 2005 <code>CAMERA_TRACK_RECTANGLE</code>, 2010 <code>CAMERA_STOP_TRACKING</code>, 2510 <code>LOGGING_START</code>, 2511 <code>LOGGING_STOP</code>, 2600 <code>CONTROL_HIGH_LATENCY</code>, 10001 <code>DO_ADSB_OUT_IDENT</code></p>
<p><strong>User-defined</strong>: 31010–31014 <code>MAV_CMD_USER_1..5</code> — described as "user defined command", the counterpart to the <code>WAYPOINT_USER</code>/<code>SPATIAL_USER</code> mission items.</p>
<h2>Neither</h2>
<p>Three enum boundary markers, valid nowhere: 95 <code>NAV_LAST</code>, 159 <code>CONDITION_LAST</code>, 240 <code>DO_LAST</code>.</p>
<h2>Caveats</h2>
<p>Everything not listed above (~60 commands: <code>DO_SET_SERVO</code>, <code>DO_CHANGE_SPEED</code>, <code>DO_SET_ROI</code>, <code>IMAGE_START_CAPTURE</code>, <code>DO_PARACHUTE</code>, <code>NAV_TAKEOFF</code>, <code>NAV_LAND</code>, <code>NAV_RETURN_TO_LAUNCH</code>, the camera and gimbal setters, etc.) is legitimately dual-use.</p>
<p>Two judgement calls worth flagging. <code>CONDITION_YAW</code> (115) and <code>NAV_SET_YAW_SPEED</code> (213) look mission-flavoured but ArduPilot accepts both as commands. And the loiter family (17–19, 31) is nearly always a mission item in practice, though nothing in the spec forbids sending it as a command — I've left them out of the mission-only list on that basis.</p>
<p>Also, ArduPilot and PX4 disagree on the edges of the dual-use set, and both accept things via <code>COMMAND_LONG</code> that the spec never contemplated. If you're validating input, treat the four lists above as necessary-but-not-sufficient and check the target stack's mission-item table too.</p></body></html>
</body>
</details>
</html>

peterbarker:
Well, now you come to mention it, I was wondering whether some of these *should* available in multiple ways.

Saying "only" doesn't allow you to say, "can be used in a mission or a command - but not in an uploaded fence".

hamishwillee:
> Saying "only" doesn't allow you to say, "can be used in a mission or a command - but not in an uploaded fence".

@peterbarker Curses, POSSIBLY foiled again.

This case doesn't make sense if you have the `onlyFence`, `onlyCommand`, `onlyMission`.
- You apply `onlyFence` to the fence cases. None of those have overlap with the missions or commands anyway.
- So by default any command that doesn't have that can be used in any command or mission. 
- The benefit being we don't have to apply anything except the `onlyFence` to exclude the case.

If you want to have in missionsOnly you can add that - now it doesn't work in fence or  commands.
It falls over if you want to be able to declare something in two groups. 
Interestingly, the only case of this would be the PX4 use of landing patterns in Rally point, which you you love :-).

So right now I _think_ the `onlyXxx` case is solid and is the one that makes minimal change to the XML. It is also the one that is correct even if not applied - i.e. if we take a while to apply `onlyMission` then the XML is still "not wrong".

But it could easily break in future if we add a new type and it can be used in more than just command or missions or has any other overlap. 

So how do you feel about being more verbose?

The most flexible solution is the define the attributes `mission`, `command`, `fence`, `rally` with default `false` and apply as many as are true in each case. 
- It means additions to a lot of items, but it is non-ambiguous, extensible, and can describe all cases.
- It will make all XML wrong initially (but no invalid) because we'll be saying nothing can be used in anything. 
- NOte we're saying "allowed in X" meaning "not forbidden"

Thoughts?

XSD update https://github.com/ArduPilot/pymavlink/pull/1272

julianoes (approved):

peterbarker (changes requested):
Aaaah.... didn't we decide against this attribute?

hamishwillee:
> Aaaah.... didn't we decide against this attribute?

Yes, and once you update the XSD, this is the PR that I will update to  match it.

hamishwillee:
Closing this. Will create a new one now that https://github.com/ArduPilot/pymavlink/pull/1272 went in

Issue #2006, opened 2023-06-12
Adding support to somehow link mavlink cameras and mavlink gimbals?


Davidsastresas:
In the latest days this matter was raised with @rmackay9 and @julianoes . In gimbals with mavlink cameras this could be very useful, because some settings from gimbal could affect cameras and vice-versa. The most obvious case is the "click and point" feature getting included in QGC soon. This feature needs to know the FOV of the camera to scale the clicks appropiately, so if QGC is receiving telemetry from a mavlink camera it could deal with this appropiately.

This could be also interesting if we bring back support for camera pan/tilt for joysticks connected to QGC, for the same reason above, we would want much less speed at high levels of zoom. And this is also true for camera/gimbal management within Autopilot firmware for the same reason, to auto scale manual pan/tilt control.

It could be also useful if at some point we enable more advanced features like that cool feature DJI has of selecting an area of interest in the video feed to take a "super resolution" photo, where camera is zoomed in and gimbal is moved to take high resolution pictures of the selected area of interest, that could be later merged in a super high res picture. This is a random example but for something like this it is interesting as well that whoever is managing this ( GCS, companion computer, autopilot ) will need to manage camera triggers and gimbal movement at the same time, and will need information about FOV, etc.

@julianoes suggested extending a field in camera and gimbal information that could link them to one another, and this could be the component id of the gimbal/camera associated with it. I like this idea very much.

We thought it was worth bringing this discussion to get more input.

Thanks!

auturgy:
Linking to https://github.com/mavlink/mavlink/issues/1678

rmackay9:
In ArduPilot at the moment we always assume that the camera and gimbal instances match.  So camera1 is in gimbal1.  This has worked so far because the number of users is small and we haven't yet hit complaints from users where there are multiple cameras within a single gimbal.

Adding a field to CAMERA_INFORMATION and GIMBAL_MANAGER_INFORMATION seems OK to me although I guess in the multiple-camera-in-single-gimbal case the gimbal-device-information would only have the first camera in the field.  Perhaps it is sufficient to only add the gimbal_id to the CAMERA_INFORMATION message assuming there is no way that a single camera could be in multiple gimbals.

Davidsastresas:
> In ArduPilot at the moment we always assume that the camera and gimbal instances match. So camera1 is in gimbal1. This has worked so far because the number of users is small and we haven't yet hit complaints from users where there are multiple cameras within a single gimbal.
> 
[...]

PR #2009, opened 2023-06-17
common: add missing gimbal_device_id

julianoes:
By adding a gimbal_device_id to GIMBAL_DEVICE_ATTITUDE_STATUS we can actually identify which gimbal device is sending a message if there are multiple gimbals with the same component ID as the gimbal manager.

As identified by @Davidsastresas


Davidsastresas:
Thanks! This is really important to differentiate to which device belongs a  GIMBAL_DEVICE_ATTITUDE_STATUS when these messages come from the same compid. This can happen for gimbals that are not 100% mavlink compliant, so an autopilot is

message_definitions/v1.0/common.xml
```diff
@@ -6985,6 +6985,7 @@
       <extensions/>
       <field type="float" name="delta_yaw" units="rad" invalid="NAN">Yaw angle relating the quaternions in earth and body frames (see message description). NaN if unknown.</field>
       <field type="float" name="delta_yaw_velocity" units="rad/s" invalid="NAN">Yaw angular velocity relating the angular velocities in earth and body frames (see message description). NaN if unknown.</field>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">This field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and should be set to 0.</field>
```
auturgy, line 6988: Suggest zero indexed rather than zero if component ID sent. That way can be trimmed if only one gimbal. Not blocking.

auturgy (commented):
It would be nice to be able to trim the other extensions... oh well.

message_definitions/v1.0/common.xml
julianoes: Right, I see what you're saying :thinking:.. We would have to change it in all the gimbal messages.

auturgy: It's only one byte, not a big deal.

hamishwillee:
Just so I am clear on this:
- in another issue we discussed that each gimbal device must have its own MAVLink component identity; so if you had two gimbals attached to your autopilot then you would need the autopilot to emit HEARTBEATs for each of the gimbal devices, and the same address would be used to send GIMBAL_DEVICE_ATTITUDE_STATUS for each.
- But "some" autopilots emitting just their own heartbeat and GIMBAL_DEVICE_ATTITUDE_STATUS  with that same heartbeat. So if you have two of them there is no way to differentiate.
- With this solution the Autopilot would send out two GIMBAL_DEVICE_ATTITUDE_STATUS , with gimbal device_id being 1, 2, etc. 
- Is that right?

Thoughts.
- The conclusion of the last discussion was that you either had just one gimbal attached to the autopilot or they had to have a separate mavlink identity. What changed?
- How does a user understand the mapping to the new id? I mean currently you have https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_INFORMATION device_id which is explicitly a component id, and which you use to discover your gimbal and map to a particular gimbal device. Now you have two devices on one component.

This "feels messy". Am I missing something?

hamishwillee:
PS, if we do #2006, what would the camera indicate is its gimbal?

auturgy:
Context is slightly different. Basically it comes down to whether the gimbal is a native mavlink component or not.

julianoes:
> The agreement was that every gimbal device had to have a mavlink identity. If you just have one gimbal attached to the autopilot you're fine because it can share the autopilot identity (component id). But if you have more than one then the autopilot has to fake another gimbal device heartbeat, so that the gimbal can be addressed.

Right, so @Davidsastresas and I basically went back on that and it seems like that's not something that is likely to actually be implemented that way in PX4 and ArduPilot, hence we're back to special IDs 1..6 for that case.

julianoes:
We talked about this and @hamishwillee wants me to:
1. Make sure the device ID is noted correctly in all places in xml, so it's all consistent.
2. Write a special paragraph in the docs explaining it.

julianoes:
> 1. Make sure the device ID is noted correctly in all places in xml, so it's all consistent.

Done in last commit [427fb9b](https://github.com/mavlink/mavlink/pull/2009/commits/427fb9b988087cd20ceb1a69432d333557ad7202).

> 2. Write a special paragraph in the docs explaining it.

https://github.com/mavlink/mavlink-devguide/pull/510

Davidsastresas:
Understood. So, If I am not mistaken, from Ardupilot, which is currently using device id 1-6 for non mavlink gimbals, we should use now 2-6 instead, correct?

Just out of curiosity, In which scenario would the device id 1 be used? Because that would correspond to autopilot compid, and if the autopilot is reporting a gimbal manager it will use the magic 2-6 anyways right? Not saying it isn't right, I just want to understand the usecase for device id/compid 1

rmackay9:
Hi, thanks for all this.

I've posted on [the wiki PR ](https://github.com/mavlink/mavlink-devguide/pull/510)(which was probably the wrong place) but in any case, I think the use of "0" and then "2~6" is quite confusing.  I think we should stick with the range 1~6 unless we know of real examples of where this is going to cause problems.

julianoes:
My concern was that one is also usually the component ID of the autopilot. So if the autopilot is the gimbal manager and advertises the `gimbal_device_id` as 1 then that could be mistaken for component ID 1 which it literally is. I'm not sure if that would actually cause a problem or naturally just work anyway. Given it doesn't necessarily conflict as such I can be convinced to use the range from 1..6 indeed.

Davidsastresas:
Now that Randy mentions that, if we stick to gimbal device id 1-6, I can not think where we could have a conflict with gimbal id 1. For sending commands, they will be directed to compid 1 ( autopilot ) and gimbal device ( 1, 2 etc ). For receiving attitude_status etc, we will receive it from compid 1 and whatever gimbal device. 

We will always have the compid information anyway. The only possible conflict could be if we have a very weird setup like:

- Autopilot compid not being 1, and gimbal manager compid being 1, which is extremely odd, such mavlink configuration should never exist.

I think it is probably fine to slot those 1-6 gimbal device id for gimbal managers that live in the same compid as the autopilot. We have already 6 compid designated to gimbals anyway:

MAV_COMP_ID_GIMBAL=154
MAV_COMP_ID_GIMBAL2=171
MAV_COMP_ID_GIMBAL3=172
MAV_COMP_ID_GIMBAL4=173
MAV_COMP_ID_GIMBAL5=174
MAV_COMP_ID_GIMBAL6=175

So probably values 1-6 for non mavlink gimbals living in autopilot, and 154, 171-175 is all the values we will ever need for this. I think it is very unlikely that we could have more than those 6 mavlink gimbals +/or 6 non mavlink gimbals within the same sysid.

So basically we only need to make sure that no gimbal devices use ever compid 1-6, which is not a crazy suggestion, that is why we have the above suggested compids for gimbals anyway.

I think for me it makes sense this way, but it could be that we are missing something. Do you guys can think of any scenario where this 1-6 gimbal device id could be problematic?

julianoes:
Sounds good, thanks for thinking that through again @Davidsastresas. I will go ahead and change it (back).

rmackay9 (approved):

message_definitions/v1.0/common.xml
```diff
@@ -6935,19 +6935,21 @@
       <field type="float" name="pitch_max" units="rad" invalid="NaN">Maximum hardware pitch angle (positive: up, negative: down). NAN if unknown.</field>
       <field type="float" name="yaw_min" units="rad" invalid="NaN">Minimum hardware yaw angle (positive: to the right, negative: to the left). NAN if unknown.</field>
       <field type="float" name="yaw_max" units="rad" invalid="NaN">Maximum hardware yaw angle (positive: to the right, negative: to the left). NAN if unknown.</field>
+      <extensions/>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">This field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and should be set to 0.</field>
```
hamishwillee, line 6939:
```suggestion
      <field type="uint8_t" name="gimbal_device_id" invalid="0">This field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and must be set to 0.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -6985,6 +6987,7 @@
       <extensions/>
       <field type="float" name="delta_yaw" units="rad" invalid="NAN">Yaw angle relating the quaternions in earth and body frames (see message description). NaN if unknown.</field>
       <field type="float" name="delta_yaw_velocity" units="rad/s" invalid="NAN">Yaw angular velocity relating the angular velocities in earth and body frames (see message description). NaN if unknown.</field>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">This field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and should be set to 0.</field>
```
hamishwillee, line 6990:
```suggestion
      <field type="uint8_t" name="gimbal_device_id" invalid="0">This field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and must be set to 0.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -6985,6 +6987,7 @@
       <extensions/>
       <field type="float" name="delta_yaw" units="rad" invalid="NAN">Yaw angle relating the quaternions in earth and body frames (see message description). NaN if unknown.</field>
       <field type="float" name="delta_yaw_velocity" units="rad/s" invalid="NAN">Yaw angular velocity relating the angular velocities in earth and body frames (see message description). NaN if unknown.</field>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">This field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and must be set to 0.</field>
```
hamishwillee, line 6990: > I think for me it makes sense this way, but it could be that we are missing something. Do you guys can think of any scenario where this 1-6 gimbal device id could be problematic?

I think it is fine for the autopilot case. 

>  field is to be used if the gimbal manager and the gimbal device are the same component and hence have the same component ID. This field is then set a number between 1-6. If the component ID is separate, this field is not required and must be set to 0

I have a slight concern with the above text where a mavlink external gimbal  implements both the gimbal manager protocol? In this case they have the same component ID. So you'd be expected to use a number from 1-6 in the device id, but you don't need one to differentiate the device and manager.

The reason we need that on the autopilot is to differentiate multiple faked gimbal devices which can only have one component ID. 

Not sure what we propose here @julianoes

julianoes: I think it's ok. @Davidsastresas thought it through in https://github.com/mavlink/mavlink/pull/2009#issuecomment-1606051711, no?

hamishwillee: I don't think so. That doesn't cover the external MAVLink combined gimbal manager/device (I don't think).
It won't "break", but we should be aware that, as written, such a gimbal would have to set the device ID number 1 - 6, even though it it could perfectly well be addressed using the component ID.

julianoes: Yes, I was hoping that's clear from the description.

hamishwillee: It is completely clear and consistent. It just feels wrong because a GCS has to do a bit more work for a case where it should not have to. 

Since this is your intent. I'm happy to approve.

hamishwillee (approved):

hamishwillee:
I think we have reached consensus. Merging.

PR #2010, opened 2023-06-21
development.xml - remove mission_checksum

hamishwillee:
This PR removes MISSION_CHECKSUM .

MISSION_CHECKSUM is supposed to provide a checksum of various parts of the plan. The theory is that a ground station can calculate the same checksum for its current plan and therefore know if it has the current fence, mission, rally points. 

Unfortunately the theory is flawed because flight stacks do not necessarily store all the data in the original mission.

Replaced the functionality by https://github.com/mavlink/mavlink/pull/2012


hamishwillee:
FYI @auturgy I am merging this. It doesn't work and we don't want anyone else to try implement it.

PR #2014, opened 2023-06-25
Master pr/camera info gimbal associated

Davidsastresas:
Following #2006 issue. This commit adds a new field to CAMERA_INFORMATION to link it to a gimbal device. 

I think the convention should follow the same logic as gimbal device id, which is 1-6 for non mavlink gimbals ( gimbals devices "simulated" by an autopilot ) and the rest of the numbers should be interpreted as component ID, for gimbal devices that are actually an independent mavlink system.

I am not that familiar with mavlink, so I am not sure if this way it is all right, in the sense of how to present it so the message is trimmed if we are not using this field, please let me know if it needs to be fixed.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6715,7 +6715,8 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration).  Use 0 if not known.</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
-      <field type="uint8_t" name="gimbal_device_id" invalid="0">If this camera is associated with a gimbal, this field is the gimbal device id, following MAVLink Gimbal Protocol v2. The id is the component id of the gimbal device, or 1-6 reserved for non mavlink gimbals ). Use 0 if no gimbal is associated with the camera.</field>
+      <extensions/>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">Gimbal id of a gimbal associated with this camera. This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.</field>
     </message>
     <message id="260" name="CAMERA_SETTINGS">
       <description>Settings of a camera. Can be requested with a MAV_CMD_REQUEST_MESSAGE command.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6715,6 +6715,7 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration).  Use 0 if not known.</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">If this camera is associated with a gimbal, this field is the gimbal device id, following MAVLink Gimbal Protocol v2. The id is the component id of the gimbal device, or 1-6 reserved for non mavlink gimbals ). Use 0 if no gimbal is associated with the camera.</field>
     </message>
     <message id="260" name="CAMERA_SETTINGS">
       <description>Settings of a camera. Can be requested with a MAV_CMD_REQUEST_MESSAGE command.</description>


message_definitions/v1.0/common.xml
```diff
@@ -6715,6 +6715,7 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration).  Use 0 if not known.</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">If this camera is associated with a gimbal, this field is the gimbal device id, following mavlink gimbal protocol v2 ( gimbal device compid, or 1-6 reserved for non mavlink gimbals ). Use 0 if no gimbal associated.</field>
```
hamishwillee, line 6718: I won't put any thought into this until the status changes out of draft, but this is quick subedit.
```suggestion
      <field type="uint8_t" name="gimbal_device_id" invalid="0">If this camera is associated with a gimbal, this field is the gimbal device id, following MAVLink Gimbal Protocol v2. The id is the component id of the gimbal device, or 1-6 reserved for non mavlink gimbals ). Use 0 if no gimbal is associated with the camera.</field>
```

Davidsastresas:
Thanks for the review, I fixed it and pushed force it.

About the draft status I agree, It doesn't make sense to put a lot of time now until we understand 100% were we want to go with this change. From my perspective I am happy with it, this would allow a GCS to understand how to link a camera and a gimbal, 
 and it will also allow an autopilot to do so, this way we can scale automatically the gimbal speed/angle control based on the zoom/FOV.
 
Maybe @rmackay9 and @julianoes can take a look as well. Does it look good? does it make sense to undraft this?

Thanks!

rmackay9:
Hi @Davidsastresas,

In general, it looks OK to me but certainly the new fields be after a "<extensions/>" tag or it will break existing users of the message.

message_definitions/v1.0/common.xml
```diff
@@ -6715,6 +6715,7 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration).  Use 0 if not known.</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">If this camera is associated with a gimbal, this field is the gimbal device id, following MAVLink Gimbal Protocol v2. The id is the component id of the gimbal device, or 1-6 reserved for non mavlink gimbals ). Use 0 if no gimbal is associated with the camera.</field>
```
rmackay9, line 6718: Definitely not a blocker but if it were me I would make it more concise. Perhaps something like this:

The gimbal id of the gimbal associated with this camera.  This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.

julianoes: It would be nicer to have a camera linked to a gimbal manager, however, that would not be unique, so I guess this is ok.

Or alternatively, we could link a camera ID in gimbal_manger_information

Davidsastresas: What @julianoes said about linking a camera ID in gimbal manager information would work as well, and maybe it is neater, more elegant right?

rmackay9: @Davidsastresas @julianoes,

I think we discussed this alternative before and concluded that it wouldn't be complete because the gimbal manager info message is for a single gimbal but that gimbal could have multiple cameras in it.  There will never be a camera within two gimbals though so it makes sense to put the mapping in the camera message.  .. and we don't need both the gimbal message and the camera message to have fields referring to each other.

julianoes: Ah, good point!

julianoes (commented):
I think this is good but should be after `</extensions>`.

message_definitions/v1.0/common.xml
hamishwillee: Yes, and also it might be that other things want to end up being associated with a gimbal one day, so having the association at the other end is desirable.

hamishwillee: This adds in the extensions tag. I also updated to use @rmackay9 wording.

```suggestion
      <extensions/>
      <field type="uint8_t" name="gimbal_device_id" invalid="0">Gimbal id of a gimbal associated with this camera. This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.</field>
```

hamishwillee:
> I am not that familiar with mavlink, so I am not sure if this way it is all right, in the sense of how to present it so the message is trimmed if we are not using this field, please let me know if it needs to be fixed.

This is correct now.

FYI only @Davidsastresas MAVLink reorders message fields for sending "by default" to be larger data-size fields first, with data of same size being in its declaration order in XML. Fields after `</extensions>` are then appended to the end in XML declaration order. 
Truncation happens after re-ordering - any zero-filled fields are stripped off for sending.

You can't modify any of the fields before extensions without making the message incompatible with recipients that have not been updated. However you can add extensions, with care (they are not part of the CRC_EXTRA calculation used to match sender/recipient versions of messages).

So to make this change we have to add an extension field and then make it have a default/invalid value of 0. That way the field is truncated if it is not used, provided someone doesn't come along later and further extend the message. Further, recipients that have not updated will not get the new value. Recipients that have updated will always get 0 from senders that have not updated. 

There is some outline of this here https://mavlink.io/en/guide/define_xml_element.html#message_extensions

hamishwillee:
I'm happy with this, but I'd very much prefer not to merge this into the "standard" until there is a tested implementation that shows it does indeed to what we need.

rmackay9:
@hamishwillee,

Great.  Maybe we should squash the commits together and use the Co-Author feature in the description so it appears as if both you and DavidS worked on it.  Then I think we're OK to merge it.

I think AP will implement this in the near future... within a month I'd guess.

julianoes (approved):

message_definitions/v1.0/common.xml
```diff
@@ -6715,6 +6715,8 @@
       <field type="uint32_t" name="flags" enum="CAMERA_CAP_FLAGS" display="bitmask">Bitmap of camera capability flags.</field>
       <field type="uint16_t" name="cam_definition_version">Camera definition version (iteration).  Use 0 if not known.</field>
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
+      <extensions/>
+      <field type="uint8_t" name="gimbal_device_id" invalid="0">Gimbal id of a gimbal associated with this camera. This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.</field>
```
hamishwillee, line 6719: Does anyone mind that as soon as this is used the cam_definition_uri can no longer be truncated.

hamishwillee: If no objections, @auturgy I plan to merge this tomorrow, even though it is not prototyped. Can undo provided it never goes into a release, because it is an extension.

rmackay9: It's sad that it can't be truncated but I don't think there's anything we can do about it because I think some cameras have already implemented this message..?  Maybe I could check around with Gremsy, ViewPro, etc..

hamishwillee: PX4 will be emitting this in a release build right @julianoes - i.e. it is too late to adjust the message?

It's a bit of a pain, but I'm not too concerned about this because the expectation is that it is requested, not streamed, as it only needs to be sent once.

rmackay9: yes, as you say, it's probably too late and it won't be sent all that often so it's likely OK.

rmackay9 (approved):
Let's squash the commits of course.  I probably don't need to say this but just in case..

hamishwillee:
Done!

Davidsastresas:
Sorry, I've been busy the past weeks and I haven't spared the time to check the pending PR until now. I agree 100% with the modifications done, and I also agree that the truncation of the uri is not that critical because this is a message that is not streamed, but requested just in the handshake with the camera.

I think we will have use for this extension shortly. Thank you very much!

PR #2015, opened 2023-07-03
Add VELOCITY_LIMITS msg

marcinolokk:
This pr is part of a work which aims to enable "slow mode" on a vehicle. This mode allows to scale down the vehicle's velocities by limiting the maximum values. It is useful in scenarios such as: EASA 3m/s velocity limit, camera zooming. Corresponding work can be found [here](https://github.com/PX4/PX4-Autopilot/pull/21790)
The velocity_limits message consists of: 
 * ~horizontal velocity limit~ horizontal speed limit
 * ~vertical velocity limit~ vertical speed limit 
 * yaw rate limit

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4927,19 +4927,6 @@
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
-    <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-    <enum name="VELOCITY_LIMITS_MASK" bitmask="true">
-      <description>These encode velocity setpoints that are currently enabled.</description>
-      <entry value="1" name="VELOCITY_LIMITS_MASK_HORIZONTAL">
-        <description>Accept horizontal speed setpoint</description>
-      </entry>
-      <entry value="2" name="VELOCITY_LIMITS_MASK_VERTICAL">
-        <description>Accept vertical speed setpoint</description>
-      </entry>
-      <entry value="4" name="VELOCITY_LIMITS_MASK_YAW">
-        <description>Accept yaw rate setpoint</description>
-      </entry>
-    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
@@ -7268,15 +7255,13 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
-    <wip/>
-    <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-    <!-- TODO: Make sure the ID and naming are appropiate -->
     <message id="354" name="VELOCITY_LIMITS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. This message should be streamed at 1Hz while the lower limits apply. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
-      <field type="float" name="horizontal_speed" units="m/s">Horizontal speed in m/s</field>
-      <field type="float" name="vertical_speed" units="m/s">Vertical speed in m/s</field>
-      <field type="float" name="yaw_rate" units="rad/s">Yaw rate in rad/s</field>
-      <field type="uint8_t" name="type_mask" enum="VELOCITY_LIMITS_MASK" display="bitmask">Bitmap to indicate which velocity setpoint should be considered.</field>
+      <field type="float" name="horizontal_velocity_limit" default="NaN" units="m/s">Limit for horizontal movement in local inertial north east frame. NaN: Use vehicle default speed</field>
+      <field type="float" name="vertical_velocity_limit" default="NaN" units="m/s">Limit for vertical movement in local inertial down frame. NaN: Use vehicle default speed</field>
+      <field type="float" name="yaw_rate_limit" default="NaN" units="rad/s">Limit for vehicle turn rate around its yaw axis. NaN: Use vehicle default rate</field>
     </message>
     <message id="360" name="ORBIT_EXECUTION_STATUS">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4927,6 +4927,19 @@
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
+    <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+    <enum name="VELOCITY_LIMITS_MASK" bitmask="true">
+      <description>These encode velocity setpoints that are currently enabled.</description>
+      <entry value="1" name="VELOCITY_LIMITS_MASK_HORIZONTAL">
+        <description>Accept horizontal speed setpoint</description>
+      </entry>
+      <entry value="2" name="VELOCITY_LIMITS_MASK_VERTICAL">
+        <description>Accept vertical speed setpoint</description>
+      </entry>
+      <entry value="4" name="VELOCITY_LIMITS_MASK_YAW">
+        <description>Accept yaw rate setpoint</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
@@ -7254,6 +7267,15 @@
       <field type="uint16_t" name="array_id" instance="true">Unique ID used to discriminate between arrays</field>
       <extensions/>
       <field type="float[58]" name="data">data</field>
+    </message>
+    <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+    <!-- TODO: Make sure the ID and naming are appropiate -->
+    <message id="354" name="VELOCITY_LIMITS">
+      <description>Setpoint of horizontal speed, vertical speed and yaw rate.</description>
+      <field type="float" name="horizontal_speed" units="m/s">Horizontal speed in m/s</field>
+      <field type="float" name="vertical_speed" units="m/s">Vertical speed in m/s</field>
+      <field type="float" name="yaw_rate" units="rad/s">Yaw rate in rad/s</field>
+      <field type="uint8_t" name="type_mask" enum="VELOCITY_LIMITS_MASK" display="bitmask">Bitmap to indicate which velocity setpoint should be considered.</field>
     </message>
     <message id="360" name="ORBIT_EXECUTION_STATUS">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7258,10 +7258,10 @@
     <message id="354" name="VELOCITY_LIMITS">
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
-      <description>Temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. This message should be streamed at 1Hz while the lower limits apply. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
-      <field type="float" name="horizontal_velocity_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME LOCAL. NaN: Use vehicle default speed</field>
-      <field type="float" name="vertical_velocity_limit" default="NaN" units="m/s">Limit for vertical movement in MAV_FRAME LOCAL. NaN: Use vehicle default speed</field>
-      <field type="float" name="yaw_rate_limit" default="NaN" units="rad/s">Limit for vehicle turn rate around its yaw axis. NaN: Use vehicle default rate</field>
+      <description>Temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
+      <field type="float" name="horizontal_speed_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME_LOCAL_NED. NaN: Field not used (ignore)</field>
+      <field type="float" name="vertical_speed_limit" default="NaN" units="m/s">Limit for vertical movement in MAV_FRAME_LOCAL_NED. NaN: Field not used (ignore)</field>
+      <field type="float" name="yaw_rate_limit" default="NaN" units="rad/s">Limit for vehicle turn rate around its yaw axis. NaN: Field not used (ignore)</field>
     </message>
     <message id="360" name="ORBIT_EXECUTION_STATUS">
       <wip/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7268,6 +7268,7 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
+    <wip/>
     <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
     <!-- TODO: Make sure the ID and naming are appropiate -->
     <message id="354" name="VELOCITY_LIMITS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7259,8 +7259,8 @@
       <wip/>
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. This message should be streamed at 1Hz while the lower limits apply. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
-      <field type="float" name="horizontal_velocity_limit" default="NaN" units="m/s">Limit for horizontal movement in local inertial north east frame. NaN: Use vehicle default speed</field>
-      <field type="float" name="vertical_velocity_limit" default="NaN" units="m/s">Limit for vertical movement in local inertial down frame. NaN: Use vehicle default speed</field>
+      <field type="float" name="horizontal_velocity_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME LOCAL. NaN: Use vehicle default speed</field>
+      <field type="float" name="vertical_velocity_limit" default="NaN" units="m/s">Limit for vertical movement in MAV_FRAME LOCAL. NaN: Use vehicle default speed</field>
       <field type="float" name="yaw_rate_limit" default="NaN" units="rad/s">Limit for vehicle turn rate around its yaw axis. NaN: Use vehicle default rate</field>
     </message>
     <message id="360" name="ORBIT_EXECUTION_STATUS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7255,14 +7255,6 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
-    <message id="354" name="VELOCITY_LIMITS">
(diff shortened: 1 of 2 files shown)


hamishwillee:
Hi @marcinolokk Since this is in draft, this is just a fly-by comment:

1. I wouldn't merge this in common until there was a working prototype. If you need this merged it should go into development.xml.
2. The design and context isn't entirely clear- you need some more description of the workflow. 

   My "guess" is that a vehicle is being driven using setpoints provided by an external system. The vehicle knows it needs to slow for some reason that the thing providing the setpoints does not - so it provides direction to the system providing the setpoint, on the expectation that the setpoints will then be modified to a lower rate.
   
   It isn't clear what the vehicle should do if the supplier of setpoints doesn't obey the instruction.
   
3. The restriction to local frame seems unnecessary, but it might not be, depends on your answer to "2".

marcinolokk:
Thanks @hamishwillee for looking into it.

1. It will be the draft PR until we have a working prototype.
2. The entire workflow is still under development and its further description will be updated once it's ready. Meanwhile maybe @cmic0 could give some clarifiaction.
3. I am not sure how to refer to this one, @MaEtUgR could you take a look?

cmic0:
hey @hamishwillee!

> The design and context isn't entirely clear- you need some more description of the workflow.

Let me try to add some context (using PX4 as reference autopilot):
When flying in Position Mode, the vehicle is allowed to fly to a maximum velocity defined by a parameter. In production vehicles the maximum speed is defined by the OEM and users are in general not allowed to change this value (it's like changing the maximum speed of your car). What we want to achieve with this message is the possibility to introduce the "Speed Limiter" feature, i.e temporarily limiting the maximum speed at which the vehicle can fly when in Position mode.

In this view, it's the autopilot responsibility to implement the logic to constrain the speed ( work covered here https://github.com/PX4/PX4-Autopilot/pull/21790) - would this initial work be enough to be considered a working prototype?

This can be used for different applications:
- In case you have an obstacle avoidance system running on a companion computer, you will have a clear interface to dynamically constrain speeds based on the avoidance data
- GCSs can expose the option to the user to set custom speed limits.
- Develop a "tripod mode" that limits the vehicle's speed to 3m/s (European regulation) to fly closer to people

MaEtUgR:
> You stream it at the vehicle at a set rate. The vehicle then obeys the limit while getting the stream. If the stream stops, it can revert after a timeout to the "default" cruise speed. Is that right?

Yes exactly. One reason for this choice was that the speed can be streamed if it changes continuously e.g. by a knob operated by the user or based on constantly changing camera zoom levels and the second reason is that if whatever is setting these limits stops streaming because of a bug or link loss or whatever the defaults are back after a 2 seconds timeout. We're open for feedback but it seems like a better solution than commands to me although I know there's `DO_CHANGE_SPEED` which is used in autonomous modes and AFAIK doesn't have an effect in manually piloted modes.

hamishwillee:
@marcinolokk When this is ready for formal review, let me know or undraft. Basically if something is in draft I assume you are still working on it, so I won't chase other opinions.

>> You stream it at the vehicle at a set rate. The vehicle then obeys the limit while getting the stream. If the stream stops, it can revert after a timeout to the "default" cruise speed. Is that right?

> Yes exactly. One reason for this choice was that the speed can be streamed if it changes continuously e.g. by a knob operated by the user or based on constantly changing camera zoom levels and the second reason is that if whatever is setting these limits stops streaming because of a bug or link loss or whatever the defaults are back after a 2 seconds timeout. We're open for feedback but it seems like a better solution than commands to me although I know there's DO_CHANGE_SPEED which is used in autonomous modes and AFAIK doesn't have an effect in manually piloted modes.

Thanks @MaEtUgR. I think this approach makes sense then.

Streaming is OK if the limits can need to change rapidly, and you don't need any confirmation of success, or even that the system is trying to handle your instruction.  Is that really the case? I.e. if your vehicle ignores this command for whatever reason, do you need to know, and if so, how will you?

Or to put it another way, do we need a way to publish current limits or can we monitor the speed and if we are streaming this and it doesn't go over the limit/slows we know things are OK?

In addition, this states that the refresh rate is 1Hz refresh - how well do you think that will work with someone twiddling the zoom knob, or you're about to fly into a "full of people" zone? I put it to you that the required rates are likely to be more similar to those for offboard mode setpoints in some cases. So not dismissing the approach, just some of the statements.

----

Asides:

- You can't use `DO_CHANGE_SPEED` for this either way because it is defined as changing the speed setpoint. A setpoint is not a limit, and will immediately be overridden by any other setpoint. That's somewhat fine in a mission because the speed setpoint is usually set on a waypoint and doesn't change, but doesn't work for this case, where you would want the limit to apply even in manual modes.
[I say, "somewhat" because if you go into hold, you're changing mode and clearing that setpoint. We really need a way to set the speed in missions for the life of the mission until replaced]

marcinolokk:
@hamishwillee I will undraft the PR once it is ready. Due to the change in product requirement we stopped working on it for now.

message_definitions/v1.0/common.xml
```diff
@@ -7255,6 +7255,14 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
+    <message id="354" name="VELOCITY_LIMITS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. This message should be streamed at 1Hz while the lower limits apply. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
+      <field type="float" name="horizontal_velocity_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME LOCAL. NaN: Use vehicle default speed</field>
```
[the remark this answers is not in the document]
marcinolokk: I addressed your comments in the latest commit. Picked `MAV_FRAME_LOCAL_NED` as it looks appropriate to me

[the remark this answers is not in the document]
marcinolokk: @MaEtUgR @cmic0 Changing CRC would require additional update to our implementation that uses `VELOCITY_LIMITS` in current form right?

marcinolokk:
Hi @hamishwillee. We would like to bring back the PR. The requirements of the product changed slightly but we still want to use the same VELOCITY_LIMITS message.

> Or to put it another way, do we need a way to publish current limits or can we monitor the speed and if we are streaming this and it doesn't go over the limit/slows we know things are OK?

As for now we don't see the need for acknowledgment whether the limits are respected by the Autopilot. Please correct if I'm wrong.

>In addition, this states that the refresh rate is 1Hz refresh

I removed this part from the description as the timeout can be specific to a platform/application.

hamishwillee:
Hi @marcinolokk 

I'll have a proper think about this tomorrow, but in passing. I'll see if I can get some eyes on this over the week. Do you think you or someone else might be able to attend next weeks MAVLink call to discuss it (assuming there is some contention)?

> As for now we don't see the need for acknowledgment whether the limits are respected by the Autopilot. Please correct if I'm wrong.

That depends. Will someone be fined or die if you don't provide acknowledgements? If so then you might expect some kind of warning on limits. These kinds of things depend on use cases though.

To my mind you don't need it, and if later you do, then a new API might be considered.

>> In addition, this states that the refresh rate is 1Hz refresh

> I removed this part from the description as the timeout can be specific to a platform/application.

Makes sense to me.

message_definitions/v1.0/common.xml
```diff
@@ -7255,6 +7255,14 @@
       <extensions/>
       <field type="float[58]" name="data">data</field>
     </message>
+    <message id="354" name="VELOCITY_LIMITS">
```
hamishwillee, line 7258: Can't be in development.xml?

hamishwillee:
@auturgy What do you think of this one?

It's a message that can be streamed at the autopilot to limit maximum vertical velocity, horizontal velocity, and yaw rate (all settings optional). If the message is not received within a timeout the speed reverts to the system default - i.e. it's a fall-out like in offboard mode.

This is designed to support a slow mode or speed limit. This is useful where [regulation](https://www.bazl.admin.ch/bazl/en/home/drohnen/open/900_4kg.html) allows drones that can be speed limited to fly closer to people. It is also useful to trigger a speed limited flight when using the camera. The camera use case is why this is a message and not a command - since the limits could be conditional on zoom level, which might be controlled on control knob.

You might of course implement slow modes in different ways - such as geofences that have an associated speed limit and so on but I don't think those preclude this, or meet all the use cases.

Note that the intent is for this to go into PX4 https://github.com/PX4/PX4-Autopilot/pull/21790 (not just downstream). Would ArduPilot use something like this? What testing do you need for merging.

marcinolokk:
> Do you think you or someone else might be able to attend next weeks MAVLink call to discuss it (assuming there is some contention)?

I guess me, @MaEtUgR  or @tstastny could participate in one

>That depends. Will someone be fined or die if you don't provide acknowledgements? If so then you might expect some kind of warning on limits. These kinds of things depend on use cases though.

Currently, the lost stream is handled by the Autopilot. When a stream disruption occurs, the pilot receives a warning, and the Autopilot retains the last received limits. This behavior may vary depending on the platform or application. Does it make sense to you?

hamishwillee:
> Currently, the lost stream is handled by the Autopilot. When a stream disruption occurs, the pilot receives a warning, and the Autopilot retains the last received limits. This behavior may vary depending on the platform or application. Does it make sense to you?

My understanding is that you are streaming this message while the limit is required. If the autopilot stops getting the signal it latches the last value for a timeout. After the timeout it would remove the speed setting.

I don't think you can send a warning when the stream is disrupted - that's expected. Other than perhaps "leaving a speed limited area".

But perhaps I just don't understand what you mean by disruption.

To be clear though, I don't really see this as a problem - the sender is streaming the message. There is only a problem if the thing doing the streaming MUST know that it succeeded or not. Generally that is not the case - because if the autopilot is not getting the message you have bigger problems.

marcinolokk:
>My understanding is that you are streaming this message while the limit is required. If the autopilot stops getting the signal it latches the last value for a timeout. After the timeout it would remove the speed limit.

Sorry for lack of full context. The idea is that we stream always disregarding whether there is limit or not. If speed or yaw rate is not limited we stream **NAN** in corresponding field. By having this approach we can somehow guarantee that the limits cannot be taken off accidentally by broken stream.
Given the above, losing stream is not expected.

>There is only a problem if the thing doing the streaming MUST know that it succeeded or not. I don't see that as likely - what's it going to do in that case? If you did need that you could stream out a message "current speed limit" - it wouldn't have to be high rate.

I think it's a very valid point. We will discuss it internally

message_definitions/v1.0/common.xml
hamishwillee, line 7258:
```suggestion
    <message id="354" name="SET_VELOCITY_LIMITS">
```

message_definitions/v1.0/common.xml
```diff
@@ -7255,6 +7255,14 @@
...
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
```
hamishwillee, line 7261:
```suggestion
      <description>Set temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
```

hamishwillee:
Thanks for coming to the mav call. Just to summarize the conclusions:

- Valid use case and using a message (rather than command) makes sense.
- This needs a message to both get and set the limits. They should more or less mirror each other. This would allow a UI to be able to report when speed is limited, in particular when it is not doing the limiting.
- Streaming rates, timeouts, and whether or not they stream at the same rate would depend on the use case. 

Further
- James unsure about whether or not limits should apply in failsafes. I think we need to test those edge cases and perhaps state in the message. [my take is that speed limiting should turn off - the failsafe should be designed to always do the right thing re people so RTL or other failsafe should be safe. However it does depend a bit - if the failsafe is set to warning not sure].
- @marcinolokk 
  - Can you confirm you really want to stream the setter all the time? That makes no sense to me in terms of bandwidth usage when you might not be limiting speed. Or to put it another way, I think that it is OK to stream the message all the time, but I don't think it is reasonable to force that on all implementers for all use cases - they might have crap radios.

- **JUST OCCURED TO ME** The setter needs a target system and target component field. Otherwise you'll be speed limiting every system on your network. Not sure about the response from the flight stack - probably though.


Finally, with the above taken into consideration this can go into development.xml with both set and get versions as "agreeable in principle".

Hope that makes sense.

marcinolokk:
First of all thanks as well for the call and your input to the work. I changed msg to `SET...` and moved it to `development.xml`. Created another message named `VELOCITY_LIMITS` which seemed more intuitive than `GET_VELOCITY_LIMITS` .

>Can you confirm you really want to stream the setter all the time? That makes no sense to me in terms of bandwidth usage when you might not be limiting speed. Or to put it another way, I think that it is OK to stream the message all the time, but I don't think it is reasonable to force that on all implementers for all use cases - they might have crap radios.

Yes, in our case we would like to stream the `SET_VELOCITY_LIMITS` message all the time. However, the message itself should not enforce other users to use it the same way. One could implement other mechanism to handle it with potentially different failsafe mechanism. Unless I missed something.

>JUST OCCURED TO ME The setter needs a target system and target component field. Otherwise you'll be speed limiting every system on your network. Not sure about the response from the flight stack - probably though.

Hmm that's a valid point. So ideally setter would have: target_comp_id and target_sys_id. Getter would have: comp_id, sys_id ? @MaEtUgR  what's your take on it? I think we also didn't consider this.

message_definitions/v1.0/development.xml
```diff
@@ -429,6 +429,22 @@
       <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.</field>
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
     </message>
+    <message id="354" name="SET_VELOCITY_LIMITS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Set temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
+      <field type="float" name="horizontal_speed_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME_LOCAL_NED. NaN: Field not used (ignore)</field>
```
hamishwillee, line 436:
```suggestion
      <field type="uint8_t" name="target_system">System ID (0 for broadcast).</field>
      <field type="uint8_t" name="target_component">Component ID (0 for broadcast).</field>
      <field type="float" name="horizontal_speed_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME_LOCAL_NED. NaN: Field not used (ignore)</field>
```

hamishwillee:
Thanks for moving the messages to development.xml, and yes, the names are good for the setter/getter.

> Hmm that's a valid point. So ideally setter would have: target_comp_id and target_sys_id. Getter would have: comp_id, sys_id ? @MaEtUgR what's your take on it? I think we also didn't consider this.

The setter needs `target_system` and `target_component` as suggested [here](https://github.com/mavlink/mavlink/pull/2015/files#r1316513869). The names matter because there is a helper method in the C library that pulls them out of the payload. This allows the sender to specify a particular system/component must obey the rules.

I am not so sure about the getter. If you don't add these fields the current speed limits for the vehicle are published to all systems on the network. If you add the fields you can choose to just send it to the thing setting the limits.
Do we think that there is a case where we might only want the setter vehicle to get this information? If so we would add them, otherwise we might add them if we are not sure.

hamishwillee:
@marcinolokk @MaEtUgR Thoughts? ^^^

MaEtUgR:
I added the `target_system` and `target_component` fields to the setter message as suggested. Good catch that's really necessary and easy to forget after working with network protocols where source and destination of each packet are clear by design. For the feedback/telemetry message I would not add a target since it's not an acknowledgement targeted at any particular participant and the source is luckily part of the header: https://mavlink.io/en/guide/serialization.html#mavlink2_packet_format

message_definitions/v1.0/development.xml
```diff
@@ -429,6 +429,24 @@
       <field type="uint16_t" name="data_rate" units="MiB/s">WiFi network data rate. Set to UINT16_MAX if data_rate information is not supplied.</field>
       <field type="uint8_t" name="security" enum="WIFI_NETWORK_SECURITY">WiFi network security type.</field>
     </message>
+    <message id="354" name="SET_VELOCITY_LIMITS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Set temporary limits for horizontal speed, vertical speed and yaw rate to slow down the vehicle from the default configuration during operation. The consumer implements a timeout and returns to default limits in case the stream is lost. Accelerations are automatically adjusted accordingly.</description>
```
hamishwillee, line 435:
```suggestion
      <description>Set temporary maximum limits for horizontal speed, vertical speed and yaw rate.
        The consumer should stream the current limits in VELOCITY_LIMITS (when limits are being set).
        The consumer implements a timeout and returns to default limits in case the stream is lost.
        Accelerations are automatically adjusted accordingly.
      </description>
```

message_definitions/v1.0/development.xml
```diff
@@ -429,6 +429,24 @@
...
+      <field type="uint8_t" name="target_system">System ID (0 for broadcast).</field>
+      <field type="uint8_t" name="target_component">Component ID (0 for broadcast).</field>
+      <field type="float" name="horizontal_speed_limit" default="NaN" units="m/s">Limit for horizontal movement in MAV_FRAME_LOCAL_NED. NaN: Field not used (ignore)</field>
+      <field type="float" name="vertical_speed_limit" default="NaN" units="m/s">Limit for vertical movement in MAV_FRAME_LOCAL_NED. NaN: Field not used (ignore)</field>
+      <field type="float" name="yaw_rate_limit" default="NaN" units="rad/s">Limit for vehicle turn rate around its yaw axis. NaN: Field not used (ignore)</field>
+    </message>
+    <message id="355" name="VELOCITY_LIMITS">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Current limits for horizontal speed, vertical speed and yaw rate.</description>
```
hamishwillee, line 445:
```suggestion
      <description>Current limits for horizontal speed, vertical speed and yaw rate, as set by SET_VELOCITY_LIMITS.</description>
```

hamishwillee:
@MaEtUgR I added a couple of suggestions that I'd like merged to "cross link" the messages.
Other than that this is sufficient for merging now. Before we merge into common.xml we'll need the prototype and some thinking around the failsafe behaviour.

message_definitions/v1.0/development.xml
MaEtUgR: > The consumer implements a timeout and returns to default limits in case the stream is lost.

We currently don't have it that way because of the use case where someone is inspecting e.g. a power line and relies to go slow and if the limit is suddenly revoked because of a connection loss it might unexpectedly go fast and crash into something. It's merely a stream to be able to adjust the limit continuously/dynamically e.g. based on camera zoom or a remote knob. We keep the last limit until the user switches mode.

> Accelerations are automatically adjusted accordingly.

That's the case in our implementation. I guess it doesn't hurt to mention this since otherwise it might get forgotten by the developer of any new implementation.

hamishwillee: I have modified my proposal. I didn't like the accelerations comment. Note, this now states that the values are latched.

hamishwillee (approved):
Awesome. Thanks for your patience. This looks good to me.

marcinolokk:
Awesome. Thanks @MaEtUgR for taking care of the pr when I was away. I will merge it then

PR #2019, opened 2023-07-12
Extend MAV_COLLISION_ACTION enum to add MAV_COLLISION_ACTION_BRAKE.

Tal-Bass:
This is in support of ADSB_AVOID improvements mentioned in the following PR:
https://github.com/ArduPilot/ardupilot/pull/24284

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3552,6 +3552,9 @@
       <entry value="6" name="MAV_COLLISION_ACTION_HOVER">
         <description>Aircraft to stop in place</description>
       </entry>
+      <entry value="7" name="MAV_COLLISION_ACTION_BRAKE">
+        <description>Aircraft to switch to brake mode</description>
+      </entry>
     </enum>
     <enum name="MAV_COLLISION_THREAT_LEVEL">
       <description>Aircraft-rated danger from this threat.</description>


message_definitions/v1.0/common.xml
```diff
@@ -3552,6 +3552,9 @@
       <entry value="6" name="MAV_COLLISION_ACTION_HOVER">
         <description>Aircraft to stop in place</description>
       </entry>
+      <entry value="7" name="MAV_COLLISION_ACTION_BRAKE">
+        <description>Aircraft to switch to brake mode</description>
```
hamishwillee, line 3556: I would interpret this to mean "Aircraft to stop in place" - so how does brake mode differ from MAV_COLLISION_ACTION_HOVER?

Or to put it another way, implementers and recipients need a description that unambiously makes it clear what this will do.

Tal-Bass: MAV_COLLISION_ACTION_HOVER will switch the vehicle mode to Loiter whereas MAV_COLLISION_ACTION_BRAKE optionally descends to a safe altitude and then switches to Brake mode. Would this be better:
MAV_COLLISION_DESCEND_AND_STOP - "Aircraft to descend to a safe altitude before stopping in place".
Ideally I'd like the user to know that it switches to Brake instead of Loiter.

hamishwillee: I kind of think that `MAV_COLLISION_ACTION_ASCEND_OR_DESCEND` allows for this - i.e. 
- can the vehicle make a smarter decision about the best action than the user? If not then ascend/descend is better
- Can the autopilot make a smart decision to always land/hover? if so, then MAV_COLLISION_ACTION_ASCEND_OR_DESCEND can be implemented this wayl.

So when would it make sense to specify "you must always descend and hover"?

That said, yes, this approach would be better, but you probably need to consider what you want to happen if a vehicle does not have a GPS and cannot hold position, only altitude.

auturgy: BRAKE Mode is defined in ArduPilot, however agreed that this isn't in the Standard modes.  I think we need to wait for additional review of the proposed ArduPilot PR to gain some greater insight into the implementation challenges presented by this approach - I suspect that there'll be some additional work required to refine the API.

hamishwillee: Sounds good to me. I'm taking eyes off until that PR progresses. Wake me up when that has progressed.

Tal-Bass: Stating the description in terms of a standard mode sounds good to me. In the case of no GPS, altitude hold also sounds reasonable. Also the description for the behavior should actually be in the opposite order:
1. Descend to a safe height as defined by the flight stack. Followed by:
2. An aggressive position hold if it is supported, otherwise an aggressive altitude hold.

I do see situations in that it is better to descend/hover and never ascend. If there is an incoming aircraft that is not transmitting ADSB or is transmitting it incorrectly, then the user may want to trigger this avoidance behavior. In this case, there is no information on the incoming aircraft's altitude so it is safer to descend as rapidly as possible. Perhaps this could be an optional parameter to enable/disable ascending.

Happy to pick this up if/when the ArduPilot PR gets discussed.

hamishwillee: Cool. The order above does not make sense to me, possibly because of the use of the term "aggressive". An aggressive break is a position hold where you "really want to stop as soon as possible". There is no point doing that after descending - you're already likely slowed or stopped unless forward thrust was also applied in the descent.

So perhaps you mean "descend then hold position (or altitude if you can't)" - no aggression involved :-)

hamishwillee:
Closing as stalled.

@Tal-Bass I'm in no way opposed to this, but the discussion has stalled and the linked PR has no interest from ArduPilot. MAVLink doesn't accept things as standard unless there is an implementation and potential interest in others to standardize. 
Feel free to reopen if you can get ArduPilot (or PX4, or another flight stack) to show interest in adoption.

PR #2024, opened 2023-07-26
common:  Add camera instance field to MAV_CMD_IMAGE_START/STOP_CAPTURE

khanasif786:
I am working on enhancing camera and gimbal interface in Ardupilot, We are adding support of multiple cameras so a camera instance field is required for telling the camera in which we want to start and stop capturing images.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2117,8 +2117,21 @@
         <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
-        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.</description>
-        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
+        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.
+
+          Param1 (id) may be used to specify the target camera: 0: all cameras, 1 to 6: autopilot-connected cameras, 7-255: MAVLink camera component ID.
+          It is needed in order to target specific cameras connected to the autopilot, or specific sensors in a multi-sensor camera (neither of which have a distinct MAVLink component ID).
+          It is also needed to specify the target camera in missions.
+
+          When used in a mission, an autopilot should execute the MAV_CMD for a specified local camera (param1 = 1-6), or resend it as a command if it is intended for a MAVLink camera (param1 = 7 - 255), setting the command's target_component as the param1 value (and setting param1 in the command to zero).
+          If the param1 is 0 the autopilot should do both.
+          
+          When sent in a command the target MAVLink address is set using target_component.
+          If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
+          If addressed to a MAVLink camera, param 1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
+          If the command is broadcast (target_component is 0) then param 1 should be set to 0 (any other value should be NACKED with MAV_RESULT_DENIED). An autopilot would trigger any local cameras and forward the command to all channels.
+        </description>
+        <param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras that don't have a distinct component id (such as autopilot-attached cameras). 0: all cameras. This is used to specifically target autopilot-connected cameras or individual sensors in a multi-sensor MAVLink camera. It is also used to target specific cameras when the MAV_CMD is used in a mission</param>
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Total Images" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
         <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1), otherwise set to 0. Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2118,7 +2118,7 @@
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Total Images" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
         <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1), otherwise set to 0. Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted.</param>
@@ -2128,7 +2128,7 @@
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
         <description>Stop image capture sequence Use NaN for reserved values.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
         <param index="2" reserved="true" default="NaN"/>
         <param index="3" reserved="true" default="NaN"/>
         <param index="4" reserved="true" default="NaN"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2117,7 +2117,7 @@
         <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
-        <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
+        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.</description>
         <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Total Images" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2140,8 +2140,21 @@
         <param index="7" reserved="true" default="NaN"/>
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
-        <description>Stop image capture sequence Use NaN for reserved values.</description>
-        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
+        <description>Stop image capture sequence.
+        
+          Param1 (id) may be used to specify the target camera: 0: all cameras, 1 to 6: autopilot-connected cameras, 7-255: MAVLink camera component ID.
+          It is needed in order to target specific cameras connected to the autopilot, or specific sensors in a multi-sensor camera (neither of which have a distinct MAVLink component ID).
+          It is also needed to specify the target camera in missions.
+
+          When used in a mission, an autopilot should execute the MAV_CMD for a specified local camera (param1 = 1-6), or resend it as a command if it is intended for a MAVLink camera (param1 = 7 - 255), setting the command's target_component as the param1 value (and setting param1 in the command to zero).
+          If the param1 is 0 the autopilot should do both.
+
+          When sent in a command the target MAVLink address is set using target_component.
+          If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
+          If addressed to a MAVLink camera, param1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
+          If the command is broadcast (target_component is 0) then param 1 should be set to 0 (any other value should be NACKED with MAV_RESULT_DENIED). An autopilot would trigger any local cameras and forward the command to all channels.
+        </description>
+        <param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras that don't have a distinct component id (such as autopilot-attached cameras). 0: all cameras. This is used to specifically target autopilot-connected cameras or individual sensors in a multi-sensor MAVLink camera. It is also used to target specific cameras when the MAV_CMD is used in a mission</param>
         <param index="2" reserved="true" default="NaN"/>
         <param index="3" reserved="true" default="NaN"/>
         <param index="4" reserved="true" default="NaN"/>


rmackay9:
@hamishwillee @Davidsastresas @julianoes,

How do you guys feel about this?

I suspect that this first field has actually been reserved for this purpose similar to the [MAV_CMD_DO_CONTROL_VIDEO](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_CONTROL_VIDEO) message's "Camera ID" field.

I think what we should clarify whether the meanings (if any) for "-1" and "0".  So for example, "-1" could be "all", "0" could be primary (?), 1 could be 1st camera, 2 could be the 2nd camera, etc.  I guess it should be similar to the gimbal device id as it appears the [CAMERA_INFORMATION](https://mavlink.io/en/messages/common.html#CAMERA_INFORMATION) message.

message_definitions/v1.0/common.xml
```diff
@@ -2118,7 +2118,7 @@
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <param index="1" label="Camera Instance">Instance number of the camera on which image capturing is to be started.</param>
```
rmackay9, line 2121: The label should perhaps be "ID" to match MAV_CMD_DO_CONTROL_VIDEO.  It should perhaps also have "min=-1", "increment=1".

hamishwillee:
This is not the first time the issue has come up. Another suggestion was in https://github.com/mavlink/mavlink/pull/1749, which was to select the set of cameras for which subsequent commands apply. It was closed - not sure of the reason, but I do remember some of the concerns and suggestions.

My take on this is that specifying the ID makes sense but with some caveats:

- You can really only specify all or one particular camera - this approach means you can't specify cameras, 1 and 3 and 77 (say)
- If [MAV_CMD_IMAGE_START_CAPTURE](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE) (or stop) is sent as a _command_ it goes to the specific system and component. 
  - You would need to resolve how the ID works in this case vs source component
  - The reserved value for param 1 is 0. This must still be zero for the "default case" otherwise you will get unintended behaviour if sent from a GCS that things 0 mean nothing.
- If MAV_CMD_IMAGE_START_CAPTURE is used in a mission, currently how it behaves in the case of multiple cameras is undefined in the spec, but I think "effectively defined" by routing behaviour. 
  - Specifically I think that for this to work the command has to be sent EITHER with a broadcast component ID so that it turns on all cameras OR with the first default camera component ID and can only turn on that specific camera. The second approach is rubbish because component IDs should not imply "purpose".
  - Upshot - 0 has to mean "enable all cameras", at least in the mission case.
- Currently all addressing is via component ID in the command. How would a camera know that it was the primary camera, first camera, whatever? All they know about themselves is their component ID and system ID. It is best to avoid having to manually apply this so you'd need to either have a camera manager that assigns and ID or similar, or just use the component ID.
- Given that addressing is via a component ID, this means you can't have more than one camera direct-connected to the autopilot (same reasons as for gimbal).
- Cameras getting this won't know what to do with it without an update - i.e. in a mission you send this out. A recipient will get the message with broadcast id and know it can act on it. A current camera will act on it.
- There was discussion about addressing cameras via purpose - i.e. specify "capture on IR camera", "Capture on primary", "Capture on camera property". That requires integration work and a lot more thought.

I think this might work, given above caveats. The benefit is that things know their own ID and you can map the identity from camera information to component ID in a GCS (so nothing else needs to know the component ID in advance). Downside is that it will require update to firmware on cameras to understand that they need to do this in missions.

```
     <param index="1" label="target_component" minValue="0" maxValue="255" increment="1">Target camera component ID when sent in missions. 0: all cameras. Ignore if sent using the command protocol.</param>
```

The only other way I can see of managing this is to use some kind of camera manager, in the same way as gimbal manager. Using that approach the gimbal manager could re-route a command to a gimbal with the correct component ID.

Let's see what @Davidsastresas and @julianoes say - Julian is overseas so might take a while to reply.

rmackay9:
Hi @hamishwillee, thanks for the detailed feedback.

- I think handling sets of cameras is a step too far that this point at least.  The same thing can be accomplished with multiple mavlink or mission commands.
- If mav-cmd-image-start-capture is sent as a command to a mavlink enabled camera then I agree it should ignore the ID and just always process it.  I think this is true whether the message is sent directly (e.g. the sysid/compid are set to specifically address the camera) or sent as a broadcast (e.g. sysid=0, compid=0)
- In the case of a mission (executed on an autopilot) I think the autopilot should send the command directly to the mavlink enabled camera with the sysid/compid set.  The camera would ignore the "ID" column as mentioned above.

BTW, let's recognise that most cameras are not mavlink enabled.  Most are much simpler and just triggered via a PWM signal.  In these most common cases, the autopilot is acting like the "camera manager" you mention.

hamishwillee:
Hi @rmackay9 

I'm on other work today, so fly by ...

Yes to all.

In the case of a mission the thing that is executing the mission (autopilot) should take the ID from the camera mission item and generate an appropriately addressed camera command if using a MAVLink camera.

Do we need to think about the multi-camera-attached to flight controller case? I.e. if the autopilot sees 1-6 (say) in the id parameter it implicitly maps these to autopilot-attached cameras?

> BTW, let's recognise that most cameras are not mavlink enabled. Most are much simpler and just triggered via a PWM signal. > In these most common cases, the autopilot is acting like the "camera manager" and this is when the new ID field is particularly useful.

Yes and no. I have never seen a mavlink enabled camera, but most of the setups I see are companion computer centric, and the camera is abstracted by a camera manager. We need to support both cases. Of course you know that :-)

rmackay9:
@hamishwillee,

thanks again.

> Do we need to think about the multi-camera-attached to flight controller case? I.e. if the autopilot sees 1-6 (say) in the id parameter it implicitly maps these to autopilot-attached cameras?

Yes, in ArduPilot at least, the user has to set parameters to specify the type of each camera attached.  So for AP at least there is already an "instance" number for each camera so we will just do a mapping from this parameter value to the instance number.

hamishwillee:
@rmackay9 Just to be clear, is that "instance number" something we can just label as 1-6 in exactly the same way as gimbals. So I'm thinking.

```
<param index="1" label="target_component" minValue="0" maxValue="255" increment="1">Target camera component ID when sent in missions. 1-6: camera instance ID for autopilot-connected cameras. 0: all cameras. Ignore if sent using the command protocol.</param>
```

rmackay9:
@hamishwillee,

yes, that's what I'm thinking as well.  It's really the same issue but affecting cameras instead of gimbals.

Davidsastresas:
If I understood correctly, you guys are proposing leaving that field 1-6 for camera id when the camera is managed by Autopilot, because the rest of the time the command will make it to the camera because it will have reported using its component ID, so the commands will arrive to it directly, correct?

I like that approach. And that field, with 1-6 values, can even be useful in the case of mavlink enabled cameras with several sensors. I am thinking of a camera where you have a visible and multispectral sensor for instance. If that camera is mavlink enabled, it could understand which sensor needs to be triggered that way.

Until now, for mavlink cameras, that is managed by the extended parameter protocol. With parameters we tell the camera which sensor or group of sensors we want to trigger when we send that command. This is how Workswell cameras work for instance ( a good example of fully compliant mavlink camera ). With this approach we can also specify if that command should trigger all cameras at once. So in short, I don't think this will be an issue ever for mavlink cameras, unless we are programming a mission where we want some times to trigger one sensor and other times just other sensor, which is very unlikely I think.

So considering all the above, I think this is a change that is only important for cameras managed by an autopilot, and I think it makes sense to use 1-6 for Autopilot managed cameras.

About the possibility to mean sensor ID for mavlink cameras, maybe it is more involved, maybe for that we would need something like what was proposed in https://github.com/mavlink/mavlink/pull/1749, or maybe at some point we should rework camera protocol similar to how gimbal manager v2 work. At this point I think it would be overkill, and I think we need to see over time how this gimbal manager v2 plays with the existing systems. I am sure it will evolve. Maybe at some point gimbal manager and camera manager can be "merged", in the end of the day most use cases will have the camera attached to a gimbal.

So, in short, I am good with this change for now. I don't think it breaks anything, it solves a problem, and I don't think it makes sense to make something much more elaborated at this point. I am seeing lately a lot of movement around gimbal/camera manufacturers, so I am sure all of this will evolve a lot in the near future.

Thanks for keeping me updated!

message_definitions/v1.0/common.xml
```diff
@@ -2118,7 +2118,7 @@
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
```
hamishwillee, line 2121: Note, keeping name as Id in case we want to use it for sensors and such later.
```suggestion
        <param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID when sent in missions (camera component id, or 1 to 6 for autopilot-connected cameras that don't have a distinct component id). 0: all cameras. Ignore when using the command protocol.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -2128,7 +2128,7 @@
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
         <description>Stop image capture sequence Use NaN for reserved values.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
```
hamishwillee, line 2131:
```suggestion
        <param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID when sent in missions (camera component id, or 1 to 6 for autopilot-connected cameras that don't have a distinct component id). 0: all cameras. Ignore when using the command protocol.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -2118,7 +2118,7 @@
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
         <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
```
hamishwillee, line 2120: NAN not used for reserved values in all case, so removing that. Adding explanation of the id.


```suggestion
        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.</description>
```

message_definitions/v1.0/common.xml
rmackay9: @hamishwillee, thanks very much for the fixup in the wording.  I see now that "0" is all while before "-1" was all.  I think this is a good change.

We should make this change in the MAV_CMD_IMAGE_STOP_CAPTURE as well.

By the way, I'm sure @khanasif786 would be happy for you to just push over top of his PR.

hamishwillee:
> If I understood correctly, you guys are proposing leaving that field 1-6 for camera id when the camera is managed by Autopilot, because the rest of the time the command will make it to the camera because it will have reported using its component ID, so the commands will arrive to it directly, correct? In the case of missions, the autopilot will be responsible to re route this command ( autopilot should have then a parameter or something to assign camera instances to compid or something like that ).

@Davidsastresas Yes.

Note that I was being too mission-centric by making statements about the field like "not used in command protocol.". If an autopilot can have have up to six attached cameras the command protocol needs to be able to select those.

```
<param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID. A MAVLink camera component ID, 1 to 6 for autopilot-attached cameras, 0: for all cameras.
This is used in missions to specify a target camera for the mission item.
For MAVLink cameras the autopilot should resend the item as a MAVLink command, setting the target_component to this value (and param1 to 0).
When sent in a command, the field is used to address autopilot connected-cameras only (1-6) and should otherwise be set to zero.</param>
```

It is desirable to be able to address individual sensors on a MAVLink camera, but this is a problem with just one field - because in a mission we need this to mean "MAVLink address of the target".  There is no way to say in missions "this camera and this sensor".

I worry that if we enable addressing sensors on mavlink cameras for commands someone will insist we work out how to do it for missions too - and I don't think we can do this without an extra field. But if we want this then we could word it like:

```
<param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera/sensor ID.
This is used in missions to specify a target camera for the mission item: a MAVLink camera component ID, 1 to 6 for autopilot-attached cameras, 0: for all cameras.
For MAVLink cameras the autopilot should resend the item as a MAVLink command, setting the target_component to this value, and param1 to 0).
When sent in a command the address of the camera component is specified in target_component. This field allows you to specify a particular autopilot-connected camera (1-6) or all cameras (0). More generally, you can use it to address a particular camera sensor on a MAVLink camera that has more than one camera sensor.</param>
```

All a bit frustrating.

hamishwillee:
@Davidsastresas PS Do you know any cameras other than [Workswell](https://my.workswell.eu/link/latest/16154771971WK4XgHicukQzBYHzL4o) that implement the protocol?

message_definitions/v1.0/common.xml
hamishwillee: @rmackay9 Sorry, been thinking about this for ages and crossed over. The problem is that in command protocol you do want to be able to set 1-6. (and this says missions only)

At that point you need to decide if, when this was sent to a MAVLink camera, the value should be ignored or used to select a specific sensor. Rambling discussion in https://github.com/mavlink/mavlink/pull/2024#issuecomment-1661314890.  If we are sure of the approach we want, yes, will roll out to the stop case as well.

rmackay9: @hamishwillee,

I agree with that wording below.  By the way, using the field to hold the instance (e.g. 1~6) is useful not only when doing missions but when using real-time/immediate commands where the camera is not mavlink enabled.  It's also even useful even if the camera is mavlink enabled and the GCS just can't be bothered trying to decide whether it can talk directly to the camera or not.. it can just always talk to the autopilot.

hamishwillee: That's true, but hard to capture "generically".

Worse I'm still not sure we have this all correct for the autopilot connected cameras, because we are mixing the concepts that worked for gimbal, which has a manager.

For a start, how do you know what cameras are available? For a MAVLink camera you look for HEARTBEAT, from that you can get a `MAV_TYPE` and then query for CAMERA_INFORMATION. That won't work for autopilot-connected cameras - because unlike gimbals there is no manager in the autopilot.

Davidsastresas:
> @Davidsastresas PS Do you know any cameras other than [Workswell](https://my.workswell.eu/link/latest/16154771971WK4XgHicukQzBYHzL4o) that implement the protocol?

I think PhaseOne is also implementing the protocol. I don't think I've worked directly with any other camera that uses it, so I am not 100% sure of any other.

About the missions matter, I really think for missions it could be handled in the Autopilot and/or the mavlink camera, using parameters for the main camera, whatever. I really think at this point it would be overkill to elaborate a camera manager. I mean, it will be a good idea at some point, but I feel we need to test more the idea with gimbal manager before extending it further as it is, don't you think?

message_definitions/v1.0/common.xml
hamishwillee: OK, @rmackay9 @Davidsastresas OK. Modified proposal. It's the same thing but it clarifies the edge cases:
- command NACK if you specify an invalid param value such as ID for non-existing local camera. 
- A broadcast command should set param1 to zero/treat param 1 as zero: if you are sending to every camera, you shouldn't be setting a specific sensor anyway.

Also moved the detail to description, since there is quite a lot to say.
```suggestion
        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.

          Param1 (id) may be used to specify the target camera: 0: all cameras, 1 to 6: autopilot-connected cameras, 7-255: MAVLink camera component ID.
          It is needed in order to target specific cameras connected to the autopilot, or specific sensors in a multi-sensor camera (neither of which have a distinct MAVLink component ID).
          It is also needed to specify the target camera in missions.

          When used in a mission, an autopilot should execute the MAV_CMD for a specified local camera (param1 = 1-6), or resend it as a command if it is intended for a MAVLink camera (param1 = 7 - 255), setting the command's target_component as the param1 value (and setting param1 in the command to zero).
          If the param1 is 0 the autopilot should do both.
          
          When sent in a command the target MAVLink address is set using target_component.
          If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
          If addressed to a MAVLink camera, param 1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
          If the command is broadcast (target_component is 0) then param 1 should be set to 0 (any other value should be NACKED with MAV_RESULT_DENIED). An autopilot would trigger any local cameras and forward the command to all channels.
        </description>
```

message_definitions/v1.0/common.xml
```diff
@@ -2117,8 +2117,8 @@
         <param index="7" label="Gimbal device ID">Component ID of gimbal device to address (or 1-6 for non-MAVLink gimbal), 0 for all gimbal device components. Send command multiple times for more than one gimbal (but not all gimbals).</param>
       </entry>
       <entry value="2000" name="MAV_CMD_IMAGE_START_CAPTURE" hasLocation="false" isDestination="false">
-        <description>Start image capture sequence. Sends CAMERA_IMAGE_CAPTURED after each capture. Use NaN for reserved values.</description>
-        <param index="1">Reserved (Set to 0)</param>
+        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.</description>
```
hamishwillee, line 2120: OK, @rmackay9 @Davidsastresas OK. Modified proposal. It's the same thing but it clarifies the edge cases:
- command NACK if you specify an invalid param value such as ID for non-existing local camera. 
- A broadcast command should set param1 to zero/treat param 1 as zero: if you are sending to every camera, you shouldn't be setting a specific sensor anyway.

Also moved the detail to description, since there is quite a lot to say.
```suggestion
        <description>Start image capture sequence. CAMERA_IMAGE_CAPTURED must be emitted after each capture.

          Param1 (id) may be used to specify the target camera: 0: all cameras, 1 to 6: autopilot-connected cameras, 7-255: MAVLink camera component ID.
          It is needed in order to target specific cameras connected to the autopilot, or specific sensors in a multi-sensor camera (neither of which have a distinct MAVLink component ID).
          It is also needed to specify the target camera in missions.

          When used in a mission, an autopilot should execute the MAV_CMD for a specified local camera (param1 = 1-6), or resend it as a command if it is intended for a MAVLink camera (param1 = 7 - 255), setting the command's target_component as the param1 value (and setting param1 in the command to zero).
          If the param1 is 0 the autopilot should do both.
          
          When sent in a command the target MAVLink address is set using target_component.
          If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
          If addressed to a MAVLink camera, param 1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
          If the command is broadcast (target_component is 0) then param 1 should be set to 0 (any other value should be NACKED with MAV_RESULT_DENIED). An autopilot would trigger any local cameras and forward the command to all channels.
        </description>
```

message_definitions/v1.0/common.xml
hamishwillee: Simplifying now description is updated.

```suggestion
        <param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras that don't have a distinct component id (such as autopilot-attached cameras). 0: all cameras. This is used to specifically target autopilot-connected cameras or individual sensors in a multi-sensor MAVLink camera. It is also used to target specific cameras when the MAV_CMD is used in a mission</param>
```

hamishwillee:
@Davidsastresas @rmackay9 

Thanks. I have updated the description and param1 for one mission above and merged. It is essentially the same thing we already discussed but "spelled out". The particularly important edge case is that if the commands are broadcast, then they should not set param1 as anything other than 0 (not that I think you would want to).

If you are OK with this I would roll them into the other MAV_CMD. I would still like @julianoes to comment. Not sure if we are ready to merge after that, but you should have enough to verify/prototype.

> About the missions matter, I really think for missions it could be handled in the Autopilot and/or the mavlink camera, using parameters for the main camera, whatever. I really think at this point it would be overkill to elaborate a camera manager. I mean, it will be a good idea at some point, but I feel we need to test more the idea with gimbal manager before extending it further as it is, don't you think?

Yes. I would much prefer to solve this without having to invent camera manager concepts. But I would prefer to be sure that we don't preclude them. 

The cost of not addressing this now is that we don't know what cameras area directly attached to the autopilot.

Thank you for the camera information. I will add to the camera microservice docs.

rmackay9 (approved):

julianoes (approved):
I think this makes sense, thanks for working it all out!

message_definitions/v1.0/common.xml
```diff
@@ -2128,7 +2141,7 @@
       </entry>
       <entry value="2001" name="MAV_CMD_IMAGE_STOP_CAPTURE" hasLocation="false" isDestination="false">
         <description>Stop image capture sequence Use NaN for reserved values.</description>
```
hamishwillee, line 2143: Roll out same info as above to the "stop" image capture sequence. I suspect that generally you'll just broadcast stop, but this allows complementary flexibility to starting capture.

```suggestion
        <description>Stop image capture sequence.
        
          Param1 (id) may be used to specify the target camera: 0: all cameras, 1 to 6: autopilot-connected cameras, 7-255: MAVLink camera component ID.
          It is needed in order to target specific cameras connected to the autopilot, or specific sensors in a multi-sensor camera (neither of which have a distinct MAVLink component ID).
          It is also needed to specify the target camera in missions.

          When used in a mission, an autopilot should execute the MAV_CMD for a specified local camera (param1 = 1-6), or resend it as a command if it is intended for a MAVLink camera (param1 = 7 - 255), setting the command's target_component as the param1 value (and setting param1 in the command to zero).
          If the param1 is 0 the autopilot should do both.

          When sent in a command the target MAVLink address is set using target_component.
          If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
          If addressed to a MAVLink camera, param1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
          If the command is broadcast (target_component is 0) then param 1 should be set to 0 (any other value should be NACKED with MAV_RESULT_DENIED). An autopilot would trigger any local cameras and forward the command to all channels.
        </description>
```

message_definitions/v1.0/common.xml
```diff
@@ -2128,7 +2141,7 @@
...
-        <param index="1">Reserved (Set to 0)</param>
+        <param index="1" label="ID" minValue="-1" increment="1">Camera ID (-1: all cameras, 0: primary camera, 1: first camera and so on)</param>
```
hamishwillee, line 2144:
```suggestion
        <param index="1" label="id" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras that don't have a distinct component id (such as autopilot-attached cameras). 0: all cameras. This is used to specifically target autopilot-connected cameras or individual sensors in a multi-sensor MAVLink camera. It is also used to target specific cameras when the MAV_CMD is used in a mission</param>
```

hamishwillee:
Thanks @julianoes .

OK, rolled out the changes. I'm happy with this but the rules say we prefer to keep things unmerged until prototyped. OK for us to do that and ping us once you're sure it works for the various cases?

nexton-winjeel:
This change will work, but don't you have to apply it to _every_ camera message? Not just `MAV_CMD_IMAGE_START_CAPTURE` and `MAV_CMD_IMAGE_STOP_CAPTURE`?

(I got here from https://github.com/ArduPilot/ardupilot/issues/24955)

rmackay9:
@nexton-winjeel,

Yes, we will need to go through as many as possible of the camera related messages and hopefully add the field to them as well.  Let's not let perfection be the enemy of the good though or we will just get stuck.

hamishwillee:
>> This change will work, but don't you have to apply it to every camera message? Not just MAV_CMD_IMAGE_START_CAPTURE and MAV_CMD_IMAGE_STOP_CAPTURE?

> Yes, we will need to go through as many as possible of the camera related messages and hopefully add the field to them as well. Let's not let perfection be the enemy of the good though or we will just get stuck.

FWIW this is the cost of deciding flight stacks don't have to map hardware connected components to be "real" mavlink components. Given that choice, this is the only logical approach I can see.

It would be nice to add this to the other commands, so at least we know which ones might be problematic. Perhaps once we're sure that this approach will work, via prototyping.

julianoes (approved):

hamishwillee:
PS We discussed briefly in MAV call last night. We're still happy with this. We don't require all the other messages be updated to merge this. We'll merge once there is a tested prototype (the rest of the team were happy to merge now, but I'm annoying).

rmackay9:
We've implemented this in AP now so I think this can go in now. https://github.com/ArduPilot/ardupilot/pull/24772

PR #2026, opened 2023-08-02
Fix c_library_vX build order

hamishwillee:
This attempts to fix the build order of the prebuilt MAVLink libraries.

The order in which things are built matters. This is due to a bug in mavgen, which means that if there are any shared enum values then they only only appear the top level library that is built.
In practise this means that if you build `ardupilotmega.xml` then `ardupilotmega/ardupilotmega.h` will contain the `mav_cmd` definitions, and `common/common.h` (which is nested) will not.

To ensure that every level of library is correct you therefore need to build top level libraries first, and then any libraries that they nest. In other words, ensure that the last time a library is built, is when it the top level library. 

This is the order to get the entire library, including everything in all.xml. IOt shows what is included by the file, and what the file includes (in brackets). So libraries must be build after all XML that includes them.
Essentially this should mostly work as long as common, standard, minimal are built in that order, after everything that includes common.xml.

```
all - incl everything ()
test - no incl (all)
storm32 - incl ardupilotmega (all)
ardupilotmega  - (all, storm32)
csAirLink - no incl - (ardupilotmega)
icarous - no incl (ardupilotmega, all)
uAvionix.xml - incl common (all, ardupilotmega)
cubepilot.xml - incl common (all, ardupilotmega)
ualberta - incl common (all)
paparazzi - incl common ()
ASLUAV - incl common (all)
matrixpilot - incl common ()
python_array_test - incl common (all)
development - incl common (all)
common - incl standard (all, ardupilotmega, ualberta, uAvionix, cubepilot, asluav, AVSSUAS, development, matrixpilot, paparazzi, python_array_test )
standard - incl minimal (common, all)
minimal - no imports (standard)
```

I tested this by building this library and then comparing to all, ardupilotmega, common, minimal as built independently. 
With this order the XML is identical except for the hashes. Since these only matter for identifying the top level file and are not affected by overlaying, that does not matter.

FYI @bkueng

### DIFF
--- a/scripts/update_c_library.sh
+++ b/scripts/update_c_library.sh
@@ -60,12 +60,22 @@ rm -rf $CLIBRARY_PATH/*
 
 # generate new c headers
 echo -e "\0033[34mStarting to generate c headers\0033[0m\n"
-generate_headers development $1
-generate_headers ardupilotmega $1
-generate_headers matrixpilot $1
+generate_headers all $1
 generate_headers test $1
+generate_headers storm32 $1
+generate_headers ardupilotmega $1
+generate_headers csAirLink $1
+generate_headers uAvionix $1
+generate_headers cubepilot $1
+generate_headers ualberta $1
+generate_headers paparazzi $1
 generate_headers ASLUAV $1
+generate_headers matrixpilot $1
+generate_headers python_array_test $1
+generate_headers development $1
 generate_headers standard $1
+generate_headers common $1
+generate_headers minimal $1
 mkdir -p $CLIBRARY_PATH/message_definitions
 cp message_definitions/v1.0/* $CLIBRARY_PATH/message_definitions/.
 echo -e "\0033[34mFinished generating c headers\0033[0m\n"
--- a/scripts/update_c_library.sh
+++ b/scripts/update_c_library.sh
@@ -73,8 +73,8 @@ generate_headers ASLUAV $1
 generate_headers matrixpilot $1
 generate_headers python_array_test $1
 generate_headers development $1
-generate_headers standard $1
 generate_headers common $1
+generate_headers standard $1
 generate_headers minimal $1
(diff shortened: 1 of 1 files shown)


hamishwillee:
@julianoes @auturgy I have tested this, and Beat has built/tested QGC with the prebuilt all.xml and ardupilotmega.xml. So it appears to be "good".

Any objections to me merging based on that?

julianoes (approved):

auturgy:
Can you expand on the testing conducted?

hamishwillee:
> Can you expand on the testing conducted?

@auturgy Of course.

If I build an XML file the C generator creates sub folders for each of the dialects, containing a files dialect_name.h and mavlink.h. A user should be able to import mavlink.h for a particular dialect folder and get its definitions. The dialect subfolders should be the same whether I build a top level XML or a nested dialect (except for build timestamps).

If that worked I could just build all.xml and get all dialects, but due to pymavlink not accepting the merge-enum bugfix it does not. So we fake this by building all the dialects.

To prove this works I built each of the dialects individually in their own tree, and compared to the result of building them "overlaid" in this order. So first I compare minimal to minimal. They are identical, so I know that you can import mavlink.h in minimal folder in either case and get the same defintions.
The I repeated for standard, then common and so on, just comparing the files at the same level folder as the same dialect (so if looking at common.xml I am not interested in minimal, because that has had its mav_cmds stripped).

Beat then tried this by building QGC with ardupilotmega.xml libraries and then again with all.xml. It worked. Note that this is not a great test because all.xml has no MAV_CMD or other enums. That means it does not strip out the MAV_CMDs in lower levels, because there are none that get merged.

I'm fairly confident based on these tests and "the logic". Certainly no worse than what is there now not working.

hamishwillee:
@auturgy Can I merge this? We'll be no worse off.

Issue #2023, opened 2023-07-24
Extending MISSION_ITEM_REACHED with more detailed status


sypaq-nexton:
Throwing this one out to get some input. Currently, the [`MISSION_ITEM_REACHED`](https://mavlink.io/en/messages/common.html#MISSION_ITEM_REACHED) message only tells us when we get to a mission item. It would be nice if we could get more detailed status while executing a mission item. For example:
* [`MAV_CMD_NAV_LOITER_TURNS`](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TURNS) could report:
     * number of turns completed; and 
     * total number or number remaining.
* [`MAV_CMD_NAV_LOITER_TIME`](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_LOITER_TIME) and [`MAV_CMD_NAV_RETURN_TO_LAUNCH`](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_RETURN_TO_LAUNCH) could report:
     * time elapsed; and 
     * total time or time remaining.

I think most mission commands would be covered by one of:
1. `count_completed` and `total_count`;
1. `time_elapsed` and `total_time`; or
1. `percentage_complete`.

What would be the best way to implement this? Extend the [`MISSION_ITEM_REACHED`](https://mavlink.io/en/messages/common.html#MISSION_ITEM_REACHED) message with new fields? Extend a different message? Or add a new message (e.g. `MISSION_ITEM_REACHED_STATUS`)?

hamishwillee:
Thanks @sypaq-nexton . The short answer is that this is not the first time that this has come up, bit I don't have an answer that satisfies me. I'll add to the dev call next week for discussion.

A few thoughts:
- The place I would think first to add additional information is [MISSION_CURRENT](https://mavlink.io/en/messages/common.html#MISSION_CURRENT) as it has both total number and sequence and is streamed - so you do have some kind of progress.
- Current count of loiter and percentage time of loiter should be easy to calculate. 
- I don't think you'd append these to an existing message. The problem is that there are so many things you might append to mission_current to provide status - each one adding to the size.
- If we did this I think I'd want to either have dedicated status messages for various events, with an enum to indicate the event and a field(s) to indicate the value for the event. So mission_update with enum loiter_turn 3/6. mission_update with loiter_time enum - value in seconds or whatever. 
- Lots of other possible events interfaces - a formatted STATUSTEXT. An `EVENT`.

[...]

PR #2029, opened 2023-08-17
common.xml: MISSION_CURRENT progress notification

hamishwillee:
This attempts to design a solution #2023 for reporting the progress in individual waypoints, in particular for loiter progress.

It adds `MISSION_CURRENT.progress` and `.progress_final` which are uint_16s that represent the progress and final progress to be reached. The units of these values depend on the mission item - I have added some possible values to waypoint and loiter waypoints.

- WAYPOINT - start distance, final distance in metres
  - Do we need this? The GCS can infer this easily from telemetry data for current position relative to initial/final position.
- MAV_CMD_NAV_LOITER_UNLIM -  progress and progress_final fields to 0.
  - I don't think we need anything here. The GCS knows it is loitering forever from the type. It knows when it is travelling to the waypoint and can infer when it has reached it from the type of vehicle and radius etc. We can measure the time taken so far, and there is no end point.
  - The only thing we might provide is number of turns already done - but since that isn't useful for progress, IMO not better to set to zero.
- MAV_CMD_NAV_LOITER_TIME -  progress and progress_final fields to 0.
  - Do we need progress for this? I've said no for same reason as previous case - a GCS can work it out from knowledge of the mission item and current vehicle position. 
  - The only exception is if you connect when the vehicle has already been waiting.
- MAV_CMD_NAV_LOITER_TURNS - number of turns completed, and total number of turns.
  - It is harder for GCS to estimate complete turns but it can actually work this out.
  - Again, only a problem if you connect after starting mission or lose telemetry.
  
@julianoes @sypaq-nexton So from above, for travel between points you can work out progress from knowledge of the mission and knowledge of current position of the vehicle. 
The exceptions are loiter time and loiter turns, which you can work out from GCS if you are always connected, but not if you lose telemetry at some point. 

The values can be zero'd if we expect the info to come from the GCS so no particular additional cost.

Anyway, have a think - is this "good". Are there other things that would need non-zero values, or be better if were non-zero values?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1083,7 +1083,7 @@
       <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data. NaN and INT32_MAX may be used in float/integer params (respectively) to indicate optional/default values (e.g. to use the component's current yaw or latitude rather than a specific value). See https://mavlink.io/en/guide/xml_schema.html#MAV_CMD for information about the structure of the MAV_CMD entries</description>
       <entry value="16" name="MAV_CMD_NAV_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint. This is intended for use in missions (for guided commands outside of missions use MAV_CMD_DO_REPOSITION).
-          MISSION_CURRENT should report the progress and progress_final fields as a distance in metres.
+          MISSION_CURRENT should report the item_progress_units as a distance in metres (and set other item_progress fields appropriately).
 	</description>
         <param index="1" label="Hold" units="s" minValue="0">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
@@ -1095,7 +1095,7 @@
       </entry>
       <entry value="17" name="MAV_CMD_NAV_LOITER_UNLIM" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint an unlimited amount of time.
-          MISSION_CURRENT should report the progress and progress_final fields as 0.
+          MISSION_CURRENT can report the fields item_progress_* fields as 0.
 	</description>
         <param index="1">Empty</param>
         <param index="2">Empty</param>
@@ -1107,7 +1107,7 @@
       </entry>
       <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint for X turns.
-          MISSION_CURRENT should report the progress and progress_final fields as number of turns completed, and total number of turns.
+          MISSION_CURRENT should report the item_progress_* fields as turns/laps.
         </description>
         <param index="1" label="Turns" minValue="0">Number of turns.</param>
         <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter circle only once heading towards the next waypoint (0 = False)</param>
@@ -1122,7 +1122,7 @@
           Multicopter vehicles stop at the point (within a vehicle-specific acceptance radius).
           Forward-only moving vehicles (e.g. fixed-wing) circle the point with the specified radius/direction.
           If the Heading Required parameter (2) is non-zero forward moving aircraft will only leave the loiter circle once heading towards the next waypoint.
-          MISSION_CURRENT should report the progress and progress_final fields 0.
+          MISSION_CURRENT should report item_progress_* fields as time in seconds.
 	</description>
         <param index="1" label="Time" units="s" minValue="0">Loiter time (only starts once Lat, Lon and Alt is reached).</param>
         <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter circle only once heading towards the next waypoint (0 = False)</param>
@@ -1310,7 +1310,9 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="93" name="MAV_CMD_NAV_DELAY" hasLocation="false" isDestination="false">
-        <description>Delay the next navigation command a number of seconds or until a specified time</description>
+        <description>Delay the next navigation command a number of seconds or until a specified time
+	  MISSION_CURRENT should report item_progress_* fields as time in seconds.
+	</description>
         <param index="1" label="Delay" units="s" minValue="-1" increment="1">Delay (-1 to enable time-of-day fields)</param>
         <param index="2" label="Hour" minValue="-1" maxValue="23" increment="1">hour (24h format, UTC, -1 to ignore)</param>
         <param index="3" label="Minute" minValue="-1" maxValue="59" increment="1">minute (24h format, UTC, -1 to ignore)</param>
@@ -1340,7 +1342,9 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="112" name="MAV_CMD_CONDITION_DELAY" hasLocation="false" isDestination="false">
-        <description>Delay mission state machine.</description>
+        <description>Delay mission state machine.
+	  MISSION_CURRENT should report item_progress_* fields as time in seconds.
+	</description>
         <param index="1" label="Delay" units="s" minValue="0">Delay</param>
         <param index="2">Empty</param>
         <param index="3">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1083,7 +1083,8 @@
       <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data. NaN and INT32_MAX may be used in float/integer params (respectively) to indicate optional/default values (e.g. to use the component's current yaw or latitude rather than a specific value). See https://mavlink.io/en/guide/xml_schema.html#MAV_CMD for information about the structure of the MAV_CMD entries</description>
       <entry value="16" name="MAV_CMD_NAV_WAYPOINT" hasLocation="true" isDestination="true">
         <description>Navigate to waypoint. This is intended for use in missions (for guided commands outside of missions use MAV_CMD_DO_REPOSITION).
-          MISSION_CURRENT should report the item_progress_units as a distance in metres (and set other item_progress fields appropriately).
+          MISSION_CURRENT may optionally report item_progress_* fields as time (s) or distance (m) when travelling towards waypoint. 		
+          MISSION_CURRENT should report item_progress_* fields as time in seconds if holding at waypoint.
 	</description>
         <param index="1" label="Hold" units="s" minValue="0">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1108,7 +1108,7 @@
       </entry>
       <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS" hasLocation="true" isDestination="true">
         <description>Loiter around this waypoint for X turns.
-          MISSION_CURRENT should report the item_progress_* fields as turns/laps.
+          MISSION_CURRENT should report the item_progress_* fields as a count of turns/laps.
         </description>
         <param index="1" label="Turns" minValue="0">Number of turns.</param>
         <param index="2" label="Heading Required" minValue="0" maxValue="1" increment="1">Leave loiter circle only once heading towards the next waypoint (0 = False)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1310,7 +1310,7 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="93" name="MAV_CMD_NAV_DELAY" hasLocation="false" isDestination="false">
-        <description>Delay the next navigation command a number of seconds or until a specified time
+        <description>Delay the next navigation command a number of seconds or until a specified time.
           MISSION_CURRENT should set the item_progress_remaining field as remaining delay time (s) while waiting.
 	</description>
         <param index="1" label="Delay" units="s" minValue="-1" increment="1">Delay (-1 to enable time-of-day fields)</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5302,8 +5302,9 @@
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle (on last item, sequence == total). If the autopilot stores its home location as part of the mission this will be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
       <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
       <field type="uint8_t" name="mission_mode" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
-      <field type="uint16_t" name="progress">Progress through current mission item. This is relative to progress_final. Units depend on type of current mission item (and should be defined in each item). 0 initially and if progress_final is 0.</field>
-      <field type="uint16_t" name="progress_final" invalid="0">Number of mission items in the sequence</field>
+      <field type="uint8_t" name="item_progress_units">0: not used. 1: distance (m), 2: time (seconds), 3: count (laps)</field>
+      <field type="uint16_t" name="item_progress">Progress through current mission item. This is relative to item_progress_final. Units depend on type of current mission item and are specified in item_progress_units. 0 initially and if progress_final is 0.</field>
+      <field type="uint16_t" name="item_progress_final" invalid="0">The final state that item_progress is expected to reach. 0 if item progress information not used.</field>
     </message>
     <message id="43" name="MISSION_REQUEST_LIST">
       <description>Request the overall list of mission items from the system/component.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1082,9 +1082,8 @@
     <enum name="MAV_CMD">
       <description>Commands to be executed by the MAV. They can be executed on user request, or as part of a mission script. If the action is used in a mission, the parameter mapping to the waypoint/mission message is as follows: Param 1, Param 2, Param 3, Param 4, X: Param 5, Y:Param 6, Z:Param 7. This command list is similar what ARINC 424 is for commercial aircraft: A data format how to interpret waypoint/mission data. NaN and INT32_MAX may be used in float/integer params (respectively) to indicate optional/default values (e.g. to use the component's current yaw or latitude rather than a specific value). See https://mavlink.io/en/guide/xml_schema.html#MAV_CMD for information about the structure of the MAV_CMD entries</description>
       <entry value="16" name="MAV_CMD_NAV_WAYPOINT" hasLocation="true" isDestination="true">
-        <description>Navigate to waypoint. This is intended for use in missions (for guided commands outside of missions use MAV_CMD_DO_REPOSITION).
-          MISSION_CURRENT may optionally report item_progress_* fields as time (s) or distance (m) when travelling towards waypoint. 		
-          MISSION_CURRENT should report item_progress_* fields as time in seconds if holding at waypoint.
+        <description>Navigate to waypoint. This is intended for use in missions (for guided commands outside of missions use MAV_CMD_DO_REPOSITION).		
+          MISSION_CURRENT should set the item_progress_remaining field as remaining hold time (s) if holding at waypoint.
 	</description>
         <param index="1" label="Hold" units="s" minValue="0">Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)</param>
         <param index="2" label="Accept Radius" units="m" minValue="0">Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)</param>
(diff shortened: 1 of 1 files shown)


AndreasAntener:
What about number of repetitions from `MAV_CMD_DO_JUMP`?

hamishwillee:
> What about number of repetitions from MAV_CMD_DO_JUMP?

It is true that this isn't exposed anywhere, and if you aren't always connected to the vehicle there is no way of guessing where you are at in the jump cycle without this kind of information. 

I can kind of imagine how you might use this to show a repeating route - ie mark it in some kind of separate colour and have a counter display near the start or end. A generalization of tracking turns. Maybe. Thanks.

AndreasAntener:
We would have wished for this a few times during testing, in endurance flights after lap 15 nobody knows the count anymore for sure ;)

But I realise now that, even tought your proposal is well generalised, the issue with the jump count is that you are not staying at the same waypoint. So the MISSION_CURRENT message might not even be sent out for it.

message_definitions/v1.0/common.xml
```diff
@@ -5283,12 +5294,16 @@
...
       </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle (on last item, sequence == total). If the autopilot stores its home location as part of the mission this will be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
       <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
       <field type="uint8_t" name="mission_mode" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
+      <field type="uint16_t" name="progress">Progress through current mission item. This is relative to progress_final. Units depend on type of current mission item (and should be defined in each item). 0 initially and if progress_final is 0.</field>
+      <field type="uint16_t" name="progress_final" invalid="0">Number of mission items in the sequence</field>
```
sypaq-nexton, line 5306: Can we be more explicit and name these new fields `item_progress` and `item_progress_final`? Otherwise, people who only read the names and not the descriptions might think these relate to the mission as a whole.

julianoes:
> After thinking more about this, the jump repeat count is more a mission level (or "lap") state, not a waypoint state.

I'm pretty sure when we first implemented goto on the PX4 side we made the mission count down, so if you were to download the mission from the vehicle, the count would go down accordingly. It's not pretty but it seemed a way to transmit that information. I'm not sure that's still how it works though, I'd have to test it. And we should probably note/spec it either way.

sypaq-nexton:
Thanks for this @hamishwillee  and @julianoes.

I would prefer if the Vehicle was the source of truth for this data, rather than having the GCS infer it, for two reasons:
* it greatly simplifies the implementation of the GCS; and
* as @hamishwillee mentioned above, it means this information remains correct regardless of when the GCS connects. I can see this becoming important as gimbals/cameras become more prevalent, as I can see a use case where one operator is controlling the Vehcile and another operator is connecting periodically to monitor/control the camera.

I also think we should have an enum that defines the type for these progress fields. The specific use case I'm thinking of is `MAV_CMD_NAV_WAYPOINT`, where we have both distance to the waypoint _and_ a hold time once it reaches the waypoint. With the current design, there is no way to report both of these. 

And I think the following also need a progress:
* `MAV_CMD_NAV_DELAY`: time remaining and total time (in seconds)
* `MAV_CMD_CONDITION_DELAY`: time remaining and total time (in seconds)

and possibly also:
* `MAV_CMD_NAV_RETURN_TO_LAUNCH`: this could be either time remaining or distance remaining...

message_definitions/v1.0/common.xml
hamishwillee, line 5306: The easiest way to make these changes is to add them as suggestions, like this.
```suggestion
      <field type="uint16_t" name="item_progress">Progress through current mission item. This is relative to progress_final. Units depend on type of current mission item (and should be defined in each item). 0 initially and if progress_final is 0.</field>
      <field type="uint16_t" name="item_progress_final" invalid="0">Number of mission items in the sequence</field>
```

hamishwillee:
@julianoes I think you are saying that you get the mission lap by downloading the _whole_ mission, looking for [MAV_CMD_DO_JUMP](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_JUMP) instances. Each time a jump iterates, you count the repeat number down? 

That feels pretty heavyweight. We could use items for this and force a jump item to emit a MISSION_CURRENT with a count. Otherwise likely you'd miss them (I assume MISSION_CURRENT may emit non waypoint mission items, such as camera triggers, but generally it won't because those all get executed very quickly).

> And I think the following also need a progress: MAV_CMD_NAV_DELAY, MAV_CMD_CONDITION_DELAY

Yes. Ideal case.

> MAV_CMD_NAV_RETURN_TO_LAUNCH: this could be either time remaining or distance remaining...

Maybe. You can kind of see this on the map already - its not like delays and turns for which it is hard to work out what is going on from telemetry.

> The specific use case I'm thinking of is MAV_CMD_NAV_WAYPOINT, where we have both distance to the waypoint and a hold time once it reaches the waypoint. With the current design, there is no way to report both of these.

Maybe. Or you could say that progress is not reported for anything except the hold time. Of course you'd still want the type in this case (and I think that does make things a little easier, for the cost of just one byte).

message_definitions/v1.0/common.xml
sypaq-nexton: Looks good. Thanks!

julianoes: I would add an enum for the unit, and then time needs to have an actual unit, e.g. seconds.

hamishwillee: I've added units in places. I can add an enum if we get broad agreement on this.

message_definitions/v1.0/common.xml
```diff
@@ -5283,12 +5294,17 @@
...
+        The progress units, such as distance, time, number of turns, must be defined in the mission item (a GCS can always present them as a percentage).
       </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle (on last item, sequence == total). If the autopilot stores its home location as part of the mission this will be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
       <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
       <field type="uint8_t" name="mission_mode" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
+      <field type="uint8_t" name="item_progress_units">0: not used. 1: distance (m), 2: time (seconds), 3: count (laps)</field>
```
hamishwillee, line 5305: Note, to replace with enum if whole strategy agreed.

hamishwillee:
OK, updated in line with discussion so far. Does not consider jumps in mission.

message_definitions/v1.0/common.xml
```diff
@@ -1102,7 +1107,9 @@
         <param index="7" label="Altitude" units="m">Altitude</param>
       </entry>
       <entry value="18" name="MAV_CMD_NAV_LOITER_TURNS" hasLocation="true" isDestination="true">
-        <description>Loiter around this waypoint for X turns</description>
+        <description>Loiter around this waypoint for X turns.
+          MISSION_CURRENT should report the item_progress_* fields as turns/laps.
```
sypaq-nexton, line 1111:
```suggestion
          MISSION_CURRENT should report the item_progress_* fields as a count of turns/laps.
```

message_definitions/v1.0/common.xml
```diff
@@ -5283,12 +5299,17 @@
...
+        The progress units, such as distance, time, number of turns, must be defined in the mission item (a GCS can always present them as a percentage).
       </description>
       <field type="uint16_t" name="seq">Sequence</field>
       <extensions/>
       <field type="uint16_t" name="total" invalid="UINT16_MAX">Total number of mission items on vehicle (on last item, sequence == total). If the autopilot stores its home location as part of the mission this will be excluded from the total. 0: Not supported, UINT16_MAX if no mission is present on the vehicle.</field>
       <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
       <field type="uint8_t" name="mission_mode" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
+      <field type="uint8_t" name="item_progress_units">0: not used. 1: distance (m), 2: time (seconds), 3: count (laps)</field>
```
sypaq-nexton, line 5310:
```suggestion
      <field type="uint8_t" name="item_progress_units">0: not used. 1: distance (m), 2: time (seconds), 3: count</field>
```

Could be reporting counts for things other than laps (e.g. `MAV_CMD_IMAGE_START_CAPTURE` might report number of snapshots taken).

hamishwillee:
@sypaq-nexton We discussed in mav call last night:
  - Use case is valid and having in MISSION_CURRENT makes sense.
  - Might as well provide a percentage as well.
  - Count down is a little more useful than count up. In theory you could just have count down in the message since the recipient should know how many items there are - but even if they didn't it would tell them the useful information "how long have you got". But we figure it's easier for the recipient not to have to look for the info in the mission.
  - There was a thought that most people aren't as interested in counts etc as "how long until the mission moves to the next item" - so waypoint travel time + wait time. That is probably correct, but could be fairly hard to estimate. 
  - This is intended for things that can't be inferred, such as hold and loiter times remaining. I think we should probably restrict use to those cases - i.e. no expectation and perhaps actively discourage publishing the travel time to a waypoint. If we did that then we could allow a message to perhaps emit time rather than counts for a turns loiter.

message_definitions/v1.0/common.xml
hamishwillee: In theory yes, which is why the (laps) is in brackets. The specific example wouldn't work though - image capture persists across waypoints - i.e. the info in the mission_current would get emitted once and so could not give a count.

message_definitions/v1.0/common.xml
```diff
@@ -5283,12 +5316,19 @@
...
       <field type="uint8_t" name="mission_state" enum="MISSION_STATE" invalid="0">Mission state machine state. MISSION_STATE_UNKNOWN if state reporting not supported.</field>
       <field type="uint8_t" name="mission_mode" invalid="0">Vehicle is in a mode that can execute mission items or suspended. 0: Unknown, 1: In mission mode, 2: Suspended (not in mission mode).</field>
+      <field type="uint8_t" name="item_progress_units" enum="MISSION_ITEM_PROGRESS_UNITS">Units for remaining/original progress fields. MISSION_ITEM_PROGRESS_UNITS_NOT_USED (0) if not reporting progress. Units depend on particular mission item.</field>
+      <field type="uint8_t" name="item_progress_percentage" units="%" invalid="UINT8_MAX">0-100: Percentage remaining (countdown). 0 if progress not being reported (item_progress_units is MISSION_ITEM_PROGRESS_UNITS_NOT_USED).</field>
+      <field type="uint16_t" name="item_progress_remaining" invalid="0">Remaining time/distance/count to complete current mission item (units in item_progress_units). 0: progress not reported or item complete.</field>
```
hamishwillee, line 5331: @sypaq-nexton @julianoes @auturgy (FYI @MaEtUgR ) I've just updated the docs in line with our discussion on the dev call plus some minor mods based on what is actually needed.

This reports both percentage and remaining progress as a _countdown_, and omits the "initial count". We're really interested in how long we have to wait or our percentage through - we don't care so much what we started with - and if we do, then we can get it from the mission item.

We only report this for things that you can't infer - such as loiter/delay.

We don't report this for travel time in waypoints, which can be inferred. We don't report it for unlimited delays - they have no "progress'.

I am quite happy with this, with one minor caveat. It is harder to report "turns=3/5" as you would have to estimate by dividing by the percentage and rounding to a whole number (note, you do not have to worry about this for time based delays since remaining time or % are how you would report those). I think the saving of 2 bytes and the simplicity of what gets reported is worth it. 

Does this work for everyone?

sypaq-nexton: I can work with this. Thanks!

sypaq-nexton:
@hamishwillee wrote:
> EDIT - but then count up allows open ended counting for an unlimited loiter where we have no end point.

If you really wanted to allow counting up, you could potentially capture it in the `MISSION_ITEM_PROGRESS_UNITS`, e.g.. you could have both:
* `MISSION_ITEM_PROGRESS_UNITS_[UNIT]_ELAPSED` and
* `MISSION_ITEM_PROGRESS_UNITS_[UNIT]_REMAINING`.

But I'm not sure the value justifies the extra complexity.

hamishwillee:
> But I'm not sure the value justifies the extra complexity.

Yes. It doesn't to me.

message_definitions/v1.0/common.xml
```diff
@@ -1299,7 +1310,9 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="93" name="MAV_CMD_NAV_DELAY" hasLocation="false" isDestination="false">
-        <description>Delay the next navigation command a number of seconds or until a specified time</description>
+        <description>Delay the next navigation command a number of seconds or until a specified time
```
hamishwillee, line 1313:
```suggestion
        <description>Delay the next navigation command a number of seconds or until a specified time.
```

hamishwillee:
@auturgy @julianoes You OK with this too? Or are we waiting on a @nexton-winjeel to create an implementation and then we merge (this has to be on common.xml, so if we approve, it's live).

hamishwillee:
@nexton-winjeel We discussed this in the call last night. This has "general approval" as a workable system.

Because we don't like merging things into common.xml until there is some testing/prototyping we need to see this appearing in a flight stack/GCS before we merge. Do you have a plan/timeline for that implementation?

FYI that there is another PR for mission_current related to detecting mission changes, so the actual position of fields might change depending on which gets prototyped/accepted first.

nexton-winjeel:
@hamishwillee: It's on my roadmap. Planned for sometime this year, but I don't have a firmer date than that.

hamishwillee:
Thanks @NaterGator - much appreciated.

nexton-winjeel:
@hamishwillee, @auturgy, @julianoes:

Initial implementation for QGC and ArduPilot available at:
* https://github.com/Winjeel/qgroundcontrol/tree/upstream/add-mission-item-progress
* https://github.com/Winjeel/ardupilot/tree/upstream/add-mission-item-progress

(Note that neither of these will build because they don't include the MAVLink changes in this PR).

But with the changes in this PR, it looks like the following:
qgc-item-progress-distance
qgc-item-progress-count
qgc-item-progress-time

I'll update those branches and raise PRs once this PR is merged.

hamishwillee:
@nexton-winjeel So in a loiter you're reporting distance to destination then number of turns remaining? Didn't we agree we only reported the things you couldn't work out from GCS. I'm happy to discuss/modify, we just shouldn't do stuff we decided against. See

> We don't report this for travel time in waypoints, which can be inferred. We don't report it for unlimited delays - they have no "progress'.

nexton-winjeel:
@hamishwillee: These QGC and ArduPilot changes don't have to be the final solution. It was done this way to demonstrate that:
* Progress can be reported for various mission item types;
* Each item progress unit type can be correctly reported and displayed; and
* Mission Items can report different unit types throughout execution.

nexton-winjeel:
@hamishwillee:
> Because we don't like merging things into common.xml until there is some testing/prototyping we need to see this appearing in a flight stack/GCS before we merge. 

Does this mean you need a proof-of-concept like I've done, or that you need this functionality merged into a flight stack/GCS?

hamishwillee:
> These QGC and ArduPilot changes don't have to be the final solution. It was done this way to demonstrate that ..

I'm not saying I don't like this, I'm saying that the implementation must be compliant with the specification. That specification currently says:

- The progress units, such as distance, time, count (for example, of "remaining turns"), must be defined in the mission item.
- MISSION_CURRENT should set the item_progress_remaining field to the remaining loiter time (s) once loitering.
- Nothing is mentioned about distance, but in this discussion we explicitly said "only report on things that you can't infer".

Don't get me wrong, I like what you've done and I don't think it breaks anything, as long as there is no expectation in the GCS that this information will be provided. Just making sure that when someone does an implementation it is compliant, and if we do choose to provide distance in an implementation, the docs say that is OK.

In summary "the spec should be something all flight stacks can develop to and likely end up with the behaviour".

>> Because we don't like merging things into common.xml until there is some testing/prototyping we need to see this appearing in a flight stack/GCS before we merge.

> Does this mean you need a proof-of-concept like I've done, or that you need this functionality merged into a flight stack/GCS?

Usually it would need broad commitment from the flight stack to take your changes. If you have that, what you have done is probably enough. We'd discuss that in the dev call next week if you think you have that agreement? I.e. has Peter Barker or JamesP looked at this?

hamishwillee:
@nexton-winjeel We discussed in the call. What is needed is a PR in ArduPilot (or PX4) so that we can see discussion and assess ArduPilot's commitment to taking this. We think it's "good" but we're wary of creating orphan fixes that never end up in a flight stack.

nexton-winjeel:
@hamishwillee: ArduPilot PR here: https://github.com/ArduPilot/ardupilot/pull/25249

I'll discuss it with them in the next dev call.

hamishwillee:
Thanks! Good luck. I would like this as part of the standard.

tridge (commented):
This only works in AUTO, but the same information also makes sense in RTL, GUIDED, QRTL and many other modes. How will they be handled?
maybe this should be a new message, which is requested via message intervals by the GCS?
this proposal also doesn't work for ship landing, doesn't work for precland doesn't work for follow, in each case the destination is not in the mission. Also doesn't work for rally points, again missing the destination
also, how does the flight controller know if the user wants time or distance?
in a new msg it could have float time and float distance, with NaN for unknown

nexton-winjeel:
Following discussion with @tridge and the ArduPilot devs, we probably want a new message that looks something like:

```xml
    <message id="11045" name="NAV_CONTROLLER_PROGRESS">
      <description>Progress of the current Nav Controller item.</description>
      <field type="uint8_t" name="percentage_complete">Percentage complete. Set to 0 if unknown.</field>
      <field type="float" name="distance_remaining" units="m" invalid="NaN">Distance remaining (in metres) until the current Nav Controller item is complete. Set to NaN if unknown, or not applicable.</field>
      <field type="float" name="time_remaining" units="s" invalid="NaN">Time remaining (in seconds) until the current Nav Controller item is complete. Set to NaN if unknown, or not applicable.</field>
      <field type="float" name="count_remaining" invalid="NaN">Count remaining until the current Nav Controller item is complete. Set to NaN if unknown, or not applicable.</field>
      <!-- MAV_MISSION_TYPE field here so we can reference Rally points? -->
      <field type="uint16_t" name="seq">Sequence (mission item number). Set to 0 if not a mission item.</field>
      <field type="int32_t" name="x" invalid="INT32_MIN">X pos/Latitude of nav target. Set to INT32_MIN if no target, or to use the location of the corresponding mission item.</field>
      <field type="int32_t" name="y" invalid="INT32_MIN">Y pos/Longitude of nav target. Set to INT32_MIN if no target, or to use the location of the corresponding mission item.</field>
      <field type="float" name="alt" units="m" invalid="NaN">Altitude of nav target. Set to NaN if no target, or to use the location of the corresponding mission item.</field>
      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame of target.</field>
    </message>
```

hamishwillee:
All excellent points. There are many other situations where the GCS can't know what the vehicle thinks it is doing so this is much better.

I don't know enough about ship landing or precland to know what you need there. Can you outline what the GCS would want to present for progress and what the vehicle would need to supply?

Ditto for Follow? Isn't that an unbounded operation - we decided that you would only report for times, distances that you could know. Follow is more like an unlimited orbit. Or to put it another way "what is the progress" in this case?

> Also doesn't work for rally points, again missing the destination

So is the point here that you want to include a destination so the GCS always knows where the vehicle thinking it needs to go in cases such as rally points, where the vehicle makes a decision and does not communicate it? Nice

> also, how does the flight controller know if the user wants time or distance?

We've been saying that it isn't up to the user. Are we OK to leave this up to the flight stack? Do we want to allow both as proposed

@peterbarker One reason I was thinking in the mission context was to avoid new messages/additional flash memory cost. What is ArduPilot's appetite for a message as proposed above.

auturgy:
Discussed at Mavlink dev call again - all supportive of proposed changes/new message, waiting for update

hamishwillee:
Note, above message should be created in a new PR in development.xml. As above, happy enough with the message proposed but would like to expand on what we expect it to be for various cases such as precland and ship landing etc so that we can be sure all the information needed is present.

hamishwillee:
@nexton-winjeel Do you still have an interest in this use case? I'm closing stalled issues, but this still seems relevant to me, so if you're still a stakeholder for it I could have another go at getting some buy-in.

peterbarker:
> @nexton-winjeel Do you still have an interest in this use case? I'm closing stalled issues, but this still seems relevant to me, so if you're still a stakeholder for it I could have another go at getting some buy-in.

FWIW, I did quite like the idea.

nexton-winjeel:
We've currently implemented the changes from [this comment](https://github.com/mavlink/mavlink/pull/2029#issuecomment-1744636407) in our fork of ArduPilot. That won't get merged into ArduPilot per [this comment](https://github.com/mavlink/mavlink/pull/2029#pullrequestreview-1681114332), as the preference is for a new message that supports more that just mission item progress. 

> Do you still have an interest in this use case?

Yes, I do. Our operators greatly appreciate this capability. I'd like to get these changes merged upstream, but I don't have any capacity right now to massage it into an acceptable shape that will allow it to be merged. 

It's still on our backlog though, but it is down at the bottom of the list.

hamishwillee:
Thanks @nexton-winjeel I'll add to the 2 week-away-dev call to see what others think. 

Coming back to this with fresh eyes, I do think the proposal for a `NAV_CONTROLLER_PROGRESS` in https://github.com/mavlink/mavlink/pull/2029#issuecomment-1766214285 is a better approach. The main reason is that we keep getting getting requests for people wanting to know where they are headed outside of missions.

For example, where am I going in my rally point return. There is also some interest in knowing where I might go if I were to rally point return right now - i.e. what's my rally point target - so perhaps that a separate thing.

I can imagine thought it is harder to "contextualize". If you're in a mission processing a loiter command, it is easy for a GCS to draw the count next to the current waypoint. If this is separate from the mission item, or needs to be associated then perhaps it is a little harder to manage. 

Looking at NAV_CONTROLLER_PROGRESS I would always supply the target details if they are know - you don't know for sure that the GCS does have the current mission onboard.

nexton-winjeel:
@hamishwillee: Regarding the `NAV_CONTROLLER_PROGRESS` message, there's a draft PR to add the equivalent mission progress information using that message instead of the `MISSION_CURRENT` message:
* https://github.com/ArduPilot/ardupilot/pull/27779

hamishwillee:
>  * [Copter, AP_Mission: Send progress for individual mission items (take 2) ArduPilot/ardupilot#27779](https://github.com/ArduPilot/ardupilot/pull/27779)

Thanks so much @nexton-winjeel . So if we were to accept `NAV_CONTROLLER_PROGRESS` (or some variation of it) would it be reasonable to assume that you would work to push this into ArduPilot - i.e. to be the stakeholder?

I can work with others to drive consensus, but I have no resource to do the work needed to validate. So if there is no stakeholder to own implementation we stall.

nexton-winjeel:
Yes, I can work on getting `NAV_CONTROLLER_PROGRESS` into ArduPilot. No promises on timelines though...

hamishwillee:
@nexton-winjeel Thanks very much. I've created #2546 for the discussion (there seems to be broad support for it in the dev call).

PR #2032, opened 2023-08-18
fix RC_CHANNELS_SCALED inactive channel

rsaxvc:
Based my reading of the message types and discussion here: https://groups.google.com/g/mavlink/c/cocB5FwakRI/m/B25tXOkD3mQJ , I think this value should probably fit in an int16_t.


hamishwillee (approved):
THank you - you're absolutely right - the values are int16_t, so the inactive value should be the max value of that type.
This same information is set in the `invalid` flag, so I have updated that too.

rsaxvc:
Thank you, your updates look good to me. Should I squash them into one commit or is there anything else I can help with?

hamishwillee:
No, all good. Thanks. Much appreciated.

PR #2035, opened 2023-08-26
common.xml: display GIMBAL_MANAGER_STATUS.flags as bitmask

peterbarker:


message_definitions/v1.0/common.xml
```diff
@@ -6902,7 +6902,7 @@
     <message id="281" name="GIMBAL_MANAGER_STATUS">
       <description>Current status about a high level gimbal manager. This message should be broadcast at a low regular rate (e.g. 5Hz).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS">High level gimbal manager flags currently applied.</field>
+      <field type="uint32_t" name="flags" enum="GIMBAL_MANAGER_FLAGS" display="bitmask">High level gimbal manager flags currently applied.</field>
```
hamishwillee, line 6905: @peterbarker At some point we decided why put the bitmask entry on the user, when it is a property of the enum. So we have `<enum name="GIMBAL_MANAGER_FLAGS" bitmask="true">`.

That said, we're inconsistent on this and we should decide if we want one or the other or both. 

I'll merge for now, but do you have an opinion on this?

Issue #2040, opened 2023-09-26
CI is failing Node tests


magicrub:
CI is failing Node tests. Seems unrelated to anything from this repo, suspect a Node library dependency just expired/retired/exploded.

Example PR https://github.com/ArduPilot/mavlink/pull/329

hamishwillee:
The issue in node 12 test is the use of `descriptor?.isOwn` - the optional chaining operator appears first in node 14 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

We should drop that. Not sure about the other one.

hamishwillee:
@julianoes Any ideas on the Node 14 failure? Copied from [here](https://github.com/mavlink/mavlink/actions/runs/6179806789/job/16775320760?pr=2039) (starting from first error):
<details>
Merged enum MAV_CMD
Merged enum MAV_CMD
Merged enum MAV_CMD
Merged enum MAV_CMD
ERROR: Duplicate enum names:
	SPEED_TYPE.SPEED_TYPE_AIRSPEED = 0 @ /home/runner/work/mavlink/mavlink/message_definitions/v1.0/common.xml:3438
	SPEED_TYPE.SPEED_TYPE_AIRSPEED = 0 @ /home/runner/work/mavlink/mavlink/message_definitions/v1.0/ardupilotmega.xml:27
rm: cannot remove './javascript/implementations/mavlink_all_v1.0/local_modules': No such file or directory
Validating ../../message_definitions/v1.0/ardupilotmega.xml
Parsing ../../message_definitions/v1.0/ardupilotmega.xml
Validating /home/runner/work/mavlink/mavlink/message_definitions/v1.0/common.xml
Parsing /home/runner/work/mavlink/mavlink/message_definitions/v1.0/common.xml
Validating /home/runner/work/mavlink/mavlink/message_definitions/v1.0/uAvionix.xml
Parsing /home/runner/work/mavlink/mavlink/message_definitions/v1.0/uAvionix.xml
Validating /home/runner/work/mavlink/mavlink/message_definitions/v1.0/icarous.xml
Parsing /home/runner/work/mavlink/mavlink/message_definitions/v1.0/icarous.xml
Validating /home/runner/work/mavlink/mavlink/message_definitions/v1.0/cubepilot.xml
Parsing /home/runner/work/mavlink/mavlink/message_definitions/v1.0/cubepilot.xml
[...]

PR #2041, opened 2023-09-26
Update test_and_deploy.yml - drop node 12

hamishwillee:
The node 12 tests fail due to use of optional chaining in generated code, added in node 14. This removes node 12.
I added node 16 because we should be promoting a more recent versions. Possibly even 18.

Part of fixing #2040

### DIFF
--- a/.github/workflows/test_and_deploy.yml
+++ b/.github/workflows/test_and_deploy.yml
@@ -91,7 +91,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install future lxml
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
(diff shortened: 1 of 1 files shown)


magicrub:
Welp.. now 16 fails LOL. Do I hear 18?

magicrub:
if 18 is the current, why not include that one?

hamishwillee:
The core problem was that `npm install` behaves differently  if the package.json does not exist. In v15 and earlier it creates an empty package file and installs nothing. In node 16 it fails because the file is not in root.
The `npm install` line was incorrect and not needed - it was never installing anything, and it shouldn't. The correct version of package.json depends on what version of the generated MAVLink libraries is used, and is selected by the test script in the next step.

I added all the versions from node 14-18, and updated the node setup module to the latest. 

This should be good, but someone else should sanity check. @auturgy ? @magicrub ?

hamishwillee:
OK, no sanity check, but I believe I'm sane :-)

PR #2046, opened 2023-10-11
common.xml: add note about scaling pitch/roll if transporting command…

peterbarker:
… via command_int

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1589,8 +1589,8 @@
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5" label="Pitch Offset">Pitch offset from next waypoint, positive pitching up</param>
-        <param index="6" label="Roll Offset">Roll offset from next waypoint, positive rolling to the right</param>
+        <param index="5" label="Pitch Offset">Pitch offset from next waypoint, positive pitching up.  If this command is transported via COMMAND_INT then this field is deg*1e7.</param>
+        <param index="6" label="Roll Offset">Roll offset from next waypoint, positive rolling to the right.  If this command is transported via COMMAND_INT then this field is deg*1e7.</param>
         <param index="7" label="Yaw Offset">Yaw offset from next waypoint, positive yawing to the right</param>
       </entry>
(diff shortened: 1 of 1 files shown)


julianoes:
I think PX4 expects degrees:
https://github.com/PX4/PX4-Autopilot/blob/6dfede0806f387c6cf0dca8f5030f41592dd2bbb/src/modules/navigator/navigator_main.cpp#L665-L667

hamishwillee:
Yes, from what I can see PX4 expects degrees if used in missions, ignoring any frame information. It does not appear to be handled as a command. 

@peterbarker Two questions:
1. When would you use this as a command? By it's name MAV_CMD_DO_SET_ROI_WPNEXT_OFFSET assumes you have a next waypoint - so it doesn't look like it makes sense outside of a mission.
2. What level of granularity makes sense. I mean if this sent in an integer command you basically get degrees truncated. But I suspect that it might be rare to need 10ths of degrees or whatever? It would certainly be easier to just define as degrees and note that if sent in an int any more granularity than the nearest degree would be lost.

message_definitions/v1.0/common.xml
```diff
@@ -1589,8 +1589,8 @@
         <param index="2">Empty</param>
         <param index="3">Empty</param>
         <param index="4">Empty</param>
-        <param index="5" label="Pitch Offset">Pitch offset from next waypoint, positive pitching up</param>
-        <param index="6" label="Roll Offset">Roll offset from next waypoint, positive rolling to the right</param>
+        <param index="5" label="Pitch Offset">Pitch offset from next waypoint, positive pitching up.  If this command is transported via COMMAND_INT then this field is deg*1e7.</param>
```
tridge, line 1592: i'm not keen on this, lets leave it as integer when done as a command_int

peterbarker:
OK, agreed, it's integer degrees, thanks all!

hamishwillee:
FYI @peterbarker Updated to specify explicitly degrees, as discussed https://github.com/mavlink/mavlink/commit/70181c42fc63306ba3512666e1a8b7b782416806

Yes, I know this didn't go via a PR - a mistake. I'm going to try protect the master branch against me doing this again. Though i already had.

PR #2047, opened 2023-10-11
Add support for illuminators

amy-chen-skydio:
- Add MAV_TYPE and MAV_COMPONENT enums for illuminators
 - Remove MAV_CMD_ILLUMINATOR_ON_OFF from wip
 - Add command to configure illuminator settings
 - Add messages to report illuminator status and information

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7624,8 +7624,8 @@
     </message>
     <message id="420" name="ILLUMINATOR_STATUS">
       <description>Illuminator status</description>
-      <field type="uint32_t" name="time_boot_us" units="us">Uptime in us</field>
-      <field type="uint8_t" name="enable">0: Illuminators OFF, 1: Illuminators ON</field>      
+      <field type="uint32_t" name="uptime_ms" units="us">Time since the start-up of the illuminator in ms</field>
+      <field type="uint8_t" name="enable">0: Illuminators OFF, 1: Illuminators ON</field>
       <field type="uint32_t" name="error_status">Error</field>
       <field type="uint8_t" name="mode" enum="ILLUMINATOR_MODE">Illuminator mode</field>
       <field type="float" name="brightness" units="%">Illuminator brightness</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5072,6 +5072,9 @@
       <entry value="2" name="ILLUMINATOR_ERROR_FLAGS_OVER_TEMPERATURE_SHUTDOWN">
         <description>Illuminator over temperature shutdown error.</description>
       </entry>
+      <entry value="3" name="ILLUMINATOR_ERROR_FLAGS_THERMISTOR_FAILURE">
+        <description>Illuminator thermistor failure.</description>
+      </entry>
     </enum>
   </enums>
   <messages>
@@ -7642,9 +7645,9 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
-    <message id="420" name="ILLUMINATOR_STATUS">
+    <message id="440" name="ILLUMINATOR_STATUS">
       <description>Illuminator status</description>
-      <field type="uint32_t" name="uptime_ms" units="us">Time since the start-up of the illuminator in ms</field>
+      <field type="uint32_t" name="uptime_ms" units="ms">Time since the start-up of the illuminator in ms</field>
       <field type="uint8_t" name="enable">0: Illuminators OFF, 1: Illuminators ON</field>
       <field type="uint8_t" name="mode_bitmask" enum="ILLUMINATOR_MODE" display="bitmask">Supported illuminator modes</field>
       <field type="uint32_t" name="error_status" enum="ILLUMINATOR_ERROR_FLAGS" display="bitmask">Errors</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7645,7 +7645,7 @@
     <message id="421" name="ILLUMINATOR_INFORMATION">
       <description>Illuminator information</description>
       <field type="char[16]" name="serial_number">Serial Number</field>
-      <field type="uint32_t" name="firmware_version">Version of the illuminator firmware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff).</field>
+      <field type="uint32_t" name="firmware_version">Version of the illuminator firmware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). Use 0 if not known.</field>
       <field type="float" name="min_strobe_period" units="s">Minimum strobing period in seconds</field>
       <field type="float" name="max_strobe_period" units="s">Maximum strobing period in seconds</field>
       <field type="uint8_t" name="mode_bitmask" enum="ILLUMINATOR_MODE" display="bitmask">Supported illuminator modes</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5064,6 +5064,15 @@
         <description>Illuminator behavior is controlled by external factors: e.g. an external hardware signal</description>
       </entry>
     </enum>
+    <enum name="ILLUMINATOR_ERROR_FLAGS" bitmask="true">
+      <description>Illuminator module error flags (bitmap, 0 means no error)</description>
+      <entry value="1" name="ILLUMINATOR_ERROR_FLAGS_THERMAL_THROTTLING">
+        <description>Illuminator thermal throttling error.</description>
+      </entry>
+      <entry value="2" name="ILLUMINATOR_ERROR_FLAGS_OVER_TEMPERATURE_SHUTDOWN">
+        <description>Illuminator over temperature shutdown error.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
@@ -7626,7 +7635,7 @@
       <description>Illuminator status</description>
       <field type="uint32_t" name="uptime_ms" units="us">Time since the start-up of the illuminator in ms</field>
       <field type="uint8_t" name="enable">0: Illuminators OFF, 1: Illuminators ON</field>
-      <field type="uint32_t" name="error_status">Error</field>
+      <field type="uint32_t" name="error_status" enum="ILLUMINATOR_ERROR_FLAGS" display="bitmask">Errors</field>
       <field type="uint8_t" name="mode" enum="ILLUMINATOR_MODE">Illuminator mode</field>
       <field type="float" name="brightness" units="%">Illuminator brightness</field>
       <field type="float" name="strobe_period" units="s">Illuminator strobing period in seconds</field>
@@ -7635,9 +7644,8 @@
     </message>
     <message id="421" name="ILLUMINATOR_INFORMATION">
       <description>Illuminator information</description>
-      <field type="uint16_t" name="vendor_id">Vendor ID</field>
-      <field type="uint16_t" name="product_id">Product ID</field>
       <field type="char[16]" name="serial_number">Serial Number</field>
+      <field type="uint32_t" name="firmware_version">Version of the illuminator firmware, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff).</field>
       <field type="float" name="min_strobe_period" units="s">Minimum strobing period in seconds</field>
       <field type="float" name="max_strobe_period" units="s">Maximum strobing period in seconds</field>
       <field type="uint8_t" name="mode_bitmask" enum="ILLUMINATOR_MODE" display="bitmask">Supported illuminator modes</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1967,10 +1967,15 @@
         </description>
       </entry>
       <entry value="405" name="MAV_CMD_ILLUMINATOR_ON_OFF" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Turns illuminators ON/OFF. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
+      </entry>
+      <entry value="406" name="MAV_CMD_DO_ILLUMINATOR_CONFIGURE" hasLocation="false" isDestination="false">
+        <description>Configures illuminator settings. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
+        <param index="1" label="Mode" enum="ILLUMINATOR_MODE">Mode</param>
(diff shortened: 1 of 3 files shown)


hamishwillee:
@amy-chen-skydio Sounds reasonable - does this mean you have an implementation in place for all of this already? 

If you are not at that point yet I suggest we push all of this into development.xml so you can prototype, and back into common.xml later.

message_definitions/v1.0/common.xml
```diff
@@ -1949,11 +1949,16 @@
...
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
       </entry>
+      <entry value="406" name="MAV_CMD_DO_ILLUMINATOR_CONFIGURE" hasLocation="false" isDestination="false">
+        <description>Configures illuminator settings. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
+        <param index="1" label="Mode" enum="ILLUMINATOR_MODE">Mode</param>
+        <param index="2" label="Brightness" units="%" minValue="0" maxValue="100">0%: Off, 100%: Max Brightness</param>
+        <param index="3" label="Strobe Period" units="s">Strobe period in seconds</param>
+        <param index="4" label="Strobe Duty" units="%" minValue="1" maxValue="100">Strobe duty cycle where 100% means it is on constantly</param>
```
hamishwillee, line 1960: 0 means "not used"?

message_definitions/v1.0/common.xml
```diff
@@ -1949,11 +1949,16 @@
...
         <description>Turns illuminators ON/OFF. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
       </entry>
+      <entry value="406" name="MAV_CMD_DO_ILLUMINATOR_CONFIGURE" hasLocation="false" isDestination="false">
+        <description>Configures illuminator settings. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
+        <param index="1" label="Mode" enum="ILLUMINATOR_MODE">Mode</param>
+        <param index="2" label="Brightness" units="%" minValue="0" maxValue="100">0%: Off, 100%: Max Brightness</param>
+        <param index="3" label="Strobe Period" units="s">Strobe period in seconds</param>
```
hamishwillee, line 1959: 0 means not used?

message_definitions/v1.0/common.xml
```diff
@@ -1949,11 +1949,16 @@
...
       </entry>
       <entry value="405" name="MAV_CMD_ILLUMINATOR_ON_OFF" hasLocation="false" isDestination="false">
-        <wip/>
-        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Turns illuminators ON/OFF. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
         <param index="1" label="Enable" minValue="0" maxValue="1" increment="1">0: Illuminators OFF, 1: Illuminators ON</param>
       </entry>
+      <entry value="406" name="MAV_CMD_DO_ILLUMINATOR_CONFIGURE" hasLocation="false" isDestination="false">
+        <description>Configures illuminator settings. An illuminator is a light source that is used for lighting up dark areas external to the system: e.g. a torch or searchlight (as opposed to a light source for illuminating the system itself, e.g. an indicator light).</description>
+        <param index="1" label="Mode" enum="ILLUMINATOR_MODE">Mode</param>
```
hamishwillee, line 1957: Might it make sense to include this all in MAV_CMD_ILLUMINATOR_ON_OFF (a question, not a "direction").  That works provided we can use 0 as the default for the other values - so any other implementation won't do surprising things.

message_definitions/v1.0/common.xml
```diff
@@ -7504,6 +7521,25 @@
...
+    </message>
+    <message id="421" name="ILLUMINATOR_INFORMATION">
+      <description>Illuminator information</description>
+      <field type="uint16_t" name="vendor_id">Vendor ID</field>
+      <field type="uint16_t" name="product_id">Product ID</field>
+      <field type="char[16]" name="serial_number">Serial Number</field>
+      <field type="float" name="min_strobe_period" units="s">Minimum strobing period in seconds</field>
+      <field type="float" name="max_strobe_period" units="s">Maximum strobing period in seconds</field>
+      <field type="uint8_t" name="mode_bitmask" enum="ILLUMINATOR_MODE" display="bitmask">Supported illuminator modes</field>
+    </message>
```
hamishwillee, line 7542: Is the intent to stream these two messages or just request them ad-hoc? 

I appreciate that separating them makes sense if streamed at different rates, but if you only ever ask for this explicitly, and only on request, then might be worth merging these.

message_definitions/v1.0/common.xml
amy-chen-skydio: Our two cents is that it makes more sense to keep them separate. If they are all included in `MAV_CMD_ILLUMINATOR_ON_OFF`, it seems to contradict the original purpose of this command in the case of enable=1 with brightness=0. The illuminator would be off but enable says the illuminator is on.

amy-chen-skydio:
> @amy-chen-skydio Sounds reasonable - does this mean you have an implementation in place for all of this already?
> 
> If you are not at that point yet I suggest we push all of this into development.xml so you can prototype, and back into common.xml later.

@hamishwillee Yes, we already have an implementation in place for all of this already.

message_definitions/v1.0/common.xml
amy-chen-skydio: The intent is to ad-hoc `ILLUMINATOR_INFORMATION` but stream `ILLUMINATOR_STATUS` so that we can monitor the state of the illuminator while only requesting information that wouldn't change once at the very start.

message_definitions/v1.0/common.xml
amy-chen-skydio: Addressed in the new commit

message_definitions/v1.0/common.xml
amy-chen-skydio: Addressed in the new commit

hamishwillee:
@auturgy @julianoes The design of these messages seems reasonable/in line with what we do normally. The only thing I might do is suggest adding a field for instance, to allow multiple illuminators to be separately controlled if desired.

BUT I know nothing about illuminators. This seems like a reasonable set of things to do, but is there anyone we can ask? Is this something ArduPilot/PX4 might take?

auturgy:
I'm not clear on PID/VID: this is usually related to USB devices so whilst I think I understand the intent, not sure this is the right data. Also, time since boot/uptime needs clarification. Which is it?

hamishwillee:
@julianoes @auturgy What are your thoughts on this? It looks sane to me but I need broader view.

message_definitions/v1.0/common.xml
julianoes: I have the same comment as in https://github.com/mavlink/mavlink/pull/2063. We should use the general https://mavlink.io/en/messages/common.html#COMPONENT_METADATA or https://mavlink.io/en/messages/development.html#COMPONENT_INFORMATION_BASIC messages for generic things like name, version, etc.

amy-chen-skydio: Would this change work?

- Keep `ILLUMINATOR_STATUS` as is
- Remove `serial_number` and `firmware_version` from `ILLUMINATOR_INFORMATION` and use `COMPONENT_INFORMATION_BASIC` instead
- Keep `min_strobe_period`, `max_strobe_period`, and `mode_bitmask` as part of `ILLUMINATOR_INFORMATION` since it is not generic

julianoes: I think that goes in the right direction, yes.

Are min strobe period, max strobe period, etc. generic illuminator properties? I'm not very familiar with such, so I'm wondering if this is only applicable to whatever you're using or a general thing?

I'm just trying to prevent that the next contributor is going to come along and want to have it all in brightness and then have it report max lumen instead.

amy-chen-skydio: I have updated `ILLUMINATOR_INFORMATION` and removed `serial_number` and `firmware_version`.

And yes, they're pretty generic illuminator properties. Many of the lights our team tested for drones have a strobe function.

hamishwillee:
Are there other things _like_ strobe we might be needed ? I.e. is there control over colour etc?

@julianoes Let me know when you think this is ready. I believe our thinking in the MAV call was that this message made sense and is pretty close to right.

julianoes:
I wonder if we could have just one STATUS message, and all fields from the INFORMATION message integrated there. To save some space, we can probably use some int8 or int16 instead of floats and that way the message will be even smaller than now. That way, a ground station doesn't have to request both information and status which is certainly more cumbersome.

Also, we could consider extending the existing ILLUMINATOR command rather than adding a CONFIGURE command, but I don't have a strong opinion there. I just know that in the past having a CONTROL and a CONFIGURE command (in the case of gimbal) lead to confusion and wasn't always implemented consistently.

message_definitions/v1.0/common.xml
hamishwillee: > Also, we could consider extending the existing ILLUMINATOR command rather than adding a CONFIGURE command, but I don't have a strong opinion there. I just know that in the past having a CONTROL and a CONFIGURE command (in the case of gimbal) lead to confusion and wasn't always implemented consistently.

@julianoes I had similar concerns here - see @amy-chen-skydio response. 

I don't have a strong opinion either. Generally one message consumes less resources everywhere, and users are less likely to "omit" it. On the other hand, in theory if we were to add brightness etc to the existing message then the default values of zero would mean that updated recipients getting this command from non-updated senders would set the brightness to zero. Not great.

I'll be lead by you.

message_definitions/v1.0/common.xml
```diff
@@ -7504,6 +7530,23 @@
...
+      <field type="float" name="strobe_period" units="s">Illuminator strobing period in seconds</field>
+      <field type="float" name="strobe_duty_cycle" units="%">Illuminator strobing duty cycle</field>
+      <field type="float" name="temp_c">Temperature in Celsius</field>
+    </message>
+    <message id="421" name="ILLUMINATOR_INFORMATION">
+      <description>Illuminator information</description>
+      <field type="float" name="min_strobe_period" units="s">Minimum strobing period in seconds</field>
+      <field type="float" name="max_strobe_period" units="s">Maximum strobing period in seconds</field>
+      <field type="uint8_t" name="mode_bitmask" enum="ILLUMINATOR_MODE" display="bitmask">Supported illuminator modes</field>
+    </message>
```
hamishwillee, line 7549: > I wonder if we could have just one STATUS message, and all fields from the INFORMATION message integrated there. To save some space, we can probably use some int8 or int16 instead of floats and that way the message will be even smaller than now. That way, a ground station doesn't have to request both information and status which is certainly more cumbersome.

@julianoes I suspect that the thinking here is that `ILLUMINATOR_INFORMATION` is static and gets requested once, while `ILLUMINATOR_STATUS` is streamed. So much like the battery information v2 messages they get split and only the information needed gets sent.

The approach makes sense for battery where there are lots of batteries, lots of static data, and the higher-rate message is sent often enough to have an impact on bandwidth on slow links. Not sure about this case whether it is worth the complication. In particular ArduPilot often argue that the cost of additional flash for more messages is not worth it for a few extra fields on an existing message.

I suspect this is probably relatively low rate, there is most likely only one or two of them on most systems. So I lean towards your view.

amy-chen-skydio: That makes sense. I can move those fields into `ILLUMINATOR_STATUS` and remove `ILLUMINATOR_INFORMATION` altogether then.

message_definitions/v1.0/common.xml
julianoes: Ok good.

message_definitions/v1.0/common.xml
julianoes: Fair enough. If there are actually implementations out there that we would break, then I accept that argument.

hamishwillee:
@julianoes This appears to be ready for re-review. Do you need me to look at it too?

julianoes (approved):

hamishwillee (commented):
@julianoes This looks OK on the face of it, though I have not read all the discussion.

This pulls the COMPONENT_INFORMATION_BASIC from development into common.xml - does that mean that there is an implementation? (and also the rest of it is also going into common).

@auturgy Can you please confirm you are happy with these changes?

hamishwillee:
Further? The normal rules for inclusion into common require that:
1. Two stakeholders agree that they will implement this, or at least that they would implement it if they were interested in the feature. I think this looks like a sensible design, so we could perhaps take this one as "likely".
2. There has been a prototype and some evidence that this is properly tested and meets the broad requirements of all likely users. Normally that's a pointer to an accepted open source PR showing implementation, and sometimes test logs etc.

@amy-chen-skydio Can you provide any pointers/evidence of the implementation?

amy-chen-skydio:
> This pulls the COMPONENT_INFORMATION_BASIC from development into common.xml - does that mean that there is an implementation? (and also the rest of it is also going into common).

We have implementation using COMPONENT_INFORMATION_BASIC for illuminator as well as parachute.

> @amy-chen-skydio Can you provide any pointers/evidence of the implementation?

We do not plan on open source the drone or illuminator code so would a log of the MAVLink communication interface between our drone and the illuminator during operation be acceptable?

Thank you!

hamishwillee:
> We do not plan on open source the drone or illuminator code so would a log of the MAVLink communication interface between our drone and the illuminator during operation be acceptable?

amy-chen-skydio:
@hamishwillee

> For our due diligence, what would be great your test suite and and the logs showing the result of this test suite. If the test suite is something we can run, then it would also be useful for validating other implementations. Would this be possible.

We have a python test script that we use to validate against our illuminator modules. Does that work? If so, is there somewhere I should upload the python script and logs to or just attach them here?

> When this is done, do you think you could create a simple microservice doc following the [same pattern as this one](https://mavlink.io/en/services/terrain.html) for the illuminators? We didn't need when this was just a single command, but now that the interactions are a little more complex it would be good to document this?

What would be a good way to share the microservice doc? Should I open a PR on https://github.com/mavlink/mavlink-devguide or attach the markdown file here?

Also is there a way to reserve message ID? I see a recently-merged message in `development.xml` is now using the message ID we were planning on using for `ILLUMINATOR_STATUS`. We will be locking down our illuminator firmware soon so want to make sure that whatever ends up going into `common.xml` is consistent with our illuminator firmware.

hamishwillee:
THanks @amy-chen-skydio 

> We have a python test script that we use to validate against our illuminator modules. Does that work? If so, is there somewhere I should upload the python script and logs to or just attach them here?

Yes, please attach them here, along with any instructions needed to use/interpret them.

> What would be a good way to share the microservice doc? Should I open a PR on https://github.com/mavlink/mavlink-devguide or attach the markdown file here?

Opening a PR would be excellent.

> Also is there a way to reserve message ID? I see a recently-merged message in development.xml is now using the message ID we were planning on using for ILLUMINATOR_STATUS. We will be locking down our illuminator firmware soon so want to make sure that whatever ends up going into common.xml is consistent with our illuminator firmware.

That's annoying - sorry. Usually PRs don't overlap each other. 

HOw long before you think you will have the test data vs when you need the Id? 
If you're going to have that data soon you won't need the ID. If you are going to be a while I'd suggest pushing your new messages into development.xml, and we can merge early. Then allow it to move to common.xml when the results are provided.

amy-chen-skydio:
@hamishwillee wanted to follow up to see if you have had a chance to review the documents I attached above. We can also attend the dev call meeting on 4/10 to help answer any additional questions you have.

hamishwillee:
Hi @amy-chen-skydio 
Sorry for delay - I only work on this project a few days a week. Will look at this today/tomorrow.

hamishwillee (approved):
@julianoes This all looks good to me - @amy-chen-skydio has provided pymavlink test scripts for both sides of the illuminator interaction. It's not as good as having a flight stack to test against, but it seems to me that this is what we asked for, and about the best we can hope for when we're not working against open source. Do you have any other comments before I merge?

@amy-chen-skydio Thank you for this. When you add the microservice docs, I propose we put the scripts under https://github.com/mavlink/mavlink-devguide/tree/master/assets in a sub folder `services/illuminators/` - then you can reference them in your docs in some way - say as a see -also.

julianoes (approved):
Thanks for all the work and follow ups!

hamishwillee:
Thanks @julianoes @amy-chen-skydio . Merging now. Please post the corresponding docs in a timely manner :-)

FYI only we discussed the test code in the call last night. MAVLink probably need to formalise the requirements here so that next time it is easier. For example, it might make sense to require test code as MAVSDK plugins - that makes it easy to write both the client and server side code in a way that can then be taken by others and adapted for their needs.

Issue #2043, opened 2023-10-09
Every geofence with its own action


Igor-Misic:
Can we add this feature to geofences? (@hamishwillee I have seen your comment)

https://github.com/mavlink/mavlink/blob/b295e70eefa5eb4eee4cf6531e33f4b4c4e472d6/message_definitions/v1.0/development.xml#L200

I was thinking of adding `fenceAction` to param7 here:
https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2340
https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2351
https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2362
https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2373

With this, we can select different actions in QGC and then send them to PX4. When geofence is breached we can do different actions.

hamishwillee:
Hi @Igor-Misic 

The command [MAV_CMD_SET_FENCE_BREACH_ACTION](https://mavlink.io/en/messages/development.html#MAV_CMD_SET_FENCE_BREACH_ACTION) is supposed to do what you need - i.e. it is **how** you add an action for each separate fence.

> As part of a mission protocol plan it sets the fence action for the _next complete geofence definition_ *after* the command. 
> - Note: A fence action defined in a plan will override the default system setting (even if the system-default is [FENCE_ACTION_NONE](https://mavlink.io/en/messages/development.html#FENCE_ACTION_NONE)
> - Note: Every geofence in a plan can have its own action; if no fence action is defined for a particular fence the system-default will be used. 
> - Note: The flight stack should reject a plan or command that uses a geofence action that it does not support and send a STATUSTEXT with the reason.

Basically you add this item before you declare a circular or polygon fence and it applies only to the next fence. If you have a fence definition made up of a bunch of [MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION](https://mavlink.io/en/messages/common.html#MAV_CMD_NAV_FENCE_POLYGON_VERTEX_INCLUSION) you would declare this once and it would apply for the whole fence that follows, but not (say) the next circular fence.

I think this is better than using a reserved part of a vertix definition because a polygon is made up of multiple vertices, and the rules would necessarily be more complex.

No one has implemented this which is why it is in development.xml. That means we can discuss varying it if there is something flawed in the approach proposed.

[...]

PR #2048, opened 2023-10-16
Re-use param index 3 for Fence Action

Igor-Misic:
This PR is connected to: https://github.com/mavlink/mavlink/issues/2043

It adds the option to use `index param 3` for sending fence action, so each geofence can have its breach action. 

Feasibility development was done and with these changes, we can have multiple steps of breaching actions. One useful example is a case where the vehicle flies uncontrollably outside the last fence we can terminate the vehicle, to prevent the vehicle from entering a forbidden area, etc.
image

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2333,7 +2333,7 @@
         </description>
         <param index="1" label="Vertex Count" minValue="3" increment="1">Polygon vertex count</param>
         <param index="2" label="Inclusion Group" minValue="0" increment="1">Vehicle must be inside ALL inclusion zones in a single group, vehicle must be inside at least one group, must be the same for all points in each polygon</param>
-        <param index="3">Reserved</param>
+        <param index="3" label="Fence Action" minValue="0" increment="1">Action the vehicle will take when the fence is breached.</param>
         <param index="4">Reserved</param>
         <param index="5" label="Latitude">Latitude</param>
         <param index="6" label="Longitude">Longitude</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2333,7 +2333,7 @@
         </description>
         <param index="1" label="Vertex Count" minValue="3" increment="1">Polygon vertex count</param>
         <param index="2" label="Inclusion Group" minValue="0" increment="1">Vehicle must be inside ALL inclusion zones in a single group, vehicle must be inside at least one group, must be the same for all points in each polygon</param>
-        <param index="3">Reserved</param>
+        <param index="3" label="Fence Action" minValue="0" increment="1">Action the vehicle will take when the fence is breached.</param>
```
hamishwillee, line 2336: 1. Should be an enum (true for all cases)
2. Make it clear that in polygon all values should be the same as each other, or will be ignored.
3. Allow for the 0 case to mean system default.

hamishwillee:
@Igor-Misic We discussed this in [20231122](https://github.com/mavlink/mavlink/wiki/20231122-Dev-Meeting) dev call. Essentially we agree in principle but we won't merge until there is a prototype that demonstrates this does indeed do what is needed (i.e. we want to be sure this is really going in before we update common.xml).

Also make sure the existing comments such as https://github.com/mavlink/mavlink/pull/2048#discussion_r1385965434 are addressed.

Igor-Misic:
Thanks for talking about this topic in the discussion. At the moment I don't have the resources to push this upstream, so I'll put this in a pause then. Anyhow I think it was good progress for anyone who will want to implement multiple geofence actions.

hamishwillee:
Thanks @Igor-Misic - that's disappointing, would have been cool. But closing it is the right thing to do if we can't get it in.

PR #2050, opened 2023-10-25
common.xml: Allow FLIGHT_INFORMATION times to be either since epoch or boot

nexton-winjeel:
Using UTC timestamps leads to invalid information on a GCS if running a simulator faster than real-time. In this case, it is better to report arming and takeoff times as time since boot. The GCS can then calculate flight time as delta between `time_boot_ms` and `takeoff_time_utc`.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6828,8 +6828,8 @@
         This can be requested using MAV_CMD_REQUEST_MESSAGE.
       </description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint64_t" name="arming_time_utc" units="us" invalid="0">Timestamp at arming (time since UNIX epoch) in UTC, 0 for unknown</field>
-      <field type="uint64_t" name="takeoff_time_utc" units="us" invalid="0">Timestamp at takeoff (time since UNIX epoch) in UTC, 0 for unknown</field>
+      <field type="uint64_t" name="arming_time_utc" units="us" invalid="0">Timestamp at arming (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number. Set to 0 if unknown or not armed.</field>
+      <field type="uint64_t" name="takeoff_time_utc" units="us" invalid="0">Timestamp at takeoff (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number. Set to 0 if unknown or not flying.</field>
       <field type="uint64_t" name="flight_uuid">Universally unique identifier (UUID) of flight, should correspond to name of log files</field>
     </message>
     <message id="265" name="MOUNT_ORIENTATION">


hamishwillee (approved):
This looks sensible, and is what is done elsewhere. I guess in theory someone might be assuming it must be Unix epoc, but I suspect they would do what this changed version says.

Any objection @julianoes @auturgy ?

julianoes:
I think this is masking a different problem. If there is no UTC time on the drone, then maybe that's what's missing. I'd rather not have these fields ambiguous because it requires logic on the parsing side.

nexton-winjeel:
> I think this is masking a different problem. If there is no UTC time on the drone, then maybe that's what's missing.

There are valid use cases where UTC time isn't available on the drone. But in the specific use case I'm targeting, UTC isn't useful.

We're using this message to display to the operator how long the drone has been flying for. We display a timer, not a timestamp, because we don't want the operator to have to calculate flight time when they are concentrating on flying.

If these fields are time since boot, then all the information needed to calculate flight time is present in this one message. If we have to use these fields as UTC timestamps, then we need to calculate the flight time based on either:
* GCS UTC time (which breaks when using a simulator running faster then real-time); or
* AV UTC time as reported in another message (which complicates the system, and can lead to inconsistencies due to stale data).

> I'd rather not have these fields ambiguous because it requires logic on the parsing side.

There are already ~40 other messages that have the same wording, so the logic for this should already exist on the parsing side.

nexton-winjeel:
The other option is to replace the `time_boot_ms` field with a `boot_time_utc` field, but that's more of a breaking change...

auturgy:
Not a request, just raising for discussion: if changing the description is contentious/problematic, perhaps add extensions for the additional options, either as new fields or a bit field/flags.

nexton-winjeel:
@auturgy: Yep, new extension fields would work too.

hamishwillee:
On further thought I'm a little uncomfortable with the redefinition because all the other cases of the text are explicitly named without utc in the field name. If we really need elapsed time of flight, I'd rather we supplied that as a single new extension.

So why did people put UTC in the first place/what information do we want from a message like this?

I mean I would have thought a GCS would just want current flight duration? Why was UTC used - is it needed for logs? And if so, why not land time?

hamishwillee:
@auturgy @julianoes @nexton-winjeel This comes down to a few things IMO 
- do we ever need both UTC and from boot timing information at the same time? I.e. is it useful to know that our arming happened at 5pm on Saturdays AND it happened 5 minutes after booting, and that extra information is worth the extra bytes.
- Whether existing code relies on this being a UTC timestamp - i.e. they just convert to a date and display in logs?

If the answer to those things is probably yes we should add new fields as that is safest. If not, I can live with what the existing PR does as a good compromise. 

My leaning is to add the extra fields. Can you vote for new fields with thumbs up, and for existing PR with thumbs down. Add comment for other option.

nexton-winjeel:
I think the better option is leave this message as-is, and create a new message that does what I need (see: https://github.com/ArduPilot/mavlink/pull/336)

I don't need:
* UTC timestamps;
* microsecond precision; or
* a UUID.

I also don't want to be sending 24 extra bytes of unused data along with the 12 bytes I need.

hamishwillee:
Just FYI for team the message that suits @nexton-winjeel needs is:

```
    <message id="11045" name="FLIGHT_DURATION">
      <description>Information about flight since last arming.</description>
      <field type="uint32_t" name="arming_time_ms" units="ms" invalid="0">Timestamp at arming (time since system boot), 0 for unknown.</field>
      <field type="uint32_t" name="takeoff_time_ms" units="ms" invalid="0">Timestamp at takeoff (time since system boot), 0 for unknown.</field>
    </message>
```

@nexton-winjeel What about landing time, disarming time?

An issue here is that even if you don't need them the other message may well still be sent - the overheads are a lot larger then.
I'm not at all opposed to this if we can agree that it can replace the old message. 
@peterbarker Do you have a view on this? Does ArduPilot care about the actual time of flight in `FLIGHT_INFORMATION` or only the relative timestamps? WOuld you be open to deprecating?

nexton-winjeel:
>What about landing time, disarming time?

They would be useful, but my primary use case for this message is a flight timer that works:
* across re-connections; and
* with a simulator running faster than real-time.

> Does ArduPilot care about the actual time of flight in FLIGHT_INFORMATION or only the relative timestamps?

ArduPilot doesn't currently implement `FLIGHT_INFORMATION`.

hamishwillee:
PX4 does. 
@julianoes WOuld I be correct looking at this https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/streams/FLIGHT_INFORMATION.hpp#L70 that PX4 is emitting time since boot? In that case the only implementation we know about is emitting boot. We could therefore make the change with some confidence.

nexton-winjeel:
> PX4 does. @julianoes WOuld I be correct looking at this https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/streams/FLIGHT_INFORMATION.hpp#L70 that PX4 is emitting time since boot? In that case the only implementation we know about is emitting boot. We could therefore make the change with some confidence.

@hamishwillee: The `time_boot_ms` is time since boot. The other two fields, `arming_time_utc` and `takeoff_time_utc`,  are UTC timestamps in the PX4 implementation. These are the two fields I wanted to change.

nexton-winjeel:
Or is the PX4 `hrt_absolute_time()` function returning a time since boot?

peterbarker:
So I have this branch from 2020 for ArduPilot: https://github.com/ArduPilot/ardupilot/compare/master...peterbarker:ardupilot:wip/flight-information

see the final commit there outlining three problems I saw with that message.

julianoes:
> Or is the PX4 hrt_absolute_time() function returning a time since boot?

Correct.

hamishwillee:
@julianoes and I discussed this in the mav call. There is a lot wrong with this message. It appears to have designed with the intent of being a reliable information tool about flights - i.e. "I took a flight on this "absolute time" and it had this UUID" and you can see what my arming and takeoff time was. But it doesn't do a great job at that, and most flight stacks aren't supply UTC or UUID.

 We can either create a replacement or extend it. @peterbarker My assumption is that you would rather extend than have a new message.
 
In that case the current proposal is this.
1. Revert current changes
2. Add new fields arm_time_boot, takeoff_time_boot, land_time_boot that are uint32 and miliseconds. These are zero when not valid - i.e. on boot all would be zero
3. Add a note that the UUID should be a UUID but isn't big enough and been used by some flight stacks as a sequential number - I'm looking at you PX4.

Does that make sense @peterbarker @julianoes . If there is an appetite for a new message I'd suggest we just have arm_time_boot etc with the sole purpose of tracking flight durations

message_definitions/v1.0/common.xml
```diff
@@ -6828,8 +6828,8 @@
...
       </description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint64_t" name="arming_time_utc" units="us" invalid="0">Timestamp at arming (time since UNIX epoch) in UTC, 0 for unknown</field>
-      <field type="uint64_t" name="takeoff_time_utc" units="us" invalid="0">Timestamp at takeoff (time since UNIX epoch) in UTC, 0 for unknown</field>
+      <field type="uint64_t" name="arming_time_utc" units="us" invalid="0">Timestamp at arming (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number. Set to 0 if unknown or not armed.</field>
+      <field type="uint64_t" name="takeoff_time_utc" units="us" invalid="0">Timestamp at takeoff (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude the number. Set to 0 if unknown or not flying.</field>
       <field type="uint64_t" name="flight_uuid">Universally unique identifier (UUID) of flight, should correspond to name of log files</field>
```
hamishwillee, line 6833: Discussed with @peterbarker . Apparently ArduPilot do not use this. So another option is to document the PX4 behaviour and live with the awful names. Thoughts @julianoes @auturgy ?

```suggestion
      <field type="uint64_t" name="arming_time_utc" units="us" invalid="0">Timestamp at arming (since system boot). Set to 0 if unknown or not armed. Note, field is misnamed UTC.</field>
      <field type="uint64_t" name="takeoff_time_utc" units="us" invalid="0">Timestamp at takeoff (since system boot). Set to 0 if unknown or not flying. Note, field is misnamed UTC.</field>
      <field type="uint64_t" name="flight_uuid">Identifier for flight. Incremented for each flight.</field>
      <extensions/>
      <field type="uint32_t" name="takeoff_time_boot_ms" units="ms">Timestamp at landing (since system boot). Set to 0 at boot, and reset to zero on arming. </field>
```

hamishwillee:
@nexton-winjeel I'm closing this because it was too painful to edit, and because the discussion is now long. For the current proposal see #2067

PR #2055, opened 2023-11-13
development: add setter/getters for torque and thrust setpoints

sfuhrer:
On PX4 we removed support for [SET_ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#SET_ACTUATOR_CONTROL_TARGET) and [ACTUATOR_CONTROL_TARGET](https://mavlink.io/en/messages/common.html#ACTUATOR_CONTROL_TARGET) in https://github.com/PX4/PX4-Autopilot/pull/20837. Reason is that we now have clean interfaces into the mixer (allocator), that doesn't go through the legacy "actuator_controls" anymore. 

In this PR I propose a way to bring back messages to set and get torque and thrust setpoints. Instead of the ambiguous previous way it's now messages with clear physical meanings.

What we still would need to align on:
- should we add option to specify if setpoints are absolute (in SI units) or normalized?
- Are SET_ACTUATOR_CONTROL_TARGET and ACTUATOR_CONTROL_TARGET used by Ardupilot or could we deprecate them?

I know that removing fields from existing messages is quite impossible, but theoretically these two messages could remove the thrust fields from https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET and https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET. Attitude and thrust should really be decoupled.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -553,5 +553,37 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="150" name="SET_TORQUE_SETPOINT">
+      <description>Set the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="151" name="GET_TORQUE_SETPOINT">
+      <description>Get the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="152" name="SET_THRUST_SETPOINT">
+      <description>Set the (normalized) thrust setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">thrust setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="153" name="GET_THRUST_SETPOINT">
+      <description>Get the (normalized) thrust setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">thrust setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
+    </message>
   </messages>
 </mavlink>


message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="150" name="SET_TORQUE_SETPOINT">
+      <description>Set the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
```
hamishwillee, line 560: This says "MAV_FRAME" but you have specified in body frame. If you mean [MAV_FRAME_BODY_FRD](https://mavlink.io/en/messages/common.html#MAV_FRAME_BODY_FRD) can we change to that everywhere?

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="150" name="SET_TORQUE_SETPOINT">
```
hamishwillee, line 556: The ID is in the reserved range. Unless this needs to be here, can we find a place above 256 for these?

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+      <description>Set the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="151" name="GET_TORQUE_SETPOINT">
```
hamishwillee, line 564: By convention getters don't have get
```suggestion
    <message id="151" name="TORQUE_SETPOINT">
```

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+    </message>
+    <message id="152" name="SET_THRUST_SETPOINT">
+      <description>Set the (normalized) thrust setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">thrust setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="153" name="GET_THRUST_SETPOINT">
```
hamishwillee, line 580: Convention
```suggestion
    <message id="153" name="THRUST_SETPOINT">
```

hamishwillee:
@tridge @auturgy Do you have any opinions/thoughts on this?

hamishwillee:
PS We should deprecate the old messages too.

message_definitions/v1.0/development.xml
sfuhrer: sure, can be anywhere, I don't know how the ID assignment should work.

message_definitions/v1.0/development.xml
sfuhrer: Probably we could fix it to FRD, but curious to other inputs.

message_definitions/v1.0/development.xml
hamishwillee, line 556:
```suggestion
    <message id="405" name="SET_TORQUE_SETPOINT">
```

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
     </message>
+    <message id="150" name="SET_TORQUE_SETPOINT">
+      <description>Set the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="151" name="TORQUE_SETPOINT">
```
hamishwillee, line 564:
```suggestion
    <message id="406" name="TORQUE_SETPOINT">
```

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+      <description>Get the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="152" name="SET_THRUST_SETPOINT">
```
hamishwillee, line 572:
```suggestion
    <message id="407" name="SET_THRUST_SETPOINT">
```

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+      <description>Set the (normalized) thrust setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">thrust setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="153" name="THRUST_SETPOINT">
```
hamishwillee, line 580:
```suggestion
    <message id="408" name="THRUST_SETPOINT">
```

message_definitions/v1.0/development.xml
hamishwillee: @sfuhrer I have reassigned. Ultimately this has to go to common.xml so you just find a value in the allowed range for common https://mavlink.io/en/guide/define_xml_element.html#message_id_ranges . Normally we'd also add a placeholder in common to indicate that these are reserved.

FYI only the lower range up to 256 is reserved for MAVLink 1 messages. The range was so small that it was partially _shared_ between flight stacks and there can still be clashes if you allocate there. Generally we just keep it clear and allocate in a convenient space somewhere above. I tend to leave blocks of 5 slots free between new messages. In this case I have put the messages in one of those blocks.

sfuhrer: sweet, thanks for the explanation as well!

auturgy:
@lthall do you have a thought on this?

julianoes:
@sfuhrer what's the reasoning of having two split messages rather than one combined one? Wouldn't you always want to send/set both at the same time?

julianoes:
@sfuhrer :wave:

sfuhrer:
Thanks for the nudge @julianoes ! Yes we could think about combining thrust and torque into one message. Though in general MAVLink commands should be as atomic as possible, and thrusts and torques are in the physical world two unrelated things. 
I'm not strongly opinionated here, if anybody is please let me know and I change the PR.

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="408" name="THRUST_SETPOINT">
+      <description>Get the (normalized) thrust setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">thrust setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
```
hamishwillee, line 586: Do all of the values have to be set if the message is sent? I presume so, otherwise we should specify an invalid value. 

For all values, if they are normalized we should specify values for minValue and maxValue. Can you do that?

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="405" name="SET_TORQUE_SETPOINT">
+      <description>Set the (normalized) torque setpoint in body frame</description>
```
hamishwillee, line 557: If we decide on another frame this text will need to change from "body".

hamishwillee:
I'm OK with them split, in particular if they are not things that need to be set at the same time. 

But for this to merge we'd need to confirm the frame behaviour as per my other notes

@lthall do you have a thought on the messages?

message_definitions/v1.0/development.xml
sfuhrer: >Do all of the values have to be set if the message is sent?

I would say yes, all or none.

sfuhrer:
BTW, if we think it's a real usecase to work with absolute instead of normalized values, we could add a field that specifies that.

message_definitions/v1.0/development.xml
hamishwillee: > Done, fixed to FRD

@sfuhrer It looks to me like you've made it variable for the torque `<field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>`, but made it fixed to FRD for the other cases. Is that right?

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
+    </message>
+    <message id="406" name="TORQUE_SETPOINT">
+      <description>Get the (normalized) torque setpoint in body frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" name="x">Torque setpoint around x in MAV_FRAME</field>
+      <field type="float" name="y">Torque setpoint around y in MAV_FRAME</field>
+      <field type="float" name="z">Torque setpoint around z in MAV_FRAME</field>
```
hamishwillee, line 570: Right?
```suggestion
      <field type="float" name="x">Torque setpoint around x in MAV_FRAME_BODY_FRD</field>
      <field type="float" name="y">Torque setpoint around y in MAV_FRAME_BODY_FRD</field>
      <field type="float" name="z">Torque setpoint around z in MAV_FRAME_BODY_FRD</field>
```

message_definitions/v1.0/development.xml
hamishwillee, line 586:
```suggestion
      <field type="float" name="x">thrust setpoint around x in MAV_FRAME_BODY_FRD</field>
      <field type="float" name="y">thrust setpoint around y in MAV_FRAME_BODY_FRD</field>
      <field type="float" name="z">thrust setpoint around z in MAV_FRAME_BODY_FRD</field>
    </message>
    <message id="408" name="THRUST_SETPOINT">
      <description>Get the (normalized) thrust setpoint in body frame</description>
      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
      <field type="float" name="x">thrust setpoint around x in MAV_FRAME</field>
      <field type="float" name="y">thrust setpoint around y in MAV_FRAME</field>
      <field type="float" name="z">thrust setpoint around z in MAV_FRAME</field>
```

message_definitions/v1.0/development.xml
sfuhrer: Sorry, it looks like I forgot to push my changes. Done now, please have another look!

hamishwillee:
Looks good to me, but needs to have the formatting script run against it. @auturgy @julianoes Can we get any further comment on this or are we good to merge into development.xml?

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="405" name="SET_TORQUE_SETPOINT">
+      <description>Set the (normalized) torque setpoint in body (FRD) frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
```
julianoes, line 558: What is `time_boot_ms` for a setpoint? Have we done this anywhere else? What would we use?

It's clear coming from the autopilot but less so to the autopilot.

I see that we have the same for [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET), so maybe it's ok like that.

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" minValue="-1" maxValue="1" name="x">Torque setpoint around body-x</field>
+      <field type="float" minValue="-1" maxValue="1" name="y">Torque setpoint around body-y</field>
+      <field type="float" minValue="-1" maxValue="1" name="z">Torque setpoint around body-z</field>
```
julianoes, line 562: Can you give me an intuition what "normalized torque" is? I see that you would potentially also suggest SI units.

So for [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) we set a quaternion and rad/s.

With this, do we essentially set rad/s^2 or not? Or why is it normalized? Is the ROS world normalized, or the PX4 interface?

message_definitions/v1.0/development.xml
```diff
@@ -553,5 +553,37 @@
...
+    </message>
+    <message id="407" name="SET_THRUST_SETPOINT">
+      <description>Set the (normalized) thrust setpoint in body (FRD) frame</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="frame" enum="MAV_FRAME">Coordinate frame used for following fields.</field>
+      <field type="float" minValue="-1" maxValue="1" name="x">thrust setpoint around body-x</field>
+      <field type="float" minValue="-1" maxValue="1" name="y">thrust setpoint around body-y</field>
+      <field type="float" minValue="-1" maxValue="1" name="z">thrust setpoint around body-z</field>
```
julianoes, line 578: Same question here: how do I think about the units? And could a flight stack potentially support m/s^2?

julianoes (commented):
Sorry, I'm bringing up more questions as I'm reviewing this again.

message_definitions/v1.0/development.xml
sfuhrer: I just copied it, have no context.

message_definitions/v1.0/development.xml
julianoes: Right. Does the ROS counter part have units? There is an analog ROS message, isn't there?

sfuhrer: I don't know, would need to find it out as well, personally I don't use any of it. Was merely reacting to issues of others, eg https://github.com/PX4/PX4-Autopilot/issues/22265.

hamishwillee: OK, Hopefully @eng-86 might be able to comment (I added request for help in 
https://github.com/PX4/PX4-Autopilot/issues/22265#issuecomment-1863703235 )

julianoes: If it's not required I'm ok with this.

julianoes:
Something wrong with `maxValue` @hamishwillee ?

> /home/runner/work/mavlink/mavlink/message_definitions/v1.0/development.xml:574:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'maxValue': The attribute 'maxValue' is not allowed.

hamishwillee:
Not allowed on fields. STILL waiting on review https://github.com/ArduPilot/pymavlink/pull/811

message_definitions/v1.0/development.xml
eng-86: > What is `time_boot_ms` for a setpoint? Have we done this anywhere else? What would we use?
> 
> It's clear coming from the autopilot but less so to the autopilot.
> 
> I see that we have the same for [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET), so maybe it's ok like that.

It is probably needed for knowing the initial time to apply integration and differentiation  for PID control.

message_definitions/v1.0/development.xml
eng-86: > Can you give me an intuition what "normalized torque" is? I see that you would potentially also suggest SI units.
> 
> So for [SET_ATTITUDE_TARGET](https://mavlink.io/en/messages/common.html#SET_ATTITUDE_TARGET) we set a quaternion and rad/s.
> 
> With this, do we essentially set rad/s^2 or not? Or why is it normalized? Is the ROS world normalized, or the PX4 interface?

Since Ardupilot or PX4 firmware support different vehicles and it is not related to a specific hardware, normalizing the torques and thrusts setpoints will preserve this generality. Although in terms of research and development domains it is better not to normalize these values, this will help for well analyzing the dynamics of a specified vehicle or frame. It is a great idea to have two options (Normalized or Standard). Implementing standard methods (SI units) is more complex than normalized approach in my opinion, the control of the actuator is usually in an open loop manner. So, you must specify the dynamic parameters, actuator gain, .......
The case of attitude control is different, it is a close loop controller. 
@sfuhrer what do you think? We need more help from control system contributors.

message_definitions/v1.0/development.xml
eng-86: > Same question here: how do I think about the units? And could a flight stack potentially support m/s^2?

same answer as before

message_definitions/v1.0/development.xml
eng-86: > Right. Does the ROS counter part have units? There is an analog ROS message, isn't there?

MAVROS does not specify the unit, only **float32[8] controls**, an array of control inputs, I usually (in previous versions) use first three elements for roll, pitch, and yaw torques, and fourth term for thrust with respect to gravity vector https://docs.ros.org/en/api/mavros_msgs/html/msg/ActuatorControl.html

PR #2057, opened 2023-11-22
moved MAV_CMD_EXTERNAL_POSITION_ESTIMATE from ardupilotmega to common

RomanBapst:
Looking  for some guidance on if I should keep the original ID from the ardupilot dialect or if it should be changed.
I am aware of the ID ranges, however, I noticed there are other messages which were moved to common but the ID was kept.

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -329,6 +329,7 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <!-- 43003 MAV_CMD_EXTERNAL_POSITION_ESTIMATE moved to common.xml -->
     </enum>
     <enum name="SCRIPTING_CMD">
       <entry value="0" name="SCRIPTING_CMD_REPL_START">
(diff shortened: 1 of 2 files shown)


message_definitions/v1.0/ardupilotmega.xml
```diff
@@ -329,16 +329,6 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="43003" name="MAV_CMD_EXTERNAL_POSITION_ESTIMATE" hasLocation="true" isDestination="false">
```
hamishwillee, line 332: This looks good but add a note here like `<!-- id 43003 reserved (MAV_CMD_EXTERNAL_POSITION_ESTIMATE) moved to common.xml -->`

hamishwillee: ^^^ This makes it easier for them not to fill that id.

Do something similar (copied from) in  common.xml so it is obvious why the id is odd.

RomanBapst: @hamishwillee Done

auturgy (approved):

hamishwillee:
Thanks @RomanBapst - we discussed in the dev call and this is good.

PR #2059, opened 2023-11-28
common.xml: remove validity requirement for COMMAND_INT_ONLY

peterbarker:
... and COMMAND_LONG_ONLY

It doesn't really matter why the command is being rejected, just that it is.

Adding constraints about why an autopilot might send the message reduces the usefulness of this.

I hit this problem when writing https://github.com/ArduPilot/ardupilot/pull/25643 .  In this case the new define in ArduPilot simply removes `COMMAND_LONG` support, so if you set the new define to true in ArduPilot *all* commands will be rejected with "INT_ONLY".  We do not want to check the validity of the commands in this case, just say, "INT_ONLY".

@nexton-winjeel


hamishwillee (approved):

hamishwillee:
Absolutely!

PR #2060, opened 2023-11-29
Fixed deprecation typo for MAV_CMD_SET_MESSAGE_INTERVAL

hamishwillee:
I suspect a typo - REQUEST_DATA_STREAM was deprecated by MAV_CMD_SET_MESSAGE_INTERVAL


julianoes (approved):
I think so.

hamishwillee:
I do. But I'm OK with this because a user needs to know what they should use here.

PR #2063, opened 2023-12-06
Add additional support for parachute

amy-chen-skydio:
- Add messages and flags to report parachute status and information
- Add command `MAV_CMD_SET_PARACHUTE_ARM` to arm parachute with additional triggering methods

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1813,8 +1813,12 @@
       </entry>
       <entry value="216" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
         <description>Command to arm/disarm parachute module</description>
-        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS" >Arm Flags</param>
+        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS">Arm Flags</param>
         <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
+      </entry>
+      <entry value="217" name="MAV_CMD_SET_PARACHUTE_ARM_ALTI" hasLocation="false" isDestination="false">
+        <description>Command to set the parachute's automatic trigger system arming altitude in meters. </description>
+        <param index="1" label="Altitude" units="m">Parachute automatic trigger system arming altitude</param>
       </entry>
       <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
         <deprecated since="2020-01" replaced_by="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW"/>
@@ -5118,40 +5122,40 @@
       <entry value="2" name="PARACHUTE_ARM_FLAGS_AUTOMATIC_TRIGGER_SYSTEM">
         <description>Arming parachute's automatic trigger system</description>
       </entry>
-      <entry value="4" name="PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS">
-        <description>Arming parachute's auto-arming its automatic trigger system capability</description>
-      </entry>
-      <entry value="8" name="PARACHUTE_ARM_FLAGS_AUTO_DISARM_ATS">
-        <description>Arming parachute's auto-disarming its automatic trigger system capability</description>
-      </entry>
-      <entry value="16" name="PARACHUTE_ARM_FLAGS_MANUAL_TRIGGER_DEVICE">
+      <entry value="4" name="PARACHUTE_ARM_FLAGS_MANUAL_TRIGGER_DEVICE">
         <description>Arming parachute's manual trigger device method</description>
       </entry>
-      <entry value="32" name="PARACHUTE_ARM_FLAGS_OBC_TRIGGER_METHOD">
+      <entry value="8" name="PARACHUTE_ARM_FLAGS_OBC_TRIGGER_METHOD">
         <description>Arming parachute's onboard computer trigger method</description>
       </entry>
-      <entry value="64" name="PARACHUTE_ARM_FLAGS_GEOFENCE_TRIGGER_METHOD">
+      <entry value="16" name="PARACHUTE_ARM_FLAGS_GEOFENCE_TRIGGER_METHOD">
         <description>Arming parachute's geofence trigger method</description>
       </entry>
+      <entry value="32" name="PARACHUTE_ARM_FLAGS_RESERVED">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="64" name="PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS">
+        <description>Arming parachute's auto-arming its automatic trigger system capability</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_ARM_FLAGS_AUTO_DISARM_ATS">
+        <description>Arming parachute's auto-disarming its automatic trigger system capability</description>
+      </entry>
     </enum>
-    <enum name="PARACHUTE_DEPLOYMENT_STATUS">
+    <enum name="PARACHUTE_DEPLOYMENT_FLAGS">
       <description>Parachute module deployment flags (bitmap, 0 means no deployment)</description>
-      <entry value="0" name="PARACHUTE_DEPLOYMENT_STATUS_NONE">
-        <description>Parachute has not deployed</description>
-      </entry>
-      <entry value="1" name="PARACHUTE_DEPLOYMENT_STATUS_MANUAL_TRIGGER_DEVICE">
-        <description>Parachute deployed by manual trigger device</description>
-      </entry>
-      <entry value="2" name="PARACHUTE_DEPLOYMENT_STATUS_AUTOMATIC_TRIGGER_SYSTEM">
-        <description>Parachute deployed by automatic trigger system</description>
-      </entry>
-      <entry value="3" name="PARACHUTE_DEPLOYMENT_STATUS_DRONE_TRIGGER_METHOD">
+      <entry value="1" name="PARACHUTE_DEPLOYMENT_FLAGS_DRONE_TRIGGER_METHOD">
         <description>Parachute deployed by drone</description>
       </entry>
-      <entry value="4" name="PARACHUTE_DEPLOYMENT_STATUS_OBC_TRIGGER_METHOD">
+      <entry value="2" name="PARACHUTE_DEPLOYMENT_FLAGS_AUTOMATIC_TRIGGER_SYSTEM">
+        <description>Parachute deployed by automatic trigger system</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_DEPLOYMENT_FLAGS_MANUAL_TRIGGER_DEVICE">
+        <description>Parachute deployed by manual trigger device</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_DEPLOYMENT_FLAGS_OBC_TRIGGER_METHOD">
         <description>Parachute deployed by onboard computer</description>
       </entry>
-      <entry value="5" name="PARACHUTE_DEPLOYMENT_STATUS_GEOFENCE">
+      <entry value="16" name="PARACHUTE_DEPLOYMENT_FLAGS_GEOFENCE">
         <description>Parachute deployed by geofence method</description>
       </entry>
     </enum>
@@ -5195,6 +5199,12 @@
       </entry>
       <entry value="4096" name="PARACHUTE_ERROR_FLAGS_GPS_ERROR">
         <description>There is an error with the parachute GPS</description>
+      </entry>
+      <entry value="8192" name="PARACHUTE_ERROR_FLAGS_GS_ERROR">
+        <description>There is an error with the parachute and GS (ground station) connection</description>
+      </entry>
+      <entry value="16384" name="PARACHUTE_ERROR_FLAGS_SUBSYSTEM_ERROR">
+        <description>There is an error with the parachute and subsystem connection</description>
       </entry>
     </enum>
   </enums>
@@ -7773,7 +7783,7 @@
       <field type="uint8_t" name="battery_status" units="%">Parachute battery percentage</field>
       <field type="uint8_t" name="arm_status" enum="PARACHUTE_ARM_FLAGS" display="bitmask">Parachute arming status</field>
       <field type="uint8_t" name="charge_status">Boolean indicating charging (1) or not (0)</field>
-      <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_STATUS">Parachute deployment_status</field>
+      <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_FLAGS" display="bitmask">Parachute deployment_status</field>
       <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute ATS auto-arming/disarming altitude in meters</field>
     </message>
     <message id="431" name="PARACHUTE_INFORMATION">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1874,7 +1874,7 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="225" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
-        <description>Command to arm/disarm parachute module</description>
+        <description>Command to arm/disarm parachute module trigger sources.</description>
         <param index="1" label="Arm Flags" enum="PARACHUTE_TRIGGER_FLAGS">Arm flags of parachute trigger sources</param>
         <param index="2" label="Bitmask" enum="PARACHUTE_TRIGGER_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
       </entry>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5114,7 +5114,7 @@
         <description>Illuminator thermistor failure.</description>
       </entry>
     </enum>
-    <enum name="PARACHUTE_ARM_FLAGS">
+    <enum name="PARACHUTE_ARM_FLAGS" bitmask="true">
       <description>Parachute module arming flags (bitmap, 0 means no arming enabled)</description>
       <entry value="1" name="PARACHUTE_ARM_FLAGS_DRONE_TRIGGER_METHOD">
         <description>Arming parachute's drone trigger method</description>
@@ -5141,7 +5141,7 @@
         <description>Arming parachute's auto-disarming its automatic trigger system capability</description>
       </entry>
     </enum>
-    <enum name="PARACHUTE_DEPLOYMENT_FLAGS">
+    <enum name="PARACHUTE_DEPLOYMENT_FLAGS" bitmask="true">
       <description>Parachute module deployment flags (bitmap, 0 means no deployment)</description>
       <entry value="1" name="PARACHUTE_DEPLOYMENT_FLAGS_DRONE_TRIGGER_METHOD">
         <description>Parachute deployed by drone</description>
@@ -5159,7 +5159,7 @@
         <description>Parachute deployed by geofence method</description>
       </entry>
     </enum>
-    <enum name="PARACHUTE_ERROR_FLAGS">
+    <enum name="PARACHUTE_ERROR_FLAGS" bitmask="true">
       <description>Parachute module error flags (bitmap, 0 means no error)</description>
       <entry value="1" name="PARACHUTE_ERROR_FLAGS_BAROMETER_ERROR">
         <description>There is an error with the parachute barometer</description>
@@ -7788,8 +7788,6 @@
     </message>
     <message id="431" name="PARACHUTE_INFORMATION">
       <description>Parachute information</description>
-      <field type="uint16_t" name="vendor_id">Vendor ID</field>
-      <field type="uint16_t" name="product_id">Product ID</field>
       <field type="char[16]" name="serial_number">Serial number in ASCII characters, null terminated. All 0: field not provided.</field>
       <field type="uint32_t" name="main_system_fw_version">Parachute main system firmware version, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). Use 0 if not known.</field>
       <field type="uint32_t" name="main_system_hw_version">Parachute main system hardware version, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). Use 0 if not known.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5149,13 +5149,13 @@
         <description>Drone</description>
       </entry>
       <entry value="4" name="PARACHUTE_DEPLOYMENT_TRIGGER_MAVLINK">
-        <description>MAVLink (offboard computer)</description>
+        <description>MAVLink, e.g. from an offboard computer</description>
       </entry>
       <entry value="5" name="PARACHUTE_DEPLOYMENT_TRIGGER_GEOFENCE">
         <description>Geofence</description>
       </entry>
       <entry value="6" name="PARACHUTE_DEPLOYMENT_TRIGGER_FTS_PRECHECKING">
-        <description>FTS (flight termination system) pre-checking protocol</description>
+        <description>Flight Termination System (FTS) pre-checking protocol</description>
       </entry>
     </enum>
     <enum name="PARACHUTE_SAFETY_FLAGS" bitmask="true">
@@ -5179,22 +5179,22 @@
         <description>There is an error with the parachute IMU</description>
       </entry>
       <entry value="4" name="PARACHUTE_ERROR_FLAGS_RF_CONNECTION_ERROR">
-        <description>There is an error with the parachute RF (used for manual control)</description>
+        <description>There is an error with the parachute's RF that is used for manual control</description>
       </entry>
       <entry value="8" name="PARACHUTE_ERROR_FLAGS_LOW_POWER">
         <description>Parachute module has low power</description>
       </entry>
       <entry value="16" name="PARACHUTE_ERROR_FLAGS_FC_CONNECTION_ERROR">
-        <description>There is an error with the parachute and FC (flight controller) connection</description>
+        <description>There is an error with the connection between parachute and flight controller (FC)</description>
       </entry>
       <entry value="32" name="PARACHUTE_ERROR_FLAGS_EFTS_CONNECTION_ERROR">
-        <description>There is an error with the parachute electrical/external FTS (flight termination system)</description>
+        <description>There is an error with the connection between parachute and Electrical Flight Termination System (EFTS)</description>
       </entry>
       <entry value="64" name="PARACHUTE_ERROR_FLAGS_POD_CONNECTION_ERROR">
         <description>There is an error with the parachute pod</description>
       </entry>
       <entry value="128" name="PARACHUTE_ERROR_FLAGS_EFTS_DIAGNOSE">
-        <description>Parachute electrical/external FTS (flight termination system) diagnosis failed</description>
+        <description>Parachute Electrical Flight Termination System (EFTS) diagnosis failed</description>
       </entry>
       <entry value="256" name="PARACHUTE_ERROR_FLAGS_CHARGING_FAILED">
         <description>Parachute module charging failed</description>
@@ -5203,16 +5203,16 @@
         <description>There is an error with the parachute external power source</description>
       </entry>
       <entry value="1024" name="PARACHUTE_ERROR_FLAGS_GS_CONNECTION_ERROR">
-        <description>There is an error with the parachute and GS (ground station) connection</description>
+        <description>There is an error with the connection between parachute and Ground Station (GS)</description>
       </entry>
       <entry value="2048" name="PARACHUTE_ERROR_FLAGS_GPS_ERROR">
-        <description>There is an error with the parachute GPS</description>
+        <description>There is an error with the parachute's GPS</description>
       </entry>
       <entry value="4096" name="PARACHUTE_ERROR_FLAGS_SUBSYSTEM_CONNECTION_ERROR">
-        <description>There is an error with the parachute and subsystem connection</description>
+        <description>There is an error with the connection between parachute and subsystem (e.g. remote controller, expansion board, etc.)</description>
       </entry>
       <entry value="8192" name="PARACHUTE_ERROR_FLAGS_SUBSYSTEM_FW_ERROR">
-        <description>There is an error with the parachute subsystem firmware</description>
+        <description>There is an error with the parachute subsystem firmware (e.g. wrong firmware version)</description>
       </entry>
       <entry value="16384" name="PARACHUTE_ERROR_FLAGS_RESERVED_1">
         <description>Reserved for future use</description>
@@ -5221,19 +5221,19 @@
         <description>Reserved for future use</description>
       </entry>
       <entry value="65536" name="PARACHUTE_ERROR_FLAGS_LOGGING_ERROR">
-        <description>There is an error with the parachute module's logging</description>
+        <description>There is an error with the parachute's internal logging system</description>
       </entry>
       <entry value="131072" name="PARACHUTE_ERROR_FLAGS_MODULE_RETIRED">
-        <description>This parachute module is retired (too many deployments)</description>
+        <description>This parachute module is retired (i.e. too many deployments)</description>
       </entry>
       <entry value="262144" name="PARACHUTE_ERROR_FLAGS_GLOW_WIRE_ERROR">
         <description>There is an error with the parachute glow wire</description>
       </entry>
       <entry value="524288" name="PARACHUTE_ERROR_FLAGS_OFFBOARD_CONNECTION_ERROR">
-        <description>There is an error with the parachute and offboard computer MAVLink connection</description>
+        <description>There is an error with the MAVLink connection between parachute and offboard computer</description>
       </entry>
       <entry value="1048576" name="PARACHUTE_ERROR_FLAGS_IMU_CALIBRATION_ERROR">
-        <description>Parachute module IMU calibration failed</description>
+        <description>Parachute's internal IMU calibration failed</description>
       </entry>
     </enum>
   </enums>
@@ -7813,7 +7813,7 @@
       <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_TRIGGER">Parachute deployment status</field>
       <field type="uint8_t" name="safety_status" enum="PARACHUTE_SAFETY_FLAGS" display="bitmask">Parachute safety status</field>
       <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute Automatic Trigger System (ATS) auto-arming/disarming altitude in meters</field>
-      <field type="char[11]" name="parachute_packed_date">Parachute packed date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
+      <field type="char[11]" name="parachute_packed_date">Parachute packed date (YYYY-MM-DD) in ASCII characters, 0 terminated. All 0: field not provided.</field>
     </message>
     <message id="440" name="ILLUMINATOR_STATUS">
       <description>Illuminator status</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1808,15 +1808,6 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="215" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
-        <description>Command to arm/disarm parachute module</description>
-        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS">Arm Flags</param>
-        <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
-      </entry>
-      <entry value="216" name="MAV_CMD_SET_PARACHUTE_ARM_ALTI" hasLocation="false" isDestination="false">
-        <description>Command to set the parachute's automatic trigger system arming altitude in meters. </description>
-        <param index="1" label="Altitude" units="m">Parachute automatic trigger system arming altitude</param>
-      </entry>
       <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
         <deprecated since="2020-01" replaced_by="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW"/>
         <description>Mission command to control a camera or antenna mount, using a quaternion as reference.</description>
@@ -1881,6 +1872,15 @@
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
+      </entry>
+      <entry value="225" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
+        <description>Command to arm/disarm parachute module</description>
+        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS">Arm Flags</param>
+        <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
+      </entry>
+      <entry value="226" name="MAV_CMD_SET_PARACHUTE_ARM_ALTI" hasLocation="false" isDestination="false">
+        <description>Command to set the parachute's automatic trigger system arming altitude in meters.</description>
+        <param index="1" label="Altitude" units="m">Parachute automatic trigger system arming altitude</param>
       </entry>
       <entry value="240" name="MAV_CMD_DO_LAST" hasLocation="false" isDestination="false">
         <description>NOP - This command is only used to mark the upper limit of the DO commands in the enumeration</description>
@@ -7780,7 +7780,7 @@
       <field type="uint8_t" name="arm_status" enum="PARACHUTE_ARM_FLAGS" display="bitmask">Parachute arming status</field>
       <field type="uint8_t" name="charge_status">Boolean indicating charging (1) or not (0)</field>
       <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_FLAGS" display="bitmask">Parachute deployment_status</field>
-      <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute ATS auto-arming/disarming altitude in meters</field>
+      <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute Automatic Trigger System (ATS) auto-arming/disarming altitude in meters</field>
       <field type="char[11]" name="parachute_packed_date">Parachute packed date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
     </message>
     <message id="440" name="ILLUMINATOR_STATUS">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5137,7 +5137,7 @@
     <enum name="PARACHUTE_DEPLOYMENT_TRIGGER">
       <description>Parachute deployment trigger source</description>
       <entry value="0" name="PARACHUTE_DEPLOYMENT_TRIGGER_NONE">
-        <description>None (not triggered)</description>
+        <description>None (parachute has not deployed)</description>
       </entry>
       <entry value="1" name="PARACHUTE_DEPLOYMENT_TRIGGER_MANUAL">
         <description>Manual</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1808,6 +1808,14 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_PARACHUTE_SHUTDOWN" hasLocation="false" isDestination="false">
+        <description>Command to shutdown the parachute module</description>
+      </entry>
+      <entry value="216" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
+        <description>Command to arm/disarm parachute module</description>
+        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS" >Arm Flags</param>
+        <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
+      </entry>
       <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
         <deprecated since="2020-01" replaced_by="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW"/>
         <description>Mission command to control a camera or antenna mount, using a quaternion as reference.</description>
@@ -5102,6 +5110,93 @@
         <description>Illuminator thermistor failure.</description>
       </entry>
     </enum>
+    <enum name="PARACHUTE_ARM_FLAGS">
+      <description>Parachute module arming flags (bitmap, 0 means no arming enabled)</description>
+      <entry value="1" name="PARACHUTE_ARM_FLAGS_DRONE_TRIGGER_METHOD">
+        <description>Arming parachute's drone trigger method</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_ARM_FLAGS_AUTOMATIC_TRIGGER_SYSTEM">
+        <description>Arming parachute's automatic trigger system</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS">
+        <description>Arming parachute's auto-arming its automatic trigger system capability</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_ARM_FLAGS_AUTO_DISARM_ATS">
+        <description>Arming parachute's auto-disarming its automatic trigger system capability</description>
+      </entry>
+      <entry value="16" name="PARACHUTE_ARM_FLAGS_MANUAL_TRIGGER_DEVICE">
+        <description>Arming parachute's manual trigger device method</description>
+      </entry>
+      <entry value="32" name="PARACHUTE_ARM_FLAGS_OBC_TRIGGER_METHOD">
+        <description>Arming parachute's onboard computer trigger method</description>
+      </entry>
+      <entry value="64" name="PARACHUTE_ARM_FLAGS_GEOFENCE_TRIGGER_METHOD">
+        <description>Arming parachute's geofence trigger method</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_DEPLOYMENT_STATUS">
+      <description>Parachute module deployment flags (bitmap, 0 means no deployment)</description>
+      <entry value="0" name="PARACHUTE_DEPLOYMENT_STATUS_NONE">
+        <description>Parachute has not deployed</description>
+      </entry>
+      <entry value="1" name="PARACHUTE_DEPLOYMENT_STATUS_MANUAL_TRIGGER_DEVICE">
+        <description>Parachute deployed by manual trigger device</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_DEPLOYMENT_STATUS_AUTOMATIC_TRIGGER_SYSTEM">
+        <description>Parachute deployed by automatic trigger system</description>
+      </entry>
+      <entry value="3" name="PARACHUTE_DEPLOYMENT_STATUS_DRONE_TRIGGER_METHOD">
+        <description>Parachute deployed by drone</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_DEPLOYMENT_STATUS_OBC_TRIGGER_METHOD">
+        <description>Parachute deployed by onboard computer</description>
+      </entry>
+      <entry value="5" name="PARACHUTE_DEPLOYMENT_STATUS_GEOFENCE">
+        <description>Parachute deployed by geofence method</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_ERROR_FLAGS">
+      <description>Parachute module error flags (bitmap, 0 means no error)</description>
+      <entry value="1" name="PARACHUTE_ERROR_FLAGS_BAROMETER_ERROR">
+        <description>There is an error with the parachute barometer</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_ERROR_FLAGS_IMU_ERROR">
+        <description>There is an error with the parachute IMU</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_ERROR_FLAGS_RF_ERROR">
+        <description>There is an error with the parachute RF</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_ERROR_FLAGS_LOW_POWER">
+        <description>Parachute module has low power</description>
+      </entry>
+      <entry value="16" name="PARACHUTE_ERROR_FLAGS_FC_CONNECTION_ERROR">
+        <description>There is an error with the parachute and FC (flight controller) connection</description>
+      </entry>
+      <entry value="32" name="PARACHUTE_ERROR_FLAGS_POD_ERROR">
+        <description>There is an error with the parachute pod</description>
+      </entry>
+      <entry value="64" name="PARACHUTE_ERROR_FLAGS_EFTS_ERROR">
+        <description>There is an error with the parachute electrical/external FTS (flight termination system)</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_ERROR_FLAGS_CHARGING_FAILED">
+        <description>Parachute module charging failed</description>
+      </entry>
+      <entry value="256" name="PARACHUTE_ERROR_FLAGS_LOGGING_ERROR">
+        <description>There is an error with the parachute module's logging</description>
+      </entry>
+      <entry value="512" name="PARACHUTE_ERROR_FLAGS_MODULE_RETIRED">
+        <description>This parachute module is retired (too many deployments)</description>
+      </entry>
+      <entry value="1024" name="PARACHUTE_ERROR_FLAGS_EXTERNAL_POWER_ERROR">
+        <description>There is an error with the parachute external power source</description>
+      </entry>
+      <entry value="2048" name="PARACHUTE_ERROR_FLAGS_OBC_CONNECTION_ERROR">
+        <description>There is an error with the parachute and OBC (onboard computer) connection</description>
+      </entry>
+      <entry value="4096" name="PARACHUTE_ERROR_FLAGS_GPS_ERROR">
+        <description>There is an error with the parachute GPS</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <message id="1" name="SYS_STATUS">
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -7532,6 +7637,26 @@
...
+      <field type="uint32_t" name="main_system_hw_version">Parachute main system hardware version, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). Use 0 if not known.</field>
+      <field type="uint32_t" name="subsystem_fw_version">Parachute subsystem firmware version, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). Use 0 if not known.</field>
+      <field type="uint32_t" name="subsystem_hw_version">Parachute subsystem hardware version, encoded as: (Dev &amp; 0xff) &lt;&lt; 24 | (Patch &amp; 0xff) &lt;&lt; 16 | (Minor &amp; 0xff) &lt;&lt; 8 | (Major &amp; 0xff). Use 0 if not known.</field>
+      <field type="char[11]" name="parachute_packed_date">Parachute packed date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
+      <field type="char[11]" name="manufacture_date">Manufacture date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
+    </message>
```
julianoes, line 7659: Instead of having these INFORMATION messages for every type of component, we try to generalize it using generic components. The message that could be is currently in development.xml:
https://mavlink.io/en/messages/development.html#COMPONENT_INFORMATION_BASIC

As soon as there are implementations published, we can move it to common.xml.

message_definitions/v1.0/common.xml
```diff
@@ -7532,6 +7637,26 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
+    <message id="430" name="PARACHUTE_STATUS">
+      <description>Parachute system status</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Uptime in ms</field>
+      <field type="uint32_t" name="error_status" enum="PARACHUTE_ERROR_FLAGS" display="bitmask">Errors</field>
+      <field type="uint8_t" name="battery_status" units="%">Parachute battery percentage</field>
```
julianoes, line 7644: The battery status could probably just be [BATTERY_STATUS](https://mavlink.io/en/messages/common.html#BATTERY_STATUS) from the parachute component.

julianoes (changes requested):
Thanks for the pull request!

Now, we're trying to avoid having specific messages for every component type and try to cover it with more generic component messages. For specific configuration, we can then use parameters. I have made a few inline comments around that.

The risk with specific configuration flags is that every new device/implementation of a parachute system might have slightly different settings and require more flags and status fields.

message_definitions/v1.0/common.xml
hamishwillee: HOwdy @amy-chen-skydio - some questions here for you.

amy-chen-skydio: Sorry for the delay, I have removed `battery_status` from `PARACHUTE_STATUS` in the latest change

message_definitions/v1.0/common.xml
amy-chen-skydio: I have removed `PARACHUTE_INFORMATION`. I added `manufacture_date` to `COMPONENT_INFORMATION_BASIC` since this is generic and can apply to any components?

message_definitions/v1.0/common.xml
```diff
@@ -1764,6 +1764,18 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_PARACHUTE_SHUTDOWN" hasLocation="false" isDestination="false">
+        <description>Command to shutdown the parachute module</description>
+      </entry>
```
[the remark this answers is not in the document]
amy-chen-skydio: I have removed this and will just use `MAV_CMD_PREFLIGHT_REBOOT_SHUTDOWN` instead

hamishwillee:
@julianoes I understand @amy-chen-skydio has modified/responded to your comments. Can you please recheck?

message_definitions/v1.0/development.xml
```diff
@@ -466,6 +466,7 @@
       <field type="char[24]" name="software_version">Software version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
       <field type="char[24]" name="hardware_version">Hardware version. The recommended format is SEMVER: 'major.minor.patch'  (any format may be used). The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
       <field type="char[32]" name="serial_number">Hardware serial number. The field must be zero terminated if it has a value. The field is optional and can be empty/all zeros.</field>
+      <field type="char[11]" name="manufacture_date">Manufacture date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
```
julianoes, line 469: I would probably say UTC time in seconds instead of a verbose date. Also, I wonder if this should be an extension or not.

julianoes (approved):

message_definitions/v1.0/development.xml
hamishwillee: @julianoes Or if we are going to date it them IMO this should be a `char[9]` - possibly as YYYYMMDD.

message_definitions/v1.0/common.xml
```diff
@@ -1764,6 +1764,15 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
+        <description>Command to arm/disarm parachute module</description>
+        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS">Arm Flags</param>
+        <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
+      </entry>
+      <entry value="216" name="MAV_CMD_SET_PARACHUTE_ARM_ALTI" hasLocation="false" isDestination="false">
+        <description>Command to set the parachute's automatic trigger system arming altitude in meters. </description>
```
hamishwillee, line 1773: We'd need to clean up the spurious whitespace - such as that after the full stop.

message_definitions/v1.0/common.xml
```diff
@@ -1764,6 +1764,15 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
+        <description>Command to arm/disarm parachute module</description>
+        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS">Arm Flags</param>
+        <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
```
hamishwillee, line 1770: So how does this work - I mean the enum is the same in both so why can't you just have one of these and set the value to 0 or 1 for the things you want turned on and off? 

Or is this so you don't need to know the other arm flags - you just set what you want to arm (.e.g feature A) and the bitmask, and all the other items stay the same?

message_definitions/v1.0/common.xml
```diff
@@ -1764,6 +1764,15 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
```
potaito, line 1767: Shouldn't the arming / disarming be done through the message
https://mavlink.io/en/messages/common.html#MAV_CMD_COMPONENT_ARM_DISARM

and the flags turned into parameters, as @julianoes suggested?

message_definitions/v1.0/common.xml
```diff
@@ -1764,6 +1764,15 @@
...
+        <description>Command to arm/disarm parachute module</description>
+        <param index="1" label="Arm Flags" enum="PARACHUTE_ARM_FLAGS">Arm Flags</param>
+        <param index="2" label="Bitmask" enum="PARACHUTE_ARM_FLAGS">Bitmask of arm flags (in param1) that should be modified</param>
+      </entry>
+      <entry value="216" name="MAV_CMD_SET_PARACHUTE_ARM_ALTI" hasLocation="false" isDestination="false">
```
potaito, line 1772: As far as I can tell, MAVLink tries to only have SET_ messages for things that are expected to change on every flight, maybe even multiple times, which is why parameters are unsuited. Example:
- [MAV_CMD_SET_CAMERA_MODE](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_MODE)
- [MAV_CMD_SET_CAMERA_ZOOM](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_ZOOM)
- [MAV_CMD_SET_CAMERA_FOCUS](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_FOCUS)

Is the parachute arming altitude something that the user is going to modify before every flight? And what's the reference frame for MAV_CMD_SET_PARACHUTE_ARM_ALTI, above sea level or above takeoff / home? I assume the latter, which means that the setting should not change often, meaning that parameters would also work for this.

message_definitions/v1.0/common.xml
```diff
@@ -7532,6 +7634,16 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
+    <message id="430" name="PARACHUTE_STATUS">
+      <description>Parachute system status</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Uptime in ms</field>
+      <field type="uint32_t" name="error_status" enum="PARACHUTE_ERROR_FLAGS" display="bitmask">Errors</field>
+      <field type="uint8_t" name="arm_status" enum="PARACHUTE_ARM_FLAGS" display="bitmask">Parachute arming status</field>
+      <field type="uint8_t" name="charge_status">Boolean indicating charging (1) or not (0)</field>
```
potaito, line 7642: `charge_status` does not seem specific to parachutes. Shouldn't the parachute publish [BATTERY_STATUS](https://mavlink.io/en/messages/common.html#BATTERY_STATUS) as mentioned in the previous comments instead?

message_definitions/v1.0/common.xml
```diff
@@ -7532,6 +7634,16 @@
...
+      <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_FLAGS" display="bitmask">Parachute deployment_status</field>
+      <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute ATS auto-arming/disarming altitude in meters</field>
```
potaito, line 7644: I don't think Automatic Trigger System (ATS) is a common abbreviation.

```suggestion
      <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute Automatic Trigger System (ATS) auto-arming/disarming altitude in meters</field>
```

message_definitions/v1.0/development.xml
potaito: Wouldn't a UINT64 unix timestamp give us more resolution with fewer bytes (8)?

Plus unix timestamp is a recognized international format. DD/MM/YYYY on the other hand is a localized format, not used in all countries.

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+        <description>There is an error with the parachute module's logging</description>
+      </entry>
+      <entry value="512" name="PARACHUTE_ERROR_FLAGS_MODULE_RETIRED">
+        <description>This parachute module is retired (too many deployments)</description>
+      </entry>
+      <entry value="1024" name="PARACHUTE_ERROR_FLAGS_EXTERNAL_POWER_ERROR">
+        <description>There is an error with the parachute external power source</description>
+      </entry>
+      <entry value="2048" name="PARACHUTE_ERROR_FLAGS_OBC_CONNECTION_ERROR">
+        <description>There is an error with the parachute and OBC (onboard computer) connection</description>
```
potaito, line 5079: - "onboard" is typically used for computers inside of the component. For instance a camera that has its own compute unit that can do more than just store a photo on an SD card.
- "offboard" was historically used for a standalone computer that is not the autopilot.

That's why I'm not sure what the intended meaning in this PR is.

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+        <description>Parachute deployed by manual trigger device</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_DEPLOYMENT_FLAGS_OBC_TRIGGER_METHOD">
+        <description>Parachute deployed by onboard computer</description>
+      </entry>
+      <entry value="16" name="PARACHUTE_DEPLOYMENT_FLAGS_GEOFENCE">
+        <description>Parachute deployed by geofence method</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_ERROR_FLAGS" bitmask="true">
```
potaito, line 5043: Most of these errors are already captured in [SYS_STATUS](https://mavlink.io/en/messages/common.html#SYS_STATUS), which is also used by the autopilot to communicate issues with Baro, IMU, GPS etc.

potaito (changes requested):

message_definitions/v1.0/development.xml
julianoes: @potaito exactly!

message_definitions/v1.0/common.xml
julianoes: Good point!

message_definitions/v1.0/development.xml
hamishwillee: The only reason for a stamp in this format is human readability, which need not be a deciding factor. This sounds generally useful, and as there are concerns with this PR, let's add it separately https://github.com/mavlink/mavlink/pull/2078

amy-chen-skydio: Agreed and will remove this change in this PR

message_definitions/v1.0/common.xml
amy-chen-skydio: Yes, that's exactly it. With both, we will be able to just arm/disarm whatever features you want without while the other items just stay the same.

message_definitions/v1.0/common.xml
amy-chen-skydio: Made the change in the latest commit!

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+      <entry value="1" name="PARACHUTE_ARM_FLAGS_DRONE_TRIGGER_METHOD">
+        <description>Arming parachute's drone trigger method</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_ARM_FLAGS_AUTOMATIC_TRIGGER_SYSTEM">
+        <description>Arming parachute's automatic trigger system</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_ARM_FLAGS_MANUAL_TRIGGER_DEVICE">
+        <description>Arming parachute's manual trigger device method</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_ARM_FLAGS_OBC_TRIGGER_METHOD">
```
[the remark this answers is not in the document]
jenke-skydio: this is intended to be the drone's onboard computer

hamishwillee:
@potaito @julianoes There has been a round of changes to this. Can you please re-review and confirm that your concerns have been addressed, and highlight any that have not.

@amy-chen-skydio Thanks for your patience. When this converges, I have the same comments to this as in this comment for the illuminators PR (how to validate/documentation required): https://github.com/mavlink/mavlink/pull/2047#issuecomment-1944819358

jenke-skydio (commented):
Addressed all open comments.  Please take a look and Amy can make changes accordingly

hamishwillee:
@julianoes @potaito Can you please respond to @jenke-skydio 's comments?

amy-chen-skydio:
> @julianoes @potaito Can you please respond to @jenke-skydio 's comments?

@potaito Wanted to follow up to see if you've had a chance to look over @jenke-skydio's comments. Thank you!

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+        <description>Arming parachute's onboard computer trigger method</description>
+      </entry>
+      <entry value="16" name="PARACHUTE_ARM_FLAGS_GEOFENCE_TRIGGER_METHOD">
+        <description>Arming parachute's geofence trigger method</description>
+      </entry>
+      <entry value="32" name="PARACHUTE_ARM_FLAGS_RESERVED">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="64" name="PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS">
+        <description>Arming parachute's auto-arming its automatic trigger system capability</description>
```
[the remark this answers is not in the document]
potaito: No, I still don't get it :sweat_smile: 

> auto-arm mode ... is armed

Is the auto-arm mode enabled? Is that what's intended?

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+    <enum name="PARACHUTE_ERROR_FLAGS" bitmask="true">
+      <description>Parachute module error flags (bitmap, 0 means no error)</description>
+      <entry value="1" name="PARACHUTE_ERROR_FLAGS_BAROMETER_ERROR">
+        <description>There is an error with the parachute barometer</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_ERROR_FLAGS_IMU_ERROR">
+        <description>There is an error with the parachute IMU</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_ERROR_FLAGS_RF_ERROR">
+        <description>There is an error with the parachute RF</description>
```
potaito, line 5052: What's "RF" ?

message_definitions/v1.0/common.xml
potaito: Fair enough!

message_definitions/v1.0/common.xml
potaito: @jenke-skydio how is the connection to the offboard computer defined? How should a parachute monitor its state? Is the parachute expected to look for heartbeat messages that have the [MAV_TYPE_ONBOARD_CONTROLLER](https://mavlink.io/en/messages/common.html#MAV_TYPE_ONBOARD_CONTROLLER) type?

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="64" name="PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS">
+        <description>Arming parachute's auto-arming its automatic trigger system capability</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_ARM_FLAGS_AUTO_DISARM_ATS">
+        <description>Arming parachute's auto-disarming its automatic trigger system capability</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_DEPLOYMENT_FLAGS" bitmask="true">
```
potaito, line 5025: Just double-checking: `PARACHUTE_DEPLOYMENT_FLAGS` is 0 until the parachute pops, at which point it will have the flag describing the reason for the popping, right?

If that's the case, it should be an enum.

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
jenke-skydio: Yes, this is enabling the auto-arm mode.  How about: "Enabling parachute automatic trigger system's (ATS) auto-arm mode".

When parachute ATS is in auto-arm mode, for example, a vehicle may takeoff with the ATS disarmed then later detect that the vehicle has gone over a certain altitude threshold and automatically arm.  This would allow the ATS to automatically deploy the chute if frefall was detected.  I think we can make this pretty clear with a state diagram in the MAVLink devguide when we add some documentation for this microservice

message_definitions/v1.0/common.xml
jenke-skydio: this is referring to the radio, we can change to `PARACHUTE_ERROR_FLAGS_RADIO_ERROR` and `There is an error with the parachute radio`

message_definitions/v1.0/common.xml
jenke-skydio: connection to offboard computer is optional, when its there it must send `HEARTBEAT` and optionally `SYSTEM_TIME`

message_definitions/v1.0/common.xml
hamishwillee: There is still a question here about what does a parachute need a radio for. I assume this is a parachute connection for MAVLink? What if the channel is USB? If it is a MAVLink channel, how will this be sent when the link is dead. Note, I have not read this thoroughly, so perhaps answered elsewhere.

message_definitions/v1.0/common.xml
hamishwillee: There isn't really a place in in xml file for this kind of information. I expect you'll need to create a parachute service doc that explains the models and use cases that are addressed.

message_definitions/v1.0/common.xml
hamishwillee: or BATTERY_STATUS_V2 would be better, though not yet supported by anyone (see development.xml). It uses most of the same fields.

message_definitions/v1.0/common.xml
jenke-skydio: its common for a parachute system to consist of a vehicle integrated module with an optional radio as well as an optional transmitter to be used on ground by a visual observer to deploy the chute (this is intended to align with emerging standards such as [ASTM-F3322](https://www.astm.org/f3322-22.html)).  While it is feasible that MAVLink is used for this link, it is not required.  In the case where the system is more of a self-contained system, this error would indicate a radio error internal to the system such that it would be feasible for a the vehicle or ground station to display some status or take some action

message_definitions/v1.0/common.xml
jenke-skydio: yes 100%.  We plan to write a up parachute microservice documentation [in the MAVLink devguide](https://github.com/mavlink/mavlink-devguide/tree/master/en/services)

message_definitions/v1.0/common.xml
potaito: That makes sense. Parachute, integrated system with its own radio :+1:

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
potaito: Yes, this is much better:
> Enabling parachute automatic trigger system's (ATS) auto-arm mode

Than this:
> Arming parachute's auto-arming its automatic trigger system capability

:+1:

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
+    <enum name="PARACHUTE_ARM_FLAGS" bitmask="true">
+      <description>Parachute module arming flags (bitmap, 0 means no arming enabled)</description>
+      <entry value="1" name="PARACHUTE_ARM_FLAGS_DRONE_TRIGGER_METHOD">
+        <description>Arming parachute's drone trigger method</description>
```
hamishwillee, line 5001: This means a control signal - PWM or whatever?

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+      <entry value="32" name="PARACHUTE_ARM_FLAGS_RESERVED">
+        <description>Reserved for future use.</description>
+      </entry>
+      <entry value="64" name="PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS">
+        <description>Arming parachute's auto-arming its automatic trigger system capability</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_ARM_FLAGS_AUTO_DISARM_ATS">
+        <description>Arming parachute's auto-disarming its automatic trigger system capability</description>
+      </entry>
+    </enum>
```
hamishwillee, line 5024: A suggestion - note, query on what FC case means. Also why a manual trigger? Why do you need that?


```suggestion
    <enum name="PARACHUTE_TRIGGER_FLAGS" bitmask="true">
      <description>Enable/disable parachute trigger sources and modes</description>
      <entry value="1" name="PARACHUTE_TRIGGER_FLAGS_FC">
        <description>Flight controller trigger (e.g. PWM, DroneCan, RC Control)</description>
      </entry>
      <entry value="2" name="PARACHUTE_TRIGGER_FLAGS_ATS">
        <description>Automatic trigger system (ATS)</description>
      </entry>
      <entry value="4" name="PARACHUTE_TRIGGER_FLAGS_MANUAL">
        <description>Manual trigger</description>
      </entry>
      <entry value="8" name="PARACHUTE_TRIGGER_FLAGS_MAVLINK">
        <description>MAVLink trigger (offboard computer)</description>
      </entry>
      <entry value="16" name="PARACHUTE_TRIGGER_FLAGS_GEOFENCE">
        <description>Geofence trigger</description>
      </entry>
      <entry value="32" name="PARACHUTE_TRIGGER_FLAGS_RESERVED">
        <description>Reserved for future use.</description>
      </entry>
      <entry value="64" name="PARACHUTE_TRIGGER_FLAGS_ATS_AUTO_ARM">
        <description>Auto-arming of parachute automatic trigger system (ATS)</description>
      </entry>
      <entry value="128" name="PARACHUTE_TRIGGER_FLAGS_ATS_AUTO_DISARM">
        <description>Auto-disarming of parachute automatic trigger system (ATS)</description>
      </entry>
    </enum>
```

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: I have simplified (I hope) the whole thing in https://github.com/mavlink/mavlink/pull/2063/files#r1577246968

> When parachute ATS is in auto-arm mode, for example, a vehicle may takeoff with the ATS disarmed then later detect that the vehicle has gone over a certain altitude threshold and automatically arm. 

If the ATS is doing an altitude check then it could simply decide it won't trigger/fully arm until until it reaches the desired altitude - so you don't need an extra state. After all, you'd have to configure the height to trigger and so on anyway. 
If it is the Vehicle doing the altitude check, then  the vehicle can get to the specific altitude and enable ATS. 

Either way, I don't see why you need it. Though I guess it is more flexible than the idea of `PARACHUTE_ARM_FLAGS_AUTO_ARM_ATS_AT_ALTITUDE

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: @potaito Yes, they are all used interchangeably. 

Historically Offboard and Companion computer, and more recently mission computer, were used to differentiate the general computer from the FC ("the board") on a drone. Onboard computer is used for the same thing, but in this case "board" refers to the fact that the computer travels on the drone.

In future we'll see more "boards" where there is no visible separation between FC and companion, so we do need to think of a consistent terminology to refer to the bit that runs the RTOS/Flight stack, and the bit that provides general computing. 

For now, I'd say Offboard is the most commonly used.  For this case, in my suggestion I've renamed the flag to be "MAVLink".

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4995,99 @@
...
+      <entry value="4" name="PARACHUTE_DEPLOYMENT_FLAGS_MANUAL_TRIGGER_DEVICE">
+        <description>Parachute deployed by manual trigger device</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_DEPLOYMENT_FLAGS_OBC_TRIGGER_METHOD">
+        <description>Parachute deployed by onboard computer</description>
+      </entry>
+      <entry value="16" name="PARACHUTE_DEPLOYMENT_FLAGS_GEOFENCE">
+        <description>Parachute deployed by geofence method</description>
+      </entry>
+    </enum>
```
hamishwillee, line 5042: I agree with @potaito - should be an enum if only one can be true.

```suggestion
    <enum name="PARACHUTE_DEPLOYMENT_TRIGGER" bitmask="true">
      <description>Parachute deployment trigger source</description>
      <entry value="0" name="PARACHUTE_DEPLOYMENT_TRIGGER_NONE">
        <description>None (not triggered)</description>
      </entry>
      <entry value="1" name="PARACHUTE_DEPLOYMENT_TRIGGER_DRONE">
        <description>Drone</description>
      </entry>
      <entry value="2" name="PARACHUTE_DEPLOYMENT_TRIGGER_ATS">
        <description>Automatic trigger system (ATS)</description>
      </entry>
      <entry value="3" name="PARACHUTE_DEPLOYMENT_TRIGGER_MANUAL">
        <description>Manual</description>
      </entry>
      <entry value="4" name="PARACHUTE_DEPLOYMENT_TRIGGER_MAVLINK">
        <description>MAVLink (offboard computer)</description>
      </entry>
      <entry value="5" name="PARACHUTE_DEPLOYMENT_TRIGGER_GEOFENCE">
        <description>Geofence</description>
      </entry>
    </enum>
```

message_definitions/v1.0/common.xml
hamishwillee: So is this what you mean by the manual trigger mode?

jenke-skydio: Yes.  A manual trigger could come from the optional transmitter described above or could also come from the vehicle ground station via UI button and passed along from ground station to vehicle to parachute system via MAVLink

message_definitions/v1.0/common.xml
jenke-skydio: This is intended to be a MAVLink command originated from the flight controller.   It is meant to be differentiated from the "offboard computer".  Seems OK to allow other types of signals (PWM, etc) but would be good to additionally call out MAVLink

message_definitions/v1.0/common.xml
jenke-skydio: this seems reasonable (FYI @amy-chen-skydio ).  I would add an additional note to to include `MAVLink` in addition to `PWM, DroneCan, RC Control` in the description of `PARACHUTE_TRIGGER_FLAGS_FC`

Also see discussion above about `PARACHUTE_ERROR_FLAGS_RADIO_ERROR` where we talked about the manual trigger

message_definitions/v1.0/common.xml
jenke-skydio: i believe that is correct (CC @amy-chen-skydio )

message_definitions/v1.0/common.xml
jenke-skydio: this seems reasonable (CC @amy-chen-skydio )

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
jenke-skydio: The aim here is to allow the parachute system to automatically detect the altitude threshold and arm rather than the flight controller doing it.  I generally agree that in most cases you may want to have the flight controller deal with this logic.

message_definitions/v1.0/common.xml
jenke-skydio: I would advocate for `BATTERY_STATUS` here since we only battery percentage which can be covered by `battery_remaining` and a Boolean indicating charging (1) or not (0) which can be covered by `charge_state`

message_definitions/v1.0/common.xml
hamishwillee: Hmmm. So the drone looks at the MAV_TYPEs of incoming triggers and uses them to determine if the command is from the FC or the Companion computer, or a GCS (say)? Sounds complicated. If you mean this though, we should say it, and replace my name for onboard/MAVLink.

hamishwillee: So how does it know it's dealing with a geofence trigger? Is this a MAVLink geofence, or an internal one, or something else?

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: So is the disarm case because you might want to turn off ATS on the takeoff or landing, but not the converse?

message_definitions/v1.0/common.xml
jenke-skydio: > Hmmm. So the drone looks at the MAV_TYPEs of incoming triggers and uses them to determine if the command 

is from the FC or the Companion computer, or a GCS (say)?
Yes.  I can get more detail on this but i believe the idea is around both control of configuration such that various triggers could be enabled & disabled in different scenarios as well as being able to log & audit the source of a trigger event


> So how does it know it's dealing with a geofence trigger? Is this a MAVLink geofence, or an internal one, or something else?

I need to check on this -> @amy-chen-skydio do you know?

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
jenke-skydio: yes

message_definitions/v1.0/common.xml
jenke-skydio: geofence is intended to leverage MAVLink mission protocol to transmit geofence data to the parachute.  The parachute would then automatically trigger based on its internal GPS system.  Again, the goal here would be for a parachute system to be able to operate autonomously even if a vehicle were not operating correctly

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
        <description>There is an error with the parachute RF (used for manual control)</description>
```

message_definitions/v1.0/common.xml
hamishwillee: FWIW that makes sense, since it is understood.

BATTERY_STATUS_V2 provides the current field (positive or negative) to indicate charging. It's a much better message. FYI we are also proposing FUEL_STATUS which can be uses as a synonym which uses the same approach https://github.com/mavlink/mavlink/pull/2107

message_definitions/v1.0/common.xml
hamishwillee: I've updated https://github.com/mavlink/mavlink/pull/2063/files#r1577246968 to say this.

But probably need to say more - in particular in accompanying docs. In particular, I assume:
1. Parachute gets geofence from vehicle, not GCS.
2. Parachute [monitors for changes in the plan](https://mavlink.io/en/services/mission.html#detecting-missionplan-changes) and minimally gets the latest on boot (if opaque id not supported).

jenke-skydio: yes, both are correct: geofence comes from vehicle and should be updated on boot and both to be included on microservices documentation (fyi @amy-chen-skydio )

hamishwillee: @amy-chen-skydio FYI, I'm eyes off this until there are updates.

message_definitions/v1.0/common.xml
amy-chen-skydio: `MAV_CMD_SET_PARACHUTE_ARM_ALTI` has been removed in the latest commit

message_definitions/v1.0/common.xml
amy-chen-skydio: Removed `charge_status` in the latest commit and will use `BATTERY_STATUS` instead

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
amy-chen-skydio: Renamed and updated the description per [suggestion](https://github.com/mavlink/mavlink/pull/2063#discussion_r1577246968)

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
amy-chen-skydio: Renamed and updated the description per [suggestion](https://github.com/mavlink/mavlink/pull/2063#discussion_r1577246968)

message_definitions/v1.0/common.xml
amy-chen-skydio: Updated the description per suggestion above

message_definitions/v1.0/common.xml
amy-chen-skydio: Updated to be used as enum instead of flags

message_definitions/v1.0/common.xml
amy-chen-skydio: Updated per above suggestion

message_definitions/v1.0/common.xml
amy-chen-skydio: @hamishwillee updated per [suggestion](https://github.com/mavlink/mavlink/pull/2063#discussion_r1577246968)

message_definitions/v1.0/common.xml
amy-chen-skydio: Updated per above suggestion in latest commit

message_definitions/v1.0/common.xml
```diff
@@ -1873,6 +1873,11 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="225" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
+        <description>Command to arm/disarm parachute module</description>
```
hamishwillee, line 1877:
```suggestion
        <description>Command to arm/disarm parachute module trigger sources.</description>
```

hamishwillee:
@julianoes @potaito I believe that @amy-chen-skydio has addressed all your concerns. To my mind the descriptions for each trigger source are now clear, and this makes sense.

Are you OK with it?

hamishwillee:
@amy-chen-skydio Assuming that everyone is happy with this, we'll need similar tests/validations as for the illuminators to get this in.

julianoes (approved):

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+        <description>There is an error with the parachute RF (used for manual control)</description>
+      </entry>
+      <entry value="8" name="PARACHUTE_ERROR_FLAGS_LOW_POWER">
+        <description>Parachute module has low power</description>
+      </entry>
+      <entry value="16" name="PARACHUTE_ERROR_FLAGS_FC_CONNECTION_ERROR">
+        <description>There is an error with the parachute and FC (flight controller) connection</description>
+      </entry>
+      <entry value="32" name="PARACHUTE_ERROR_FLAGS_EFTS_CONNECTION_ERROR">
+        <description>There is an error with the parachute electrical/external FTS (flight termination system)</description>
```
potaito, line 5191: Why "electrical/external" FTS? It's always electrical, isn't it? And can this refer to internal and external Flight Termination Systems at the same time?

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="64" name="PARACHUTE_ERROR_FLAGS_POD_CONNECTION_ERROR">
+        <description>There is an error with the parachute pod</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_ERROR_FLAGS_EFTS_DIAGNOSE">
+        <description>Parachute electrical/external FTS (flight termination system) diagnosis failed</description>
```
potaito, line 5197:
```suggestion
        <description>Parachute electrical/external Flight Termination System (FTS) diagnosis failed</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="256" name="PARACHUTE_ERROR_FLAGS_CHARGING_FAILED">
+        <description>Parachute module charging failed</description>
+      </entry>
+      <entry value="512" name="PARACHUTE_ERROR_FLAGS_EXTERNAL_POWER_ERROR">
+        <description>There is an error with the parachute external power source</description>
+      </entry>
+      <entry value="1024" name="PARACHUTE_ERROR_FLAGS_GS_CONNECTION_ERROR">
+        <description>There is an error with the parachute and GS (ground station) connection</description>
```
potaito, line 5206:
```suggestion
        <description>There is an error with the connection between parachute and Ground Station</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="2048" name="PARACHUTE_ERROR_FLAGS_GPS_ERROR">
+        <description>There is an error with the parachute GPS</description>
```
potaito, line 5209:
```suggestion
        <description>There is an error with the parachute's GPS</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="4096" name="PARACHUTE_ERROR_FLAGS_SUBSYSTEM_CONNECTION_ERROR">
+        <description>There is an error with the parachute and subsystem connection</description>
```
potaito, line 5212:
```suggestion
        <description>There is an error with the connection between parachute its internal subsystem</description>
```

What kind of subsystem actually? Internal to the parachute? Isn't this then just an internal error in general?

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="8192" name="PARACHUTE_ERROR_FLAGS_SUBSYSTEM_FW_ERROR">
+        <description>There is an error with the parachute subsystem firmware</description>
```
potaito, line 5215: What kinds of error is this referring to? Example? Wrong firmware version?

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="16384" name="PARACHUTE_ERROR_FLAGS_RESERVED_1">
+        <description>Reserved for future use</description>
+      </entry>
+      <entry value="32768" name="PARACHUTE_ERROR_FLAGS_RESERVED_2">
+        <description>Reserved for future use</description>
+      </entry>
+      <entry value="65536" name="PARACHUTE_ERROR_FLAGS_LOGGING_ERROR">
+        <description>There is an error with the parachute module's logging</description>
```
potaito, line 5224:
```suggestion
        <description>There is an error with the parachute's internal logging system</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="131072" name="PARACHUTE_ERROR_FLAGS_MODULE_RETIRED">
+        <description>This parachute module is retired (too many deployments)</description>
```
potaito, line 5227:
```suggestion
        <description>This parachute module is retired. Too many deployments</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="262144" name="PARACHUTE_ERROR_FLAGS_GLOW_WIRE_ERROR">
+        <description>There is an error with the parachute glow wire</description>
+      </entry>
+      <entry value="524288" name="PARACHUTE_ERROR_FLAGS_OFFBOARD_CONNECTION_ERROR">
+        <description>There is an error with the parachute and offboard computer MAVLink connection</description>
```
potaito, line 5233:
```suggestion
        <description>There is an error with the MAVLink connection between parachute and offboard computer</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="1048576" name="PARACHUTE_ERROR_FLAGS_IMU_CALIBRATION_ERROR">
+        <description>Parachute module IMU calibration failed</description>
```
potaito, line 5236:
```suggestion
        <description>Parachute's internal IMU calibration failed</description>
```

No need to introduce "Parachute module" as another term in my opinion

message_definitions/v1.0/common.xml
```diff
@@ -7671,6 +7805,16 @@
...
     </message>
+    <message id="430" name="PARACHUTE_STATUS">
+      <description>Parachute system status</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Uptime in ms</field>
+      <field type="uint32_t" name="error_status" enum="PARACHUTE_ERROR_FLAGS" display="bitmask">Errors</field>
+      <field type="uint8_t" name="arm_status" enum="PARACHUTE_TRIGGER_FLAGS" display="bitmask">Parachute arming status</field>
+      <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_TRIGGER">Parachute deployment status</field>
+      <field type="uint8_t" name="safety_status" enum="PARACHUTE_SAFETY_FLAGS" display="bitmask">Parachute safety status</field>
+      <field type="uint8_t" name="ats_arm_altitude" units="m">Parachute Automatic Trigger System (ATS) auto-arming/disarming altitude in meters</field>
+      <field type="char[11]" name="parachute_packed_date">Parachute packed date (DD/MM/YYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
```
potaito, line 7816: Didn't we want to change this date to uint64 for less data?

If you really want to keep the string, I'd change it to YYYY-MM-DD according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

> The [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format YYYY-MM-DD (2024-05-22) is intended to harmonize these formats and ensure accuracy in all situations. Many countries have adopted it as their sole official date format, though even in these areas writers may adopt abbreviated formats that are no longer recommended.

potaito (commented):
Structure looks good to me. Only one comment about the date as a string still.

The rest are suggestions for cleaner phrasing.

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+        <description>Auto-arming of parachute automatic trigger system (ATS). This allows a parachute to enable ATS after detecting that it has reached a desired altitude.</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_TRIGGER_FLAGS_ATS_AUTO_DISARM">
+        <description>Auto-disarming of parachute automatic trigger system (ATS). This allows a parachute to disable ATS after detecting that it is below a desired altitude.</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_DEPLOYMENT_TRIGGER">
+      <description>Parachute deployment trigger source</description>
+      <entry value="0" name="PARACHUTE_DEPLOYMENT_TRIGGER_NONE">
+        <description>None (not triggered)</description>
```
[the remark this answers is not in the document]
amy-chen-skydio: Not all the deployments are triggered automatically, such as manual trigger.

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+        <description>Manual</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_DEPLOYMENT_TRIGGER_ATS">
+        <description>Automatic trigger system (ATS)</description>
+      </entry>
+      <entry value="3" name="PARACHUTE_DEPLOYMENT_TRIGGER_DRONE">
+        <description>Drone</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_DEPLOYMENT_TRIGGER_MAVLINK">
+        <description>MAVLink (offboard computer)</description>
```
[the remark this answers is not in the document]
hamishwillee: @amy-chen-skydio Are you able to directly merge these suggestions? I.e. have you got a button to commit the changes above? If so, that's the best way to merge changes, because we can see what you accepted and merged and what you did not - it makes tracking changes way easier.

If you do not have that option then there is something wrong with your github repo setup. There actually is anyway, since normally I can directly edit PRs, and I can't do so for yours.

message_definitions/v1.0/common.xml
hamishwillee: @potaito I'm OK with or without brackets, but if we're not using brackets we should have a proper sentence.

```suggestion
        <description>This parachute is retired (it has has been deployed too many times).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+    <enum name="PARACHUTE_ERROR_FLAGS" bitmask="true">
+      <description>Parachute module error flags (bitmap, 0 means no error)</description>
+      <entry value="1" name="PARACHUTE_ERROR_FLAGS_BAROMETER_ERROR">
+        <description>There is an error with the parachute barometer</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_ERROR_FLAGS_IMU_ERROR">
+        <description>There is an error with the parachute IMU</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_ERROR_FLAGS_RF_CONNECTION_ERROR">
+        <description>There is an error with the parachute RF (used for manual control)</description>
```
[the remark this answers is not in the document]
amy-chen-skydio: Updated the wording to `There is an error with the parachute's RF that is used for manual control`

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
amy-chen-skydio: @hamishwillee Sorry I completely missed this comment before I committed the changes. I will do this in the future.

message_definitions/v1.0/common.xml
amy-chen-skydio: I've updated the description. This should refer to Electrical Flight Termination System

message_definitions/v1.0/common.xml
amy-chen-skydio: This error is designed for different subsystems of Parachute Recovery System (PRS), it doesn't have to be internal to the parachute. For example, the subsystem could be remote controller or expansion board.

I've updated the description to `There is an error with the connection between parachute and subsystem (e.g. remote controller, expansion board, etc.)`

message_definitions/v1.0/common.xml
amy-chen-skydio: Yes, wrong firmware version is an example of this. I've update the description to `There is an error with the parachute subsystem firmware (e.g. wrong firmware version)`

message_definitions/v1.0/common.xml
amy-chen-skydio: I've updated the description to `This parachute module is retired (i.e. too many deployments)`

message_definitions/v1.0/common.xml
amy-chen-skydio: I've updated to use ISO 8601 format YYYY-MM-DD

message_definitions/v1.0/common.xml
hamishwillee: I agree!

hamishwillee:
@potaito All your comments addressed?

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: Next time :-)

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
potaito: So when set to "None" there is no way for the parachute to pop? It's effectively disabled then, isn't it?

message_definitions/v1.0/common.xml
potaito: Got it, thank you. To me a subsystem was something that's on the same device

potaito:
@hamishwillee 
Except for this one: https://github.com/mavlink/mavlink/pull/2063#discussion_r1609573608

But it's not a deal breaker, so LGTM. Thanks for the patience.

potaito (approved):
LGTM

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
amy-chen-skydio: This enum is used to indicate what the trigger source was after a parachute has deployed. So "None" means that no parachute has deployed.

amy-chen-skydio:
> @amy-chen-skydio Assuming that everyone is happy with this, we'll need similar tests/validations as for the illuminators to get this in.

@hamishwillee should I upload the tests and microservice here or should we create a PR for [mavlink-devguide](https://github.com/Skydio/mavlink-devguide) directly?

message_definitions/v1.0/common.xml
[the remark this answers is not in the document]
hamishwillee: @amy-chen-skydio I didn't find this confusing, but perhaps we could improve like this?

```suggestion
        <description>None (parachute has not deployed)</description>
```

hamishwillee:
@amy-chen-skydio
1. Let's address @potaito final comment like this: https://github.com/mavlink/mavlink/pull/2063/files#r1626779291
2. W.r.t. the tests, could you please create the mavlink devguide entry and cross link - that way we can merge them together and you don't have to upload twice.

hamishwillee:
Excellent. So from my perspective this is just waiting on docs  - for the tests.

amy-chen-skydio:
@hamishwillee here's the [PR](https://github.com/mavlink/mavlink-devguide/pull/546) for the docs and tests :pray:

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
         <description>Illuminator thermistor failure.</description>
       </entry>
     </enum>
+    <enum name="PARACHUTE_TRIGGER_FLAGS" bitmask="true">
+      <description>Parachute trigger sources.</description>
+      <entry value="1" name="PARACHUTE_TRIGGER_FLAGS_MANUAL">
+        <description>Manual trigger (ground based control via parachute-specific RF channel)</description>
+      </entry>
```
hamishwillee, line 5114: I think we need a more specific name for this because there are multiple manual triggers.
Note that this is intended for your remote control unit, but as described it could also be (in theory) a button on any old RC controller.

```suggestion
      <entry value="1" name="PARACHUTE_TRIGGER_FLAGS_MANUAL_RF_CHANNEL">
        <description>Manual trigger from parachute-specific RF channel</description>
      </entry>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      <entry value="2" name="PARACHUTE_TRIGGER_FLAGS_ATS">
+        <description>Automatic trigger system (ATS)</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_TRIGGER_FLAGS_FC">
+        <description>Flight controller trigger (e.g. MAVLink from FC, PWM, DroneCan, RC Control)</description>
```
hamishwillee, line 5119: While looking at the [docs PR](https://github.com/mavlink/mavlink-devguide/pull/546) it occurred to me that this has to play nice with the existing behaviour of [MAV_CMD_DO_PARACHUTE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PARACHUTE).

That command enables/disables automatic parachute triggering, or commands immediate triggering. The idea is that you perhaps want to turn off automatic triggering in some cases. While your new command allows configuration of various sources we need to consider that this would be a "master" command - i.e. if this is set, I would expect it to be the same as if `PARACHUTE_TRIGGER_FLAGS` was 0 except for "manual options", including `PARACHUTE_TRIGGER_FLAGS_MANUAL_RF_CHANNEL` and `PARACHUTE_TRIGGER_FLAGS_OFFBOARD`

There is overlap here with these flags. 

For this case

PARACHUTE_TRIGGER_FLAGS_MANUAL

What should happen if MAV_CMD_DO_PARACHUTE enables auto-triggering, but you have set PARACHUTE_TRIGGER_FLAGS_FC disabled?

What should happen if PARACHUTE_TRIGGER_FLAGS_MANUAL is

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      <entry value="1" name="PARACHUTE_DEPLOYMENT_TRIGGER_MANUAL">
+        <description>Manual</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_DEPLOYMENT_TRIGGER_ATS">
+        <description>Automatic trigger system (ATS)</description>
+      </entry>
+      <entry value="3" name="PARACHUTE_DEPLOYMENT_TRIGGER_DRONE">
+        <description>Drone</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_DEPLOYMENT_TRIGGER_MAVLINK">
```
hamishwillee, line 5151: Name should be consistent with the trigger option right?

```suggestion
      <entry value="4" name="PARACHUTE_DEPLOYMENT_TRIGGER_OFFBOARD">
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
         <description>Illuminator thermistor failure.</description>
       </entry>
     </enum>
+    <enum name="PARACHUTE_TRIGGER_FLAGS" bitmask="true">
```
hamishwillee, line 5110: While looking at the [docs PR](https://github.com/mavlink/mavlink-devguide/pull/546) it occurred to me that this has to play nice with the existing behaviour of [MAV_CMD_DO_PARACHUTE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PARACHUTE).

That command enables/disables automatic parachute triggering, or commands immediate triggering. The idea is that you perhaps want to turn off automatic triggering in some cases. While your new command allows configuration of various sources we need to consider that this would be a "master" command - i.e. if this is set, I would expect it to be the same as if `PARACHUTE_TRIGGER_FLAGS` was 0 except for "manual options", including `PARACHUTE_TRIGGER_FLAGS_MANUAL_RF_CHANNEL` and `PARACHUTE_TRIGGER_FLAGS_OFFBOARD`

We need to be more precise here about what should happen if there is overlap. Thoughts?

hamishwillee: More important, I think the manual trigger from [MAV_CMD_DO_PARACHUTE](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_PARACHUTE) should always be obeyed.

message_definitions/v1.0/common.xml
```diff
@@ -7671,6 +7805,16 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
+    <message id="430" name="PARACHUTE_STATUS">
+      <description>Parachute system status</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Uptime in ms</field>
+      <field type="uint32_t" name="error_status" enum="PARACHUTE_ERROR_FLAGS" display="bitmask">Errors</field>
+      <field type="uint8_t" name="arm_status" enum="PARACHUTE_TRIGGER_FLAGS" display="bitmask">Parachute arming status</field>
+      <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_TRIGGER">Parachute deployment status</field>
```
hamishwillee, line 7813:
```suggestion
      <field type="uint8_t" name="deployment_status" enum="PARACHUTE_DEPLOYMENT_TRIGGER">Parachute deployment reason/status.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="64" name="PARACHUTE_TRIGGER_FLAGS_ATS_AUTO_ARM">
+        <description>Auto-arming of parachute automatic trigger system (ATS). This allows a parachute to enable ATS after detecting that it has reached a desired altitude.</description>
+      </entry>
+      <entry value="128" name="PARACHUTE_TRIGGER_FLAGS_ATS_AUTO_DISARM">
+        <description>Auto-disarming of parachute automatic trigger system (ATS). This allows a parachute to disable ATS after detecting that it is below a desired altitude.</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_DEPLOYMENT_TRIGGER">
+      <description>Parachute deployment trigger source</description>
```
hamishwillee, line 5138: This is named as "trigger", but what we're reporting here is a _reason_ for deployment. So can we name it to clearly differentiate it from the trigger flags - which are "setters"? Here I have suggested "PARACHUTE_DEPLOY_REASON", because that is what it is. Need to to change everywhere.

```suggestion
    <enum name="PARACHUTE_DEPLOY_REASON">
      <description>Reason for parachute deployment (reported in PARACHUTE_STATUS)</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      <entry value="0" name="PARACHUTE_DEPLOYMENT_TRIGGER_NONE">
+        <description>None (parachute has not deployed)</description>
+      </entry>
+      <entry value="1" name="PARACHUTE_DEPLOYMENT_TRIGGER_MANUAL">
+        <description>Manual</description>
+      </entry>
+      <entry value="2" name="PARACHUTE_DEPLOYMENT_TRIGGER_ATS">
+        <description>Automatic trigger system (ATS)</description>
+      </entry>
+      <entry value="3" name="PARACHUTE_DEPLOYMENT_TRIGGER_DRONE">
```
hamishwillee, line 5148: See comment above. I have concerns about this.

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
         <description>Illuminator thermistor failure.</description>
       </entry>
     </enum>
+    <enum name="PARACHUTE_TRIGGER_FLAGS" bitmask="true">
+      <description>Parachute trigger sources.</description>
```
hamishwillee, line 5111: Mentioning the thing that uses it makes cross linking in docs a bit better.

```suggestion
      <description>Parachute trigger sources (used to enable specific trigger sources in MAV_CMD_SET_PARACHUTE_ARM).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="5" name="PARACHUTE_DEPLOYMENT_TRIGGER_GEOFENCE">
+        <description>Geofence</description>
+      </entry>
+      <entry value="6" name="PARACHUTE_DEPLOYMENT_TRIGGER_FTS_PRECHECKING">
+        <description>Flight Termination System (FTS) pre-checking protocol</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_SAFETY_FLAGS" bitmask="true">
+      <description>Parachute module safety-related flags.</description>
```
hamishwillee, line 5162:
```suggestion
      <description>Parachute module safety-related flags (reported in PARACHUTE_STATUS).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+      </entry>
+      <entry value="2" name="PARACHUTE_SAFETY_FLAGS_ON_GROUND">
+        <description>This is used to indicate that the parachute's own sensor has confirmed it is stably on the ground.</description>
+      </entry>
+      <entry value="4" name="PARACHUTE_SAFETY_FLAGS_GEOFENCE_MISSION_SET">
+        <description>This is used to indicate that the parachute module has downloaded geofence mission successfully and can be triggered by geofence source.</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_ERROR_FLAGS" bitmask="true">
+      <description>Parachute module error flags (bitmap, 0 means no error)</description>
```
hamishwillee, line 5174:
```suggestion
      <description>Parachute module error flags (bitmap, 0 means no error).  Reported in PARACHUTE_STATUS.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -5102,6 +5107,135 @@
...
+        <description>MAVLink, e.g. from an offboard computer</description>
+      </entry>
+      <entry value="5" name="PARACHUTE_DEPLOYMENT_TRIGGER_GEOFENCE">
+        <description>Geofence</description>
+      </entry>
+      <entry value="6" name="PARACHUTE_DEPLOYMENT_TRIGGER_FTS_PRECHECKING">
+        <description>Flight Termination System (FTS) pre-checking protocol</description>
+      </entry>
+    </enum>
+    <enum name="PARACHUTE_SAFETY_FLAGS" bitmask="true">
```
hamishwillee, line 5161: I'm OK with calling this safety flags, but should we perhaps consider `PARACHUTE_STATUS_FLAGS`?

message_definitions/v1.0/common.xml
```diff
@@ -1873,6 +1873,11 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="225" name="MAV_CMD_SET_PARACHUTE_ARM" hasLocation="false" isDestination="false">
```
hamishwillee, line 1876: I wonder if this is more clear? Thoughts

```suggestion
      <entry value="225" name="MAV_CMD_SET_PARACHUTE_TRIGGERS" hasLocation="false" isDestination="false">
```

hamishwillee:
This PR has conflicts and lots of unaddressed discussion from more than a year ago. Closing as stale. Would happilly take another PR that provides a response to the suggestions.

@bkueng @TomasTwardzik I see these are in the Auterion dialect. Unless there is flight-stack code supporting these, probably doesn't need to be moved to common.

PR #2069, opened 2024-01-08
common: battery_status gets state of health field

rmackay9:
This adds a new state_of_health percentage field to the BATTERY_STATUS message.

DroneCAN batteries at least report and the format of this field was derived somewhat from [this DroneCAN message](https://github.com/dronecan/DSDL/blob/master/uavcan/equipment/power/1092.BatteryInfo.uavcan).  The largest difference between this field and the corresponding DSDL message is for MAVLink I think we should use zero for unknown instead of 127 (or is it perhaps -1) because the field is an extension.

The corresponding ArduPilot mavlink repo PR is here https://github.com/ArduPilot/mavlink/pull/348
The ArduPilot flight code PR that makes use of this change is here: https://github.com/ArduPilot/ardupilot/pull/25926

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6395,6 +6395,7 @@
       <field type="uint16_t[4]" name="voltages_ext" units="mV" invalid="[0]">Battery voltages for cells 11 to 14. Cells above the valid cell count for this battery should have a value of 0, where zero indicates not supported (note, this is different than for the voltages field and allows empty byte truncation). If the measured value is 0 then 1 should be sent instead.</field>
       <field type="uint8_t" name="mode" enum="MAV_BATTERY_MODE">Battery mode. Default (0) is that battery mode reporting is not supported or battery is in normal-use mode.</field>
       <field type="uint32_t" name="fault_bitmask" display="bitmask" enum="MAV_BATTERY_FAULT">Fault/health indications. These should be set when charge_state is MAV_BATTERY_CHARGE_STATE_FAILED or MAV_BATTERY_CHARGE_STATE_UNHEALTHY (if not, fault reporting is not supported).</field>
+      <field type="uint8_t" name="state_of_health" units="%">State of Health (SOH) estimate.  Typically 100 at the time of manufacture and will decrease over time and use.  0 if unknown.</field>
     </message>
     <message id="148" name="AUTOPILOT_VERSION">
       <description>Version and capability of autopilot software. This should be emitted in response to a request with MAV_CMD_REQUEST_MESSAGE.</description>


message_definitions/v1.0/common.xml
```diff
@@ -6395,6 +6395,7 @@
       <field type="uint16_t[4]" name="voltages_ext" units="mV" invalid="[0]">Battery voltages for cells 11 to 14. Cells above the valid cell count for this battery should have a value of 0, where zero indicates not supported (note, this is different than for the voltages field and allows empty byte truncation). If the measured value is 0 then 1 should be sent instead.</field>
       <field type="uint8_t" name="mode" enum="MAV_BATTERY_MODE">Battery mode. Default (0) is that battery mode reporting is not supported or battery is in normal-use mode.</field>
       <field type="uint32_t" name="fault_bitmask" display="bitmask" enum="MAV_BATTERY_FAULT">Fault/health indications. These should be set when charge_state is MAV_BATTERY_CHARGE_STATE_FAILED or MAV_BATTERY_CHARGE_STATE_UNHEALTHY (if not, fault reporting is not supported).</field>
+      <field type="uint8_t" name="state_of_health" units="%">State of Health (SOH) estimate.  Typically 100 at the time of manufacture and will decrease over time and use.  0 if unknown.</field>
```
hendjoshsr71, line 6398: Should be 255 for unknown since that is not possible

```suggestion
      <field type="uint8_t" name="state_of_health" units="%">State of Health (SOH) estimate.  Typically 100 at the time of manufacture and will decrease over time and use.  255 if unknown.</field>
```

hendjoshsr71:
This RFC is likely relevant : https://github.com/mavlink/rfcs/pull/19

Generally, adding fields to Battery_Status has been a nonstarter since it is already too large of a message.... (Nothing has been added in 4 years to it.)

Here is the newer Battery_Status_V2 message : https://mavlink.io/en/messages/development.html#BATTERY_STATUS_V2
It doesn't include state of health since that changes quite slowly.

I did check that the SMART_BATTERY_INFO message Doesn't Include state of health : https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO

I would either add State of Health to that message. Or create a new improved Smart_Battery_Info that mirrors dronecan. This SMART_BATTERY_INFO message has not been merged in either PX4 or AP yet.

@hamishwillee probably more work needs to be done on figuring out to match the newer CAN messages which are saner IMO.

message_definitions/v1.0/common.xml
rmackay9: Thanks for the feedback.  I agree that 0 isn't ideal but because it's an extension, if we make it 255 then it will always consume the maximum bandwidth on the telemetry link.  Also for those publishers that don't implement the field (or even know anything about it) the value will appear as 0 to the consumer so the consumer really needs to assume that 0 is unknown I think.

rmackay9:
@hendjoshsr71,

OK, thanks for that.

BATTERY_STATUS_V2 looks OK although it's massive scope creep because AP doesn't implement this message and it means we would need to organise a migration to the new message.

I'm much less keen on the SMART_BATTERY_INFO because the field is relevant to non-smart batteries.

hamishwillee:
Thanks @hendjoshsr71 . @rmackay9 It's worth having a peak at the RFC for an understanding of the history and design. It was developed in discussion with @dakejahl who was doing the CAN battery messages at the time. It was widely discussed within both communities and agreed _ready for prototyping_ by all parties - signed off by @tridge . 

The driver was those continuously growing arrays of cells in BATTERY_STATUS that were anecdotally using a significant portion of links for vehicles with more than a few batteries. Adding a field like this temperature field would immediately mean that all of the preceding fields are not zero-byte truncated. 

Nothing has happened AFAIK in terms of development since, because it looks like the CAN side has evolved a bit since. Some of that looks inferrable though - e.g. temperature in Kelvin vs cDegC and so forth.

With respect to health, this would be inferred from https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO using `capacity_full_specification` and `capacity_full` - this second accounts for degradation, so you can estimate how far down the path your battery is. Note though that there are faults that warn of specific battery issues.

The RFC proposes three messages.
-  The BATTERY_STATUS_V2 is the high rate message. It gets rid of all the battery information that is static. It also gets rid of the cell information. Most consumers don't care, and growing that every few years is what screwed up the BATTERY_STATUS message and makes it unextendable.
-  [SMART_BATTERY_INFO](https://mavlink.io/en/messages/common.html#SMART_BATTERY_INFO) message is for all the static or near-static information. The name is legacy. If we can show there are no implementers (I have a feeling that was supposed to be ARK) then we can change. 
- A not yet fully defined message with cell information that was not included in the BATTERY_STATUS_V2. This is waiting on tridge to PR the change, but obviously not much pressure for that - @WickedShell was the only really interested party - this is mostly for debug and would not be enabled by default.

Anyway, I think we need to bite the bullet and change to new messages, and I'm not alone. But you're right it needs to be managed. Perhaps have a chat with @auturgy and @tridge . 

We can change BATTERY_STATUS_V2, and if no one has implemented SMART_BATTERY_INFO we should rename that too or create a new version and deprecate the first. 
Next step would be as Josh says to make sure we are still compatible with CAN.

I'm away for two weeks. Would someone like to propose the ideal message? Sticking with a high rate update/low rate update message makes sense for MAVLink. The messages don't have to be identical, but it should be possible to capture the information in CAN using MAVLink types, or we should explicitly state what types we want.

rmackay9:
I've created an initial proposal for the BATTERY_INFO message in this PR https://github.com/mavlink/mavlink/pull/2070

PR #2070, opened 2024-01-10
Common: rename SMART_BATTERY_INFO to BATTERY_INFO and add SOH

rmackay9:
As a follow up to https://github.com/mavlink/mavlink/pull/2069, here is a proposal for the BATTERY_INFO message:

1. renamed SMART_BATTERY_INFO to BATTERY_INFO based on the assumption that there were no users of the old message
2. removed the extension tag because if we're going to rename the message then old consumers can't use it anyway, so no need for the last few fields to be extensions
3. added a state_of_health field just below the capacity fields which I think are all related

I suppose that more analysis is required to confirm that this message is complete before we can merge.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6379,7 +6379,7 @@
       <field type="float" name="yaw_rate" units="rad/s">Angular rate in yaw axis</field>
     </message>
     <message id="147" name="BATTERY_STATUS">
-      <description>Battery information. Updates GCS with flight controller battery status. Smart batteries also use this message, but may additionally send SMART_BATTERY_INFO.</description>
+      <description>Battery information. Updates GCS with flight controller battery status. Smart batteries also use this message, but may additionally send BATTERY_INFO.</description>
       <field type="uint8_t" name="id" instance="true">Battery ID</field>
       <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
       <field type="uint8_t" name="type" enum="MAV_BATTERY_TYPE">Type (chemistry) of the battery</field>
@@ -7357,7 +7357,7 @@
       <field type="int32_t" name="y">Y coordinate of center point.  Coordinate system depends on frame field: local = x position in meters * 1e4, global = latitude in degrees * 1e7.</field>
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
-    <message id="370" name="SMART_BATTERY_INFO">
+    <message id="370" name="BATTERY_INFO">
       <description>Smart Battery information (static/infrequent update). Use for updates from: smart battery to flight stack, flight stack to GCS. Use BATTERY_STATUS for smart battery frequent updates.</description>
       <field type="uint8_t" name="id" instance="true">Battery ID</field>
       <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
@@ -7371,7 +7371,6 @@
       <field type="uint16_t" name="discharge_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when discharging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="charging_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when charging. If not supplied set to UINT16_MAX value.</field>
       <field type="uint16_t" name="resting_minimum_voltage" units="mV" invalid="UINT16_MAX">Minimum per-cell voltage when resting. If not supplied set to UINT16_MAX value.</field>
-      <extensions/>
       <field type="uint16_t" name="charging_maximum_voltage" units="mV" invalid="0">Maximum per-cell voltage when charged. 0: field not provided.</field>
       <field type="uint8_t" name="cells_in_series" invalid="0">Number of battery cells in series. 0: field not provided.</field>
       <field type="uint32_t" name="discharge_maximum_current" units="mA" invalid="0">Maximum pack discharge current. 0: field not provided.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7364,6 +7364,7 @@
       <field type="uint8_t" name="type" enum="MAV_BATTERY_TYPE">Type (chemistry) of the battery</field>
       <field type="int32_t" name="capacity_full_specification" units="mAh" invalid="-1">Capacity when full according to manufacturer, -1: field not provided.</field>
       <field type="int32_t" name="capacity_full" units="mAh" invalid="-1">Capacity when full (accounting for battery degradation), -1: field not provided.</field>
+      <field type="int8_t" name="state_of_health" units="%" invalid="-1">State of Health (SOH) estimate.  Typically 100 at the time of manufacture and will decrease over time and use. -1: field not provided.</field>
       <field type="uint16_t" name="cycle_count" invalid="UINT16_MAX">Charge/discharge cycle count. UINT16_MAX: field not provided.</field>
       <field type="char[16]" name="serial_number" invalid="[0]">Serial number in ASCII characters, 0 terminated. All 0: field not provided.</field>
       <field type="char[50]" name="device_name" invalid="[0]">Static device name in ASCII characters, 0 terminated. All 0: field not provided. Encode as manufacturer name then product name separated using an underscore.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7358,6 +7358,7 @@
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
     <message id="370" name="BATTERY_INFO">
+      <wip/>
       <description>
         Battery information that is static, or requires infrequent update.
         This message should requested using MAV_CMD_REQUEST_MESSAGE and/or streamed at very low rate.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7358,25 +7358,31 @@
       <field type="float" name="z" units="m">Altitude of center point. Coordinate system depends on frame field.</field>
     </message>
     <message id="370" name="BATTERY_INFO">
-      <description>Smart Battery information (static/infrequent update). Use for updates from: smart battery to flight stack, flight stack to GCS. Use BATTERY_STATUS for smart battery frequent updates.</description>
+      <description>
+        Battery information that is static, or requires infrequent update.
+        This message should requested using MAV_CMD_REQUEST_MESSAGE and/or streamed at very low rate.
+        BATTERY_STATUS_V2 is used for higher-rate battery status information.
+      </description>
       <field type="uint8_t" name="id" instance="true">Battery ID</field>
-      <field type="uint8_t" name="battery_function" enum="MAV_BATTERY_FUNCTION">Function of the battery</field>
-      <field type="uint8_t" name="type" enum="MAV_BATTERY_TYPE">Type (chemistry) of the battery</field>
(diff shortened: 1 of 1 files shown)


hamishwillee:
@rmackay9 Just back after holidays - sorry for delay. 

The big thing is to make sure this is not used by anyone. 
Then yes, we should do some due diligence to make sure that it is complete - that means capturing everything exported by DroneCAN.

I've posted in https://discord.com/channels/899843920987562014/902774723925147649/threads/1202135873668255764 - but if you can ask around that would be good.

dakejahl (commented):
LGTM

For the sake of comparing against the DroneCAN BatteryPeriodic message we added last year:
- **device_name**: this is named **name** in DC
- **serial_number**: this is 32 bytes in DC
- **manufacture_date**: this is 9 bytes in DC
- **capacity_full_specification**: this is named **design_capacity** in DC and is a float32 (Ah)
- **capacity_full**: named **full_charge_capacity** in DC and is a float32 (Ah)


**Included in DC, missing here**
- nominal_voltage
- charging_maximum_current

**Included here, missing in DC**
- weight
- id
- battery function
- type
- resting_minimum_voltage

While we're in here it might be good to at least align the names/types/units.

hamishwillee:
Thanks @dakejahl . I found [BatteryInfo](https://dronecan.github.io/Specification/7._List_of_standard_data_types/#batteryinfo), but where is the definition of DroneCAN `BatteryPeriodic`?

dakejahl:
@hamishwillee I only mention those DC messages since there was a lot of thought put into them and they're relatively new. We don't necessarily need to make mavlink/dronecan match since they're intended for different recipients and comm links (groundstand over radio vs robotic system over CAN). But if we're going to make a new message just to change the name and add one field we might want to take the opportunity to review the message as a whole and decided if anything else should change.

hamishwillee:
@dakejahl @rmackay9 I had a chat to @auturgy about this and we agreed to fully align the two messages (in terms of names, types, units, invalid values) for existing fields and then add the additional fields from the smart battery message. 
For a message that is only be sent on request (i.e. on seeing BATTERY_STATUS_V2 with a new id) or very low rate it makes sense to align and reduce the chance of conversion errors.

So that's what is done - the only differences are:
- Additional fields 
  - id - needed to map to specific batteries and in BATTERY_STATUS
  - battery function and type - from original BATTERY_STATUS
  - resting_minimum_voltage and weight - which were part of the smart battery requirements and I presume are desired?
- uint8 strintgs in CAN are char in MAVLink
- float used for all floats as there is no float16 in MAVLink
- ordered by size of type and with char last to potentially allow truncation. 

@rmackay9 Note that to match UAVCAN this `state_of_health` modified to `uint` (not int) and invalid value is now the max value of the type.

hamishwillee (approved):
Self merging as discussed last week.

auturgy:
Lgtm

rmackay9:
Thanks very much!

message_definitions/v1.0/common.xml
```diff
@@ -7357,26 +7357,33 @@
...
+      <field type="float" name="nominal_voltage" units="V" invalid="0">Battery nominal voltage. Used for conversion between Wh and Ah. 0: field not provided.</field>
+      <field type="float" name="discharge_maximum_current" units="A" invalid="0">Maximum pack discharge current. 0: field not provided.</field>
+      <field type="float" name="discharge_maximum_burst_current" units="A" invalid="0">Maximum pack discharge burst current. 0: field not provided.</field>
+      <field type="float" name="design_capacity" units="Ah" invalid="0">Fully charged design capacity. 0: field not provided.</field>
+      <field type="float" name="full_charge_capacity" units="Ah" invalid="NaN">Predicted battery capacity when fully charged (accounting for battery degradation). NAN: field not provided.</field>
+      <field type="char[9]" name="manufacture_date" invalid="[0]">Manufacture date (DDMMYYYY) in ASCII characters, 0 terminated. All 0: field not provided.</field>
+      <field type="char[32]" name="serial_number" invalid="[0]">Serial number in ASCII characters, 0 terminated. All 0: field not provided.</field>
```
MaEtUgR, line 7385: @rmackay9 @hamishwillee  May I politely ask why it was changed to `char[32]`? I'm not against that just wanted to clarify if you have a certain case of serial number definition or a list of possible serial numbers or any other specific protocol definition you wanted to cover with such a large string. Generally the binary 2^256 or even just the alphanumeric 62^32 number of serial number combinations seems pretty overkill to me. Of course we send the info low frequency so resources should not be the problem but it's still a bit a wildcard for a more specific definition.

I've seen:
- [SMBus](https://github.com/PX4/PX4-Autopilot/blob/cac01339018b9beddcf669f7f2f67461c37fefd0/src/lib/drivers/smbus_sbs/SBS.hpp#L227-L228) `uint16_t`
- [DroneCAN DSDL BatteryInfo](https://github.com/dronecan/DSDL/blob/0856687a5acb42b55d29bbad8f4fc9e19b4989fc/uavcan/equipment/power/1092.BatteryInfo.uavcan#L63) `uint32_t`
- [DroneCan DSDL CUAV CBAT](https://github.com/dronecan/DSDL/blob/0856687a5acb42b55d29bbad8f4fc9e19b4989fc/cuav/equipment/power/20300.CBAT.uavcan#L21) `uint16_t`
- [DroneCAN DSDL Ardupilot BatteryPeriodic](https://github.com/dronecan/DSDL/blob/0856687a5acb42b55d29bbad8f4fc9e19b4989fc/ardupilot/equipment/power/20011.BatteryPeriodic.uavcan#L6) `char[32]`
- Cyphal Battery (can't find atm) `char[16]`
- Proprietary battery manufacturer 1 `uint32_t`
- Proprietary battery manufacturer 2 `char[14]`

Is `char[32]` the common denominator we settle on 🤔 And do we agree on an `itoa(uintXX_t)` conversion for integers?
I'm asking because I intend to change the PX4 internal definition to be as compatible as possible while not wasting resources or providing an unclear definition.

hamishwillee: @MaEtUgR Not sure. I think:
- we triggered off [DroneCAN DSDL Ardupilot BatteryPeriodic](https://github.com/dronecan/DSDL/blob/0856687a5acb42b55d29bbad8f4fc9e19b4989fc/ardupilot/equipment/power/20011.BatteryPeriodic.uavcan#L6) 
- I suspect the main thinking was that this is a very low rate or on-request message so it didn't matter so much if we allowed any ascii char and some length. Given the low rate better not to constrain the possible serial number formats.
- _think_ the `itoa` conversions should be OK for integers as long as the string is 0 terminated. @peterbarker can you confirm that.

FYI If I was to look at this again I'd consider whether some of this information should be in https://mavlink.io/en/messages/common.html#COMPONENT_INFORMATION_BASIC instead.

peterbarker: Is there any particular reason to assume the entry in that message in numeric?  ie. `atoi` may make be making an invalid assumption.

I do realise it says "serial number" - but "serial numbers" often include leading or trailing letters etc

hamishwillee: Obviously MAVLink can't control whether batteries (or anything) has a **unique** serial number. IMO we shouldn't enforce that it is numeric either, though obviously Cyphal, SMBUS etc have taken a different view. I think this is too late to change, even if we can agree that we should.

However it is reasonable that if it is numeric _and we can assume that it will always be base 10 number_ it should be possible to clearly defined the conversion between the data types.
I think we can assume that - it would be easy but weird to do anything else.

Anyway, what I think we can do here is make the statement in docs for serial numbers like:

> Unique numeric serial numbers should be used if at all possible, as these can be converted to the integer fields used for transporting serial numbers over other internal busses, such as DroneCan, Cyphal, and SMBUS. The recommended conversions in this case can be done with the C `itoa` and `atoi` functions assuming the serial number is base 10. The bridging behaviour when using alphanumeric serial numbers is undefined.

Would that do what you need?

@hendjoshsr71 Does this make sense to you?

peterbarker:
Results from our DevCall:
 - integers should be turned into their ASCII representation for transport via mavlink
 - atoi on these char[32] fields is not a valid thing to do

We probably need to renumber this message now; 1 in 65535 chance of a GCS accepting the other form of the packet.

message_definitions/v1.0/common.xml
hamishwillee: @MaEtUgR et al, there _might_ be an opportunity to change this https://github.com/mavlink/mavlink/pull/2173

hendjoshsr71: Serial numbers should allow for chars as it is very common for devices. I believe the Tattu DroneCAN batteries used letters in their serial number initially. 

Here is my Tattu DroneCAN battery using the old DroneCAN message power.BatteryInfo
"TAA16KP12S15C_Z220506001400220"
First half is model number and the second is serial number.

hamishwillee:
Thanks @peterbarker @hendjoshsr71 

> atoi on these char[32] fields is not a valid thing to do

I guess most of the time a serial number is coming from uavcan and can translate to MAVLink. Not so often we tell the uavcan battery its serial number. So this makes sense.

> We probably need to renumber this message now; 1 in 65535 chance of a GCS accepting the other form of the packet.

I'm really not sure. 
- there is no doubt we _should_ have created a new message originally
- I doubt there would be any impact on leaving things as they are, but I know there is likely to be some from changing the ids
- But if there is a problem from this later, much harder to fix then than now.

Accepting comments in https://github.com/mavlink/mavlink/pull/2173

message_definitions/v1.0/common.xml
MaEtUgR: My previously highest requirement for a battery serial number was a `uint32_t`, but based on these definitions, longer alphanumeric serials must exist. I was asking for context on why `BATTERY_INFO` uses `char[32]`. It seems this was influenced by DroneCAN and, specifically, @hendjoshsr71’s example in https://github.com/dronecan/DSDL/pull/7#issuecomment-1149461594. Nothing against that definition.

@hamishwillee I was confused by https://github.com/mavlink/mavlink/pull/2173 a new message would’ve been cleaner, but reverting might’ve caused more churn. The discussion seems to have happened in a Discord I can’t access.

Looking ahead, I believe we should move away from `SMART_BATTERY_INFO` and adopt `BATTERY_INFO` with char[32], as I reflected in this PX4 change:
https://github.com/PX4/PX4-Autopilot/pull/24960

hamishwillee: @MaEtUgR Re ^^^

1. The question of serial numbers comes up again and again.
   - I personally prefer `uint32_t` and a predefined byte mapping to a semantic version because this is what most people use, and because it makes the field useful for more than just presentation in a GUI and direct comparison of a particular value.
   -  The choice to adopt `char[33]` followed the dronecan message (as you say) and also was agreed for https://mavlink.io/en/messages/common.html#COMPONENT_INFORMATION_BASIC. 
   - I can't tell you exactly "why" other than we decided to go for flexibility - too many "what ifs" - what if I want to use a git hash. What if I have letters in my serial number blah blah.

> @hamishwillee I was confused by https://github.com/mavlink/mavlink/pull/2173 a new message would’ve been cleaner, but reverting might’ve caused more churn. The discussion seems to have happened in a Discord I can’t access.

1. Initially we renamed the SMART_BATTTERY message to BATTERY_INFO rather than adding a BATTERY_INFO message because our initial investigation showed there were no consumers).
   That seemed reasonable at the time (and was agreed the stakeholders but it was dumb.
2. Some time later a user of SMART_BATTERY message was discovered. There was pressure to revert, so I did.
3. Then moved BATTERY_INFO to new id.

Upshot, I screwed this up in step 1. The following stuff occurred because we decided it was better to accept churn now to avoid being in an unfixable mess later.

> Looking ahead, I believe we should move away from SMART_BATTERY_INFO and adopt BATTERY_INFO with char[32], as I reflected in this PX4 change:

Yes. We should dump SMART_BATTERY_INFO as soon as possible. ... and adopt BATTERY_STATUS_V2. There is a third message proposed to include cell information dropped from BATTERY_STATUS. It turns out no one really cares much about that though.

MaEtUgR: Thanks for the clear answer! I’ll do my best to align PX4 with these interface improvements step by step. 🙏👍

PR #2074, opened 2024-01-18
development.xml: merge both MAV_CMD enums together

KonradRudin:
### Problem
the <enum name="MAV_CMD"> element with the same name existed twice in the parent enums element in the development.xml. Some xml parser seem to have a problem with that. Thus grouping them together.

Only cosmetic change, no wording message are changed.


hamishwillee (approved):
Thanks very much @KonradRudin - that was unintended. It would be hard to prevent with the current parser.

PR #2078, opened 2024-02-07
COMPONENT_INFORMATION_BASIC - add manufacturer date

hamishwillee:
Manufacturer date has been requested in https://github.com/mavlink/mavlink/pull/2063, and seems generally useful.

This adds it as a unix timestamp in seconds.

FYI @julianoes @potaito @auturgy

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -459,6 +459,7 @@
       <description>Basic component information data. Should be requested using MAV_CMD_REQUEST_MESSAGE on startup, or when required.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint64_t" name="capabilities" enum="MAV_PROTOCOL_CAPABILITY" display="bitmask">Component capability flags</field>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/development.xml
```diff
@@ -459,6 +459,7 @@
       <description>Basic component information data. Should be requested using MAV_CMD_REQUEST_MESSAGE on startup, or when required.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint64_t" name="capabilities" enum="MAV_PROTOCOL_CAPABILITY" display="bitmask">Component capability flags</field>
+      <field type="uint64_t" name="time_manufacture_us" units="us">Date of manufacture as a UNIX Epoch time (since 1.1.1970) in us.</field>
```
julianoes, line 462: We could do seconds since 1970 in an uint32 valid until 2106.

hamishwillee: Probably better. Fact is that it's a field that we only care about in terms of probably days or months
```suggestion
      <field type="uint32_t" name="time_manufacture_s" units="s" invalid="0">Date of manufacture as a UNIX Epoch time (since 1.1.1970) in seconds.</field>
```

hamishwillee:
This agreed in the dev call.

PR #2079, opened 2024-02-08
common: add set-camera-source command

rmackay9:
This adds a mavlink command to change the active camera lens (e.g. which camera lens is being streamed to the GCS).  This is useful for cameras with multiple lenses like the ViewPro and Xacti cameras.

ArduPilot already has support for doing this through an auxiliary switch and the corresponding ArduPilot PRs are:

- AP mavlink: https://github.com/ArduPilot/mavlink/pull/351
- AP flightcode: https://github.com/ArduPilot/ardupilot/pull/26172

I think the main question is whether we are happy with the "device id" field which I've modelled after the similar "Gimbal device id" field found in many gimbal related commands and messages including [MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/common.xml#L2109)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2093,7 +2093,7 @@
         <description>Set camera source. Changes the camera's active sources on cameras with multiple image sensors.</description>
         <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras.</param>
         <param index="2" label="primary source" enum="CAMERA_SOURCE">Primary Source</param>
-        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source. If non-zero the second source will be displayed as picture-in-picture</param>
+        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source. If non-zero the second source will be displayed as picture-in-picture.</param>
       </entry>
       <entry value="600" name="MAV_CMD_JUMP_TAG" hasLocation="false" isDestination="false">
         <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2091,9 +2091,9 @@
       </entry>
       <entry value="534" name="MAV_CMD_SET_CAMERA_SOURCE" hasLocation="false" isDestination="false">
         <description>Set camera source. Changes the camera's active sources on cameras with multiple image sensors.</description>
-        <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras</param>
+        <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras.</param>
         <param index="2" label="primary source" enum="CAMERA_SOURCE">Primary Source</param>
-        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source.  Non-zero for Picture-in-picture</param>
+        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source. If non-zero the second source will be displayed as picture-in-picture</param>
       </entry>
       <entry value="600" name="MAV_CMD_JUMP_TAG" hasLocation="false" isDestination="false">
         <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2089,6 +2089,11 @@
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
         <param index="2" label="Usage" enum="STORAGE_USAGE_FLAG">Usage flags</param>
       </entry>
+      <entry value="534" name="MAV_CMD_SET_CAMERA_LENS" hasLocation="false" isDestination="false">
+        <description>Set camera lens. Changes the active camera lens on cameras with multiple lens.</description>
+        <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras</param>
+        <param index="2" label="Lens">Lens</param>
+      </entry>
       <entry value="600" name="MAV_CMD_JUMP_TAG" hasLocation="false" isDestination="false">
         <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
         <param index="1" label="Tag" minValue="0" increment="1">Tag.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2089,10 +2089,11 @@
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
         <param index="2" label="Usage" enum="STORAGE_USAGE_FLAG">Usage flags</param>
       </entry>
-      <entry value="534" name="MAV_CMD_SET_CAMERA_LENS" hasLocation="false" isDestination="false">
-        <description>Set camera lens. Changes the active camera lens on cameras with multiple lens.</description>
+      <entry value="534" name="MAV_CMD_SET_CAMERA_SOURCE" hasLocation="false" isDestination="false">
+        <description>Set camera source. Changes the camera's active sources on cameras with multiple image sensors.</description>
         <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras</param>
-        <param index="2" label="Lens">Lens</param>
+        <param index="2" label="primary source" enum="CAMERA_SOURCE">Primary Source</param>
+        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source.  Non-zero for Picture-in-picture</param>
       </entry>
       <entry value="600" name="MAV_CMD_JUMP_TAG" hasLocation="false" isDestination="false">
         <description>Tagged jump target. Can be jumped to with MAV_CMD_DO_JUMP_TAG.</description>
@@ -3856,6 +3857,21 @@
       </entry>
       <entry value="6" name="FOCUS_TYPE_AUTO_CONTINUOUS">
         <description>Continuous auto focus. Mainly used for dynamic scenes. Abbreviated as AF-C.</description>
+      </entry>
+    </enum>
+    <enum name="CAMERA_SOURCE">
+      <description>Camera sources for MAV_CMD_SET_CAMERA_SOURCE</description>
+      <entry value="0" name="CAMERA_SOURCE_DEFAULT">
+        <description>Default camera source.</description>
+      </entry>
+      <entry value="1" name="CAMERA_SOURCE_RGB">
+        <description>RGB camera source.</description>
+      </entry>
+      <entry value="2" name="CAMERA_SOURCE_IR">
+        <description>IR camera source.</description>
+      </entry>
+      <entry value="3" name="CAMERA_SOURCE_NDVI">
+        <description>NDVI camera source.</description>
       </entry>
     </enum>
     <enum name="PARAM_ACK">


rmackay9:
Hi @hamishwillee,

Thanks for the review and feedback.

I suspect the other team bypasses the autopilot completely and directly communicates with the camera.  The most obvious advantage to my approach of adding a new command is that it means that autonomous missions can change the lens as well.

dogmaphobic:
> @dogmaphobic Can you confirm how sensors (say IR/Visible light) are selected on a single camera - is it a definition file thing?

The vehicle presents multiple cameras. It's very unlikely you have one camera that can switch lenses. Each "lens" is a camera with its own sensor. The ground station can select which one to use or, in the case of streaming video, it can stream any number of them simultaneously. A good example is where you have a camera with a thermal sensor and another with a regular sensor. You can can composite both streams and blend them providing both real color and thermal.

This is all explained in [MAVLink Camera Protocol documentation](https://mavlink.io/en/services/camera.html). Everything is done through MAVLink and there is no reason missions cannot handle it. On the other hand, this implies a vehicle with a companion computer responsible for the payload. If this message is supposed to bypass the payload manager, how will the flight controller know how to trigger a different camera (assuming in this case it's all a dumb/trigger by wire system)?

rmackay9:
Hi @hamishwillee, @dogmaphobic

Three different cameras that I've worked with ([Xacti](https://ardupilot.org/copter/docs/common-xacti-gimbal.html), [Siyi ZT30](https://shop.siyi.biz/products/siyi-zt30), [ViewPro](https://ardupilot.org/copter/docs/common-viewpro-gimbal.html)) support switching the active "lens" (or "image sensor" or whatever you want to call it).  Normally that selection is RGB or IR but sometimes it is a picture-in-picture combination of the two.  This selection affects what is "streamed" (or output on the camera's HDMI port in the case of the Xacti) but also and what pictures are taken.  At least two of the three do not support streaming from both lenses/sensors at the same time.

I considered using MAV_CMD_VIDEO_START_STREAMING but this doesn't seem correct because (as mentioned) it would also affect what pictures are taken.  Also changing the lens/sensor does not actually start/stop streaming.

MatejFranceskin:
Having a gimbal and a camera with multiple sensors is a very common setup. We expose a "Camera" parameter in camera definition file with multiple values, for example: "EO", "IR". There are specific parameters for each sensor that are active only when particular sensor is chosen and this can be also addressed in camera definition XML with conditional parameters.
@rmackay9 The question is what are you trying to achieve besides switching the active sensor that cannot be done with the above approach?

hamishwillee:
Thanks very much @dogmaphobic @MatejFranceskin @rmackay9 

@MatejFranceskin Good - that's what I thought. We need to document this use case (irrespective of what comes of this discussion). Would it be possible for you to send me a camera definition file (or attach) with this field, or update https://mavlink.io/en/services/camera_def.html#full_example to add this to the example?

Further, don't we need to freeze the ID in the common parameters (https://mavlink.io/en/services/camera_def.html#common-parameters) so that the UI can "know" what this is and move the accompanying widget around if needed?

@rmackay9 Just to add to what @MatejFranceskin said above: the case of Picture-in-Picture can also be done as an option in the definition file.

In summary as I understand it:

- This is the "right" way to do the use case of a multi-sensor/lens camera that can have a single sensor active at a time. That makes sense because it is still just one camera.
- This will work in missions provided you don't need to change the mode half way through in a mission item. I guess it is possible that you might for a very long mission. Am I missing something here?
- If multiple sensors can be used at the same time, then MAVLink would expect them to be treated as separate cameras.

You could also treat these as separate cameras, but your implementation would need to expose the sensors as separate cameras. The problem here is that there appears to be an assumption that all cameras are available all the time. If the sensors are mutually exclusive that is not the case. So you'd get a list of all sensors at startup and then some would refuse a command if the sensor is not available. 
The UI would be ugly, because you'd have no easy way to know which sensor is active.

FWIW The camera API has been designed to expose the simple subset you'd need in most cameras in messages, and do the many optional settings in the camera definition. This use case still feels like one that would be best dealt with by the camera definition file - in particular because we're already seeing many possible combinations you might want:
- IR, Visible Light, X-Ray :-)
- Picture in picture of any combination.
- Whatever else comes up.

rmackay9:
Hi, I must admit that I do not yet have much experience with the camera definition files but it's not clear to me how a user could add a command to a mission that would allow changing the lens (aka image sensor).  I think that only MAV_CMDs can be added to missions.

hamishwillee:
> Hi, I must admit that I do not yet have much experience with the camera definition files but it's not clear to me how a user could add a command to a mission that would allow changing the lens (aka image sensor) held within a camera definition file. I think that only MAV_CMDs can be added to missions.

Is there an actual need for different lenses within the same mission?

Either way it is a good question. Same as "how would we swap IR to Visible light in a mission"

rmackay9:
Hi @hamishwillee,

Yes, I think it should be possible to switch the active lens (aka image sensors) during a mission.  We already have commands for zooming, focusing, taking pictures or recording video.  Why wouldn't we have commands for changing the active lens?  The most likely the case would be where the command is placed at the beginning of the mission to ensure the correct lens is used.

For example yesterday I performed two identical missions with the Xacti, one with a picture-in-picture RGB and then again with just RGB.  I used ArduPilot's built in support for changing the lens but having it built into the mission means I have one less thing to worry about.

Re using the zoom command, it's probably obvious but just in case, we're not talking about physically changing a lens on the front of a camera like some professional photographers do.  This is just changing the active sensor between RGB and IR.

hamishwillee:
@rmackay9 Thanks. I don't want us to fix this for lens and then have to come back and sort it out for IR/Visible light too, and whatever comes next. So some generic solution would be good, but that needs to work sensibly with the camera definitions, and allow a GCS to display what is active and not active. 
On the other hand, generic commands can make the mission hard to read and be usable on any system.

Let's see what @MatejFranceskin and @dogmaphobic might suggest for the mission case.

rmackay9:
@hamishwillee,

Thanks for that feedback.

Re the generic solution, command_long and command_int are already generic messages so there is almost no cost to adding new items to the MAV_CMD list so I don't think we need to worry too much about the slippery slope.  Even if we did worry about that though, I think this lens (or active sensor) is still common enough that we should have a command for it.

Anway, I don't mean to be too pushy.  Happy to wait for more feedback.

hamishwillee:
@MatejFranceskin Repeating - we need to document the IR/Visible light use case
-  Would it be possible for you to send me a camera definition file (or attach) with this field, or update https://mavlink.io/en/services/camera_def.html#full_example to add this to the example?
- Do you have any thoughts on how you might set the lens/sensor case in a mission without creating a new command?

hamishwillee:
@rmackay9 The cost of many commands is that we have a simple API now, and we don't want to complicate it unless needed. There are millions of settings you might want to add. We'll do it if we have to. 

**I've added to agenda for next week's meeting - @MatejFranceskin @dogmaphobic your attendance or solutions welcome.**

If we do it this way we would need to:
- Decide if this is specific to "lens" or can be used for "ir/visible light" - essentially its a way of saying that there are a number of mutually exclusive sensors you can select - I'd call it "MAV_CMD_SET_SENSOR"
- Agree this is worth making part of the basic capabilities 
- We will need to add a flag and cross link: https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS

By generic solution I mean a command that could be used to hook into the existing camera definition system so that you could set any feature you like via a single command. We could do this by:
- defining `CAM_LENS` as a [common parameter](https://mavlink.io/en/services/camera_def.html#common-parameters) in the definition file - i.e. a parameter of this name will always be used for setting CAM_LENS (if supported) on any camera
- You could then add support for [MAV_CMD_DO_SET_PARAMETER](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_PARAMETER) specifying the value of that parameter in your mission.
  - We'd probably make this a bit easier by providing an enum you could use.
- Ore we could have a specific set camera feature kind of command that does the same thing - the difference being that the GCS understand this is for cameras, and if supporting the camera definition could fetch possible options for the UI.

nexton-winjeel:
@rmackay9, @hamishwillee:

The MAVLink camera protocol already has the concept of streams. I think rather than making this about lenses, what is needed is a `MAV_CMD_VIDEO_SET_ACTIVE_STREAM` command which would change the active stream but wouldn't change the streaming state.

We've bumped into this issue before, and we've worked around it using `MAV_CMD_VIDEO_START_STREAMING`/`MAV_CMD_VIDEO_STOP_STREAMING`, but that's not ideal.

The specific use case we had was:
* a payload with EO and IR cameras;
* payload records snaphots to onboard storage in response to `MAV_CMD_IMAGE_START_CAPTURE`;
    * snapshot is taken of current stream;
* video stream output is disabled to reduce radio bandwidth;
* fly a mission to take snaphots from both cameras.

With the current camera protocol, there is no way to specify which stream to take a snapshot from without also enbling output from that stream. (Note that this particular issue could also be fixed by adding a `Stream ID` field to `MAV_CMD_IMAGE_START_CAPTURE`.)

hamishwillee:
Thanks @nexton-winjeel - that idea certainly has possibilities!  I can see how you might use either a stream id in image capture or by adding the "set active stream command".

How would you identify a particular video stream as a lens? i.e. in https://mavlink.io/en/messages/common.html#VIDEO_STREAM_INFORMATION we need to find out our set of lenses to use.

nexton-winjeel:
@hamishwillee:
> How would you identify a particular video stream as a lens?

You associate each lens with a stream ID, and output a separate `VIDEO_STREAM_INFORMATION` message (with appropriate `name`) for each stream.

nexton-winjeel:
Note that the above solution is the simplest way to address @rmackay9's use case given the current MAVLink design.

The real issue is that the video streams  (per [`VIDEO_STREAM_INFORMATION`](https://mavlink.io/en/messages/common.html#VIDEO_STREAM_INFORMATION)) currently conflate two items, which should probably be separate:
1. Video _channels_, i.e. what am I looking at (EO, IR, picture-in-picture); 
2. Video _outputs_, i.e. how do I receive and decode the video.

In @rmackay9's use case, he has multiple channels but only a single output (as seems to be the case with most gimbals).

hamishwillee:
@dogmaphobic @MatejFranceskin Suggestion above - use streams as a "proxy" for lenses. Thoughts?

If you like this:
1. Preference for setting stream on the image capture.
2. ... or having new message to set the active stream.

nexton-winjeel:
@hamishwillee: Setting the active stream without changing the streaming state needs to be either:
* a new message; or 
* a new param on the [`MAV_CMD_VIDEO_START_STREAMING`](https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_START_STREAMING) command. 

Setting it via a new param in [`MAV_CMD_IMAGE_START_CAPTURE`](https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE) doesn't work for cameras that _do_ have a stream but _don't_ support image capture.

hamishwillee:
@nexton-winjeel 

Thank you. I can see you have a much clearer view on this than me. So here's a bit of a ramble to compose my thinking and see what our options are now. If we can overcome confliction issue now and keep this simple, that would be better.

So the proposal around current streams design looks workable:
- We can get id of streams for lenses. We should probably define a "channel" (currently status) for lenses to make parsing easier.
  - This is going to be messy though, because if you support streaming multiple video types you'll have a big matrix.
- I _think_ setting the stream id in the image capture command or a new `set_video_stream(s)_for_image_capture` could both work. 
  - If we just want to capture from a single stream then probably adding id to the image capture command makes sense. Perhaps with 0 meaning don't change or reset to default? 
  - `set_video_stream(s)_for_image_capture` makes sense as a name to clarify we're setting the image to capture, not starting a new stream.
  - Can we support image capture of multiple streams?
- Looking at the current design it seems to me that `VIDEO_STREAM_STATUS` is misused. That should really just be status - active, standby, faulted. VIDEO_STREAM_TYPE should be VIDEO_STREAM_ENCODING_TYPE, and we should have a new type for the channel VIDEO_STREAM_CHANNEL_TYPE with IR, Visible, Lens, whatever.
  - Is it worth making this split now?
  - And if not, is it worth add lens as a flag?

How would we design to deconflate this - and could this be done in a compatible way right now that makes things better/easier for @rmackay9 and does not lock us into complications with streams at least for this use case.

What I'm thinking is:
- Add [CAMERA_INFORMATION.channel_count](https://mavlink.io/en/messages/common.html#CAMERA_INFORMATION) field which tells how many channels you support - or 0 if not supported.
- New message `CAMERA_CHANNEL` which you can query in the same way as the video status/information. But you don't need to bother if `channel_count==0` or `1`
- Add _channel id_ for image capture command
- For the `VIDEO_INFORMATION` we could include the camera channel used if we wanted, instead of having in status. 

I am a bit concerned that things like  IR/Visible light can be combined with lens 1 to 5?

rmackay9:
Thanks for trying to help clarify this.  If possible I think we shouldn't use the word "channel".  Cameras have lenses and image sensors which people can guess the meaning of while "channel" is just a bit too abstract I think.

hamishwillee:
@rmackay9 Cameras have  image sensors (Visible light/IR) but they can also have composites of those - such as Picture in Picture. So that's more "channel like". I guess we could treat that as a virtual sensor. 
Logically I also quite like separating lenses from our sensors, because even though right now you'll have one or the other, one day you might well have both.

But then you need a way to set both. Realistically you could set the combinations to be used by the video system too. 

It might be that we're better off just using streams because the abstraction isn't worth the complication. Let's see what Nick says.

I've also asked @MatejFranceskin if he can come to the next MAV call to work this out.

rmackay9:
@hamishwillee,

Sorry, I replied too quickly.  After thinking about his some more, it is similar to a TV channel.  The ViewPro for example accepts a number from 0 to 5 which produces an output which can be 0:RGB, 1:Thermal, 2: PIP(RGB+Thermeal), 3:PIP(Thermal+RGB), etc.  The numbers are quite arbitrary in that they don't tell you anything about what you're going to get which is similar to a TV channel.

BTW, ViewPro calls it an "Image Sensor"
image

Xacti calls it a "SensorMode"
image

hamishwillee:
Sensor mode works for me as a name if we go down this path. Interesting that they add PIP "both ways".

nexton-winjeel:
@hamishwillee: 
> What I'm thinking is:
> - Add [CAMERA_INFORMATION.channel_count](https://mavlink.io/en/messages/common.html#CAMERA_INFORMATION) field which tells how many channels you support - or 0 if not supported.
> - New message CAMERA_CHANNEL which you can query in the same way as the video status/information. But you don't need to bother if channel_count==0 or 1
> - Add channel id for image capture command
> - For the VIDEO_INFORMATION we could include the camera channel used if we wanted, instead of having in status.
> - I am a bit concerned that things like IR/Visible light can be combined with lens 1 to 5?

That doesn't really help. The current `CAMERA_INFORMATION` and `VIDEO_STREAM_INFORMATION` are more suited to describing point-and-shoot cameras that are attached to a drone. They're missing some info for gimbals with multiple integrated cameras (I'll flesh it out more in #2083). 

For now, I think the best solution is to add the minimum needed to support @rmackay9's use case, which is either:
* a way to switch the active stream without changing the streaming state; or
* a way to specify which stream the `CAMERA_IMAGE_CAPTURE` command will use.

hamishwillee:
> For now, I think the best solution is to add the minimum needed to support @rmackay9's use case, which is either:
> 
> - a way to switch the active stream without changing the streaming state; or
> - a way to specify which stream the CAMERA_IMAGE_CAPTURE command will use.

I agree, and with selecting the second option as a way to do it. 
I just think we need to make it clear in that command what is being captured - i.e. the camera is not taking a particular frame in that stream, it is most likely just capturing raw sensor data. We're only using the stream to identify the particular configuration of the camera (lens/ir/whatever) in which the picture is taken

hamishwillee:
We discussed this in https://github.com/mavlink/mavlink/wiki/20240228-Dev-Meeting and agreed that the most pragmatic way out of this is to accept an approach like this. Action on @rmackay9 to modify name to reflect it isn't a lens (is a source), to set options as an enum, and Picture-in-picture via second param.

FWIW I haven't got a clear idea of exactly what this will look in terms of the pip parameter - mostly because if you have 4 sources say, not sure how this would specify two of them and their order.

This will do the same thing as setting IR/visible light etc in the camera definition file and will set behaviour for the camera in both video and image capture.

As an aside, we did discuss doing this by setting a stream and also by providing direct access to setting the parameters. The parameter option was agreed to be most flexible, but requires quite a lot of work in GCS to make useful/meaningful for users. The selected choice is more pragmatic.

rmackay9:
Thanks for the discussion yesterday.  I've rebased on master and added a follow-up commit with what I think we agreed upon.  All comments/changes welcome and feel free to squash the commits together if you like (or I can if you prefer).

message_definitions/v1.0/common.xml
```diff
@@ -2089,6 +2089,12 @@
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
         <param index="2" label="Usage" enum="STORAGE_USAGE_FLAG">Usage flags</param>
       </entry>
+      <entry value="534" name="MAV_CMD_SET_CAMERA_SOURCE" hasLocation="false" isDestination="false">
+        <description>Set camera source. Changes the camera's active sources on cameras with multiple image sensors.</description>
+        <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras</param>
```
hamishwillee, line 2094: Consistency
```suggestion
        <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -2089,6 +2089,12 @@
...
+        <param index="2" label="primary source" enum="CAMERA_SOURCE">Primary Source</param>
+        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source.  Non-zero for Picture-in-picture</param>
```
hamishwillee, line 2096:
```suggestion
        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source. If non-zero the second source will be displayed as picture-in-picture-in-picture</param>
```

rmackay9:
@hamishwillee, thanks for the review, I've committed the suggestions.

message_definitions/v1.0/common.xml
```diff
@@ -2089,6 +2089,12 @@
         <param index="1" label="Storage ID" minValue="0" increment="1">Storage ID (1 for first, 2 for second, etc.)</param>
         <param index="2" label="Usage" enum="STORAGE_USAGE_FLAG">Usage flags</param>
       </entry>
+      <entry value="534" name="MAV_CMD_SET_CAMERA_SOURCE" hasLocation="false" isDestination="false">
+        <description>Set camera source. Changes the camera's active sources on cameras with multiple image sensors.</description>
+        <param index="1" label="device id">Component Id of camera to address or 1-6 for non-MAVLink cameras, 0 for all cameras.</param>
+        <param index="2" label="primary source" enum="CAMERA_SOURCE">Primary Source</param>
+        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source. If non-zero the second source will be displayed as picture-in-picture</param>
```
hamishwillee, line 2096:
```suggestion
        <param index="3" label="secondary source" enum="CAMERA_SOURCE">Secondary Source. If non-zero the second source will be displayed as picture-in-picture.</param>
```

hamishwillee (approved):
I think this is as discussed. Approving. @julianoes @auturgy @MatejFranceskin Any last comments before this is merged?

(@rmackay9 Not expecting any, just making sure that we're all on board)

julianoes (approved):
Nice, looks good now!

auturgy (approved):

hamishwillee:
Thanks for merging. Docs for this in https://github.com/mavlink/mavlink-devguide/pull/530. There is a question/issue in that.

PR #2081, opened 2024-02-16
common.xml: add a command to deal with safety switch

peterbarker:
decoding the old MODE message is a terrible way to do this

ArduPilot PR is here: https://github.com/ArduPilot/mavlink/pull/352

Please do not merge this PR yet, it's for-discussion.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2397,6 +2397,16 @@
         <param index="7">Reserved (set to 0)</param>
       </entry>
       <!-- END of UAVCAN range -->
+      <entry value="5300" name="MAV_CMD_DO_SET_SAFETY_SWITCH_STATE" hasLocation="false" isDestination="false">
+        <description>Change state of safety switch.</description>
+        <param index="1" label="Desired State" enum="SAFETY_SWITCH_STATE">New safety switch state.</param>
+        <param index="2">Empty.</param>
+        <param index="3">Empty.</param>
+        <param index="4">Empty</param>
+        <param index="5">Empty.</param>
+        <param index="6">Empty.</param>
+        <param index="7">Empty.</param>
+      </entry>
       <entry value="10001" name="MAV_CMD_DO_ADSB_OUT_IDENT" hasLocation="false" isDestination="false">
         <description>Trigger the start of an ADSB-out IDENT. This should only be used when requested to do so by an Air Traffic Controller in controlled airspace. This starts the IDENT which is then typically held for 18 seconds by the hardware per the Mode A, C, and S transponder spec.</description>
         <param index="1">Reserved (set to 0)</param>
@@ -4984,6 +4994,17 @@
       </entry>
       <entry value="5" name="MISSION_STATE_COMPLETE">
         <description>Mission has executed all mission items.</description>
+      </entry>
+    </enum>
+    <enum name="SAFETY_SWITCH_STATE">
+      <description>
+	Possible safety switch states.
+      </description>
+      <entry value="0" name="SAFETY_SWITCH_STATE_SAFE">
+        <description>Safety switch is engaged and vehicle should be safe to approach.</description>
+      </entry>
+      <entry value="1" name="SAFETY_SWITCH_STATE_DANGEROUS">
+        <description>Safety switch is NOT engaged and motors, propellers and other actuators should be considered active.</description>
       </entry>
     </enum>
   </enums>


message_definitions/v1.0/common.xml
```diff
@@ -2397,6 +2397,16 @@
         <param index="7">Reserved (set to 0)</param>
       </entry>
       <!-- END of UAVCAN range -->
+      <entry value="5300" name="MAV_CMD_DO_SET_SAFETY_SWITCH_STATE" hasLocation="false" isDestination="false">
+        <description>Change state of safety switch.</description>
+        <param index="1" label="Desired State" enum="SAFETY_SWITCH_STATE">New safety switch state.  SAFETY_SWITCH_STATE_UNKNOWN should be ignored.</param>
```
rmackay9, line 2402: I think we can remove the SAFETY_SWITCH_STATE_UNKNOWN reference here.  I'm not sure why a caller would want to send in a unknown desired state for the switch.

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4996,17 @@
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
+    <enum name="SAFETY_SWITCH_STATE">
+      <description>
+	Possible states the safety switch can be in.
```
rmackay9, line 5001: nitpick: we could change this wording to just "safety switch states" maybe.

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4996,17 @@
...
+      </description>
+      <entry value="0" name="SAFETY_SWITCH_STATE_SAFEISH">
```
rmackay9, line 5003: I definitely like the idea of easy to understand states but some other possibilities are:

0:Engaged
1:Disengaged

The above is probably close to what people would expect but it is difficult to understand which is which.

Maybe we should just change, "SAFETY_SWITCH_STATE_SAFE**ISH**" to simply "SAFETY_SWITCH_STATE_SAFE"?

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4996,17 @@
...
+        <description>Safety switch is engaged and vehicle safe-ish to approach..</description>
+      </entry>
+      <entry value="1" name="SAFETY_SWITCH_STATE_DANGEROUS">
+        <description>Safety switch is NOT engaged and whirling blades of death and/or flamethrowers may be active.</description>
```
rmackay9, line 5007: This could be changed to, "Safety switch is NOT engaged and motors and propellers may be active"

message_definitions/v1.0/common.xml
```diff
@@ -4986,6 +4996,17 @@
         <description>Mission has executed all mission items.</description>
       </entry>
     </enum>
+    <enum name="SAFETY_SWITCH_STATE">
+      <description>
+	Possible states the safety switch can be in.
+      </description>
+      <entry value="0" name="SAFETY_SWITCH_STATE_SAFEISH">
+        <description>Safety switch is engaged and vehicle safe-ish to approach..</description>
```
rmackay9, line 5004: This could be changed to, "Safety switch is engaged and motors and propellers are not active"

peterbarker:
On Fri, 16 Feb 2024, Randy Mackay wrote:

> I think we can remove the SAFETY_SWITCH_STATE_UNKNOWN reference here. I'm not sure why a caller would want to send in a unknown desired state for the switch.

Will remove it; I added it thinking it would be useful for reporting (not 
done in this message/ArduPilot change.

peterbarker:
On Fri, 16 Feb 2024, Randy Mackay wrote:

> I definitely like the idea of easy to understand states but some other possibilities are:
> 
> 0:Engaged
> 1:Disengaged
> 
> The above is probably close to what people would expect but it is difficult to understand which is which.

Exactly right.  And this is something people should be absolutely sure 
about IMO.

> Maybe we should just change, "SAFETY_SWITCH_STATE_SAFEISH" to simply "SAFETY_SWITCH_STATE_SAFE"?

I *know* that the vehicle is only safe-ish with the safety switch in the 
"you're likely to keep all your fingers" position.  In ArduPilot we have a 
mask saying that some things are allowed through, for example.

I guess that's not really the safety switch's fault.  Will change to SAFE.

peterbarker:
On Fri, 16 Feb 2024, Randy Mackay wrote:

> This could be changed to, "Safety switch is engaged and motors and propellers are not active"

It could be, but could be an absolute lie as well.  I'm trying not to give 
any guarantees in these comments.

message_definitions/v1.0/common.xml
peterbarker: Done

message_definitions/v1.0/common.xml
peterbarker: Done

message_definitions/v1.0/common.xml
peterbarker: Done

message_definitions/v1.0/common.xml
peterbarker: Done

message_definitions/v1.0/common.xml
peterbarker: I've tempered this one a little to "should not be active".

rmackay9 (approved):
This looks good to me.

I don't think we need to add an element of uncertainty to the descriptions.  It's not a legal contract, it's just a definition of what should be happening when things are going well.

message_definitions/v1.0/common.xml
```diff
@@ -2397,6 +2397,16 @@
         <param index="7">Reserved (set to 0)</param>
       </entry>
       <!-- END of UAVCAN range -->
+      <entry value="5300" name="MAV_CMD_DO_SET_SAFETY_SWITCH_STATE" hasLocation="false" isDestination="false">
+        <description>Change state of safety switch.</description>
```
hamishwillee, line 2401: I think this is good but perhaps should be explicit about our expectation on the flight stack. Specifically this shouldn't be a proxy for failsafe, or something you couldn't physically trigger if you wanted. My concern here is the same as disarming in flight. Would recommending only allowing when disarmed work, or landed?

```suggestion
        <description>
          Change state of safety switch.
          Autopilots should NACK a refuse a request to engage unless disarmed.
        </description>
```

TLDR:
1. Point of a safety switch is a first step in making vehicle able to arm, so sharp things don't spin. Many vehicles don't use it and I've always found the concept a bit scary - having to go near something with spinny bits.
2. But if we're going to have GCS setting this anyway, then it doesn't make sense to make it hard and error prone. So I'm for.
3. But for safety better if this can't be done in the air.

hamishwillee:
@julianoes @auturgy - what do you think?

hamishwillee:
PS This would also make it easier to switch to a prearm mode on PX4.

message_definitions/v1.0/common.xml
peterbarker: >     1. Point of a safety switch is a first step in making vehicle able to arm, so sharp things don't spin. Many vehicles don't use it and I've always found the concept a bit scary - having to go near something with spinny bits.

Considering it's all software, there's just no difference between arming and the safety switch.  If anything, the safety switch is marginally less useful than the arming concept as most vehicles don't have an external mechanism for arming, but there often is a physical safety switch.  Putting your hand into the middle of the whirling blades of death to actively disable the safety system *always* struck me as a silly idea....

Also, the safety switch is often one of the last steps in arming a vehicle on ArduPilot.  You can let the vehicle run through all of its checks (calibration checks, AHRS happiness checks) while the vehicle is in its "safe to approach" safety-switch state.

>     3. But for safety better if this can't be done in the air.

That's actually a difficult problem for planes as whether they're flying or not is pure heuristics (for the vast majority of vehicles).

In terms of the suggested change, "Autopilots should NACK a refuse a request to engage unless disarmed" - that's a really good idea.  We *do* have options in ArduPilot to disable the safety switch while the vehicle is armed, and it *is* the default on Copter and Plane (but not on some other vehicles e.g. tracker).  Physical faults with safety switches cost some vehicles...

However, armed-with-safety-on is actually a valid state to be in, and part of some people's workflows.  Not saying it's a great workflow, but we can't break it!

Another reason for allowing safety switch change when armed is that the vehicle's behaviour can subtly change once the vehicle is armed (estimators can change thresholds, for example), and it can be useful to test that stuff out while having *some* level of protection against motors spinning up.

I wouldn't mind more words in there about  general expectations around safety switch stuff, but I don't think we should be prescriptive about behaviours.

Perhaps:
```
        <description>
          Change state of safety switch.
          When the safety switch is in the "safe" position the vehicle should be safe to approach - e.g motor movement inhibited, pumps disabled.
        </description>
```

peterbarker:
> PS This would also make it easier to switch to a prearm mode on PX4.

This is a direct replacement for `SET_MODE.base_mode & MAV_MODE_FLAG_DECODE_POSITION_SAFETY` and (`MAV_CMD_DO_SET_MODE param1 == MAV_MODE_FLAG_DECODE_POSITION_SAFETY` combined with `MAV_CMD_DO_SET_MODE param2` == 0 or 1 to turn on/off.

It doesn't bare in *more* functionality, just disentangles it from mode setting!

message_definitions/v1.0/common.xml
hamishwillee: @peterbarker It isn't quite the same as messing with the modes etc - you've gone from making it quite hard to make the vehicle fall out the air to something that can be done trivially and possibly by accident.

Perhaps we should adopt the same solution as https://mavlink.io/en/messages/common.html#MAV_CMD_COMPONENT_ARM_DISARM - see the `force` field. Essentially you can engage this with 0 by default and the safety switch will only work when disarmed. But you need to use this when armed you can set a magic number to make it disarm any time. Far harder to say "but I didn't understand the risk".

Thoughts?

I've added this to the agenda to discuss in mav call next week if not resolved earlier.

peterbarker:
On Tue, 20 Feb 2024, Hamish Willee wrote:

> @peterbarker It isn't quite the same as messing with the modes etc - you've
> gone from making it quite hard to make the vehicle fall out the air to
> something that can be done trivially and possibly by accident.

How so?  This doesn't change the user interface around this functionality, 
just the way the GCS communicates the command to the vehicle.

> Perhaps we should adopt the same solution as
> https://mavlink.io/en/messages/common.html#MAV_CMD_COMPONENT_ARM_DISARM -
> see the force field. Essentially you can engage this with 0 by default and
> the safety switch will only work when disarmed. But you need to use this
> when armed you can set a magic number to make it disarm any time. Far harder
> to say "but I didn't understand the risk".
> 
> Thoughts?

We're not actually making this any more or less dangerous, just changing 
the way the command is conveyed.

The force option is quite an interesting one.  At the moment ArduPilot 
would just reject the command if you were armed and hadn't said it was OK 
to use the safety switch.

Adding "force" here would require UI changes in GCS, *and* changes to the 
autopilot software to understand the "force" option.  I think we could 
reasonably add "force" later if people think it's a good idea?

> I've added this to the agenda to discuss in mav call next week if not
> resolved earlier.

Thanks!  I'll bring up the "force" thing at tonight's DevCall.

hamishwillee:
> We're not actually making this any more or less dangerous, just changing the way the command is conveyed.

Correct. It is already too dangerous, and this is your opportunity to do the right thing and make it safer :-)

@julianoes @auturgy may take a different view.

rmackay9:
[MAV_CMD_DO_FLIGHT_TERMINATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_FLIGHTTERMINATION) also doesn't have a magic number to ensure that the caller really means it.

hamishwillee:
> [MAV_CMD_DO_FLIGHT_TERMINATION](https://mavlink.io/en/messages/common.html#MAV_CMD_DO_FLIGHTTERMINATION) also doesn't have a magic number to ensure that the caller really means it.

Yes, but the intent and result is clear from the name. That's not so true with arm or safety switch.

peterbarker:
On Tue, 20 Feb 2024, Hamish Willee wrote:

>       We're not actually making this any more or less dangerous, just changing the way the command is conveyed.
> Correct. It is already too dangerous, and this is your opportunity to do the right thing and make it safer :-)

Aaahh... adding a "force" option will make it less safe, not more safe :-)

Unless you need to get things going in a hurry, of course.

hamishwillee:
> On Tue, 20 Feb 2024, Hamish Willee wrote: We're not actually making this any more or less dangerous, just changing the way the command is conveyed. Correct. It is already too dangerous, and this is your opportunity to do the right thing and make it safer :-)
> Aaahh... adding a "force" option will make it less safe, not more safe :-) Unless you need to get things going in a hurry, of course.

That doesn't make sense to me. By default you have force off, so that this can only be triggered when vehicles are in a disarmed state, which is suitable for most vehicles. On a vehicle that requires the you to be able to set this "whenever" you choose to set the `force` option. Some random who doesn't know what the vehicle needs is likely to start without the force option.

But perhaps I'm being dim. Not my call.

@julianoes @auturgy What are your thoughts on this.

julianoes:
I'm ok with this. But I always wonder why we call it safety switch. Isn't it a safety button?

hamishwillee (approved):
OK, this is agreed OK by @julianoes and @auturgy so it is going in. 

@julianoes No idea why we call it a safety switch. It was called that when I started in this business. Button is more accurate since the signal is transitory, but unless there is broad agreement I'm not going to change our naming of this.

hamishwillee:
Arrrggg. @peterbarker I just pushed this in by accident, after you said hold off. Had I taken a bit more time I would have suggested that this should be WIP until implemented and/or in development.

Should I move it out again and put it in develpment.xml for now? (i.e. I will do this unless you say "keep it in")

SJayYang:
@peterbarker Curious - how is this message handled on the PX4 side?

peterbarker:
> @peterbarker Curious - how is this message handled on the PX4 side?

PX4-Autopilot doesn't seem to reference the enumeration names this PR introduced.  ArduPilot merged support some time ago - it will be in our next stable release.

I'll be merging the MAVProxy support for it soon after that stable release (if not before).  https://github.com/ArduPilot/MAVProxy/pull/1558

hamishwillee:
@SJayYang As @peterbarker  says, this is not implemented in PX4 - the command ACKs with unsupported.
If you're interested in providing an implementation, please cross link to here.

PR #2082, opened 2024-02-19
common: MAV_CMD_DO_SET_SYS_CMP_ID: Add first version

patrickelectric:
Allow changes of the system id and component id

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -356,9 +356,9 @@
           This allows moving of a system and all its components to a new system id, or moving a particular component to a new system/component id.
           Recipients must reject command addressed to broadcast system ID.
         </description>
-        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID that should be used by the device. Zero (0) is ignored, System ID should not changed.</param>
-        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID that should be used by the device. Zero (0) can be used to ignore this field.</param>
-        <param index="3" label="Reboot">Do a component reboot after ID change. Any value different than zero (0) will result in a vehicle reboot.</param>
+        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID for target component(s). 0: ignore and reject command (broadcast system ID not allowed).</param>
+        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID for target component(s). 0: ignore (component IDs don't change).</param>
+        <param index="3" label="Reboot">Reboot components after ID change. Any non-zero value triggers the reboot.</param>
         <param index="4" reserved="true" default="NaN"/>
       </entry>
       <entry value="12900" name="MAV_CMD_ODID_SET_EMERGENCY" hasLocation="false" isDestination="false">
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -350,6 +350,17 @@
         <param index="6" reserved="true"/>
         <param index="7" reserved="true"/>
       </entry>
+      <entry value="610" name="MAV_CMD_DO_SET_SYS_CMP_ID" hasLocation="false" isDestination="false">
+        <description>
+          Set system and component id.
+          This allows moving of a system and all its components to a new system id, or moving a particular component to a new system/component id.
+          Recipients must reject command addressed to broadcast system ID.
+        </description>
+        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID that should be used by the device. Zero (0) is ignored, System ID should not changed.</param>
+        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID that should be used by the device. Zero (0) can be used to ignore this field.</param>
+        <param index="3" label="Reboot">Do a component reboot after ID change. Any value different than zero (0) will result in a vehicle reboot.</param>
+        <param index="4" reserved="true" default="NaN"/>
+      </entry>
       <entry value="12900" name="MAV_CMD_ODID_SET_EMERGENCY" hasLocation="false" isDestination="false">
         <description>Used to manually set/unset emergency status for remote id.
           This is for compliance with MOC ASTM docs, specifically F358 section 7.7: "Emergency Status Indicator".


message_definitions/v1.0/common.xml
```diff
@@ -1378,6 +1378,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="175" name="MAV_CMD_DO_SET_SYS_CMP_ID" hasLocation="false" isDestination="false">
```
hamishwillee, line 1381: @patrickelectric Can you please update this value to some unused value in the range 300 to 10000 (these lower ranges are shared for legacy reasons, and can result in conflicts).

So what's the use case/interaction that you need this for?

I mean that it is true that there is no way to set these values via MAVLink. However you have to have an ID already for it to be changed, and that's got to be allocated by someone. FWIW there have been proposals of auto-allocation of ids on a network from a pool by some allocator system. Not progressed beyond discussion.

patrickelectric: Thanks for the info @hamishwillee!

I just updated the message as requested.

This message tries to cover the problem where there is no standard way for people to change devices or system ID.

In the case where a vehicle contains multiple components (hardware or software) that talks via mavlink, a message like this one is valuable, a single:

```py
# pseudocode
send("MAV_CMD_DO_SET_SYS_CMP_ID", sysID=2, cmpID=0)
```

Would result in all components on a vehicle infrastructure to change to a desirable sysID while maintaining the componentID. Now all components on the vehicle will have the same system ID.

Doing that in the route fix part of the problem, but you still have the problem of intercommunication of mavlink data in the same "bus" or channel, where a vehicle will receive mavlink messages from others, and a systemID collision can happen.

I believe that the message is still valid and hand for fleets that share the same network and are not bigger than a couple of hundreds vehicles.

hamishwillee: Thanks @patrickelectric

So
- if you already know the id of the target system (set with `target_system`) this is to allows you to change the system ID of all components in a system, keeping all component ids the same
- if you already know the id of the target system (set with `target_system`) and if of the chosen component (`target_component`) you can change the component id and system to any other component and system.

But if you did something like this to the broadcast Id/s you'd set all systems to 0.

```
send("MAV_CMD_DO_SET_SYS_CMP_ID", sysID=2, cmpID=0)
```

I can see a use - ie you could use this to configure something with a system id and component via MAVLink if it was the only thing on the network. Risky though!

is that about right?

julianoes (approved):
Cool!

message_definitions/v1.0/common.xml
```diff
@@ -2098,6 +2098,16 @@
...
+      <entry value="610" name="MAV_CMD_DO_SET_SYS_CMP_ID" hasLocation="false" isDestination="false">
+        <description>Set system and component id.</description>
+        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID that should be used by the device.</param>
+        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID that should be used by the device. Zero (0) can be used to ignore this field.</param>
+        <param index="2" label="Reboot">Do a component reboot after ID change. Any value different than zero (0) will result in a vehicle reboot.</param>
+        <param index="4">Empty</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
```
hamishwillee, line 2110: @patrickelectric Discussed in [20240228-Dev-Meeting](https://github.com/mavlink/mavlink/wiki/20240228-Dev-Meeting). We love it - in particular it provides a common mavlink id for setting ids of components which we lack - many components just use the default id for their type, so you might get clashes. This allows reconfiguration via mavlink if you can attach the component to the network.

Main concerns:
- This should go into development.xml first and move to common after implementation. Can you please move it? (and I can then merge).
- We should be explicit that the command must be rejected if received with a broadcast system id (0). Otherwise you could fire this at a network and screw up all system ids. If you want to use this to explic
- Docs can be made a little more clear on use
- Param 2 duplicated.

This should

```suggestion
      <entry value="610" name="MAV_CMD_DO_SET_SYS_CMP_ID" hasLocation="false" isDestination="false">
        <description>Set system and component id.
        This allows moving of a system and all its components to a new system id, or moving a particular component to a new system/component id.
        Recipients must reject command addressed to broadcast system ID.
        </description>
        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID that should be used by the device.</param>
        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID that should be used by the device. Zero (0) can be used to ignore this field.</param>
        <param index="3" label="Reboot">Do a component reboot after ID change. Any value different than zero (0) will result in a vehicle reboot.</param>
        <param index="4" reserved="true" default="NaN"/>
      </entry>
```

patrickelectric: > This should go into development.xml first and move to common after implementation. Can you please move it? (and I can then merge).

Done

> We should be explicit that the command must be rejected if received with a broadcast system id (0). Otherwise you could fire this at a network and screw up all system ids. If you want to use this to explic

Yes, that's correct. I put "System ID" parameter with `minValue=1` for that reason. Thanks for improving the description about that.

message_definitions/v1.0/common.xml
patrickelectric: > if you already know the id of the target system (set with target_system) this is to allows you to change the system ID of all components in a system, keeping all component ids the same
if you already know the id of the target system (set with target_system) and if of the chosen component (target_component) you can change the component id and system to any other component and system.

Yes, you could also broadcast `target_system` as zero (0) to trigger a "system ID" change on all components on the network.

Sorry if my pseudocode example did not took that in consideration. I'll use [mavlink2rest](https://github.com/mavlink/mavlink2rest) syntax for now on.

```ts
// Change all components on the network to the new "System ID" (4)
{
  "header": { // I'm a GCS, look at me
    "system_id": 255,
    "component_id": 1,
    "sequence": 1
  },
  "message": {
    "type": "COMMAND_LONG",
    "param1": 4, // All vehicles / Components that match "target_system" and "target_component" need to change their "System ID" to the new value (4) 
    "param2": 0, // Zero (0) is used to ignore this field, the "Component ID" will not change 
    "param3": 1, // Perform a reboot after the operation is complete
    "param4": 0,
    "command": {
      "type": "MAV_CMD_DO_SET_SYS_CMP_ID"
    },
    "target_system": 0, // Zero (0) is used for broadcast, all "System IDs" vehicles need to process the operation
    "target_component": 0, // Zero (0) is used for broadcast, all components with the matched "target_system" need to process the operation 
    "confirmation": 1 // Send an ack before reboot
  }
}

// Change all components with "System ID" of 1 to 4. Keeping their original "Component ID"
{
  "header": { // I'm a GCS, look at me
    "system_id": 255,
    "component_id": 1,
    "sequence": 1
  },
  "message": {
    "type": "COMMAND_LONG",
    "param1": 4, // All vehicles / Components that match "target_system" and "target_component" need to change their "System ID" to the new value (4) 
    "param2": 0, // Zero (0) is used to ignore this field, the "Component ID" will not change 
    "param3": 1, // Perform a reboot after the operation is complete
    "param4": 0,
    "command": {
      "type": "MAV_CMD_DO_SET_SYS_CMP_ID"
    },
    "target_system": 1, // All components system "System ID" equal to 1 should deal with this action
    "target_component": 0, // Zero (0) is used for broadcast, all components with the matched "target_system" need to process the operation 
    "confirmation": 1 // Send an ack before reboot
  }
}

// Change only the component with "System ID" with the value 1 to 4.
{
  "header": { // I'm a GCS, look at me
    "system_id": 255,
    "component_id": 1,
    "sequence": 1
  },
  "message": {
    "type": "COMMAND_LONG",
    "param1": 4, // All vehicles / Components that match "target_system" and "target_component" need to change their "System ID" to the new value (4) 
    "param2": 194, // Change "Component ID" to new value (`MAV_COMP_ID_ONBOARD_COMPUTER4`)
    "param3": 1, // Perform a reboot after the operation is complete
    "param4": 0,
    "command": {
      "type": "MAV_CMD_DO_SET_SYS_CMP_ID"
    },
    "target_system": 4, // All components with the "System ID" equal to 4 should deal with this action
    "target_component": 192, // Only the component 192 (`MAV_COMP_ID_ONBOARD_COMPUTER2`) should process this command 
    "confirmation": 1 // Send an ack before reboot
  }
}
```


> But if you did something like this to the broadcast Id/s you'd set all systems to 0.
send("MAV_CMD_DO_SET_SYS_CMP_ID", sysID=2, cmpID=0)
I can see a use - ie you could use this to configure something with a system id and component via MAVLink if it was the only thing on the network. Risky though!

No, that's not correct. My idea was to change all components on the network to the same "System ID" (2), in the example, where the current "Component ID" does not change.
This is better explained on the second example.

Thinking in a better way, we should allow "System ID" with value zero (0) in the "parameter 1", in that case, we should ignore the field and only do what "parameter 2" ("Component ID") tells us.
This will help with the "Component ID" changes where repeating `target_system` in "parameter 1" will not be required.

message_definitions/v1.0/development.xml
```diff
@@ -350,6 +350,17 @@
...
       </entry>
+      <entry value="610" name="MAV_CMD_DO_SET_SYS_CMP_ID" hasLocation="false" isDestination="false">
+        <description>
+          Set system and component id.
+          This allows moving of a system and all its components to a new system id, or moving a particular component to a new system/component id.
+          Recipients must reject command addressed to broadcast system ID.
+        </description>
+        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID that should be used by the device. Zero (0) is ignored, System ID should not changed.</param>
+        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID that should be used by the device. Zero (0) can be used to ignore this field.</param>
+        <param index="3" label="Reboot">Do a component reboot after ID change. Any value different than zero (0) will result in a vehicle reboot.</param>
```
hamishwillee, line 361: A few minor changes ...
```suggestion
        <param index="1" label="System ID" minValue="1" maxValue="255" increment="1">New system ID for target component(s). 0: ignore and reject command (broadcast system ID not allowed).</param>
        <param index="2" label="Component ID" minValue="0" maxValue="255" increment="1">New component ID for target component(s). 0: ignore (component IDs don't change).</param>
        <param index="3" label="Reboot">Reboot components after ID change. Any non-zero value triggers the reboot.</param>
```

hamishwillee (approved):
@patrickelectric Sorry for the delay - I only work Weds/Thurs.

This looks great. I modified it a little for English and consistency with other commands. Merging now as pre-approved by the dev call.

When you have a working implementation that shows this does what you need then we can review again and move into common.xml.

Thanks!

PR #2083, opened 2024-02-21
common.xml: Add "Stream ID" param to MAV_CMD_IMAGE_START_CAPTURE

nexton-winjeel:
As noted in [this comment](https://github.com/mavlink/mavlink/pull/2079#issuecomment-1955724877):

> With the current camera protocol, there is no way to specify which stream to take a snapshot from without also enabling output from that stream.

This change will allow a camera to take a snapshot from one stream while another stream is being output.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2135,7 +2135,7 @@
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Total Images" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
         <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1), otherwise set to 0. Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted.</param>
-        <param index="5" reserved="true"/>
+        <param index="5" label="Stream ID" minValue="0" increment="1">Video Stream to capture (0 for current active stream)</param>
         <param index="6" reserved="true"/>
         <param index="7" reserved="true" default="NaN"/>
       </entry>


message_definitions/v1.0/common.xml
```diff
@@ -2135,7 +2135,7 @@
         <param index="2" label="Interval" units="s" minValue="0">Desired elapsed time between two consecutive pictures (in seconds). Minimum values depend on hardware (typically greater than 2 seconds).</param>
         <param index="3" label="Total Images" minValue="0" increment="1">Total number of images to capture. 0 to capture forever/until MAV_CMD_IMAGE_STOP_CAPTURE.</param>
         <param index="4" label="Sequence Number" minValue="1" increment="1">Capture sequence number starting from 1. This is only valid for single-capture (param3 == 1), otherwise set to 0. Increment the capture ID for each capture command to prevent double captures when a command is re-transmitted.</param>
-        <param index="5" reserved="true"/>
+        <param index="5" label="Stream ID" minValue="0" increment="1">Video Stream to capture (0 for current active stream)</param>
```
hamishwillee, line 2138: @nexton-winjeel I don't think the discussion has fully resolved.
1. Should/could we add a channel concept now, and use channel id here https://github.com/mavlink/mavlink/pull/2079#issuecomment-1958432641
2. We are not actually capturing the video stream, we're capturing the source channel of the video stream. The difference is that to me this implies we're capturing an image with all the properties of the stream, such as resolution, but the camera might have a different setting for image capture. 
3. What if there are multiple active streams? I don't _think_ this is prohibited.

nexton-winjeel: Yeah, this really depends on what happens in #2079. 

> 1. Should/could we add a channel concept now, and use channel id here https://github.com/mavlink/mavlink/pull/2079#issuecomment-1958432641

If we're going to move to a channel concept, then we need to flesh that design out before we modify this command.

I'll detail it more in #2079, but conceptually, we probably need to capture:
* sensors, i.e. the actual image capturing component;
* channels, i.e. the conversion of raw sensor imagery into an encoded video;
* outputs, i.e. the transport stream sending out the encoded video

> 2. We are not actually capturing the video stream, we're capturing the source channel of the video stream. The difference is that to me this implies we're capturing an image with all the properties of the stream, such as resolution, but the camera might have a different setting for image capture.

Not necessarily. We have a payload that can capture a snapshot of either:
* the raw sensor image; or
* the streamed image.

But we can't currently make this distinction using only MAVLink commands. 

What I would probably do is make `param 6` an enum, e.g. `0: Capture output, 1: Capture channel, 2: Capture sensor`, and `param 5` can be the relevant ID.

> 3. What if there are multiple active streams? I don't think this is prohibited.

Not prohibited, but highly camera dependent. As above, we have a payload with multiple sensors, multiple channels and multiple outputs:
* It can capture raw sensor snapshots at any time.
* It _can't_ capture channel snapshots directly.
* It can capture output snapshots _only if_ that output is actively streaming.

For this PR, I'm trying to make `param 5` backward compatible. My assumption was that most cameras would only snapshot the active stream(s). But maybe we need to distinguish between _active_ streams and _all_ streams?

hamishwillee: > My assumption was that most cameras would only snapshot the active stream(s).

I think there are probably quite a few missions flown with cameras with no streaming going on at all, taking photos using whatever their default settings are.

Currently in MAVLink the image capture command means, AFAIK, exactly that - capture using your current settings to your storage card - don't care about the stream. We're now talking about the case where the camera perhaps has IR and Visible light sensors - again, whatever is streaming, the command is capture raw sensor snapshot.

The lens thing feels orthogonal.

nexton-winjeel: > I think there are probably quite a few missions flown with cameras with no streaming going on at all, taking photos using whatever their default settings are.

Yes. The reason I am using the term "stream" is because, right now, the only way to indicate that a camera has multiple sensors is by publishing multiple streams.

You can add an extra field to `CAMERA_INFORMATION` that specifies the number of sensors. But then how do you specify the sensor size, resolution and focal length of each sensor?

> Currently in MAVLink the image capture command means, AFAIK, exactly that - capture using your current settings to your storage card - don't care about the stream. We're now talking about the case where the camera perhaps has IR and Visible light sensors - again, whatever is streaming, the command is capture raw sensor snapshot.

I would argue that right now it is undefined, and therefore implementation defined. In terms of multi-camera payloads, I have seen all three implementations:
1. Snapshot captures raw sensor;
2. Snapshot captures processed video; and
3. Snapshot can select between raw and processed imagery.

hamishwillee: Hmmm. Fair points. Another question/confirmation - say we define video streams for the lens' and then get a command to stream the video. That's well and good if we really can stream the video, but what if we have a camera that doesn't allow video streaming and we still want to publish video streams for the purpose of setting the camera sensor?

nexton-winjeel: Yeah, my solution does assume that you can also stream video. I would say the solution is either:
1. Return `MAV_RESULT_UNSUPPORTED` for the streaming commands
    * Hopefully drivers this apply to would also have the [`CAMERA_CAP_FLAGS_HAS_VIDEO_STREAM`](https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS_HAS_VIDEO_STREAM) capabilty bit un-set.
 2. Rather than publishing multiple streams, publish each sensor as a separate Camera.

nexton-winjeel: I think we should close this and just use the solution in #2079 (although it should probably be made more generic, e.g. rename to `SET_IMAGE_CAPTURE_SOURCE`).

hamishwillee:
Agreed (by me and in the dev call). Thank you for all your thoughtful discussion of this.

PR #2087, opened 2024-02-29
common: Add webrtc option in video stream

patrickelectric:
Modern vehicles can provide a WebRTC video stream though a webpage that can be accessible via GCS software.
This change allow the GCS to identify and provide a way for the user to access it.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7394,7 +7394,7 @@
       <field type="uint16_t" name="rotation" units="deg">Video image rotation clockwise.</field>
       <field type="uint16_t" name="hfov" units="deg">Horizontal Field of view.</field>
       <field type="char[32]" name="name">Stream name.</field>
-      <field type="char[160]" name="uri">Video stream URI (TCP or RTSP URI ground station should connect to) or port number (UDP port ground station should listen to).</field>
+      <field type="char[160]" name="uri">Video stream URI (TCP, RTSP, HTTP URI that the ground station should connect to) or port number (UDP port ground station should listen to).</field>
       <extensions/>
       <field type="uint8_t" name="encoding" enum="VIDEO_STREAM_ENCODING">Encoding of stream.</field>
       <field type="uint8_t" name="camera_device_id" default="0" minValue="0" maxValue="6">Camera id of a non-MAVLink camera attached to an autopilot (1-6).  0 if the component is a MAVLink camera (with its own component id).</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3976,7 +3976,7 @@
       <entry value="4" name="VIDEO_STREAM_TYPE_WHEP">
         <description>Stream is WHEP (WebRTC-HTTP Egress Protocol)</description>
       </entry>
-      <entry value="4" name="VIDEO_STREAM_TYPE_WEBRTC">
+      <entry value="5" name="VIDEO_STREAM_TYPE_WEBRTC">
         <description>Stream is WebRTC</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3976,6 +3976,9 @@
       <entry value="4" name="VIDEO_STREAM_TYPE_WHEP">
         <description>Stream is WHEP (WebRTC-HTTP Egress Protocol)</description>
       </entry>
+      <entry value="4" name="VIDEO_STREAM_TYPE_WEBRTC">
+        <description>Stream is WebRTC</description>
+      </entry>
     </enum>
     <enum name="VIDEO_STREAM_ENCODING">
       <description>Video stream encodings</description>


patrickelectric:
Relevant for https://github.com/mavlink/mavlink-camera-manager @joaoantoniocardoso

hamishwillee:
From the little I know about WebRTC there's a signaling server somewhere that magically connects two endpoints so that they can talk directly. Once they have exchanged enough data about supported codecs and transports to set up a peer connection, they can stream video and audio data at each other via that connection in the form of media tracks. 

There's an example on MDN where this gets talked about https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Signaling_and_video_calling

So what exactly in this architecture is the URL that is being supplied? Is anything else needed by the recipient or sender of this URL to use it? My concern is that knowing that this is an RTC url and having the URL might not be enough for anyone to understand how they go about using the URL. So I'm after context. 

Hope that makes sense.

patrickelectric:
> From the little I know about WebRTC there's a signaling server somewhere that magically connects two endpoints so that they can talk directly. Once they have exchanged enough data about supported codecs and transports to set up a peer connection, they can stream video and audio data at each other via that connection in the form of media tracks.
> 
> There's an example on MDN where this gets talked about https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Signaling_and_video_calling
> 
> So what exactly in this architecture is the URL that is being supplied? Is anything else needed by the recipient or sender of this URL to use it? My concern is that knowing that this is an RTC url and having the URL might not be enough for anyone to understand how they go about using the URL. So I'm after context.
> 
> Hope that makes sense.

Hi @hamishwillee, that's 100% correct.

In the future we would like to expand that in a better way related to the signaling. 
But for now, we are planning for the mavlink-camera-manager to provide a webpage with the video source, where it can be rendered by the GCS with an iframe or webengine. 
In the end will be a webpage provided by the camera manager itself that already encapsulates all the signaling logic.
We could change this from `WebRTC` to `WebSite` ? Not 100% sure.

hamishwillee:
Thank you. IMO I think that Website with a description that the URL points that the website that can be embedded as an iframe (possibly updated via a WebRTC steam) would be better. My reasoning is that 
- if you use WebRTC now there are too many questions like mine above
- This would make it harder to define a more "pure" WebRTC in future
- In fact you don't care how the page gets its video stream, so we shouldn't mandate WebRTC.

message_definitions/v1.0/common.xml
```diff
@@ -3976,6 +3976,9 @@
       <entry value="4" name="VIDEO_STREAM_TYPE_WHEP">
         <description>Stream is WHEP (WebRTC-HTTP Egress Protocol)</description>
       </entry>
+      <entry value="4" name="VIDEO_STREAM_TYPE_WEBRTC">
```
```suggestion
      <entry value="5" name="VIDEO_STREAM_TYPE_WEBRTC">
```

hamishwillee:
@patrickelectric Is this still of interest? I'm still not keen about this but if you'd like it to go to the MAV call I can see what others think. If not, then I will close it next time I'm reviewing PRs and I see this note.

patrickelectric:
@joaoantoniocardoso, do we want this ? Is this something that we'll need in the future ?

joaoantoniocardoso:
I believe the idea of HTTP/webpage can be generally interesting, especially for browser-based GCS: they'd get an implementation-free URL that plays the video, abstracting all media handling.

The bad part is that, because of this abstraction, it seems to me that recording on topside becomes more difficult, so the overall value seems limited.

While MCM can potentially benefit users with this, I'd advocate adding something like `VIDEO_STREAM_TYPE_ZENOH` instead -- h264/h265 via Zenoh, which MCM already offers but has no way to advertise/offer it via MAVLink.

Thanks

hamishwillee:
I feel uncomfortable mixing the transport with the encoding. Do we need a separate field for that?

joaoantoniocardoso:
> I feel uncomfortable mixing the transport with the encoding. Do we need a separate field for that?

I'm not sure if I understood.

There is the `VIDEO_STREAM_ENCODING` (h264/h265) available already, but there's no way to tell the clients to subscribe to a given zenoh topic.

I'm proposing a variant of `VIDEO_STREAM_TYPE` named `VIDEO_STREAM_TYPE_ZENOH` for that.

Thanks

hamishwillee:
@joaoantoniocardoso Apologies. You didn't understand because my comment because it didn't make much sense. 

The criteria I try to aim for with anything in MAVLink is that :
- an addition is useful and usable to a consumer.
  So, if a consumer were to get an encoding of `VIDEO_STREAM_TYPE_ZENOH`, with everything else provided by MAVLink, would a GCS be able to decode the stream _without any prior knowledge of a particular setup_? What I don't want is for a GCS implementer to have to talk to BlueROV to get the video feed. For example , I imagine you still need a URI - is the format of that URI standardized and predictable, are there any other assumptions. and so on. You will know far more than me on this.
- it doesn't mix concerns - i.e. we don't mix encoding with transport

This is why I was open to WHEP - as I understand it the protocols to get the video are all well defined and supported by the info in MAVLink. It is why I was opposed to WebRTC - too many user-defined things that create vendor lock in.

Does that make sense?

PR #2088, opened 2024-03-03
Common: add MAV_CMD_DO_RETURN_PATH_START

IamPete1:
This adds support for a new "land rejoin" waypoint. This acts as a marker to the autopilot that it can rejoin a mission at any point after this marker and before a landing or MAV_CMD_DO_LAND_START.

AP demo: https://youtu.be/SoHXRTD0vEs?si=pHTvRDMcOsStv4l6

AP PR: https://github.com/ArduPilot/ardupilot/pull/26383

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1528,6 +1528,8 @@
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
       <entry value="188" name="MAV_CMD_DO_RETURN_PATH_START" hasLocation="true" isDestination="false">
+        <wip/>
+        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
         <description>Mission item to specify the start of a failsafe/landing return-path segment (the end of the segment is the next MAV_CMD_DO_LAND_START item).
           A vehicle that is using missions for landing (e.g. in a return mode) will join the mission on the closest path of the return-path segment (instead of MAV_CMD_DO_LAND_START or the nearest waypoint).
           The main use case is to minimize the failsafe flight path in corridor missions, where the inbound/outbound paths are constrained (by geofences) to the same particular path.
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1527,6 +1527,25 @@
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
+      <entry value="188" name="MAV_CMD_DO_RETURN_PATH_START" hasLocation="true" isDestination="false">
+        <description>Mission item to specify the start of a failsafe/landing return-path segment (the end of the segment is the next MAV_CMD_DO_LAND_START item).
+          A vehicle that is using missions for landing (e.g. in a return mode) will join the mission on the closest path of the return-path segment (instead of MAV_CMD_DO_LAND_START or the nearest waypoint).
+          The main use case is to minimize the failsafe flight path in corridor missions, where the inbound/outbound paths are constrained (by geofences) to the same particular path.
+          The MAV_CMD_NAV_RETURN_PATH_START would be placed at the start of the return path.
+          If a failsafe occurs on the outbound path the vehicle will move to the nearest point on the return path (which is parallel for this kind of mission), effectively turning round and following the shortest path to landing.
+          If a failsafe occurs on the inbound path the vehicle is already on the return segment and will continue to landing.
+          The Latitude/Longitude/Altitude are optional, and may be set to 0 if not needed.
+          If specified, the item defines the waypoint at which the return segment starts.
+          If sent using as a command, the vehicle will perform a mission landing (using the land segment if defined) or reject the command if mission landings are not supported, or no mission landing is defined. When used as a command any position information in the command is ignored.
+        </description>
+        <param index="1">Empty</param>
+        <param index="2">Empty</param>
+        <param index="3">Empty</param>
+        <param index="4">Empty</param>
+        <param index="5" label="Latitude">Latitudee. 0: not used.</param>
+        <param index="6" label="Longitude">Longitudee. 0: not used.</param>
+        <param index="7" label="Altitude" units="m">Altitudee. 0: not used.</param>
+      </entry>
       <entry value="189" name="MAV_CMD_DO_LAND_START" hasLocation="true" isDestination="false">
         <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot where a sequence of mission items that represents a landing starts.
 	  It may also be sent via a COMMAND_LONG to trigger a landing, in which case the nearest (geographically) landing sequence in the mission will be used.


message_definitions/v1.0/common.xml
```diff
@@ -1528,6 +1528,20 @@
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
+      <entry value="188" name="MAV_CMD_DO_LAND_REJOIN" hasLocation="true" isDestination="false">
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot that it can rejoin a mission at any point after this marker and before a landing or MAV_CMD_DO_LAND_START.
+          It may also be sent via a COMMAND_LONG to trigger a landing.
+          The autopilot will then rejoin the mission at the closet point following it to the landing.
+          The Latitude/Longitude/Altitude is optional, and may be set to 0 if not needed.
```
hamishwillee, line 1535: 1. How are they used if they are specified? In similar cases these would act as a waypoint as well as their other function. So if you were just flying a mission this would look like a waypoint
2. What would happen if these were sent and the MAV_CMD was used in a command?

message_definitions/v1.0/common.xml
```diff
@@ -1528,6 +1528,20 @@
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
+      <entry value="188" name="MAV_CMD_DO_LAND_REJOIN" hasLocation="true" isDestination="false">
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot that it can rejoin a mission at any point after this marker and before a landing or MAV_CMD_DO_LAND_START.
+          It may also be sent via a COMMAND_LONG to trigger a landing.
```
hamishwillee, line 1533: Why COMMAND_LONG? Generally we're recommending COMMAND_INT by default nowdays because you can package almost anything in a command int and get good precision on lat/lon, but that is not true if you try package in a command_lon (i.e. far more commands will work at their best in a COMMAND_INT if implemented in the autopilot).

message_definitions/v1.0/common.xml
```diff
@@ -1528,6 +1528,20 @@
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
+      <entry value="188" name="MAV_CMD_DO_LAND_REJOIN" hasLocation="true" isDestination="false">
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot that it can rejoin a mission at any point after this marker and before a landing or MAV_CMD_DO_LAND_START.
```
hamishwillee, line 1532: This will need more context - mostly because it is hard from this text to understand how this is different from other land commands.

My understanding from the video is that the prime use case is that you want to land the vehicle following some part of the defined mission - perhaps following either some kind of land command, or switching into a mission return mode.
So you specify this `MAV_CMD_DO_LAND_REJOIN` in the mission. If you specify it with position this is the waypoint at which it applies. If you specify it without position then it applies on the previous waypoint. Either way, the section of the mission between the rejoin and the final landing defines the parts of a mission that you are happy for the vehicle to rejoin the mission.

Then imagine you're flying around, you get low on batteries and invoke the mission return mode, the vehicle will fly towards whatever waypoint it thinks is most optimal (or closes?) in that section of the mission to rejoin it.
Is that correct?

What's the benefit over flying to the closest waypoint and fast forwarding, or flying to the landing waypoint? Is it that you perhaps know that anywhere in the particular identified part of the mission it is safe to rejoin the mission, but perhaps it is not safe to do so by flying to the landing waypoint or closest waypoint?

- What happens if there are several of these rejoins in the mission?
- Would it made sense to define rejoin zones rather than "from here to the end"?


What happens if I send this in in a command protocol? Is it exactly the same as sending a DO_LAND. I suspect not, so where in this case does the mission get rejoined?

message_definitions/v1.0/common.xml
IamPete1: Right, if its in a mission it would just behave like a normal way point. If sent as a `MAV_CMD` they are not used.

I'm not particularity attached to this behavior, its just a copy across from `MAV_CMD_DO_LAND_START`.

message_definitions/v1.0/common.xml
IamPete1: Again, just a copy from `MAV_CMD_DO_LAND_START`. Happy to change it. AP accepts both.

message_definitions/v1.0/common.xml
IamPete1: > Then imagine you're flying around, you get low on batteries and invoke the mission return mode, the vehicle will fly towards whatever waypoint it thinks is most optimal (or closes?) in that section of the mission to rejoin it.
> Is that correct?

Correct.

In AP it currently just flys to waypoint at the end of closest mission leg. If there are several rejoin legs it will pick the closest. I do have a branch where it picks the leg that will result in the closest landing. For example the closest leg might be a long leg to a landing point which is further away. It might be better to fly a little further up front and rejoin a shorter mission leg. In any-case, that is not implemented in the current PR, In that case I had a parameter option rather than encoding that into the mission somehow. 


> What's the benefit over flying to the closest waypoint and fast forwarding, or flying to the landing waypoint? Is it that you perhaps know that anywhere in the particular identified part of the mission it is safe to rejoin the mission, but perhaps it is not safe to do so by flying to the landing waypoint or closest waypoint?

Flying to the closest waypoint would mean that half the time you backtrack to get to that waypoint before continuing on towards the landing. Flying directly to the landing is what `MAV_CMD_DO_LAND_START` does, you can get the same "fly to closest waypoint" behavior  by having do land start markers after each waypoint on the rejoin leg. But again you sometimes end up backtracking. If you just have one do `MAV_CMD_DO_LAND_START` before the landing then you fly there in a straight line. 



> * Would it made sense to define rejoin zones rather than "from here to the end"?

We sort of do have that, the rejoin zone starts at this new waypoint and ends at a landing or a `MAV_CMD_DO_LAND_START`. So you can have multiple zones in one mission.



> What happens if I send this in in a command protocol? Is it exactly the same as sending a DO_LAND. I suspect not, so where in this case does the mission get rejoined?

It will look for the `MAV_CMD_DO_LAND_REJOIN` in the mission and rejoin. If there is no `MAV_CMD_DO_LAND_REJOIN` then you will get a `MAV_RESULT_FAILED`. In AP this is the same behavior as if you were to switch into the Auto-RTL mode (except that Auto-RTL mode will look for a `MAV_CMD_DO_LAND_START` to jump to if there is no rejoin).

message_definitions/v1.0/common.xml
IamPete1: I have changed to mention `COMMAND_INT` rather then `COMMAND_LONG` and updated the wording a little.

message_definitions/v1.0/common.xml
```diff
@@ -1528,6 +1528,19 @@
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
+      <entry value="188" name="MAV_CMD_DO_LAND_REJOIN" hasLocation="true" isDestination="false">
```
hamishwillee, line 1531: Maybe? Because you're joining, not rejoin.

```suggestion
      <entry value="188" name="MAV_CMD_DO_LAND_JOIN" hasLocation="true" isDestination="false">
```

message_definitions/v1.0/common.xml
```diff
@@ -1528,6 +1528,19 @@
...
+        <description>Mission command to perform a landing. This is used as a marker in a mission to tell the autopilot that it can rejoin a mission at any point after this marker and before a landing or MAV_CMD_DO_LAND_START.
+          It may also be sent via a COMMAND_INT, the autopilot will then rejoin the mission at the closet point after a MAV_CMD_DO_LAND_REJOIN mission item.
+          The Latitude/Longitude/Altitude is optional, and may be set to 0 if not needed.
```
hamishwillee, line 1534:
```suggestion
        <description>
          Mission item to specify the start of a land-rejoin segment (the end of the segment is set by the next following MAV_CMD_LAND_REJOIN command).
          A vehicle that is using missions for landing (e.g. in a return mode) can join the mission at any point within any land-rejoin segment.
          This allows more flexible definition of the landing path than joining the mission path at either the closest waypoint (which might require backtracking) or the MAV_CMD_DO_LAND_START point.
          The Latitude/Longitude/Altitude are optional, and may be set to 0 if not needed.
          If specified the item defines the waypoint at which the join start.
          If not specified, the land-rejoin segment starts on the previous waypoint.
          If sent using as a command, the vehicle will perform a mission landing (using the land segment if defined) or reject the command if mission landings are not supported, or no mission landing is defined. When used as a command any position information in the command is ignored.          
```

message_definitions/v1.0/common.xml
```diff
@@ -1528,6 +1528,19 @@
...
+          It may also be sent via a COMMAND_INT, the autopilot will then rejoin the mission at the closet point after a MAV_CMD_DO_LAND_REJOIN mission item.
+          The Latitude/Longitude/Altitude is optional, and may be set to 0 if not needed. 
+        </description>
+        <param index="1">Empty</param>
+        <param index="2">Empty</param>
+        <param index="3">Empty</param>
+        <param index="4">Empty</param>
+        <param index="5" label="Latitude">Latitude</param>
+        <param index="6" label="Longitude">Longitude</param>
+        <param index="7" label="Altitude" units="m">Altitude</param>
```
hamishwillee, line 1542: Always a bit weird because 0 is a valid lat/lon
```suggestion
        <param index="5" label="Latitude">Latitude. 0: not used.</param>
        <param index="6" label="Longitude">Longitude. 0: not used.</param>
        <param index="7" label="Altitude" units="m">Altitude. 0: not used.</param>
```

message_definitions/v1.0/common.xml
hamishwillee: OK. I've proposed version in https://github.com/mavlink/mavlink/pull/2088/files#r1522423206

I still can't visualise the benefit. I mean I can see that you don't want to backtrack, so going to closest waypoint behind you doesn't work, and further than you might not want to go straight to the nearest land waypoint behind you. But that isn't stopped by this - if it picks the closest segment, that is just as likely in both cases to be the one behind you anyway.

But perhaps just haven't flown enough. I'll raise this in the dev call to discuss.

hamishwillee:
FWIW I am still not sure about this. There's concern about back tracking, but that's sorted out by allowing your vehicle free reign to decide how it wants to land, not having these segments.

Anyway, we'll need to see what comments fall out of your ArduPilot PR before we progress this. In the meantime I've added comments to improve the desciption.

IamPete1:
I'm going to have another try at explaining why this is needed. 

This is a typical example mission where you would want this. It is a "out and back" for a delivery. There is a airspace restriction, the vehicle must stay within the fence. In such cases breach of fence generally requires flight termination with parachute release.

image

For the sake of example the vehicle got to this point following the normal mission, then for some reason the mission has to be aborted, the vehicle must return and land.

image

I could RTL, but then I breach the airspace restrictions:
image

I could use do land start, but I have to add a do land start waypoint after every point on the return leg, this adds half again to my number of waypoints.

image

And do land start results in flying to the end waypoint first before backtacking because we were more than half way along that mission leg, This means that my abort does not move the vehicle towards safer state where it is closer to landing.

image

We could enable pathplanning and use fence path planning within auto. I would setup the the mission like this with one do-land-start before the landing:
image

That gets us this:
image

We we make it back via a short route and don't breach the fence. But we do come close to the fence, this could be fudged with the margin parameter. We also ignore any altitude change along that path because the planning is just linear in altitude. We also don't get nice S-Curve navigation.

The final way is with smart RTL mode. That gets us this:
image

That seems good, however there are a number of issues. Because the vehicle is following breadcrumbs it has dropped we don't know the path it will will take ahead of time. This also means position errors on the output bound leg are compounded on the return leg.

Rejoin is better than all the other methods because:

- The path is known before takeoff. (SRTL can't do this)
- Starts returning immediately (do land start can't do this).
- We follow the normal mission points, so we can do change of altitude on the return (SRLT and path planning can't do this)
- You can have condition yaw to line up, or do change speeds to slow down in the landing phase (SRTL cant do this).

hamishwillee:
@IamPete1 Thanks - appreciate you taking the time to educate me. I'll have a look at this next week when I'm back. FYI, I would still wait for the PR in ArduPilot to be broadly accepted - for stuff to be in common it needs to be accepted in at least one flight stack and accepted that another one would consider doing so.

hamishwillee:
Hmmm, so I can see that you can't specify a safe return mode "fly a reverse mission back from wherever you are to home" because if you were near the end you'd have to go all the way to the beginning. Similarly, you can't specify "fast forward from where you are to end" because that might take you forward unnecessarily if you're in the "heading out" bit. 
You could run some path planning that calculated both options and took the shortest option of those two? I guess this doesn't work if your takeoff and land position need to be slightly different for some reason.

But anyhow, so you want to cut across from wherever you are to the closest path on the return journey, without cutting across anywhere you're not allowed to go. Right?

So where do you put your land rejoins to make this happen? I mean if you put it on 1 and your problem happens before 8 then the vehicle can choose to rejoin wherever it likes. But why wouldn't it choose to fly across direct to the end, like your RTL example? 
If you put that rejoin at 8 then similarly if the error happens early you'd rejoin further away than you want.

So don't you have to put these land rejoins everywhere. I'm sorry I'm not getting it.

hamishwillee:
@IamPete1 OK, so @peterbarker walked me through this. I think I get it now. The thing I was finding hard to grasp was understanding the specific use cases where this is useful and understanding where you set the rejoin.

I'm going to summarise here to get it clear in my mind. This will duplicate a lot of what you have already said.

1. The main use case for this is where you're following a constrained path (geofence bounded) along a pipeline, returning through that same pipeline. The waypoints on both paths would be very close or overlaying
2. If a failsafe occurs in your mission, you want to return to landing along the shortest return path. If you were on the outbound run this would be to cut to the adjacent return path. If you were on the inbound run, it would be just to return along your current path.
3. You could do this by adding multiple DO_LAND_START along the return path. 
   - If you had a failsafe in the outbound path, you'd head to the nearest DO_LAND_START. Since that is on the return path, and you have corresponding waypoints next to you on the inbound and outbound paths, you'll essentially just start heading back the way you came. 
   - If you had a failsafe on the return path you'd just follow the trail of DO_LAND_START ahead of you.
   - But it's a lot of work to add these at every return point.
 4. The REJOIN "somewhat" emulates this behaviour by specifying that on failsafe the vehicle can join the path anywhere between the rejoin and the next DO_LAND_START.
    So considering the previous example you put your REJOIN at the end of the outbound path/start of the return path.
    - On outbound failsafe the vehicle moves to the return path that is closest to it and continues along that.
    - On inbound failsafe it just continues on the path.
 
With that understanding I can see why this would be useful for this use case. It would be easy to screw up in some cases - in particular outside of this corridor design, but anywhere that your inbound and outbound paths might be separated by a geofence is a problem.

The name is not great. Peter suggested  `MAV_CMD_NAV_RETURN_PATH_START` which I prefer. Thoughts?

message_definitions/v1.0/common.xml
hamishwillee: OK, so here is updated attempt to make this clear

```suggestion
        <description>
          Mission item to specify the start of a failsafe/landing return-path segment (the end of the segment is the next MAV_CMD_DO_LAND_START item).
          A vehicle that is using missions for landing (e.g. in a return mode) will join the mission on the closest path of the return-path segment (instead of MAV_CMD_DO_LAND_START or the nearest waypoint).
          
          The main use case is to minimize the failsafe flight path in corridor missions, where the inbound/outbound paths are constrained (by geofences) to a particular path.
          The MAV_CMD_NAV_RETURN_PATH_START would be placed at the start of the return path.
          If a failsafe occurs on the outbound path the vehicle will move to the nearest point on the return path (which is parallel for this kind of mission), effectively turning round and following the shortest path to landing.
          If a failsafe occurs on the inbound path the vehicle is already on the return segment and will continue to landing.
          
          The Latitude/Longitude/Altitude are optional, and may be set to 0 if not needed.
          If specified, the item defines the waypoint at which the return segment starts.
          If sent using as a command, the vehicle will perform a mission landing (using the land segment if defined) or reject the command if mission landings are not supported, or no mission landing is defined. When used as a command any position information in the command is ignored.
```

sfuhrer:
> MAV_CMD_NAV_RETURN_PATH_START

That name sounds also better to me, it's more descriptive.

So in the "narrow out-/inbound corridor example you would set it to the waypoint at the start of the return leg, have I got that correct?

hamishwillee:
> So in the "narrow out-/inbound corridor example you would set it to the waypoint at the start of the return leg, have I got that correct?

@sfuhrer Yes!

hamishwillee:
@IamPete1 Next step would be for you to look at the suggestions above and ideally merge them. I get it now and I think those updates capture it. Once this is broadly accepted that it will be merged into ArduPilot we can discuss merging. We can do so earlier if you want to move it into development.xml.

message_definitions/v1.0/common.xml
IamPete1: I think it should still be a "DO" because the concept is some similar to "DO_LAND_START".

message_definitions/v1.0/common.xml
IamPete1: Done.

message_definitions/v1.0/common.xml
IamPete1: Done

message_definitions/v1.0/common.xml
IamPete1: Renamed to `MAV_CMD_DO_RETURN_PATH_START`

IamPete1:
> @IamPete1 Next step would be for you to look at the suggestions above and ideally merge them. I get it now and I think those updates capture it. Once this is broadly accepted that it will be merged into ArduPilot we can discuss merging. We can do so earlier if you want to move it into development.xml.

Thanks for the reminder, I have updated using the suggestions except the name which is "DO" rather than "NAV". 

I took this to the AP dev call a few weeks ago and got some supportive comments that we would mege it. (https://github.com/ArduPilot/ardupilot/pull/26383#issuecomment-2005372885 ect) From my end it needs to merge here, then I can open a PR that ports the MAVLink change to AP's MAVlink branch once that is merged we can merge it into AP.

message_definitions/v1.0/common.xml
hamishwillee: From https://mavlink.io/en/guide/define_xml_element.html#command_names

> - MAV_CMD_NAV_: NAV commands are used for navigation/movement commands (commands to go to a particular waypoint or move in a particular way).
> - MAV_CMD_DO_: DO commands are used for setting modes, changing altitude or speed etc.

I think on balance you're right. It can act as a waypoint, but it is not specifically a command to navigate somewhere.

hamishwillee (approved):
I think this makes sense.

@peterbarker @julianoes Note that this now named MAV_CMD_DO_RETURN_PATH_START, which makes it a DO command rather than a NAV command.

Any objections to merging?

peterbarker:
On Wed, 27 Mar 2024, Hamish Willee wrote:

> Any objections to merging?

Not from me.

julianoes (approved):

message_definitions/v1.0/common.xml
```diff
@@ -1527,6 +1527,25 @@
         <param index="6" label="Actuator 6" minValue="-1" maxValue="1">Actuator 6 value, scaled from [-1 to 1]. NaN to ignore.</param>
         <param index="7" label="Index" minValue="0" increment="1">Index of actuator set (i.e if set to 1, Actuator 1 becomes Actuator 7)</param>
       </entry>
+      <entry value="188" name="MAV_CMD_DO_RETURN_PATH_START" hasLocation="true" isDestination="false">
```
hamishwillee, line 1530: WIPing it - though really it should go in development.xml. @IamPete1 you can remove this once your PR is accepted into ArduPilot

```suggestion
      <entry value="188" name="MAV_CMD_DO_RETURN_PATH_START" hasLocation="true" isDestination="false">
        <wip/>
        <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
```

hamishwillee:
THanks all. Merged as WIP.

PR #2091, opened 2024-03-07
DO_FENCE_ENABLE when sent as a command

hamishwillee:
The MAV_CMD_DO_FENCE_ENABLE states that it is a mission item, but it can, and is sent as a command. This attempts to clarify the expected behaviour in this case. 

The points of interest being that:
- mission settings persist in the mission context, even on flight stacks that pause by switching mode to "Hold": on return to the mission the mission setting would be used.
- As a command the value only persists until mode change or reboot.
- A command can affect a mission

Originates from discussion with @andyp1per in https://github.com/ArduPilot/mavlink/pull/349#discussion_r1482155395

@sfuhrer @peterbarker Does this makes sense? Is it what is done? That's my assumption on what you'd want to happen but I haven't tested.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1707,7 +1707,8 @@
           Enable the geofence.
           This can be used in a mission or via the command protocol.
           The persistence/lifetime of the setting is undefined.
-          Depending on flight stack implementation it may persist until superseded (even across reboots), or it may revert to a system default at the end of a mission or on reboot.
+          Depending on flight stack implementation it may persist until superseded, or it may revert to a system default at the end of a mission.
+          Flight stacks typically reset the setting to system defaults on reboot.
 	</description>
         <param index="1" label="Enable" minValue="0" maxValue="2" increment="1">enable? (0=disable, 1=enable, 2=disable_floor_only)</param>
         <param index="2" label="Types" enum="FENCE_TYPE">Fence types to enable or disable as a bitmask. A value of 0 indicates that all fences should be enabled or disabled. This parameter is ignored if param 1 has the value 2</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1703,7 +1703,13 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
-        <description>Mission command to enable the geofence</description>
+        <description>
+          Enable the geofence.
+          This can be used in a mission or via the command protocol.
+          The effects persist while the mission is being executed, until superseded by another mission item or the mission completes.
+          When used as a command (outside of mission) the value persists until superseded by another command, a mode change (which reverts system back to system defaults), or reboot.
+          When used as a command while a mission is running the value is treated as though it were a mission item, and will change the current mission value.
+	</description>      
         <param index="1" label="Enable" minValue="0" maxValue="2" increment="1">enable? (0=disable, 1=enable, 2=disable_floor_only)</param>
         <param index="2" label="Types" enum="FENCE_TYPE">Fence types to enable or disable as a bitmask. A value of 0 indicates that all fences should be enabled or disabled. This parameter is ignored if param 1 has the value 2</param>
         <param index="3">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1708,7 +1708,7 @@
           This can be used in a mission or via the command protocol.
           The effects persist while the mission is being executed, until superseded by another mission item or the mission completes.
           When used as a command (outside of mission) the value persists until superseded by another command, a mode change (which reverts system back to system defaults), or reboot.
-          When used as a command while a mission is running the value is treated as though it were a mission item, and will change the current mission value.
+          When used as a command the fence will change to the set value (even while a mission is running).
 	</description>      
         <param index="1" label="Enable" minValue="0" maxValue="2" increment="1">enable? (0=disable, 1=enable, 2=disable_floor_only)</param>
         <param index="2" label="Types" enum="FENCE_TYPE">Fence types to enable or disable as a bitmask. A value of 0 indicates that all fences should be enabled or disabled. This parameter is ignored if param 1 has the value 2</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1706,10 +1706,9 @@
         <description>
           Enable the geofence.
           This can be used in a mission or via the command protocol.
-          The effects persist while the mission is being executed, until superseded by another mission item or the mission completes.
-          When used as a command (outside of mission) the value persists until superseded by another command, a mode change (which reverts system back to system defaults), or reboot.
-          When used as a command the fence will change to the set value (even while a mission is running).
-	</description>      
+          The persistence/lifetime of the setting is undefined.
+          Depending on flight stack implementation it may persist until superseded (even across reboots), or it may revert to a system default at the end of a mission or on reboot.
+	</description>
         <param index="1" label="Enable" minValue="0" maxValue="2" increment="1">enable? (0=disable, 1=enable, 2=disable_floor_only)</param>
         <param index="2" label="Types" enum="FENCE_TYPE">Fence types to enable or disable as a bitmask. A value of 0 indicates that all fences should be enabled or disabled. This parameter is ignored if param 1 has the value 2</param>
         <param index="3">Empty</param>


sfuhrer (approved):
Looks correct to me. The same way as we have implemented the SET_SPEED from what I see.

hamishwillee:
Thanks @sfuhrer !

@peterbarker Does this work for you too?

message_definitions/v1.0/common.xml
```diff
@@ -1703,7 +1703,13 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
-        <description>Mission command to enable the geofence</description>
+        <description>
+          Enable the geofence.
+          This can be used in a mission or via the command protocol.
+          The effects persist while the mission is being executed, until superseded by another mission item or the mission completes.
```
peterbarker, line 1709: Pretty sure that's not ArduPilot's behaviour; there may be bits and pieces which will affect it (automatic fence disabling and whatnot), but there's nothing necessarily magic  "mission has finished, undo fence stuff!" things going.

message_definitions/v1.0/common.xml
```diff
@@ -1703,7 +1703,13 @@
...
+          When used as a command (outside of mission) the value persists until superseded by another command, a mode change (which reverts system back to system defaults), or reboot.
```
peterbarker, line 1710: Mode changes don't affect ArduPilot fences unless you're doing really clever stuff.

message_definitions/v1.0/common.xml
```diff
@@ -1703,7 +1703,13 @@
...
+          When used as a command while a mission is running the value is treated as though it were a mission item, and will change the current mission value.
```
peterbarker, line 1711: Not sure what "mission value" means in this context.

message_definitions/v1.0/common.xml
hamishwillee: @peterbarker I thought the _general expectation_ was that "what happens in mission land, stays in mission land". So setting a speed, an ROI or whatever is mission-specific and gets undone at the end. 

But if not, would you be happy with:

```suggestion
          The effects persist while the mission is being executed or until superseded by another mission item (whether or not effects persist in other modes/after mission completion depends on the flight stack).
```

message_definitions/v1.0/common.xml
hamishwillee: What I think you're saying is that there is just one fence setting/parameter for geofence and that is used everywhere - right?

message_definitions/v1.0/common.xml
hamishwillee: Good point. I don't think it applies to argupilot. What this means is "if you send a command (and it is supported) and the mission is running the fence setting will change to whatever you set it with using the command - effectively this is the same though you had set the fence in the mission.

```suggestion
          When used as a command the fence will change to the set value (even while a mission is running).
```

message_definitions/v1.0/common.xml
peterbarker: > @peterbarker I thought the _general expectation_ was that "what happens in mission land, stays in mission land". So setting a speed, an ROI or whatever is mission-specific and gets undone at the end.

Yeah, pretty sure ROI won't be reset on ArduPilot either :-)  Actually doing what you're suggesting is problematic from an embedded-development POV in that you need to keep arbitrary amounts of old state around to revert to.  And if you re-enter mission mode then perhaps go back to whatever the mission was dictating... messy :-)

> But if not, would you be happy with:

Yep!

message_definitions/v1.0/common.xml
peterbarker: Yes, correct.  So e.g. if the mission enables the fence the user can send a command via mavlink in that disables it.  Or they can toggle a switch on the flight controller.  Or a script could enable/disable.  Or you could use a physical switch on the vehicle....

Last-one-in wins.

message_definitions/v1.0/common.xml
```diff
@@ -1703,7 +1703,13 @@
...
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
-        <description>Mission command to enable the geofence</description>
+        <description>
+          Enable the geofence.
+          This can be used in a mission or via the command protocol.
+          The effects persist while the mission is being executed, until superseded by another mission item or the mission completes.
+          When used as a command (outside of mission) the value persists until superseded by another command, a mode change (which reverts system back to system defaults), or reboot.
+          When used as a command the fence will change to the set value (even while a mission is running).
+	</description>
```
hamishwillee, line 1712: Thanks @peterbarker. Sadly all we can actually say is:

```suggestion
        <description>
          Enable the geofence.
          This can be used in a mission or via the command protocol.
          The persistence/lifetime of the setting is undefined.
          Depending on flight stack implementation it may persist until superseded (even across reboots), or it may revert to a system default at the end of a mission or on reboot.
	</description>
```

message_definitions/v1.0/common.xml
hamishwillee: MOdified - no point saying all this if it is not how things are.

message_definitions/v1.0/common.xml
```diff
@@ -1703,7 +1703,12 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="207" name="MAV_CMD_DO_FENCE_ENABLE" hasLocation="false" isDestination="false">
-        <description>Mission command to enable the geofence</description>
+        <description>
+          Enable the geofence.
+          This can be used in a mission or via the command protocol.
+          The persistence/lifetime of the setting is undefined.
+          Depending on flight stack implementation it may persist until superseded (even across reboots), or it may revert to a system default at the end of a mission or on reboot.
+	</description>
```
hamishwillee, line 1711: FYI @sfuhrer @peterbarker OK, complete rewrite. ArduPilot just stores the value in a param so it persists forever. I'm not sure whether it is better to standardize that behaviour or to make it "up to the flight stack". I have decided on the later here, but I'd like your opinion - I mean for end users it is kind of better to know that if they set this in a mission it will persist, than to know that this "might" be the case.

sfuhrer: I guess that's the best you can do, yes agree.

hamishwillee: >I don't think we would want to store it in a param on PX4, for me it makes more sense to reset on reboot.

@sfuhrer Thanks. I don't think it should be a param because I strongly disagree that mission settings should escape a mission. Missions should be self-contained, and on exiting a mission your vehicle should be in a known state.

If settings persist outside the mission though then I'm not sure how I feel about resetting on reboot. On one hand I like the idea that rebooting always returns the vehicle to a known state. But you could argue that if the user expects this to be persistent, then the known state is whatever they set last.

But my personal philosophy does not count - we don't update the spec in a way that forces a flight stack to be incompatible without broad agreement.

hamishwillee: @peterbarker Work for you now?

peterbarker:
On Wed, 20 Mar 2024, Hamish Willee wrote:

> > +        <description>
> +          Enable the geofence.
> +          This can be used in a mission or via the command protocol.
> +          The persistence/lifetime of the setting is undefined.
> +          Depending on flight stack implementation it may persist until superseded (even across reboots), or it may revert to a system default at the end of a mission or on reboot.
> +	</description>
> 
> @peterbarker Work for you now?

About as good as it's going to get ATM :-/

Note that ArduPilot has had some behaviour where non-parameter mavlink 
commands have changed persistent state on the vehicle where they 
shouldn't.  We are cleaning those up as time goes on.


I don't think ArduPilot persists any fence state based on these mavlink 
commands, 'though.... we initialise state from parameters but don't store 
the state back into those parameters.

hamishwillee:
> I don't think ArduPilot persists any fence state based on these mavlink commands, 'though.... we initialise state from parameters but don't store the state back into those parameters.

Oh, I understood from your comment that ArduPilot wrote to a parameter. So you reset on boot too?

peterbarker:
On Thu, 21 Mar 2024, Hamish Willee wrote:

>       I don't think ArduPilot persists any fence state based on these mavlink commands, 'though.... we initialise state from parameters but don't store the state back into those parameters.
> 
> Oh, I understood from your comment that ArduPilot wrote to a parameter. So you reset on boot too?

Net outcome, yes.

Not really a matter of "reset on boot" so much as "at boot the value is 
initialised from parameter" and that parameter's not changed by these 
commands.

message_definitions/v1.0/common.xml
hamishwillee:
```suggestion
        <description>
          Enable the geofence.
          This can be used in a mission or via the command protocol.
          The persistence/lifetime of the setting is undefined.
          Depending on flight stack implementation it may persist until superseded, or it may revert to a system default at the end of a mission.
          Flight stacks typically reset the setting to system defaults on reboot.
	</description>
```

hamishwillee:
Thanks @peterbarker - that's what I understood. I've updated to make it clear that "most flight stacks update on reboot". Going to merge now, as this is hand wavey, but more correct than before.

PR #2092, opened 2024-03-07
common.xml: deprecate the _INT frames

peterbarker:
the frame of an altitude is not dependent on how the altitude is being transported (how other fields in the same message are being transported).

ArduPilot has treated these frames to be the same as the non-_INT equivalents in most places for a very long time.

We can save some confusion and perhaps a small amount of flash space if we remove the _INT frames

---

HamishW Edit

This marks MAV_FRAME_GLOBAL_INT, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT as deprecated in favour of using their non-`_INT` alternatives, and simplifies the descriptions of scaling in the various messages.

The `_INT` variants were clearly created to allow specification of frames where the values of lat/lon are scaled, to go with the corresponding ones that don't make any statements about scaling. However these are used mainly in COMMAND_INT which explicitly specifies the scaling anyway - so the two values are synonyms (also true elsewhere these are used).

ArduPilot and MAVSDK, and presumably PX4 tend to implement code such that the INT versions are just synonyms. But sometimes one option or the other is forgotten, and this results in bugs. The idea here is to deprecate the pointless options, and eventually stop them being sent by GCS.

- [ ] On merge, update dev guide - see https://github.com/mavlink/mavlink/pull/2092#issuecomment-2008529832

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -260,8 +260,8 @@
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>
-        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</description>
+        <description>Global (WGS84) coordinate frame + MSL altitude (mean sea level). Position values are scaled by 1E7.</description>
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
         <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_RELATIVE_ALT"/>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5801,7 +5801,7 @@
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot). The rationale for the timestamp in the setpoint is to allow the system to compensate for the transport delay of the setpoint. This allows the system to compensate processing latency.</field>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL_INT = 5, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT = 6, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT = 11</field>
+      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL = 0, MAV_FRAME_GLOBAL_RELATIVE_ALT = 3, MAV_FRAME_GLOBAL_TERRAIN_ALT = 10 (MAV_FRAME_GLOBAL_INT, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT are allowed synonyms, but have been deprecated)</field>
       <field type="uint16_t" name="type_mask" enum="POSITION_TARGET_TYPEMASK" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle.</field>
       <field type="int32_t" name="lat_int" units="degE7">X Position in WGS84 frame</field>
       <field type="int32_t" name="lon_int" units="degE7">Y Position in WGS84 frame</field>
@@ -5818,7 +5818,7 @@
     <message id="87" name="POSITION_TARGET_GLOBAL_INT">
       <description>Reports the current commanded vehicle position, velocity, and acceleration as specified by the autopilot. This should match the commands sent in SET_POSITION_TARGET_GLOBAL_INT if the vehicle is being controlled this way.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot). The rationale for the timestamp in the setpoint is to allow the system to compensate for the transport delay of the setpoint. This allows the system to compensate processing latency.</field>
-      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL_INT = 5, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT = 6, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT = 11</field>
+      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL = 0, MAV_FRAME_GLOBAL_RELATIVE_ALT = 3, MAV_FRAME_GLOBAL_TERRAIN_ALT = 10 (MAV_FRAME_GLOBAL_INT, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT are allowed synonyms, but have been deprecated)</field>
       <field type="uint16_t" name="type_mask" enum="POSITION_TARGET_TYPEMASK" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle.</field>
       <field type="int32_t" name="lat_int" units="degE7">X Position in WGS84 frame</field>
       <field type="int32_t" name="lon_int" units="degE7">Y Position in WGS84 frame</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -259,14 +259,13 @@
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</description>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL">Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</deprecated>
         <description>Global (WGS84) coordinate frame + MSL altitude (mean sea level). Position values are scaled by 1E7.</description>
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
-        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_RELATIVE_ALT"/>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_RELATIVE_ALT">Use MAV_FRAME_GLOBAL_RELATIVE_ALT in COMMAND_INT (and elsewhere) as a synonymous replacement.</deprecated>
         <description>
           Global (WGS84) coordinate frame (scaled) + altitude relative to the home position.
-          First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude with 0 being at the altitude of the home position.
         </description>
       </entry>
       <entry value="7" name="MAV_FRAME_LOCAL_OFFSET_NED">
@@ -284,8 +283,8 @@
         <description>Global (WGS84) coordinate frame with AGL altitude (altitude at ground level).</description>
       </entry>
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
-        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_TERRAIN_ALT"/>
-        <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_TERRAIN_ALT">Use MAV_FRAME_GLOBAL_TERRAIN_ALT in COMMAND_INT (and elsewhere) as a synonymous replacement.</deprecated>
+        <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (altitude at ground level).</description>
       </entry>
       <entry value="12" name="MAV_FRAME_BODY_FRD">
         <description>FRD local frame aligned to the vehicle's attitude (x: Forward, y: Right, z: Down) with an origin that travels with vehicle.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -260,9 +260,11 @@
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>
         <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_RELATIVE_ALT"/>
         <description>
           Global (WGS84) coordinate frame (scaled) + altitude relative to the home position.
           First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude with 0 being at the altitude of the home position.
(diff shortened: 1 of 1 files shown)


peterbarker:
I've created a PR on the ArduPilot repository for discussion at our next DevCall.

peterbarker:
We discussed at our DevCall and we think this is a good change.  Just waiting for some discussion upstream.

hamishwillee:
The `_INT` cases explicitly specify that lat/lon are scaled 1E7, while the others imply they are not. PX4 handles the case to make the decision about the way to decode the values. 

I'll raise in the meeting this evening.

nexton-winjeel:
I wasn't in the ArduPilot dev call, but my assumption is the problem is:
* the `frame` is independent from the types used to transport the position fields in a message.
* the latitude/longitude scaling should be determined from the type of the field:
    * no scaling if the field is a `float`;
    * `1E7` scaling if the field is `int32_t`.

peterbarker:
On Tue, 12 Mar 2024, Hamish Willee wrote:

> The _INT cases explicitly specify that lat/lon are scaled 1E7, while
> the others imply they are not. PX4 handles the case to make the
> decision about the way to decode the values.

I'd be a little worried if PX4 could accept a 1e0 in an _INT 
message - but that's what you're implying :-P

peterbarker:
> You're basically always scaling when sending in a [COMMAND_INT](https://mavlink.io/en/messages/common.html#COMMAND_INT) irrespective of the `_INT` type, possibly because that command says what the scaling of lat/lon should be and otherwise the frames are the same. 

Correct.  `COMMAND_INT` specifying the values as being 1e7 *and* the frame specifying a multiplier leads to a natural conundrum of which to trust.  So rely on the message (in this case `COMMAND_INT`)  to specify the multiplier as *it* knows what the storage for the data looks like.

> You're also saying that in the other places MAV_FRAME is used, the differences specified in `_INT` are irrelevant. Right?

Yes.  Many years ago we found *many* bugs in the way we treat these frames - leading to a lack of correct multiplication or incorrect application thereof.

```
commit b0ea662c9b72d1db4ab74ba83eb8b80611c87881
Author: Peter Barker <pbarker@barker.dropbear.id.au>
Date:   Tue Dec 3 17:11:22 2019 +1100

    AP_Mission: consider _INT frames equivalent to non _INT variants
    
    AP_Mission: consider TERRAIN_ALT_INT equivalent to TERRAIN_ALT
    
    AP_Mission: consider RELATIVE_ALT_INT equivalent to RELATIVE_ALT
    
    AP_Mission: consider GLOBAL_INT equivalent to GLOBAL
```

> This made sense to @julianoes who has implemented this and treated those frames as synonymous.

Nice!

>     1. I'm wondering if we should remove the field mappings in the frame definitions, as it is "by comparison" that they are confusing.
>        So for example we currently have for the non-INT type:
>        > Global (WGS84) coordinate frame + altitude relative to the home position. First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home position.
>        
>        
>        I read that as "the values are unscaled" if you're using this frame when I compare to the `_INT` version.
>        Could we reasonably have:
>        > Global (WGS84) coordinate frame + altitude relative to the home position (altitudes are positive and 0 is the altitude of the home position).

I agree.

>        In any command that uses this we'd need to make sure the scaling was specified or implied.

Yes.

> Sorry if I'm being dim. I'm a bit worried about touching frames - it always ends badly.

:-)

Maarrk:
> We can save some confusion

I guess I am the confused person :smile:

If I can chip in as mostly a user, and a very rare and novice contributor to either project, this would make my life easier. Recently we were trying to use a new feature of ArduPilot and ran exactly into the issue this change would prevent.

This is my current understanding, relying mostly on documentation:

> I assume that [MAV_FRAME](https://mavlink.io/en/messages/common.html#MAV_FRAME) sent in COMMAND_INT should behave the same way for [MAV_FRAME_GLOBAL](https://mavlink.io/en/messages/common.html#MAV_FRAME_GLOBAL) and [MAV_FRAME_GLOBAL_INT](https://mavlink.io/en/messages/common.html#MAV_FRAME_GLOBAL_INT) ([a search of the repo](https://github.com/search?q=repo%3AArduPilot%2Fardupilot%20MAV_FRAME_GLOBAL_INT&type=code) shows them together), but only the first one was accepted in the check.

quoted from ArduPilot/ardupilot#26387

hamishwillee:
@Maarrk Yes. The theory is that these two frames should always be synonymous when used in a COMMAND_INT. However because they both exist and this is not always obvious, not every usage correctly allows either frame to be used.

Peter wants to clarify this. I'm supportive. I just know that frame definitions have historically caused much pain to modify.

Maarrk:
Do you think there could be a mention of that in the [command microservice documentation](https://mavlink.io/en/services/command.html#location-commands-and-frame-types)? Especially since it explicitly forbids using an unexpected frame:

> Note that it is an error to ignore the specified frame and use some other arbitrary frame.

Proposed addition (I can make the PR to devguide if you think it's OK):

> The message type `COMMAND_INT` implies encoding of global location as integers, deprecating separate frames ending with `_INT`. For backwards compatibility, flight stacks are advised to support the deprecated `_INT` frames and treat them the same as the `MAV_FRAME_GLOBAL` and `MAV_FRAME_GLOBAL_RELATIVE_ALT` counterpart.

I hope it's not too verbose, but past me would appreciate it.

Thank you both for taking this on and for your commitment to stability!

hamishwillee:
Thanks @Maarrk 
Yes, if this goes in we will need to update the devguide - thanks for pointing that out. However I want this PR to get merged before we make any changes in the guide. When that happens I will make some changes, using your words as inspriration.

message_definitions/v1.0/common.xml
```diff
@@ -5815,7 +5818,7 @@
...
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot). The rationale for the timestamp in the setpoint is to allow the system to compensate for the transport delay of the setpoint. This allows the system to compensate processing latency.</field>
-      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL_INT = 5, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT = 6, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT = 11</field>
+      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL = 0, MAV_FRAME_GLOBAL_RELATIVE_ALT = 3, MAV_FRAME_GLOBAL_TERRAIN_ALT = 10 (MAV_FRAME_GLOBAL_INT, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT are allowed synonyms, but have been deprecated)</field>
       <field type="uint16_t" name="type_mask" enum="POSITION_TARGET_TYPEMASK" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle.</field>
       <field type="int32_t" name="lat_int" units="degE7">X Position in WGS84 frame</field>
       <field type="int32_t" name="lon_int" units="degE7">Y Position in WGS84 frame</field>
```
hamishwillee, line 5824:
```suggestion
      <field type="int32_t" name="lat_int" units="degE7">Latitude in WGS84 frame</field>
      <field type="int32_t" name="lon_int" units="degE7">Longitude in WGS84 frame</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -5798,7 +5801,7 @@
...
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
-      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL_INT = 5, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT = 6, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT = 11</field>
+      <field type="uint8_t" name="coordinate_frame" enum="MAV_FRAME">Valid options are: MAV_FRAME_GLOBAL = 0, MAV_FRAME_GLOBAL_RELATIVE_ALT = 3, MAV_FRAME_GLOBAL_TERRAIN_ALT = 10 (MAV_FRAME_GLOBAL_INT, MAV_FRAME_GLOBAL_RELATIVE_ALT_INT, MAV_FRAME_GLOBAL_TERRAIN_ALT_INT are allowed synonyms, but have been deprecated)</field>
       <field type="uint16_t" name="type_mask" enum="POSITION_TARGET_TYPEMASK" display="bitmask">Bitmap to indicate which dimensions should be ignored by the vehicle.</field>
       <field type="int32_t" name="lat_int" units="degE7">X Position in WGS84 frame</field>
       <field type="int32_t" name="lon_int" units="degE7">Y Position in WGS84 frame</field>
```
hamishwillee, line 5807:
```suggestion
      <field type="int32_t" name="lat_int" units="degE7">Latitude in WGS84 frame</field>
      <field type="int32_t" name="lon_int" units="degE7">Longitude in WGS84 frame</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -260,9 +260,11 @@
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>
         <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
```
hamishwillee, line 264: @peterbarker How about this? I've removed the specific field names which tie this to COMMAND_INT.
I'd like to remove the scaling information but then it would be less obvious what's happening if people are looking at this.

We really want to discourage people calling this, so that in future we can rename and remove it. Any thoughts on that?

```suggestion
        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>MAV_FRAME_GLOBAL should be used instead in COMMAND_INT and elsewhere this frame might be used</description>
        <description>Global (WGS84) coordinate frame + MSL altitude. Position values are scaled by 1E7. Altitude is relative to mean sea level (MSL)</description>
```

hamishwillee: And for MAV_FRAME_GLOBAL would change

```
      <entry value="0" name="MAV_FRAME_GLOBAL">
      </entry>
```

TO

```
      <entry value="0" name="MAV_FRAME_GLOBAL">
        <description>Global (WGS84) coordinate frame + mean sea level (MSL) altitude.</description>
      </entry>
```

peterbarker:
On Tue, 19 Mar 2024, Hamish Willee wrote:

> We really want to discourage people calling this, so that in future we can rename and remove it. Any thoughts on that?


Sounds good.  We need a very long time-frame for removing this stuff, of 
course.  We need a couple of years before we can suggest to GCS operators 
to not send in the _INT frames to allow firmware implementors to make 
their adjustments.

And then we need a couple more years before we can remove the framesfrom 
the firmwares...

message_definitions/v1.0/common.xml
```diff
@@ -253,16 +253,17 @@
...
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</description>
+        <description>Global (WGS84) coordinate frame + MSL altitude (mean sea level). Position values are scaled by 1E7.</description>
       </entry>
       <entry value="6" name="MAV_FRAME_GLOBAL_RELATIVE_ALT_INT">
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_RELATIVE_ALT"/>
         <description>
           Global (WGS84) coordinate frame (scaled) + altitude relative to the home position.
           First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude with 0 being at the altitude of the home position.
```
hamishwillee, line 269:
```suggestion
        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_RELATIVE_ALT">Use MAV_FRAME_GLOBAL_RELATIVE_ALT in COMMAND_INT (and elsewhere) as a synonymous replacement.</description>
        <description>
          Global (WGS84) coordinate frame (scaled) + altitude relative to the home position.
```

message_definitions/v1.0/common.xml
```diff
@@ -253,16 +253,17 @@
...
           Global (WGS84) coordinate frame + altitude relative to the home position.
-          First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home position.
         </description>
       </entry>
       <entry value="4" name="MAV_FRAME_LOCAL_ENU">
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL"/>Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</description>
```
hamishwillee, line 262:
```suggestion
        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL">Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</deprecated>
```

message_definitions/v1.0/common.xml
```diff
@@ -280,9 +281,10 @@
         <description>This is the same as MAV_FRAME_BODY_FRD.</description>
       </entry>
       <entry value="10" name="MAV_FRAME_GLOBAL_TERRAIN_ALT">
-        <description>Global (WGS84) coordinate frame with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees, second value / y: longitude in degrees, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
+        <description>Global (WGS84) coordinate frame with AGL altitude (altitude at ground level).</description>
       </entry>
       <entry value="11" name="MAV_FRAME_GLOBAL_TERRAIN_ALT_INT">
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_TERRAIN_ALT"/>
         <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
```
hamishwillee, line 288:
```suggestion
        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL_TERRAIN_ALT">Use MAV_FRAME_GLOBAL_TERRAIN_ALT in COMMAND_INT (and elsewhere) as a synonymous replacement.</deprecated>
        <description>Global (WGS84) coordinate frame (scaled) with AGL altitude (altitude at ground level).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -253,19 +253,19 @@
...
-          First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the home position.
         </description>
       </entry>
       <entry value="4" name="MAV_FRAME_LOCAL_ENU">
         <description>ENU local tangent frame (x: East, y: North, z: Up) with origin fixed relative to earth.</description>
       </entry>
       <entry value="5" name="MAV_FRAME_GLOBAL_INT">
-        <description>Global (WGS84) coordinate frame (scaled) + MSL altitude. First value / x: latitude in degrees*1E7, second value / y: longitude in degrees*1E7, third value / z: positive altitude over mean sea level (MSL).</description>
+        <deprecated since="2024-03" replaced_by="MAV_FRAME_GLOBAL">Use MAV_FRAME_GLOBAL in COMMAND_INT (and elsewhere) as a synonymous replacement.</deprecated>
+        <description>Global (WGS84) coordinate frame + MSL altitude (mean sea level). Position values are scaled by 1E7.</description>
```
hamishwillee, line 263:
```suggestion
        <description>Global (WGS84) coordinate frame (scaled) + altitude relative to mean sea level (MSL).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -242,7 +242,7 @@
       Some deprecated frames do not follow these conventions (e.g. MAV_FRAME_BODY_NED and MAV_FRAME_BODY_OFFSET_NED).
  </description>
       <entry value="0" name="MAV_FRAME_GLOBAL">
-        <description>Global (WGS84) coordinate frame + MSL altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude over mean sea level (MSL).</description>
+        <description>Global (WGS84) coordinate frame + MSL altitude (altitude above mean sea level).</description>
```
hamishwillee, line 245:
```suggestion
        <description>Global (WGS84) coordinate frame + altitude relative to mean sea level (MSL).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -280,10 +280,11 @@
         <description>This is the same as MAV_FRAME_BODY_FRD.</description>
       </entry>
       <entry value="10" name="MAV_FRAME_GLOBAL_TERRAIN_ALT">
-        <description>Global (WGS84) coordinate frame with AGL altitude (at the waypoint coordinate). First value / x: latitude in degrees, second value / y: longitude in degrees, third value / z: positive altitude in meters with 0 being at ground level in terrain model.</description>
+        <description>Global (WGS84) coordinate frame with AGL altitude (altitude at ground level).</description>
```
hamishwillee, line 281: Note, I have removed mention of waypoint because the location depends on context - it doesn't have to be a waypoint.

I removed the mention of terrain map because again, this depends on context - it might come from a LIDAR or similar.

hamishwillee:
@julianoes Can you sanity check this one too? As part of that, can you check whether PX4 does indeed use these as synonyms and support either in the cases where the command is sent?

I'll ping in Auterion too.

julianoes (approved):

julianoes:
I doubt PX4 does it correct yet, we'll have to fix that.

hamishwillee:
@auturgy @peterbarker Any final words on this one? If not, I'll merge (probably next week). Feedback from Silvan and Alessandro in Auterion positive.

hamishwillee (approved):

peterbarker:
> @auturgy @peterbarker Any final words on this one? If not, I'll merge (probably next week). Feedback from Silvan and Alessandro in Auterion positive.

LGTM, thanks!

Perhaps fewer commits would be nice, but that's it.

auturgy:
Commits will get squashed on merge. Probably no need to wait until next week.

hamishwillee:
OK. I'm feeling brave.,

PR #2094, opened 2024-03-13
GIMBAL_DEVICE_FLAGS_YAW* replace the lock

hamishwillee:
The `GIMBAL_DEVICE_FLAGS_YAW_LOCK` is retained for backwards compatibility. This is captured in https://github.com/mavlink/mavlink-devguide/pull/531 but perhaps here as well @Davidsastresas @julianoes ?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -553,6 +553,7 @@
         <description>Lock pitch angle to absolute angle relative to horizon (not relative to vehicle). This is generally the default with a stabilizing gimbal.</description>
       </entry>
       <entry value="16" name="GIMBAL_DEVICE_FLAGS_YAW_LOCK">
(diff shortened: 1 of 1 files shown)


julianoes (commented):
But that's not quite right. It's still used as a control flag, so it still signals whether yaw is locked or not.

hamishwillee:
@julianoes Can you edit it to capture the point then? Or should I just close this?

julianoes:
It certainly can't be marked deprecated because it's still used.

Davidsastresas:
So maybe we can only comment in there that the preferred method is using the other flags. But as we have it in the docs, I don't see completely wrong the mavlink description as it is now either.

hamishwillee:
Thanks! Closing this as unnecessary - the gimbal device status message itself captures the important parts, and as you say the flag is not deprecated in all cases.

PR #2096, opened 2024-03-17
Python Example

harunkurtdev:
Python UDP example

### DIFF
--- a/examples/python/README.md
+++ b/examples/python/README.md
@@ -1,0 +1,17 @@
+# Simple Python example
+
+Simple example receiving and sending MAVLink v2 over UDP based on POSIX APIs (e.g. Linux, BSD, macOS).
+
+## Install PyMAVLink
+
+
+```bash
+pip install pymavlink
+```
+
+## Run example
+
+```bash
+cd example/python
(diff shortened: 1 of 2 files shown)


Maarrk:
Hi, did you know that the devguide already has an article about [using MAVLink with Python](https://mavlink.io/en/mavgen_python/)? I'm not sure adding this example with few explanations helps more than the full documentation there.

hamishwillee:
I'm not opposed to this kind of example (though of course more explanation is always nice). 

My issue is that because we don't have many "full" examples people might take this as a good example of how we recommend doing things. If using the MAVLink C libraries you might well do something like this to set up your connection. because the C libraries implement very little of the connection semantics.

But Pymavlink is different. If you're using `mavutil` you do the setup for  any particular connection using `mavlink_connection()` and passing the connection string - like `master = mavutil.mavlink_connection("/dev/ttyACM0", baud=115200)`.
The example uses `mavutil` for other things, just not for this main case.

So my first question is "When would a user want to directly implement the connection semantics like this?".

Then we need to provide that context in the example, and also link from sidebar etc.

PR #2097, opened 2024-03-18
common: trying to make sense of radius unit

julianoes:
This fixes the unit of the tracking "point" radius. My choice is relative to frame width and not pixels because camera and stream might not have the same resolution.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2149,7 +2149,7 @@
 
           When used in a mission, an autopilot should execute the MAV_CMD for a specified local camera (param1 = 1-6), or resend it as a command if it is intended for a MAVLink camera (param1 = 7 - 255), setting the command's target_component as the param1 value (and setting param1 in the command to zero).
           If the param1 is 0 the autopilot should do both.
-          
+
           When sent in a command the target MAVLink address is set using target_component.
           If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
           If addressed to a MAVLink camera, param 1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -2207,7 +2207,7 @@
         <description>If the camera supports point visual tracking (CAMERA_CAP_FLAGS_HAS_TRACKING_POINT is set), this command allows to initiate the tracking.</description>
         <param index="1" label="Point x" minValue="0" maxValue="1">Point to track x value (normalized 0..1, 0 is left, 1 is right).</param>
         <param index="2" label="Point y" minValue="0" maxValue="1">Point to track y value (normalized 0..1, 0 is top, 1 is bottom).</param>
-        <param index="3" label="Radius" minValue="0" maxValue="1">Point radius (normalized 0..1, 0 is image left, 1 is image right).</param>
+        <param index="3" label="Radius" minValue="0" maxValue="1">Point radius (normalized 0..1, 0 is one pixel, 1 is full image width).</param>
```
hamishwillee, line 2210: I think what it is saying now is that the point radius can be some proportion of the total width.
- If the value is 0 then the radius is not "0", it is the smallest selectable picture element (pixel). That makes sense to me because with a normalized radius there is otherwise no way to specify "just one pixel"
- If the value is 1 then it is the a radius that as many pixels as make up the image width. 
  That seems fine too. Though if the image has a much greater height that width the radius might not be as big as hoped in some cases.

Is that right? If so, this seems correct to me. Do you see any risk?

julianoes: I think the value 1 is a bit odd, but it should be clear at least.

hamishwillee: Yes it is. Definitely the old version is just wrong/unimplementable. I'm going to merge.

hamishwillee (approved):
Approving - the old version definitely wrong, and this version makes sense.

julianoes:
Thanks @hamishwillee!

PR #2100, opened 2024-04-02
[WIP] common.xml: add new frame MAV_FRAME_GLOBAL_VEHICLERELATIVE_ALT

peterbarker:
this is expected to be used to tell a vehicle to take off to 20m above current altitude.

ArduPlane already actually does this for some commands, but uses an inappropriate altitude frame.

Should also be useful for "go up 20m".

ArduPilot's PR is here: https://github.com/ArduPilot/mavlink/pull/358

For discussion at this point only, please.  Need to make sure this is fit-for-purpose in the ArduPilot codebase :-)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -321,6 +321,9 @@
       <entry value="21" name="MAV_FRAME_LOCAL_FLU">
         <description>FLU local tangent frame (x: Forward, y: Left, z: Up) with origin fixed relative to earth. The forward axis is aligned to the front of the vehicle in the horizontal plane.</description>
       </entry>
+      <entry value="22" name="MAV_FRAME_GLOBAL_ALT_VEHICLE_RELATIVE">
+        <description>Global (WGS84) coordinate frame + altitude positive up relative to the vehicle's current altitude. This frame is useful for commands such as "take off 20m above current altitude".</description>
+      </entry>
     </enum>
     <enum name="MAVLINK_DATA_STREAM_TYPE">
       <entry value="0" name="MAVLINK_DATA_STREAM_IMG_JPEG">


message_definitions/v1.0/common.xml
```diff
@@ -321,6 +321,9 @@
       <entry value="21" name="MAV_FRAME_LOCAL_FLU">
         <description>FLU local tangent frame (x: Forward, y: Left, z: Up) with origin fixed relative to earth. The forward axis is aligned to the front of the vehicle in the horizontal plane.</description>
       </entry>
+      <entry value="22" name="MAV_FRAME_GLOBAL_VEHICLERELATIVE_ALT">
+        <description>Global (WGS84) coordinate frame + altitude relative to the vehicle's current altitude. First value / x: latitude, second value / y: longitude, third value / z: positive altitude with 0 being at the altitude of the vehicle's current location.  This frame is useful for commands such as "take off 20m above current altitude."</description>
```
hamishwillee, line 325: For consistency with other globals.

```suggestion
        <description>Global (WGS84) coordinate frame + altitude positive up relative to the vehicle's current altitude. This frame is useful for commands such as "take off 20m above current altitude".</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -321,6 +321,9 @@
       <entry value="21" name="MAV_FRAME_LOCAL_FLU">
         <description>FLU local tangent frame (x: Forward, y: Left, z: Up) with origin fixed relative to earth. The forward axis is aligned to the front of the vehicle in the horizontal plane.</description>
       </entry>
+      <entry value="22" name="MAV_FRAME_GLOBAL_VEHICLERELATIVE_ALT">
```
hamishwillee, line 324: How about this? Mostly I find it hard to parse "VEHICLE_RELATIVE"

```suggestion
      <entry value="22" name="MAV_FRAME_GLOBAL_ALT_VEHICLE_RELATIVE">
```

hamishwillee:
@peterbarker I think it's a great idea. @julianoes @auturgy ANy thoughts?

FYI, have converted to draft and added WIP to title to stop me trying to merge it. Move to ready for review when it is ready to merge.

message_definitions/v1.0/common.xml
peterbarker: > How about this? Mostly I find it hard to parse "VEHICLERELATIVE" as one word

Yeah, I was trying to group them together as it'd usually be "vehicle-relative altitude".  But you're absolutely right.

peterbarker:
This one got yelled down at ArduPilot DevCall - Randy doesn't like it.

Within the ArduPilot code Plane is using `MAV_FRAME_LOCAL_OFFSET_NED` (it only needs the altitude); that's a frame which moves with the vehicle, so "D" should be relative to vehicle current location.  Perhaps that will have to do for now....

hamishwillee:
What Randy is saying will of course "work" but I don't like it very much. Mostly because if you're working in global frames it isn't likely to be obvious that you need to switch to this local frame to get the desired behaviour. Also you may have to do lots of messing around to work out correct lat/lon mapping from your local frame. Doable, but harder than it needs to be.

But MAVLink doesn't invent things without an interested party who wants to implement them. So if you're not going to do this, let me know and we can close this.

julianoes (commented):
I'd like this one.

rmackay9:
Yeah, sorry, I'm really not a fan of this one.  I'm happy with the general idea of adding support for altitudes and Locations specified as an offset from the vehicle but we should not underestimate the amount of effort required to do that.  The safety feature in AP that led to this PR doesn't really need the new altitude frame.

hamishwillee:
> Yeah, sorry, I'm really not a fan of this one. I'm happy with the general idea of adding support for altitudes and Locations specified as an offset from the vehicle but we should not underestimate the amount of effort required to do that. The safety feature in AP that led to this PR doesn't really need the new altitude frame.

Thanks @rmackay9 

From my perspective this is a clear improvement on the "API definition". If we added this there would be no requirement for you to implement it - the other frame is valid, it's just nothing that anyone would ever think of when commanding a vehicle or setting a mission item.

But, we normally wouldn't do so unless there is interest from at least one flight stack to implement. Otherwise it is clutter, because there is no easy way (at least in missions) for users to know what frames are supported by a flight stack.

@peterbarker @julianoes Unless you have some intent to implement or know someone who might, we can close this?

PR #2101, opened 2024-04-11
Fix ILLUMINATOR_ERROR_FLAGS

amy-chen-skydio:
`ILLUMINATOR_ERROR_FLAGS` is meant to be bit flags so `ILLUMINATOR_ERROR_FLAGS_THERMISTOR_FAILURE` value should be 4 (bit 2) instead of 3


tom-skydio (approved):

hamishwillee (approved):
Whoops. Thanks very much.

PR #2102, opened 2024-04-17
common: extend EFI_STATUS with a 'fuel_remaining' field

TSC21:
Adding a field to the `EFI_STATUS` message so that one is able to monitor how much fuel remains in a fuel tank. This is as, if not more, important than the fuel consumed, although both values are interdependent. The value that represents the tank as being full should be present at the autopilot, either hard-coded, or, preferentially, through a parameter.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6603,6 +6603,7 @@
       <extensions/>
       <field type="float" name="ignition_voltage" units="V">Supply voltage to EFI sparking system.  Zero in this value means "unknown", so if the supply voltage really is zero volts use 0.0001 instead.</field>
       <field type="float" name="fuel_pressure" units="kPa">Fuel pressure. Zero in this value means "unknown", so if the fuel pressure really is zero kPa use 0.0001 instead.</field>
+      <field type="float" name="fuel_remaining" units="cm^3" invalid="0">The calculated amount of fuel remaining in the tank, determined by subtracting the fuel consumed from the initial full tank value. A value of zero indicates an "unknown" quantity; use -1 to represent empty.</field>
     </message>
     <!-- MESSAGE IDs 180 - 229: Space for custom messages in individual projectname_messages.xml files -->
     <message id="230" name="ESTIMATOR_STATUS">


message_definitions/v1.0/common.xml
```diff
@@ -6603,6 +6603,7 @@
       <extensions/>
       <field type="float" name="ignition_voltage" units="V">Supply voltage to EFI sparking system.  Zero in this value means "unknown", so if the supply voltage really is zero volts use 0.0001 instead.</field>
       <field type="float" name="fuel_pressure" units="kPa">Fuel pressure. Zero in this value means "unknown", so if the fuel pressure really is zero kPa use 0.0001 instead.</field>
+      <field type="float" name="fuel_remaining" units="cm^3" invalid="0">The calculated amount of fuel remaining in the tank, determined by subtracting the fuel consumed from the initial full tank value. A value of zero indicates an "unknown" quantity; use -1 to represent empty.</field>
```
hamishwillee, line 6606: @CarlEngel @julianoes @auturgy @robertlong13 @peterbarker Any comments/objections on this addition to EFI_STATUS (you have all touched this message at some point.

CarlEngel: I would propose using grams rather than volume since the weight is what affects the flight time of most UAVs. Most EFI systems measure the fuel consumption in grams. The volume measurement can change quite drastically with temperature. I would propose changing the fuel consumed value to grams as well. But this would require a much bigger change of the flight controller code.

TSC21: @CarlEngel my only thing against what you suggest is that would introduce inconsistency in thr message itself, since the `fuel_consumed` is also in cm^3, and I don't believe we can change that now as that would break existent implementations. I would prefer to keep the current units for the sake of consistency.

julianoes: What is cm^3. Never seen this unit. That being said I never had to mess with fuel in UAVs.

Given this is a float, we don't need something small for resolution but could just go with SI, so liters or kg or whatever.

auturgy:
My only concern is that this is an inferred value, not an output from the EFI system (ie it varies from the intent of the message).  Can the specific implementation be explained?  Whilst also a misuse of a message, ArduPilot uses the battery subsystem to report fuel remaining, as this directly interfaced with the relevant failsafes etc.  Unless the fuel remaining is an output of the EFI, in which case this extension would be appropriate, I think it would be better for the GCS's if the same approach was used across flight stacks. https://ardupilot.org/plane/docs/common-fuel-sensors.html

peterbarker:
Self-consistency is good.  I'd suggest sticking to ccs.

If we wanted to burn a few bytes/s we could allow the autopilot to provide the conversion from ccs to grams.  If the autopilot even knows, which stock ArduPilot certainly doesn't.  Presumably this would be the (temperature-compensated) specific gravity of the fuel - which would need to be configured.

TSC21:
Everyone, thanks for the review. It seems that we got into a situation where everyone has sort of different opinions about a specific change, which although perfectly fine, this again introduces the problem where there's seems to be no alignment to bring this forward.

I am going to try addressing everyone's review in this comment, so at least I provide my perspective:

@julianoes correct cm³ is not SI. But also, fuel_consumed is already in cm³, so as Peter says, trying to keep the consistency in the message is better, IMO, trying to make part of it correct at this point (we unfortunately have a lot of cases in MAVLink where this already happened).

@auturgy I don't necessarily agree with your statement where you state this is only about the output of an EFI system. I don't see where's is that necessarily stated and why can't we just assume this is similar to the BATTERY_STATUS message. I honestly don't think that the Ardupilot's approach to this should be taken as a standard approach to what every autopilot + gcs is doing because, and as you state, using the BATTERY_STATUS is also a message misuse. Both systems, being batteries or EFI, or even jet engines, should be treated differently, although similar failsafes can be taken into account. A situation where an hybrid system, with both EFI and battery, is used (not the case btw), then you want different information to be sent for each system, inferfed or not.

@peterbarker I share the same opinion as you and that is why I decided to keep cm³ as a unit. Regarding the extra field with the data in grams, we can sure do that, but don't want to overextend at this point - a problem at a time makes it easier to solve.

julianoes:
> But also, fuel_consumed is already in cm³, so as Peter says, trying to keep the consistency in the message is better

Ah ok, that's fine then.

hamishwillee:
Let's discuss in the MAV call this evening. My two bits for now:
- cm3 fine for me.
- if we want weight as well, how would this information be used?
- The fact that ArduPilot uses battery information for reflecting fuel and trigger levels was a pragmatic short term design decision with suboptimal results. Specifically this was done because it was easy and meant that ArduPilot did not need to think about the failsafe paths, not because it is the best for users. There are options here.
- I'm not concerned about fuel levels being guessed provided we tell the GCS and give some way to compensate. This is why BATTERY_STATUS_V2 provides [MAV_BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL](https://mavlink.io/en/messages/development.html#MAV_BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL) to indicate that values are measured/trusted.

hamishwillee:
We discussed this in the mav call last night. The concern with adding this here is that:
- EFI status reflects the status of the EFI component, not the fuel tank - it is more like and ESC monitoring message than a battery. It has been designed to match the DroneCan message. 
- Adding a fuel remaining so we can use it like a fuel monitor is poor design, in particular since there are many other fuel systems than EFI, and we'd then be using a semantically wrong message for reporting fuel on those.
- The approach used by ArduPilot to overload the battery status is also not great - it is semantically "wrong" for other fuel types, and the current solution doesn't report the values properly.

We think the right solution is a new generic fuel status message. The concept being that it can be used for any fuel system to report the levels. Design for that is here: https://github.com/mavlink/mavlink/pull/2107

Would welcome feedback - @TSC21 if you are not interested in adopting this replacement let me know. We only merge new XML entities that have a stakeholder who is interested in doing an implementation.

Closing this.

PR #2103, opened 2024-04-18
Improved docs from XML generation_toolchain

hamishwillee:
The current docs generation toolchain uses XSL conversion of the XML to HTML, followed by some massaging of the produced HTML to produce some final output.

This is very clever and all, but inherently limited for generating docs, because:
1. Files don't know about each other - we can't, for example, create a combined file, even if we want.
2. The output is HTML. That should work in markdown, but rendering is spotty.

So what this does is copy the ideas behind mavgen parser (but using beautifulsoup) to parse the XML into Python, and then generate markdown output. 

This replaces the top level files. 
Includes workflow to build the files, and start of workflow to deploy files. However no point finishing that until the old version is disabled, which depends on CI that relies on Daniel Agar. Waiting

### DIFF
--- a/.github/workflows/docs_build_and_deploy.yml
+++ b/.github/workflows/docs_build_and_deploy.yml
@@ -1,0 +1,59 @@
+name: Build markdown from XML and deploy
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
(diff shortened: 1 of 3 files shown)


hamishwillee:
Self merging.

PR #2106, opened 2024-04-24
style: add EditorConfig and format some files

flyingthingsintothings:
Add an EditorConfig to make sure editors know how to correctly edit and save files. Also strip trailing whitespace in some files that don't need it, as configured in the EditorConfig file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,40 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = tab
+tab_width = 4
+insert_final_newline = true
+
+[*.xml]
+tab_width = 2
+trim_trailing_whitespace = true
+
+[*.php]
+insert_final_newline = false
+
+[*.css]
+insert_final_newline = false
+
+[*.py]
+# Otherwise Python language server complains
+trim_trailing_whitespace = true
+
+[*.xsl]
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = true
+tab_width = 2
+
+[*.json]
+tab_width = 2
+trim_trailing_whitespace = true
+
+[*.{yml,yaml}]
+tab_width = 2
+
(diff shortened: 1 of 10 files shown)


flyingthingsintothings:
I checked files with grep and find to see what extensions there were, which files used what conventions and tried to add all the conventions in the EditorConfig file to enforce a more consistent style in the future. I also limited `trim_trailing_whitespace` to files that would normally never need trailing whitespace.

Having this file is especially handy when this code is used as a submodule where another EditorConfig file might set different settings. The editor always uses the EditorConfig that is closest to the edited file, which would be the one added here.

hamishwillee:
@flyingthingsintothings Thank you so much - this is great. In particular, should fix some of those spurious formatting errors when working with new message definitions,

@julianoes I'm approving. But good to have a second pair of eyes so would appreciate you could do a check too.

hamishwillee (approved):

.editorconfig
```diff
@@ -0,0 +1,40 @@
...
+[*.json]
+tab_width = 2
+trim_trailing_whitespace = true
+
+[*.{yml,yaml}]
+tab_width = 2
+
+[*.{c,cpp,cc,h,hpp}]
+indent_style = tab
+tab_width = 8
```
julianoes, line 39: I'd probably do 4 here.

julianoes (commented):
Fine with me.

julianoes:
@hamishwillee would be nice if your editor wouldn't leave trailing whitespace on save.

flyingthingsintothings:
I added some rules to remove trailing whitespace on save but didn't do it for all files as I know it can be meaningful in some languages.

julianoes (approved):

hamishwillee:
> @hamishwillee would be nice if your editor wouldn't leave trailing whitespace on save.

It certainly would :-)

Thanks @flyingthingsintothings - merging.

PR #2107, opened 2024-04-25
FUEL_STATUS message proposal

hamishwillee:
This is a proposal for a message to report fuel levels in a generic way that can be displayed by a GCS, and used for autopilot fuel-level failsafes. Note that ArduPilot simply uses the battery messages fields for this, with the consequence that the values displayed are inaccurate. This message can be used for all sorts of fuels.

The design is similar to the battery messages but has some differences. 

1.  Consumed/remaining levels are _measured_ values or NaN.

    Battery messages are designed around power monitors that can only measure consumed charge, and that hence report consumed charge and make an assumption that the battery is full for estimating/providing remaining charge. A smart battery tracks the amounts effectively, so they have a flag there to indicate when the values can be trusted.

   By contrast a fuel system might measure consumed fuel, or remaining fuel, or both. Rather than make assumptions, the fuel message reports what is measured and allows the consumer to make appropriate assumptions and provide feedback to the user.
 
2. Fuel types have different units - an enum provides the units that should be displayed by the GCS.

   The current enum suggests normalized units for "unknown" - but maybe this should be an invalid value or a default value for a liquid? Or perhaps suggest that flow isn't supplied in this case?

   Other enums are for battery (matching battery_status_v2 values) and liquid (in ml, and ml/s).

   What other types do we need? Gas? What units?

3. This doesn't include temperature. I'm open to the idea of extra fields as long as they are "broadly useful". Someone would have to decide what that use is, and commit to implementing.
4. This has no flags for status. I like the simplicity. Arguably we might have some basic flags like fatal fault, warning fault or whatever to allow an autopilot to react to fault behaviours without knowing what they are.

Comments welcome.

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -420,6 +420,9 @@
       <entry value="2" name="MAV_FUEL_TYPE_LIQUID">
         <description>A generic liquid fuel. Fuel levels are in millilitres (ml). Fuel rates are in millilitres/second.</description>
       </entry>
+      <entry value="3" name="MAV_FUEL_TYPE_GAS">
+        <description>A gas tank. Fuel levels are in kilo-Pascal (kPa), and flow rates are in milliliters per second (ml/s).</description>
+      </entry>
     </enum>
   </enums>
   <messages>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -523,7 +523,7 @@
       <field type="float" name="remaining_fuel" invalid="NaN">Remaining fuel until empty (measured). The value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
       <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Percentage of remaining fuel, relative to full. Values: [0-100], UINT8_MAX: field not provided.</field>
       <field type="float" name="flow_rate" invalid="NaN">Positive value when emptying/using, and negative if filling/replacing. NaN: field not provided.</field>
-      <field type="uint32_t" enum="MAV_FUEL_TYPE">Fuel type. Defines units for fuel capacity and consumption fields above.</field>
+      <field type="uint32_t" name="fuel_type" enum="MAV_FUEL_TYPE">Fuel type. Defines units for fuel capacity and consumption fields above.</field>
     </message>
     <message id="414" name="GROUP_START">
       <description>Emitted during mission execution when control reaches MAV_CMD_GROUP_START.</description>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -502,7 +502,7 @@
       <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Remaining battery energy. Values: [0-100], UINT8_MAX: field not provided.</field>
       <field type="uint32_t" name="status_flags" display="bitmask" enum="MAV_BATTERY_STATUS_FLAGS">Fault, health, readiness, and other status indications.</field>
     </message>
-    <message id="369" name="FUEL_STATUS">
+    <message id="371" name="FUEL_STATUS">
       <description>Fuel status.
         This message provides "generic" fuel level information for display in a GCS and for triggering failsafes in an autopilot.
 	The fuel type and associated units for fields in this message are defined in the enum MAV_FUEL_TYPE.
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -523,6 +523,7 @@
       <field type="float" name="remaining_fuel" invalid="NaN">Remaining fuel until empty (measured). The value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
       <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Percentage of remaining fuel, relative to full. Values: [0-100], UINT8_MAX: field not provided.</field>
       <field type="float" name="flow_rate" invalid="NaN">Positive value when emptying/using, and negative if filling/replacing. NaN: field not provided.</field>
+      <field type="float" name="temperature" units="K" invalid="NaN">Fuel temperature. NaN: field not provided.</field>
       <field type="uint32_t" name="fuel_type" enum="MAV_FUEL_TYPE">Fuel type. Defines units for fuel capacity and consumption fields above.</field>
     </message>
     <message id="414" name="GROUP_START">
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -414,13 +414,10 @@
       <entry value="0" name="MAV_FUEL_TYPE_UNKNOWN">
         <description>Not specified. Fuel levels are normalized (i.e. maximum is 1, and other levels are relative to 1.</description>
       </entry>
-      <entry value="1" name="MAV_FUEL_TYPE_BATTERY">
-        <description>A battery. Fuel levels are in Ampere-hours (Ah), and flow rates are in Amps (A)</description>
-      </entry>
-      <entry value="2" name="MAV_FUEL_TYPE_LIQUID">
+      <entry value="1" name="MAV_FUEL_TYPE_LIQUID">
         <description>A generic liquid fuel. Fuel levels are in millilitres (ml). Fuel rates are in millilitres/second.</description>
       </entry>
-      <entry value="3" name="MAV_FUEL_TYPE_GAS">
+      <entry value="2" name="MAV_FUEL_TYPE_GAS">
         <description>A gas tank. Fuel levels are in kilo-Pascal (kPa), and flow rates are in milliliters per second (ml/s).</description>
       </entry>
     </enum>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -409,6 +409,18 @@
         <description>Channel data may be out of date. This is set when the receiver is unable to validate incoming data from the transmitter and has therefore resent the last valid data it received.</description>
       </entry>
     </enum>
+    <enum name="MAV_FUEL_TYPE">
+      <description>Fuel types for use in FUEL_TYPE. Fuel types specify the units for the maximum, available and consumed fuel, and for the flow rates.</description>
+      <entry value="0" name="MAV_FUEL_TYPE_UNKNOWN">
+        <description>Not specified. Fuel levels are normalized (i.e. maximum is 1, and other levels are relative to 1.</description>
+      </entry>
+      <entry value="1" name="MAV_FUEL_TYPE_BATTERY">
+        <description>A battery. Fuel levels are in Ampere-hours (Ah), and flow rates are in Amps (A)</description>
+      </entry>
+      <entry value="2" name="MAV_FUEL_TYPE_LIQUID">
+        <description>A generic liquid fuel. Fuel levels are in millilitres (ml). Fuel rates are in millilitres/second.</description>
+      </entry>
+    </enum>
   </enums>
   <messages>
     <!-- Transactions for parameter protocol -->
@@ -489,6 +501,29 @@
       <field type="float" name="capacity_remaining" units="Ah" invalid="NaN">Remaining charge (until empty). NaN: field not provided. Note: If MAV_BATTERY_STATUS_FLAGS_CAPACITY_RELATIVE_TO_FULL is unset, this value is based on the assumption the battery was full when the system was powered.</field>
       <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Remaining battery energy. Values: [0-100], UINT8_MAX: field not provided.</field>
       <field type="uint32_t" name="status_flags" display="bitmask" enum="MAV_BATTERY_STATUS_FLAGS">Fault, health, readiness, and other status indications.</field>
+    </message>
+    <message id="369" name="FUEL_STATUS">
+      <description>Fuel status.
+        This message provides "generic" fuel level information for display in a GCS and for triggering failsafes in an autopilot.
+	The fuel type and associated units for fields in this message are defined in the enum MAV_FUEL_TYPE.
+
+	The reported `consumed_fuel` and `remaining_fuel` must only be supplied if measured: they must not be inferred from the `maximum_fuel` and the other value.
+        A recipient can assume that if these fields are supplied they are accurate.
+        If not provided, the recipient can infer `remaining_fuel` from `maximum_fuel` and `consumed_fuel` on the assumption that the fuel was initially at its maximum (this is what battery monitors assume).
+	Note however that this is an assumption, and the UI should prompt the user appropriately (i.e. notify user that they should fill the tank before boot).
+
+	This kind of information may also be sent in fuel-specific messages such as BATTERY_STATUS_V2.
+	If both messages are sent for the same fuel system, the ids and corresponding information must match.
+
+	This should be streamed (nominally at 0.1 Hz).
+      </description>
+      <field type="uint8_t" name="id" instance="true">Fuel ID. Must match ID of other messages for same fuel system, such as BATTERY_STATUS_V2.</field>
+      <field type="float" name="maximum_fuel">Capacity when full. Must be provided.</field>
+      <field type="float" name="consumed_fuel" invalid="NaN">Consumed fuel (measured). This value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="float" name="remaining_fuel" invalid="NaN">Remaining fuel until empty (measured). The value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Percentage of remaining fuel, relative to full. Values: [0-100], UINT8_MAX: field not provided.</field>
+      <field type="float" name="flow_rate" invalid="NaN">Positive value when emptying/using, and negative if filling/replacing. NaN: field not provided.</field>
+      <field type="uint32_t" enum="MAV_FUEL_TYPE">Fuel type. Defines units for fuel capacity and consumption fields above.</field>
     </message>
     <message id="414" name="GROUP_START">
       <description>Emitted during mission execution when control reaches MAV_CMD_GROUP_START.</description>


hamishwillee:
PS, this is counter proposal to https://github.com/mavlink/mavlink/pull/2102 - @peterbarker @auturgy @julianoes @TSC21 

Like everything else, it only goes in if there  is commitment to implement from one stakeholder and general interest from two stakeholders.

message_definitions/v1.0/development.xml
```diff
@@ -490,6 +502,29 @@
...
+
+	This should be streamed (nominally at 0.1 Hz).
+      </description>
+      <field type="uint8_t" name="id" instance="true">Fuel ID. Must match ID of other messages for same fuel system, such as BATTERY_STATUS_V2.</field>
+      <field type="float" name="maximum_fuel">Capacity when full. Must be provided.</field>
+      <field type="float" name="consumed_fuel" invalid="NaN">Consumed fuel (measured). This value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="float" name="remaining_fuel" invalid="NaN">Remaining fuel until empty (measured). The value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Percentage of remaining fuel, relative to full. Values: [0-100], UINT8_MAX: field not provided.</field>
+      <field type="float" name="flow_rate" invalid="NaN">Positive value when emptying/using, and negative if filling/replacing. NaN: field not provided.</field>
+      <field type="uint32_t" enum="MAV_FUEL_TYPE">Fuel type. Defines units for fuel capacity and consumption fields above.</field>
```
hamishwillee, line 526:
```suggestion
      <field type="uint32_t" name="fuel_type" enum="MAV_FUEL_TYPE">Fuel type. Defines units for fuel capacity and consumption fields above.</field>
```

TSC21:
Hi @hamishwillee. Thanks for pushing for a discussion and for putting effort in providing an alternative. The alternative does seem reasonable and I like it. I am currently discussing it with the team.

However, it will be impossible on our side to commit to implement this in a reasonable time frame and provide it upstream. For that same reason, we are probably going with my previous solution and when we have the bandwidth, we will give this one a try.

hamishwillee:
> we are probably going with my previous solution and when we have the bandwidth, we will give this one a try.

Then you'll have to fork common, because there isn't support to have this extension as part of the standard. 

I'll close this and leave the other issue open until we either get bored of it or there is an implentation.

TSC21:
> > we are probably going with my previous solution and when we have the bandwidth, we will give this one a try.
> 
> Then you'll have to fork common or add it as a dialect, because there isn't support to have the extension as part of the standard.
> 
> I'll leave this open "for a bit" in the hope that we get a stakeholder

We are fine doing that for now. Thanks Hamish

hamishwillee:
@baobrien Alignment with DroneCan is desirable, but it would be interesting to understand why they thought this would be useful? 

This is on hold until we have someone to say they will implement.

TSC21:
> @baobrien Alignment with DroneCan is desirable, but it would be interesting to understand why they thought this would be useful?

I don't think we are going to get an answer to that from who implemented it, but what is known is that fuel density and viscosity depend on temperature, and it changes fuel flow rate as well as combustion characteristics inside jet and fuel injection engines. So a temperature field would help monitor the quality of the fuel - critical because the properties can change engine performance and efficiency in a significant way. One can then use this information to change the fuel, change engine params or plan repairs if the temperatures indicate any trouble.

> This is on hold until we have someone to say they will implement.

I'm on it. Will have updates next week.

hamishwillee:
Cheers. Presumably pressure might have impact on some fuels too?

message_definitions/v1.0/development.xml
```diff
@@ -490,6 +502,29 @@
...
+	If both messages are sent for the same fuel system, the ids and corresponding information must match.
+
+	This should be streamed (nominally at 0.1 Hz).
+      </description>
+      <field type="uint8_t" name="id" instance="true">Fuel ID. Must match ID of other messages for same fuel system, such as BATTERY_STATUS_V2.</field>
+      <field type="float" name="maximum_fuel">Capacity when full. Must be provided.</field>
+      <field type="float" name="consumed_fuel" invalid="NaN">Consumed fuel (measured). This value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="float" name="remaining_fuel" invalid="NaN">Remaining fuel until empty (measured). The value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="uint8_t" name="percent_remaining" units="%" invalid="UINT8_MAX">Percentage of remaining fuel, relative to full. Values: [0-100], UINT8_MAX: field not provided.</field>
+      <field type="float" name="flow_rate" invalid="NaN">Positive value when emptying/using, and negative if filling/replacing. NaN: field not provided.</field>
```
hamishwillee, line 525:
```suggestion
      <field type="float" name="flow_rate" invalid="NaN">Positive value when emptying/using, and negative if filling/replacing. NaN: field not provided.</field>
      <field type="float" name="temperature" units="K" invalid="NaN">Fuel temperature. NaN: field not provided.</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -490,6 +502,30 @@
...
+      <field type="float" name="temperature" units="K" invalid="NaN">Fuel temperature. NaN: field not provided.</field>
```
hamishwillee, line 526: FYI, added temperature in `K` here as precise map to uavcan (except that mavlink doesn't support float 16).

message_definitions/v1.0/development.xml
F11ff: - [ ] ****

hamishwillee:
Thanks @Pedro-Roque - we can add pressure as well. Please add a github suggestion for any changes you think are needed.

Pedro-Roque:
> Thanks @Pedro-Roque - we can add pressure as well. Please add a github suggestion for any changes you think are needed.
> 
> Do we know anything about hydrogen fuel cells? Can they reasonably be covered by this?

Afaik, they take liquid hydrogen and air to produce electricity, so it should work with a liquid type for the hydrogen tank and with pressure for the air supply.

message_definitions/v1.0/development.xml
```diff
@@ -490,6 +502,30 @@
...
+
+	This kind of information may also be sent in fuel-specific messages such as BATTERY_STATUS_V2.
+	If both messages are sent for the same fuel system, the ids and corresponding information must match.
+
+	This should be streamed (nominally at 0.1 Hz).
+      </description>
+      <field type="uint8_t" name="id" instance="true">Fuel ID. Must match ID of other messages for same fuel system, such as BATTERY_STATUS_V2.</field>
+      <field type="float" name="maximum_fuel">Capacity when full. Must be provided.</field>
+      <field type="float" name="consumed_fuel" invalid="NaN">Consumed fuel (measured). This value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
+      <field type="float" name="remaining_fuel" invalid="NaN">Remaining fuel until empty (measured). The value should not be inferred: if not measured set to NaN. NaN: field not provided.</field>
```
Pedro-Roque, line 523: Not sure about this and "consumed fuel", regarding it being available only if measured - if these re not measured at all, then in which case would we need a maximum_fuel indication?

I think consumed can be inferred (in thruster-based systems, you can infer how much you used depending on how long were thrusters open).

message_definitions/v1.0/development.xml
```diff
@@ -409,6 +409,18 @@
...
     </enum>
+    <enum name="MAV_FUEL_TYPE">
+      <description>Fuel types for use in FUEL_TYPE. Fuel types specify the units for the maximum, available and consumed fuel, and for the flow rates.</description>
+      <entry value="0" name="MAV_FUEL_TYPE_UNKNOWN">
+        <description>Not specified. Fuel levels are normalized (i.e. maximum is 1, and other levels are relative to 1.</description>
+      </entry>
+      <entry value="1" name="MAV_FUEL_TYPE_BATTERY">
+        <description>A battery. Fuel levels are in Ampere-hours (Ah), and flow rates are in Amps (A)</description>
+      </entry>
+      <entry value="2" name="MAV_FUEL_TYPE_LIQUID">
```
Pedro-Roque, line 420: would add one more entry:
```
      <entry value="3" name="MAV_FUEL_TYPE_GAS">
        <description>A gas tank. Fuel levels are in bar (bar), equivalent to 100.000,00 Pascal , and flow rates are in liters-per-minute (l/s)</description>
      </entry>
```

message_definitions/v1.0/development.xml
TSC21: @hamishwillee I think the clarity we need here is that these are all measured values, and for the sake of inferring one or the other in the end system, at least one of them should be provided. If you don't provided either, this message is not useful, as @Pedro-Roque said. So we need to clarify somewhere in the message that either the consumed or the remaining fuel (or both, if available) need to be measured so the other can be inferred (or not, if not found useful) in the end system.

message_definitions/v1.0/development.xml
```diff
@@ -409,6 +409,18 @@
...
+        <description>A generic liquid fuel. Fuel levels are in millilitres (ml). Fuel rates are in millilitres/second.</description>
+      </entry>
```
hamishwillee, line 422:
```suggestion
      </entry>
      <entry value="3" name="MAV_FUEL_TYPE_GAS">
        <description>A gas tank. Fuel levels are in kilo-Pascal (kPa), and flow rates are in milliliters per second (ml/s).</description>
      </entry>
```

message_definitions/v1.0/development.xml
```diff
@@ -409,6 +409,21 @@
...
+      <entry value="3" name="MAV_FUEL_TYPE_GAS">
+        <description>A gas tank. Fuel levels are in kilo-Pascal (kPa), and flow rates are in milliliters per second (ml/s).</description>
```
hamishwillee, line 424: @Pedro-Roque OK, added fuel levels in in kPa. The options currently in the [XSD schema file](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd) were:

```
    <!-- pressure -->
    <xs:enumeration value="Pa"/>       <!-- Pascal -->
    <xs:enumeration value="hPa"/>      <!-- hecto-Pascal -->
    <xs:enumeration value="kPa"/>      <!-- kilo-Pascal -->
    <xs:enumeration value="mbar"/>     <!-- millibar -->
```

message_definitions/v1.0/development.xml
hamishwillee: Thanks. Done, see below

message_definitions/v1.0/development.xml
hamishwillee: @Pedro-Roque @TSC21 So yes, the intent here is that you only supply values that you measure. A recipient can then infer the other values. The point here though is that you don't supply the inferred value so that the recipient _knows_ what is inferred and what is not. 

This overcomes the problems we had in battery messages where you couldn't tell whether the values were inferred or not, which meant that a GCS wouldn't know to warn the user that they must fully charge in order to get a reliable remaining charge value. 

Isn't this all captured in the description?

> 	The reported `consumed_fuel` and `remaining_fuel` must only be supplied if measured: they must not be inferred from the `maximum_fuel` and the other value.
        A recipient can assume that if these fields are supplied they are accurate.
        If not provided, the recipient can infer `remaining_fuel` from `maximum_fuel` and `consumed_fuel` on the assumption that the fuel was initially at its maximum (this is what battery monitors assume).
	Note however that this is an assumption, and the UI should prompt the user appropriately (i.e. notify user that they should fill the tank before boot).

message_definitions/v1.0/development.xml
```diff
@@ -409,6 +409,21 @@
...
+      </entry>
```
hamishwillee, line 425: REmoving battery - we're not using fuel tanks for those.

```suggestion
      <entry value="1" name="MAV_FUEL_TYPE_LIQUID">
        <description>A generic liquid fuel. Fuel levels are in millilitres (ml). Fuel rates are in millilitres/second.</description>
      </entry>
      <entry value="2" name="MAV_FUEL_TYPE_GAS">
        <description>A gas tank. Fuel levels are in kilo-Pascal (kPa), and flow rates are in milliliters per second (ml/s).</description>
      </entry>
```

hamishwillee:
@auturgy (and others). We discussed in last MAV Call removing the enum and splitting the message into a gas-specific and liquid-specific messages. We rejected because this is likely to be extended with specific information about particular fuels being used. I don't think the options are mutually exclusive.

The reason I'd like to explore this is that with separate messages the units are explicit in the messages and can be read by the ground station without any "special" code. 

What we have now means that a GCS will have to code special handling for this message to check the enum and work out the units to display. If I add a new fuel type the GCS author should probably need to check the units even if they shouldn't change.

Note that this does not preclude having an enum to identify specific fuel types. It just means that the enum will only indicate the fuel, and not the units. It also means there would potentially be two of these enums for gas/liquid. If we do this we'd just create an enum with value 0 for each case "fuel_unknown".

Downside is that you have two messages, which is more costly in firmware/code handling. I also presume for some fuels you might have to send both messages - e.g. fuel cells https://github.com/mavlink/mavlink/pull/2107#issuecomment-2114657121.

My feeling is that separate `GAS_FUEL_STATUS` and `LIQUID_FUEL_STATUS` messages are a cleaner design, but that either will work and is better what we have now.

message_definitions/v1.0/development.xml
TSC21: @hamishwillee I believe what is not clear is that the user of the message needs to necessary either set the `remaining_fuel` or the `consumed_fuel`, or both, because, if you use this message and don't set either, then the message is useless. It's not so much about the inferred values, but more about having at least one of these fields filled. And also, you can infer the `consumed_fuel` from the `remaining_fuel` and `maximum_fuel` (it really depends on the sort of sensor you have in the tank). So I believe further clarification should be added to the message.

TSC21:
> @auturgy (and others). We discussed in last MAV Call removing the enum and splitting the message into a gas-specific and liquid-specific messages. We rejected because this is likely to be extended with specific information about particular fuels being used. I don't think the options are mutually exclusive.
> 
> The reason I'd like to explore this is that with separate messages the units are explicit in the messages and can be read by the ground station without any "special" code.
> 
> What we have now means that a GCS will have to code special handling for this message to check the enum and work out the units to display. If I add a new fuel type the GCS author should probably need to check the units even if they shouldn't change.
> 
> Note that this does not preclude having an enum to identify specific fuel types. It just means that the enum will only indicate the fuel, and not the units. It also means there would potentially be two of these enums for gas/liquid. If we do this we'd just create an enum with value 0 for each case "fuel_unknown".
> 
> Downside is that you have two messages, which is more costly in firmware/code handling. I also presume for some fuels you might have to send both messages - e.g. fuel cells [#2107 (comment)](https://github.com/mavlink/mavlink/pull/2107#issuecomment-2114657121).
> 
> My feeling is that separate `GAS_FUEL_STATUS` and `LIQUID_FUEL_STATUS` messages are a cleaner design, but that either will work and is better what we have now.

I am not opposed to separate messages, but that is going to be a lot more work to happen by the implementer in the autopilot side. I don't think necessarily determining the units based on the fuel type is a bad thing. But well, if you guys thing it's better to split, then let's do it. I just wouldn't like that this takes much longer to define, since there's work and capabilities depending on this being set - over-engineering the message is probably not helpful, as we don't know if it is right until we actually try it (and that is why it is in the `development` dialect).

Pedro-Roque:
I'm not totally sure why having an extra message definition makes sense (or, in that regard, multiple as we increase the vehicle fueling options).  

To me, the code that handles that enum is not more than an if condition / switch case to access if the fuel type is type I want to or not, but the rest of the "digestion" of said message is the same irrespective of the fuel type. So why duplicate it into different callbacks that will essentially do the same?

hamishwillee:
@Pedro-Roque It's a mavlink thing. If we have separate messages then we specify the units in the message. A GCS can automate display of those units if it so wishes based on parsing of the message. As it is now, you can't write generic code to get those units - you have to code specifically. 

I don't expect this argument to outweigh the cost of implementation and flash - just doing 'due diligence'.

Are you both happy with the current message (@TSC21 )?

@auturgy Can you run this past ArduPilot dev team or add approval if you have already done so?

TSC21:
> @Pedro-Roque It's a mavlink thing. If we have separate messages then we specify the units in the message. A GCS can automate display of those units if it so wishes based on parsing of the message. As it is now, you can't write generic code to get those units - you have to code specifically.
> 
> I don't expect this argument to outweigh the cost of implementation and flash - just doing 'due diligence'.
> 
> Are you both happy with the current message (@TSC21 )?
> 
> @auturgy Can you run this past ArduPilot dev team or add approval if you have already done so?

Good to go on my side @hamishwillee! Thanks for the wotk bringing this forward

TSC21 (approved):

hamishwillee:
@auturgy This message is defined in common but it has an implementation PR in PX4 with broad support https://github.com/PX4/PX4-Autopilot/pull/23137 and interest from ArduPilot too (but no PR). 

What conditions do you have before this can be merged? We could mark it WIP or move it to development - either is OK with me given the implementation PR.

hamishwillee:
> @hamishwillee the message is actually already defined in development.xml.

@TSC21 I'm clearly going (more) blind. I've updated my original message to pretend this never happened. 
I've pinged @auturgy offline - I suspect he is away. I expect this to go in latest next week's MAV call, since he has already expressed broad support.

TSC21:
> > @hamishwillee the message is actually already defined in development.xml.
> 
> @TSC21 I'm clearly going (more) blind. I've updated my original message to pretend this never happened. I've pinged @auturgy offline - I suspect he is away. I expect this to go in latest next week's MAV call, since he has already expressed broad support.

Hey @hamishwillee! Are we good to unblock this and merge it?

hamishwillee:
I haven't had final confirmation from @auturgy , but this looks good and he was supportive last time this came up. I'm going to merge because this is going into development.xml, and can therefore change if we find anything seriously broken during prototyping.

hamishwillee:
No evidence of implementation, so we propose to remove this in https://github.com/mavlink/mavlink/pull/2170 in the near future. We'd love to see this done, because it would fix annoyances in fuel reporting coming from faking that with BATTERY statuses. But we can't unless it has a stakeholder.

TSC21:
> No evidence of implementation, so we propose to remove this in #2170 in the near future. We'd love to see this done, because it would fix annoyances in fuel reporting coming from faking that with BATTERY statuses. But we can't unless it has a stakeholder.

Evidence of implementation is here: https://github.com/PX4/PX4-Autopilot/pull/23137. And this is actively being used with jet propulsione fixed wing drones.

auturgy:
Let's move it to common: as we discussed, it would be great to have a better way than wielding the "we're going to delete this" hammer to close the loop between development.xml and implementation, but the hammer approach appears to have worked in this case. Happy to take suggestions on better ways.

TSC21:
> Let's move it to common: as we discussed, it would be great to have a better way than wielding the "we're going to delete this" hammer to close the loop between development.xml and implementation, but the hammer approach appears to have worked in this case. Happy to take suggestions on better ways.

I understand the "hammer" piece, but this one was a case where that wasn't really necessary. A quick look at the PR above will tell you where this was referenced and anyone can see the PX4 Autopilot PR there. So this is not even a case where the implementation wasn't highlighted (could have been).

PR #2114, opened 2024-05-17
Fix missing increment and max value typo

alielbashir:
This PR changes the label for maxValue to be `max` rather than `min`, and also fixes the bug which causes increment to be shown when maxValue is set, even if there's no increment

## Screenshot
image


hamishwillee (approved):
Thank you - good catch

PR #2116, opened 2024-05-18
ADSB_VEHICLE.squawk octal code is sent as if decimal

jokalode:
Adds documentation that field `squawk` of `ADSB_VEHICLE` message is encoded in decimal, not in octal.

The ADS-B drivers of PX4 and ArduPilot receive the squawk as octal number, but convert it to decimal when creating the MAVLink `ADSB_VEHICLE` message.

This may come as a surprise to aerospace engineers who tend to interpret Squawk as an octal number. In turn, this can lead to a lack of recognition of emergency statuses (safety issue) or mis-identification of emergency statuses (less critical, but still a safety issue).

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6808,7 +6808,7 @@
       <field type="uint8_t" name="emitter_type" enum="ADSB_EMITTER_TYPE">ADSB emitter type.</field>
       <field type="uint8_t" name="tslc" units="s">Time since last communication in seconds</field>
       <field type="uint16_t" name="flags" enum="ADSB_FLAGS" display="bitmask">Bitmap to indicate various statuses including valid data fields</field>
-      <field type="uint16_t" name="squawk">Squawk code</field>
+      <field type="uint16_t" name="squawk">Squawk code. Note that the code is in decimal: e.g. 7700 (general emergency) is encoded as binary 0b0001_1110_0001_0100, not(!) as 0b0000_111_111_000_000</field>
     </message>
     <message id="247" name="COLLISION">
       <description>Information about a potential collision</description>


message_definitions/v1.0/common.xml
```diff
@@ -6808,7 +6808,7 @@
       <field type="uint8_t" name="emitter_type" enum="ADSB_EMITTER_TYPE">ADSB emitter type.</field>
       <field type="uint8_t" name="tslc" units="s">Time since last communication in seconds</field>
       <field type="uint16_t" name="flags" enum="ADSB_FLAGS" display="bitmask">Bitmap to indicate various statuses including valid data fields</field>
-      <field type="uint16_t" name="squawk">Squawk code</field>
+      <field type="uint16_t" name="squawk">Squawk code. Note that the code is in decimal: e.g. 7700 (general emergency) is encoded as binary 0b0001_1110_0001_0100, not(!) as 0b0000_111_111_000_000</field>
```
hamishwillee, line 6811: Just want to make this very clear.
```suggestion
      <field type="uint16_t" name="squawk">Squawk code. The 4 digit octal squawk code is sent as though it was a decimal integer. For example, the octal squawk code 7700 (general emergency) is 4032 in decimal, but is sent as 7700 (0b0001111000010100).</field>
```

hamishwillee:
Thanks very much @jokalode . @peterbarker @julianoes This seems like a good thing to me to clarify. Are you OK with it.

I have checked that PX4 writes any data it receives over mavlink to UORB without any kind of base conversions, and the drivers seem to do the same thing  - so this is correct for PX4.

peterbarker:
I think this is correct, but @magicrub is the one to know about this field properly as far as ArduPilot is concerned

hamishwillee (approved):
Approving because the Ardupilot code comments explicitly state that the number is handled as decimal and you can see that each digit is processed to check it is 7 or less. @auturgy also seemed to think this was right. 

However I'll hold off merging to give @WickedShell / @peterbarker a chance to comment.

hamishwillee:
Thanks @jokalode - really appreciate you posting this. Merged now.

PR #2118, opened 2024-05-23
Image/video capture commands do not set mode

hamishwillee:
This updates the docs to note that the image and video capture commands do not set the camera mode.
So if you're in video mode and you start video capture the command may fail (depending on whether the camera supports separate image/video modes and whether it allows capture of the opposite type in each mode.)

@julianoes @Davidsastresas @rmackay9 

FYI The camera API allows a GCS to handle cameras for the manual use case because you can query what the camera mode type is, whether separate modes are supported, and whether the camera allows capture in the opposite mode. 

I'm not sure what we can or should do for missions though.
If you're using a camera that allows capture in any mode there is no issue. If you're using a camera that supports separate image/video capture and you're in the wrong mode then the capture will fail, ideally with a status update. That's it.

What a flight stack should do IMO is check the camera mode support as part of mission feasibility checks with associated sensible status/event updates to the GCS as to why the mission is invalid.

1. Should we suggest feasibility check in the docs?
2. Is there a standard behaviour when you get a command while you're already executing a command? 
   - I.e. if I'm executing an image capture, and I get another image capture command, what should the camera do?
   - My thinking is that the command should always be rejected.. If you send another command to capture images on the same camera it should be rejected - you would first need to stop the existing capture. Reason being is that it is more predictable. 
   - This does mean that missions that have a start image capture for a particular target should probably  be preceded by a stop capture.
   - Should we state those things?
   
My concern ^^^ is that a GCS might do a simplistic mission implementation that assumes cameras can all capture images in all modes, and that the camera will always allow image capture commands to be overridden by later commands (the second is not specified)

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2186,6 +2186,10 @@
           If addressed specifically to an autopilot: param1 should be used in the same way as it is for missions (though command should NACK with MAV_RESULT_DENIED if a specified local camera does not exist).
           If addressed to a MAVLink camera, param 1 can be used to address all cameras (0), or to separately address 1 to 7 individual sensors. Other values should be NACKed with MAV_RESULT_DENIED.
           If the command is broadcast (target_component is 0) then param 1 should be set to 0 (any other value should be NACKED with MAV_RESULT_DENIED). An autopilot would trigger any local cameras and forward the command to all channels.
+
+          Note that this command does not explicitly set the camera mode.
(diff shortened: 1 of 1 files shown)


julianoes (approved):

Davidsastresas:
The change in docs looks good to me too.

However, about the behaviour when a camera is under capture and we send another capture command, I think it would be better to take the new command and forget about the old one automatically. If user/mission sends a new capture command probably we are not interested any longer in the old ones.

Reason for this is that otherwise we would need a lot of extra waypoints on surveys, which are already very waypoint consuming.

hamishwillee:
> If user/mission sends a new capture command probably we are not interested any longer in the old ones.

I'm OK with that too. The important thing is to specify it. What do cameras/flight stacks do now?

PR #2120, opened 2024-05-29
common - multiplier values for azimuth and eph/epv

### LINKED ISSUE #2062: Proposed addition/clarification of units for EPH, EPV, Azimuth
I was looking into doing an update to the Wireshark plugin to show the units of the data being displayed, and in the case of some particularly obscure units, to show a 'human readable' number alongside the raw value.
And this led me to notice a couple of things with the units that I think could be improved:

1/ EPH & EPV values have no unit set.
The eph and epv fields of GPS[2]_RAW_INT, HIL_GPS messages are (according to description) "unitless * 100", but there is no unit defined to be able to decode the value.
I would propose using a new unit of "1E2" - as "1" seems to be an accepted symbol for "unitless" (at least according to a quick internet search!).

Note that the eph and epv fields of HIGH_LATENCY2 do have units of "dm" defined. I guess that either the unit is incorrectly set (and it should again be some multiplier of 'unitless'), or it is misnamed, and it actually representing position uncertainly.

2/ Satellite Azimuth unit doesn't consider scale factor
The satellite_azimuth field of the GPS_STATUS message has a unit of "deg", but the description says "0: 0 deg, 255: 360 deg".
To be consistent with other units, and allow automatic value decoding, I would propose that a specific unit is used here too.
e.g. "degS8" (meaning "Degree Scaled to 8 bit integer"), or "deg/360*255" or any sensible simplification thereof.

I would also question if 255 is really supposed to be 360deg - as this means a value of 0 and 255 represents the same angle.
Perhaps what was intended is that 255 represents 360-(360/256) = 358.59375 ?

3/ Minor typo in schema comment
The mavschema.xsd file includes a comment next to the degE5 and degE7 units as "degrees * 10E5" and "degrees * 10E5".
To be fussy, this should either be written as "degrees * 10^5" (i.e. "^" not "E"), or "degrees * 1E5" (i.e. "1" not "10").
I know this is very pedantic and only applies in a comment, so I've only included it here as I thought it could be cleaned up alongside addition of proposed new units in case one or more of the above is taken forward!

hamishwillee:
We added a `multiplier` attribute to indicate that the value is scaled. The value is the amount you multiply the received value to get the actual value in the specified units (if there are any).

This adds multiplier for eph/epv of 1E-2 and multiplier of `360/255` for satellite_azimuth

Fixes #2062

Related to https://github.com/ArduPilot/pymavlink/pull/885, https://github.com/ArduPilot/pymavlink/pull/894

Docs update in https://github.com/mavlink/mavlink-devguide/pull/540

@shancock884 @peterbarker @julianoes This OK?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -5218,8 +5218,8 @@
       <field type="int32_t" name="lat" units="degE7">Latitude (WGS84, EGM96 ellipsoid)</field>
       <field type="int32_t" name="lon" units="degE7">Longitude (WGS84, EGM96 ellipsoid)</field>
       <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up. Note that virtually all GPS modules provide the MSL altitude in addition to the WGS84 altitude.</field>
-      <field type="uint16_t" name="eph" invalid="UINT16_MAX">GPS HDOP horizontal dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
-      <field type="uint16_t" name="epv" invalid="UINT16_MAX">GPS VDOP vertical dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="eph" invalid="UINT16_MAX" multiplier="1E-2">GPS HDOP horizontal dilution of position (unitless * 100). If unknown, set to: UINT16_MAX</field>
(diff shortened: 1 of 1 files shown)


julianoes (approved):

shancock884:
Looks good to me :-).

hamishwillee:
OK, we're doing this then.

PR #2121, opened 2024-05-30
common: add camera_device_id to CAMERA_INFORMATION

julianoes:
By adding a camera_device_id to a CAMERA_INFORMATION message, we allow an autopilot to be a proxy for a MAVLink or non-MAVLink camera.

The idea is similar to a gimbal manager implemented by an autopilot.

@hamishwillee we likely need it in a few other places and will have to think the edge cases.

This was @rmackay9's idea when trying to figure out how QGC can detect a gimbal that's proxied by the autopilot.

Also see https://github.com/mavlink/qgroundcontrol/pull/11574

An alternative would be for the autopilot to send out heartbeats from two component IDs, something that no one seems to want to do. PX4 doesn't even though it sends out messages as the camera component, and ArduPilot is reluctant to do such a hack. Plus it screws with sequence numbers unless two channels are occupied.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6909,6 +6909,7 @@
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
       <extensions/>
       <field type="uint8_t" name="gimbal_device_id" invalid="0">Gimbal id of a gimbal associated with this camera. This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.</field>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied camera, or 1-6 for non mavlink cameras. Use 0 if no camera is associated with the component.</field>
     </message>
     <message id="260" name="CAMERA_SETTINGS">
       <description>Settings of a camera. Can be requested with a MAV_CMD_REQUEST_MESSAGE command.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6909,7 +6909,7 @@
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
       <extensions/>
       <field type="uint8_t" name="gimbal_device_id" invalid="0">Gimbal id of a gimbal associated with this camera. This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.</field>
-      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied camera, or 1-6 for non mavlink cameras. Use 0 if no camera is associated with the component.</field>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
     </message>
     <message id="260" name="CAMERA_SETTINGS">
       <description>Settings of a camera. Can be requested with a MAV_CMD_REQUEST_MESSAGE command.</description>


rmackay9:
Thanks very much @julianoes for creating this PR.

Here are some other messages and commands that I think probably also need the camera_device_id field:

- https://mavlink.io/en/messages/common.html#CAMERA_SETTINGS
- https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_TRACK_POINT
- https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_TRACK_RECTANGLE
- https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_STOP_TRACKING
- https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_START_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_STOP_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_CAM_TRIGG_DIST
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_ZOOM
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_FOCUS

These commands already have support (provided for reference in case it helps with wording)

- https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_STOP_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_MODE
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_SOURCE

These commands are old style and we can ignore them I think:

- MAV_CMD_DO_DIGICAM_CONFIGURE:
- MAV_CMD_DO_DIGICAM_CONTROL:

message_definitions/v1.0/common.xml
```diff
@@ -6909,6 +6909,7 @@
       <field type="char[140]" name="cam_definition_uri">Camera definition URI (if any, otherwise only basic functions will be available). HTTP- (http://) and MAVLink FTP- (mavlinkftp://) formatted URIs are allowed (and both must be supported by any GCS that implements the Camera Protocol). The definition file may be xz compressed, which will be indicated by the file extension .xml.xz (a GCS that implements the protocol must support decompressing the file). The string needs to be zero terminated.  Use a zero-length string if not known.</field>
       <extensions/>
       <field type="uint8_t" name="gimbal_device_id" invalid="0">Gimbal id of a gimbal associated with this camera. This is the component id of the gimbal device, or 1-6 for non mavlink gimbals. Use 0 if no gimbal is associated with the camera.</field>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied camera, or 1-6 for non mavlink cameras. Use 0 if no camera is associated with the component.</field>
```
hamishwillee, line 6912: Just to be clear, this is going to be 1-6 and represent the camera id of a non-MAVLink camera associated with the component emitting the message (or zero if the thing emitting the message is a Mavlink camera) - right? 

So a GCS will get the MAVLink component address of any camera from the message source, and further refine the address by getting this ID, which is then used in commands as the corresponding camera ID param -right?

Assuming that is the case, how about 

```suggestion
      <field type="uint8_t" name="camera_device_id" invalid="0" minValue="0" maxValue="6">Camera id of a camera proxied by this component. This allows identification of 1-6 cameras attached to an autopilot or other component. 0 if the component doesn't proxy cameras and the camera can be addressed purely by component id.</field>
```

hamishwillee:
@julianoes I added a comment which I think perhaps makes this clearer (?). Are you happy to add ids for the ones @rmackay9 suggests (they make sense to me).

message_definitions/v1.0/common.xml
julianoes: I would say it's like gimbals.

- 0: not set.
- 1-6: non-mavlink camera proxied
- 7-255: MAVLink camera proxied with component ID (although I agree this one might be tricky and could lead to conflicts)

hamishwillee: Ah, so you mean that an autopilot might emit this message for a MAVLink connected camera that it "manages". So for example, you might have an autopilot with a private channel to the camera. It gets or has some kind of camera information from the camera, but either way, it does not pass that on but instead generates information on behalf of the camera?

I guess it does allow you a mechanism to support older mavlink cameras via the new API. You'd want to be very sure that the port wasn't being forwarded because otherwise you might get multiple things coming out of a camera.

In this case the text makes sense. I think this makes this more obvious for dummies?

```suggestion
      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera proxied by this component. This is the component id of a proxied MAVLink camera, or 1-6 for non MAVLink cameras.  0 if the component doesn't proxy cameras and the camera can be addressed purely by component id. NOTE, this is intended primarily for autopilots that have cameras physically attached, or that have a private channel to a MAVLink camera.</field>
```

hamishwillee: PS @julianoes You didn't answer my question about whether you'll add the other ids. You're welcome to push back on @rmackay9 or me, but if you push to me it won't happen this week.

rmackay9: Hi @hamishwillee, yes, this change is for use where the autopilot controls the camera normally using a non-MAVLink protocol.

TBH, I prefer @julianoes's original wording and it is extremely similar to what we have for gimbals.

julianoes: Correct, except gimbals _can_ be forwarded if that's the setup and then they are referred to with the MAVLink compid. That works because they are gimbal devices. But we don't have camera devices, so I think we exclude that for simplicity.

hamishwillee: OK, thanks very much

> TBH, I prefer @julianoes's original wording and it is extremely similar to what we have for gimbals.

@rmackay9 How about this very minor amendment?

```suggestion
      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if no camera is associated with the component.</field>
```

@julianoes 
Perhaps a short section in https://mavlink.io/en/services/camera.html on component-attached cameras and gimbals would be better than me trying to shoehorn details here?

rmackay9: Hi there, thanks again for all this.  I think these most recent changes to the wording are definitely an improvement.

I'm less sure about whether the last line, "Use 0 if no camera is associated with the component" makes it any easier for consumers/publishers so if it were me I would just remove that line but I don't mind too much.

hamishwillee: Thanks @rmackay9 - we do indicate that 0 is the `invalid` value, but I don't think it hurts to spell out what this means in the text (it is usual practise). Let me have a last go on that sentence, because it seems odd to say "no camera is associated with" when we kind of mean "the component is a camera"

Though I wouldn't mind 

```suggestion
      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
```

hamishwillee: @julianoes You OK with that last change. Note this is still in draft, and I won't look at it again until you decide it is ready.

rmackay9: the last sentence looks good to me I think.

julianoes: I think that's ok, thanks @hamishwillee and @rmackay9!

hamishwillee (approved):

hamishwillee:
@rmackay9 Decided to merge this in dev call because without it an autopilot can't advertise its directly attached cameras, which makes discovery impossible.

The remainder of the things you suggested are in https://github.com/mavlink/mavlink/issues/2134 - either you or Julian welcome to take those on.

rmackay9:
Hi @hamishwillee,

Great stuff, it's a start!

FrankvVeelen:
Does this also potentially fix the fact that in a mission command no camera id's can be specified to start taking pictures? It should, right?

julianoes:
Not sure @FrankvVeelen, I would guess not.

rmackay9:
@FrankvVeelen,

I'm not sure which flight stack your talking about but IMAGE_START/STOP_CAPTURE messages already included the camera instance and AP makes use of these to allow specifying which camera is used to take pictures

hamishwillee:
@FrankvVeelen The latest version of PX4 also allows uses the ids in those messages if specified - i.e. if it encounters the command in a mission and that value is set, then it will resend the command to the component id specified.  In older versions it just triggers the connected camera (if any) or sends the command to the component with ID 100.

What this does is allow you to detect what vehicle-connected cameras there are - because with this the autopilot can output CAMERA_INFORMATION for connected cameras and give them the "fake" ids 1-6. Note that PX4 does not do this, but if it outputs any CAMERA_INFORMATION you know it can only be for the single connected camera.

PR #2122, opened 2024-06-11
Added MAV_CMD_EXTERNAL_WIND_ESTIMATE to development

RomanBapst:
When taking off in GNSS denied environments with a fixed wing or a VTOL airplane, wind dead-reckoning performance strongly relies on an accurate wind estimate. Since the wind states are not really observable without any velocity / position aiding, it makes sense to allow the operator to externally set the wind states in order to improve dead-reckoning performance.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1809,6 +1809,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and strength. Can be used to to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" label="Wind strength" units="m/s" minValue="0" >Horizontal wind strength.</param>
+        <param index="2" label="Wind strength accuracy" units="m/s" >Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360"> Azimuth (relative to true north) from where the wind is blowing.</param>
+        <param index="4" label="Wind direction accuracy" units="degrees">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
       <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
         <deprecated since="2020-01" replaced_by="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW"/>
         <description>Mission command to control a camera or antenna mount, using a quaternion as reference.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1811,8 +1811,8 @@
       </entry>
       <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
         <description>Provides an external estimate of wind direction and strength. Can be used to to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
-        <param index="1" label="Wind strength" units="m/s" minValue="0" >Horizontal wind strength.</param>
-        <param index="2" label="Wind strength accuracy" units="m/s" >Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <param index="1" label="Wind strength" units="m/s" minValue="0">Horizontal wind strength.</param>
+        <param index="2" label="Wind strength accuracy" units="m/s">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
         <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360"> Azimuth (relative to true north) from where the wind is blowing.</param>
         <param index="4" label="Wind direction accuracy" units="degrees">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
         <param index="5">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1810,9 +1810,9 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
-        <description>Provides an external estimate of wind direction and strength. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
-        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind strength.</param>
-        <param index="2" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
+        <param index="2" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
         <param index="3" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
         <param index="4" label="Direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
         <param index="5">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1811,10 +1811,10 @@
       </entry>
       <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
         <description>Provides an external estimate of wind direction and strength. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
-        <param index="1" label="Wind strength" units="m/s" minValue="0">Horizontal wind strength.</param>
-        <param index="2" label="Wind strength accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
-        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
-        <param index="4" label="Wind direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
+        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind strength.</param>
+        <param index="2" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <param index="3" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
+        <param index="4" label="Direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1810,11 +1810,11 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
-        <description>Provides an external estimate of wind direction and strength. Can be used to to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <description>Provides an external estimate of wind direction and strength. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
         <param index="1" label="Wind strength" units="m/s" minValue="0">Horizontal wind strength.</param>
-        <param index="2" label="Wind strength accuracy" units="m/s">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
-        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360"> Azimuth (relative to true north) from where the wind is blowing.</param>
-        <param index="4" label="Wind direction accuracy" units="degrees">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
+        <param index="2" label="Wind strength accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
+        <param index="4" label="Wind direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1809,16 +1809,6 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
-      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
-        <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
-        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
-        <param index="2" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
-        <param index="3" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
-        <param index="4" label="Direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
-        <param index="5">Empty</param>
-        <param index="6">Empty</param>
-        <param index="7">Empty</param>
-      </entry>
       <entry value="220" name="MAV_CMD_DO_MOUNT_CONTROL_QUAT" hasLocation="false" isDestination="false">
         <deprecated since="2020-01" replaced_by="MAV_CMD_DO_GIMBAL_MANAGER_PITCHYAW"/>
         <description>Mission command to control a camera or antenna mount, using a quaternion as reference.</description>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -294,7 +294,7 @@
       </entry>
       <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
         <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
-        <param index="1" type="float" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
+        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
         <param index="2" type="float" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
         <param index="3" type="float" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
         <param index="4" type="float" label="Direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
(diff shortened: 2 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1809,6 +1809,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and strength. Can be used to to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" label="Wind strength" units="m/s" minValue="0">Horizontal wind strength.</param>
+        <param index="2" label="Wind strength accuracy" units="m/s">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
```
hamishwillee, line 1815:
```suggestion
        <param index="2" label="Wind strength accuracy" units="m/s"  invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1809,6 +1809,16 @@
...
+        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360"> Azimuth (relative to true north) from where the wind is blowing.</param>
```
hamishwillee, line 1816:
```suggestion
        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1809,6 +1809,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and strength. Can be used to to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
```
hamishwillee, line 1813:
```suggestion
        <description>Provides an external estimate of wind direction and strength. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -1809,6 +1809,16 @@
...
+        <param index="1" label="Wind strength" units="m/s" minValue="0">Horizontal wind strength.</param>
+        <param index="2" label="Wind strength accuracy" units="m/s">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360"> Azimuth (relative to true north) from where the wind is blowing.</param>
+        <param index="4" label="Wind direction accuracy" units="degrees">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
```
hamishwillee, line 1817:
```suggestion
        <param index="4" label="Wind direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -1809,6 +1809,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and strength. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" label="Wind strength" units="m/s" minValue="0">Horizontal wind strength.</param>
+        <param index="2" label="Wind strength accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
+        <param index="3" label="Wind direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
+        <param index="4" label="Wind direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
```
hamishwillee, line 1817: Should it be wind speed? Perhaps

```suggestion
        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind strength.</param>
        <param index="2" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind strength. Set to NaN if unknown.</param>
        <param index="3" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
        <param index="4" label="Direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
```

hamishwillee:
@peterbarker @auturgy @julianoes  Any comments?

hamishwillee:
Can you run format on this? Also, please put it into development.xml - we move things into common.xml when there is a validated implementation.

RomanBapst:
@hamishwillee Done!

message_definitions/v1.0/development.xml
```diff
@@ -292,6 +292,16 @@
         <param index="7" label="Altitude/Z">Center point altitude MSL/Z coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
         INT32_MAX or NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" type="float" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
```
peterbarker, line 297:
```suggestion
        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
```

peterbarker: type is determined by the message transport (cmd_long or command_int)

message_definitions/v1.0/development.xml
```diff
@@ -292,6 +292,16 @@
...
+        <param index="2" type="float" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
+        <param index="3" type="float" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
+        <param index="4" type="float" label="Direction accuracy" units="degrees" invalid="NaN">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
```
peterbarker, line 300:
```suggestion
        <param index="4" type="float" label="Direction accuracy" units="degrees">Estimated 1 sigma accuracy of wind direction. Set to NaN if unknown.</param>
```

Can't technically specify and invalid value for these as it's transport dependent.  But I'm all good with your "Set to Nan" comment.

peterbarker (changes requested):
This LGTM apart from the technical problems.

I don't think the plumbing in ArduPilot would be particularly good actually *using* this, mind you.  There's questions around how and when you use it...  wind speed is usually an output from an estimator, so does this become a default when the estimator is unable to provide one?  An input to the estimator and let it sort it out?

Should there be a way to "cancel" the setting of the airspeed, or specify a timeout on it?

RomanBapst:
> There's questions around how and when you use it... wind speed is usually an output from an estimator, so does this become a default when the estimator is unable to provide one? An input to the estimator and let it sort it out?

So this could be similar to resetting the estimator position based on an external observation, see MAV_CMD_EXTERNAL_POSITION_ESTIMATE. If the estimator has an aid source which allows it to estimate wind, then obviously it should do that. If it does not have that e.g. during GPS denied while wind dead-reckoning, then the wind states will just more or less remain at those values.
So think about it more like a reset rather than anything else. Also it can be very beneficial to provide an initial wind estimate to the estimator before takeoff, especially when you are flying without an airspeed sensor and are using synthetic airspeed.

RomanBapst:
> Should there be a way to "cancel" the setting of the airspeed, or specify a timeout on it?

I don't think so. Of course the ground station should make sure that the values are constrained to some sensible values.

peterbarker:
@RomanBapst turns out we have a forthcoming PR against ArduPilot for a very similar message.

That proposed command definition is here: https://github.com/ArduPilot/mavlink/compare/130a836efbfef0eb3287a92cd5e187de7facdce2...32e62f34b6e6247177e414a01ab22a4e762a47cd#diff-ee2e911cfc72d7f4a67b192556ec00841d7d166482938c7571c4e9fc90683999

The implementation branch is here: https://github.com/priseborough/ardupilot/commits/pr-ekfWindStateUpdate-rebase02/

My reading of the two messages is that:
 - yours is a superset of Paul's proposed command - the shape of the command is better.
 - Paul has chosen a better name for his version of this message (`MAV_CMD_EXTERNAL_WIND_ETIMATE` as it aligns with `MAV_CMD_EXTERNAL_POSITION_ESTIMATE`

Would you agree with that assessment?

RomanBapst:
> Would you agree with that assessment?

@hamishwillee 
@peterbarker Yes, I agree with that. I think Paul has not included the accuracy of wind speed and direction, which I would definitely do.

Please correct me if I'm wrong but would the following be a feasible way forward:
1) I rename the command to MAV_CMD_EXTERNAL_WIND_ESTIMATE
2) I change the command ID to 43004?
3) I use param 3 and 4 for the accuracy values to make it compatible with Paul's message?

peterbarker:
>     1. I rename the command to MAV_CMD_EXTERNAL_WIND_ESTIMATE

Yep, thanks!

>     2. I change the command ID to 43004?

Having it right next to the equivalent command for position seems nice to me, so a renumber would be good from my perspective.

>     3. I use param 3 and 4 for the accuracy values to make it compatible with Paul's message?

Suit yourself on that one.  I marginally prefer your ordering.

message_definitions/v1.0/development.xml
hamishwillee: FWIW. Well the transports we support allow for NaN on param 1, 2, 3, 4, 6, so in theory you could, and I would like that. But the XSD doesn't allow it.

message_definitions/v1.0/development.xml
```diff
@@ -292,6 +292,16 @@
         <param index="7" label="Altitude/Z">Center point altitude MSL/Z coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
         INT32_MAX or NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="215" name="MAV_CMD_DO_SET_WIND" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
+        <param index="2" type="float" label="Wind speed accuracy" units="m/s" invalid="NaN">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
```
hamishwillee, line 298:
```suggestion
        <param index="2" type="float" label="Wind speed accuracy" units="m/s">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
```

peterbarker:
On Tue, 18 Jun 2024, Hamish Willee wrote:

> FWIW. Well the transports we support allow for NaN on param 1, 2, 3, 4, 6, so in theory you could, and I would like that. But the XSD doesn't allow it.

It would also pre-suppose all possible transports for the parameters. 
Which might be OK.

hamishwillee:
@peterbarker It might be OK, but this is not a can of worms we need now.

@RomanBapst What you and Peter have discussed all sounds good to me.

hamishwillee:
OK, went for a run, and had more thoughts. Don't want to overcomplicate this, but I guess it is worth making sure the questions have been asked...

My understanding of this is that the operator sets a speed and direction, and a 1 sigma accuracy on the estimate.

1. Is the accuracy a measure of how accurate the estimate is right now, or a measure of the likely variance in the value across the whole time you are flying?
   - If it is the former, what happens if the estimate gets too far off?
   - Actually, in general, what if the sigma isn't right?
3. How is the pilot estimating this?
4. At what range is a wind estimate likely to be useful? i.e. would this be useful only over 2 m/s? 
5. As @peterbarker said, how long will this estimate last/ be useful for, and what is the trigger for the user to know they probably should reset it? Normally we'd say at least that the value resets on reboot. PX4 expects some parameters to reset on mode change.
6. Does the pilot need to know that an estimate is set, or **not set**, and what the estimate is? I.e. I show up and start flying, does the system need to prompt for an estimate?
7. If there are multiple operators, do they need to know what is set? So do we need a message to indicate the current estimate? 

One last thought, is that is this purely for operator supplied guestimates, or are there other forms of external estimators that this might be useful for? Normally a command is for low rate values where you don't care if the value is persistent, or persists when it is no longer true. For an estimator I'd normally expect a more failsafe approach where the value is streamed in a message at some rate, and is not longer trusted when the data stops coming. The thing being streamed at would normally output the value it is using too. That may not be what we have here.

hamishwillee:
@priseborough Could you advise on this discussion?

julianoes:
I'm a bit late to the party but I would think this should be a broadcast message, sent at some regular rate e.g. as the data comes from an anemometer. In other words this is an environmental information that you can subscribe to if you need it. Or am I misunderstanding this?

hamishwillee:
@julianoes Agree - that's what I was thinking with "For an estimator I'd normally expect a more failsafe approach where the value is streamed in a message at some rate, and is not longer trusted when the data stops coming. The thing being streamed at would normally output the value it is using too. "

But it does depend a bit - if the operator really is only updating the value once a flight (say) with an estimate and variance for whole flight, then having this streamed would be more flexible but more bandwidth intensive. I guess there is nothing to stop the system tuning the rate based on the variance.

Hence the questions.

nexton-winjeel:
Late to the party, but a couple of comments:
* Command should probably include an `altitude` field.
* This is specifically the horizontal wind estimate. Will we ever need to supply the vertical wind speed estimate? 

@julianoes: 
>  but I would think this should be a broadcast message, sent at some regular rate

Isn't that just the [`WIND_COV`](https://mavlink.io/en/messages/common.html#WIND_COV) message?

hamishwillee:
> n't that just the [`WIND_COV`](https://mavlink.io/en/messages/common.html#WIND_COV) message?

That's emitted, but with system id/component id added it might work.

peterbarker:
On Wed, 19 Jun 2024, Hamish Willee wrote:

>       n't that just the WIND_COV message?
> 
> That's emitted, but with system id/component id added it might work.

Broadcast wind doesn't make a lot of sense to me without a location tacked 
onto it.  If you've got two vehicles on the same link a hundred kilometres 
apart....

RomanBapst:
@hamishwillee @julianoes Instead of trying to answer all the related subquestions let me maybe try to explain the original intent once more.
The main reason why we need this message is to allow the operator of a VTOL or Fixed Wing vehicle to reset the EKF horizontal wind states to some values. Those values could come from the operator sticking out his finger and making a guess, or he could look them up on windy or whatever.
The point is just that currently, when flying GPS denied, the estimator will not learn wind and even if the operator sets the wind roughly correct then it will make a huge performance difference.

If it helps, this command does EXACTLY the same thing as https://mavlink.io/en/messages/common.html#MAV_CMD_EXTERNAL_POSITION_ESTIMATE does for the global position, it's just wind states instead of position states.

For  MAV_CMD_EXTERNAL_POSITION_ESTIMATE we also have an operator who guesses the position of the vehicle via some source (landmarks) and then sends a command to the autopilot to reset the filter states to this observation. It makes sense to add accuracy values in this case as well.

I am not sure about broadcast message, I guess we could do the same argument for MAV_CMD_EXTERNAL_POSITION_ESTIMATE, no?
When I think about broadcast, I feel like the message is meant to serve as an observation to the filter which can be fused.
The intent of this command and MAV_CMD_EXTERNAL_POSITION_ESTIMATE is rather to trigger a state reset.

priseborough:
> OK, went for a run, and had more thoughts. Don't want to overcomplicate this, but I guess it is worth making sure the questions have been asked...
> 
> My understanding of this is that the operator sets a speed and direction, and a 1 sigma accuracy on the estimate.
> 
> 1. Is the accuracy a measure of how accurate the estimate is right now, or a measure of the likely variance in the value across the whole time you are flying?

It should be at the time of the measurement. It will be used by the EKF to set the wind state variances.

>    
>    * If it is the former, what happens if the estimate gets too far off?
>    * Actually, in general, what if the sigma isn't right?

The EKF will need to provide whatever upper and lower limits are appropriate for filter  stability, but as always - Garbage In Garbage Out


> 2. How is the pilot estimating this?

Not our problem, but could be a portable weather station (maybe for close in low altitude operations when close in, but wind gradient effects would need to be factored in) or weather data from a service.

> 3. At what range is a wind estimate likely to be useful? i.e. would this be useful only over 2 m/s?

For  all values. It improves dead reckoning accuracy regardless of wind speed.

> 4. As @peterbarker said, how long will this estimate last/ be useful for, and what is the trigger for the user to know they probably should reset it? Normally we'd say at least that the value resets on reboot. PX4 expects some parameters to reset on mode change.

It would reset on reboot as per all other EKF state estimates. Remember that as long as the EKF has GPS it forms its own wind estimate. 

> 5. Does the pilot need to know that an estimate is set, or **not set**, and what the estimate is? I.e. I show up and start flying, does the system need to prompt for an estimate?

The EKF wind estimates are already available at the ground station.

> 6. If there are multiple operators, do they need to know what is set? So do we need a message to indicate the current estimate?

See above regarding ground station access to EKF wind estimate.


> 
> One last thought, is that is this purely for operator supplied guestimates, or are there other forms of external estimators that this might be useful for? Normally a command is for low rate values where you don't care if the value is persistent, or persists when it is no longer true. For an estimator I'd normally expect a more failsafe approach where the value is streamed in a message at some rate, and is not longer trusted when the data stops coming. The thing being streamed at would normally output the value it is using too. That may not be what we have here.

This would be used to reset the wind states and covariances so it doesn't matter if data stops coming. Data is trusted at the described accuracy when it is received and used to do the reset. After that the EKF is responsible for the growth rate of wind velocity state variances, ie how quickly it assumes the  wind speed to change.

This is an accuracy enhancement. The more often it gets updated, the better, but bear in mind that it's most likely use is to extend the period of time we can operate without GPS before position drift builds up to an unsafe level. For this reason an update every few minutes would likely be sufficient when operating at altitude.

message_definitions/v1.0/development.xml
```diff
@@ -292,6 +292,16 @@
         <param index="7" label="Altitude/Z">Center point altitude MSL/Z coordinate according to MAV_FRAME. If no MAV_FRAME specified, MAV_FRAME_GLOBAL is assumed.
         INT32_MAX or NaN: Use current vehicle altitude.</param>
       </entry>
+      <entry value="215" name="MAV_CMD_EXTERNAL_WIND_ESTIMATE" hasLocation="false" isDestination="false">
```
peterbarker, line 295: I was suggesting this change IDs to be right next to the existing location function.

Are you thinking otherwise?

message_definitions/v1.0/development.xml
```diff
@@ -292,6 +292,16 @@
...
+        <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
+        <param index="1" label="Wind speed" units="m/s" minValue="0">Horizontal wind speed.</param>
+        <param index="2" type="float" label="Wind speed accuracy" units="m/s">Estimated 1 sigma accuracy of wind speed. Set to NaN if unknown.</param>
+        <param index="3" type="float" label="Direction" units="degrees" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
```
peterbarker, line 299:
```suggestion
        <param index="3" type="float" label="Direction" units="deg" minValue="0" maxValue="360">Azimuth (relative to true north) from where the wind is blowing.</param>
```

similarly elsewhere
.... is `type` valid attribute, I wonder?

message_definitions/v1.0/development.xml
RomanBapst: @peterbarker Sorry my bad, that was my intention as well

RomanBapst:
@hamishwillee @peterbarker I think I addressed everything. I assume the plan was still to keep this in development, right?

peterbarker:
> @hamishwillee @peterbarker I think I addressed everything. I assume the plan was still to keep this in development, right?

Yep.  Might not take too long to qualify to move out by the sounds of it...

message_definitions/v1.0/development.xml
```diff
@@ -376,6 +376,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="43004" name="MAV_CMD_EXTERNAL_WIND_ESTIMATE" hasLocation="false" isDestination="false">
+        <description>Provides an external estimate of wind direction and speed. Can be used to provide a more accurate wind estimate in the case where the vehicle is wind dead-reckoning.</description>
```
hamishwillee, line 380: Thanks @priseborough / @peterbarker 

How about we capture @priseborough last comment here - I appreciate you all share an understanding of this, but readers in future might be ignorant like me.

```suggestion
        <description>Set an external estimate of wind direction and speed.
          This is used to provide an initial wind estimate to the estimator (EKF) in the case where the vehicle is wind dead-reckoning, extending the time when operating without GPS before before position drift builds to an unsafe level. The command might reasonably be sent every few minutes when operating at altitude.
          The value is cleared when the estimate resets itself.
        </description>
```

hamishwillee:
Other than https://github.com/mavlink/mavlink/pull/2122/files#r1652124013, are we all good with this? @auturgy ? I plan to merge because I think this makes heaps of sense with all the explanation provided (much appreciated).

message_definitions/v1.0/development.xml
RomanBapst: @hamishwillee I am ok with the description you provided. I think it's quite specific to the wind dead-reckoning use case and how the estimator deals with this message. Again, I have no problem with this.
As a side note, the message could also be used to initialise the wind states of the estimator on the ground, which in turn could be useful to implement preflight checks which e.g. make sure that the vehicle is facing the wind...
From my side, good to go!

hamishwillee: @RomanBapst Thanks. I've reworded a little to make it clear that this "might be used for", which makes it clear that there might be other use cases. I don't think we can or should go into all of them.

RomanBapst: @hamishwillee Nice. Thank you

hamishwillee (approved):
Merging - we have broad agreement for this. Note, it's going into development so changes are still allowed if needed.

Issue #2124, opened 2024-06-24
The attribute 'multiplier' is not allowed error


Aleksandr-Sopliakov:
The commit introducing the problem is https://github.com/mavlink/mavlink/commit/f1d42e2774cae767a1c0651b0f95e3286c587257

generating C++ dialect gives the following output:

```sh
ERROR: /home/wiz29/.cache/bazel/_bazel_wiz29/16df959b0e695745ca275009ddca6358/sandbox/linux-sandbox/275/execroot/_main/external/_main~third_party_extensions~mavlink/message_definitions/v1.0/common.xml:5221:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'multiplier': The attribute 'multiplier' is not allowed.
/home/wiz29/.cache/bazel/_bazel_wiz29/16df959b0e695745ca275009ddca6358/sandbox/linux-sandbox/275/execroot/_main/external/_main~third_party_extensions~mavlink/message_definitions/v1.0/common.xml:5222:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'multiplier': The attribute 'multiplier' is not allowed.
/home/wiz29/.cache/bazel/_bazel_wiz29/16df959b0e695745ca275009ddca6358/sandbox/linux-sandbox/275/execroot/_main/external/_main~third_party_extensions~mavlink/message_definitions/v1.0/common.xml:5240:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'multiplier': The attribute 'multiplier' is not allowed.
/home/wiz29/.cache/bazel/_bazel_wiz29/16df959b0e695745ca275009ddca6358/sandbox/linux-sandbox/275/execroot/_main/external/_main~third_party_extensions~mavlink/message_definitions/v1.0/common.xml:6147:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'multiplier': The attribute 'multiplier' is not allowed.
[...]

PR #2125, opened 2024-06-25
Add multiplier field to docs, if present

hamishwillee:
The multiplier attribute was recently added to XSD. You multiply the value received and the multiplier value to get the actual value sent. This is only defined in a few values currently.

This adds support to messages and commands, such that a new column is added for multiplier if the attribute is used on any field/param.

This came out of this discussion #2124


hamishwillee:
Self merging - passes CI and tested.

PR #2127, opened 2024-06-27
common.xml: Add "encoding" field to VIDEO_STREAM_INFORMATION

nexton-winjeel:
Cameras that support H.265 are becoming more common. We need a way to specify the encoding of the video stream.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3828,6 +3828,18 @@
         <description>Stream is h.264 on MPEG TS (URI gives the port number)</description>
       </entry>
     </enum>
+    <enum name="VIDEO_STREAM_ENCODING">
+      <description>Video stream encodings</description>
+      <entry value="0" name="VIDEO_STREAM_ENCODING_UNKNOWN">
+        <description>Stream encoding is unknown</description>
+      </entry>
+      <entry value="1" name="VIDEO_STREAM_ENCODING_H264">
+        <description>Stream encoding is H.264</description>
+      </entry>
+      <entry value="2" name="VIDEO_STREAM_ENCODING_H265">
+        <description>Stream encoding is H.265</description>
+      </entry>
+    </enum>
     <enum name="CAMERA_TRACKING_STATUS_FLAGS">
       <description>Camera tracking status flags</description>
       <entry value="0" name="CAMERA_TRACKING_STATUS_FLAGS_IDLE">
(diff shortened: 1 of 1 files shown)


nexton-winjeel:
My preference is to do this with a new field, but I can change it to extend the existing `VIDEO_STREAM_TYPE` enum.

hamishwillee:
@nexton-winjeel So just to be clear, we currently have a bunch of stream types that mix up control and transport layers with encoding. You're proposing to separate the encoding.

That makes sense to me, but I'd like to be sure that our existing `VIDEO_STREAM_TYPE` don't imply an encoding, or will still have the default encoding after this. 

For that to work, all the `VIDEO_STREAM_TYPE` have to be assuming h.264 right? Which is why your `VIDEO_STREAM_ENCODING` zero value is h.264? With this change you could now specify VIDEO_STREAM_ENCODING_H265 as the encoding format on any of the other types.

So the risk here is the change from `VIDEO_STREAM_TYPE_MPEG_TS_H264` to `VIDEO_STREAM_TYPE_MPEG_TS`. That's potentially a recompile change, which I am OK with as long as we inform people. I don't think it will harm any recipients _right now_ because everything will be sending H264.

What happens when they start sending HT265? Are we assuming all recipients will either update when they see our change or fail happily? If we go this approach would be we be better to deprecate the existing value and create a new one for `VIDEO_STREAM_TYPE_MPEG_TS`?

@julianoes I know little about this. If you're around would appreciate your opinion.

nexton-winjeel:
> For that to work, all the VIDEO_STREAM_TYPE have to be assuming h.264 right? Which is why your VIDEO_STREAM_ENCODING zero value is h.264? With this change you could now specify VIDEO_STREAM_ENCODING_H265 as the encoding format on any of the other types.

Right now, it's just undefined. In the case of QGroundControl, it tries to auto-detect the encoding, which works until you change the encoding while the stream is connected.

So should I change the `0` element to `VIDEO_STREAM_ENCODING_UNKNOWN` (or `VIDEO_STREAM_ENCODING_AUTO`)?

julianoes (approved):

hamishwillee:
> So should I change the 0 element to VIDEO_STREAM_ENCODING_UNKNOWN (or VIDEO_STREAM_ENCODING_AUTO)?

So currently it isn't undefined in one case - VIDEO_STREAM_TYPE_MPEG_TS_H264, but given that recipients (at least one) is autodetecting I think we can get away with `VIDEO_STREAM_ENCODING_UNKNOWN` - my preference because that is the normal default enum value. I guess we might note that HT64 is usually the MPEG_TS option.

Does that sound reasonable?

nexton-winjeel:
@hamishwillee, @julianoes: I've added the `VIDEO_STREAM_ENCODING_UNKNOWN` value.

hamishwillee:
@rmackay9 @auturgy Can you please look at this PR, which modifies the `VIDEO_STREAM_INFORMATION` to split the video encoding from its container format/transport. 

This is a compile time change for anything that uses `VIDEO_STREAM_TYPE_MPEG_TS_H264`, but is unlikely to break very much since apparently recipients autodetect the encoding currently anyway.

Something like this was required to handle the fact that we now have different encodings being used. The alternative was to create a matrix of values in the original enum. That would be small now, but scales poorly with new encodings or containers. 

Any objections?

hamishwillee (approved):

julianoes (approved):

PR #2130, opened 2024-07-01
allow for GUIDED_CHANGE_HEADING to reset heading back the default

timtuxworth:
As per discussion in ArduPilot PR  27365 https://github.com/ArduPilot/ardupilot/pull/27365#issuecomment-2198767773

### DIFF
--- a/message_definitions/v1.0/ardupilotmega.xml
+++ b/message_definitions/v1.0/ardupilotmega.xml
@@ -22,6 +22,7 @@
     <enum name="HEADING_TYPE">
       <entry value="0" name="HEADING_TYPE_COURSE_OVER_GROUND"/>
       <entry value="1" name="HEADING_TYPE_HEADING"/>
+      <entry value="2" name="HEADING_TYPE_DEFAULT"/>
     </enum>
     <!-- ardupilot specific MAV_CMD_* commands -->
     <enum name="MAV_CMD">


hamishwillee:
Thanks @timtuxworth . Normally for ardupilot.xml changes you'd push this into the ArduPilot/mavlink fork, and it would then eventually be pushed up here, whereas for common.xml changes you'd make them here and they'd be pulled down to the ardupilot fork.

@peterbarker @tridge  I'm happy to merge this if you can confirm it is all good with you?

timtuxworth:
I added it as "3" just to continue the sequence. I guess I could have used "-1" but it didn't seem to fit the pattern. Happy to change it, or add the change to ardupilot/mavlink just LMK.

hamishwillee:
I don't have a problem with that value - we generally don't use negative values in enums in MAVLink, but I don't think there is an explicit restriction. If we were designing this from scratch I'd probably make the new value `0` because for enums we tend to make the default be the default or nothing action. Too late now.

If it were me I  would probably also add a description for the whole enum that mentions MAV_CMD_GUIDED_CHANGE_HEADING - the docs infrastructure will turn that into a link 

But that's up to you/@peterbarker etc - we don't manage dialects like ardupilotmega.xml in this repo - we just accept them if submitted by someone who I know has got broad agreement to do so and they reflect the dialect repo.

PR #2131, opened 2024-07-03
common.xml: correct description of GLOBAL_POSIITON_INT.relative_alt f…

peterbarker:
…rame

I believe this is true for both ArduPilot and PX4-Autopilot


hamishwillee (approved):
@julianoes This looks correct to me from https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/mavlink/streams/GLOBAL_POSITION_INT.hpp#L93 , and if so this is good to clarify.

Can you sanity check that I'm right, and PX4 does use home as the reference here as it appears?

julianoes (approved):
Correct!

PR #2132, opened 2024-07-03
Yaw Orbit Behaviour - add an option to select system default

Claudio-Chies:
Elaborated on the use of Orbit yaw behaviour.
see associated PR: https://github.com/PX4/PX4-Autopilot/pull/23358

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -843,7 +843,7 @@
         <description>Yaw controlled by RC input.</description>
       </entry>
       <entry value="5" name="ORBIT_YAW_BEHAVIOUR_UNCHANGED">
-        <description>Yaw behaviour is not changed from current or vehicle default behaviour.</description>
+        <description>Vehicle uses current yaw behaviour (unchanged). The vehicle-default yaw behaviour is used if this value is specified when orbit is first commanded.</description>
       </entry>
     </enum>
     <enum name="WIFI_CONFIG_AP_RESPONSE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -842,6 +842,9 @@
       <entry value="4" name="ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED">
         <description>Yaw controlled by RC input.</description>
       </entry>
+      <entry value="5" name="ORBIT_YAW_BEHAVIOUR_UNCHANGED">
+        <description>Yaw behaviour is not changed from current or vehicle default behaviour.</description>
+      </entry>
     </enum>
     <enum name="WIFI_CONFIG_AP_RESPONSE">
       <description>Possible responses from a WIFI_CONFIG_AP message.</description>
@@ -1244,7 +1247,7 @@
         <description>Start orbiting on the circumference of a circle defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
         <param index="1" label="Radius" units="m">Radius of the circle. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
-        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle. NaN: use vehicle orbit yaw behavior</param>
+        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
         <param index="4" label="Orbits" units="rad" minValue="0" default="0">Orbit around the centre point for this many radians (i.e. for a three-quarter orbit set 270*Pi/180). 0: Orbit forever. NaN: Use vehicle default, or current value if already orbiting.</param>
         <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
         <param index="6" label="Longitude/Y">Center point longitude (if no MAV_FRAME specified) / Y coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1244,7 +1244,7 @@
         <description>Start orbiting on the circumference of a circle defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
         <param index="1" label="Radius" units="m">Radius of the circle. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
-        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
+        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle. NaN: use vehicle orbit yaw behavior</param>
         <param index="4" label="Orbits" units="rad" minValue="0" default="0">Orbit around the centre point for this many radians (i.e. for a three-quarter orbit set 270*Pi/180). 0: Orbit forever. NaN: Use vehicle default, or current value if already orbiting.</param>
         <param index="5" label="Latitude/X">Center point latitude (if no MAV_FRAME specified) / X coordinate according to MAV_FRAME. INT32_MAX (or NaN if sent in COMMAND_LONG): Use current vehicle position, or current center if already orbiting.</param>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -1244,7 +1244,7 @@
         <description>Start orbiting on the circumference of a circle defined by the parameters. Setting values to NaN/INT32_MAX (as appropriate) results in using defaults.</description>
         <param index="1" label="Radius" units="m">Radius of the circle. Positive: orbit clockwise. Negative: orbit counter-clockwise. NaN: Use vehicle default radius, or current radius if already orbiting.</param>
         <param index="2" label="Velocity" units="m/s">Tangential Velocity. NaN: Use vehicle default velocity, or current velocity if already orbiting.</param>
-        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
+        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle. NaN: use vehicle orbit yaw behavior</param>
```
hamishwillee, line 1247: No, set this with an enum value. The point of using an enum is to restrict the allowed values, so using a value like NaN here is unconventional.

Normally we'd set the `0` value to a system default, but it is too late to change now. Something like:

```
      <entry value="5" name="ORBIT_YAW_BEHAVIOUR_DEFAULT">
        <description>Yaw behaviour set to vehicle default.</description>
      </entry>
```

Claudio-Chies: the problem is if we expand the enum we might run into problems if the QGC version works with the expanded enum, and the PX4 firmware not, if we look in how the yaw behavior is implemented [Link](https://github.com/PX4/PX4-Autopilot/blob/4c5ce7af6b3c1a53c419b6c5d7686d6df87f34d1/src/modules/flight_mode_manager/tasks/Orbit/FlightTaskOrbit.cpp#L337)
then there wouldn't be a switch case for the new enum entry. 
Now if we use NaN (as used in the param 2, velocity) then it will skip the assignment in applyCommandParameters(), and use the _yaw_behaviour set in the constructor.

hamishwillee:
Possibly. Fix QGC.

message_definitions/v1.0/common.xml
hamishwillee: Then you need to fix either PX4 or QGC for this. We don't mess with the conventions of MAVLink if possible.

PX4 should treat values it does not understand as invalid and reject the request.

sfuhrer: >unless the design of MAVLink/messages itself is broken.

The thing is that I would argue that the design of this message IS broken. The orbit yaw behavior is best known by the vehicle, as it is governed by its hardware and gimbal capabilities. Some vehicles have the gimbal fixed (without yaw actuation) and thus need to point it to the center using the vehicle yaw, while others need to keep the body yaw aligned with the direction of travel due to more stability/efficiency when reducing side slip (most extremely for helicopters or VTOLs in hover).

This was an oversight when designing this message, as otherwise we need to assume that the operator always selects the right option when planning the Orbit (assuming this yaw options are one day exposed in an UI). 

>so using a value like NaN here is unconventional.
If this a strict requirement then okay, I guess we can append a new enum value. @Claudio-Chies If I read it correctly then older PX4 versions will just always use FRONT_TO_CIRCLE" in case of an unknown enum setting (`default` case). We can live with that.

hamishwillee: > The thing is that I would argue that the design of this message IS bro

No, I mean the design of MAVLink itself is broken, not this message. I will add this to the mav call to see what others think, but I am opposed to having just one message do something different than all others in MAVLink.

hamishwillee: @sfuhrer If you add a default value to the enum you can code for this properly going forward. I'm still waiting on comments from @auturgy and @julianoes . If I get none before, will raise in MAVLink call next week.

MaEtUgR: > yaw aligned with the direction of travel due to more stability/efficiency when reducing side slip (most extremely for helicopters or VTOLs in hover). This was an oversight when designing this message

I agree, I did not think of this case at all when designing the message. But I thought about the possibility of not defining any option if the operator has no specific opinion and wants the vehicle to do the "best thing". Sadly I didn't write this possibility out in the message definition.

Making a separate enum option for "use vehicle orbit yaw behavior" seems sensible since there is no enum option in MAVLink that can take any value outside of the options listed in the enum. Most of them (currently 160 in common.xml) are unsigned integers, 29 in common.xml are part of the command parameters which are outside of exceptions sent as float value and could hence be NAN. But that's rather a side effect and we should not exploit that to avoid confusion.

I think it's important to be able to not specify what Yaw behavior option you want to command. This is especially interesting when the vehicle is already orbiting with a certain yaw behavior and you only want to change e.g. the velocity using the command and not change the yaw behavior. The command is otherwise designed in a way that you can not specify parameters you don't want to change.

I suggest to specify also the case you don't want to set any behavior 
```suggestion
        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle. NaN: Retain already used behavior, vehicle standard if not yet in orbit</param>
```

and then according to the request by @hamishwillee make it an enum option just not with value zero but rather the next free ID 5:
image
Then this option would always need to be specified and 5 should be the default that preserves the meaning of no change, no specific requirement for yaw behavior.

For context: So far this message is WIP and to my knowledge used by PX4, QGC (hardcoded value without having UI to chose behavior https://github.com/mavlink/qgroundcontrol/pull/11650) and MAVSDK (https://mavsdk.mavlink.io/main/en/cpp/api_reference/classmavsdk_1_1_action.html#classmavsdk_1_1_action_1ad9dd7c5e85dda1ae188df75998375c92)

hamishwillee: @MaEtUgR Yes, it is still marked as WIP, but we should remove that because that's a hint to the mavlink stakeholders that this message can be changed in a breaking way - and that is no longer the case right?

Note, this is not implemented by ArduPilot.

@Claudio-Chies To make this proposed change you'd revert the current changes:

```suggestion
        <param index="3" label="Yaw Behavior" enum="ORBIT_YAW_BEHAVIOUR">Yaw behavior of the vehicle.</param>
```

Then we'd add the two new cases under discussion (I think this is what you mean Matthias).

```
      <entry value="5" name="ORBIT_YAW_BEHAVIOUR_DEFAULT">
        <description>Yaw behaviour set to vehicle default.</description>
      </entry>
      <entry value="6" name="ORBIT_YAW_BEHAVIOUR_UNCHANGED">
        <description>Yaw behaviour is not changed.</description>
      </entry>
```

hamishwillee: @julianoes Thinks having `ORBIT_YAW_BEHAVIOUR_UNCHANGED` is unnecessary - you have all the state.

MAV call more generally thinks it would be better not to use NaN. SO proposal is above, but with "6" removed.

MaEtUgR: > MAV call more generally thinks it would be better not to use NaN. SO proposal is above, but with "6" removed.

Makes sense 👍

I would also add only one new option 5 when it's not already orbiting it does default behavior if it's already orbiting with a certain behavior it's kept unchanged. That's at least what I would expect.

@hamishwillee Thanks for bringing it up and the clarifications.

hamishwillee: @MaEtUgR I'm not entirely certain what you mean by:

> when it's not already orbiting it does default behavior if it's already orbiting with a certain behavior it's kept unchanged. 

I understood this to mean that you wanted an option for `ORBIT_YAW_BEHAVIOUR` that would allow you to specify the other possible parameters such as the radius, orbits and centre location without having to re-specify the yaw behaviour.
For example, after sending the vehicle to a particular position, radius, yaw behaviour, you decide you just want to change the radius, so you send the command again but set all the parameters you want unchanged to NaN and `ORBIT_YAW_BEHAVIOUR` to the "remain unchanged" value.

This was what I was proposing as option 6:
```
      <entry value="6" name="ORBIT_YAW_BEHAVIOUR_UNCHANGED">
        <description>Yaw behaviour is not changed.</description>
      </entry>
```

Julian suggested that you already know what your state is, so you could just re-enter the value(s) of your current orbit. 

So now I'm confused about what you think we meant by `ORBIT_YAW_BEHAVIOUR_UNCHANGED`

Claudio-Chies: my understanding from my discussions with @MaEtUgR is that we would have something like option 6, which you proposed.
if the vehicle is already orbiting, it will not change it, if its initializing an orbit, it uses the vehicle default.

message_definitions/v1.0/common.xml
```diff
@@ -842,6 +842,9 @@
       <entry value="4" name="ORBIT_YAW_BEHAVIOUR_RC_CONTROLLED">
         <description>Yaw controlled by RC input.</description>
       </entry>
+      <entry value="5" name="ORBIT_YAW_BEHAVIOUR_UNCHANGED">
+        <description>Yaw behaviour is not changed from current or vehicle default behaviour.</description>
```
hamishwillee, line 846:
```suggestion
        <description>Vehicle uses current yaw behaviour (unchanged). The vehicle-default yaw behaviour is used if this value is specified when orbit is first commanded.</description>
```

message_definitions/v1.0/common.xml
hamishwillee: G'day @Claudio-Chies 

Ah, so you're saying that you only want to set the system default once, when you first initialize the yaw. Otherwise you're either going to be explicit about the specific value or send unchanged. That works.

@julianoes @auturgy I think this makes sense, and should not break anything. You're OK with that?

hamishwillee (approved):

Claudio-Chies:
@hamishwillee perfect, thanks. can you merge this PR? i don't have write access.

message_definitions/v1.0/common.xml
julianoes: Yes, that's good!

julianoes (approved):

hamishwillee:
@Claudio-Chies Happy to merge now - was waiting on independent review by stakeholders :-). It isn't enough that I think it is a good idea :-).

PR #2137, opened 2024-07-31
comment - use markdown in replaced by to fix rendering

### LINKED ISSUE #2136: MAV_CMD_DO_SET_ROI is replace by MAV_CMD_DO_SET_ROI?
At <https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_ROI>, it says:
> MAV_CMD_DO_SET_ROI (201) — [DEP]
> DEPRECATED: Replaced By MAV*CMD_DO_SET_ROI*\* (2018-01)

Is MAV_CMD_DO_SET_ROI really deprecated? if yes, what should I use instead to rotate ardupilot?

hamishwillee:
The docs were rending this item as italic when in markdown as shown. 

> MAV_CMD_DO_SET_ROI (201) — [DEP]
> DEPRECATED: Replaced By MAV*CMD_DO_SET_ROI*\* (2018-01)

This forces the markdown to code. SHould be good because this information is not rendered in the docs from the generators.

Fixes #2136 

@peterbarker Make sense to you?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1254,7 +1254,7 @@
         <param index="7" label="Altitude/Z">Center point altitude (MSL) (if no MAV_FRAME specified) / Z coordinate according to MAV_FRAME. NaN: Use current vehicle altitude.</param>
       </entry>
       <entry value="80" name="MAV_CMD_NAV_ROI" hasLocation="true" isDestination="false">
-        <deprecated since="2018-01" replaced_by="MAV_CMD_DO_SET_ROI_*"/>
+        <deprecated since="2018-01" replaced_by="`MAV_CMD_DO_SET_ROI_*`"/>
         <description>Sets the region of interest (ROI) for a sensor set or the vehicle itself. This can then be used by the vehicle's control system to control the vehicle attitude and the attitude of various sensors such as cameras.</description>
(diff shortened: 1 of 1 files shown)


julianoes (approved):

zyxkad:
shouldn't this considers as fix the markdown translater :|  A lot other commands also have simulur problem.

hamishwillee:
@zyxkad No. The markdown translator treats it as markdown, which is what I want, and the things that make this render incorrectly are valid mardown.. I can't fix it in the translator because the translator would have to _know_ that this particular markdown should be ignored while all other things do get treated as markdown.

I could fix it in the source in another way of course, but this is the least disruptive for all other users of the source, such as MAVLink and the generated docs, which expect plain text.

hamishwillee:
@zyxkad So which other commands are having similar problems? The only case I am seeing is MAVCMD_NAV_FENCE - and for similar reasons.

zyxkad:
> No. The markdown translator treats it as markdown, which is what I want, and the things that make this render incorrectly are valid mardown.

what I think is `_` between words should not convert to styles, that's what Github and most markdown parsers do
For example MAV_CMD_DO_SET_ROI_*

hamishwillee:
The renderer here is the static site generator "gitbook" - so we might change to another renderer (and plan to eventually), but that's out of scope for this discussion. Other SSGs may have their own foibles about rendering markdown.

PR #2139, opened 2024-08-01
Docs generation improvements

hamishwillee:
This makes common.xml only display XML entities in full for included files - other dialects only include a heading.

This also improves the heading text for dialects, listing the source repo where known.

There are some that still need to be confirmed.


hamishwillee:
Self merging. Passes tests.

PR #2142, opened 2024-08-09
common.xml: Add missing camera IDs to camera commands and messages

### LINKED ISSUE #2134: Add missing camera ids to camera commands
We now use a camera_device_id field to allow addressing of components that are attached to autopilots and to allow commands used in missions to be addressed to a specific camera.

The following messages/commands should probably have camera_device_id field:

- https://mavlink.io/en/messages/common.html#CAMERA_SETTINGS
- https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_TRACK_POINT
- https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_TRACK_RECTANGLE
- https://mavlink.io/en/messages/common.html#MAV_CMD_CAMERA_STOP_TRACKING
- https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_START_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_VIDEO_STOP_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_DO_SET_CAM_TRIGG_DIST
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_ZOOM
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_FOCUS

These commands already have support (provided for reference in case it helps with wording)

- https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_START_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_IMAGE_STOP_CAPTURE
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_MODE
- https://mavlink.io/en/messages/common.html#MAV_CMD_SET_CAMERA_SOURCE

These commands are old style and we can ignore them I think:

- MAV_CMD_DO_DIGICAM_CONFIGURE:
- MAV_CMD_DO_DIGICAM_CONTROL:

This falls out of https://github.com/mavlink/mavlink/pull/2121#issuecomment-2143980798 by @rmackay9 .

I'm assigning to @rmackay9 and @julianoes 

We also need docs related to this philosophy - does anyone want to take that, or should I?

nexton-winjeel:
* Adds a "Target Camera ID" param to Camera commands that don't already have it;
* Adds a `camera_device_id` field to Camera messages that don't already have it; and
* ~Adds a new Camera capability flag `CAMERA_CAP_FLAGS_USES_CAMERA_IDS` to report that the camera is using camera IDs rather than component IDs.~

FYI: @hamishwillee, @julianoes, @rmackay9 

Fixes #2134


julianoes:
Thanks for that @nexton-winjeel. I need to review this in  a quiet minute.

message_definitions/v1.0/common.xml
```diff
@@ -2108,10 +2108,11 @@
         <description>Reset all camera settings to Factory Default</description>
         <param index="1" label="Reset" minValue="0" maxValue="1" increment="1">0: No Action 1: Reset all settings</param>
         <param index="2">Reserved (all remaining params)</param>
+        <param index="3" label="Target Camera ID" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras that don't have a distinct component id (such as autopilot-attached cameras). 0: all cameras. This is used to specifically target autopilot-connected cameras or individual sensors in a multi-sensor MAVLink camera. It is also used to target specific cameras when the MAV_CMD is used in a mission</param>
```
julianoes, line 2111: This should be param 2. We don't need to keep Reserved as param 2.

message_definitions/v1.0/common.xml
```diff
@@ -3806,6 +3811,9 @@
       <entry value="2048" name="CAMERA_CAP_FLAGS_HAS_TRACKING_GEO_STATUS">
         <description>Camera supports tracking geo status (CAMERA_TRACKING_GEO_STATUS).</description>
       </entry>
+      <entry value="4096" name="CAMERA_CAP_FLAGS_USES_CAMERA_IDS">
+        <description>Camera has multiple capture devices that are addressed by camera ID rather than component ID.</description>
+      </entry>
```
julianoes, line 3816: I'm not sure whether we want this, just because it is redundant information, right? When [CAMERA_INFORMATION](https://mavlink.io/en/messages/common.html#CAMERA_INFORMATION) is sent out with these flags, we already know about the camera ID from the `camera_device_id`, so we don't need the flag.

I'm saying this because it's easy to forget one or the other as soon as you have two ways to signal it. And that then means the other side needs to parse both and make some sort of guess based on the two pieces of information. Should it be an OR, or AND. I'd rather just have one field to debug.

julianoes (commented):
Thanks for doing the work here @nexton-winjeel! See my inline comments.

message_definitions/v1.0/common.xml
nexton-winjeel: Fixed.

message_definitions/v1.0/common.xml
nexton-winjeel: Yeah, you're right. I've removed it.

julianoes (approved):
Good from my side, thanks @nexton-winjeel!

rmackay9:
I haven't checked every messages but those that I did check looked OK, thanks!

hamishwillee:
Awesome. Can we update the labels and docs to match for the cases that have already been done? Such as [MAV_CMD_IMAGE_START_CAPTURE](http://localhost:4000/en/services/camera.html#MAV_CMD_IMAGE_START_CAPTURE)
[MAV_CMD_IMAGE_STOP_CAPTURE](http://localhost:4000/en/services/camera.html#MAV_CMD_IMAGE_STOP_CAPTURE), which use `id` rather than `Target camera id`. 

Anything that can be changed to be consistent without breaking anything should be.

hamishwillee:
@nexton-winjeel @julianoes @rmackay9 @Davidsastresas There are open questions in https://github.com/mavlink/mavlink-devguide/pull/517 that I would like addressed before we merge this.

Specifically:
- confirm it is OK not to be able to address cameras attached to a companion computer in a mission
- decide how we get non-MAVLink cameras to emit a CAMERA_INFORMATION. We could just ping the whole system with the request, or we could decide that we only care about companions and autopilots, and ping those for CAMERA_INFORMATION.
  This comes up because our current process only queries components that identify as cameras, not autopilots.

Davidsastresas:
>   This comes up because our current process only queries components that identify as cameras, not autopilots.

I answered in the other thread, but about this one, I think in QGC 4.4 Julian added recently support to query for autopilots compids too

message_definitions/v1.0/common.xml
```diff
@@ -1721,7 +1721,7 @@
         <param index="1" label="Distance" units="m" minValue="0">Camera trigger distance. 0 to stop triggering.</param>
         <param index="2" label="Shutter" units="ms" minValue="-1" increment="1">Camera shutter integration time. -1 or 0 to ignore</param>
         <param index="3" label="Trigger" minValue="0" maxValue="1" increment="1">Trigger camera once immediately. (0 = no trigger, 1 = trigger)</param>
-        <param index="4">Empty</param>
+        <param index="4" label="Target Camera ID" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras that don't have a distinct component id (such as autopilot-attached cameras). 0: all cameras. This is used to specifically target autopilot-connected cameras or individual sensors in a multi-sensor MAVLink camera. It is also used to target specific cameras when the MAV_CMD is used in a mission</param>
```
hamishwillee, line 1724: @nexton-winjeel As per https://github.com/mavlink/mavlink-devguide/pull/517#issuecomment-2290360512 we're only going to allow the id to be used by an autopilot. If you want to attach multiple cameras to a companion, or have some other setup, then the component must implement each as a unique mavlink component. That means this has to change. Can you apply this everywhere?

```suggestion
        <param index="4" label="Target Camera ID" minValue="0" maxValue="255" increment="1">Target camera ID. 7 to 255: MAVLink camera component id. 1 to 6 for cameras attached to the autopilot, which don't have a distinct component id. 0: all cameras. This is used to target specific autopilot-connected cameras. It is also used to target specific cameras when the MAV_CMD is used in a mission.</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -7101,6 +7114,8 @@
       <field type="float" name="dist" units="m" invalid="NaN">Distance between camera and tracked object. NAN if unknown</field>
       <field type="float" name="hdg" units="rad" invalid="NaN">Heading in radians, in NED. NAN if unknown</field>
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
+      <extensions/>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
```
hamishwillee, line 7118: @julianoes Is this right? I.e.  this will now either be the id of the autopilot hardwired camera OR 0.

The implication of the previous text was that the autopilot could proxy a mavlink camera too. My assumption is that you never do that - MAVLink camera messages are forwarded, and this is only used as a proxy for autopilot-connected cameras.

```suggestion
      <field type="uint8_t" name="camera_device_id" default="0" minValue="0" maxValue="6">Camera id of a non-MAVLink camera attached to an autopilot (1-6).  0 if the component is a MAVLink camera (with its own component id).</field>
```

julianoes: Yes, I think so.

hamishwillee: @nexton-winjeel And all message id fields should be like this pattern.

nexton-winjeel:
Updated:
* replaced `id` with `Target Camera ID` in `MAV_CMD_IMAGE_START_CAPTURE` and `MAV_CMD_IMAGE_STOP_CAPTURE`
* updated description of "Target Camera ID" params
* updated description of "camera_device_id" fields

hamishwillee (approved):
Looks great to me. @julianoes Can you sanity check too?

hamishwillee:
@julianoes Except I'm a bit sleepy and hit the merge button :-(. I'm OK with that, this is sane.

rmackay9:
nice!!

PR #2143, opened 2024-08-12
MAV_CMD_DO_SET_WIND - Set a custom wind directions

python36:
When flying without a GPS receiver, drift correction does not work properly because the autopilot does not have information about the wind speed and direction.
I can get wind information using online services (e.g. windy.com) and with this command set a custom wind speed and direction, which will greatly improve drift correction.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2466,6 +2466,17 @@
         <param index="6">Empty.</param>
         <param index="7">Empty.</param>
       </entry>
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thank you, we already have a wind setting message in development.xml: https://mavlink.io/en/messages/development.html#MAV_CMD_EXTERNAL_WIND_ESTIMATE

If you think this does not meet your needs feel free to reopen. Note, Peter Barker probably a good person to chat about this with in the ArduPilot community.

python36:
@hamishwillee many thanks

PR #2144, opened 2024-08-14
common: add capability flags for camera and gimbal

julianoes:
Having these two capability flags means that ground stations don't have to ask any discovered components for CAMERA_INFORMATION and GIMBAL_MANAGER_INFORMATION messages and can instead only do so when they see this capability flag set.

This means however, that each ground station now needs to request either AUTOPILOT_VERSION or COMPONENT_INFORMATION_BASIC from any MAVLink component that it discovers, so it's only slightly better.

The alternative would be to make it explicit that cameras and gimbals need to send out CAMERA_CAPTURE_STATUS and GIMBAL_MANAGER_STATUS messages respsectively, to signal implementing a camera/gimbal manager.

FYI @Davidsastresas

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3230,6 +3230,14 @@
           Note that either this flag or MAV_PROTOCOL_CAPABILITY_PARAM_ENCODE_BYTEWISE should be set if the parameter protocol is supported.
         </description>
       </entry>
+      <entry value="262144" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_CAMERA">
+        <description>This component implements/is a camera. This means the CAMERA_INFORMATION, and other messages can be requested.
+        </description>
+      </entry>
+      <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER">
+        <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
+        </description>
+      </entry>
     </enum>
     <enum name="MAV_MISSION_TYPE">
       <description>Type of mission items being requested/sent in mission protocol.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3230,11 +3230,7 @@
           Note that either this flag or MAV_PROTOCOL_CAPABILITY_PARAM_ENCODE_BYTEWISE should be set if the parameter protocol is supported.
         </description>
       </entry>
-      <entry value="262144" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_CAMERA">
-        <description>This component implements/is a camera. This means the CAMERA_INFORMATION, and other messages can be requested.
-        </description>
-      </entry>
-      <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER">
+      <entry value="262144" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER">
         <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
         </description>
       </entry>


hamishwillee:
Sorry, I missed this. Can it wait until Wednesday?

julianoes:
@hamishwillee I only added this after the call, as discussed.

Davidsastresas:
Thanks Julian, this is good!

hamishwillee:
Prior to this, a GCS discovers gimbal manages by sending a broadcast [MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/services/gimbal_v2.html#MAV_CMD_REQUEST_MESSAGE) for [GIMBAL_MANAGER_INFORMATION](https://mavlink.io/en/services/gimbal_v2.html#GIMBAL_MANAGER_INFORMATION). Every gimbal manager should respond with GIMBAL_MANAGER_INFORMATION. That includes autopilot attached gimbals.

Prior to this a GCS discovers cameras by looking for HEARTBEAT with type of camera, and then requests CAMERA_INFORMATION. There has been a proposal to perhaps also add a request to query AUTOPILOT components, which are the only components we're allowing to "own a camera".

1. How is adding a capability bit and querying individual components for their capabilities better than these two approaches?
2. Which of these approaches is best?

My thinking on this is that a protocol bit makes sense if you want to be able to recognise a component "is a thing" or "supports a thing" and you can't infer that from the type. If you can infer it from the type then having a multi-step process like this is flawed because (other than being multi step) you can already know it from the type, and you know it better, since people are more likely to get the type of a thing right than the id of a protocol bit (lots of historical evidence).

So can we infer from the type? Yes we can. In the devcall we made it a "rule" that we only allow camera components and cameras attached to autopilot are allowed - not cameras attached to companions (those have to be camera components).
So the only place you'd have to use this id would be on the autopilot. In that case, why not just also send the request command to the autopilot as well?

IMO the best approach for both gimbal and camera would be the camera  process + requesting cameras from the autopilot as well. This is more robust that the gimbal process because that only discusses a single discovery phase on first connection, and because if you miss a gimbal message then you don't know you missed it - because you aren't checking types, only messages. Hope that makes sense.

Thoughts?

julianoes:
@hamishwillee I can see that the flag is a duplication from the component ID. However, the component ID shouldn't matter in theory.

Now you say just request camera information from cameras AND the autopilot. I think that would work, but I'm not sure how that works for gimbal managers. Gimbal managers don't have component IDs, only gimbal devices do. So we would have to add component IDs for gimbal managers, right?

hamishwillee:
> @hamishwillee I can see that the flag is a duplication from the component ID. However, the component ID shouldn't matter in theory.

No, it's a duplication of the MAV_TYPE, not the component id. I don't think I mentioned component id at all. Doesn't change your argument :-)

> Now you say just request camera information from cameras AND the autopilot. 

Yes. To be very clear, currently the camera spec says [request CAMERA_INFORMATION from camera components based on type](https://mavlink.io/en/services/camera.html#camera-discoveryconnection).
My proposal is to extend this to also query the autopilot (based on its type). 
It's a minimal change that should work with existing setups and adds just one more request to the process.
This is a lot more lightweight than checking the protocol bit and then sending the message, since we know that everything that is a MAV_TYPE _must_ support CAMERA_INFORMATION.

We could add a bit for supporting cameras to the **autopilot** and only query the autopilot if it has attached cameras. In that case you'd be adding the bit to [MAV_PROTOCOL_CAPABILITY](https://mavlink.io/en/messages/common.html#MAV_PROTOCOL_CAPABILITY) and it would be something like `MAV_PROTOCOL_CAPABILITY_COMPONENT_ATTACHED_CAMERA`.
Don't think it is needed.

> ... I'm not sure how that works for gimbal managers. Gimbal managers don't have component IDs, only gimbal devices do. So we would have to add component IDs for gimbal managers, right?

You're right, it wouldn't work since it relies being able to know the type.

The current process for gimbals is that:
- a GCS should broadcast a request for `GIMBAL_MANAGER_INFORMATION`
- a gimbal manager should broadcast a request for `GIMBAL_DEVICE_INFORMATION`

IN both cases all systems that support gimbal managers emit the `GIMBAL_MANAGER_INFORMATION`.
In most systems this will be things of MAV_TYPE of gimbal, and the autopilot, though I guess we don't know that for sure.

You're proposing instead that the GCS check for this protocol bit on all components, then only send GIMBAL_MANAGER_INFORMATION request targeted to those things that say they support it?

Right now nothing will set this bit. So GCS should just assume there is no gimbal manager? How do we manage the crossover time?

julianoes:
> You're proposing instead that the GCS check for this protocol bit on all components, then only send GIMBAL_MANAGER_INFORMATION request targeted to those things that say they support it?

My goal was to avoid unnecessary traffic to all components. But it sounds like it's not easy to avoid.

> Right now nothing will set this bit.

You are indeed right that this is not backwards compatible so we're already stuck with having to send out requests.

hamishwillee:
> My goal was to avoid unnecessary traffic to all components. But it sounds like it's not easy to avoid.

In terms of traffic  you're basically comparing:
- one broadcast command to get gimbal manager info that will result in multiple unnecessary ACKS back from non-gimbal components, vs 
- multiple requests to each gimbal manager (once you have established which are manager) with multiple ACKs back but those should only be valid ones.

I don't "count" the messages to discover the bit, because that discovery should really be done anyway.
  
So assuming a system with 1 gimbal manager and 5 other mavlink components we're comparing 6 messages (1 cmd + 5 Ack) with 2 messages (1 out, 1 in).. Of course the numbers change with more managers or more components, but I suspect that this will definitely reduce the traffic for gimbals.

The question for me though is whether it is worth it? This is a command, so you're only sending it once.
So this has more to do with "do we always want to be sure we discovered all gimbal managers + the complexity" vs simplicity of current approach.

> You are indeed right that this is not backwards compatible so we're already stuck with having to send out requests.

Any ideas on how to migrate from one to the other?

One last option is to _specify_ that a gimbal can only be mounted on a component that identifies as a gimbal device, autopilot, companion computer, (and maybe something that identifies as a gimbal manager component), and just query those components. I suspect that this will work just as well. Traffic a bit higher, but still low, and you get the reliability of individual commands.

How do you want to proceed this?

auturgy:
@rmackay9 your review/input would be appreciated here

rmackay9:
I haven't read everything thoroughly above but I two things to keep in mind:

- If we go with the idea of the GCS broadcasting a request for GIMBAL_MANAGER_INFORMATION then an AP autopilot could reply with multiple messages and each would have its "gimbal_device_id" filled in with a unique number (from 1 to 6)
- when the AP autopilot is in control of the camera gimbals it won't send any messages with a component id of CAMERA or GIMBAL

message_definitions/v1.0/common.xml
```diff
@@ -3230,6 +3230,14 @@
...
         </description>
       </entry>
+      <entry value="262144" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_CAMERA">
+        <description>This component implements/is a camera. This means the CAMERA_INFORMATION, and other messages can be requested.
+        </description>
+      </entry>
+      <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER">
+        <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
+        </description>
+      </entry>
```
hamishwillee, line 3240:
```suggestion
      <entry value="262144" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER">
        <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
        </description>
      </entry>
```

hamishwillee (approved):
We had a chat about this in the dev call. 

We don't need the _camera_ bit but @julianoes thinks it is not to late for the gimbal bit.

The proposal is:

1. Monitor for new heartbeats
2. If `MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER` set, then request [GIMBAL_MANAGER_INFORMATION](https://mavlink.io/en/messages/common.html#GIMBAL_MANAGER_INFORMATION)
3. Always request autopilot for `MAV_PROTOCOL_CAPABILITY_COMPONENT_IMPLEMENTS_GIMBAL_MANAGER` (at least for current versions). This is needed because there are versions in the wild we can't update.

This would break anything in the wild that is not an autopilot that implements the gimbal manager protocol. Note that we _think_ this might affect one gimbal. In theory it could affect a gimbal manager attached to a companion computer but we don't know of any such setups now.

While it would be good to let them know about this, we think a break at this point is worth the cost of such a break.

@julianoes @auturgy @rmackay9 I support this. Any arguments against me merging?

rmackay9:
Hi, I don't understand fully but I'm happy to trust your judgement and fix anything that needs fixing on the AP side should that arise

hamishwillee:
OK, merging. Everyone has had a chance to comment.

PR #2145, opened 2024-08-24
common: add camera-thermal-status

rmackay9:
This adds a new message to allow sending a thermal camera's min and max temperatures (including the location on the image) to the ground station in real time.  This is useful for search & rescue and firefighting applications because the ground station can overlay a rectangle over the hotspot along with its temperature

I've tried to make it consistent with other CAMERA messages including the tracking message in the hopes that that makes it easier to merge.  I've also included the camera_device_id field consistent with PR https://github.com/mavlink/mavlink/pull/2142

Over on the AP side we've recently been improving our support for thermal cameras including the Siyi ZT6, ZT30 and the Topotek gimbals.  This has been tested on real-hardware using a Siyi ZT6

The equivalent AP PR is here https://github.com/ArduPilot/mavlink/pull/369

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3807,7 +3807,7 @@
         <description>Camera supports tracking geo status (CAMERA_TRACKING_GEO_STATUS).</description>
       </entry>
       <entry value="4096" name="CAMERA_CAP_FLAGS_HAS_THERMAL_RANGE">
-        <description>Camera supports absolute thermal range (request CAMERA_THERMAL_RANGE with MAV_CMD_REQUEST_MESSAGE).</description>
+        <description>Camera supports absolute thermal range (request CAMERA_THERMAL_RANGE with MAV_CMD_REQUEST_MESSAGE) (WIP).</description>
       </entry>
     </enum>
     <enum name="VIDEO_STREAM_STATUS_FLAGS" bitmask="true">
@@ -3819,7 +3819,7 @@
         <description>Stream is thermal imaging</description>
       </entry>
       <entry value="4" name="VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED">
-        <description>Stream can report absolute thermal range (see CAMERA_THERMAL_RANGE)</description>
+        <description>Stream can report absolute thermal range (see CAMERA_THERMAL_RANGE). (WIP).</description>
       </entry>
     </enum>
     <enum name="VIDEO_STREAM_TYPE">
@@ -7109,6 +7109,8 @@
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
     <message id="277" name="CAMERA_THERMAL_RANGE">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="stream_id" invalid="0" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7113,7 +7113,7 @@
       <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
       <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
-      <field type="uint8_t" name="stream_id" invalid="0" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="stream_id" minValue="1" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
       <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
       <field type="float" name="max" units="degC">Temperature max.</field>
       <field type="float" name="max_point_x" invalid="NaN">Temperature max point x value (normalized 0..1, 0 is left, 1 is right), NAN if unknown.</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7109,7 +7109,7 @@
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
     <message id="277" name="CAMERA_THERMAL_RANGE">
-      <description>Camera absolute thermal range.  This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream.  Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
+      <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="stream_id" invalid="0" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
       <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3806,6 +3806,9 @@
       <entry value="2048" name="CAMERA_CAP_FLAGS_HAS_TRACKING_GEO_STATUS">
         <description>Camera supports tracking geo status (CAMERA_TRACKING_GEO_STATUS).</description>
       </entry>
+      <entry value="4096" name="CAMERA_CAP_FLAGS_HAS_THERMAL_RANGE">
+        <description>Camera supports absolute thermal range (request CAMERA_THERMAL_RANGE with MAV_CMD_REQUEST_MESSAGE).</description>
+      </entry>
     </enum>
     <enum name="VIDEO_STREAM_STATUS_FLAGS" bitmask="true">
       <description>Stream status flags (Bitmap)</description>
@@ -3814,6 +3817,9 @@
       </entry>
       <entry value="2" name="VIDEO_STREAM_STATUS_FLAGS_THERMAL">
         <description>Stream is thermal imaging</description>
+      </entry>
+      <entry value="4" name="VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED">
+        <description>Stream can report absolute thermal range (see CAMERA_THERMAL_RANGE)</description>
       </entry>
     </enum>
     <enum name="VIDEO_STREAM_TYPE">
@@ -7101,6 +7107,18 @@
       <field type="float" name="dist" units="m" invalid="NaN">Distance between camera and tracked object. NAN if unknown</field>
       <field type="float" name="hdg" units="rad" invalid="NaN">Heading in radians, in NED. NAN if unknown</field>
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
+    </message>
+    <message id="277" name="CAMERA_THERMAL_RANGE">
+      <description>Camera absolute thermal range.  This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream.  Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="stream_id" invalid="0" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
+      <field type="float" name="max" units="degC">Temperature max.</field>
+      <field type="float" name="max_point_x" invalid="NaN">Temperature max point x value (normalized 0..1, 0 is left, 1 is right), NAN if unknown.</field>
+      <field type="float" name="max_point_y" invalid="NaN">Temperature max point y value (normalized 0..1, 0 is top, 1 is bottom), NAN if unknown.</field>
+      <field type="float" name="min" units="degC">Temperature min.</field>
+      <field type="float" name="min_point_x" invalid="NaN">Temperature min point x value (normalized 0..1, 0 is left, 1 is right), NAN if unknown.</field>
+      <field type="float" name="min_point_y" invalid="NaN">Temperature min point y value (normalized 0..1, 0 is top, 1 is bottom), NAN if unknown.</field>
     </message>
     <message id="280" name="GIMBAL_MANAGER_INFORMATION">
       <description>Information about a high level gimbal manager. This message should be requested by a ground station using MAV_CMD_REQUEST_MESSAGE.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7114,7 +7114,7 @@
       <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="stream_id" minValue="1" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
-      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
+      <field type="uint8_t" name="camera_device_id" default="0" minValue="0" maxValue="6">Camera id of a non-MAVLink camera attached to an autopilot (1-6).  0 if the component is a MAVLink camera (with its own component id).</field>
       <field type="float" name="max" units="degC">Temperature max.</field>
       <field type="float" name="max_point_x" invalid="NaN">Temperature max point x value (normalized 0..1, 0 is left, 1 is right), NAN if unknown.</field>
       <field type="float" name="max_point_y" invalid="NaN">Temperature max point y value (normalized 0..1, 0 is top, 1 is bottom), NAN if unknown.</field>


nexton-winjeel:
Couple of comments:
* As it stands, the `CAMERA_THERMAL_STATUS_FLAGS` field adds no value.
    * It can't be used to signal whether to request this message if it is _inside_ this message. Would be better to extend [CAMERA_THERMAL_STATUS_FLAGS](https://mavlink.io/en/messages/common.html#VIDEO_STREAM_STATUS_FLAGS) instead.
    * The "Unavailable" case can be signalled by setting the `min` and `max` fields to `NaN`.
* The `*_point_*` fields are under-specified (although this criticism applies to other messages like [CAMERA_TRACKING_IMAGE_STATUS](https://mavlink.io/en/messages/common.html#CAMERA_TRACKING_IMAGE_STATUS)):
    * Are the points reported points in relation to the IR sensor? Or the output video stream?
        * To make life easy on a GCS everything should be reported in terms of video stream, so probably need to include a `stream_id` field.
    * The specific use cases that cause confusion are where:
        * the sensor size is not the same as the video stream size; and/or
        * the centre of the sensor is not the same as the centre of the video stream.
    * Examples of the above include:
        1. IR cameras with digital zoom and e-stab;
        2. EO/IR systems with different sensor sizes that do blending, i.e. an EO image with a smaller IR overlay at the centre of the image.

rmackay9:
Hi @nexton-winjeel, thanks for the feedback.

I'm totally happy of course to remove the camera-thermal-status-flags especially if others agree as well

Re the points fields, from an autopilot point of view they are related to an IR sensor so it doesn't actually matter if the video is streaming or not.  I don't think that AP will be able to fill in the stream_id unless it makes an assumption that there's a one-to-one relationship between camera lenses and streams.  I can add it if others want but I'm not sure we will be able to actually fill it in.

Re scaling, the points are in the range of 0 to 1 so the resolution of the pictures or video stream don't matter I think.

Thanks again!  Very happy to have the discussion!

nexton-winjeel:
@rmackay9: 
> Re the points fields, from an autopilot point of view they are related to an IR sensor so it doesn't actually matter if the video is streaming or not

It does if you want to allow the GCS to "overlay a rectangle over the hotspot along with its temperature" (from the description).

> Re scaling, the points are in the range of 0 to 1 so the resolution of the pictures or video stream don't matter I think.

The _resolution_ doesn't matter, but the image extents do. I'll see if I can explain the issue a bit more clearly.

In a camera with digital zoom, you get the following:
```
+-----------------------------+
| Sensor Image                |
|                             |
|     +-----------------+     |
|     | Display Image   |     |
|     |                 |     |
|     |                 |     |
|     +-----------------+     |
|                             |
|                             |
+-----------------------------+
```
In this case, the centre of the Sensor Image is the same as the centre of the Display image.
* If the `x` and `y` positions are reported in terms of Display Image pixels, the GCS can easily draw the hotspot in the correct location.
* If the `x` and `y` positions are reported in terms of Sensor Image pixels, the GCS needs to calculate the scale of the Display Image in relation to the Sensor Image based on the reported zoom.
    * However, it only needs to calculate in regards to _digital_ zoom. If the camera has both physical and digital zoom, then this can't be calculated as MAVLink only reports a single, aggregate zoom level.

In an camera with digital zoom _and_ electronic stabilisation, the Display Image can move around within the Sensor Image, such as in the following:
```
+-----------------------------+
| Sensor Image                |
|                             |
|                             |
|                             |
|  +-----------------+        |
|  | Display Image   |        |
|  |                 |        |
|  |                 |        |
|  +-----------------+        |
+-----------------------------+
```
In this case, the centre of the Sensor Image is _not_ the same as the centre of the Display image.
* If the `x` and `y` positions are reported in terms of Display Image pixels, the GCS can easily draw the hotspot in the correct location.
* If the `x` and `y` positions are reported in terms of Sensor Image pixels, the GCS needs to calculate:
    * the scale of the Display Image in relation to the Sensor Image based on the reported zoom.
    * the offset of the Display image in relation to the Sensor Image. This is not reported in MAVLink.

hamishwillee:
Before I look at the messages, I would like to confirm how this should be used.

What I understand is that you want to be able to show the max and min temperature points on a live video stream. Here is a snapshot of what that might look like. How is this useful? I mean the minimal temperature is anywhere that is not hot - it's background. The colouring of a typical heatmap already shows you the hot spots - how does knowing that one particular point is the hottest help, and wouldn't that be obvious in most cases  from what you see in the image?

image

- Do cameras generally supply this information?
- Would it be better for a GCS to work this out from the video stream they are getting?
- Are we ever going to want more than max and min? ie. all points over X degrees, or in top 10%

W.r.t. the message I tend to agree that it would be better to add enable detection to another message, though why not https://mavlink.io/en/messages/common.html#CAMERA_CAP_FLAGS so we can get it in https://mavlink.io/en/messages/common.html#CAMERA_INFORMATION ?

W.r.t. what the points is relevant to, I _think_ @nexton-winjeel has certainly found a case where there is ambiguity. My initial reaction to this was that you need a stream ID. However we don't really want to send this for every stream. Unlike https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE , https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL doesn't allow you to specify a particular instance of a message - i.e. in this case for a particular stream id, and I don't particularly want to shove this information into VIDEO_STREAM_STATUS. So more thinking required.

message_definitions/v1.0/common.xml
```diff
@@ -7102,6 +7111,18 @@
       <field type="float" name="hdg" units="rad" invalid="NaN">Heading in radians, in NED. NAN if unknown</field>
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
+    <message id="277" name="CAMERA_THERMAL_STATUS">
+      <description>Camera thermal status, sent while thermal camera is capturing absolute temperatures. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval.</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
```
hamishwillee, line 7117: In https://github.com/mavlink/mavlink/pull/2142/files#r1717794836 for messages I proposed we do this as below. This assumes we're only allowing attached cameras to the autopilot to be proxied. Companions etc need to implement them as MAVLink components.

```suggestion
      <field type="uint8_t" name="camera_device_id" default="0" minValue="0" maxValue="6">Camera id of a non-MAVLink camera attached to an autopilot (1-6).  0 if the component is a MAVLink camera (with its own component id).</field>
```

rmackay9:
Hi @hamishwillee,

The key enhancement that this exposes is the absolute temperature reading.  Most cameras by default will essentially adjust/scale the brightness of the picture and/or live video stream so that the warmest object is bright and the coldest object is dark (depending upon the palette of course).  This is useful for humans trying to understand what they're looking at and so they don't bump into things but it's almost useless for application where we're searching for something of a known temperature (e.g. a human or a fire).

With this output to the GCS the user doesn't need to even look at the video stream most of the time.  They can just watch a graph of the live temperature and wait for it to spike.  I've included a picture of this in the AP PR that pulls this info from the Siyi ZT6. 
https://github.com/ArduPilot/ardupilot/pull/27915

Having the min/max also theoretically makes it possible for the GCS to calculate the temperature of every pixel so the user can hover over a pixel and see its temperature.  This has not been attempted yet though.

Re cameras that support this, the Siyi ZT6 and ZT30 support this.  I expect that the [Xacti](https://ardupilot.org/copter/docs/common-xacti-gimbal.html) will shortly and then others will follow.  It is somewhat amazing that absolute temperatures have been ignored for this long considering how important they are.

hamishwillee:
@rmackay9 Thank you very much for the explanation. 

It does seems reasonable that you might want to be able to get absolute temperatures from an image (for a range of reasons), and this is probably the lowest bandwidth cost way of doing it - since the processing would be pushed to the GCS. If we wanted to support getting per-pixel data I'd probably include the MAVFTP url of the pixel map in the message so getting that would be a choice.

For this particular use case you probably only need the maximum value, since you can compare this to the minimum temperature of whatever you are looking for and trigger a warning/alarm. No need to watch for random spikes. In fact for this particular use case you don't even "really" need the stream id or to know the mapping to your image, since it's mostly a hint that you need to look. 

But that's as an aside.

Looking at the specific message with this in mind:
1. How do we know when to stream it and when to stop streaming it? 
   - There are a bunch of ways we can do this, but since this is (depending on your response to @nexton-winjeel comments) dependent on stream, then perhaps a  new flag in [VIDEO_STREAM_STATUS .flags](https://mavlink.io/en/messages/common.html#VIDEO_STREAM_STATUS) - like `VIDEO_STREAM_STATUS_FLAGS_THERMAL_TEMP_RANGE`?
   - Benefit is that we only stream this while we have a video stream AND flag is set. If stream dies or gets turned off we can request new rate.
2. `CAMERA_THERMAL_STATUS` seems very general for a specific message. Maybe `STREAM_THERMAL_RANGE`?
3. Do we need to match the timestamp to a particular part of the video? Probably not. SHould maybe note that this isn't provided for that reason.
4. Some more description about this absolute temp use case.
5. Do we need the camera device ID if this is associated with a stream ID?

rmackay9:
Hi @hamishwillee,

Great, thanks for the feedback.

1. I think we could add this to the camera capabilities rather than the video stream capabilities.  It just not clear to me at the moment how we get the stream-id.  It's very much a capability of the camera (or camera lens) just like zoom or focus.
2. re naming, how about CAMERA_THERMAL_RANGE?  Personally I think the first word is the most important and it's a CAMERA related message.  "STREAM" is more of an adjective describing how often it is sent so I don't think this word should be first.  It could potentially be VIDEO_STREAM_THERMAL_RANGE but still, from my point of view, it's a camera feature
3. I think we should send the time if we have it.  It's probably not critical but aren't we always sending a time these days?
4. sure
5. I think it's associated with a camera device id but I really don't understand where stream ids come from.  maybe a stream is a one-to-one mapping with camera lens

dakejahl:
I'm a little confused why this would go in mavlink at all? If you are finding hot spots in an image then I would assume you're likely already using something like OpenCV to do it. Why not just add the overlay directly to the video stream before publishing the stream?

hamishwillee:
@rmackay9

Videos streams belong to a particular camera and are discovered by querying a camera component. There is no way to know the camera directly from the video stream message, but you can store that relationship as part of the discovery ([how video streams work](https://mavlink.io/en/services/camera.html#video_streaming)). 

The streams are allocated by the camera. The numbers are arbitrary. But they are important, because only specific streams will be being displayed by the GCS (only one currently).

Whether we use streams or camera ids, we would still need to extend the request interval command to support setting the streaming for that message for a particular instance.

 1. My understanding of @nexton-winjeel point is that we need to resolve this message to video stream ids in order to plot on the video feed, and for the more generic case of working out the absolute temperatures of all the pixels.
   It is not sufficient to have this solely associated with a camera, because a camera can have multiple streams that come from different sensors, zooms, whatever.
   
I'm only opposed to using `STATUS` for a very specific kind of status. I think `CAMERA_THERMAL_RANGE`  might be OK, but I would _prefer_ VIDEO_STREAM_THERMAL_RANGE because this range is specific to a particular stream. If you change to a stream with a different zoom that message is no longer "true".


YOu need `min` to get the absolute range for the image. So yes, we need it. Was just commenting that it wasn't essential for your use case.

hamishwillee:
> I'm a little confused why this would go in mavlink at all? If you are finding hot spots in an image then I would assume you're likely already using something like OpenCV to do it. Why not just add the overlay directly to the video stream before publishing the stream? I'm not trying to criticize, I am genuinely curious! Because this way seems like way more work.

@dakejahl Randy answered this above in https://github.com/mavlink/mavlink/pull/2145#issuecomment-2313862921

This is an easy way to find that there are hot spots without having to analyse or watch the image.

The hotspots in a thermal image are all relative - so you could have a bunch of hot spots, but none of them help you know what is likely to be a human or a fire so you'd have to check all of them. With this feature you can calibrate your range for a search by knowing the absolute values that those hot spots map to.

dakejahl:
> so the user can hover over a pixel and see its temperature

potential interactive overlay :+1: 

> They can just watch a graph of the live temperature and wait for it to spike

viewing mavlink data as live plot to see delta temperature spike :+1: 

That makes sense from an ease of use perspective, a plot and a spike is much more helpful. Thanks @hamishwillee

rmackay9:
So just to clarify the changes I should make:

1. rename CAMERA_THERMAL_STATUS to CAMERA_THERMAL_RANGE
2. remove the status field and CAMERA_THERMAL_STATUS_FLAGS enum
3. update the camera_device_id as suggested by @hamishwillee above
4. add a new value to the camera capabilities enum.  Perhaps we will call it CAMERA_CAP_FLAGS_HAS_THERMAL_RANGE

anything I've missed?

hamishwillee:
Let's see what @nexton-winjeel says. But we can't design this around the Siyi ZT6  - if it is possible for another camera to come along with a different mapping for the displayed stream we know this would break. And it is.

@julianoes Any thoughts on this?

FrauBluher:
I would rather see video telemetry in a muxed video stream so that presentation timestamp and telemetry timestamp can be easily synced. KLV data in an mpegts stream is one such application of telemetry in stream that could support this. That said, I can appreciate this information being available as mavlink.

Adding a pts field or some kind of field which can be used for frame synching would be helpful IMO.

An option:
https://en.m.wikipedia.org/wiki/Presentation_timestamp

hamishwillee:
@FrauBluher Something like this would be great, in particular if the kinds of time synced metadata of interest became bigger.

My concern is that:
- the approach would have to be standardized for the camera API and work with the different video stream types
- We'd have to standardize what data was sent and how to encode decode
- GCS and cameras would have to be reasonably able to adopt/send this format.

All of that is possible, but it would require broad buy-in from multiple stakeholders: I don't think I could get support now to mandate that, mostly because we're only talking about one small bit of data that only needs to be weakly associated with the video frames. If someone came up with an end to end proposal that showed how this would work and costed the effort then maybe.

By contrast this data can quite easily be sent right now over MAVLink. It may feel short sighted, but it is pragmatic

Given that, is there any other frame related metadata we think we'd like right now? IF so we should consider if it should be part of this message or another. The more data that needs to be sent and the more tightly it needs to be tied to frames, the bigger the pull for a more generic solution.

hamishwillee:
@rmackay9 If this has to be associated with video, and the geometry of the video can be independent of the camera sensor, then this cannot be done generically if the data is associated only with a camera - it has to be the stream. 
For this particular camera you could always send the same data for all streams.
This is how the camera API works - video data is associated with video streams.

Messages for video will have to have camera_device_Id and stream_id or there is no way to resolve streams from autopilot-attached cameras (so also add camera device id to  https://mavlink.io/en/messages/common.html#VIDEO_STREAM_INFORMATION. )

Further, we'll have to update https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL so that param3-6 can be used to identify the specific camera and stream that we want to have sent out - because otherwise there is no way an autopilot to know which camera and/or stream id you might be interested in getting a stream for. 

For discovery I think add `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` to [VIDEO_STREAM_STATUS .flags](https://mavlink.io/en/messages/common.html#VIDEO_STREAM_STATUS).

So:
1. Cameras would always set `VIDEO_STREAM_STATUS.flags = VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` for any stream that is providing thermal range data. In you case, the camera would iterate the streams when thermal data is being captured and set this flag for all of them.
2. GCS would monitor this flag on either all streams or just the current displayed stream and request interval for the stream/camera id to stream/stop streaming `CAMERA_THERMAL_RANGE`
3. I suspect your camera will only have this one stream, so the GCS is only monitoring one thing.

Message would look like this:

```xml
    <message id="277" name="CAMERA_THERMAL_RANGE">
      <description>Camera video stream absolute thermal range.
        This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only send it for the current active stream.
        Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).
      </description>
      <field type="uint8_t" name="stream_id">Video Stream ID (1 for first, 2 for second, etc.). Messages with same value are from the same source (instance).</field>
      <field type="uint8_t" name="camera_device_id" default="0" minValue="0" maxValue="6">Camera id of a non-MAVLink camera attached to an autopilot (1-6).  0 if the component is a MAVLink camera (with its own component id).</field>
      <field type="uint8_t" name="status" enum="CAMERA_THERMAL_STATUS_FLAGS">Thermal status.</field>
      <field type="float" name="max" units="degC">Temperature max.</field>
      <field type="float" name="max_point_x" invalid="NaN">Temperature max point x value (normalized 0..1, 0 is left, 1 is right), NAN if unknown.</field>
      <field type="float" name="max_point_y" invalid="NaN">Temperature max point y value (normalized 0..1, 0 is top, 1 is bottom), NAN if unknown.</field>
      <field type="float" name="min" units="degC">Temperature min.</field>
      <field type="float" name="min_point_x" invalid="NaN">Temperature min point x value (normalized 0..1, 0 is left, 1 is right), NAN if unknown.</field>
      <field type="float" name="min_point_y" invalid="NaN">Temperature min point y value (normalized 0..1, 0 is top, 1 is bottom), NAN if unknown.</field>
    </message>
```

That's what I think is consistent with the architecture of the API, but it would be good to get @julianoes and @Davidsastresas and @nexton-winjeel to confirm I'm not missing anything.

And we should still consider what other metadata we might want before finalizing anything.

Sorry this is more complicated than you want: on the plus side, it still works for your future cameras.

hamishwillee:
And this should go in development.xml.

rmackay9:
Hi @hamishwillee,

Thanks for the help and suggestions.  That message format looks good to me except that we can drop the "status" field I think because the individual fields can be set to NaN (this was @nexton-winjeel's suggestion).

I'll update the PR shortly, thanks!

P.S. I can't think of other things to add at this moment.  We will have at least one more camera (the Xacti) appear with absolute temperature support before this goes live.  I'll also ping some other camera manufacturers to ask if they are planning absolute thermal support.

hamishwillee:
> Thanks for the help and suggestions. That message format looks good to me except that we can drop the "status" field I think because the individual fields can be set to NaN (this was @nexton-winjeel's suggestion).

Doh. Of course. Removed in version above. This shouldn't be streamed if it isn't valid, but of course it might be, so setting values to NaN makes sense. I also allow the ids values to be set to invalid.

I might add more context in the description as a suggestion after PR updated.

One thing that occurs to me is that it will be impossible to fully timesync this to a video - you'll only ever be approximate at the max-min pixel when you overlay. Doesn't matter maybe here, but some metadata it would, which is why the comment https://github.com/mavlink/mavlink/pull/2145#issuecomment-2314928993.
I don't know much about videos, but is there any external way to provide a sync to to the stream so you could more accurately determine when metadata should be applied [I might be overthinking this]

nexton-winjeel:
Sorry for the delay in getting back to this one.

@hamishwillee (https://github.com/mavlink/mavlink/pull/2145#issuecomment-2313790290):
> * Do cameras generally supply this information?

Thermal cameras _usually_ supply this information, but how they supply it differs (might be a published message, might be a register that can be read, might be embedded in the image metadata).

> * Would it be better for a GCS to work this out from the video stream they are getting?

The GCS can't work it out from the imagery alone. The pixel values in the image will provide _relative_ temperature across the image, but there is no way to determine _absolute_ temperatures without a reference point, either from:
* this message; or
* embedded metadata (as @FrauBluher mentioned)

> * Are we ever going to want more than max and min? ie. all points over X degrees, or in top 10%

Probably. But that should be in a separate message. This is sufficient for now.

@rmackay9 (https://github.com/mavlink/mavlink/pull/2145#issuecomment-2314309090):
> 1. I think we could add this to the camera capabilities rather than the video stream capabilities. It just not clear to me at the moment how we get the stream-id. It's very much a capability of the camera (or camera lens) just like zoom or focus.
> 2. re naming, how about CAMERA_THERMAL_RANGE? Personally I think the first word is the most important and it's a CAMERA related message. "STREAM" is more of an adjective describing how often it is sent so I don't think this word should be first. It could potentially be VIDEO_STREAM_THERMAL_RANGE but still, from my point of view, it's a camera feature

As long as you include the stream ID, I don't care what you call it. But it isn't usable without a stream ID. (And calling it `CAMERA_THERMAL_RANGE` matches `CAMERA_TRACKING_IMAGE_STATUS` and friends).

As to why I need the stream ID: I've got a camera system on my desk that has a 4K EO sensor and a 720p IR sensor. It can _simultaneously_ output:
1. The 4K EO imagery;
2. The 720p IR imagery;
3. A 1080p blended image consisting of the IR image overlaid on the centre of the EO imagery. 

The min and max temperatures and points are different for each of these streams.

@hamishwillee (https://github.com/mavlink/mavlink/pull/2145#issuecomment-2314346694):
> My understanding of @nexton-winjeel point is that we need to resolve this message to video stream ids in order to plot on the video feed, and for the more generic case of working out the absolute temperatures of all the pixels.
> It is not sufficient to have this solely associated with a camera, because a camera can have multiple streams that come from different sensors, zooms, whatever.

Yes. See above.

@hamishwillee (https://github.com/mavlink/mavlink/pull/2145#issuecomment-2316375570): 
> For discovery I think add VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED to VIDEO_STREAM_STATUS .flags.

:+1: 

> Message would look like this:

:+1: 

But I think this needs to change:
* "but a GCS may choose to only send it for the current active stream."
to:
* "but a camera may choose to only send it for the current active stream."

rmackay9:
Thanks for all this feedback.  I've updated the PR now with these changes:

1. renamed to CAMERA_THERMAL_RANGE
2. removed status field
3. added stream id field.  Note that I've added an "invalid=0" argument
4. added a camera capabilities flag to CAMERA_CAP_FLAGS called CAMERA_CAP_FLAGAS_HAS_THERMAL_RANGE
5. added a video stream status flag called VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED

I have left this in common.xml instead of moving to development.xml but I can move it if that's what people prefer.  I think we've had enough discussion that it's "graduated" and can be in common.xml.

I've lightly tested this on the AP flight code and it seems to work.

rmackay9:
@hamishwillee

No huge pressure of course but what do you think?  I'm a little more keen than normal on the timing because I've got a search and rescue competition coming up in just over a month and I plan to use this

message_definitions/v1.0/common.xml
```diff
@@ -7102,6 +7108,18 @@
       <field type="float" name="hdg" units="rad" invalid="NaN">Heading in radians, in NED. NAN if unknown</field>
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
+    <message id="277" name="CAMERA_THERMAL_RANGE">
+      <description>Camera absolute thermal range.  This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream.  Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
```
hamishwillee, line 7112: Just convention, single space.
```suggestion
      <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -7102,6 +7108,20 @@
       <field type="float" name="hdg" units="rad" invalid="NaN">Heading in radians, in NED. NAN if unknown</field>
       <field type="float" name="hdg_acc" units="rad" invalid="NaN">Accuracy of heading, in NED. NAN if unknown</field>
     </message>
+    <message id="277" name="CAMERA_THERMAL_RANGE">
+      <wip/>
+      <!-- This message is work-in-progress and it can therefore change. It should NOT be used in stable production environments. -->
+      <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="stream_id" invalid="0" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
```
hamishwillee, line 7116: > added stream id field. Note that I've added an "invalid=0" argument

This is good - all streams must have an ID. 

I was thinking we could allow this for cameras where you _know_ that the information is the same for all streams, OR if you choose not to publish the positions (just max/min). But this is cleaner.

hamishwillee, line 7116: > added stream id field. Note that I've added an "invalid=0" argument

When would you this message without an id? This should either be set to a min value or we need to be explicit about what such a value would mean.

Is the intent that you allow a camera to emit this if it knows that the information is the same for all streams? you could also use this for the case where you don't care about the video stream and just want to send max-min for the camera values. IN this case you would need to NaN any values that aren't true for all streams (i.e. set position values to Nan, but you can still get the max and min).

So it would either be:

```suggestion
      <field type="uint8_t" name="stream_id" minValue="1" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
```

Or

```suggestion
      <field type="uint8_t" name="stream_id" invalid="0" instance="true">Video Stream ID (1 for first, 2 for second, etc.). 0 indicates that this the information is accurate for all stream ids associated with the camera.</field>
```

hamishwillee:
@rmackay9 
1. I have added #2148 as well because otherwise the status can't be requested.
2. Generally prefer in development.xml. We have so many WIP things that never go into autopilots. But on assumption that this will, I have tagged as WIP. Let me know when you have an implementation in a release and we can remove it. 
3. See https://github.com/mavlink/mavlink/pull/2145/files#r1742872340

>> 4. added a camera capabilities flag to CAMERA_CAP_FLAGS called CAMERA_CAP_FLAGAS_HAS_THERMAL_RANGE
>> 5. added a video stream status flag called VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED
>
> I'm unsure if we really need (5) above but it's there.

I think having both makes sense, the one you actually _need_ is actually `5`. 

@julianoes @peterbarker @nexton-winjeel This looks good to me. Any objection to merging as WIP? @rmackay9 Another benefit of development.xml is that I would merge it there straight away, but for common.xml we need broader acceptance.

hamishwillee:
@nexton-winjeel 

> But I think this needs to change:
> "but a GCS may choose to only send it for the current active stream."
> to:
> "but a camera may choose to only send it for the current active stream."

Randy changed this appropriately: " a GCS may choose to only request it for the current active stream."

rmackay9:
Hi @hamishwillee, thanks again for your help.

As discussed separately, here's a link to the AP flight code change that will go in as soon as we can get this mavlink changed merged https://github.com/ArduPilot/ardupilot/pull/27915

nexton-winjeel:
Looks good to me.

message_definitions/v1.0/common.xml
```diff
@@ -7102,6 +7108,20 @@
...
+      <description>Camera absolute thermal range. This can be streamed when the associated `VIDEO_STREAM_STATUS.flag` bit `VIDEO_STREAM_STATUS_FLAGS_THERMAL_RANGE_ENABLED` is set, but a GCS may choose to only request it for the current active stream. Use MAV_CMD_SET_MESSAGE_INTERVAL to define message interval (param3 indicates the stream id of the current camera, or 0 for all streams, param4 indicates the target camera_device_id for autopilot-attached cameras or 0 for MAVLink cameras).</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="stream_id" minValue="1" instance="true">Video Stream ID (1 for first, 2 for second, etc.)</field>
+      <field type="uint8_t" name="camera_device_id" invalid="0">Camera id of a camera associated with this component. This is the component id of a proxied MAVLink camera, or 1-6 for a non-MAVLink camera attached to the component. Use 0 if the component is a camera (not something else providing access to a camera).</field>
```
hamishwillee, line 7117: New form we're using now proxy cameras can _only_ be on autopilot, as per https://github.com/mavlink/mavlink/pull/2142
```suggestion
      <field type="uint8_t" name="camera_device_id" default="0" minValue="0" maxValue="6">Camera id of a non-MAVLink camera attached to an autopilot (1-6).  0 if the component is a MAVLink camera (with its own component id).</field>
```

julianoes (approved):
LGTM

hamishwillee:
Thanks all for your input and help.

rmackay9:
We've added the AP side implementation now so I think we could remove the WIP label https://github.com/ArduPilot/ardupilot/pull/27915.  No huge pressure of course.  Txs again!

hamishwillee:
Thanks @rmackay9 . We'd normally wait for this to be  fully tested - i.e. the point at which you are sure no changes will be required. Are you at that point? Otherwise maybe wait until after the competition?

rmackay9:
sure, we can wait a bit to see ..

Davidsastresas:
Hi, just bumping this thread in case it got forgotten! It is something that eventually we will need so it would be good to keep advancing even if slow. Thanks!

rmackay9:
From my point of view, this is fully baked.  We used it very successfully in this year's [Japan Innovation Challenge](https://discuss.ardupilot.org/t/tap-j-at-the-japan-innovation-challenge-2024/125750).  The ability to get the real time maximum absolute temperature was critical in this success.  There's a pic in the middle of the linked blog post showing what it looked like on Mission Planner when we flew over a dummy.

hamishwillee:
Thanks both. Added proposal to remove WIP tagging in https://github.com/mavlink/mavlink/pull/2189

PR #2146, opened 2024-08-27
development.xml: add GNSS message

peterbarker:
intended as a replacement for GPS_RAW_INT and GPS2_RAW, adding an instance number to stretch to >2 GPSs via mavlink

TALKING POINT ONLY - do not merge!

### DIFF
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -120,6 +120,18 @@
         <description>Reserved (not used). If set, this will indicate that an additional status field exists for higher status values.</description>
       </entry>
     </enum>
+    <enum name="MAV_GNSS_DATUM">
+      <description>Possible datums location is specified in GNSS message.
+      </description>
+      <entry value="0" name="MAV_GNSS_DATUM_INVALID">
+        <description>Invalid datum; position not valid.
+        </description>
+      </entry>
+      <entry value="1" name="MAV_GNSS_DATUM_WGS84">
+        <description>WGS84 dataum.
+        </description>
+      </entry>
+    </enum>
     <!-- The MAV_CMD enum entries describe either: -->
     <!--  * the data payload of mission items (as used in the MISSION_ITEM_INT message) -->
     <!--  * the data payload of mavlink commands (as used in the COMMAND_INT and COMMAND_LONG messages) -->
@@ -582,6 +594,26 @@
       <field type="uint8_t" name="system_status_summary" minValue="0" maxValue="10" invalid="UINT8_MAX">An abstract value representing the overall status of the receiver, or 255 if not available.</field>
       <field type="uint8_t" name="gnss_signal_quality" minValue="0" maxValue="10" invalid="UINT8_MAX">An abstract value representing the quality of incoming GNSS signals, or 255 if not available.</field>
       <field type="uint8_t" name="post_processing_quality" minValue="0" maxValue="10" invalid="UINT8_MAX">An abstract value representing the estimated PPK quality, or 255 if not available.</field>
+    </message>
+    <message id="442" name="GNSS">
+      <!-- a replacement for GPS_RAW_INT and GPS2_RAW which includes an instance number -->
+      <description>GNSS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_above_datum" units="m">Altitude. Positive for up. Use NaN for a 2D fix.</field>
+      <field type="uint8_t" name="datum" enum="MAV_GNSS_DATUM">Datum used for position.</field>
+      <field type="float" name="ground_speed" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
+      <field type="uint16_t" name="cog" units="cdeg" invalid="UINT16_MAX">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="float" name="h_acc" units="m">1-Sigma Position accuracy.  Set to NaN if not known.</field>
+      <field type="float" name="v_acc" units="m">1-Sigma Altitude accuracy.  Set to NaN if not known.</field>
+      <field type="float" name="vel_acc" units="m">1-Sigma Speed accuracy.  Set to NaN if not known.</field>
+      <field type="float" name="hdg_acc" units="deg">1-Sigma Heading / track accuracy.  Set to NaN if not known.</field>
     </message>
     <!-- Target info from a sensor on the target -->
     <message id="510" name="TARGET_ABSOLUTE">
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -608,7 +608,7 @@
       <field type="uint8_t" name="datum" enum="MAV_GNSS_DATUM">Datum used for position.</field>
       <field type="float" name="ground_speed" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
       <field type="uint16_t" name="cog" units="cdeg" invalid="UINT16_MAX">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
-      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint8_t" name="satellites_used">Number of satellites used in the PVT solution. If unknown, set to 255</field>
       <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
       <field type="float" name="h_acc" units="m">1-Sigma Position accuracy.  Set to NaN if not known.</field>
       <field type="float" name="v_acc" units="m">1-Sigma Altitude accuracy.  Set to NaN if not known.</field>


message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up.</field>
+      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="vel" units="cm/s">GPS ground speed. If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint8_t" name="dgps_numch">Number of DGPS satellites</field>
+      <field type="uint32_t" name="dgps_age" units="ms">Age of DGPS info</field>
```
peterbarker, line 757: Do we really need these?

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+    <message id="442" name="GNSS">
+      <description>GPS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up.</field>
+      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
```
peterbarker, line 752: Should we kill these?  They've always sucked.

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint16_t" name="vel" units="cm/s">GPS ground speed. If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint8_t" name="dgps_numch">Number of DGPS satellites</field>
+      <field type="uint32_t" name="dgps_age" units="ms">Age of DGPS info</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="int32_t" name="alt_ellipsoid" units="mm">Altitude (above WGS84, EGM96 ellipsoid). Positive for up.</field>
```
peterbarker, line 759: Should probably be grouped with the AMSL altitude, above

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint32_t" name="h_acc" units="mm">Position uncertainty.</field>
+      <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty.</field>
```
peterbarker, line 761: 2^32mm is a lot of mm.
```
You have: 2^32 millimetres
You want: metres
	* 4294967.3
	/ 2.3283064e-07
You have: 2^16 millimetres
You want: metres
	* 65.536
	/ 0.015258789
```

Is 65m enough, so use uint16_t?

Change this to micrometres for future-proofing?

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint32_t" name="vel_acc" units="mm">Speed uncertainty.</field>
```
peterbarker, line 762: Units are wrong on this.

Similar questions on size and units as for the positions

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="442" name="GNSS">
+      <description>GPS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
```
peterbarker, line 749: Should we really lock ourselves into int32_t?  Time for uint64_t?

hamishwillee: What's the difference in precision you would get from a `uint64_t`? Is this a useful?

- Say the precision is mm, what are the use cases where this would be handy?
- Would this imply a corresponding change needed for other position messages?
- We use scaled ints because of rounding errors that would result in use of corresponding floats. What if we used a bigger real number format?

message_definitions/v1.0/development.xml
hamishwillee: I don't have the context to know what "sux" means. If no one is using them and they aren't fit for any of the use cases that people have then sure. Is here a replacement/alternative?

message_definitions/v1.0/development.xml
hamishwillee: Could we use micrometers for anything realistic? I mean breathing on the vehicle would change the result wildly. I don't know if 65 m is enough.

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint8_t" name="dgps_numch">Number of DGPS satellites</field>
+      <field type="uint32_t" name="dgps_age" units="ms">Age of DGPS info</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="int32_t" name="alt_ellipsoid" units="mm">Altitude (above WGS84, EGM96 ellipsoid). Positive for up.</field>
+      <field type="uint32_t" name="h_acc" units="mm">Position uncertainty.</field>
+      <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty.</field>
+      <field type="uint32_t" name="vel_acc" units="mm">Speed uncertainty.</field>
+      <field type="uint32_t" name="hdg_acc" units="degE5">Heading / track uncertainty</field>
+    </message>
```
hamishwillee, line 764: https://mavlink.io/en/messages/common.html#GPS_RAW_INT also has yaw  extension

hamishwillee:
So just to be clear, this is to test the appetite for a modified https://mavlink.io/en/messages/common.html#GPS_RAW_INT with an instance number, and some (possible) future proofing? 

How would be manage the cross over - vehicle streams this by default but GCS can request old version if it wants?

message_definitions/v1.0/development.xml
peterbarker: Yes, these are unitless values telling you how precise your GPS is being right now.  My understanding is that there *is* an actual technical definition for these but also that not all GPS units out there use that standard.

The accuracy fields below - in SI units - are so much better

message_definitions/v1.0/development.xml
peterbarker: 32-bit latitudes at the equator is ~8mm granularity.  There are GPS out there with better than that right now.

We could potentially ignore this problem again for now and stick with 32-bits.

We could go to a float double or 64-bit integer.  Not having scaling would be nice, I guess.

All seems rather wasteful when we probably only need another few bits.  We could split a uint8_t to add 4 bits to each of the lat/lng to add more precision, perhaps.  Drone telemetry bandwidth isn't growing at a vast rate, for the most part.

peterbarker:
> So just to be clear, this is to test the appetite for a modified https://mavlink.io/en/messages/common.html#GPS_RAW_INT with an instance number, and some (possible) future proofing?
> 
> Note, I'm very much in favour - if we're going to add another message we should do it with an instance number.
> 
> How would be manage this? I mean GCS that have been updated will be fine - if they see this message they know it is used for all GNSS instances. Those that haven't won't know about the extra instance, but would know about the original messages.
> 
> The ideal would be that if a GCS doesn't see the messages it understand it should request them. Then we could stream this new message by default and a GCS that hasn't been updated could request the old ones.

Yes, getting from here to there on such a fundamental message will be interesting.  Not quite as fundamental as GLOBAL_POSITION_INT, but still...

Phase one might be GCSs can request it, meaning they can support it and tell the autopilot to stop streaming the other.  Phase two might be to switch things around - stop streaming `GPS_RAW_INT` and `GPS2_RAW` and stream this instead.

message_definitions/v1.0/development.xml
hamishwillee: I guess this is one way to get some bytes back. We shouldn't duplicate the information, so if we can get it in another way, then yes, removal makes sense.

message_definitions/v1.0/development.xml
hamishwillee: What are the use cases that need more accuracy than 8mm? If we increase this we're presumably also going to have to modify the ranges for the uncertainties, and also for any corresponding setter commands. Is it worth it?

My gut feel on this is that the vast majority do not need that kind of resolution. If that becomes a requirement I'd suggest we could invent another message that did the job and allow the flight stack to choose which to send based on either the GPS it has onboard or a configuration setting.

hamishwillee:
>> Yes, getting from here to there on such a fundamental message will be interesting. Not quite as fundamental as GLOBAL_POSITION_INT, but still...

> Phase one might be GCSs can request it, meaning they can support it and tell the autopilot to stop streaming the other. Phase two might be to switch things around - stop streaming GPS_RAW_INT and GPS2_RAW and stream this instead.

Yes. 

What we'd really like to do is get the GCS to request all the things they know about and need all the time (or if not sent by default). Then this kind of switch would be easy.

But if we do what you're suggesting, that will at least be true for this message in future, since GCS will be requesting it.

message_definitions/v1.0/development.xml
peterbarker: > https://mavlink.io/en/messages/common.html#GPS_RAW_INT also has yaw extension

Is it different to the yaw field I already put in here?

hamishwillee: Nope, I'm just blind.

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="442" name="GNSS">
+      <description>GPS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
```
tridge, line 746: should say that zero is assumed to the the first sensor

message_definitions/v1.0/development.xml
tridge: i'd prefer to leave it as int32_t

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up.</field>
+      <field type="uint16_t" name="eph">GPS HDOP horizontal dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="epv">GPS VDOP vertical dilution of position (unitless). If unknown, set to: UINT16_MAX</field>
+      <field type="uint16_t" name="vel" units="cm/s">GPS ground speed. If unknown, set to: UINT16_MAX</field>
```
tridge, line 753: we should make this m/s float. We likely will have vehicles flying at over 655 m/s in the lifetime of this msg

message_definitions/v1.0/development.xml
tridge: i think we can remove them

message_definitions/v1.0/development.xml
tridge: should say what to send if we only have one height, probably best to send as equal if undulation not known

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint8_t" name="dgps_numch">Number of DGPS satellites</field>
+      <field type="uint32_t" name="dgps_age" units="ms">Age of DGPS info</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="int32_t" name="alt_ellipsoid" units="mm">Altitude (above WGS84, EGM96 ellipsoid). Positive for up.</field>
+      <field type="uint32_t" name="h_acc" units="mm">Position uncertainty.</field>
```
tridge, line 760: might as well make these float, uint32_t doesn't win us anything, and means we can use NaN for unknown

message_definitions/v1.0/development.xml
tridge: again float, and NaN for unknown

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
+      <field type="uint32_t" name="v_acc" units="mm">Altitude uncertainty.</field>
+      <field type="uint32_t" name="vel_acc" units="mm">Speed uncertainty.</field>
+      <field type="uint32_t" name="hdg_acc" units="degE5">Heading / track uncertainty</field>
```
tridge, line 763: float and NaN

message_definitions/v1.0/development.xml
tridge: hmm, maybe float and NaN for unknown?

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
...
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="442" name="GNSS">
+      <description>GPS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="int32_t" name="alt" units="mm">Altitude (MSL). Positive for up.</field>
```
tridge, line 750: better to make this float I think. Maybe NaN for a 2D fix?

tridge (changes requested):

tridge:
@peterbarker should we consider using one of the precious remaining IDs under 255 for this, given how key it is? some systems still send MAVLink1

tridge:
@peterbarker I also think we should add gps_week and gps_week_ms, that is the real key to time handling with GNSS

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +740,27 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="442" name="GNSS">
+      <description>GPS data.</description>
```
hamishwillee, line 744:
```suggestion
      <description>GNSS data.</description>
```

message_definitions/v1.0/development.xml
julianoes: I don't like this field named `_acc`. We should make it clear whether this is a standard deviation or variance.

And if we do call it `_acc` for accuracy, then the comment should be accuracy, and not uncertainty.

message_definitions/v1.0/development.xml
julianoes: We should probably not copy them over as is. The description says one thing and the value name something else. This was done wrongly back in the day as far as I'm aware.

We should make this clear, and not confusing/wrong. I would vote to keep eph and epv for sure. HDOP and VDOP could be included as well but I'm not too sure whether that's used much these days.

message_definitions/v1.0/development.xml
julianoes, line 744:
```suggestion
      <description>GNSS data.</description>
```

And we need to comment of what it replaces, at least before moving to common I suppose.

julianoes (commented):
I'm in favor of this. I'm just wondering what additional fields we should add, e.g. UTC GTPS time.

message_definitions/v1.0/development.xml
hamishwillee: Are you saying "state this is 0 indexed" or "if this is 1 indexed, 0 would would also represent the first sensor"?

I ask because the existing messages are effectively 1-indexed based on their names. Messages like BATTERY_STATUS that use an index are sending 1 for the first battery (at least on PX4). There are also cases where indexing from `1` is better - for example if you're requesting a message for a particular index, then `0` index has to mean "all messages", so to get index 0 you would have to request `1` here (all a bit messy).

peterbarker: OK, so PX4 are likely to want to index from 1.

ArduPilot indexes `BATTERY_STATUS` from zero, so we're already different there.

I do see the problem here - I'll raise a PR supporting requesting BATTERY_STATUS in ArduPilot to highlight the problem.  Given the lack of GCS support, I'll probably suggest that we support "-1" for all instances when requesting streams/one-shots.  We'll probably need to get GCSs to start sending "-1" through by default before we start to support per-instance-streams/one-shots.

@nexton-winjeel 's PR here: https://github.com/ArduPilot/ardupilot/pull/28353 highlights that the assumption that "0" means "all instances" is problematic; for a mavlink message that should spit out a heartbeat, not one of every message the autopilot supports :-)

For now I'm going to leave this without a comment.

message_definitions/v1.0/development.xml
peterbarker: Done

message_definitions/v1.0/development.xml
peterbarker: Done

message_definitions/v1.0/development.xml
peterbarker: Removed

message_definitions/v1.0/development.xml
peterbarker: I've left this in as two separate fields for now, both float and NaNable.

I think @hamishwillee 's idea of using an enumeration has merit!

@hamishwillee yes, GPSs can supply AMSL and geoid heights.  Sometimes they even tell you which they're supplying.

message_definitions/v1.0/development.xml
peterbarker: Done

message_definitions/v1.0/development.xml
peterbarker: I've changed this to (eg.):
```
      <field type="float" name="h_acc" units="m">1-Sigma Position accuracy.  Set to NaN if not known.</field>
```

(I believe that's what uBlox report, for example)

message_definitions/v1.0/development.xml
peterbarker: Fixed

message_definitions/v1.0/development.xml
peterbarker: Done

peterbarker:
> I'm in favor of this. I'm just wondering what additional fields we should add, e.g. UTC GTPS time.

GTPS?  I guess you meant GPS here?

@tridge suggested gps week and milliseconds, the traditional GPS thing.  Sound OK to you?

I do find it a little strange the original messages didn't have these... but do note `GPS_RTK` and `GPS2_RTK` *do* have them.

Have to come back and create `GNSS_RTK` later....

message_definitions/v1.0/development.xml
```diff
@@ -792,5 +792,24 @@
       <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no GCS in control).</field>
       <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this message instance defines the default controlling GCS for the whole system.</field>
     </message>
+    <message id="442" name="GNSS">
+      <!-- a replacement for GPS_RAW_INT and GPS2_RAW which includes an instance number -->
+      <description>GNSS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
```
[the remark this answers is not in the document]
peterbarker: tridge thinks not

message_definitions/v1.0/development.xml
```diff
@@ -792,5 +792,24 @@
...
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_ellipsoid" units="m">Altitude (above WGS84, EGM96 ellipsoid). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="vel" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
+      <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
```
peterbarker, line 808: Do we want to retain these semantics or some up with some new ones?

The old semantics were partly brought about because this was an extension field

message_definitions/v1.0/development.xml
hamishwillee: OK. So no change on this one (at least for now).

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
hamishwillee: I don't know what this is used for, and hence why the selection of microseconds originally. But if you know that microseconds are not needed then yes please, change.

hamishwillee:
> > I'm in favor of this. I'm just wondering what additional fields we should add, e.g. UTC GTPS time.
> 
> GTPS? I guess you meant GPS here?
> 
> @tridge suggested gps week and milliseconds, the traditional GPS thing. Sound OK to you?
> 
> I do find it a little strange the original messages didn't have these... but do note `GPS_RTK` and `GPS2_RTK` _do_ have them.
> 
> Have to come back and create `GNSS_RTK` later....

I don't understand what they would are used for, but sure if they are useful to someone and you are happy to take the cost of transport.

@julianoes Comments to this?

message_definitions/v1.0/development.xml
hamishwillee: It is usual to provide "data not supplied now", but not to split the nuance "not supported" and "not supplied now. Is this nuance useful? If not, then we should do:

```suggestion
      <field type="uint16_t" name="yaw" units="cdeg" invalid="UINT16_MAX">Yaw in earth frame from north. 0: North. Use UINT16_MAX if this GNSS is not providing yaw.</field>
```

If it is useful, then what do you suggest?

message_definitions/v1.0/development.xml
```diff
@@ -792,5 +792,24 @@
...
+    <message id="442" name="GNSS">
+      <!-- a replacement for GPS_RAW_INT and GPS2_RAW which includes an instance number -->
+      <description>GNSS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_ellipsoid" units="m">Altitude (above WGS84, EGM96 ellipsoid). Positive for up.  Use NaN for a 2D fix.</field>
```
hamishwillee, line 804: @peterbarker We really need to have the discussion in https://github.com/mavlink/mavlink/pull/2168 to decide what these should be. 

Essentially WGS84 is an ellipsoid, EGM96 is a geoid, and MSL is an ambiguous term that somewhat matches the EGM96 geoid, and which ublox approximates using something that is not actually MSL and is (possibly) a worse approximation than EGM96 to MSL.

YOu can't say and altitude is (above WGS84, EGM96 ellipsoid) because those are different references, and only one can be true.

What we probably mean is MSL = whatever uBlox spits out. For alt_ellipsoid, we probably mean `alt_egm96_geoid`. But who knows. 

Probably we should everywhere state that MSL should be interpreted as alt_egm96_geoid since that is at least a consistent and believable reference. OR we should say it is uBlox MSL.

Anyway, the way I would do it IF we are going to move to a consistent reference.

```suggestion
      <field type="float" name="alt_egm96_geoid" units="m">Altitude (EGM96 geoid). Positive for up.  Use NaN for a 2D fix.</field>
```

or we move to this definition for alt by default and start emitting it instead of ublox MSL

```suggestion
      <field type="float" name="alt" units="m">Altitude (EGM96 geoid). Positive for up. Use NaN for a 2D fix.</field>
```

Or if we want to stick with MSL but just give it a clear definition and do this consistently everywhere

```suggestion
      <field type="float" name="alt" units="m">Altitude (uBlox MSL). Positive for up. Use NaN for a 2D fix.</field>
      <field type="float" name="alt_egm96_geoid" units="m">Altitude (EGM96 geoid). Positive for up.  Use NaN for a 2D fix.</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -792,5 +792,24 @@
...
+      <field type="float" name="vel" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
+      <field type="uint16_t" name="cog" units="cdeg">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
```
hamishwillee, line 806:
```suggestion
      <field type="uint16_t" name="cog" units="cdeg" invalid="UINT16_MAX">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
```

peterbarker:
> @peterbarker I also think we should add gps_week and gps_week_ms, that is the real key to time handling with GNSS

I've discussed this with tridge and he's expressed the opinion that if we've done without them so far in the other messages they're probably not required here either.

message_definitions/v1.0/development.xml
```diff
@@ -674,6 +686,26 @@
...
+      <!-- a replacement for GPS_RAW_INT and GPS2_RAW which includes an instance number -->
+      <description>GNSS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_above_datum" units="m">Altitude. Positive for up. Use NaN for a 2D fix.</field>
+      <field type="uint8_t" name="datum" enum="MAV_GNSS_DATUM">Datum used for position.</field>
```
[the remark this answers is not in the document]
peterbarker: I'm not really in favour of "GPS-specific AMSL" as a reference frame; we may as well start sending `mav_message_ublox` and `mav_message_nmea` instead :-)

message_definitions/v1.0/development.xml
```diff
@@ -674,6 +686,26 @@
...
+      <field type="float" name="vel" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
+      <field type="uint16_t" name="cog" units="cdeg" invalid="UINT16_MAX">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="float" name="h_acc" units="m">1-Sigma Position accuracy.  Set to NaN if not known.</field>
+      <field type="float" name="v_acc" units="m">1-Sigma Altitude accuracy.  Set to NaN if not known.</field>
+      <field type="float" name="vel_acc" units="m">1-Sigma Speed accuracy.  Set to NaN if not known.</field>
+      <field type="float" name="hdg_acc" units="deg">1-Sigma Heading / track accuracy.  Set to NaN if not known.</field>
```
hamishwillee, line 707:
```suggestion
      <field type="float" name="h_acc" units="m" invalid="NaN">1-Sigma Position accuracy. Set to NaN if not known.</field>
      <field type="float" name="v_acc" units="m" invalid="NaN">1-Sigma Altitude accuracy. Set to NaN if not known.</field>
      <field type="float" name="vel_acc" units="m" invalid="NaN">1-Sigma Speed accuracy. Set to NaN if not known.</field>
      <field type="float" name="hdg_acc" units="deg" invalid="NaN">1-Sigma Heading / track accuracy.  Set to NaN if not known.</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -674,6 +686,26 @@
...
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_above_datum" units="m">Altitude. Positive for up. Use NaN for a 2D fix.</field>
+      <field type="uint8_t" name="datum" enum="MAV_GNSS_DATUM">Datum used for position.</field>
+      <field type="float" name="vel" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
+      <field type="uint16_t" name="cog" units="cdeg" invalid="UINT16_MAX">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
+      <field type="uint16_t" name="yaw" units="cdeg">Yaw in earth frame from north. Use 0 if this GPS does not provide yaw. Use 65535 if this GPS is configured to provide yaw and is currently unable to provide it. Use 36000 for north.</field>
+      <field type="float" name="h_acc" units="m" invalid="NaN">1-Sigma Position accuracy. Set to NaN if not known.</field>
```
hamishwillee, line 704: Just FMI, this means "1-sigma speed accuracy: This means that 68% of the time, the actual speed of the aircraft will be within one standard deviation of the measured speed."

I presume this is the normal/most useful way to provide the information for consumers?

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
hamishwillee: @peterbarker I strongly agree, and yet  "GPS-specific AMSL" is what we send now, and have sent forever - it's just that historically we got lucky because the vast majority used UBlox and were sending `mav_message_ublox`.

Anyway, I don't have a problem with this conceptually. I just think we need to look at this problem holistically
- what are you expecting setters to do? I.e. I got an alt in datum "whatever" but I now have to set an alt in ublox MSL, because that is what all the commands say. Who converts?
- What about the old GPS messages, and all the other ones?

If we solve this problem for this message but not everywhere else then we're complicating the problem.

hamishwillee:
@peterbarker @tridge There's a proposal for a "tangential" message to support publishing a global position from any source, not necessarily a GNSS here. Would appreciate your thoughts on whether it makes sense to try make this general, or whether this is a reasonable separate use case: https://github.com/mavlink/mavlink/pull/2256#discussion_r2075123964

peterbarker:
Working on this one again a little.  Wondering if we should try to reduce the packet size and use float16 for the accuracies.

ArduPilot and pymavlink has good support for float16, so I think we'd be OK there.

Would make life harder for others, 'though.

message_definitions/v1.0/development.xml
```diff
@@ -674,6 +686,26 @@
...
+      <description>GNSS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_above_datum" units="m">Altitude. Positive for up. Use NaN for a 2D fix.</field>
+      <field type="uint8_t" name="datum" enum="MAV_GNSS_DATUM">Datum used for position.</field>
+      <field type="float" name="vel" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
```
dakejahl:
```suggestion
      <field type="float" name="ground_speed" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
```

dakejahl (commented):
Looking forward to this. One nitpick I have is if we're going to spell out `satellites_visible` we should also do the same for `heading_accuracy` and others.

message_definitions/v1.0/development.xml
peterbarker: Fair call, done.

ThomasRigi:
I like this. I found it through https://github.com/dakejahl/px4_ideas/blob/main/gnss/data_reporting_over_mavlink_and_dronecan.md coming from [RFC 30](https://github.com/mavlink/rfcs/pull/30), which I also like. ;) 

We've been internally using something very similar already. Glad to see that we're not the only ones that don't like how GPS_RAW_INT and GPS2_RAW are handled. 

One additional point is that we need an extra flag that we simply called `used` (but open for more different naming) to indicate if the corresponding GNSS is actively used in the position fusion algorithm. I think this is interesting information on any setup, but in particular on dual- or multi-GNSS setups that might also blend the different GNSS raw data. Knowing which GNSS(s) is(/are) actively used in the EKF (or knowing that none is used...) provides important information to us and our clients.

message_definitions/v1.0/development.xml
```diff
@@ -583,6 +595,26 @@
...
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
+      <field type="float" name="alt_above_datum" units="m">Altitude. Positive for up. Use NaN for a 2D fix.</field>
+      <field type="uint8_t" name="datum" enum="MAV_GNSS_DATUM">Datum used for position.</field>
+      <field type="float" name="ground_speed" units="m/s">GPS ground speed. If unknown, set to: NaN</field>
+      <field type="uint16_t" name="cog" units="cdeg" invalid="UINT16_MAX">Course over ground (NOT heading, but direction of movement): 0.0..359.99 degrees. If unknown, set to: UINT16_MAX</field>
+      <field type="uint8_t" name="satellites_visible">Number of satellites visible. If unknown, set to 255</field>
```
dakejahl:
```suggestion
      <field type="uint8_t" name="satellites_used">Number of satellites used in the PVT solution. If unknown, set to 255</field>
```
Used is what matters. In practice this field has been stuffed with the used count anyway, so we should correct the name

peterbarker:
On Thu, 20 Aug 2026, Thomas Stauber wrote:

> One additional point is that we need an extra flag that we simply called used (but open for more different naming) to indicate if the corresponding GNSS is actively used in the position
> fusion algorithm. I think this is interesting information on any setup, but in particular on dual- or multi-GNSS setups that might also blend the different GNSS raw data. Knowing which
> GNSS(s) is(/are) actively used in the EKF (or knowing that none is used...) provides important information to us and our clients.

Interesting idea, and I can see the utility.

However, we don't usually include that information in the sensor data - 
and a simple boolean really can't contain the information required.

For example, ArduPilot can run several cores with different data sets. 
It's probably *very* useful information to the operator as to whether it's 
being used by the core whose data source set nominates it - but that core 
may not currently be providing attitude to the vehicle.

I'd presume the boolean flag would follow the active estimator on the 
vehicle - that's straight forward.  But can't see how it works with 
multiple data sources.

There's also be some fun around aggregated GPS data.  ArduPilot's 
"blended" backend would *possibly* light up "used" for all things being 
blended.  A little bit of interesting plumbing there for us - but the 
general class of "GPS sensor isn't actually used directly by the 
estimator" might be more awkward in other implementations.

What-an-estimator-is-using is probably best left to messages specific to 
the estimator.  This *has* been proposed before in various places (e.g 
https://github.com/ArduPilot/mavlink/pull/428)

ThomasRigi:
Fair point with it being better suited in an estimator-related topic. I found [ESTIMATOR_SENSOR_FUSION_STATUS](https://mavlink.io/en/messages/development.html#ESTIMATOR_SENSOR_FUSION_STATUS) introduced in https://github.com/mavlink/mavlink/pull/2439, but I don't understand how you would use that message (yet) to state which of the two GNSS are used in an instance, just that "at least one" GNSS is used via [ESTIMATOR_SENSOR_FUSION_SOURCE](https://mavlink.io/en/messages/development.html#ESTIMATOR_SENSOR_FUSION_SOURCE). I will follow-up over there to not clutter this PR.

edit: I think for as long as you don't use multi-instance estimators (or only report on the instance that is actually active), you can use ESTIMATOR_SENSOR_FUSION_STATUS for my intentions. -> 
e.g. to say that the estimator is using both GNSS, it would be `active[ESTIMATOR_SENSOR_FUSION_SOURCE_GPS] = 3` (0b11)

message_definitions/v1.0/development.xml
```diff
@@ -583,6 +595,26 @@
...
     </message>
+    <message id="442" name="GNSS">
+      <!-- a replacement for GPS_RAW_INT and GPS2_RAW which includes an instance number -->
+      <description>GNSS data.</description>
+      <field type="uint64_t" name="time_usec" units="us">Timestamp (UNIX Epoch time or time since system boot). The receiving end can infer timestamp format (since 1.1.1970 or since system boot) by checking for the magnitude of the number.</field>
+      <field type="uint8_t" name="id" instance="true">GNSS receiver id. Must match instance ids of other messages from same receiver.</field>
+      <field type="uint8_t" name="fix_type" enum="GPS_FIX_TYPE">GPS fix type.</field>
+      <field type="int32_t" name="lat" units="degE7">Latitude (WGS84)</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude (WGS84)</field>
+      <field type="float" name="alt" units="m">Altitude (MSL). Positive for up.  Use NaN for a 2D fix.</field>
```
dakejahl, line 606: "MSL" doesn't say which geoid. Receivers typically use a coarse built-in table (Septentrio: the 1991 STANAG 4294 model; u-blox reportedly a 10°x10° EGM96 grid), so this value can be metres off true EGM96/EGM2008.

Suggest the description say: geoid model is receiver-dependent, and `alt_above_datum - alt` is the undulation the receiver applied. That lets a consumer undo it and apply a better model.

Optional: a small `geoid_model` enum (unknown / receiver built-in / EGM96 / EGM2008 / manual) so a GCS can tell a coarse MSL from a good one.

message_definitions/v1.0/development.xml
```diff
@@ -583,6 +595,26 @@
...
+      <field type="float" name="alt_above_datum" units="m">Altitude. Positive for up. Use NaN for a 2D fix.</field>
```
dakejahl, line 607: Suggest "Height above the ellipsoid of `datum`". A datum isn't a surface you can be above.

Also worth stating for both altitude fields: NaN if the receiver doesn't provide it, never copy one into the other. Some DroneCAN nodes send the same value in both Fix2 height fields, and ArduPilot needs a heuristic to detect it.

message_definitions/v1.0/development.xml
```diff
@@ -120,6 +120,18 @@
         <description>Reserved (not used). If set, this will indicate that an additional status field exists for higher status values.</description>
       </entry>
     </enum>
+    <enum name="MAV_GNSS_DATUM">
+      <description>Possible datums location is specified in GNSS message.
+      </description>
+      <entry value="0" name="MAV_GNSS_DATUM_INVALID">
+        <description>Invalid datum; position not valid.
```
dakejahl, line 127: This mixes datum validity with fix validity, which `fix_type` already covers. Suggest `MAV_GNSS_DATUM_UNKNOWN` instead.

message_definitions/v1.0/development.xml
```diff
@@ -120,6 +120,18 @@
...
+        </description>
+      </entry>
+      <entry value="1" name="MAV_GNSS_DATUM_WGS84">
+        <description>WGS84 dataum.
```
dakejahl, line 131: Typo: "dataum".

With RTK the solution is in the base station's frame, not WGS84, and receivers report which one. Septentrio's list is a reasonable starting set: WGS84/ITRF, ETRS89, NAD83, NAD83_PA, NAD83_MA, GDA94, GDA2020, plus user-defined.

dakejahl (commented):
A few notes on the altitude/datum fields, prompted by a customer trying to recover precise elevation from an RTK receiver. Reporting both heights plus a datum is a real improvement over GPS_RAW_INT. Related: #2167.

message_definitions/v1.0/development.xml
peterbarker: > This mixes datum validity with fix validity, which `fix_type` already covers. Suggest `MAV_GNSS_DATUM_UNKNOWN` instead.

I think you're just objecting to the name here, not the "position not valid" thing?  You're 300m up from.... something.... doesn't really help navigate an aircraft, after all.

dakejahl: Yeah, different word choice would reduce possible confusion

PR #2147, opened 2024-09-04
Clarify fence vertex count and order

hamishwillee:
The behaviour of fence vertex commands was ambiguous because "fence count" could be "number of vertices" or "the current vertix in the polygon definition". It actually means "the number of vertices in this polygon, and is shared by all vertices in the polygon".

I also spelled out that that vertices in the polygon must be sent sequentially. This should make it clear how this works - all the vertices in a polygon are sent in order, and share the vertix count number.


hamishwillee:
Closing. Current version in master is at least as good as this.

PR #2148, opened 2024-09-04
MAV_CMD_SET_MESSAGE_INTERVAL - additional params for requestor

hamishwillee:
This adds parameter details to [MAV_CMD_SET_MESSAGE_INTERVAL](https://mavlink.io/en/messages/common.html#MAV_CMD_SET_MESSAGE_INTERVAL ) to mirror [MAV_CMD_REQUEST_MESSAGE](https://mavlink.io/en/messages/common.html#MAV_CMD_REQUEST_MESSAGE) .

Specifically, request message allows you to get a particular instance of a message. So if you're only interested in the message for Battery_status instance 2 say, then that's the one you'll get. This allows you to do the same thing for intervals, if you're only interested in setting the interval for one thing. Note, I'm not sure you'll ever need this, and it would be a pain to implement, but the facility is there.

Further, you might want to configure other information that decides what values you get back, and this updates the remaining params to be used in this way.

This is needed for https://github.com/mavlink/mavlink/pull/2145, which wants to be able to specify that a streamed message belongs to a specific camera and video stream.

@peterbarker @julianoes Does this make sense to you?

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2042,6 +2042,10 @@
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM.</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
         <param index="2" label="Interval" units="us" minValue="-1" increment="1">The interval between two messages. -1: disable. 0: request default rate (which may be zero).</param>
+        <param index="3" label="Req Param 3">Use for index ID, if required. Otherwise, the use of this parameter (if any) must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="4" label="Req Param 4">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="5" label="Req Param 5">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="6" label="Req Param 6">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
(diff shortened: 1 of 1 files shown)


peterbarker:
Makes sense to me.

OTOH, this is a pretty tough ask for ArduPilot's GCS library.  It's not impossible, but we already expend a bunch of embedded resources on the existing message-rate stuff, and extending the rates to be per-instance is going to be a bunch more RAM and CPU.  And a heap of complexity.

There's also the problem that existing autopilot implementations will ignore that field, so a GCS trying to set a rate for a specific message ID will inadvertently set the rate for all instances.  If we want to change the meaning of param3 like this we'd best get some code in place which rejects this message if any of params 3, 4, 5 or 6 are non-zero.

hamishwillee:
>  If we want to change the meaning of param3 like this we'd best get some code in place which rejects this message if any of params 3, 4, 5 or 6 are non-zero.

Yes, setting [MAV_RESULT_DENIED](https://mavlink.io/en/messages/common.html#MAV_RESULT_DENIED) in this case would be a minimal implementation before you even start supporting streaming particular messages with more params.

julianoes (approved):

hamishwillee:
Thanks. Merging.

hamishwillee:
PX4 implementation of "make it reject with denied ACK if an unused param is non-zero in https://github.com/PX4/PX4-Autopilot/pull/23643

Note, this shows an incompatible use of the API, which is to set the data rates (?). Might be some pushback to use param 3 for something else.

peterbarker:
> PX4 implementation of "make it reject with denied ACK if an unused param is non-zero in [PX4/PX4-Autopilot#23643](https://github.com/PX4/PX4-Autopilot/pull/23643)
> 
> Note, this shows an incompatible use of the API, which is to set the data rates (?). Might be some pushback to use param 3 for something else.

I did a keyhole PR against ArduPilot here for the specific case we're looking at: https://github.com/ArduPilot/ardupilot/pull/28019

PR #2150, opened 2024-09-05
common.xml: adjust field instance descriptions

peterbarker:
also point out flag value can be different based on transport

@julianoes whoops.  I completely missed the instances-start-at-zero thing...

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2042,10 +2042,10 @@
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM.</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
         <param index="2" label="Interval" units="us" minValue="-1" increment="1">The interval between two messages. -1: disable. 0: request default rate (which may be zero).</param>
-        <param index="3" label="Req Param 3">Use for index ID, if required. Otherwise, the use of this parameter (if any) must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="3" label="Req Param 3">Use for index ID, if required. Otherwise, the use of this parameter (if any) must be defined in the requested message. By default assumed not used (0).  When used as an index ID, 0 means "all instances", "1" means "instance 0", "2" means "instance 1" etc</param>
         <param index="4" label="Req Param 4">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
-        <param index="5" label="Req Param 5">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
-        <param index="6" label="Req Param 6">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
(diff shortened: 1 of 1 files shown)


julianoes (approved):

message_definitions/v1.0/common.xml
```diff
@@ -2042,10 +2042,10 @@
         <description>Set the interval between messages for a particular MAVLink message ID. This interface replaces REQUEST_DATA_STREAM.</description>
         <param index="1" label="Message ID" minValue="0" maxValue="16777215" increment="1">The MAVLink message ID</param>
         <param index="2" label="Interval" units="us" minValue="-1" increment="1">The interval between two messages. -1: disable. 0: request default rate (which may be zero).</param>
-        <param index="3" label="Req Param 3">Use for index ID, if required. Otherwise, the use of this parameter (if any) must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="3" label="Req Param 3">Use for index ID, if required. Otherwise, the use of this parameter (if any) must be defined in the requested message. By default assumed not used (0).  When used as an index ID, 0 means "all instances", "1" means "instance 0", "2" means "instance 1" etc</param>
```
hamishwillee, line 2045: Yes, good to clarify that when used as an index the number and instance might not match. Does using `instance 0` clearly imply "first item in sequence does not need to be 1"? Would this be more clear?

```suggestion
        <param index="3" label="Req Param 3">Use for index ID, if required. Otherwise, the use of this parameter (if any) must be defined in the requested message. By default assumed not used (0).  When used as an index ID, 0 means "all instances", "1" means the first instance in the sequence (the emitted message will have an id of 0 if message ids are 0-indexed, or 1 if index numbers start from one).</param>
```

message_definitions/v1.0/common.xml
```diff
@@ -2042,10 +2042,10 @@
...
         <param index="4" label="Req Param 4">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
-        <param index="5" label="Req Param 5">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
-        <param index="6" label="Req Param 6">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0).</param>
+        <param index="5" label="Req Param 5">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0/NaN).</param>
+        <param index="6" label="Req Param 6">The use of this parameter (if any), must be defined in the requested message. By default assumed not used (0/NaN).</param>
```
hamishwillee, line 2048: @peterbarker The reserved value of a command has to be set when you define the message or it defaults to 0 in all cases - https://mavlink.io/en/guide/define_xml_element.html#reserved. I can be set to 0 or NaN, but it can't be both.

The reserved value is required, whether specified or implied, because you need to know when you define a meaning that the "do nothing" value will be what is sent by default by existing systems. So for all of these it has to be zero, because no reserved value was set. 

I think we could in theory allow NaN as a default value for 3/4, because we're not actually giving this a meaning yet. Risky though.

hamishwillee:
@peterbarker Sorry for the delay on this one. I merged my clarification that the sequence of indexes might be 0 or 1-based, otherwise it is the same.

PR #2154, opened 2024-09-19
common.xml: GIMBAL_MANAGER_FLAGS display as bitmask

pv42:
Similar to #2035

> At some point we decided why put the bitmask entry on the user, when it is a property of the enum. So we have <enum
> name="GIMBAL_MANAGER_FLAGS" bitmask="true">.
> That said, we're inconsistent on this and we should decide if we want one or the other or both.

I personally think it should be explicitly stated in both places for consitency with the existing code as not specifying it breaks some code that depends on these definition files (I am coming from [mavlink/rust-mavlink](https://github.com/mavlink/rust-mavlink/)).

Since however the `ILLUMINATOR_STATUS` message merged in #2047 uses the none-`bitmask` enum `ILLUMINATOR_MODE` as both `bitmask=true` and without it I can see that specifying this should probably be moved to the message field definition consistenly in the future for all existing enums.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -7171,7 +7171,7 @@
       <description>High level message to control a gimbal's attitude. This message is to be sent to the gimbal manager (e.g. from a ground station). Angles and rates can be set to NaN according to use case.</description>
       <field type="uint8_t" name="target_system">System ID</field>
       <field type="uint8_t" name="target_component">Component ID</field>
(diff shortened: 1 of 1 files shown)


hamishwillee (approved):
I think it should probably be in both places too. Thanks.

@julianoes You OK with this?

julianoes (approved):

julianoes:
Thansk!

Issue #2155, opened 2024-09-26
TERRAIN_REPORT?


enok71:
I'm trying to understand the terrain protocol described in https://mavlink.io/en/services/terrain.html. I'm starting to think there is some sort of confusion in either the functionality or in the documentation:
 
The lat, lon position in a TERRAIN_REPORT is supposed to match that in a preceeding TERRAIN_CHECK if a TERRAIN_CHECK triggered it. I.e. the lat, lon is in general not the position of the drone. The terrain_height must be the terrain height at the lat, lon coordinates (not at the drone's position). On the other hand "current_height" is clearly the drone's altitude. But over what? Over the terrain_height at lat, lon? Or over the terrain_height at the drone's current position?

TERRAIN_REPORT can also be streaming back after a TERRAIN_REQUEST in parallel with the TERRAIN_DATA flow. But what lat, lon positions are then intended to be transmitted? Something that should match the position of tiles transmitted in TERRAIN_DATA messages I guess (if the purpose is to help debug the TERRAIN_DATA flow). Still very unclear though. What lat, lon, terrain_height and current_height are emitted in this case?

It seems that ardupilot streams TERRAIN_REPORT messages for a different reason, namely as a small message containing the drone's current altitude over ground. The lat, lon and terrain_height is then always those of at the drone's position. However, this use of TERRAIN_REPORT is not in accordance with the documentation at https://mavlink.io/en/services/terrain.html. Seems like a third way of using TERRAIN_REPORT not described in https://mavlink.io/en/services/terrain.html?

What is the right way of using TERRAIN_REPORT according to you? Should the ardupilot way of using it be described or at least mentioned in your documentation?

hamishwillee:
G'day, been on holiday sorry.

The information is based on this discussion https://github.com/mavlink/mavlink-devguide/pull/259#issuecomment-651316489. 
It is possible, and even likely that parts of it are wrong or inconsistent with what a flight stack might be doing.
As you say it is important to capture both expected behaviour and differences. 

> The lat, lon position in a TERRAIN_REPORT is supposed to match that in a preceding TERRAIN_CHECK if a TERRAIN_CHECK triggered it. I.e. the lat, lon is in general not the position of the drone. The terrain_height must be the terrain height at the lat, lon coordinates (not at the drone's position). On the other hand "current_height" is clearly the drone's altitude. But over what? Over the terrain_height at lat, lon? Or over the terrain_height at the drone's current position?

[...]

PR #2157, opened 2024-10-03
TERRAIN_REPORT clarifications

hamishwillee:
This updates `TERRAIN_REPORT` with clarifications falling out of the discussion in https://github.com/mavlink/mavlink/issues/2155

Specifically, it makes it clear that:
- current_height is always relative to the tile at the vehicle position.
- When streamed, lat, lon, terrain_height, current_height are optional , but if specified are relative to the vehicle, not some other arbitrary tile. This is because when streaming the only thing actually important "generally" is the pending/loaded fields.
- When requested, they are  the values of the specified tile, except for current_height. NOTE, pending/loaded should always be supplied and match the previous/last TERRAIN_REQUEST.

@peterbarker Can you check this is correct - a few minor confirmations inline.

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -6425,14 +6425,23 @@
       <field type="int32_t" name="lon" units="degE7">Longitude</field>
     </message>
     <message id="136" name="TERRAIN_REPORT">
-      <description>Streamed from drone to report progress of terrain map download (initiated by TERRAIN_REQUEST), or sent as a response to a TERRAIN_CHECK request. See terrain protocol docs: https://mavlink.io/en/services/terrain.html</description>
-      <field type="int32_t" name="lat" units="degE7">Latitude</field>
-      <field type="int32_t" name="lon" units="degE7">Longitude</field>
-      <field type="uint16_t" name="spacing">grid spacing (zero if terrain at this location unavailable)</field>
-      <field type="float" name="terrain_height" units="m">Terrain height MSL</field>
-      <field type="float" name="current_height" units="m">Current vehicle height above lat/lon terrain height</field>
-      <field type="uint16_t" name="pending">Number of 4x4 terrain blocks waiting to be received or read from disk</field>
-      <field type="uint16_t" name="loaded">Number of 4x4 terrain blocks in memory</field>
+      <description>Report progress of terrain map download (initiated by TERRAIN_REQUEST), and/or confirm that a particular tile is stored in memory (as a response to a TERRAIN_CHECK request).
+
+        The message should be streamed from the drone in order to report how much of the tile map specified in the previous TERRAIN_REQUEST has been downloaded, using the pending and loaded fields.
+        When streamed, the lat, lon, terrain_height, and current_height fields are optional, but if supplied must reflect the tile at the current vehicle position.
+
+        The message should be emitted in response to a TERRAIN_CHECK request in order to confirm that a particular tile at a lat/lon is present in UAV memory.
+        In this case the lat, lon, and terrain_height reflect the values for the requested tile, or NaN if the tile is not present in UAV memory.
+        Note that the current_height is always relative to the terrain height at the current vehicle location (not at the specified tile).
+
+        See terrain protocol docs: https://mavlink.io/en/services/terrain.html.</description>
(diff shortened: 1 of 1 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -6425,14 +6425,23 @@
...
+
+        The message should be streamed from the drone in order to report how much of the tile map specified in the previous TERRAIN_REQUEST has been downloaded, using the pending and loaded fields.
+        When streamed, the lat, lon, terrain_height, and current_height fields are optional, but if supplied must reflect the tile at the current vehicle position.
+
+        The message should be emitted in response to a TERRAIN_CHECK request in order to confirm that a particular tile at a lat/lon is present in UAV memory.
+        In this case the lat, lon, and terrain_height reflect the values for the requested tile, or NaN if the tile is not present in UAV memory.
+        Note that the current_height is always relative to the terrain height at the current vehicle location (not at the specified tile).
+
+        See terrain protocol docs: https://mavlink.io/en/services/terrain.html.</description>
+      <field type="int32_t" name="lat" units="degE7">Latitude. When streamed, of the tile at the current vehicle position. When requested, of the tile specified in the corresponding TERRAIN_CHECK.</field>
```
hamishwillee, line 6438: Note, "of the tile" for the current vehicle position. My understanding is that a tile has a position and this is what we report. NOT the GPS coordinates of the vehicle when streaming. 

@peterbarker Can you confirm - we're not streaming the _vehicle_ position from GPS but the tile position that corresponds?

message_definitions/v1.0/common.xml
```diff
@@ -6425,14 +6425,23 @@
...
+      <field type="int32_t" name="lon" units="degE7">Longitude. When streamed, of the tile at the current vehicle position. When requested, of the tile specified in the corresponding TERRAIN_CHECK.</field>
+      <field type="uint16_t" name="spacing" invalid="0">Grid spacing. 0 if terrain at this location unavailable.</field>
+      <field type="float" name="terrain_height" units="m" invalid="NaN">Terrain height MSL. NaN if terrain at this location unavailable.</field>
+      <field type="float" name="current_height" units="m" invalid="NaN">Current vehicle height above ground relative to lat/lon terrain height. NaN if height not supplied.</field>
+      <field type="uint16_t" name="pending">Number of 4x4 terrain blocks waiting to be received or read from disk. The set of all terrain blocks is defined by the previous TERRAIN_REQUEST.</field>
+      <field type="uint16_t" name="loaded">Number of 4x4 terrain blocks in memory. The set of all terrain blocks is defined by the previous TERRAIN_REQUEST.</field>
```
hamishwillee, line 6444: Note, neither of these are optional. I am expecting you'd supply these in both check and in streaming.

message_definitions/v1.0/common.xml
```diff
@@ -6425,14 +6425,23 @@
...
+
+        The message should be emitted in response to a TERRAIN_CHECK request in order to confirm that a particular tile at a lat/lon is present in UAV memory.
+        In this case the lat, lon, and terrain_height reflect the values for the requested tile, or NaN if the tile is not present in UAV memory.
+        Note that the current_height is always relative to the terrain height at the current vehicle location (not at the specified tile).
+
+        See terrain protocol docs: https://mavlink.io/en/services/terrain.html.</description>
+      <field type="int32_t" name="lat" units="degE7">Latitude. When streamed, of the tile at the current vehicle position. When requested, of the tile specified in the corresponding TERRAIN_CHECK.</field>
+      <field type="int32_t" name="lon" units="degE7">Longitude. When streamed, of the tile at the current vehicle position. When requested, of the tile specified in the corresponding TERRAIN_CHECK.</field>
+      <field type="uint16_t" name="spacing" invalid="0">Grid spacing. 0 if terrain at this location unavailable.</field>
```
hamishwillee, line 6440: I assume this is optional when streaming.

hamishwillee:
@peterbarker Just pushing review of this back to your attention. I know it has only been a week.

hamishwillee:
@peterbarker Just pushing review of this back to your attention.

hamishwillee:
We should also note that the altitude reference of the data in messages is documented to be MSL, but the ardupilot server (anecdotally) uses EGM96, and the PX4 server uses some other reference-  EGM2008 or similar. 
QGC appears to just grab the data and send it to the recipient  - which implies that the recipient must match the data.
That's not ideal, but important to document.

Issue #2098, opened 2024-03-18
Could we have some kind of sub-protocol to handle multi-GCS configurations in a graceful manner?


Davidsastresas:
I have seen this situation recursively. Multi GCS configurations where the "request control" functionality is needed.

Right now, if the autopilot supports it by a parameter or similar, like Ardupilot does, we can do workarounds playing with GCS-GCS custom mavlink messaging. But we don't have any mavlink standard for this, and I think we could have, it is a nice feature.

I am thinking about something like:
- A new message, broadcasted by the UAV, indicating which sysid is in control of the UAV. This message could be broadcasted at a very small rate, as long as the autopilot sends a new one as soon as control changes. Besides for the protocol itself, this would be good for displaying on a GCS if it is currently in control or not of the vehicle it is connected to.
- A GCS wanting to get control, would get the above message, extract the system id in control, and send a command requesting control.
- The GCS currently in control would receive the message from the other GCS, and could display a popup for the user informing of the other GCS wanting control, and it could approve or decline it.
- In case the GCS in control accepts the control request, it sends a new MAV_CMD to the autopilot, to command the change.
- Autopilot would broadcast immediately the message in point 1, so both GCS would be updated of the change, and could indicate it to the user.

All the above is inspired by how we work with it on gimbal protocol v2 and GIMBAL_MANAGER_STATUS.

Would something like this be valuable? I would like to hear your thoughts @hamishwillee @julianoes

hamishwillee:
Thanks @Davidsastresas 

This has come up before, both for multi GCS control of components (e.g. separate flight and gimbal control) and also for multi GCS observers.

There is a (closed) proposal here https://github.com/mavlink/mavlink/pull/1954 (discussion rambled off-topic but the core bits may be useful). I can't remember the detail, but I think it was conceptually similar. It was closed by the author and I don't remember why.

I think it is a reasonable use case - if you need it and want to provide an implementation then very happy to discuss/take this further.

Davidsastresas:
Thank you @hamishwillee, very interesting. I read that proposal. There were some very interesting topics in that discussion. I think I miss the point at the ending, as you guys reference something you discussed in the dev call, and I have no details about it.

From what I understood from that thread, the way they presented it on its totality would mean a bunch of other changes in the mavlink protocol in general. This is not necessarily bad of course, some very valid points were mentioned there, and they seemed to have spent a bunch of time studying the situation.

[...]

PR #2158, opened 2024-10-08
development.xml: First prototype for multi-GCS operation.

Davidsastresas:
Implement basic functionality for a MAV to broadcast information about the current MAVLink component (usually a GCS) in control of it and a new command to request this control. This follows the discussion in https://github.com/mavlink/mavlink/issues/2098, specifically the latest comments @hamishwillee suggested in https://github.com/mavlink/mavlink/issues/2098#issuecomment-2360036137 about the "cannot be taken over bit", as it makes more sense than the previous approach suggested.

### Description       
A controlled vehicle should only accept MAVLink commands (or command-like messages) that are sent by the controlling GCS, or from other components with the same system id. Commands from other system should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
        
The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission. The autopilot should grant control to the GCS if the system does not require takeover permission or is uncontrolled and ACK the request with MAV_RESULT_ACCEPTED. The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.

If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED. The autopilot should then send this same command to the current owning GCS in order to notify of the request.

The owning GCS would then ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
Note that the pilots of both GCS should co-ordinate safe handover offline.
           
It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended). In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot). The flow is otherwise the same.

#### New messages:
- CONTROL_STATUS: Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL:

    - sysid_in_control: System ID of GCS MAVLink component in control (0: no one in control)
    - takeover_allowed: 1: Requests for control will be granted. 0: Requests for control will be rejected, but the controlling GCS will be notified (and may release control or allow takeover).
   
#### New commands:
- MAV_CMD_REQUEST_OPERATOR_CONTROL: Request control of a system.

    - param 1: Sysid requesting control - System ID of GCS requesting control. 0 when command sent from GCS to autopilot (autopilot determines requesting GCS sysid from message header). Sysid of GCS requesting control when command sent by autopilot to controlling GCS.
    - param 2: Action - 0: Release control, 1: Request control.
    - param 3: Allow takeover - Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.

##### Diagram 1 - 2 GCS, one of them asks for control without enforcing being asked before other GCS takes over (MAV_CMD_REQUEST_OPERATOR_CONTROL param 3 = 1):
```mermaid
sequenceDiagram;
    participant GCS - 253
    participant Vehicle - 1
    participant GCS - 250
    note over Vehicle - 1: Broadcast CONTROL_STATUS: Nobody in control <br> - Sysid In control: 0
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 0
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 0 
    Note over GCS - 253: GCS 253 wants to request control <br> not enforcing asking for takeover
    GCS - 253->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover)
    Vehicle - 1->>GCS - 253: COMMAND_ACK
    Note over Vehicle - 1: Control is transfered to GCS 253
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 1 (takeover allowed automatically)
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 1 (takeover allowed automatically)
    Note over GCS - 253: It is now in control of vehicle 1
    Note over GCS - 250: Now GCS 250 wants to request control. <br> As GCS 253 didn't enforce being asked, <br> it will be granted automatically.
    GCS - 250->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover, although irrelevant right now)
    Vehicle - 1->>GCS - 250: COMMAND_ACK
    Note over Vehicle - 1: As asking for takeover wasn't enforced by GCS 253, <br> control is granted automatically.
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 250 <br> - takeover_allowed: 1 (takeover allowed automatically)
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 250 <br> - takeover_allowed: 1 (takeover allowed automatically)
```

##### Diagram 2 - Now one of them asks for control enforcing being asked before other GCS takes over. When the other GCS takes over, the original GCS in control just requests again control without enforcing being asked before. This situation is interesting so the vehicle is never with no GCS in control. An alternative approach is in diagram 3.
```mermaid
sequenceDiagram;
    participant GCS - 253
    participant Vehicle - 1
    participant GCS - 250
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 0
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 0 
    Note over GCS - 253: GCS 253 wants to request control <br> ENFORCING asking for takeover first
    GCS - 253->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 0 (Do not allow automatic takeover, ask first)
    Vehicle - 1->>GCS - 253: COMMAND_ACK
    Note over Vehicle - 1: Control is transfered to GCS 253
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 0 (automatic takeover not allowed, ask first)
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 0 (automatic takeover not allowed, ask first)
    Note over GCS - 253: It is now in control of vehicle 1
    Note over GCS - 250: Now GCS 250 wants to request control
    GCS - 250->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover, although irrelevant at this point)
    Vehicle - 1->>GCS - 250: COMMAND_ACK
    Vehicle - 1->>GCS - 253: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 250 (Sysid requesting control) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover, although irrelevant at this point)
    Note over GCS - 253: Popup appears in GCS comunicating the request from GCS 250. <br>User accepts. Requesting control again without enforcint being asked for takeover
    GCS - 253->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover)
    Vehicle - 1->>GCS - 253: COMMAND_ACK
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 1 (automatic takeover allowed)
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 1 (automatic takeover allowed)
    Note over GCS - 250: As vehicle is reporting now automatic takeover <br> is allowed, we are free to request control  
    GCS - 250->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 0 (do not allow automatic takeover, although irrelevant at this point)
    Vehicle - 1->>GCS - 250: COMMAND_ACK
    Note over Vehicle - 1: Control will be granted automatically now
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 250 <br> - takeover_allowed: 1 (allow automatic takeover)
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 250 <br> - takeover_allowed: 1 (allow automatic takeover)
```
##### Diagram 3 - Now one of them asks for control enforcing being asked before other GCS takes over. When the other GCS takes over, the original GCS in control just releases control. This might be desirable in some situations, but leaves the vehicle with no owner until the second GCS requests control:
```mermaid
sequenceDiagram;
    participant GCS - 253
    participant Vehicle - 1
    participant GCS - 250
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 0
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 0 
    Note over GCS - 253: GCS 253 wants to request control <br> ENFORCING asking for takeover first
    GCS - 253->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 0 (Do not allow automatic takeover, ask first)
    Vehicle - 1->>GCS - 253: COMMAND_ACK
    Note over Vehicle - 1: Control is transfered to GCS 253
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 0 (automatic takeover not allowed, ask first)
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 253 <br> - takeover_allowed: 0 (automatic takeover not allowed, ask first)
    Note over GCS - 253: It is now in control of vehicle 1
    Note over GCS - 250: Now GCS 250 wants to request control
    GCS - 250->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover, although irrelevant at this point)
    Vehicle - 1->>GCS - 250: COMMAND_ACK
    Vehicle - 1->>GCS - 253: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 250 (Sysid requesting control) <br> - param 2: 1 (request control) <br> - param 3: 1 (allow automatic takeover, although irrelevant at this point)
    Note over GCS - 253: Popup appears in GCS comunicating the request from GCS 250. <br>User accepts. Releasing control
    GCS - 253->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 0 (release control) <br> - param 3: 1 (allow automatic takeover, although it doesn't apply as we drop control)
    Vehicle - 1->>GCS - 253: COMMAND_ACK
    Note over Vehicle - 1: Nobody in control now
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 0 <br> - takeover_allowed: 1 (automatic takeover allowed)
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 0 <br> - takeover_allowed: 1 (automatic takeover allowed)
    Note over GCS - 250: As vehicle is reporting now nobody in control <br> We are free to request control  
    GCS - 250->>Vehicle - 1: MAV_CMD_REQUEST_OPERATOR_CONTROL: <br> - param 1: 0 (Sysid requesting control not needed, can be deducted from sender) <br> - param 2: 1 (request control) <br> - param 3: 0 (do not allow automatic takeover, although irrelevant at this point)
    Vehicle - 1->>GCS - 250: COMMAND_ACK
    Note over Vehicle - 1: Control will be granted automatically now, as <br> nobody is in control
    Vehicle - 1->>GCS - 253: CONTROL_STATUS <br> - sysid_in_control: 250 <br> - takeover_allowed: 1 (allow automatic takeover)
    Vehicle - 1->>GCS - 250: CONTROL_STATUS <br> - sysid_in_control: 250 <br> - takeover_allowed: 1 (allow automatic takeover)
```

We could expand and add more diagrams for the documentation, but I wanted to double check this looks good first.

For awareness @rmackay9 @julianoes.

Thanks!

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2860,6 +2860,9 @@
       <entry value="9" name="MAV_RESULT_COMMAND_UNSUPPORTED_MAV_FRAME">
         <description>Command is invalid because a frame is required and the specified frame is not supported.</description>
       </entry>
+      <entry value="10" name="MAV_RESULT_PERMISSION_DENIED">
+        <description>Command is valid, but the sender is not authorized to control this MAV component. Sender should use MAV_CMD_REQUEST_OPERATOR_CONTROL accordingly to get control, more details in this command's description. (WIP).</description>
+      </entry>
     </enum>
     <enum name="MAV_MISSION_RESULT">
       <description>Result of mission operation (in a MISSION_ACK message).</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3245,6 +3245,10 @@
         <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
         </description>
       </entry>
+      <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_ACCEPTS_GCS_CONTROL">
+        <wip/>
+        <description>This component sets its controlling GCS from a CONTROL_STATUS with the flag GCS_CONTROL_STATUS_FLAGS_SYSTEM_GCS set. Note that if a component can also be separately controlled by a GCS it will stream CONTROL_STATUS.</description>
+      </entry>
     </enum>
     <enum name="MAV_MISSION_TYPE">
       <description>Type of mission items being requested/sent in mission protocol.</description>
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2861,7 +2861,8 @@
         <description>Command is invalid because a frame is required and the specified frame is not supported.</description>
       </entry>
       <entry value="10" name="MAV_RESULT_PERMISSION_DENIED">
-        <description>Command is valid, but the sender is not authorized to control this MAV component. Sender should use MAV_CMD_REQUEST_OPERATOR_CONTROL accordingly to get control, more details in this command's description. (WIP).</description>
+        <wip/>
+        <description>Sender is not authorized to control this MAV component. Control may be requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
       </entry>
     </enum>
     <enum name="MAV_MISSION_RESULT">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3248,7 +3248,7 @@
       </entry>
       <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_ACCEPTS_GCS_CONTROL">
         <wip/>
-        <description>Component supports GCS control (via MAV_CMD_REQUEST_OPERATOR_CONTROL).</description>
+        <description>Component supports locking control to a particular GCS independent of its system (via MAV_CMD_REQUEST_OPERATOR_CONTROL).</description>
       </entry>
     </enum>
     <enum name="MAV_MISSION_TYPE">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -3248,7 +3248,7 @@
       </entry>
       <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_ACCEPTS_GCS_CONTROL">
         <wip/>
-        <description>This component sets its controlling GCS from a CONTROL_STATUS with the flag GCS_CONTROL_STATUS_FLAGS_SYSTEM_GCS set. Note that if a component can also be separately controlled by a GCS it will stream CONTROL_STATUS.</description>
+        <description>Component supports GCS control (via MAV_CMD_REQUEST_OPERATOR_CONTROL).</description>
       </entry>
     </enum>
     <enum name="MAV_MISSION_TYPE">
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -406,7 +406,7 @@
            In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
            The flow is otherwise the same.
         </description>
-        <param index="1" label="sysid requesting control">System ID MAVLink component requesting control (Usually a GCS).</param>
+        <param index="1" label="Sysid requesting control">System ID of GCS requesting control.</param>
         <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
         <param index="3" label="control action">0: Release control, 1: Request control.</param>
         <param index="4" label="allow_takeover">Allow other systems to take control of the MAV for which we are requesting control automatically, or enforce asking us first (0: don't allow takeover, ask current owner first, 1: allow takeover automatically) </param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -389,23 +389,34 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="43005" name="MAV_CMD_REQUEST_OPERATOR_CONTROL" hasLocation="false" isDestination="false">
-        <description>Request control of a system.
-        
-          A controlled vehicle should only accept MAVLink commands (or command-like messages) that are sent by the controlling GCS, or from other components with the same system id. Commands from other system should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
+        <description>Request GCS control of a system (or of a specific component in a system).
 
-          The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
-          It is important to remember targeting a particular component, and avoiding using 0 targetting (all components). Only components emitting CONTROL_STATUS should be targeted, usually these will be autopilots, but could be other components like onboard computers.
-          The autopilot should grant control to the GCS if the system does not require takeover permission (or is uncontrolled) and ACK the request with MAV_RESULT_ACCEPTED.
-          The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.
+          A controlled system should only accept MAVLink commands and command-like messages that are sent by its controlling GCS, or from other components with the same system id.
+          Commands from other systems should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
+          Command-like messages should be ignored (or rejected if that is supported by their associated protocol).
 
-          If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
-          The autopilot should then send this same command to the current owning GCS in order to notify of the request.
-          The owning GCS would ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
+          GCS control of the whole system is managed via a single component that we will refer to here as the "system manager component".
+          This component streams the CONTROL_STATUS message and sets the GCS_CONTROL_STATUS_FLAGS_SYSTEM_MANAGER flag.
+          Other components in the system should monitor for the CONTROL_STATUS message with this flag, and set their controlling GCS to match its published system id.
+          A GCS that wants to control the system should also monitor for the same message and flag, and address the MAV_CMD_REQUEST_OPERATOR_CONTROL to its component id.
+          Note that integrators are required to ensure that there is only one system manager component in the system (i.e. one component emitting the message with GCS_CONTROL_STATUS_FLAGS_SYSTEM_MANAGER set).
+   
+          The MAV_CMD_REQUEST_OPERATOR_CONTROL command is sent by a GCS to the system manager component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
+
+          The system manager component should grant control to the GCS if the system does not require takeover permission (or is uncontrolled) and ACK the request with MAV_RESULT_ACCEPTED.
+          The system manager component should then stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own controlling GCS to match that of the system manager component.
+
+          If the system manager component cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
+          The system manager component should then send this same command to the current owning GCS in order to notify of the request.
+          The owning GCS would ACK with MAV_RESULT_ACCEPTED, and might choose to release control of the vehicle, or re-request control with the takeover bit set to allow permission. 
           Note that the pilots of both GCS should co-ordinate safe handover offline.
-
-          It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
-          In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
-          The flow is otherwise the same.
+          
+          Note that in most systems the only controlled component will be the "system manager component", and that will be the autopilot.
+          However separate GCS control of a particular component is also permitted, if supported by the component.
+          In this case the GCS will address MAV_CMD_REQUEST_OPERATOR_CONTROL to the specific component it wants to control.
+          The component will then stream CONTROL_STATUS for its controlling GCS (it must not set GCS_CONTROL_STATUS_FLAGS_SYSTEM_MANAGER).
+          The component should fall back to the system GCS (if any) when it is not directly controlled, and may stop emitting CONTROL_STATUS.
+          The flow is otherwise the same as for requesting control over the whole system.
         </description>
         <param index="1" label="Sysid requesting control">System ID of GCS requesting control. 0 when command sent from GCS to autopilot (autopilot determines requesting GCS sysid from message header). Sysid of GCS requesting control when command sent by autopilot to controlling GCS.</param>
         <param index="2" label="Action">0: Release control, 1: Request control.</param>
@@ -418,8 +429,8 @@
     </enum>
     <enum name="GCS_CONTROL_STATUS_FLAGS" bitmask="true">
       <description>CONTROL_STATUS flags.</description>
-      <entry value="1" name="GCS_CONTROL_STATUS_FLAGS_MASTER_COMPID">
-        <description>When a GCS sends MAV_CMD_REQUEST_OPERATOR_CONTROL to acquire control of a whole system it needs to address the component emitting this flag. Should be set by only one component in the system, normally an autopilot.</description>
+      <entry value="1" name="GCS_CONTROL_STATUS_FLAGS_SYSTEM_MANAGER">
+        <description>If set, this CONTROL_STATUS publishes the controlling GCS for the whole system. If unset, the CONTROL_STATUS indicates the controlling GCS for just the component emitting the message. Note that to request control of the system a GCS should send MAV_CMD_REQUEST_OPERATOR_CONTROL to the component emitting CONTROL_STATUS with this flag set.</description>
       </entry>
       <entry value="2" name="GCS_CONTROL_STATUS_FLAGS_TAKEOVER_ALLOWED">
         <description>Takeover allowed (requests for control will be granted). If not set requests for control will be rejected, but the controlling GCS will be notified (and may release control or allow takeover).</description>
@@ -778,8 +789,8 @@
     </message>
     <message id="512" name="CONTROL_STATUS">
       <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
-      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
-      <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this component defines the default GCS controller for the whole system</field>
+      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no GCS in control).</field>
+      <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this message instance defines the default controlling GCS for the whole system.</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -768,7 +768,7 @@
     </message>
     <message id="512" name="CONTROL_STATUS">
       <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
-      <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
+      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
       <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
     </message>
   </messages>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -406,7 +406,7 @@
           In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
           The flow is otherwise the same.
         </description>
-        <param index="1" label="Sysid requesting control">System ID of GCS requesting control. It is only needed when an autopilot is forwarding the control request to the current GCS in control. Otherwise, from GCS to autopilot, it should be 0 as it can be deducted from the sender system ID</param>
+        <param index="1" label="Sysid requesting control">System ID of GCS requesting control. 0 when command sent from GCS to autopilot (autopilot determines requesting GCS sysid from message header). Sysid of GCS requesting control when command sent by autopilot to controlling GCS.</param>
         <param index="2" label="Action">0: Release control, 1: Request control.</param>
         <param index="3" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
         <param index="4">Empty</param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -409,7 +409,7 @@
         <param index="1" label="Sysid requesting control">System ID of GCS requesting control.</param>
         <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
         <param index="3" label="Action">0: Release control, 1: Request control.</param>
-        <param index="4" label="allow_takeover">Allow other systems to take control of the MAV for which we are requesting control automatically, or enforce asking us first (0: don't allow takeover, ask current owner first, 1: allow takeover automatically) </param>
+        <param index="4" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -770,7 +770,7 @@
     <message id="512" name="CONTROL_STATUS">
       <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
       <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
-      <field type="uint8_t" name="takeover_allowed">1: Requests for control will be granted. 0: Requests for control will be rejected, but the controlling GCS will be notified (and may release control or allow takeover).</field>
+      <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this component defines the default GCS controller for the whole system</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -767,7 +767,7 @@
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
     <message id="512" name="CONTROL_STATUS">
-      <description>Current status about the MAVLink component in control of this MAV. It is meant to be broadcasted at a low rate (1 hz or less), and also wehenever the fields on this message changes</description>
+      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
       <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
       <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -768,7 +768,6 @@
     </message>
     <message id="512" name="CONTROL_STATUS">
       <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
-      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
       <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
       <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
     </message>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -388,6 +388,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="43005" name="MAV_CMD_REQUEST_OPERATOR_CONTROL" hasLocation="false" isDestination="false">
+        <description>Sent by a MAVLink component (usually GCS) to a vehicle to attempt to request control of it, or sent by a vehicle to the MAVLink component in control to ask for permission to grant control to a different MAVLink component.</description>
+        <param index="1" label="sysid requesting control">System ID MAVLink component requesting control (Usually a GCS).</param>
+        <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
+        <param index="3" label="control action">0: Release control, 1: Request control.</param>
+        <param index="4" label="allow_takeover">Allow other systems to take control of the MAV for which we are requesting control automatically, or enforce asking us first (0: don't allow takeover, ask current owner first, 1: allow takeover automatically) </param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
     </enum>
     <enum name="TARGET_ABSOLUTE_SENSOR_CAPABILITY_FLAGS" bitmask="true">
       <description>These flags indicate the sensor reporting capabilities for TARGET_ABSOLUTE.</description>
@@ -739,6 +749,12 @@
       <field type="float[4]" name="q_target">Quaternion of the target's orientation from the target's frame to the TARGET_OBS_FRAME (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
+    </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Current status about the MAVLink component in control of this MAV. It is meant to be broadcasted at a low rate (1 hz or less), and also wehenever the fields on this message changes</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
+      <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
     </message>
   </messages>
 </mavlink>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -407,9 +407,9 @@
            The flow is otherwise the same.
         </description>
         <param index="1" label="Sysid requesting control">System ID of GCS requesting control. It is only needed when an autopilot is forwarding the control request to the current GCS in control. Otherwise, from GCS to autopilot, it should be 0 as it can be deducted from the sender system ID</param>
-        <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
-        <param index="3" label="Action">0: Release control, 1: Request control.</param>
-        <param index="4" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
+        <param index="2" label="Action">0: Release control, 1: Request control.</param>
+        <param index="3" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
+        <param index="4">Empty</param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
         <param index="7">Empty</param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -406,7 +406,7 @@
            In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
            The flow is otherwise the same.
         </description>
-        <param index="1" label="Sysid requesting control">System ID of GCS requesting control.</param>
+        <param index="1" label="Sysid requesting control">System ID of GCS requesting control. It is only needed when an autopilot is forwarding the control request to the current GCS in control. Otherwise, from GCS to autopilot, it should be 0 as it can be deducted from the sender system ID</param>
         <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
         <param index="3" label="Action">0: Release control, 1: Request control.</param>
         <param index="4" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -392,19 +392,19 @@
         <description>Request control of a system.
         
           A controlled vehicle should only accept MAVLink commands (or command-like messages) that are sent by the controlling GCS, or from other components with the same system id. Commands from other system should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
-        
-           The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
-           The autopilot should grant control to the GCS the system does not require takeover permission or is uncontrolled and ACK the request with MAV_RESULT_ACCEPTED.
-           The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.
+
+          The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
+          The autopilot should grant control to the GCS if the system does not require takeover permission (or is uncontrolled) and ACK the request with MAV_RESULT_ACCEPTED.
+          The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.
 
           If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
           The autopilot should then send this same command to the current owning GCS in order to notify of the request.
-          The owning GCS would then ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
+          The owning GCS would ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
           Note that the pilots of both GCS should co-ordinate safe handover offline.
-           
-           It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
-           In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
-           The flow is otherwise the same.
+
+          It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
+          In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
+          The flow is otherwise the same.
         </description>
         <param index="1" label="Sysid requesting control">System ID of GCS requesting control. It is only needed when an autopilot is forwarding the control request to the current GCS in control. Otherwise, from GCS to autopilot, it should be 0 as it can be deducted from the sender system ID</param>
         <param index="2" label="Action">0: Release control, 1: Request control.</param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -767,7 +767,7 @@
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
     <message id="512" name="CONTROL_STATUS">
-      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
+      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
       <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
       <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
     </message>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -394,6 +394,7 @@
           A controlled vehicle should only accept MAVLink commands (or command-like messages) that are sent by the controlling GCS, or from other components with the same system id. Commands from other system should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
 
           The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
+          It is important to remember targeting a particular component, and avoiding using 0 targetting (all components). Only components emitting CONTROL_STATUS should be targeted, usually these will be autopilots, but could be other components like onboard computers.
           The autopilot should grant control to the GCS if the system does not require takeover permission (or is uncontrolled) and ACK the request with MAV_RESULT_ACCEPTED.
           The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.
 
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -408,7 +408,7 @@
         </description>
         <param index="1" label="Sysid requesting control">System ID of GCS requesting control.</param>
         <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
-        <param index="3" label="control action">0: Release control, 1: Request control.</param>
+        <param index="3" label="Action">0: Release control, 1: Request control.</param>
         <param index="4" label="allow_takeover">Allow other systems to take control of the MAV for which we are requesting control automatically, or enforce asking us first (0: don't allow takeover, ask current owner first, 1: allow takeover automatically) </param>
         <param index="5">Empty</param>
         <param index="6">Empty</param>
--- a/message_definitions/v1.0/development.xml
+++ b/message_definitions/v1.0/development.xml
@@ -389,7 +389,23 @@
         <param index="7">Empty</param>
       </entry>
       <entry value="43005" name="MAV_CMD_REQUEST_OPERATOR_CONTROL" hasLocation="false" isDestination="false">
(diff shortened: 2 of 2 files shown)


hamishwillee:
Thanks very much @Davidsastresas ! This looks as we discussed. I'll raise it in the mav call this evening and also try get some eyes on this.

Note that in the second case the owning pilot could alternatively just release control, leaving the vehicle in a position with no owner. That might even be a more clear flow.

A few comments for readers to capture a few additional interesting points from https://github.com/mavlink/mavlink/issues/2098:

- A main role `CONTROL_STATUS` role is to advertise the owner of the system so that _components_ can determine who their owner is. This allows them to self-configure what ground station can send commands.
- This implementation explicitly does not require a passkey - security, if required, should be another layer.
- A ground station can release control without it being taken over. A drone can decide it has no owner if the original owner is disconnected. 
- The model is based around the idea that 
  - in some cases the pilots are co-located, and it is reasonable they just lean over to each other and suggest control is about to be grabbed - in which case each GCS that requests control is granted it immediately. Note that the GCS can know if such a request will probably succeed from the `CONTROL_STATUS`.
  - Pilots in different locations might want a GCS prompt that control is wanted. You could do that offline somehow since the control status tells you who the owner is. But this model allows you to be prompted when control is desired so you can release it when ready.
- There is an earlier set of messages (not commands) for this purpose: [CHANGE_OPERATOR_CONTROL](https://mavlink.io/en/messages/common.html#CHANGE_OPERATOR_CONTROL), [CHANGE_OPERATOR_CONTROL_ACK](https://mavlink.io/en/messages/common.html#CHANGE_OPERATOR_CONTROL_ACK)
  - No docs but these allow a GCS to set and release control, and have a passkey. Ownership appears to always be granted if the passkey is correct, and their is no "discussion mechanism". 
  - Very much appears to be designed for multiple GCS control over a single drone that is considered to have no components.

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +750,11 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Current status about the MAVLink component in control of this MAV. It is meant to be broadcasted at a low rate (1 hz or less), and also wehenever the fields on this message changes</description>
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
+      <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
+      <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permision before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
```
julianoes, line 757:
```suggestion
      <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="43005" name="MAV_CMD_REQUEST_OPERATOR_CONTROL" hasLocation="false" isDestination="false">
+        <description>Sent by a MAVLink component (usually GCS) to a vehicle to attempt to request control of it, or sent by a vehicle to the MAVLink component in control to ask for permission to grant control to a different MAVLink component.</description>
+        <param index="1" label="sysid requesting control">System ID MAVLink component requesting control (Usually a GCS).</param>
+        <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
```
julianoes, line 394: In my thinking I would add component ID as well in case you have e.g.
- a ground station consisting of both QGC and MAVSDK,
- or two MAVLink pieces running on a companion, e.g. one for a camera and one for some mission planning.

That way you **can** set it specifically if you want to, or just set compid to 0 for all.

hamishwillee: @julianoes 
1. @julianoes Are you talking about adding a component id for the controller or for the target, or for both? 
    FWIW 
   - we decided operator might control a camera separately from the vehicle, but they wouldn't control the companion separately from the drone.
   - I don't consider onboard companion computer control to be "an operator". I think it is up to the system to resolve those levels of control. If we can do it without breaking stuff then sure.
   - we decided that this was about handover from one GCS to another and shared control wasn't needed. Worth revisiting this use case
2. If we add component ids here, we will also need to add them to the CONTROL_STATUS
3. All components in a system monitor the autopilot component of their current system for CONTROL_STATUS and accept commands/command-like messages from that system.
   - If we allow for multiple controllers, we would modify this such that a specifically controlled component would have to emit  a separate CONTROL_STATUS showing its owner. It would have to monitor for its owner and and if it loses and owner it should revert to the main autopilot controller, if any - right?
   - When we discussed this, the assumption was that the separately controlled component would have its own sysid. I think it still all works if we allow things to be set to zero.
5. This currently assumes complete handover of control from one GCS to another, not shared control. 
   - A use case was say a mavlink-enabled joystick not connected to your GCS. This is an odd one because it might have the same system id as the vehicle, or be represented as a GCS. 
6. The thinking we had originally was around having different controllers for a particular component. The idea was that by default all components would monitor the autopilot component 

One point of clarification is that  that system components still need to accept commands from other parts of the system "by default". Otherwise a mavlink camera would ignore mission commands from the autopilot, and your companion computer would not be able to control the vehicle because a GCS already has control. IF you had a standalone "mavlink joystick" it would not be able to control the vehicle unless it was part of the system.
Further, this means we will need to make sure that the autopilots explicitly lock operator control to mean "from a GCS, not from a particular system id or system id/component id combination".

Upshot, I think that it is better to add at least the support for target component id now if we can, but we need to look at use cases with flows showing who each component thinks is its owner in each case.  Let's see what David says.

julianoes: 1. Yes, that's what I mean.

> Otherwise a mavlink camera would ignore mission commands from the autopilot

I was assuming this was for flight controls only. Cameras and gimbals would be separate (gimbal already has its own "in control" protocol.

> and your companion computer would not be able to control the vehicle because a GCS already has control.

That would have been my assumption.

hamishwillee: Yes, so with this anything from the same system id should always be obeyed by default IMO.

"1" was not a yes, no answer. It was a) component id for target, component id for controller or both. I think that you're suggesting component id for the target so that we can have different GCS controlling different parts of the system.

julianoes:
Right, so then can we specify what APIs exactly are determined by this protocol? Is it MANUAL_CONTROL? Is it also RC_CHANNELS_OVERRIDE? Is it sending commands where to go? If it is about commands, then a companion absolutely also able to send things like DO_REPOSITION. And then there is offboard/guided as you bring up. Sorry, more questions than answers.

message_definitions/v1.0/development.xml
hamishwillee: @Davidsastresas Re https://github.com/mavlink/mavlink/pull/2158#issuecomment-2405087871 

> If I am not mistaken, we are talking just about adding support for component ids on systems in control right? 

This is my question above ^^^. I interpreted it to mean "the system being controlled".  

Why? 
- Because I don't consider the companion computer to be an operator, and I don't think there should ever be a case where the GCS is not the ultimate thing in control. IMO If you're same system id as the the target you can send commands and they should be obeyed and it is up to the companion and the autopilot to work out their own control mechanism. 
- The case where a GCS has multiple components is by definition a system. You still have the same operator in the same location. We only need a manager if the operators are separate. If they need separate control we could just give the parts separate system ids. If they had the same system id then I still wouldn't do it this way - I'd add a component id to https://mavlink.io/en/messages/common.html#MANUAL_CONTROL and have the different parts send messages to different targets.
- So for me it is a complication that does not appear to have a valid use case.

However I can see the use for being able to separately control different components. Artificial, but perhaps we have a drone with separate "thingies" that can benefit from independent operator control and that control might come from a different GCS. In this case having a component id to specify the particular target to be controlled by GCS 1 might be  useful.

OF course it might just be that I understand how the above case could work, and I don't see how you could safely manage the multiple controlling component case _and_ have components in the same system talking to each other.

Upshot, we need to decide if this is for operator control or a generic ownership model at the component level. I'd prefer it to be the former. But lets see what @julianoes says.

hamishwillee:
> Right, so then can we specify what APIs exactly are determined by this protocol? Is it MANUAL_CONTROL? Is it also RC_CHANNELS_OVERRIDE? Is it sending commands where to go? If it is about commands, then a companion absolutely also able to send things like DO_REPOSITION. And then there is offboard/guided as you bring up. Sorry, more questions than answers.

What makes sense to me is that a component will accept commands only from its own system and from the owning system (and owning component if we support that) and reject commands from any other system id (and component). In other words it does not matter what command is sent. Further, it should also reject any command-like message, such as MANUAL_CONTROL that is not from the owning system/component.. 

A companion is in the same system so all its commands are obeyed, all the time unless a separate mechanism is put in place. Ditto commands to a mavlink camera or gimbal - an autopilot or companion can always control them, but the GCS cannot.
IF we're going to do anything else you'll have to answer those "more questions".

Davidsastresas:
> IF we're going to do anything else you'll have to answer those "more questions".

I think that is the key. If we want to substantially add more complexity versus what it is already presented, we will have to address a bunch of those scenarios, and ultimately it will delay a lot adoption.

I personally think that going ahead with a minimalist solution will see adoption much quicker. I also think we can afford to be mistaken right now as we are planning this over develompent.xml.

I see this evolving in the future to support all the functionality related with components id we were talking about, it would be great, but overall I am pretty aligned with @hamishwillee in https://github.com/mavlink/mavlink/pull/2158#discussion_r1796136970.

So, if @julianoes is fine with it, I would propose to close this as it is, or maybe add component id just for the GCS ( system in control ) part, as that would not add a lot of extra complexity.

julianoes:
Ok, that sounds fine @Davidsastresas and @hamishwillee. Sorry for blocking this, let's move onwards with what you suggested!

Davidsastresas:
Great, thanks Julian!

@hamishwillee, how should we proceed, are we happy with the current definition? Should I proceed with a PR to mavlink docs? And if so, should I start a new section for this? I don't think it quite fits in non of the existing ones.

Thanks!

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,16 @@
         <param index="6">Empty</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="43005" name="MAV_CMD_REQUEST_OPERATOR_CONTROL" hasLocation="false" isDestination="false">
+        <description>Sent by a MAVLink component (usually GCS) to a vehicle to attempt to request control of it, or sent by a vehicle to the MAVLink component in control to ask for permission to grant control to a different MAVLink component.</description>
```
hamishwillee, line 392: Let's capture the changes here and then merge into development. 

Note that it also suggests new result MAV_RESULT_PERMISSION_DENIED  to indicate that you can't send a system commands, because you don't have ownership.
Also clarifies what control actually means.

This adds the concept of a separate component being controlled (but not separate components in control). 

Is this a good summary?

```suggestion
        <description>Request control of a system.
        
          A controlled vehicle should only accept MAVLink commands (or command-like messages) that are sent by the controlling GCS, or from other components with the same system id. Commands from other system should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
        
           The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
           The autopilot should grant control to the GCS the system does not require takeover permission or is uncontrolled and ACK the request with MAV_RESULT_ACCEPTED.
           The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.

          If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
          The autopilot should then send this same command to the current owning GCS in order to notify of the request.
          The owning GCS would then ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
          Note that the pilots of both GCS should co-ordinate safe handover offline.
           
           It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
           In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
           The flow is otherwise the same.
        </description>
```

hamishwillee: Note, we can always have more docs in the guide, but we should be fairly specific here too where we can.

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,16 @@
...
+        <param index="1" label="sysid requesting control">System ID MAVLink component requesting control (Usually a GCS).</param>
```
hamishwillee, line 393:
```suggestion
        <param index="1" label="Sysid requesting control">System ID of GCS component requesting control (Usually a GCS).</param>
```

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,16 @@
...
+        <param index="2" label="sysid control requested">System ID MAVLink component for which control is being requested (Usually a vehicle).</param>
+        <param index="3" label="control action">0: Release control, 1: Request control.</param>
```
hamishwillee, line 395:
```suggestion
        <param index="3" label="Action">0: Release control, 1: Request control.</param>
```

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,16 @@
...
+        <param index="4" label="allow_takeover">Allow other systems to take control of the MAV for which we are requesting control automatically, or enforce asking us first (0: don't allow takeover, ask current owner first, 1: allow takeover automatically) </param>
```
hamishwillee, line 396:
```suggestion
        <param index="4" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
```

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,16 @@
...
+        <param index="5">Empty</param>
```
hamishwillee, line 397: This could be target component id, but I don't think needed for same reasons as https://github.com/mavlink/mavlink/pull/2158/files#r1802192636 ?

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +750,11 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Current status about the MAVLink component in control of this MAV. It is meant to be broadcasted at a low rate (1 hz or less), and also wehenever the fields on this message changes</description>
```
hamishwillee, line 754:
```suggestion
      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
```

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +750,11 @@
...
+      <field type="uint32_t" name="time_boot_ms" units="ms">Timestamp (time since system boot).</field>
```
hamishwillee, line 755: Why is this useful?

hamishwillee:
> @hamishwillee, how should we proceed, are we happy with the current definition? Should I proceed with a PR to mavlink docs? And if so, should I start a new section for this? I don't think it quite fits in non of the existing ones.

@Davidsastresas As we are broadly in agreement I gave the message a detail review. I think we can do away with the "target ids" and the timestamp - thoughts. Look carefully at those changes and accept them if they make sense. If not, we iterate.

Yes, a docs PR would be good and yes, that could be a new protocol under here: https://mavlink.io/en/services/

I think @auturgy thinks this is broadly reasonable and so it can go into development.xml once we are happy. But we still need broad buy-in from ardupilot etc. Have you chatted to Tridge and PeterB?

message_definitions/v1.0/development.xml
Davidsastresas: This looks great, super clear, thank you!

message_definitions/v1.0/development.xml
Davidsastresas: Thank you. Indeed we can leave it at 0, mavlink truncation will make this message slightly more compact because of that right? I pushed a new commit with this fix.

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
Davidsastresas: You are totally right, we can remove this one. Thanks!

message_definitions/v1.0/development.xml
Davidsastresas: That's right, as we are removing Target Sysid this should not be needed either.

Davidsastresas:
Thanks for the comments @hamishwillee. I agree with all your suggestions, so I committed all of them plus:

- Removing timestamp in CONTROL_STATUS
- Sysid requesting control only needed for autopilot->GCS, when forwarding the control request. Otherwise it can be deducted from sender and must be 0.
- Sysid control requested isn't needed, it can be deducted in all scenarios, so we are removing that one too.

I also updated the first comment in this PR, as well as the diagrams accordingly.

Thanks!

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,32 @@
...
+
+          If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
+          The autopilot should then send this same command to the current owning GCS in order to notify of the request.
+          The owning GCS would then ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
+          Note that the pilots of both GCS should co-ordinate safe handover offline.
+           
+           It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
+           In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
+           The flow is otherwise the same.
+        </description>
```
hamishwillee, line 408: Fix indentation (sorry!). Note that this recommends targeting the specific autopilot or other component id. You can use 0 but that would take control over all ids. It is not generally a good idea to target all components with a command because ACK handling is messy.

```suggestion
          A controlled vehicle should only accept MAVLink commands (or command-like messages) that are sent by the controlling GCS, or from other components with the same system id. Commands from other system should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).

          The command is sent by a GCS to the autopilot component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.
          The autopilot should grant control to the GCS if the system does not require takeover permission (or is uncontrolled) and ACK the request with MAV_RESULT_ACCEPTED.
          The autopilot should stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own owner to match that of the autopilot.

          If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
          The autopilot should then send this same command to the current owning GCS in order to notify of the request.
          The owning GCS would ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
          Note that the pilots of both GCS should co-ordinate safe handover offline.

          It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
          In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
          The flow is otherwise the same.
        </description>
```

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,32 @@
...
+        <param index="1" label="Sysid requesting control">System ID of GCS requesting control. It is only needed when an autopilot is forwarding the control request to the current GCS in control. Otherwise, from GCS to autopilot, it should be 0 as it can be deducted from the sender system ID</param>
```
hamishwillee, line 409: Typo "deducted". I have modified a little to my taste.

```suggestion
        <param index="1" label="Sysid requesting control">System ID of GCS requesting control. 0 when command sent from GCS to autopilot (autopilot determines requesting GCS sysid from message header). Sysid of GCS requesting control when command sent by autopilot to controlling GCS.</param>
```

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +766,10 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
+      <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
```
hamishwillee, line 771:
```suggestion
      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +766,10 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change.</description>
```
hamishwillee, line 770: This improves cross linking

```suggestion
      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
```

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +766,10 @@
...
+      <field type="uint8_t" name="sysid_in_control">System ID of MAVLink component in control (0: no one in control)</field>
+      <field type="uint8_t" name="takeover_allowed">If current MAVLink component in control requires asking permission before taking control of the MAV emitting this message (0: takeover not allowed, ask current owner first, 1: takeover allowed automatically)</field>
```
hamishwillee, line 772: Simplified?
```suggestion
      <field type="uint8_t" name="takeover_allowed">1: Requests for control will be granted. 0: Requests for control will be rejected, but the controlling GCS will be notified (and may release control or allow takeover).</field>
```

hamishwillee:
Thanks @Davidsastresas Looking good! Some tweaks added as suggestions (FYI for some reason your repo does not allow me to accept these myself).

Can you update minimal.xml `MAV_RESULT` with `MAV_RESULT_PERMISSION_DENIED`. This is likely to be a little controversial (since it impacts "common"). But I think we need it.

timtuxworth:
>   * Unless our companion computer is using manual_control or rc_override ( which usually isn't a good idea for a companion computer ), the only way it will control our vehicle is in guided mode, sending guided rate or position commands, correct? In which case, we would need to expand this whole control idea to those guided commands as well in the corresponding autopilots, so they are not accepted from a Mavlink component not in control.

This is definitely not a valid assumption. To do anything useful the companion computer has to be able to send a range of different MavLink commands.

hamishwillee:
> > * Unless our companion computer is using manual_control or rc_override ( which usually isn't a good idea for a companion computer ), the only way it will control our vehicle is in guided mode, sending guided rate or position commands, correct? In which case, we would need to expand this whole control idea to those guided commands as well in the corresponding autopilots, so they are not accepted from a Mavlink component not in control.
> 
> This is definitely not a valid assumption. To do anything useful the companion computer has to be able to send a range of different MAVLink commands.

@timtuxworth True. The current design makes it explicit that the controller we're talking about here is an external system. A controlled system will accept commands from any component in the same system (e.g. companion computer) but only commands from the particular external system that is in control (GCS).

timtuxworth:
> > > * Unless our companion computer is using manual_control or rc_override ( which usually isn't a good idea for a companion computer ), the only way it will control our vehicle is in guided mode, sending guided rate or position commands, correct? In which case, we would need to expand this whole control idea to those guided commands as well in the corresponding autopilots, so they are not accepted from a Mavlink component not in control.
> > 
> > 
> > This is definitely not a valid assumption. To do anything useful the companion computer has to be able to send a range of different MAVLink commands.
> 
> @timtuxworth True. The current design makes it explicit that the controller we're talking about here is an external system. A controlled system will accept commands from any component in the same system (e.g. companion computer) but only commands from the particular external system that is in control (GCS).

What will happen with a Cube Red? A Cube Red has 2 AP's internally each with their own SYSID_THISMAV, a companion computer in a vehicle with a Cube Red needs to control both APs.

julianoes (approved):

julianoes:
@timtuxworth I would say CubeRed will open all sorts of cans full of worms, so that's probably out of the scope of this PR.

hamishwillee:
> will happen with a Cube Red? A Cube Red has 2 AP's internally each with their own SYSID_THISMAV, a companion computer in a vehicle with a Cube Red needs to control both APs.

As @julianoes says, you're probably screwed. I couldn't begin to work out a solution without a clear model of how the APs, companion computer and other components are supposed to share control over the single entity that is the vehicle.

message_definitions/v1.0/development.xml
Davidsastresas: Thanks. What you say about targeting all sysid/compid with 0 is important. Should we make that explicit in the docs?

message_definitions/v1.0/development.xml
Davidsastresas: Much clearer, thanks!

message_definitions/v1.0/development.xml
Davidsastresas: Committed, thanks!

message_definitions/v1.0/development.xml
Davidsastresas: Committed, good idea, I agree.

Davidsastresas:
I agree with what Hamish and Julian said about cubeRed issue. Wouldn't make more sense to use same sysid for both autopilot cores, but different comp IDs? They both belong to the same vehicle which by definition should mean same system ID. Not sure if we have right now any component that could be used for this, maybe we can have a couple or more component IDs for autopilots to handle such scenarios. In any case, I also think it is out of the scope of this PR, but If that discussion ever appears again I am interested in contributing.

@hamishwillee I don't know why you can not modify this branch. I gave you admin access to the original repo this PR comes from, so you should be able to commit directly now. I will address the docs now. Thanks!

timtuxworth:
> I agree with what Hamish and Julian said about cubeRed issue. Wouldn't make more sense to use same sysid for both autopilot cores, but different comp IDs? They both belong to the same vehicle which by definition should mean same system ID. Not sure if we have right now any component that could be used for this, maybe we can have a couple or more component IDs for autopilots to handle such scenarios. In any case, I also think it is out of the scope of this PR, but If that discussion ever appears again I am interested in contributing.

It does seem to make sense that the two independent (but co-operating) AP's in the Cube Red should have the same SYSID, but a different component Id, but this doesn't seem to be supported by ArduPilot, which doesn't seem to have a configurable component id (is it hard coded to "1"?) - am I right about that @tridge ?

message_definitions/v1.0/development.xml
hamishwillee: Probably. Do you want to propose some text.

hamishwillee:
> It does seem to make sense that the two independent (but co-operating) AP's in the Cube Red should have the same SYSID, but a different component Id, but this doesn't seem to be supported by ArduPilot, which doesn't seem to have a configurable component id (is it hard coded to "1", "200"?) - am I right about that @tridge ?

Just FWIW this is "one of the problems" I was talking about. In theory a component should be identified by `MAV_TYPE` - an autopilot is any of the frame type values, components are their specific type. The MAV_COMPONENT values are supposed to be like a default - i.e. [MAV_COMP_ID_AUTOPILOT1 ](https://mavlink.io/en/messages/common.html#MAV_COMP_ID_AUTOPILOT1) is normally used as this reduces the likelihood of id clashes, but it should be fine to use any id at all if you wanted.

Both ArduPilot and PX4 "mostly" assume they can trust the ID as though it were hard coded.

hamishwillee:
@Davidsastresas @timtuxworth I can see a way forward for this particular issue/protocol assuming:
- the two FCs are independent MAVLink components
- All **other** components in the system can infer their controlling GCS from just one autopilot

The problem is determining how components should work out their owning GCS. At the moment they determine this from the `CONTROL_STATUS` emitted by the autopilot. 

What we could do instead is add a flag to `CONTROL_STATUS` indicating "component in active control" - i.e. this is the CONTROL_STATUS that all other components use to determine their status by default. This would be set by just one component, nominally the autopilot, but it could be a companion computer. 

Everything then works as before - any autopilot that wants to can specify a separate controlling GCS. A companion computer can talk to either AP and visa versa because they are all in the same system.

The integrator will need to make sure that only one of the APs or the Companion is setting the bit at a time. Something in cube red would need to manage that.

timtuxworth:
What happens if the GCS "in command" goes down/loses connection/fails/disconnects unexpectedly for some reason? If the "requires permission" flag is on, I think you get stuck with an autopilot with no GCS, or am I missing something?

I think the AP should be able to release a GCS if it doesn't receive heartbeats or ACKs of some kind for some defined timeout.

If this happens in the middle of a "in command" request (i.e. a GCS requests control, then goes down before sending an ACK), control should remain with the previously active GCS I'm thinking.

hamishwillee:
If the autopilot loses it controlling GCS then it becomes uncontrolled (is the current discussion).

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +766,10 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
+      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
+      <field type="uint8_t" name="takeover_allowed">1: Requests for control will be granted. 0: Requests for control will be rejected, but the controlling GCS will be notified (and may release control or allow takeover).</field>
```
hamishwillee, line 772: @Davidsastresas If we decide to accept the model where the component in control of a system does not have to be a particular autopilot - i.e. it could be one of multiple, or it could be the companion computer then we would have to modify this a little. The change would be to have ONLY ONE `CONTROL_STATUS` advertise as the "master". This would be emitted by default by an autopilot but an integrator would need to manage this. The implication for a flight stack would have a parameter to define whether it is the master or not, default on.

The change then would be to make this a flag bitmask so.

```suggestion
      <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this component defines the default GCS controller for the whole system</field>
```

And we'd have to have an enum like this:

```xml
    <enum name="GCS_CONTROL_STATUS_FLAGS" bitmask="true">
      <description>CONTROL_STATUS flags.</description>
      <entry value="1" name="GCS_CONTROL_STATUS_FLAGS_DEFAULT_GCS"/>
        <description>This message indicates GCS that controls the all components in the system by default. Should be set by only one component in the system, normally an autopilot.</description>
      </entry>
      <entry value="2" name="GCS_CONTROL_STATUS_FLAGS_TAKEOVER">
        <description>Takeover allowed (requests for control will be granted). If not set requests for control will be rejected, but the controlling GCS will be notified (and may release control or allow takeover).</description>
      </entry>
    </enum>
```

This does complicate the implementation a little since 
- an autopilot will now have to decide whether it emits the message or not, and if it does, whether it sets the active flag (it would by default).
- Components in the system will now need to check the bit to decide whether to use the message to determine the GCS in control. I don't see this as much harder than checking to see if the message is from autopilot.
- GCS will have to address its requests to the component that emits the CONTROL_STATUS rather than specifically to an autopilot.

The main benefit is that now you could make the companion computer the main controller, which would be convenient in some setup. Also it supports this multiple FC case.

@julianoes What do you think?

julianoes (approved):

message_definitions/v1.0/development.xml
hamishwillee: @Davidsastresas We discussed this in the dev call and I think I convinced everyone that it is worth making these changes.

Can you update the PR with that (and the new enum) and the command ack result value, with "WIP" on it. I will propose updated text for the description tomorrow.

message_definitions/v1.0/development.xml
Davidsastresas: I did this https://github.com/mavlink/mavlink/pull/2158/commits/9d7d74b449f6f791e06e4cad6e56d3db0f39e3e1.

Davidsastresas:
Thanks for all the work @hamishwillee. I pushed the changes requested, although I modified some of the definitions slightly, especially that new GCS_CONTROL_STATUS_FLAGS_MASTER_COMPID flag, for me something like this made more sense as the explanation but still I am not 100% satisfied with it, do you mind going over it and let me know your thoughts?

When the new changes are confirmed, I will upload new diagrams ( that could be used in the docs too ) for the following situations:
- More than one component emitting CONTROL_STATUS, to represent the use of the new flags.
- Situation where an autopilot looses link ( stops receiving HEARTBEAT ) from the current controlling system, to release control.
- Update the existing ones accordingly with new changes, more notably the new control_status flags.

I will put on hold the docs PR https://github.com/mavlink/mavlink-devguide/pull/556 for a moment until we figure out all of this if that is fine. Thanks!

message_definitions/v1.0/development.xml
Davidsastresas: I addressed all these fronts, but I am leaving this conversation open, in case something additional appears, to keep it a bit tidier.

message_definitions/v1.0/common.xml
```diff
@@ -2860,6 +2860,9 @@
       <entry value="9" name="MAV_RESULT_COMMAND_UNSUPPORTED_MAV_FRAME">
         <description>Command is invalid because a frame is required and the specified frame is not supported.</description>
       </entry>
+      <entry value="10" name="MAV_RESULT_PERMISSION_DENIED">
+        <description>Command is valid, but the sender is not authorized to control this MAV component. Sender should use MAV_CMD_REQUEST_OPERATOR_CONTROL accordingly to get control, more details in this command's description. (WIP).</description>
+      </entry>
```
hamishwillee, line 2865: Simplified and added WIP tag.

I've removed "command is valid" because it might not be. We haven't specified response order, but I'd actually hope that a command would be rejected without checking its validity. We should recommend that the check is done before others in the docs.

```suggestion
      <entry value="10" name="MAV_RESULT_PERMISSION_DENIED">
        <wip/>
        <description>Sender is not authorized to control this MAV component. Control may be requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
      </entry>
```

hamishwillee: @Davidsastresas I do have merge rights! Thank you.

message_definitions/v1.0/development.xml
```diff
@@ -740,5 +776,10 @@
       <field type="float[4]" name="q_sensor">Quaternion of the sensor's orientation from TARGET_OBS_FRAME to vehicle-carried NED. (Ignored if set to (0,0,0,0)) (w, x, y, z order, zero-rotation is 1, 0, 0, 0)</field>
       <field type="uint8_t" name="type" enum="LANDING_TARGET_TYPE">Type of target</field>
     </message>
+    <message id="512" name="CONTROL_STATUS">
+      <description>Information about GCS in control of this MAV. This should be broadcast at low rate (nominally 1 Hz) and emitted when ownership or takeover status change. Control over MAV is requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
+      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no one in control)</field>
+      <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this component defines the default GCS controller for the whole system</field>
```
hamishwillee, line 782: Full stops. Just for consistency with the rest of this PR (note, we're not across the messages)
```suggestion
      <field type="uint8_t" name="sysid_in_control">System ID of GCS MAVLink component in control (0: no GCS in control).</field>
      <field type="uint8_t" name="flags" enum="GCS_CONTROL_STATUS_FLAGS" display="bitmask">Control status. For example, whether takeover is allowed, and whether this message instance defines the default controlling GCS for the whole system.</field>
```

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,42 @@
...
+          Note that the pilots of both GCS should co-ordinate safe handover offline.
+
+          It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
+          In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
+          The flow is otherwise the same.
+        </description>
+        <param index="1" label="Sysid requesting control">System ID of GCS requesting control. 0 when command sent from GCS to autopilot (autopilot determines requesting GCS sysid from message header). Sysid of GCS requesting control when command sent by autopilot to controlling GCS.</param>
+        <param index="2" label="Action">0: Release control, 1: Request control.</param>
+        <param index="3" label="Allow takeover">Enable automatic granting of ownership on request (by default reject request and notify current owner). 0: Ask current owner and reject request, 1: Allow automatic takeover.</param>
```
hamishwillee, line 412: I wonder if there is any benefit in inverting this to "deny takeover" and ditto to the flag. We kind of what the "default" to be "you can takeover if you want". What do you think @Davidsastresas .

message_definitions/v1.0/common.xml
```diff
@@ -3242,6 +3245,10 @@
         <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
         </description>
       </entry>
+      <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_ACCEPTS_GCS_CONTROL">
```
hamishwillee, line 3248: @Davidsastresas @julianoes I added this because you can know if a component can be separately controlled but there is no way to know if it implements this protocol in the normal way (i.e. it will monitor for CONTROL_STATUS and set is owner).

Do we need this? Possibly not since you can't change your behaviour if the component doesn't support the command.

What do you think?

message_definitions/v1.0/development.xml
Davidsastresas: I think that is how it was initially planned in the discussion on the initial issue thread, but if I recall correctly you suggested to do this the other way around, as it is now. I don't remember exactly but I think as it is now it made more sense for some reason. But I am open to change that if needed of course.

message_definitions/v1.0/common.xml
Davidsastresas: It is interesting to have, although we would know anyway if such component emits its own CONTROL_STATUS. I personally don't see an obvious need for this at the moment, but it doesn't hurt either. Having this could help us spot from a GCS if a component with this flag is not sending CONTROL_STATUS, and we could try to request it. Other than that I don't see any further reason now.

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,42 @@
...
+        <param index="4">Empty</param>
+        <param index="5">Empty</param>
+        <param index="6">Empty</param>
+        <param index="7">Empty</param>
+      </entry>
+    </enum>
+    <enum name="GCS_CONTROL_STATUS_FLAGS" bitmask="true">
+      <description>CONTROL_STATUS flags.</description>
+      <entry value="1" name="GCS_CONTROL_STATUS_FLAGS_MASTER_COMPID">
+        <description>When a GCS sends MAV_CMD_REQUEST_OPERATOR_CONTROL to acquire control of a whole system it needs to address the component emitting this flag. Should be set by only one component in the system, normally an autopilot.</description>
```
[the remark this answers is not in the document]
Davidsastresas: Much clearer, thanks!

message_definitions/v1.0/development.xml
```diff
@@ -388,6 +388,42 @@
...
+
+          If the autopilot cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
+          The autopilot should then send this same command to the current owning GCS in order to notify of the request.
+          The owning GCS would ACK with MAV_RESULT_ACCEPTED, and might choose to release ownership of the vehicle, or re-request ownership with the takeover bit set to allow permission. 
+          Note that the pilots of both GCS should co-ordinate safe handover offline.
+
+          It is also possible for a GCS to request control of a specific component rather than the whole system (though this is not recommended).
+          In this case the command should be sent to the specific component, which should also emit CONTROL_STATUS to indicate its owner (this component would ignore the CONTROL_STATUS set by the autopilot).
+          The flow is otherwise the same.
+        </description>
```
hamishwillee, line 409: Very hard to rewrite this to make it non-autopilot centric.

```suggestion
        <description>Request GCS control of a system (or of a specific component in a system).

          A controlled system should only accept MAVLink commands and command-like messages that are sent by its controlling GCS, or from other components with the same system id.
          Commands from other systems should be rejected with MAV_RESULT_PERMISSION_DENIED (except for this command, which may be acknowledged with MAV_RESULT_ACCEPTED if control is granted).
          Command-like messages should be ignored (or rejected if that is supported by their associated protocol).

          GCS control of the whole system is managed via a single component that we will refer to here as the "system manager component".
          This component streams the CONTROL_STATUS message and sets the GCS_CONTROL_STATUS_FLAGS_SYSTEM_GCS flag.
          Other components in the system should monitor for the message with this flag, and set their controlling GCS to match the value published in CONTROL_STATUS.
          A GCS that wants to control the system should also monitor for the same message and flag, and address the MAV_CMD_REQUEST_OPERATOR_CONTROL to its component id.
          Note that integrators are required to ensure that there is only one system manager component in the system (i.e. one component emitting the message with GCS_CONTROL_STATUS_FLAGS_SYSTEM_GCS set).
          Normally this will be the autopilot, but it does not have to be.
       
          Separate GCS control of a particular component is also permitted, if supported by the component.
          The GCS will address MAV_CMD_REQUEST_OPERATOR_CONTROL to the specific component it wants to control.
          The component will then stream CONTROL_STATUS for its controlling GCS (it must not set GCS_CONTROL_STATUS_FLAGS_SYSTEM_GCS).
          The component should fall back to the system GCS (if any) when it is not directly controlled, and may stop emitting CONTROL_STATUS.

          Note that in most systems the only controlled component will be the "system manager component", and that will be the autopilot.
   
          The MAV_CMD_REQUEST_OPERATOR_CONTROL command is sent by a GCS to the system manager component to request or release control of a system, specifying whether subsequent takeover requests from another GCS are automatically granted, or require permission.

          The system manager component should grant control to the GCS if the system does not require takeover permission (or is uncontrolled) and ACK the request with MAV_RESULT_ACCEPTED.
          The system manager component should then stream CONTROL_STATUS indicating its controlling system: all other components with the same system id should monitor this message and set their own controlling GCS to match that of the system manager component.

          If the system manager component cannot grant control (because takeover requires permission), the request should be rejected with MAV_RESULT_PERMISSION_DENIED.
          The system manager component should then send this same command to the current owning GCS in order to notify of the request.
          The owning GCS would ACK with MAV_RESULT_ACCEPTED, and might choose to release control of the vehicle, or re-request control with the takeover bit set to allow permission. 
          Note that the pilots of both GCS should co-ordinate safe handover offline.
        </description>
```

message_definitions/v1.0/common.xml
hamishwillee: @Davidsastresas The flag is/was just "I will read CONTROL_STATUS from the system manager and use it as my component GCS in control".
It isn't "I support direct control, so will emit CONTROL_STATUS". On further reflection it isn't useful.

However it could be useful if we state that a component only emits CONTROL_STATUS when it is actively controlled, and otherwise falls back to not emitting the CONTROL_STATUS and takes the value of the system setting (I've actually said it does that). That way if a thing is not going to be independently controlled we don't have all those message flying around.

Instead you would check which can be controlled directly using the protocol bit. 

Note that the system controller should always emit CONTROL_STATUS

message_definitions/v1.0/common.xml
```diff
@@ -3242,6 +3246,10 @@
...
+        <wip/>
+        <description>This component sets its controlling GCS from a CONTROL_STATUS with the flag GCS_CONTROL_STATUS_FLAGS_SYSTEM_GCS set. Note that if a component can also be separately controlled by a GCS it will stream CONTROL_STATUS.</description>
```
hamishwillee, line 3251:
```suggestion
        <description>Component supports GCS control (via MAV_CMD_REQUEST_OPERATOR_CONTROL).</description>
```

message_definitions/v1.0/common.xml
hamishwillee: Modified to match this in https://github.com/mavlink/mavlink/pull/2158/files#r1814195309

message_definitions/v1.0/common.xml
```diff
@@ -3242,6 +3246,10 @@
         <description>This component implements/is a gimbal manager. This means the GIMBAL_MANAGER_INFORMATION, and other messages can be requested.
         </description>
       </entry>
+      <entry value="524288" name="MAV_PROTOCOL_CAPABILITY_COMPONENT_ACCEPTS_GCS_CONTROL">
+        <wip/>
+        <description>Component supports GCS control (via MAV_CMD_REQUEST_OPERATOR_CONTROL).</description>
```
hamishwillee, line 3251:
```suggestion
        <description>Component supports locking control to a particular GCS control (via MAV_CMD_REQUEST_OPERATOR_CONTROL).</description>
```

message_definitions/v1.0/development.xml
hamishwillee: Let's leave. There may be more feedback!

message_definitions/v1.0/development.xml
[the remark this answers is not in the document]
hamishwillee: @Davidsastresas Glad you like it. I did modify the name again, after you saw this!

hamishwillee:
OK, hands off for now. Lets see what general thinking is.

Davidsastresas:
I think I will start working on a prototype for Ardupilot and QGC. I think most of what we have now makes sense to give it a try. I will keep you all posted. Thanks!

hamishwillee (approved):
We discussed in the MAV call - good enough to prototype! Thanks for your patience.

message_definitions/v1.0/common.xml
```diff
@@ -2860,6 +2860,10 @@
       <entry value="9" name="MAV_RESULT_COMMAND_UNSUPPORTED_MAV_FRAME">
         <description>Command is invalid because a frame is required and the specified frame is not supported.</description>
       </entry>
+      <entry value="10" name="MAV_RESULT_PERMISSION_DENIED">
+        <wip/>
+        <description>Sender is not authorized to control this MAV component. Control may be requested using MAV_CMD_REQUEST_OPERATOR_CONTROL.</description>
+      </entry>
```
julianoes, line 2866: Why not re-use MAV_RESULT_DENIED? It creates quite a bit of churn to have more MAV_RESULTs in things like MAVSDK.

hamishwillee: Because it is wrong "Command is invalid (is supported but has invalid parameters). Retrying same command and parameters will not work."

The command is valid so we'd be providing false information.

We might use [MAV_RESULT_FAILED](https://mavlink.io/en/messages/common.html#MAV_RESULT_FAILED)	Command is valid, but execution has failed. This is used to indicate any non-temporary or unexpected problem, i.e. any problem that must be fixed before the command can succeed/be retried. For example, attempting to write a file when out of memory, attempting to arm when sensors are not calibrated, etc.

Though in this case we would also have to add the chance that the user does not have permission. @Davidsastresas What do you think? It is obviously harder for users to work out what is going on but on the other hand, it might be less churn for everyone.

hamishwillee: We do need to decide, so we can remove this asap if it is not used

julianoes: I'm just against adding more results for new commands because it means everyone handling acks now has deal with that result, even if it's in theory unlikely.

In an ideal world we could have specific errors for each command but that would defeat the point of a generic command/ack protocol.

From the description of RESULT_DENIED,  I can see that it's not right. Instead we should use RESULT_FAILED as @Davidsastresas suggests, or RESULT_TEMPORARILY_REJECTED, I'd say.

Davidsastresas: Great, so I think @hamishwillee and me are open to this and see @julianoes point, so I think we should go forward with removing this new command, and maybe re-evaluate in the future.

The only question is if we use RESULT_FAILED or RESULT_TEMPORARILY_REJECTED. Reading both descriptions, I think RESULT_FAILED is better suited. RESULT_TEMPORARILY_REJECTED seems to me like something that will be alright in short time, without operator input. On the other hand, RESULT_FAILED indicates that something needs to be fixed, requiring some action taken.

I will open a new PR shortly with this and a couple of new changes to keep the discussion. Thank you both!

PR #2161, opened 2024-10-31
Standard modes - move to common.xml

hamishwillee:
Standard modes are now in PX4 v1.15. So no longer in development.

- PX4 is a full-ish implementation of the concept except it omits the vehicle-specific modes. Have asked Beat to do those.
- QGC implements this and you can test this now for PX4 v1.15.
- To test the dynamic modes you have to use daily build and companion computer setup.

Ardupilot PR for same thing: https://github.com/ArduPilot/ardupilot/pull/28566 - 

This moves the relevant messages. 

Corresponding docs work https://github.com/mavlink/mavlink-devguide/pull/557

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -1958,6 +1958,18 @@
         <param index="6" label="Pitch Angle" units="deg" minValue="-180" maxValue="180" default="0">Fixed pitch angle that the camera will hold in oblique mode if the mount is actuated in the pitch axis.</param>
         <param index="7">Empty</param>
       </entry>
+      <entry value="262" name="MAV_CMD_DO_SET_STANDARD_MODE" hasLocation="false" isDestination="false">
+        <description>Enable the specified standard MAVLink mode.
+          If the mode is not supported the vehicle should ACK with MAV_RESULT_FAILED.
+        </description>
+        <param index="1" label="Standard Mode" enum="MAV_STANDARD_MODE">The mode to set.</param>
+        <param index="2" reserved="true" default="0"/>
+        <param index="3" reserved="true" default="0"/>
+        <param index="4" reserved="true" default="0"/>
+        <param index="5" reserved="true" default="0"/>
+        <param index="6" reserved="true" default="0"/>
+        <param index="7" reserved="true" default="NaN"/>
+      </entry>
       <entry value="300" name="MAV_CMD_MISSION_START" hasLocation="false" isDestination="false">
         <description>start running a mission</description>
         <param index="1" label="First Item" minValue="0" increment="1">first_item: the first mission item to run</param>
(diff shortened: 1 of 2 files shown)


message_definitions/v1.0/common.xml
```diff
@@ -7733,6 +7733,36 @@
       <field type="uint16_t" name="sequence_oldest_available">Oldest Sequence number that is still available after the sequence set in REQUEST_EVENT.</field>
       <field type="uint8_t" name="reason" enum="MAV_EVENT_ERROR_REASON">Error reason.</field>
     </message>
+    <message id="435" name="AVAILABLE_MODES">
+      <description>Get information about a particular flight modes.
+        The message can be enumerated or requested for a particular mode using MAV_CMD_REQUEST_MESSAGE.
+        Specify 0 in param2 to request that the message is emitted for all available modes or the specific index for just one mode.
+        The modes must be available/settable for the current vehicle/frame type.
+        Each modes should only be emitted once (even if it is both standard and custom).
```
hamishwillee, line 7741:
```suggestion
        Each modes should only be emitted once (even if it is both standard and custom).
        The current mode should be emitted in CURRENT_MODE.
        AVAILABLE_MODES_MONITOR should be emitted to indicate that the available modes have changed and AVAILABLE_MODES should be re-requested.
```

message_definitions/v1.0/common.xml
```diff
@@ -1400,6 +1400,7 @@
       </entry>
       <entry value="176" name="MAV_CMD_DO_SET_MODE" hasLocation="false" isDestination="false">
         <description>Set system mode.</description>
+        <deprecated since="2024-11" replaced_by="MAV_CMD_DO_SET_STANDARD_MODE">MAV_CMD_DO_SET_STANDARD_MODE allows setting both standard and custom modes</deprecated>
```
hamishwillee, line 1403: Note, we should change all of these to superseded ...

hamishwillee, line 1403:
```suggestion
        <description>Set custom system mode. Use MAV_CMD_DO_SET_STANDARD_MODE to set a standard mode.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -7733,6 +7845,36 @@
...
     </message>
+    <message id="435" name="AVAILABLE_MODES">
+      <description>Get information about a particular flight modes.
+        The message can be enumerated or requested for a particular mode using MAV_CMD_REQUEST_MESSAGE.
+        Specify 0 in param2 to request that the message is emitted for all available modes or the specific index for just one mode.
+        The modes must be available/settable for the current vehicle/frame type.
+        Each modes should only be emitted once (even if it is both standard and custom).
+      </description>
+      <field type="uint8_t" name="number_modes">The total number of available modes for the current vehicle type.</field>
+      <field type="uint8_t" name="mode_index">The current mode index within number_modes, indexed from 1.</field>
```
hamishwillee, line 7856:
```suggestion
      <field type="uint8_t" name="mode_index">The current mode index within number_modes, indexed from 1. The index is not persistent, and may change between reboots or if the set of modes change.</field>
```

message_definitions/v1.0/common.xml
```diff
@@ -7733,6 +7845,36 @@
...
+      <field type="uint8_t" name="standard_mode" enum="MAV_STANDARD_MODE">Standard mode.</field>
+      <field type="uint32_t" name="custom_mode">A bitfield for use for autopilot-specific flags</field>
```
hamishwillee, line 7858: Should this be invalid value `0` ? Is that invalid for a custom mode?

julianoes:
> Standard modes are now in PX4 v1.15. So no longer in development.

> We really need a better process to finalize validation before the releases. Grrr.

While I'm not specifically against standardizing this, and given there was not follow up to change it, I'd say it's probably ready to go in, however I want to make a more general point: 
**If PX4 releases with development.xml that's PX4's problem. We are still allowed to change this, not a MAVLink devs problem.**

auturgy:
"If PX4 releases with development.xml that's PX4's problem. We are still allowed to change this, not a MAVLink devs problem".
Agreed.
The PX4 implementation has very little to do with the intent of standard modes, and is focused on using ROS2 for new non-standard modes.  From a quick inspection it uses 4 of the standard modes - which indicates that we've missed the mark on standard modes.  I think we need to have another look at what we want to achieve with standard modes - which should have nothing to do with new non-standard modes from ROS2 - before this can be brought into common.xml.  I'd be more inclined to back it out completely at this point.

hamishwillee:
> "If PX4 releases with development.xml that's PX4's problem. We are still allowed to change this, not a MAVLink devs problem".

The process is supposed to be a proposal into WIP state which is "largely agreed", prototyping and presentation of the implementation, sign off.
The expectation is that the prototyping will highlight problems with the proposal/design and we iterate the spec, or that it works, and that final sign off is largely a rubber stamp.

So yes, I agree that we should be able to change things if prototyping shows it doesn't work. 

What my (onging but quite extensive) testing shows here is that what we have is something that is working really well but is still incomplete.
- Not all modes are emitted as standard modes
- The AVAILABLE_MODES_MONITOR is not being emitted - I have to follow that up.

> As an aside, you might remember that we don't actually provide tools to allow flight stacks to easy control integration of WIP stuff - and Beat specifically asked for this. For evidence this embarrassment ... https://github.com/ArduPilot/pymavlink/pull/240#top

> The PX4 implementation has very little to do with the intent of standard modes, and is focused on using ROS2 for new non-standard modes.

Irrelevant, as long as it is compliant with the RFC as agreed and accepted.

> From a quick inspection it uses 4 of the standard modes - which indicates that we've missed the mark on standard modes. 

Yes, it implements MAV_STANDARD_MODE_RETURN_HOME, MAV_STANDARD_MODE_MISSION, MAV_STANDARD_MODE_LAND, MAV_STANDARD_MODE_TAKEOFF

It omits MAV_STANDARD_MODE_POSITION_HOLD, MAV_STANDARD_MODE_ORBIT, MAV_STANDARD_MODE_CRUISE, MAV_STANDARD_MODE_ALTITUDE_HOLD, MAV_STANDARD_MODE_SAFE_RECOVERY

I don't think we've missed the mark. I think these need to be implemented. At that point  the APIs become useful. Perhaps ArduPilot might push the long running PR Peter has on this in as well?

We can reasonably request that those be added as a test criterion for taking into common.xml.
The argument being that this prototype is incomplete.
Though, in fairness we haven't _required_ that all modes be implemented, the use case is broken if we do not.

> I think we need to have another look at what we want to achieve with standard modes - which should have nothing to do with new non-standard modes from ROS2 - before this can be brought into common.xml. I'd be more inclined to back it out completely at this point.

Again, what PX4 did for the API is irrelevant as long as it does what we intended. 
It does so modulo missing some modes.

At this point I think you'd have to prove that the design doesn't work to make it worthwhile revisting.

hamishwillee:
In summary, let's discuss in the meeting, but I think we don't have a full prototype. It isn't worth revisiting the design but it is worth pushing for the  missing modes to be added.

hamishwillee:
@auturgy @julianoes I had a chat to @bkueng about this.

1. QGC is generating its UI from the standard modes API in daily (which relies on development.xml). So this has and is being tested by users now. [I am going to play with this]
2. As noted, it only does the "generic modes" currently - this was a first prototype. Beat is going to look into what it would take to add support for the position, hold and cruise modes - and check that PX4 would comply with definitions as they are (no ETA yet).
3. The (optional) monitor message is emitted only after the mode changes and is then streamed at a very low rate. If you're testing against vanilla PX4 this doesn't get emitted. This was discussed but I had forgotten - will update docs appropriately.
4. I've asked Beat if there was anything else in the implementation that was discovered.

So this is getting quite a lot of real world testing.

message_definitions/v1.0/common.xml
```diff
@@ -5136,6 +5148,106 @@
...
+        </description>
+      </entry>
+      <entry value="3" name="MAV_STANDARD_MODE_CRUISE">
+        <description>Cruise mode (manual).
+          Position-controlled and stabilized manual mode.
+          When sticks are released vehicles return to their level-flight orientation and hold their original track against wind and external forces.
+          Fixed-wing (FW) vehicles level orientation and maintain current track and altitude against wind and external forces.
+          Hybrid MC/FW ("VTOL") vehicles first transition to FW mode (if needed) but otherwise behave in the same way as MC vehicles.
+          Multicopter (MC) vehicles must not support this mode.
+          Other vehicle types must not support this mode (this may be revisited through the PR process).
```
hamishwillee, line 5190: This also makes sense for boats and subs. Are we OK with this addition?

```suggestion
          Rovers and boat vehicles maintain current track and altitude against wind and external forces.
          Submarine vehicles level orientation and maintain current track and altitude against external forces.
          Other vehicle types, including multicopters, must not support this mode (this may be revisited through the PR process).
```

Note the restriction on MC might be a bit harsh. There isn't any reason why an MC or car couldn't choose to implement this mode, it just doesn't make much sense because unlike in an FW or boat, you can't fly forward "forever".
On the other hand, until someone wants it, I guess there is no pull.

hamishwillee:
[2022](https://github.com/mavlink/mavlink/wiki/20241120-Dev-Meeting) - This is mostly waiting on a full implementation of the vehicle-specific modes - promised by BeatK at some point. ArduPilot implementation would trigger this too.

As a side-note, the original intent was to make it easy to use standard modes so that ground stations can work better with unknown flight stacks. The current design makes it too easy to use custom modes, so we've somewhat achieved the goal since a GCS can discover modes and somewhat make a common UI, but not as well as if flight stacks do move to custom modes where possible.

hamishwillee:
REminder, we will also need to merge in changes from recovery mode being merged with RTL, if that happens.

auturgy (approved):
As you know, I'm not completely onboard, but let's get this in.

julianoes (approved):

message_definitions/v1.0/common.xml
hamishwillee: Closing this. Not agreed at point of merging into common. If needed, should be added to follow up PR.

message_definitions/v1.0/common.xml
hamishwillee: Accepting this as it is captured in the original spec.

message_definitions/v1.0/common.xml
hamishwillee: Accepting. This is just cross linking to the other relevant message.

message_definitions/v1.0/common.xml
hamishwillee: Probably, but not accepting in this PR - since I don't know if 0 or uintmax could be valid custom modes.

hamishwillee:
Thanks @julianoes and @auturgy - merging.

PR #2162, opened 2024-11-04
add rc_type_crsf

benjinne:
Based on recommendation to add for CRSF binding
https://github.com/PX4/PX4-Autopilot/pull/23294#issuecomment-2303311359

### DIFF
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -2027,8 +2027,8 @@
       </entry>
       <entry value="500" name="MAV_CMD_START_RX_PAIR" hasLocation="false" isDestination="false">
         <description>Starts receiver pairing.</description>
-        <param index="1" label="Spektrum">0:Spektrum.</param>
-        <param index="2" label="RC Type" enum="RC_TYPE">RC type.</param>
+        <param index="1" label="RC Type" enum="RC_TYPE">RC type.</param>
+        <param index="2" label="RC Sub Type" enum="RC_SUB_TYPE">RC sub type.</param>
       </entry>
       <entry value="510" name="MAV_CMD_GET_MESSAGE_INTERVAL" hasLocation="false" isDestination="false">
         <deprecated since="2022-04" replaced_by="MAV_CMD_REQUEST_MESSAGE"/>
@@ -4014,14 +4014,23 @@
     </enum>
     <enum name="RC_TYPE">
       <description>RC type</description>
-      <entry value="0" name="RC_TYPE_SPEKTRUM_DSM2">
+      <entry value="0" name="RC_TYPE_SPEKTRUM">
+        <description>Spektrum</description>
+      </entry>
+      <entry value="1" name="RC_TYPE_CRSF">
+        <description>CRSF</description>
+      </entry>
+    </enum>
+    <enum name="RC_SUB_TYPE">
+      <description>RC type</description>
+      <entry value="0" name="RC_SUB_TYPE_SPEKTRUM_DSM2">
         <description>Spektrum DSM2</description>
       </entry>
-      <entry value="1" name="RC_TYPE_SPEKTRUM_DSMX">
+      <entry value="1" name="RC_SUB_TYPE_SPEKTRUM_DSMX">
         <description>Spektrum DSMX</description>
       </entry>
-      <entry value="2" name="RC_TYPE_CRSF">
-        <description>Crossfire Protocol</description>
+      <entry value="2" name="RC_SUB_TYPE_SPEKTRUM_DSMX8">
+        <description>Spektrum DSMX8</description>
       </entry>
     </enum>
     <enum name="POSITION_TARGET_TYPEMASK" bitmask="true">
--- a/message_definitions/v1.0/common.xml
+++ b/message_definitions/v1.0/common.xml
@@ -4020,6 +4020,9 @@
       <entry value="1" name="RC_TYPE_SPEKTRUM_DSMX">
         <description>Spektrum DSMX</description>
       </entry>
+      <entry value="2" name="RC_TYPE_CRSF">
+        <description>Crossfire Protocol</description>
+      </entry>
     </enum>
     <enum name="POSITION_TARGET_TYPEMASK" bitmask="true">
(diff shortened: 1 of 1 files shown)


hamishwillee:
Thanks @benjinne .

I created a simpler version in https://github.com/mavlink/mavlink/pull/2163 before I saw this.

I think it this might be better.  If it were compatible, I'd just have RC_TYPE with all the types and sub types in param 1

message_definitions/v1.0/common.xml
```diff
@@ -4006,12 +4006,24 @@
...
-      <entry value="0" name="RC_TYPE_SPEKTRUM_DSM2">
+      <entry value="0" name="RC_TYPE_SPEKTRUM">
+        <description>Spektrum</description>
+      </entry>
+      <entry value="1" name="RC_TYPE_CRSF">
+        <description>CRSF</description>
+      </entry>
+    </enum>
+    <enum name="RC_SUB_TYPE">
+      <description>RC type</description>
```
hamishwillee, line 4017: Allows cross linking
```suggestion
      <description>RC sub-type. Used in MAV_CMD_START_RX_PAIR.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -4006,12 +4006,24 @@
     </enum>
     <enum name="RC_TYPE">
       <description>RC type</description>
```
hamishwillee, line 4008:
```suggestion
      <description>RC type. Used in MAV_CMD_START_RX_PAIR.</description>
```

message_definitions/v1.0/common.xml
```diff
@@ -4005,13 +4005,28 @@
...
+    </enum>
+    <enum name="RC_SUB_TYPE">
+      <description>RC sub-type of types defined in RC_TYPE. Used in MAV_CMD_START_RX_PAIR.</description>
+      <entry value="0" name="RC_SUB_TYPE_SPEKTRUM_DSM2">
         <description>Spektrum DSM2</description>
       </entry>
-      <entry value="1" name="RC_TYPE_SPEKTRUM_DSMX">
+      <entry value="0" name="RC_SUB_TYPE_UNSPECIFIED">
+        <description>Sub type not specified (there may be no sub types).</description>
+      </entry>
```
hamishwillee, line 4023: It is good to specify a 0 case for enums, defining the default. Normally we'd assume unknown, none, or all. Ideally we'd want to say "just try it"

benjinne: Thanks, I agree, I used zero as a magic number, but this explains it

hamishwillee (approved):
Thanks @benjinne As per https://github.com/mavlink/mavlink/pull/2163#issuecomment-2460873200 I think this is a good change in that it reflects the way that flight stacks are using the message currently - we can't break that.

@julianoes FYI, I'm going to merge. For all the same radio types as previously the same command params would be set for the same radio. @benjinne Also has a QGC implementation, so if there are issues, that will catch them.

message_definitions/v1.0/common.xml
```diff
@@ -4013,13 +4013,28 @@
...
```
hamishwillee, line 4031: Toolchain not letting me do this as duplicate value 0

```suggestion
```

message_definitions/v1.0/common.xml
```diff
@@ -4013,13 +4013,25 @@
...
         <description>Spektrum DSM2</description>
       </entry>
-      <entry value="1" name="RC_TYPE_SPEKTRUM_DSMX">
+      <entry value="1" name="RC_SUB_TYPE_SPEKTRUM_DSMX">
         <description>Spektrum DSMX</description>
       </entry>
+      <entry value="2" name="RC_SUB_TYPE_SPEKTRUM_DSMX8">
+        <description>Spektrum DSMX8</description>
+      </entry>
     </enum>
```
hamishwillee, line 4035: Either this, or this.
```
      <entry value="3" name="RC_SUB_TYPE_UNSPECIFIED">
        <description>Sub type not specified (there may be no sub types).</description>
      </entry>
```

hamishwillee, line 4035: This is one option for handling the unspecified sub type. If you don't specify then you can set the third value to zero.
What I don't like about this is that it isn't the zero'th value. 


```suggestion
      <entry value="3" name="RC_SUB_TYPE_UNSPECIFIED">
        <description>Sub type not specified (there may be no sub types).</description>
      </entry>
    </enum>
```

message_definitions/v1.0/common.xml
```diff
@@ -4013,13 +4013,25 @@
...
+      <description>RC type. Used in MAV_CMD_START_RX_PAIR.</description>
+      <entry value="0" name="RC_TYPE_SPEKTRUM">
+        <description>Spektrum</description>
+      </entry>
+      <entry value="1" name="RC_TYPE_CRSF">
+        <description>CRSF</description>
+      </entry>
+    </enum>
+    <enum name="RC_SUB_TYPE">
+      <description>RC sub-type of types defined in RC_TYPE. Used in MAV_CMD_START_RX_PAIR.</description>
```
hamishwillee, line 4025:
```suggestion
      <description>RC sub-type of types defined in RC_TYPE. Used in MAV_CMD_START_RX_PAIR. Ignored if value does not correspond to the set RC_TYPE.</description>
```

hamishwillee:
> I fail to see how this was not a breaking change in the spec. It's a good change, and the discussion referenced seems pretty solid.
> 
> If a client was sending a value of 1 in param1 then they used to start a bind for DSMX. After this PR they would start a bind for CRSF.

Er, before this change if they were sending a value of 1 in param1 they were sending an unsupported value:

```
        <param index="1" label="Spektrum">0:Spektrum.</param>
```

But perhaps I am missing something. 

Glad you can fix this on ArduPilot end either way. Thank you. 

I was not certain about this type/subtype either - I wanted the system to abstract and binding stuff so you could just say "bind". I think I was convinced that it would be helpful to do it this way to inform the flight stack of what special code it need to run, and for a sender to be able to know that a particular bind attempt failed.

Certainly open to having a "try anything" enum value.

PR #2165, opened 2024-11-06
VELOCITY_LIMITS - move to common.xml

hamishwillee:
SET_VELOCITY_LIMITS
VELOCITY_LIMITS


hamishwillee:
@MaEtUgR This is the API that supports slow mode. I understand it is in a release. For it to move to common.xml there has to be some way to test and verify this - can I test in QGC?

hamishwillee:
[2022](https://github.com/mavlink/mavlink/wiki/20241120-Dev-Meeting) - this is waiting on @MaEtUgR to confirm the messages are in the desired format and provide validation. My understanding is that they may require more work

